PR #847, opened 2019-07-05
chore: update toolchain version and output created config files

### LINKED ISSUE #743: Upgrade containerd to 1.2.7

### LINKED ISSUE #760: Output created files in `osctl config generate`
We don't output information around what was done or where the newly generated files are, so a user would have no idea that master-{1,2,3}.yaml, worker.yaml, and talosconfig were generated in their pwd. 
We should do better.

rsmitty:
Decided to combine two very small changes (which I'm now grumpy at myself for doing). 

First, we'll update the toolchain image versions to allow for the use of a new containerd and runc. Also updated go.mod and go.sum to make use of newer containerd version. Closes #743 and #744.

Second, I added the bit of logic to `osctl config generate` to determine the working directory and let the user know that we created the various yaml files there. Closes #760.

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -6,6 +6,7 @@ package cmd
 
 import (
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"net"
@@ -131,6 +132,11 @@ var configGenerateCmd = &cobra.Command{
 			helpers.Fatalf("failed to generate PKI and tokens: %v", err)
 		}
 
+		workingDir, err := os.Getwd()
+		if err != nil {
+			helpers.Fatalf("failed to fetch current working dir: %v", err)
+		}
+
 		var udType generate.Type
 		for idx, master := range strings.Split(args[1], ",") {
 			input.Index = idx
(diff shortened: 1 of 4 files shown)


cmd/osctl/cmd/config.go
```diff
@@ -145,12 +151,14 @@ var configGenerateCmd = &cobra.Command{
 			if err = writeUserdata(input, udType, "master-"+strconv.Itoa(idx+1)); err != nil {
 				helpers.Fatalf("failed to generate userdata for %s: %v", "master-"+strconv.Itoa(idx+1), err)
 			}
+			fmt.Println("created file", workingDir+"/master-"+strconv.Itoa(idx+1)+".yaml")
```
smira: should this go to `stderr`?

rsmitty: I didn't _think_ so. I followed what we do with the output for `osctl cluster create`.

andrewrynhard:
@rsmitty In the future we will want to break these into two separate PRs. Since the commit header will go into the changelog, we will also want a short summary of the feature in the header. For example: `feat: upgrade containerd to v1.2.7 and runc to v1.0-rc8`.

PR #854, opened 2019-07-07
docs: Add API2 proposal

bradbeam:


andrewrynhard:
Some notes from today's meeting:

- We need versioned APIs for granular endpoints
- The Composition API is our "public" API
- We need to figure out how this works with authn/authz. For example, should each service do auth or should OSD?

andrewrynhard:
@bradbeam Can you clean this up to make it accurate? It could be useful as architecture documentation for future engineers.

andrewrynhard:
Closing in favor of #1814.

PR #855, opened 2019-07-08
docs: add machine configuration proposal

rsmitty:


docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Abstract
+
+This proposal will outline how we'll handle the passing of machine configurations in Talos, as we're aiming to simplify the options and keep the user from having to worry about the minutiae of kubeadm and all of the options.
+
+## Background
+
+I think the easiest way to background this is to take a look at each of the machine configs that we currently have templates for. When looking at these files, it's somewhat self-explanatory on what is available to tweak, but it also gives a good idea of what is similar between the three types of Talos nodes: init (the first master), control plane (any other masters), and workers. I've also appended some additional fields that we use for certain platforms like Packet.
+
+#### Init:
```
andrewrynhard: I don't think these examples are needed here. It just makes things super verbose. I simple sentence stating that our current config is bigger than need be should be good enough.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+
+Author(s): Spencer Smith (@rsmitty)
+
+## Abstract
+
+This proposal will outline how we'll handle the passing of machine configurations in Talos, as we're aiming to simplify the options and keep the user from having to worry about the minutiae of kubeadm and all of the options.
+
+## Background
+
+I think the easiest way to background this is to take a look at each of the machine configs that we currently have templates for. When looking at these files, it's somewhat self-explanatory on what is available to tweak, but it also gives a good idea of what is similar between the three types of Talos nodes: init (the first master), control plane (any other masters), and workers. I've also appended some additional fields that we use for certain platforms like Packet.
```
andrewrynhard: I think it is worth pointing out that we are going explicitly away from the term "user data" in order to distinguish ourselves from `cloud-init`.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+  root:
+    device: /dev/sda
+    size: 2048000000
+    rootfs: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/rootfs.tar.gz
+  data:
+    device: /dev/sda
+    size: 4096000000
+```
+
+I'm hoping that most kubernetes-specific knobs that need to be tweaked can be done with extra args. The machine and cluster top-level keys are new and correspond to things that need to be configured per-machine, as well as things that need to be set at the cluster level. The install and networking layouts will likely change as other proposals are implemented in Talos v0.2.0 so those are left unchanged for now.
```
andrewrynhard: Aside from the machine config layout, I think I'd like to see something around versioning and we intend on implementing backwards compatibility, config verification, config translation, how far back we will support a config, etc.

andrewrynhard:
Just thought of this. We should like at removing the need for specifying the CNI.

docs/proposals/20190708-MachineConfig.md
rsmitty: will remove control plane and worker configs. keep init.

docs/proposals/20190708-MachineConfig.md
rsmitty: pre-v1.0, we'll support n-1 on versions for machine config.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+      key: "{{ .Certs.OsKey }}"
+  kubernetes:
+    ca:
+      crt: "{{ .Certs.K8sCert }}"
+      key: "{{ .Certs.K8sKey }}"
+services:
+  init:
+    cni: flannel
+  kubeadm:
+    initToken: {{ .InitToken }}
```
Ulexus: Should this even be exposed at all?  This is quite ephemeral (default 24-hours), and doesn't have anything to do with the PKI infrastructure (so far as I know).

andrewrynhard: Since we have no mechanism to tell `kubeadm` about any new tokens, we actually disable the TTL and make the token never expire. Happy to revisit. The thing that makes expiring `kubeadm` tokens difficult is automation. 🤔 I wonder if we can set tokens on the fly with CAPI.

Ulexus: Ultimately, we'll need to revisit the perpetual tokens.  This should be something that trustd can handle:  generating new kubeadm tokens to add new machines (which can be done by the kubeadm CLI tool, so it really _should_ be an API).

For now, though, if it makes it easier, it doesn't really hurt anything.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+    schedulerExtraArgs: {}
+    networking:
+      dnsDomain: {{ .ServiceDomain }}
+      podSubnet: {{ index .PodNet 0 }}
+      serviceSubnet: {{ index .ServiceNet 0 }}
+networking:
+  os:
+    devices:
+      - interface: lo
+        cidr: x.x.x.x/32
```
Ulexus: cidr should definitely be an array, if presented so

docs/proposals/20190708-MachineConfig.md
andrewrynhard: 100% Agree... this should absolutely be an API. So would it be something like the trust token is the single long standing token and then by virtue of the fact that the trustd token is valid, that a machine can request a token?

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+    apiServerExtraArgs: {}
+    controllerManagerExtraArgs: {}
+    schedulerExtraArgs: {}
+    networking:
+      dnsDomain: {{ .ServiceDomain }}
+      podSubnet: {{ index .PodNet 0 }}
+      serviceSubnet: {{ index .ServiceNet 0 }}
+networking:
+  os:
+    devices:
```
Ulexus: I'm not sure that there is a separation between "devices" and "interfaces".  There are certainly interfaces which are not devices, for instance, and there can be devices which are used in multiple interfaces.  Personally, I would just remove "devices," collapsing interfaces down a level.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+  controlPlane:
+    apiServerExtraArgs: {}
+    controllerManagerExtraArgs: {}
+    schedulerExtraArgs: {}
+    networking:
+      dnsDomain: {{ .ServiceDomain }}
+      podSubnet: {{ index .PodNet 0 }}
+      serviceSubnet: {{ index .ServiceNet 0 }}
+networking:
+  os:
```
Ulexus: Need a `routes` field, too.

```
routes:
  - "2001:db8::10:0/64 via 2001:db8::0:2bd9 metric 10"
  - "0.0.0.0/0 via 192.168.24.1"
  - "10.66.254.0/24 via 10.254.254.1 dev tap0"
```

I haven't looked at the API for adding these, though, so they may need to be more structured than the iproute2 commandline format above.

Ulexus: The routes could be better placed (or perhaps additionally, optionally placed) under each interface, as well.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+    devices:
+      - interface: lo
+        cidr: x.x.x.x/32
+      - interface: eth0
+        dhcp: true
+install:
+  wipe: false
+  force: true
+  boot:
+    device: /dev/sda
```
Ulexus: When multiple volume types specify the same device, as in the case here, does the system just understand that the disk should be partitioned out in some way?  If wipe is false and there is an existing partition table, what happens?

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+networking:
+  os:
+    devices:
+      - interface: lo
+        cidr: x.x.x.x/32
+      - interface: eth0
+        dhcp: true
+install:
+  wipe: false
+  force: true
```
Ulexus: What does force mean?

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,293 @@
...
+    size: 1024000000
+    kernel: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/vmlinuz
+    initramfs: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/initramfs.xz
+  root:
+    device: /dev/sda
+    size: 2048000000
+    rootfs: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/rootfs.tar.gz
+  data:
+    device: /dev/sda
+    size: 4096000000
```
Ulexus: If size is not defined (or '0'), will the system use the "rest" of the disk for the data volume?  (Really useful to have this be so)

docs/proposals/20190708-MachineConfig.md
andrewrynhard: This could be named better, but it means that the filesystem will be forcefully formatted, vs a wipe where the block device is zeroed out.

andrewrynhard: This essentially translates into the `-f` flag to `mkfs.xfs`.

docs/proposals/20190708-MachineConfig.md
andrewrynhard: `size` should actually be removed for the data(ephemeral) device because Talos will _always_ resize it to the size of the disk. IMO we should remove this from the data definition.

docs/proposals/20190708-MachineConfig.md
Ulexus: Pretty much, yes.

Down the road, of course, I'd still like even the trustd token to at least have an _optional_ expiry and provide a facility via osctl to generate new ones (to be used in future machine configs).  This is why I wanted to be able to bind in and register machine identifiers, so that we could have ephemeral trustd tokens:  once a machine is registered, its own machine ID acts as a token.  New machines require new tokens, within a deployment window.  That may not work well for cloud environments with AutoScaling (which is why it should be optional), but it's a very nice optional security level.

docs/proposals/20190708-MachineConfig.md
andrewrynhard: We have some intelligence in the system that knows how to partition the block device if `boot` and `data` specify the same device. If wipe is false and there is an existing partition table, Talos will write over that partition table regardless. `wipe` basically indicates that the disk should be zeroed out before partitioning.

docs/proposals/20190708-MachineConfig.md
andrewrynhard: I think I agree with this. 👍

docs/proposals/20190708-MachineConfig.md
andrewrynhard: Agree.

docs/proposals/20190708-MachineConfig.md
Ulexus: In that case, "force" should probably be moved to be a volume-level option, as "wipe" there.

"wipe" here, then becomes ambiguous.  Does that mean to wipe the partition table?  If so, does that mean for each device listed amongst the volume "device"s?

We had discussed the precept that if Talos is on a disk, it "owns" that disk, meaning the disk and all its partitions.  Leaving "wipe" (partitions) (and to a lesser extent the "wipe" (filesystems)) seems to suggest that to not be true.

I'm not saying it's necessarily a problem to include the option, but it makes the logic more complicated.

My default assumption would be to sum up the list of "devices" for whom "wipe" is enabled, determine which are non-partition specifications (/dev/xvda vs /dev/xvda3 or /dev/nvme0n1 vs /dev/nvme0n1p3), and repartition those disks accordingly.

However, even then, it's a bit hairy.  Just the name 'device' leaves open to discussion whether that _should_ be a disk or a partition.  I think I would just plan for that to be "disk" rather than "device," and always expect to "own" the disk by repartitioning it as needed.

Leaving partitions open to the user is just problematic.  What if there is a partition clash?  What if the existing partition table is dos and there are already 4 primary partitions?

CoreOS did this well:  it expected a single disk which would be unilaterally repartitioned into its own schema.  Then, if you wanted to add additional storage volumes, you did so separately, specifying the parameters to do so.

andrewrynhard: Makes sense. These options might be better left out for the "owned" disk, but available for "extra" disks.

docs/proposals/20190708-MachineConfig.md
Ulexus: That still seems to be bit wonky.  Why would you zero data at the top level (and not at the volume level)?  I could understand that from a data security perspective, but that seems to be rather out-of-scope here.

andrewrynhard: That is essentially it, purely for security purposes. It has zero bearing on the Talos functionality so I agree that it is out of scope.

docs/proposals/20190708-MachineConfig.md
Ulexus: Essentially, I'm following the pretty well-thought design of systemd-networkd here.  The name of an interface can be arbitrary.  It can be bound to a device by name (unreliable) or by mac (reliable, but requires the foreknowledge).  Addresses are bound to interfaces.  Routes can be bound to interfaces.

This can then be extended to support more esoteric devices, such as bridges, tunnels, bonded links, etc. by adding the 'type' field and specific requirements of each type. The equivalent of a networkd `.netdev` file should be added as `os.devices` to create new logical devices.

For now, though, this presents a reasonable first round organization.

docs/proposals/20190708-MachineConfig.md
Ulexus: Collapsing this comment in favour of the better-described layout bound to the cidr (line 250) thread.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,204 @@
...
+      crt: "{{ .Certs.K8sCert }}"
+      key: "{{ .Certs.K8sKey }}"
+    token: abc.1234
+## All configs that are specific to this machine only
+## go in machine key
+machine:
+  nodeType: init, controlplane, worker
+  kubeletExtraArgs: {}
+  network:
+    interfaces:
```
rsmitty: Updated with interfaces example provided by @Ulexus, as well as included the routes under the eth1 device as an example.

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,202 @@
...
+## We've removed the size of the "data" partition and renamed it to "ephemeral". Also removed the root partition
+install:
+  wipe: false
+  force: true
+  boot:
+    device: /dev/sda
+    size: 1024000000
+    kernel: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/vmlinuz
+    initramfs: http://139.178.69.21:8080/assets/talos/release/v0.1.0-beta.0/initramfs.xz
+  ephemeral:
```
rsmitty: Renamed to ephemeral and removed size for "data" partition

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,206 @@
...
+  controlPlane:
+    ips: [x.x.x.x, y.y.y.y, z.z.z.z]
+    apiServerExtraArgs: {}
+    controllerManagerExtraArgs: {}
+    schedulerExtraArgs: {}
+    network:
+      dnsDomain: {{ .ServiceDomain }}
+      podSubnet: {{ index .PodNet 0 }}
+      serviceSubnet: {{ index .ServiceNet 0 }}
+    ## An example of how we might allow for override of image locations with local registries
```
rsmitty: Exposed the images we'll use for control plane to allow folks to use their own repos or different versions (though we may want to say that's not really supported but may work)

docs/proposals/20190708-MachineConfig.md
```diff
@@ -0,0 +1,223 @@
...
+    ephemeral:
+      device: /dev/sda
+
+## All cluster-wide configs go here. These are generally things that get 
+## injected to kubeadm, or affect the behavior of the kubernetes control 
+## plane as a whole.
+cluster:
+  clusterName: {{ .ClusterName }}
+  controlPlane:
+    ips:
```
Ulexus: We should also have some mechanism by which we can specify a DNS name or load balancer IP for clients to use.  Perhaps `canonicalAddress` or `advertisedAddress`?  This is important at a minimum for TLS certs (subject alternative names), but also for more advanced network configurations where some internal components need to be able to bootstrap before the service node is available.

Ulexus: Perhaps that should be under `apiServer`?  Or better, `clusterAddress` next to `clusterName`.

PR #856, opened 2019-07-09
chore: Add buildkit service to drone

bradbeam:
This adds a buildkit instance local to the drone build. This should allow
us to get more consistency with our builds by keeping the build context
local to the node.

This also changes the buildkit image store to be backed by memory.
This should have a positive impact on our build times since any important
artifact is exported.

Lastly, this reverts the `-p 1` setting in our test script to let tests run
at full speed.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,6 +15,14 @@ services:
         path: /dev
       - name: tmp
         path: /tmp
+  - name: buildkit
+    image: moby/buildkit:v0.5.1
+    privileged: true
+    volumes:
+      - name: buildkitsock
+        path: /run/buildkit
+      - name: buildkit-storage
+        path: /var/lib/buildkit
 
 steps:
   - name: fetch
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -28,101 +38,127 @@ steps:
     image: autonomy/build-container:latest
     pull: always
     environment:
-      BUILDKIT_HOST: tcp://buildkitd.ci.svc:1234
+      BUILDKIT_HOST: unix:///run/buildkit/buildkitd.sock
```
andrewrynhard: I don’t think we need this. I had a PR the other day that just used the default path. Might be nice to not have to worry about setting this env var.

bradbeam: I set this because our makefile defaults to `tcp://0.0.0.0:1234` to account for the local setup. Something to work through.

.drone.yml
```diff
@@ -28,101 +36,143 @@ steps:
...
       BINDIR: /usr/local/bin
     commands:
       - make lint
+    volumes:
+      - name: buildkitsock
+        path: /run/buildkit
     depends_on:
       - fetch
 
+  # We'll use this step to load the hyperkube image to save
```
andrewrynhard: not a fan of this... this is getting more and more down the path of managing dependencies on our own, which we already got bit by with the rootfs/test.

Dockerfile
```diff
@@ -191,8 +191,10 @@ COPY --from=rootfs-archive /rootfs.tar.gz /rootfs.tar.gz
 # The test target performs tests on the source code.
 
 FROM base AS test
-COPY --from=rootfs-build / /rootfs
-ENV PATH /rootfs/bin:$PATH
+#COPY --from=rootfs-build / /rootfs
```
andrewrynhard: Let's stick with copying the rootfs. The only reason this works right now is because `containerd` is actually not linked against `libseccomp` (mistakenly), but that is fixed with the new `bldr` stuff I have been working on.

.drone.yml
bradbeam: This is to speed up the integration test, not for any other reason. We can kick off the docker pull in parallel with the build steps so when we get to running the integration test we don't need to spend the time waiting on docker pull.

andrewrynhard: Yeah I get why, just adds in yet another place we need to be mindful of when upgrading Kubernetes. I'd rather not have this step. Breaking out every piece is already diverging from the original design to have dependencies managed in on place. I'd rather not go further down that path.

PR #862, opened 2019-07-12
feat: add install flag for extra kernel args

andrewrynhard:
In addition to adding a flag, this adds a field to the user data that allows
for extra kernel arguments to be specified.

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -184,6 +184,14 @@ func (c *cmdline) Append(k string, v string) {
 	insert(&c.Parameters, k, v)
 }
 
+// AppendAll appends a set of kernel parameters.
+func (c *cmdline) AppendAll(args []string) error {
+	parameters := parse(strings.Join(args, " "))
+	c.Parameters = append(c.Parameters, parameters...)
+
+	return nil
+}
+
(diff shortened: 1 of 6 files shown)


internal/pkg/kernel/kernel.go
```diff
@@ -184,6 +185,23 @@ func (c *cmdline) Append(k string, v string) {
...
+func (c *cmdline) AppendAll(args []string) error {
+	for _, arg := range args {
+		parts := strings.Split(arg, "=")
+		switch len(parts) {
+		case 1:
+			c.Append(parts[0], "")
+		case 2:
+			c.Append(parts[0], parts[1])
+		default:
+			return errors.New("expected kernel argument in key=value format")
```
smira: I believe argument might contain `=`, so it's more correct to do `parts := strings.SplitN(arg, "=", 2)` 

and use "" if parts[1] is missing

andrewrynhard: If I understand you, I think we don't need to do this because `c.Append` adds the `=`.

andrewrynhard: In order to support one key with multiple values, we use the follow type:

```
type Parameter struct {
	key    Key
	values []string
}
```
So the `=` doesn't exist until the `String()` method is called.

PR #866, opened 2019-07-13
feat: use new pkgs for initramfs and rootfs

andrewrynhard:
This brings in the newly compiled libraries and binaries from our new
pkg builds. Since we no longer create symlinks for directories that need to be
writable, we leverage overlayfs for very specific paths
(e.g. /etc/kubernetes, /etc/cni). This allows us to keep our base rootfs as
read-only, and then have approved paths in the rootfs that we allow to be
writable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,46 @@
 # syntax = docker/dockerfile:1.1-experimental
 
-ARG KERNEL_IMAGE
-ARG TOOLCHAIN_IMAGE
-ARG ROOTFS_IMAGE
-ARG INITRAMFS_IMAGE
+ARG TOOLS
+FROM $TOOLS AS tools
+ENV PATH /toolchain/bin
+RUN ["/toolchain/bin/mkdir", "/bin", "/tmp"]
+RUN ["/toolchain/bin/ln", "-svf", "/toolchain/bin/bash", "/bin/sh"]
+RUN ["/toolchain/bin/ln", "-svf", "/toolchain/etc/ssl", "/etc/ssl"]
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.16.0
 
-# The proto target generates code from protobuf service definitions.
+# The build target creates a container that will be used to build Talos source
+# code.
 
-ARG TOOLCHAIN_IMAGE
-FROM ${TOOLCHAIN_IMAGE} AS proto-build
-WORKDIR /osd
-COPY ./internal/app/osd/proto ./proto
-RUN protoc -I/usr/local/include -I./proto --go_out=plugins=grpc:proto proto/api.proto
-WORKDIR /trustd
-COPY ./internal/app/trustd/proto ./proto
-RUN protoc -I/usr/local/include -I./proto --go_out=plugins=grpc:proto proto/api.proto
-WORKDIR /init
-COPY ./internal/app/init/proto ./proto
-RUN protoc -I/usr/local/include -I./proto --go_out=plugins=grpc:proto proto/api.proto
-
-FROM scratch AS proto
-COPY --from=proto-build /osd/proto/api.pb.go /internal/app/osd/proto/
-COPY --from=proto-build /trustd/proto/api.pb.go /internal/app/trustd/proto/
-COPY --from=proto-build /init/proto/api.pb.go /internal/app/init/proto/
-
-# The base provides a common image to build the Talos source code.
-
-ARG TOOLCHAIN_IMAGE
-FROM ${TOOLCHAIN_IMAGE} AS base
-ENV GOPATH /toolchain/gopath
-RUN mkdir -p ${GOPATH}
+FROM scratch AS build
+COPY --from=tools / /
+SHELL ["/toolchain/bin/bash", "-c"]
+ENV PATH /toolchain/bin:/toolchain/go/bin
 ENV GO111MODULE on
 ENV GOPROXY https://proxy.golang.org
 ENV CGO_ENABLED 0
 WORKDIR /src
+
+# The generate target generates code from protobuf service definitions.
+
+FROM build AS generate-build
+WORKDIR /osd
+COPY ./internal/app/osd/proto ./proto
+RUN protoc -I./proto --go_out=plugins=grpc:proto proto/api.proto
+WORKDIR /trustd
+COPY ./internal/app/trustd/proto ./proto
+RUN protoc -I./proto --go_out=plugins=grpc:proto proto/api.proto
+WORKDIR /init
+COPY ./internal/app/init/proto ./proto
+RUN protoc -I./proto --go_out=plugins=grpc:proto proto/api.proto
+FROM scratch AS generate
+COPY --from=generate-build /osd/proto/api.pb.go /internal/app/osd/proto/
+COPY --from=generate-build /trustd/proto/api.pb.go /internal/app/trustd/proto/
+COPY --from=generate-build /init/proto/api.pb.go /internal/app/init/proto/
+
+# The base target provides a container that can be used to build all Talos
+# assets.
+
+FROM build AS base
 COPY ./go.mod ./
 COPY ./go.sum ./
 RUN go mod download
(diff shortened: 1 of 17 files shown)


internal/app/init/internal/rootfs/rootfs.go
```diff
@@ -66,6 +66,12 @@ func Prepare(s string, inContainer bool, data *userdata.UserData) (err error) {
 		}
 	}
 
+	// Create required directories that are not part of FHS.
+	for _, p := range []string{"/etc/kubernetes/manifests", "/etc/cni", "/var/lib/kubelet", "/var/log/pods", "/usr/libexec/kubernetes"} {
```
andrewrynhard: I found that we were creating the same well known paths in the `PreFunc` of multiple services. Since we were mistakenly running with a writable rootfs, it went unnoticed that these paths should be created _before_ switch rooting.

internal/app/init/main.go
```diff
@@ -195,6 +195,46 @@ func root() (err error) {
 		if _, err = kmsg("[talos]"); err != nil {
 			return fmt.Errorf("failed to setup logging to /dev/kmsg: %v", err)
 		}
+
+		log.Println("mounting overlay for /etc/kubernetes")
```
andrewrynhard: This is the new way of having writable paths on top of the `rootfs` without the need for symlinks and/or extra partitions.

andrewrynhard:
This mostly works, the only thing I need to figure out is why calico-node fails with:

```
[I] $ kubectl logs -n kube-system calico-node-nvdlf -f
2019-07-13 18:01:49.927 [INFO][8] startup.go 256: Early log level set to info
2019-07-13 18:01:49.927 [INFO][8] startup.go 272: Using NODENAME environment for node name
2019-07-13 18:01:49.927 [INFO][8] startup.go 284: Determined node name: alpha-1-dev-ewr1
2019-07-13 18:01:49.928 [INFO][8] k8s.go 228: Using Calico IPAM
2019-07-13 18:01:49.928 [INFO][8] startup.go 316: Checking datastore connection
2019-07-13 18:02:19.929 [INFO][8] startup.go 331: Hit error connecting to datastore - retry error=Get https://10.96.0.1:443/api/v1/nodes/foo: dial tcp 10.96.0.1:443: i/o timeout
2019-07-13 18:02:50.930 [INFO][8] startup.go 331: Hit error connecting to datastore - retry error=Get https://10.96.0.1:443/api/v1/nodes/foo: dial tcp 10.96.0.1:443: i/o timeout
```

internal/app/init/pkg/system/runner/containerd/containerd_test.go
```diff
@@ -68,7 +68,7 @@ func (suite *ContainerdSuite) SetupSuite() {
 	args := &runner.Args{
 		ID: "containerd",
 		ProcessArgs: []string{
-			"/rootfs/bin/containerd",
```
andrewrynhard: You will see a number of similar changes in this PR. The reason is because we used to add the rootfs to `/rootfs`, but since `containerd` is now linked against `libseccomp` we need to place everything at `/` to make sure that the  search paths are good.

Dockerfile
```diff
@@ -92,134 +109,97 @@ ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
...
+COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
+RUN ["/toolchain/bin/cleanup.sh", "/rootfs"]
+FROM scratch AS rootfs-base
+COPY --from=rootfs-build /rootfs /
+FROM build AS rootfs-archive
+COPY --from=rootfs-base / /rootfs
 WORKDIR /rootfs
-RUN tar -cvpzf /rootfs.tar.gz .
-
+RUN tar -cpzf /rootfs.tar.gz .
```
andrewrynhard: I remove the verbose flags because before I was running `cleanup.sh` there were a lot of source code files and I realized that printing out each one when tarring was slowing down builds dramatically.

hack/cleanup.sh
```diff
@@ -0,0 +1,30 @@
+#!/toolchain/bin/bash
```
andrewrynhard: Add comments to make clear what is going on in this script.

Dockerfile
```diff
@@ -92,134 +109,100 @@ ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
...
-ARG INITRAMFS_IMAGE
-FROM ${INITRAMFS_IMAGE} AS initramfs-build
-WORKDIR /
-COPY --from=init-build /init /init
-
-ARG TOOLCHAIN_IMAGE
-FROM ${TOOLCHAIN_IMAGE} AS initramfs-archive
-COPY --from=initramfs-build / /initramfs
+FROM scratch AS kernel
+COPY --from=docker.io/autonomy/kernel:2ac99a0 /boot/vmlinuz /vmlinuz
```
bradbeam: nit, should modules be included with this?

bradbeam: Also is there much value in having a specific `kernel` target versus using `autonomy/kernel`?

Dockerfile
```diff
@@ -92,134 +109,100 @@ ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
...
+
+FROM tools AS initramfs-build
+COPY --from=docker.io/autonomy/fhs:8467184 / /rootfs
+COPY --from=docker.io/autonomy/ca-certificates:20f39f7 / /rootfs
+COPY --from=docker.io/autonomy/dosfstools:767dee6 / /rootfs
+COPY --from=docker.io/autonomy/musl:9bc7430 / /rootfs
+COPY --from=docker.io/autonomy/syslinux:85e1f9c / /rootfs
+COPY --from=docker.io/autonomy/xfsprogs:5e50579 / /rootfs
+COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
+RUN ["/toolchain/bin/cleanup.sh", "/rootfs"]
```
bradbeam: nit, let's add a newline here to separate `initramfs-build` and `initramfs-base`

Dockerfile
```diff
@@ -92,134 +109,100 @@ ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
...
+FROM scratch AS initramfs-base
+COPY --from=initramfs-build /rootfs /
+COPY --from=init /init /init
+FROM build AS initramfs-archive
```
bradbeam: nit, let's add a newline here to separate `initramfs-base` and `initramfs-archive`

Dockerfile
andrewrynhard: Reason we don't do that is because we use the `local` exporter to drop the `vmlinuz` artifact into `./build`, and add the modules here will mean that they get exported locally too.

andrewrynhard: For the same reason mentioned about, we need a `target` we can hit in order to export the `vmlinuz`.

Dockerfile
```diff
@@ -239,10 +228,25 @@ COPY --from=hashicorp/packer:1.4.2 /bin/packer /bin/packer
...
 ARG TAG
 ENV VERSION ${TAG}
 ENTRYPOINT ["entrypoint.sh"]
+
+# The test target performs tests on the source code.
+
+FROM base AS test
+RUN ["/toolchain/bin/unlink", "/etc/ssl"]
+COPY --from=rootfs-base / /
+RUN ["/toolchain/bin/ln", "-svf", "/toolchain/bin/bash", "/bin/sh"]
```
bradbeam: nit, let's move this above the copy, should limit the changes needed when rootfs changes ( i know, link is super lightweight )

Dockerfile
```diff
@@ -239,10 +228,25 @@ COPY --from=hashicorp/packer:1.4.2 /bin/packer /bin/packer
...
+COPY hack/golang/test.sh /bin
+
+# The lint target performs linting on the source code.
+
+FROM base AS lint
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.16.0
```
bradbeam: maybe as a todo for later is move this step to bldr/pkgs so we can skip out on the install each time

Dockerfile
```diff
@@ -92,134 +109,106 @@ ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
...
-COPY --from=kernel /modules /
-COPY images /usr/images
-
-ARG TOOLCHAIN_IMAGE
-FROM ${TOOLCHAIN_IMAGE} AS rootfs-archive
-COPY --from=rootfs-build / /rootfs
+# The rootfs target provides the Talos rootfs image.
+
+FROM tools AS rootfs-build
+COPY --from=docker.io/autonomy/fhs:8467184 / /rootfs
```
smira: it probably might make sense in the future to use some more human-readable tag names? so that it's easier to track when deps are going to be updated. certainly not for this PR, but as a future todo

internal/app/init/main.go
```diff
@@ -188,13 +189,41 @@ func initram() (err error) {
 	return nil
 }
 
+func createOverlay(path string) error {
+	log.Printf("mounting overlay for %s\n", path)
+
+	parts := strings.Split(path, "/")
+	prefix := strings.Join(parts[1:], "-")
+	diff := fmt.Sprintf("/var/%s-diff", prefix)
+	workdir := fmt.Sprintf("/var/%s-workdir", prefix)
```
bradbeam: Just making note for later, we should consider moving all of these to a specific dir/prefix ( ex, `/var/talos`, `/var/talos-overlays`, ... )

Dockerfile
andrewrynhard: Yeah I'm hoping that we can use SHAs while still in alpha and then cut official releases of the images. There is also probably some more optimization we can do around setting these dynamically using a dependency algorithm in `bldr`. I have some ideas here. Actually I was hoping to chat with you about them.

PR #869, opened 2019-07-15
chore: setup gce for e2e builds

rsmitty:


rsmitty:
Note: don't merge without machined in .drone.yaml

PR #874, opened 2019-07-16
fix: Fix integration of extra kernel args

bradbeam:

### DIFF
--- a/cmd/osctl/cmd/install_linux.go
+++ b/cmd/osctl/cmd/install_linux.go
@@ -91,6 +91,6 @@ func init() {
 	installCmd.Flags().StringVar(&device, "device", "", "The path to the device to install to")
 	installCmd.Flags().StringVar(&endpoint, "userdata", "", "The value of "+constants.KernelParamUserData)
 	installCmd.Flags().StringVar(&platform, "platform", "", "The value of "+constants.KernelParamPlatform)
-	installCmd.Flags().StringSliceVar(&extraKernelArgs, "extra-kernel-arg", []string{}, "Extra argument to pass to the kernel")
+	installCmd.Flags().StringArrayVar(&extraKernelArgs, "extra-kernel-arg", []string{}, "Extra argument to pass to the kernel")
 	rootCmd.AddCommand(installCmd)
 }
--- a/hack/installer/entrypoint.sh
+++ b/hack/installer/entrypoint.sh
@@ -29,7 +29,7 @@ function setup_raw_device(){
 }
 
 function install_talos() {
-  osctl install --bootloader="${WITH_BOOTLOADER}" --device="${DEVICE}" --platform="${TALOS_PLATFORM}" --userdata="${TALOS_USERDATA}"
+  osctl install --bootloader="${WITH_BOOTLOADER}" --device="${DEVICE}" --platform="${TALOS_PLATFORM}" --userdata="${TALOS_USERDATA}" ${EXTRA_ARGS}
 }
 
 function create_iso() {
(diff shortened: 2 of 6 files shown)


cmd/osctl/cmd/install_linux.go
```diff
@@ -91,6 +91,6 @@ func init() {
 	installCmd.Flags().StringVar(&device, "device", "", "The path to the device to install to")
 	installCmd.Flags().StringVar(&endpoint, "userdata", "", "The value of "+constants.KernelParamUserData)
 	installCmd.Flags().StringVar(&platform, "platform", "", "The value of "+constants.KernelParamPlatform)
-	installCmd.Flags().StringSliceVar(&extraKernelArgs, "extra-kernel-arg", []string{}, "Extra argument to pass to the kernel")
+	installCmd.Flags().StringArrayVar(&extraKernelArgs, "extra-kernel-arg", []string{}, "Extra argument to pass to the kernel")
```
bradbeam: When using a `StringSliceVar`, it splits on a comma. This breaks settings that include a comma such a baud rate ( `ttyS1,115200n8` ). Switching to `StringArrayVar` keeps the comma intact.

internal/pkg/kernel/kernel.go
```diff
@@ -20,10 +20,10 @@ func NewDefaultCmdline() *cmdline {
 	cmdline.Append("pti", "on")
 	// TODO(andrewrynhard): Add slub_debug=P. See https://github.com/talos-systems/talos/pull/157.
 	cmdline.Append("consoleblank", "0")
-	cmdline.Append("console", "ttyS1,115200n8")
+	//cmdline.Append("console", "ttyS1,115200n8")
```
bradbeam: Leaving this comment in for discussion. I'm not sure of the best place to migrate these parameters to.

bradbeam: These are currently called inline:

```
internal//app/init/internal/platform/iso/iso.go:	cmdline := kernel.NewDefaultCmdline()
internal//app/init/internal/platform/baremetal/baremetal.go:	cmdline := kernel.NewDefaultCmdline()
internal//app/init/internal/platform/cloud/packet/packet.go:	cmdline := kernel.NewDefaultCmdline()
```

andrewrynhard: I wonder if we need `--append-kernel-arg` and `--set-kernel-arg`. Then use `cmdline.Set(...)`. 🤔 not the prettiest. Or somehow in the `NewDefaultCmdline` we handle this.

hack/installer/entrypoint.sh
```diff
@@ -86,14 +87,17 @@ case "$1" in
...
+          ;;
         n )
           TALOS_RAW="/out/${OPTARG}.raw"
           ;;
         p )
           TALOS_PLATFORM=${OPTARG}
           echo "Using kernel parameter talos.platform=${TALOS_PLATFORM}"
           ;;
-       r )
+        r )
```
smira: extra space?

internal/pkg/kernel/kernel.go
andrewrynhard: The quick win is that we just be explicit in our image creation (i.e. just use `-e` in the `Makefile`).

hack/installer/entrypoint.sh
andrewrynhard: I think this just aligns with the other cases.

bradbeam: ^ yeah just a little alignment fix

PR #875, opened 2019-07-16
fix: generate GCE images with a boot loader

andrewrynhard:
The -b flag needs to be supplied to the installer's script in order to
install a boot loader to the disk.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ generate: buildkitd
 
 .PHONY: talos-gce
 talos-gce:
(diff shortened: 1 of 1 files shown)


andrewrynhard:
I was confused, but in fairness its the `entrypoint.sh` that is confusing :). Feel s more natural to have the `-b` flag be required to create the image with the boot loader.

rsmitty:
Note: don't merge w/out machined in .drone.yml

PR #878, opened 2019-07-17
fix(init): Fix hostname retrieval in azure

bradbeam:
This bit needed better handling around the http headers.


bradbeam:
note, there is still a bug with hostname configuration. 
Even though we retrieve the correct hostname from the metadata service here, the dhcp routines explicitly set the hostname based on the dhcp response. This hostname from dhcp response matches up with the ip address, so we'll need to give some thought on how to handle this properly.

andrewrynhard:
Ok I will add label to remind us not to merge this until we figure that out.

PR #880, opened 2019-07-17
feat: allow specification of mtu for cluster create

rsmitty:
This PR adds the ability to set mtu for the cluster create networks.
Default is 1440, which seems to be the default for calico.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -29,6 +29,7 @@ import (
 var (
 	clusterName string
 	image       string
+	networkMTU  string
 	workers     int
 )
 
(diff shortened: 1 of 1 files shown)


cmd/osctl/cmd/cluster.go
```diff
@@ -329,6 +333,7 @@ func saveConfig(input *generate.Input) (err error) {
 
 func init() {
 	clusterUpCmd.Flags().StringVar(&image, "image", "docker.io/autonomy/talos:"+version.Tag, "the image to use")
+	clusterUpCmd.Flags().StringVar(&networkMTU, "mtu", "1440", "MTU of the docker bridge network")
```
andrewrynhard: We should probably default to `1500` like the docker daemon does.

rsmitty: Updated.

PR #881, opened 2019-07-17
fix: Truncate hostname if necessary

bradbeam:
We should only set the hostname to the actual host name instead of FQDN.
This hasnt been much of an issue, but GCE does return the FQDN for the
hostname field in dhcp.

### DIFF
--- a/internal/pkg/network/dhcp.go
+++ b/internal/pkg/network/dhcp.go
@@ -7,7 +7,9 @@ package network
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/talos-systems/talos/internal/pkg/constants"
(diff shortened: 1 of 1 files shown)


internal/pkg/network/dhcp.go
```diff
@@ -118,6 +120,13 @@ func (service *Service) dhclient4(ctx context.Context, ifname string, modifiers
 				hostname = *kernHostname
 			}
 
+			// Truncate hostname to be betta
+			// Allow IP addrs to be valid hostnames for the time being
+			if ok := net.ParseIP(hostname); ok != nil {
+				// Pull out the first part of a potential FQDN
+				hostname = strings.Split(hostname, ".")[0]
```
smira: to be on the safe side, I would check first that hostname is not empty, as split returns empty slice in that case

https://play.golang.org/p/3BS4TfAMXxT

smira: sorry, I confused myself, it's still one-item slice.

PR #882, opened 2019-07-17
fix: use existing logic to perform reset

andrewrynhard:
This PR moves the reset API to the init API definition.
It leverages the same code we use for upgrades.

Signed-off-by: Andrew Rynhard <andrew@andrewrynhard.com>
(cherry picked from commit 5d8ee0a3a5fb5e927513f710fb7e462a34bf3446)

fix(init): Fix routes endpoint

Temporary workaround while we get more information on the
specifics for what is failing.

Ref: #795
(cherry picked from commit 58537faa8b9a41a938248afe54eba955c60d39db)

fix(init) mount root partition as read-only

This uses the correct mount flag for read-only.
We mistakenly had the flag for opening a file as read-only.

Signed-off-by: Andrew Rynhard <andrew@andrewrynhard.com>
(cherry picked from commit d4a59b7c1433e31736656589282487c01080f8d1)

fix: return non-nil response in reset

The gRPC response will fail to be decoded because our reply is nil.

Signed-off-by: Andrew Rynhard <andrew@andrewrynhard.com>
(cherry picked from commit c40802b12257399874461d93ab025b51ce655cb1)

fix: mount the rootfs as read-only

There was not a clean commit we could pull this from so the changes were
made directly onto the release branch.

### DIFF
--- a/internal/app/init/internal/reg/reg.go
+++ b/internal/app/init/internal/reg/reg.go
@@ -121,7 +121,7 @@ func (r *Registrator) Reset(ctx context.Context, in *empty.Empty) (data *proto.R
 		return data, err
 	}
 
-	return data, err
+	return &proto.ResetReply{}, err
 }
 
 // ServiceList returns list of the registered services and their status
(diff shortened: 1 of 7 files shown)


andrewrynhard:
This is a number of bug fixes that needed to be back ported into 0.1.

PR #885, opened 2019-07-17
feat: allow mtu specification for network devices

rsmitty:
This PR is needed so we can specify an MTU of 1460 for GCE VMs

### DIFF
--- a/internal/pkg/network/setup.go
+++ b/internal/pkg/network/setup.go
@@ -69,7 +69,7 @@ func SetupNetwork(data *userdata.UserData) (err error) {
 // Maybe look at adjusting this to accept an interface value from a kernel arg
 func defaultNetworkSetup() (err error) {
 	log.Println("bringing up lo")
-	if err = ifup("lo"); err != nil {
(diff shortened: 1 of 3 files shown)


internal/pkg/network/setup.go
```diff
@@ -92,7 +92,7 @@ func defaultNetworkSetup() (err error) {
 	return nil
 }
 
-func ifup(ifname string) (err error) {
+func ifup(ifname string, mtu int) (err error) {
```
smira: should we consider option interface (or refactoring interface to be more of an object?) to support more parameters in the future, and also avoid passing `0` as default value?

rsmitty: Yeah, this is a bit of a stopgap for now. This will likely get rewritten as part of the networkd rewrite that Brad is working on.

bradbeam: Yeah, there's a significant refactor to this network code that needs to happen and we just haven't gotten around to yet. I think for now this should be fine.

PR #888, opened 2019-07-18
chore: run tests in the buildkit itself

smira:
This relies on two PRs to the buildkit, which aren't merged yet, so I
had to do some overrides to apply them:

* https://github.com/moby/buildkit/pull/1081
* https://github.com/moby/buildkit/pull/1085

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 TOOLS ?= autonomy/tools:b4e3778
 
 # TODO(andrewrynhard): Move this logic to a shell script.
-BUILDKIT_VERSION ?= v0.5.0
+BUILDKIT_VERSION ?= master@sha256:455f06ede03149051ce2734d9639c28aed1b6e8b8a0c607cb813e29b469a07d6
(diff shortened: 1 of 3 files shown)


smira:
Forgot to remove `dev-test` target, I'll do it in the next push, want to see if it works in drone

Makefile
```diff
@@ -1,7 +1,7 @@
 TOOLS ?= autonomy/tools:b4e3778
 
 # TODO(andrewrynhard): Move this logic to a shell script.
-BUILDKIT_VERSION ?= v0.5.0
+BUILDKIT_VERSION ?= master@sha256:455f06ede03149051ce2734d9639c28aed1b6e8b8a0c607cb813e29b469a07d6
```
andrewrynhard: I can update our current deployment to match this.

andrewrynhard: Updated and it looks like this works!

smira: oh, that's awesome!

let me rebase and squash it. still waiting on dockerfile PR, and we can update `Dockerfile` slug back to upstream

PR #896, opened 2019-07-22
feat: attempt to connect to all trustd endpoints when downloading PKI

### LINKED ISSUE #891: Trustd should try all endpoints when downloading PKI
The code is self-explanatory. This fix would make the control plane much more resilient to failures.
```
// NewGenerator initializes a Generator with a preconfigured grpc.ClientConn.
func NewGenerator(data *userdata.UserData, port int) (g *Generator, err error) {
	if len(data.Services.Trustd.Endpoints) == 0 {
		return nil, fmt.Errorf("at least one root of trust endpoint is required")
	}

	creds, err := basic.NewCredentials(data.Services.Trustd)
	if err != nil {
		return nil, err
	}

	// TODO: In the case of failure, attempt to generate the identity from
	// another RoT.
	var conn *grpc.ClientConn
	conn, err = basic.NewConnection(data.Services.Trustd.Endpoints[0], port, creds)
	if err != nil {
		return nil, err
	}
	client := proto.NewTrustdClient(conn)

	return &Generator{
		client: client,
	}, nil
}
```

rsmitty:
This PR will connect to each trustd endpoint specified, returning once successful. Closes #891.

### DIFF
--- a/internal/pkg/grpc/gen/gen.go
+++ b/internal/pkg/grpc/gen/gen.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/talos-systems/talos/internal/app/trustd/proto"
 	"github.com/talos-systems/talos/internal/pkg/grpc/middleware/auth/basic"
 	"github.com/talos-systems/talos/pkg/crypto/x509"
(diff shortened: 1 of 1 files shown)


internal/pkg/grpc/gen/gen.go
```diff
@@ -33,18 +33,22 @@ func NewGenerator(data *userdata.UserData, port int) (g *Generator, err error) {
...
+		return &Generator{client: client}, nil
 	}
-	client := proto.NewTrustdClient(conn)
 
-	return &Generator{
-		client: client,
-	}, nil
+	// We were unable to connect to any trustd endpoint
+	// Return error from last attempt.
+	return nil, err
```
smira: does it make sense to combine all the errors into `multierror`, as the error talking to the last endpoint might be not the same as the others, potentially hiding the issue? (e.g. last endpoint is down, but others have other issues, and the fact that it's down gives zero info on the previous failures)

rsmitty: Good thought, let me look at how we've used that package previously.

rsmitty: @smira updated. see what you think now.

PR #897, opened 2019-07-22
chore: fix image builds on tags

andrewrynhard:
The GCE and Azure steps need to run in serial since they both output the
same artifact name.


.drone.yml
```diff
@@ -338,8 +339,8 @@ volumes:
     temp: {}
 
 trigger:
-  cron: 
-    exclude: [ nightly ]
```
andrewrynhard: Editor auto formats YAML

PR #902, opened 2019-07-23
chore: allow to run tests only for specified packages

smira:
This allows to do `make test TESTPKGS=./internal/app/machined`.

Also update Dockerfile slug as
https://github.com/moby/buildkit/pull/1081 was merged into master.

### DIFF
--- a/hack/golang/test.sh
+++ b/hack/golang/test.sh
@@ -2,24 +2,23 @@
 
 set -e
 
-CGO_ENABLED=1
-
 perform_tests() {
-  echo "Performing tests"
-  go test -v -covermode=atomic -coverprofile=coverage.txt ./...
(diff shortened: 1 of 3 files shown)


hack/golang/test.sh
```diff
@@ -2,24 +2,23 @@
 
 set -e
 
-CGO_ENABLED=1
-
 perform_tests() {
-  echo "Performing tests"
-  go test -v -covermode=atomic -coverprofile=coverage.txt ./...
+  echo "Performing tests on $1"
+  GODEBUG=gocachetest=1 go test -v -covermode=atomic -coverprofile=coverage.txt "$1"
```
smira: oops this is not intentional (`GODEBUG`), I was trying to debug cache, and left it by mistake

PR #903, opened 2019-07-23
chore: implement first version of CRI runner

smira:
It runs containers via CRI interface in a pod sandbox. This is the very
first version:  I tried not to introduce any changes to common runner
interface.

There should be some CRI-speficic options for the runner (like polling
interval, as it doesn't have nice `Wait()` API), plus my plan so far is
to use OCI as the common layer for container options, so that we can
analyze OCI and translate to CRI (when possible, return errors when
option is not implemented).

CRI interface doesn't have a concept of 'unpacking' an image, so we
probably need to unpack via containerd API (or any other
runtime-specific API) by targeting CRI namespace.

### DIFF
--- a/internal/app/machined/pkg/system/runner/cri/cri_test.go
+++ b/internal/app/machined/pkg/system/runner/cri/cri_test.go
@@ -1,0 +1,338 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package cri_test
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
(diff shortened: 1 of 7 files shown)


internal/app/machined/pkg/system/runner/cri/cri_test.go
```diff
@@ -0,0 +1,338 @@
...
+
+	suite.Assert().NoError(r.Run(MockEventSink))
+
+	logContents := suite.getLogContents("logtest.log")
+
+	suite.Assert().Contains(string(logContents), "Test 1\n")
+	suite.Assert().Contains(string(logContents), "Test 2\n")
+}
+
+// func (suite *CRISuite) TestStopFailingAndRestarting() {
```
andrewrynhard: Is this commented on purpose?

smira: yes, I copy-pasted it from containerd tests, but I can't enable it yet, as CRI runner doesn't support mounts, so my plan is to enable it once mounts are supported

PR #913, opened 2019-07-25
chore: enable and use properly go build cache

smira:
There are two parts of the change:

1. Use `go generate` to generate file with version tags instead of
passing them as linker flags. (This might need some cleanup).

2. Remove `-a` to `go build`, clean up cache mounts.


smira:
Going to close this one, cache part was merged in #915, while `go generate` for version looks ugly to me and I don't see any good reason to use it

PR #914, opened 2019-07-26
refactor: move setup logic into machined

andrewrynhard:
The responsibility of init should only be to mount the rootfs. This
change moves Talos specific logic into machined. This will allow us to
define a version of Talos in a single binary instead of split across
two. This will enable cleaner upgrades and helps make the codebase
easier to reason about.


andrewrynhard:
This is one more step into the direction of refactoring `init` and `machined`. I want to keep the scope of this PR to only moving existing code from `init` into `machined` and follow up with some more refactor PRs that are focused on specific areas of `machined`.

internal/app/machined/main.go
```diff
@@ -128,26 +132,94 @@ func createOverlay(path string) error {
...
+		if err = p.Prepare(data); err != nil {
+			return err
+		}
+
+		var initializer *mount.Initializer
+		if initializer, err = mount.NewInitializer(""); err != nil {
+			return err
+		}
+
+		// Mount the owned partitions.
```
bradbeam: Should we move these bits earlier? Might not be a 'this pr' issue, but curious if there's some slight reorganizations of things we can do.

andrewrynhard: We probably can. I have been playing around with some ideas about how we can clean things up. The problem is that things are very much required to run in serial and over time it has become unclear about subgroups within those steps that might actually be able to be ran in parallel. I plan on fixing this in my next PR.

PR #919, opened 2019-07-28
refactor: split machined into phases

andrewrynhard:
This change aims to provide more structure to the implementation of machined.
It introduces the notion of a phase that consists of any number of stages.
Phases are executed in serial, while the stages the make up a phase are
executed in parallel.


andrewrynhard:
This is a massive change because we have let `init`/`machined` sort of fall apart of time. Happy to do video review. The best place to start IMO is to look at `machined`'s `main.go`.

internal/app/machined/internal/mount/mount.go
```diff
@@ -0,0 +1,198 @@
...
+// nolint: dupl
+func mountpoints() (mountpoints *mount.Points, err error) {
+	mountpoints = mount.NewMountPoints()
+	for _, name := range []string{constants.DataPartitionLabel, constants.BootPartitionLabel} {
+		var target string
+		switch name {
+		case constants.DataPartitionLabel:
+			target = constants.DataMountPoint
+		case constants.BootPartitionLabel:
+			target = constants.BootMountPoint
```
bradbeam: probably overkill, but should we add a `default: to return error` ?

internal/app/machined/internal/phase/api/acpi.go
```diff
@@ -0,0 +1,47 @@
...
+			log.Printf("sync done")
+			return
+		case <-time.After(time.Second):
+		}
+		if i != 0 {
+			log.Printf("waiting %d more seconds for sync to finish", i)
+		}
+	}
+
+	log.Printf("sync hasn't completed in time, aborting...")
```
bradbeam: i know it's not part of the actual pr itself, but we probably want to drop the `, aborting...`. makes it unclear if the shutdown is being aborted or the sync.

internal/app/machined/internal/phase/api/api.go
```diff
@@ -0,0 +1,109 @@
...
+		case <-api.RebootCh:
+			log.Printf("reboot via API received, rebooting...")
+			rebootCmd = unix.LINUX_REBOOT_CMD_RESTART
+		}
+	}()
+
+	return nil
+}
+
+func (task *API) container(platform platform.Platform, data *userdata.UserData) (err error) {
```
bradbeam: nit, unsure if it's worth DRY this part since it's also used in the standard bit.

internal/app/machined/internal/mount/mount.go
andrewrynhard: I have some local changes to clean up all the mount logic, was planning on a follow-up PR. I will take a closer look at this then.

internal/app/machined/internal/phase/network/network.go
```diff
@@ -0,0 +1,39 @@
...
+// NewUserDefinedNetworkTask initializes and returns an UserDefinedNetwork task.
+func NewUserDefinedNetworkTask() phase.Task {
+	return &UserDefinedNetwork{}
+}
+
+// RuntimeFunc returns the runtime function.
+func (task *UserDefinedNetwork) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
+	switch mode {
+	case runtime.Standard:
+		return task.runtime
```
bradbeam: feels odd that we don't do anything for the container here, but i guess it makes sense. we'd always see the DHCP logs/errors previously.

internal/app/machined/internal/phase/install/install.go
```diff
@@ -0,0 +1,59 @@
...
+func (task *Install) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
+	switch mode {
+	case runtime.Standard:
+		return task.runtime
+	default:
+		return nil
+	}
+}
+
+func (task *Install) runtime(platform platform.Platform, data *userdata.UserData) (err error) {
```
bradbeam: nit, should we change `platform` variable name to something else so it doesn't overlap with package name?

internal/app/machined/internal/phase/api/api.go
andrewrynhard: fixed

internal/app/machined/internal/phase/network/network.go
andrewrynhard: Yeah setting up networking defined by the machine config in a container doesn't make sense so I left it out. So we return `nil` here to indicate to the phase runner that there is nothing to run.

internal/app/machined/internal/phase/phase.go
```diff
@@ -0,0 +1,111 @@
...
+	platform, err := platform.NewPlatform()
+	if err != nil {
+		return nil, err
+	}
+
+	mode := runtime.Standard
+	switch platform.(type) {
+	case *container.Container:
+		mode = runtime.Container
+	default:
```
bradbeam: nit, seems kind of odd to overload default to do kmsg.Setup instead of being focused on mode/platform.
unfortunately the alternative would feel awkward as well i think, so perhaps this is the better.

internal/app/machined/internal/phase/install/install.go
andrewrynhard: Since we don't depend on the platform package any point after, then its fine as is. avoids having a non-descriptive `p` floating around. But happy to change if you feel strongly about it.

internal/app/machined/internal/phase/phase.go
```diff
@@ -0,0 +1,111 @@
...
+				var f RuntimeFunc
+				if f = task.RuntimeFunc(r.mode); f == nil {
+					// A task is not defined for this runtime mode.
+					continue
+				}
+				if err := f(r.platform, r.data); err != nil {
+					result = multierror.Append(result, err)
+				}
+			}
+		}(phase)
```
bradbeam: should we also pass in result? not sure what the best practice around this would be. all or nothing?

internal/app/machined/internal/phase/phase.go
andrewrynhard: This was mostly because I don't know what `log`'s default are and rather than mucking with those I chose to setup logging to `/dev/kmsg` only when we need to.

internal/app/machined/internal/phase/phase.go
andrewrynhard: Not sure on the _why_ but the linter only yelled at me when `phase` wasn't passed in.

internal/app/machined/internal/phase/rootfs/mount_overlays.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/talos-systems/talos/internal/app/machined/internal/runtime"
+	"github.com/talos-systems/talos/pkg/userdata"
+
+	"golang.org/x/sys/unix"
+)
+
+const (
+	// SystemVarPath is the path to write runtime system related files and
+	// directories.
+	SystemVarPath = "/var/system"
```
bradbeam: Should we move this to internal/pkg/constants? Seems like this and roots.etc might be good candidates for having these bits defined there.

andrewrynhard: I debated that and what I landed on is that I'm starting to favor the idea of constants that are relevant to a particular domain (rootfs) in this case should live in that package. That being said I think the the global `constants.go` file has its' place as well. So just a long winded way of saying "I'm torn". :)

internal/app/machined/internal/phase/userdata/save_userdata.go
```diff
@@ -0,0 +1,49 @@
...
+func NewSaveUserDataTask() phase.Task {
+	return &SaveUserData{}
+}
+
+// RuntimeFunc returns the runtime function.
+func (task *SaveUserData) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
+	return task.runtime
+}
+
+func (task *SaveUserData) runtime(platform platform.Platform, data *userdata.UserData) (err error) {
```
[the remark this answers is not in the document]
andrewrynhard: Unfortunately `!os.IsNotExist(err)` does not guarantee that that file exists. `err` could be a number of other errors.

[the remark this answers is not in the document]
andrewrynhard: To make it a little cleaner I can do an early return.

PR #920, opened 2019-07-29
feat: break up osctl cluster create and basic/e2e tests

rsmitty:
This PR will break cluster create apart from the other steps in
integration tests. It will allow us to run the cluster create, then use
it for parallel e2e builds in different cloud environments.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -386,7 +386,23 @@ services:
         path: /tmp
 
 steps:
-  - name: gce
+  - name: osctl-cluster-create
+    image: autonomy/build-container:latest
+    pull: always
+    environment:
+      BUILDKIT_HOST: tcp://buildkitd.ci.svc:1234
+      BINDIR: /usr/local/bin
+      TAG: latest
+    commands:
+      - make osctl-linux
+      - make osctl-cluster-create
+    volumes:
+      - name: dockersock
+        path: /var/run
+      - name: tmp
+        path: /tmp
+
+  - name: gce-image
     image: autonomy/build-container:latest
     pull: always
     environment:
@@ -403,8 +419,10 @@ steps:
         path: /var/run
       - name: dev
         path: /dev
+    depends_on:
+      - osctl-cluster-create
 
-  - name: e2e-integration
+  - name: gce-e2e-integration
     image: autonomy/build-container:latest
     pull: always
     environment:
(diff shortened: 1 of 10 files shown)


hack/test/manifests/gce-cluster.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: cluster.k8s.io/v1alpha1
 kind: Cluster
 metadata:
   annotations: null
-  name: talos-e2e
+  name: talos-e2e-gce
```
rsmitty: Renamed all cluster refs to include gce so it's unique vs. azure later

hack/test/e2e-runner.sh
```diff
@@ -0,0 +1,25 @@
+
+e2e_run() {
```
rsmitty: Helper functions that get sourced by each e2e script.

hack/test/e2e-platform.sh
```diff
@@ -0,0 +1,65 @@
+source ./hack/test/e2e-runner.sh
+
+## Download kustomize and template out capi cluster, then deploy it
```
rsmitty: Broke this out into its own script, as it was a fair amount of logic. This will create the cluster in a given `$PLATFORM`, which will be gce/azure for now.

.drone.yml
```diff
@@ -386,7 +386,23 @@ services:
         path: /tmp
 
 steps:
-  - name: gce
+  - name: osctl-cluster-create
```
rsmitty: Broke the osctl cluster into its own step. The GCE and Azure e2e tests will both use the same CAPI deployment on this cluster for their testing.

.drone.yml
```diff
@@ -386,7 +386,23 @@ services:
...
+    image: autonomy/build-container:latest
```
bradbeam: maybe later see if we can switch to autonomy/installer and skip osctl-linux build?

hack/test/e2e-integration.sh
```diff
@@ -2,63 +2,35 @@
...
-cleanup() {
- e2e_run "kubectl delete machine talos-e2e-master-0 talos-e2e-master-1 talos-e2e-master-2
-          kubectl scale machinedeployment talos-e2e-workers --replicas=0
-          kubectl delete machinedeployment talos-e2e-workers
-          kubectl delete cluster talos-e2e"
- ${OSCTL} cluster destroy --name integration
- rm -rf ${TMP}
-}
-trap cleanup EXIT
+## ClusterAPI Provider Talos (CAPT)
```
bradbeam: maybe consolidate these (+kubectl versions) in e2e.sh

PR #922, opened 2019-07-30
chore: add AMI build

andrewrynhard:
This will build AMIs and publish them to our official account on a
release.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -230,7 +230,6 @@ steps:
     when:
       event: tag
     depends_on:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -230,7 +230,6 @@ steps:
     when:
       event: tag
     depends_on:
-      ## Should change to e2e once we get things more stable
```
andrewrynhard: This is not true anymore for this pipeline.

PR #927, opened 2019-07-30
chore: add 'talos-cli' container image

smira:
This is more of PoC now, plan is to integrate this with `osctl` to
provide easy way to get access kubectl to the running Talos cluster.

I haven't added to the list of 'images' being pushed so that we can play
a bit with it before it becomes official.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,16 @@ talos: buildkitd
 osctl-cluster-create:
 	@TAG=$(TAG) ./hack/test/$@.sh
 
+.PHONY: cli
+cli: buildkitd
+	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
+		build \
+		--output type=docker,dest=build/$@.tar,name=docker.io/autonomy/$@:$(TAG) \
+		--opt target=$@ \
+		--opt build-arg:KUBECTL_VERSION=$(KUBECTL_VERSION) \
+		$(COMMON_ARGS)
+	@docker load < build/$@.tar
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -211,6 +211,16 @@ talos: buildkitd
 		$(COMMON_ARGS)
 	@docker load < build/$@.tar
 
+.PHONY: talos-cli
+talos-cli: buildkitd
+	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
+		build \
+		--output type=docker,dest=build/$@.tar,name=docker.io/autonomy/$@:$(TAG) \
```
andrewrynhard: I know right now we are namespace with `autonomy` but I plan on fixing that. I think I
'd prefer this to just be called `cli`. Then when we move to something like GH packages, or a registry we host, we can have the full image be something like `talos/cli:latest`.

smira:
TODO: use `KUBECTL_ARCHIVE` from `Makefile`

andrewrynhard:
Sorry just now catching up on PRs. Is this ready?

smira:
I don't think it is, let's keep it as a PR. I have the plan, but I wasn't
able to implement everything for it.

сб, 10 авг. 2019 г., 9:15 Andrew Rynhard <notifications@github.com>:

andrewrynhard:
@smira Is this still something you want to get done?

smira:
We can close the PR, I can get back to it, but looks like it's not top priority

PR #928, opened 2019-07-30
fix: Remove references to NewRoot in install code.

bradbeam:
Since we revamped how init and machine startup happens, we no longer juggle
roots in the same manner. These refereces are no longer needed.


andrewrynhard:
@bradbeam Per our discussion on slack, I'm gonna close this.

For anyone else, the short version is that this will break `osctl install`. I am working on a refactor of our installer code and it should fix this.

PR #929, opened 2019-07-30
feat: Update networkd

bradbeam:
This introduces a rewrite of our networkd implementation. This update
includes moving off our fork of our dhcp library, replacement of the
underlying netlink library, and more granular control of how we
interact with the network stack


bradbeam:
Still working through some nuances, but able to successfully boot up a machine in packet with this.

PR #931, opened 2019-07-30
feat(osctl): allow configurable number of masters to `cluster create`

smira:
This allows to run tiny Talos clusters (which is sometimes nice for
local testing), e.g. with just a single master and zero workers.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -31,7 +31,10 @@ var (
 	image       string
 	networkMTU  string
 	workers     int
+	masters     int
 )
+
+const baseNetwork = "10.5.0.%d"
 
 // clusterCmd represents the cluster command
 var clusterCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


andrewrynhard:
Awesome. This has been in the back of mind every day for the last 2 weeks.

smira:
> LGTM. I have one nit. We should make sure we set --masters=3 in our basic integration test.

yep, I figured that out while testing it locally :))

cmd/osctl/cmd/cluster.go
```diff
@@ -335,6 +335,7 @@ func init() {
 	clusterUpCmd.Flags().StringVar(&image, "image", "docker.io/autonomy/talos:"+version.Tag, "the image to use")
 	clusterUpCmd.Flags().StringVar(&networkMTU, "mtu", "1500", "MTU of the docker bridge network")
 	clusterUpCmd.Flags().IntVar(&workers, "workers", 1, "the number of workers to create")
+	clusterUpCmd.Flags().IntVar(&masters, "masters", 3, "the number of masters to create")
```
andrewrynhard: I liked the idea of defaulting to 1. I think HA locally has few benefits outside of our basic integration test. Not a blocker for me though.

smira: I was afraid of breaking something else or someone's else automation by changing defaults here.

PR #932, opened 2019-07-31
refactor: improve installation reliability

andrewrynhard:
This change aims to make installations more unified and reliable. It
introduces the concept of a mountpoint manager that is capable of
mounting, unmounting, and moving a set of mountpoints in the correct
order.


internal/app/init/main.go
```diff
@@ -31,23 +33,40 @@ func initram() (err error) {
...
-	if err = initializer.Switch(); err != nil {
+	if err = switchroot.Switch(constants.NewRoot, virtual); err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func recovery() {
+	if r := recover(); r != nil {
```
smira: not related to this PR, but I remember we had a disk sync handler on the return path from `main()` (probably more for machined, but still), do we need it still or it was removed on purpose?

andrewrynhard: At this point `init` is only the initramfs and no disks have been mounted, so I _think_ we can skip that step. That being said it doesn't hurt to call `sync`.

smira: I don't see it in machined either. I can add it back if you don't mind

andrewrynhard: Its handled in a task now: 
https://github.com/talos-systems/talos/blob/master/internal/app/machined/internal/phase/api/api.go#L68
https://github.com/talos-systems/talos/blob/master/internal/app/machined/internal/phase/api/acpi.go#L41

smira: oh, I missed that!

I noticed a data race in phase runner, I can send  a PR to fix it.

I also thought we might try to unify that though with any other panic-like behavior to get just a single exit path. I can try to do something later on it.

andrewrynhard: +1 on both!

cmd/osctl/cmd/install_linux.go
```diff
@@ -71,7 +63,8 @@ var installCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		if err = install.Install(cmdline.String(), data); err != nil {
+		i := installer.NewInstaller(cmdline, data)
```
andrewrynhard: The changes in this PR essentially boil down to making installations this simple.

internal/app/machined/main.go
```diff
@@ -68,11 +69,11 @@ func run() (err error) {
 			userdatatask.NewExtraFilesTask(),
 		),
 		phase.NewPhase(
-			"installation",
-			install.NewInstallTask(),
+			"platform tasks",
```
bradbeam: nit, probably for a followup pr. but it'd be nice to move these phases names to constants or something a bit more well defined.

internal/pkg/mount/manager/virtual/virtual.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// MountPoints returns the mountpoints required to boot the system.
+func MountPoints() (mountpoints *mount.Points, err error) {
+	virtual := mount.NewMountPoints()
+	virtual.Set("dev", mount.NewMountPoint("devtmpfs", "/dev", "devtmpfs", unix.MS_NOSUID, "mode=0755"))
+	virtual.Set("proc", mount.NewMountPoint("proc", "/proc", "proc", unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV, ""))
+	virtual.Set("sys", mount.NewMountPoint("sysfs", "/sys", "sysfs", 0, ""))
+	virtual.Set("run", mount.NewMountPoint("tmpfs", "/run", "tmpfs", 0, ""))
+	virtual.Set("tmp", mount.NewMountPoint("tmpfs", "/tmp", "tmpfs", 0, ""))
```
bradbeam: not for this pr.
curious if we should set a limit for these tmpfs filesystems. since they're entirely backed by memory it makes me al little concerned a rogue process could 💣 💥 a system

andrewrynhard: We can certainly set the size. Agree that is for a followup as I'm not sure how big we should go just yet, or if we should base it on percentage.

internal/app/machined/main.go
andrewrynhard: Yeah I think this might even change TBH. Sort of a quick and dirty hack to have _some_ way to identify a phase for now. Still thinking through the phase/task bits.

PR #933, opened 2019-07-31
fix: set mtu value regardless of interface state

rsmitty:
This PR will fix a bug we encountered in GCE, where the interface was
already up and the MTU value wasn't getting set.

### DIFF
--- a/internal/pkg/network/setup.go
+++ b/internal/pkg/network/setup.go
@@ -102,19 +102,20 @@ func ifup(ifname string, mtu int) (err error) {
 	case netlink.OperUnknown:
 		fallthrough
 	case netlink.OperDown:
-		if mtu > 0 {
-			if err = netlink.LinkSetMTU(link, mtu); err != nil {
-				return err
-			}
-		}
 		if err = netlink.LinkSetUp(link); err != nil && err != syscall.EEXIST {
 			log.Printf("im failing here in operdown for %s", ifname)
 			return err
 		}
 	case netlink.OperUp:
-		return nil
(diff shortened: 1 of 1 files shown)


internal/pkg/network/setup.go
```diff
@@ -102,21 +102,22 @@ func ifup(ifname string, mtu int) (err error) {
...
 			log.Printf("im failing here in operdown for %s", ifname)
 			return err
 		}
 	case netlink.OperUp:
-		return nil
 	default:
 		return errors.Errorf("cannot handle current state of %s: %s", ifname, attrs.OperState.String())
 	}
 
+	//Configure MTU if specified
```
andrewrynhard: nit: Space before `//` and comment.

internal/pkg/network/setup.go
```diff
@@ -102,21 +102,22 @@ func ifup(ifname string, mtu int) (err error) {
 	case netlink.OperUnknown:
 		fallthrough
 	case netlink.OperDown:
-		if mtu > 0 {
-			if err = netlink.LinkSetMTU(link, mtu); err != nil {
-				return err
-			}
-		}
 		if err = netlink.LinkSetUp(link); err != nil && err != syscall.EEXIST {
 			log.Printf("im failing here in operdown for %s", ifname)
```
smira: lol on the log message here

andrewrynhard: lol, yes that needs to be fixed ... although brad is rewriting `networkd` so it will probably go away then.

PR #934, opened 2019-07-31
fix(init): flip concurrency of tasks/services, fix small issues 

smira:
Phases should run sequentially, while tasks concurrently in a phase.

There are two potential issues fixed:

1. `result` multierror was updated inside goroutine without any
synchronization, so this is a data race
2. panic inside task/phase runner might happen and as unhandled panic in a
goroutine aborts whole process, this might lead to a system halt as
as the 'machined' exits

### DIFF
--- a/internal/app/machined/internal/phase/phase.go
+++ b/internal/app/machined/internal/phase/phase.go
@@ -6,7 +6,8 @@ package phase
 
 import (
 	"log"
-	"sync"
+	goruntime "runtime"
+	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


internal/app/machined/internal/phase/phase.go
```diff
@@ -52,33 +51,49 @@ func NewRunner(data *userdata.UserData) (*Runner, error) {
...
-				if err := f(r.platform, r.data); err != nil {
-					result = multierror.Append(result, err)
-				}
-			}
-		}(phase)
-		wg.Wait()
-		if result != nil {
-			return result.ErrorOrNil()
-		}
+		r.runPhase(phase, errCh)
```
bradbeam: is there much difference in 
```
go r.runPhase
```
vs
```
runPhase{
  go func() {
...
  }
}
```

if there isn't, i think I'd prefer the first form so it's clear that r.runPhase is being launched as a goroutine.

andrewrynhard:
> This adds tests and changes logic slightly: if some task in a phase errors,
this stops phase run (but not other phases runs). I thought it makes
more sense as tasks are run sequentially.

I think you meant phases are ran sequentially and tasks run concurrently? IMO the behavior that makes sense is that the machine should about startup if _any_ tasks fails as they are are deemed critical to get the machine into the expected state. Maybe I'm just misunderstanding you.

internal/app/machined/internal/phase/phase.go
smira: there is no difference, I agree, this is more clear

smira:
> I think you meant phases are ran sequentially and tasks run concurrently? IMO the behavior that makes sense is that the machine should abort startup if _any_ tasks fails as they are are deemed critical to get the machine into the expected state. Maybe I'm just misunderstanding you.
> 
> Just to clarify, phases are dependent on the previous phase to be successful, while tasks stand on their own within a phase.

Per discussion in Slack: actually phases/tasks concurrency were inverted in the `master`, this PR addresses that.

internal/app/machined/internal/phase/phase.go
```diff
@@ -50,37 +51,69 @@ func NewRunner(data *userdata.UserData) (*Runner, error) {
...
+
+	for range phase.tasks {
+		err := <-errCh
+		if err != nil {
+			log.Printf("[phase]: %s error running task: %s", phase.name, err)
+		}
+		result = multierror.Append(result, err)
+	}
+
+	log.Printf("[phase]: %s done, %s", phase.name, time.Since(start))
```
andrewrynhard: Nice

smira: I thought it might be nice to see which phase takes more time.

andrewrynhard: Yeah I'm curious to see as well.

PR #939, opened 2019-08-01
docs: add boot sequence proposal

andrewrynhard:
This is a proposal that introduces the notion of a boot sequencer.

### DIFF
--- a/docs/proposals/20190801-BootSequence.md
+++ b/docs/proposals/20190801-BootSequence.md
@@ -1,0 +1,113 @@
+# Proposal: Boot Sequence
+
+Author: Andrew Rynhard
+
+## Abstract
+
+Ensuring proper order of tasks in boot and shutdown events is currently a problem.
+This document proposes a boot sequence management system that can be used to boot and shutdown a node in the proper order.
+
+## Background
+
+The main routine of `machined` was very script-like in the sense that it was a very long sequence of conditionals and functions.
+Lacking any kind of abstraction, a lot of code was repeated, conditionals were all over the place and hard to make sense of.
+This poses problems when we start to look at what it would take to bring a node back to its original state of having zero mounts, networking, processes (except PID 1 of course), etc.
+Undoing what is done only exacerbates all the previously mentioned problems.
+
+## Proposal
+
+This proposal builds on the notion of phases, tasks, and services by outlining yet another abstraction that ties these together.
+I propopse an `struct` aptly named `BootSequencer`.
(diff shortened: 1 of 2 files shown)


docs/proposals/20190801-BootSequence.md
```diff
@@ -0,0 +1,113 @@
...
+type Status int
+
+const (
+    Pass Status = iota
+    Fail
+)
+
+type Level struct {
+    status Status
+    phases []phase.Phase
```
smira: should it be `phases [][]phase.Phase`?

docs/proposals/20190801-BootSequence.md
```diff
@@ -0,0 +1,113 @@
...
+    boot.Register(&Phase0Task{}, 0)
+}
+```
+
+Notice that `phase.Runner` is embedded.
+This serves to enforce the idea that a `BootSequencer` is a wrapper around a `phase.Runner`, that adds some extra functionality on top of it.
+The `BootSequencer` will perform the following tasks:
+
+- register phases and build the execution order
+- create any event handling channels
```
smira: can we instead have a singleton-like standalone object act as service bus components can subscribe to and publish to?

e.g. ACPI listener can publish shutdown even there, while machined's main.go can listen on it to trigger shutdown?

does it have part of boot sequence, or we can think of that as standalone component?

andrewrynhard: Yeah I like this idea. This object I think more or less will encompass all of what `machined` does. The main function could just set this thing up and then wait for a signal. Maybe this standalone component contains the boot sequencer.

docs/proposals/20190801-BootSequence.md
andrewrynhard: I think so ... building out an example go file and will be sure it compiles. I will copy over here when done.

smira:
Another potential approach to the boot sequence and cleanup might be borrowing concept of typical Go HTTP middleware:

```go
type Phase func(somedata *Data, nextPhase phase) error

func examplePhase func(somedata *Data, nextPhase error) {
     if err := doSomething(); err != nil {
         return err
     }
     defer cleanup()

     return nextPhase(somedata)
}
```

in this case phases are totally free to implement cleanup the way they want, and cleanup happens "naturally" as `defer` objects are run during stack unwinding either via normal return sequence or due to the panic.

andrewrynhard:
This is outdated, and not really worth salvaging.

PR #943, opened 2019-08-02
chore(init): rearrange phase handling to push shutdown to main

smira:
This re-arranges phases a bit so that shutdown actions are pushed back
to the top-level main.go of machined.

Small rudimentary event.Bus is introduce to facilitate event passing
(shutdown/restart) between various machined components and main.go. This
might be not the best implementation, just something to allow this
message passing without global variables or such.

Machined API was refactored to run as goroutine service.

ACPI & signal handlers re-built as phase tasks, and activated for
non-container, container modes respectively.

As part of the fix, now `docker stop` triggers correct shutdown of Talos
(not a big deal, but good for testing).


smira:
This doesn't match proposal in #939, it can be thought as interim step in the evolution of the boot flow

andrewrynhard:
> This doesn't match proposal in #939, it can be thought as interim step in the evolution of the boot flow

Perfectly fine... proposal is still a WIP and made completely change. This is still a step in the right direction.

PR #944, opened 2019-08-02
chore: add tests for event.Bus

smira:
Small tests to make sure code works as expected.


smira:
#946 to fix the error from test run (I wonder why it got triggered more often with this change)

PR #945, opened 2019-08-02
chore: fix data race in goroutine runner

smira:
Discovered with `go test -race`:

```
WARNING: DATA RACE
Read at 0x00c0000cf2f8 by goroutine 25:
  github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/goroutine.(*goroutineRunner).Stop()
      /home/smira/Documents/autonomy/talos/internal/app/machined/pkg/system/runner/goroutine/goroutine.go:111 +0x3e
  github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/goroutine_test.(*GoroutineSuite).TestStop()
      /home/smira/Documents/autonomy/talos/internal/app/machined/pkg/system/runner/goroutine/goroutine_test.go:115 +0x345
  runtime.call32()
      /usr/local/go/src/runtime/asm_amd64.s:519 +0x3a
  reflect.Value.Call()
      /usr/local/go/src/reflect/value.go:308 +0xc0
  github.com/stretchr/testify/suite.Run.func2()
      /home/smira/Documents/go/pkg/mod/github.com/stretchr/testify@v1.3.1-0.20190311161405-34c6fa2dc709/suite/suite.go:133 +0x2ec
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:865 +0x163

Previous write at 0x00c0000cf2f8 by goroutine 26:
  github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/goroutine.(*goroutineRunner).Run()
      /home/smira/Documents/autonomy/talos/internal/app/machined/pkg/system/runner/goroutine/goroutine.go:65 +0xcb
  github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/goroutine_test.(*GoroutineSuite).TestStop.func3()
      /home/smira/Documents/autonomy/talos/internal/app/machined/pkg/system/runner/goroutine/goroutine_test.go:104 +0x4a
```

### DIFF
--- a/internal/app/machined/pkg/system/runner/goroutine/goroutine.go
+++ b/internal/app/machined/pkg/system/runner/goroutine/goroutine.go
@@ -48,6 +48,8 @@ func NewRunner(data *userdata.UserData, id string, main FuncMain, setters ...run
 		opts: runner.DefaultOptions(),
 	}
 
+	r.ctx, r.ctxCancel = context.WithCancel(context.Background())
+
 	for _, setter := range setters {
 		setter(r.opts)
 	}
@@ -62,8 +64,6 @@ func (r *goroutineRunner) Open(ctx context.Context) error {
 
 // Run implements the Runner interface.
 func (r *goroutineRunner) Run(eventSink events.Recorder) error {
-	r.ctx, r.ctxCancel = context.WithCancel(context.Background())
-
 	r.wg.Add(1)
 	defer r.wg.Done()
 
@@ -111,6 +111,8 @@ func (r *goroutineRunner) Stop() error {
 	r.ctxCancel()
 
 	r.wg.Wait()
+
+	r.ctx, r.ctxCancel = context.WithCancel(context.Background())
 
 	return nil
 }


smira:
> Wonder why we don't run `-race` in `hack/golang/test.sh`. Seems like this should be a thing we run by default?

This is a long story, but in short race detector doesn't work under muslc. I remember there was even golang issue to fix that, but afaik it's not done.

I thought of two things:

1. Running `go test -race` via regular golang image. This doesn't cover containerd tests, but better than nothing.

2. Building race-enabled Talos image and using it for basic-integration.

bradbeam:
> This is a long story, but in short race detector doesn't work under muslc.

Ah yeah that one. 

> Running go test -race via regular golang image. This doesn't cover containerd tests, but better than nothing.

This might be a pretty good idea. With the recent changes you made to buildkit to speed things up, we should be able to run both tests as parallel without much impact to overall build times.

smira:
Race-enabled Talos image (just all the binaries compiled with `-race`) should be even more fun, as it might encounter races in the code not covered with tests...

PR #954, opened 2019-08-06
chore: add race-enabled test run

smira:
As Go race detector doesn't work under libmuscl, use stock glibc-based
golang container.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -145,6 +145,20 @@ steps:
     depends_on:
       - build.Lint
 
+  - name: test.UnitTestRace
+    image: autonomy/build-container:latest
+    pull: always
+    environment:
+      BUILDKIT_HOST: tcp://buildkitd.ci.svc:1234
+      BINDIR: /usr/local/bin
+    commands:
+      - make test-race
+    volumes:
+      - name: dockersock
+        path: /var/run/
+    depends_on:
+      - build.Lint
+
 # Phase 3
   - name: build.Rootfs
     image: autonomy/build-container:latest
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 # syntax = docker/dockerfile-upstream:1.1.2-experimental
 
+# Meta args applied to stage base names.
+
 ARG TOOLS
+ARG GO_VERSION
+
+# The tools target provides base toolchain for the build.
+
 FROM $TOOLS AS tools
 ENV PATH /toolchain/bin
 RUN ["/toolchain/bin/mkdir", "/bin", "/tmp"]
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -267,6 +267,16 @@ RUN --security=insecure --mount=type=cache,id=testspace,target=/tmp --mount=type
 FROM scratch AS test
 COPY --from=test-runner /src/coverage.txt /coverage.txt
 
+# The test-race target performs tests with race detector.
+
+FROM golang:1.12 AS test-race
```
smira: probably golang version here should be pushed to the Makefile

Dockerfile
```diff
@@ -267,6 +267,16 @@ RUN --security=insecure --mount=type=cache,id=testspace,target=/tmp --mount=type
...
+COPY --from=base /src /src
+COPY --from=base /go/pkg/mod /go/pkg/mod
```
smira: re-using already downloaded & verified modules

smira:
Go issue for the reference: https://github.com/golang/go/issues/14481

.drone.yml
```diff
@@ -190,6 +190,20 @@ steps:
...
+    environment:
+      BUILDKIT_HOST: tcp://buildkitd.ci.svc:1234
+      BINDIR: /usr/local/bin
+    commands:
+      - make test-race
+    volumes:
+      - name: dockersock
+        path: /var/run/
+    depends_on:
+      - lint
```
bradbeam: I think the depends_on will need to be changed to `rootfs` or `osd` and `proxyd` ( think those are the two containers the tests depend on )

smira: test-race can't run containerd because it doesn't have Talos rootfs, so in fact it's the same as doing `go test` locally on your development tree. 

I think it's not quite possible to mix those as race detector depends on glibc, and our code is all built with musl

Dockerfile
smira: I tried to but I can't make it accept `1.12` or full `golang:1.12` as a build arg, giving up. seems that it works only for the first `FROM` in the `Dockerfile`, but it breaks `FROM $tools` if I move this target about `tools`

.drone.yml
```diff
@@ -145,6 +145,20 @@ steps:
     depends_on:
       - build.Lint
 
+  - name: test.UnitTestRace
```
smira: not sure about the placement here. 

I placed it near the other steps with same deps, but at the same time it makes sense to place it close to `test.UnitTest` (without race)

smira:
Running in `golang` glibc-based container is ugly, but I think value of race-enabled unit-tests outweighs the ugliness of the solution. This shouldn't inflate build times, as tests run early while `rootfs` is being built, and some steps are going to be amortized over time (build cache, downloaded golang image, etc.)

I would really like to fix `tsan` (Thread Safety Analyzer) on `muslc` so that we can run race detector on Talos (including binaries, so that we can build Docker image `talos-race` and catch moar races!)

Dockerfile
smira: resolved the issue, also documented things a bit so that next one won't have confusion :)

PR #955, opened 2019-08-07
fix(initramfs): Allow data partition to grow

bradbeam:
This fix ensures that we always grow the data partition during an installation.

Signed-off-by: Brad Beam <brad.beam@talos-systems.com>

PR #960, opened 2019-08-08
fix(machined): Clean up installation process

bradbeam:
This also includes a fix for #955 which had the unintended side effect
of breaking image creation ( since it would attempt to grow the filesystem
always ).

The refactor standardizes around looking for the DATA and ESP labels to
discover any existing installations/filesystems. If none are found, an
installation will proceed -- for both image creation and bare metal.
During bootup, the DATA partition will always attempt to expand/grow.

### DIFF
--- a/internal/app/machined/internal/platform/baremetal/baremetal.go
+++ b/internal/app/machined/internal/platform/baremetal/baremetal.go
@@ -13,6 +13,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/talos-systems/talos/internal/pkg/installer"
 	"github.com/talos-systems/talos/internal/pkg/kernel"
+	"github.com/talos-systems/talos/internal/pkg/mount"
+	"github.com/talos-systems/talos/internal/pkg/mount/manager"
+	"github.com/talos-systems/talos/internal/pkg/mount/manager/owned"
 	"github.com/talos-systems/talos/pkg/blockdevice/probe"
 	"github.com/talos-systems/talos/pkg/constants"
 	"github.com/talos-systems/talos/pkg/userdata"
@@ -88,9 +91,27 @@ func (b *BareMetal) Initialize(data *userdata.UserData) (err error) {
 		return err
 	}
 
-	i := installer.NewInstaller(cmdline, data)
-	if err = i.Install(); err != nil {
-		return errors.Wrap(err, "failed to install")
+	// Attempt to discover a previous installation
+	// An err case should only happen if no partitions
+	// with matching labels were found
+	var mountpoints *mount.Points
+	mountpoints, err = owned.MountPointsFromLabels()
+	if err != nil {
+		// No previous installation was found, attempt an install
+		i := installer.NewInstaller(cmdline, data)
+		if err = i.Install(); err != nil {
+			return errors.Wrap(err, "failed to install")
+		}
+
+		mountpoints, err = owned.MountPointsFromLabels()
+		if err != nil {
+			return err
+		}
+	}
+
+	m := manager.NewManager(mountpoints)
+	if err = m.MountAll(); err != nil {
+		return err
 	}
 
 	return nil
--- a/internal/pkg/installer/bootloader/syslinux/syslinux.go
+++ b/internal/pkg/installer/bootloader/syslinux/syslinux.go
@@ -132,7 +132,7 @@ func WriteSyslinuxCfg(base, path string, syslinuxcfg *Cfg) (err error) {
 		return err
 	}
 
-	log.Println("writing syslinux.cfg to disk")
+	log.Printf("writing %s to disk", path)
 	if err = ioutil.WriteFile(path, wr.Bytes(), 0600); err != nil {
 		return err
 	}
--- a/internal/pkg/installer/installer.go
+++ b/internal/pkg/installer/installer.go
@@ -7,7 +7,6 @@ package installer
 import (
 	"io"
 	"io/ioutil"
-	"log"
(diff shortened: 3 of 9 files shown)


internal/app/machined/internal/platform/baremetal/baremetal.go
```diff
@@ -88,9 +91,27 @@ func (b *BareMetal) Initialize(data *userdata.UserData) (err error) {
 		return err
 	}
 
-	i := installer.NewInstaller(cmdline, data)
-	if err = i.Install(); err != nil {
-		return errors.Wrap(err, "failed to install")
+	// Attempt to discover a previous installation
```
andrewrynhard: This feels like a very heavy pattern to have to follow for every cloud platform. It is only just a copy paste thing, but I was hoping we could come up with something that keeps things simple.

internal/app/machined/internal/platform/baremetal/baremetal.go
```diff
@@ -88,9 +91,27 @@ func (b *BareMetal) Initialize(data *userdata.UserData) (err error) {
...
+	// An err case should only happen if no partitions
+	// with matching labels were found
+	var mountpoints *mount.Points
+	mountpoints, err = owned.MountPointsFromLabels(true)
+	if err != nil {
```
andrewrynhard: I don't think determining if an installation has occurred from this error is the right route. The probing logic is complex and it could really be any number of issues that causes the probing to fail.

internal/pkg/installer/bootloader/syslinux/syslinux.go
```diff
@@ -106,6 +106,8 @@ func Install(base string, config interface{}) (err error) {
 
 	paths := []string{filepath.Join(base, "syslinux", "syslinux.cfg"), filepath.Join(base, "EFI", "syslinux", "syslinux.cfg")}
 	for _, path := range paths {
+		// TODO seems like there's some duplication bits around syslinux labels here
```
andrewrynhard: It installs for EFI and BIOS.

internal/pkg/installer/installer.go
```diff
@@ -216,37 +188,3 @@ func wipe(manifest *manifest.Manifest) (err error) {
 
 	return nil
 }
-
-func exists(devpath string) (bool, error) {
```
andrewrynhard: IMO this should not be removed. Determining if an installation is present based solely on labels isn't reliable. I also had plans in the back on my mind to expand on the file that gets dropped into the boot partition to include some metadata about the install.

internal/pkg/mount/manager/owned/owned.go
```diff
@@ -14,59 +14,33 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// MountPointsForDevice returns the mountpoints required to boot the system.
-func MountPointsForDevice(devpath string) (mountpoints *mount.Points, err error) {
+// MountPointsFromLabels returns the mountpoints required to boot the system.
+func MountPointsFromLabels(grow bool) (mountpoints *mount.Points, err error) {
```
andrewrynhard: Not a big fan of the boolean argument. This function is used in a number of places, and I feel like we shouldn't have to specify whether or not the partition should be resized every time its' invoked.

internal/app/machined/internal/platform/baremetal/baremetal.go
bradbeam: This is only implemented for bare-metal like platforms where we need to perform an install inline. VM based cloud providers don't have any of this additional work ( but all do call `owned.MountPointsFromLabels()`).

internal/app/machined/internal/platform/baremetal/baremetal.go
bradbeam: decision was based on the error being thrown here. 
https://github.com/talos-systems/talos/blob/master/internal/pkg/mount/manager/owned/owned.go#L68

It seemed reasonable that if probe stuff failed for a more specific ( bad disk? ) reason that it'd fail during the installation process when we try to read/write to the disk.

internal/pkg/installer/installer.go
bradbeam: > Determining if an installation is present based solely on labels isn't reliable

Why's that? The only instance I can think of is if the installation itself failed. But now that we minimally use the disk, I'm not sure there's much risk here since it's mostly ephemeral.

internal/pkg/mount/manager/owned/owned.go
bradbeam: It's not super heavy, and there's definitely room to DRY it a bit. 
```internal//app/machined/internal/platform/aws/aws.go
internal//app/machined/internal/platform/azure/azure.go
internal//app/machined/internal/platform/baremetal/baremetal.go
internal//app/machined/internal/platform/googlecloud/googlecloud.go
internal//app/machined/internal/platform/packet/packet.go
internal//app/machined/internal/platform/vmware/vmware.go
internal//pkg/installer/installer.go
internal//pkg/mount/manager/owned/owned.go```

We could always jiggle it with variadic parameters if we want to make clever, but not sure that's really needed.

bradbeam:
> I think this PR needs to be rethought. If we depend only on MountPointsFromLabels then creating images in our CI is going to break. We need the ability to be explicit about the device and to skip looking up by label. Unless I'm missing something, I think this will break something like make talos-gce if its ran on a Talos node.

Didn't have any issues running it through CI to create the images, https://ci.dev.talos-systems.io/talos-systems/talos/2189/1/6

PR #964, opened 2019-08-08
chore: remove machined from rootfs target

andrewrynhard:
The machined dependency is not needed in the rootfs target. The
dependency is handled by buildkit.


bradbeam:
Forgot machined runs from a local binary instead of container 
ref https://github.com/talos-systems/talos/blob/master/Dockerfile#L191

PR #967, opened 2019-08-09
chore: build drone YAML via jsonnet

andrewrynhard:
This PR aims to DRY the drone config file by using Jsonnet to generate
it.


Makefile
```diff
@@ -196,50 +181,71 @@ talos-azure:
...
+.PHONY: iso
+iso:
+	@docker run --rm -i -v $(PWD)/build:/out autonomy/installer:$(TAG) iso
+
+.PHONY: container
+container: buildkitd
 	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
 		build \
-		--output type=docker,dest=build/$@.tar,name=docker.io/autonomy/$@:$(TAG) \
+		--output type=docker,dest=build/$@.tar,name=docker.io/autonomy/talos:$(TAG) \
```
bradbeam: just to verify, `dest=build/$@.tar` ( build/container.tar? ) is what we want, not `talos.tar`?

Makefile
```diff
@@ -196,50 +181,71 @@ talos-azure:
...
-.PHONY: e2e
+.PHONY: e2e-integration
 e2e-integration:
 	@TAG=$(TAG) ./hack/test/$@.sh
 
-.PHONY: test
-test: buildkitd
+.PHONY: unit-tests
+unit-tests: buildkitd
+	@mkdir -p build
```
bradbeam: doing `mkdir build` for unit tests seems unnecessary?

hack/drone.jsonnet
```diff
@@ -0,0 +1,323 @@
...
+local conformance = e2e_setup + [
+  conformance_azure,
+  conformance_gce,
+];
+
+
+local release_step ={
+  name: "release",
+  image: "plugins/github-release",
+  settings:{
```
bradbeam: did you want to try to include the `note: ` bit to automatically pull in the ami names?

hack/drone.jsonnet
```diff
@@ -0,0 +1,323 @@
...
+
+
+// CAPI.
+
+local push_image_azure = step("push-image-azure", [image_azure], creds_env_vars);
+local push_image_gce = step("push-image-gce", [image_gce], creds_env_vars);
+
+// Tests
+
+local lint = step("lint", [rootfs]);
```
bradbeam: why move lint to depend on rootfs and not clone?

hack/drone.jsonnet
```diff
@@ -0,0 +1,323 @@
...
+// Steps
+
+local apps = [
+    machined,
+    osd,
+    trustd,
+    proxyd,
+    ntpd,
+    osctl_linux,
+    osctl_darwin,
```
bradbeam: nit, but having osctl_darwin here causes it to get built during e2e tests which isn't needed. not super significant and probably causes more complexity to filter it out

Makefile
andrewrynhard: yeah, every time I have to explain to people that `talos.tar` is the container. this is to just make it more clear as an artifact that its the container.

Makefile
andrewrynhard: Not sure why that is there TBH. Its nothing new, just a rename of `test` to `unit-test`. We can probably remove it.

hack/drone.jsonnet
andrewrynhard: Shimmying that in to the way the `step` function works made things super ugly. In the `step` function I'm hoping to force us to use a single command for _most_ things. I found that it makes the steps a lot more clear and helps align the `Makefile` and `drone.jsonnet` file.

hack/drone.jsonnet
andrewrynhard: I was thinking that we could group all the "tests" to run at the same time, but its probably better to just depend on `clone`. I will update.

hack/drone.jsonnet
andrewrynhard: Yeah, it really makes the implementation uglier if we try to pull that bit out. I tried to design in in such a way that each pipeline sort of builds on the previous one. Plus this gets us closer to running a more complete e2e test.

hack/drone.jsonnet
andrewrynhard: Plus with everything we talked about today in regard to publish and AWS asset instead of the AMIs, this felt less important.

andrewrynhard:
I wanna iterate on this over the weekend and see how it performs in real life. Gonna merge, and follow up with some more cleanup as I figure out the best way to structure the Jsonnet code.

PR #952, opened 2019-08-06
chore(ci): Add e2e promotion pipeline

bradbeam:
Signed-off-by: Brad Beam <brad.beam@talos-systems.com>

PR #968, opened 2019-08-09
chore: Wait a little after cluster comes up

bradbeam:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -502,6 +502,7 @@ steps:
       BUILDKIT_HOST: ${BUILDKIT_HOST=tcp://buildkitd.ci.svc:1234}
       BINDIR: /usr/local/bin
       PLATFORM: azure
+      CONFORMANCE: run
     commands:
       - make e2e-integration
     volumes:
@@ -541,6 +542,7 @@ steps:
       BUILDKIT_HOST: ${BUILDKIT_HOST=tcp://buildkitd.ci.svc:1234}
       BINDIR: /usr/local/bin
       PLATFORM: gce
+      CONFORMANCE: run
     commands:
       - make e2e-integration
     volumes:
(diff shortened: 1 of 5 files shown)


hack/test/osctl-cluster-create.sh
```diff
@@ -53,6 +53,9 @@ run "timeout=\$((\$(date +%s) + ${TIMEOUT}))
 	   sleep 5
 	 done"
 
+## Give a moment for things to stabilize before applying manifests
+sleep 10
```
andrewrynhard: Not for this PR, but curious what we can wait on via `kubectl wait` before we apply this. Only started seeing this after 1.16 move.

.drone.yml
```diff
@@ -488,6 +488,7 @@ steps:
       BUILDKIT_HOST: ${BUILDKIT_HOST=tcp://buildkitd.ci.svc:1234}
       BINDIR: /usr/local/bin
       PLATFORM: azure
+      CONFORMANCE: run
```
bradbeam: Accidentally dropped these in #952

hack/test/conformance.sh
```diff
@@ -7,19 +7,23 @@ source ./hack/test/e2e-runner.sh
 echo "Master CIS Conformance:"
 e2e_run "export KUBECONFIG=${KUBECONFIG}-${PLATFORM}-capi
          kubectl apply -f /e2emanifests/cis-kube-bench-master.yaml
-         kubectl wait --timeout=60s --for=condition=complete job/kube-bench-master > /dev/null
+         kubectl wait --timeout=300s --for=condition=complete job/kube-bench-master > /dev/null
```
bradbeam: Seeing issues in azure e2e where 60s isn't enough.

hack/test/e2e-integration.sh
```diff
@@ -85,8 +86,7 @@ e2e_run "timeout=\$((\$(date +%s) + ${TIMEOUT}))
 		 done"
 
 ## Run conformance tests if var is not null
-if [ -n "${CONFORMANCE}" ]
-then
+if [ ${CONFORMANCE:-"dontrun"} == "run" ]; then
```
bradbeam: -n conformance fails with our options above ( set -eou pipefail ), so we need to ensure it has some default value so they can be properly skipped

PR #971, opened 2019-08-09
fix: stalls in local Docker cluster boot

smira:
Problem was triggered by udevd trigger, root cause is not clear, but
workaround is to disable it for container mode.

Implement CPU/mem limits for `osctl cluster create`, apply defaults,
bump defaults for cicd.

### DIFF
--- a/internal/app/machined/internal/phase/services/services.go
+++ b/internal/app/machined/internal/phase/services/services.go
@@ -23,17 +23,19 @@ func NewServicesTask() phase.Task {
 
 // RuntimeFunc returns the runtime function.
 func (task *Services) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
-	return task.runtime
+	return func(platform platform.Platform, data *userdata.UserData) error {
+		return task.runtime(data, mode)
+	}
 }
 
-func (task *Services) runtime(platform platform.Platform, data *userdata.UserData) (err error) {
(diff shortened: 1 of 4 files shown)


internal/app/machined/internal/phase/services/services.go
```diff
@@ -23,28 +23,38 @@ func NewServicesTask() phase.Task {
 
 // RuntimeFunc returns the runtime function.
 func (task *Services) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
-	return task.runtime
```
andrewrynhard: Pretty cool to see phases/tasks/runtime mode make this super simple.

hack/test/osctl-cluster-create.sh
```diff
@@ -27,7 +27,7 @@ run() {
 	 	k8s.gcr.io/hyperkube:${KUBERNETES_VERSION} -c "${1}"
 }
 
-${OSCTL} cluster create --name integration --image ${TALOS_IMG} --mtu 1440
+${OSCTL} cluster create --name integration --image ${TALOS_IMG} --mtu 1440 --cpus 4.0
```
andrewrynhard: Sorry. My refactor moved this to the `basic-integration.sh` script.

smira: wow, git was able to rebase it automatically

PR #982, opened 2019-08-11
fix: enclose server address is bracks if IPv6

### LINKED ISSUE #980: [ipv6] osctl config: need brackets around IP when IPv6
When the target is an IPv6 address, it needs to be enclosed with square brackets before appending the port number.

This is when _using_ `osctl` as a client, rather than as a config generator.

Ulexus:
Fixes #980


Ulexus:
This may not be the best place or name for the common function `FormatAddress`.  Open to better ideas.

andrewrynhard:
This works for now as we figure out the ipv6 story.

PR #990, opened 2019-08-11
fix(proxyd): fail to proxy after 10 attempts

andrewrynhard:
Recursion and never failing cause the proxy logic to high memory usage.


andrewrynhard:
The plan is to remove proxyd in 0.3.

PR #995, opened 2019-08-11
refactor: add more runtime modes

andrewrynhard:
In order to DRY up all installation methods and mount methods, this PR
introduces a few more runtime modes. The modes are then used to
determine the strategy for creating and or mounting the paritions.


internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -25,6 +40,101 @@ func (task *Platform) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
 }
 
 func (task *Platform) runtime(platform platform.Platform, data *userdata.UserData) (err error) {
+	switch platform.Mode() {
```
andrewrynhard: This file is the gist of things. I plan on cleaning this up a bit more but wanted to put the PR up so we can all talk about the general direction some time this week.

internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -25,6 +40,101 @@ func (task *Platform) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
...
+		unix.Reboot(int(unix.LINUX_REBOOT_CMD_RESTART))
+	case runtime.BareMetal:
+		var endpoint *string
+		if endpoint = kernel.ProcCmdline().Get(constants.KernelParamUserData).First(); endpoint == nil {
+			return errors.Errorf("failed to find %s in kernel parameters", constants.KernelParamUserData)
+		}
+
+		cmdline := kernel.NewDefaultCmdline()
+		cmdline.Append("initrd", filepath.Join("/", "default", "initramfs.xz"))
+		cmdline.Append(constants.KernelParamPlatform, "packet")
```
bradbeam: i think we'd want to drop this param.

bradbeam: or `s/packet/platform.Name()/`

internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -25,6 +40,101 @@ func (task *Platform) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
...
+		// nolint: errcheck
+		unix.Reboot(int(unix.LINUX_REBOOT_CMD_RESTART))
+	case runtime.BareMetal:
+		var endpoint *string
+		if endpoint = kernel.ProcCmdline().Get(constants.KernelParamUserData).First(); endpoint == nil {
+			return errors.Errorf("failed to find %s in kernel parameters", constants.KernelParamUserData)
+		}
+
+		cmdline := kernel.NewDefaultCmdline()
+		cmdline.Append("initrd", filepath.Join("/", "default", "initramfs.xz"))
```
bradbeam: not directly related to this pr, but do you remember why bare metal needs to handle the above differently? is it for efi? wonder if we can collapse it into defaultcmdline.

internal/app/machined/internal/phase/platform/platform.go
andrewrynhard: Thanks. Yeah this whole bit here needs cleanup. Still a bit of a PoC and hesitant to put more work into if its not a direction we wanna go.

internal/app/machined/internal/phase/platform/platform.go
andrewrynhard: This is just leftover from how we used to do root-a root-b. And it was left here because we discussed having support for multiple versions of Talos, but we haven't done anything with that idea yet. If we want one and only one then we can make this part of the default args.

internal/app/machined/internal/runtime/initializer/baremetal/baremetal.go
```diff
@@ -0,0 +1,70 @@
...
+// Initialize implements the Initializer interface.
+func (b *Baremetal) Initialize(platform platform.Platform, data *userdata.UserData) (err error) {
+	var endpoint *string
+	if endpoint = kernel.ProcCmdline().Get(constants.KernelParamUserData).First(); endpoint == nil {
+		return errors.Errorf("failed to find %s in kernel parameters", constants.KernelParamUserData)
+	}
+
+	cmdline := kernel.NewDefaultCmdline()
+	cmdline.Append("initrd", filepath.Join("/", "default", "initramfs.xz"))
+	cmdline.Append(constants.KernelParamPlatform, strings.ToLower(platform.Name()))
```
andrewrynhard: `strings.ToLower(platform.Name())` forced the rename of the `bare-metal` platform. This was a good time to align some of the other outstanding TODOs (like rename `googlecloud`  to `gcp`. GCP is technically the correct name if we are going to name the platform after the over cloud provider service (e.g AWS instead of EC2, Azure instead of Virtual Machines).

andrewrynhard:
> only nit i have is in the phases we have a combination of `task.standard` and `task.runtime`. it's not directly a result of this pr, so we could handle it in a followup one, but we should standardize on what the 'default' task name is. personally i like `task.standard` best
Agree. That was bugging me throughout this PR.

PR #997, opened 2019-08-11
fix(proxyd): do not pre-bracket IPv6 backend addrs

### LINKED ISSUE #996: proxyd now double-brackets backend IPv6 addresses
`net.JoinHostPort()` adds brackets to already-bracketed IPv6 addresses, resulting in double brackets.  This is actually probably a Go base library bug.  We can fix it here just by _not_ bracketing the address before it goes to `JoinHostPort()`.

Ulexus:
Fixes #996


andrewrynhard:
@Ulexus any way I can get you to squeeze in that log print we thought would have saved us some pain in discovering this?

PR #1002, opened 2019-08-12
chore: run unique E2E tests

andrewrynhard:
In order to run more than one instance of E2E testing at a time, we need
to ensure that all resources are unique to the run.

### DIFF
--- a/hack/test/basic-integration.sh
+++ b/hack/test/basic-integration.sh
@@ -41,6 +41,21 @@ run "timeout=\$((\$(date +%s) + ${TIMEOUT}))
 	   sleep 2
 	 done"
 
+## Wait for the init node to report in
+run "timeout=\$((\$(date +%s) + ${TIMEOUT}))
+     until kubectl get node master-1 >/dev/null
+	 do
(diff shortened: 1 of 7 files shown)


hack/test/basic-integration.sh
```diff
@@ -41,9 +41,9 @@ run "timeout=\$((\$(date +%s) + ${TIMEOUT}))
 	   sleep 2
 	 done"
 
-## Wait for all nodes to report in
+## Wait for the init node to report in
```
andrewrynhard: This is a workaround for the issues we are seeing with `kube-proxy`.

hack/test/gce-setup.sh
```diff
@@ -5,14 +5,14 @@ set -eou pipefail
 ## Setup svc acct
 echo $GCE_SVC_ACCT | base64 -d > /tmp/svc-acct.json
 apk add --no-cache python
-wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-253.0.0-linux-x86_64.tar.gz
```
andrewrynhard: This was causing a dirty working tree.

PR #1014, opened 2019-08-14
chore: update go modules to use Kubernetes v1.16.0-alpha.3

andrewrynhard:
This is not ideal, but it works. We essentially need to start using
replace statements in order to pull in the modules we need.


internal/app/machined/internal/phase/userdata/userdata.go
```diff
@@ -63,9 +63,7 @@ func (task *UserData) container(platform platform.Platform, data *userdata.UserD
 	if ok {
 		initConfiguration.ClusterConfiguration.ComponentConfigs.Kubelet.FailSwapOn = false
 		// See https://github.com/kubernetes/kubernetes/issues/58610#issuecomment-359552443
-		max := int32(0)
 		maxPerCore := int32(0)
-		initConfiguration.ClusterConfiguration.ComponentConfigs.KubeProxy.Conntrack.Max = &max
```
andrewrynhard: Deprecated.

go.mod
```diff
@@ -78,37 +58,51 @@ require (
...
+	k8s.io/apimachinery v0.0.0
+	k8s.io/client-go v0.0.0
+	k8s.io/cri-api v0.0.0
+	k8s.io/kubernetes v1.16.0-alpha.3
+	k8s.io/utils v0.0.0 // indirect
+)
+
+replace (
+	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+	github.com/opencontainers/runtime-spec v1.0.1 => github.com/opencontainers/runtime-spec v0.1.2-0.20180301181910-fa4b36aa9c99
```
andrewrynhard: This and the one above are workarounds since container does not have modules support yet.

PR #947, opened 2019-08-03
Refactor(networkd): Replace networkd with a standalone app

bradbeam:
Signed-off-by: Brad Beam <brad.beam@talos-systems.com>

PR #1017, opened 2019-08-15
feat: Allow hostname to be specified in userdata

bradbeam:
This sets up the ability to define hostname via userdata. I dont expect
this will get used publicly much, but provides a mechanism to convey
the hostname from various sources internally.

This also fixes the hostname retrieval from metadata services in azure.


bradbeam:
Intentionally leaving the network code as is since that should get addressed in #947

PR #1021, opened 2019-08-15
feat: rename DATA partition to EPHEMERAL

andrewrynhard:


andrewrynhard:
> Not something for now, but we should consider what a migration path/strategy would or should look like.

For sure. Was thinking by v0.3 we should have these types of things in place. That would be a good goal for us.

PR #1022, opened 2019-08-15
fix: verify installation definition

andrewrynhard:
This fixes the possibility of panicing on a nil pointer by running the
verification steps earlier.


internal/pkg/installer/installer.go
```diff
@@ -70,16 +54,6 @@ func (i *Installer) Install() (err error) {
 		return nil
 	}
 
-	// Verify that the target device(s) can satisify the requested options.
-
-	if err = VerifyBootDevice(i.data); err != nil {
```
andrewrynhard: These need to be moved into the `NewManifest` function so that we can verify the machine config early and avoid nil pointers.

PR #1023, opened 2019-08-15
feat(proxyd): Add gRPC server

bradbeam:
Part of the API refactor; this introduces a gRPC server for proxyd
to expose some of the internal state.

### DIFF
--- a/internal/app/proxyd/main.go
+++ b/internal/app/proxyd/main.go
@@ -8,9 +8,13 @@ import (
 	"context"
 	"flag"
 	"log"
+	"sync"
 
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/conditions"
 	"github.com/talos-systems/talos/internal/app/proxyd/internal/frontend"
+	"github.com/talos-systems/talos/internal/app/proxyd/internal/reg"
+	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/grpc/factory"
 	"github.com/talos-systems/talos/pkg/startup"
 	"github.com/talos-systems/talos/pkg/userdata"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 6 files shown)


internal/app/proxyd/main.go
```diff
@@ -48,46 +52,76 @@ func main() {
...
-	}()
 
-	// nolint: errcheck
-	r.Listen(":443")
+	}(r, &wg)
+
+	wg.Wait()
 }
 
-func init() {
```
bradbeam: we had two `init()` defined

internal/app/proxyd/main.go
```diff
@@ -48,46 +52,76 @@ func main() {
...
-		}
+	var wg sync.WaitGroup
 
-		// Discover local non loopback ips
-		ips, err := pkgnet.IPAddrs()
-		if err != nil {
-			log.Fatalf("failed to get local address: %v", err)
+	wg.Add(1)
+	// Start up reverse proxy
+	go func(r *frontend.ReverseProxy, wg *sync.WaitGroup) {
```
andrewrynhard: We can simplify this with the ListenAndServe func in the factory. Checkout trustd.

internal/app/proxyd/main.go
```diff
@@ -48,46 +52,70 @@ func main() {
...
-
-		// Overwrite defined host so we can target local apiserver
-		// and bypass the admin.conf host which is configured for proxyd
-		config.Host = ip.String() + ":6443"
-
-		clientset, err := kubernetes.NewForConfig(config)
+	}(r, &wg)
+
+	wg.Add(1)
+	// Start up gRPC server
```
andrewrynhard: This wait group isn’t needed. ListenAndServe will block

bradbeam: it's so we can launch both as goroutines and have a consistency across both listeners instead of doing
```
go r.Listen()
factory.ListenAndServe()
```

andrewrynhard: What you posted ^ looks a lot simpler and easier to follow IMO. Not a blocker for me but what you posted above is the more common approach.

PR #1026, opened 2019-08-16
feat: Add gRPC server for ntp

bradbeam:
Also migrated ntp functionality to sub package for better organization. 

### DIFF
--- a/internal/app/ntpd/main.go
+++ b/internal/app/ntpd/main.go
@@ -7,11 +7,11 @@ package main
 import (
 	"flag"
 	"log"
-	"math/rand"
-	"syscall"
-	"time"
 
-	"github.com/beevik/ntp"
+	"github.com/talos-systems/talos/internal/app/ntpd/pkg/ntp"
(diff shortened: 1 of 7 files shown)


internal/app/ntpd/main.go
```diff
@@ -56,77 +54,21 @@ func main() {
...
-// SetTime sets the system time based on the query response
-func (n *NTP) SetTime() error {
-	// Not sure if this is the right thing to do
-	if n.Response == nil {
-		return nil
-	}
-	timeval := syscall.NsecToTimeval(n.Response.Time.UnixNano())
-	return syscall.Settimeofday(&timeval)
+	log.Println("Starting ntpd")
+	errch := make(chan error)
```
bradbeam: @andrewrynhard this pattern any better than proxyd?

andrewrynhard: Yeah I like this a lot better 👍. This is nice because we can handle an error from either.

PR #1036, opened 2019-08-17
chore: Retry reread partition table if EBUSY

bradbeam:
Should help make it more robust

### DIFF
--- a/pkg/blockdevice/blockdevice.go
+++ b/pkg/blockdevice/blockdevice.go
@@ -8,6 +8,8 @@ package blockdevice
 import (
 	"bytes"
 	"os"
+	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/blockdevice/blockdevice.go
```diff
@@ -107,8 +109,16 @@ func (bd *BlockDevice) RereadPartitionTable() error {
 		return errors.Errorf("flush block device buffers: %v", ret)
 	}
 	// Reread the partition table.
-	if _, _, ret := unix.Syscall(unix.SYS_IOCTL, bd.f.Fd(), unix.BLKRRPART, 0); ret != 0 {
-		return errors.Errorf("re-read partition table: %v", ret)
+	for i := 0; i < 50; i++ {
+		if _, _, ret := unix.Syscall(unix.SYS_IOCTL, bd.f.Fd(), unix.BLKRRPART, 0); ret != 0 {
+			switch ret {
+			case syscall.EBUSY:
```
andrewrynhard: Think we still need to return the EBUSY in the case that all 50 tries fail. A simple check of the index in this block should work. If last attempt return error otherwise continue.

PR #1037, opened 2019-08-17
chore: Make installer script more robust

bradbeam:
Adds some functionality to help prevent collisions.

- Run `losetup -f <attach>` to atomically provision the next loopback
  device with a file
- Creates a tempfile for default and test named raw images to prevent
  overlap

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ base: buildkitd
 generate: buildkitd
 	$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
 		build \
-    --output type=local,dest=./ \
+		--output type=local,dest=./ \
 		--opt target=$@ \
 		$(COMMON_ARGS)
 
@@ -124,7 +124,7 @@ generate: buildkitd
 kernel: buildkitd
 	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
 		build \
-    --output type=local,dest=build \
+		--output type=local,dest=build \
 		--opt target=$@ \
 		$(COMMON_ARGS)
 	@-rm -rf ./build/modules
@@ -133,7 +133,7 @@ kernel: buildkitd
 initramfs: buildkitd
 	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
 		build \
-    --output type=local,dest=build \
+		--output type=local,dest=build \
 		--opt target=$@ \
 		$(COMMON_ARGS)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -201,7 +208,7 @@ push-image-gce:
 
 .PHONY: image-test
 image-test:
-	@docker run --rm -v /dev:/dev -v /tmp:/out --privileged $(DOCKER_ARGS) autonomy/installer:$(TAG) install -n test -r -p test -u none
```
bradbeam: Dropping the output dir mount because we don't really care about the image afterwards.

hack/installer/entrypoint.sh
```diff
@@ -16,15 +16,25 @@ EOF
...
     rm ${TALOS_RAW}
   fi
+
+  # If defaults or a test file is specified, make a temp file
+  # to prevent collisions.
+  if [ ${TALOS_RAW} == "/out/talos.raw" ] || \
+     [ ${TALOS_RAW} == "/out/test.raw" ]; then
+    TALOS_RAW="$(mktemp ${TALOS_RAW}.XXXXXX)"
+    RENAME_TEMPFILE=true
+  fi
```
bradbeam: Still not sure how I feel about the above.

Makefile
andrewrynhard: Let’s keep the output dir. I can still see the value in being able to run something like fdisk on the test image. Plus it more accurately tests the way we generate images.

hack/installer/entrypoint.sh
andrewrynhard: Agree.

hack/installer/entrypoint.sh
```diff
@@ -16,15 +16,25 @@ EOF
...
+
   dd if=/dev/zero of="${TALOS_RAW}" bs=1M count=0 seek=544
-  DEVICE=$(losetup -f)
+
   # NB: Since we use BLKRRPART to tell the kernel to re-read the partition
   # table, it is required to create a partitioned loop device. The BLKRRPART
   # command is meaningful only for partitionable devices.
-  losetup -P ${DEVICE} ${TALOS_RAW}
+  DEVICE=$(losetup --find --partscan --nooverlap --show ${TALOS_RAW})
```
andrewrynhard: This is probably the only thing we want to change in this script. All the other changes seem related to things we need to do if the test image is not output and I think we should output the test image. Also not sure I see much value in the rename stuff.

bradbeam:
While I'm not married to some of the temp file/renaming/etc bits, it does solve a use case. 

If I run multiple `make image-test` commands concurrently, I'll run into collisions in some form or another
because each command will operate on the same raw disk image. In the `setup_raw_device` func, we remove and recreate the file, which depending on the timing could nuke an already mounted image, could overwrite an image in the middle of installation, etc and cause subsequent failures. 

I _think_ given how our ci setup is, we should create `azure.raw` (azure), `talos.raw` (test), and `disk.raw` (gce), and each pipeline should run on isolated storage, so I don't expect we'll run into an issue there. 

I'll drop the other changes till we can get something better in place.

andrewrynhard:
> While I'm not married to some of the temp file/renaming/etc bits, it does solve a use case.
> 
> If I run multiple `make image-test` commands concurrently, I'll run into collisions in some form or another
> because each command will operate on the same raw disk image. In the `setup_raw_device` func, we remove and recreate the file, which depending on the timing could nuke an already mounted image, could overwrite an image in the middle of installation, etc and cause subsequent failures.
> 
> I _think_ given how our ci setup is, we should create `azure.raw` (azure), `talos.raw` (test), and `disk.raw` (gce), and each pipeline should run on isolated storage, so I don't expect we'll run into an issue there.
> 
> I'll drop the other changes till we can get something better in place.

Gotcha. Yeah definitely possible. Extreme edge case but yeah this is true. Agree, let's sit on this since the likelihood of running two test runs at once is very low.

Issue #976, opened 2019-08-10
Return error with process stdout


andrewrynhard:
When execing processes we do not set the stdout of the cmd. We should create a temporary buffer and in the case of an error set, return an error with the commands output.

PR #1039, opened 2019-08-17
feat: add standardized command runner

andrewrynhard:
This adds a command runner function that can be used everywhere we need
to exec a binary. It adds addtional logic around error handling that
will allow for viewing errors in the case of a failed command.


andrewrynhard:
Closes #976.

PR #1040, opened 2019-08-17
chore: Fix reread error value on retry

bradbeam:
Should prevent a flake with returning an error when
it actually succeeded.

Believe previously if we hit a case where we needed to retry and then succeeded we'd still throw an error from being previously set. 

### DIFF
--- a/pkg/blockdevice/blockdevice.go
+++ b/pkg/blockdevice/blockdevice.go
@@ -109,18 +109,24 @@ func (bd *BlockDevice) RereadPartitionTable() error {
 		return errors.Errorf("flush block device buffers: %v", ret)
 	}
 
+	var (
+		err error
+		ret syscall.Errno
+	)
+
 	// Reread the partition table.
-	var err error
 	for i := 0; i < 50; i++ {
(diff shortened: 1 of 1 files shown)


pkg/blockdevice/blockdevice.go
```diff
@@ -112,15 +112,17 @@ func (bd *BlockDevice) RereadPartitionTable() error {
...
-		if _, _, ret := unix.Syscall(unix.SYS_IOCTL, bd.f.Fd(), unix.BLKRRPART, 0); ret != 0 {
-			err = errors.Errorf("re-read partition table: %v", ret)
-			switch ret {
-			case syscall.EBUSY:
-				time.Sleep(100 * time.Millisecond)
-				continue
-			default:
-				return err
-			}
+		if _, _, ret := unix.Syscall(unix.SYS_IOCTL, bd.f.Fd(), unix.BLKRRPART, 0); ret == 0 {
```
andrewrynhard: `ret` will be local to this function in this case. Take a look at https://github.com/talos-systems/talos/blob/a116145c1b1e7763d1d3b866de9e7dd3e5bae990/internal/pkg/mount/mount.go#L100-L110. Might be good to align with that one. The part that makes it work is the early `return nil`.

PR #1042, opened 2019-08-17
chore(ci): Only push `latest` tags if branch is master.

### LINKED ISSUE #956: Only push latest tags on master branch
As it currently stands, we push latest tags for all branch merges. This leads to a case where we're pushing a latest installer from the release-0.1 branch and causing all kinds of e2e head scratching.

bradbeam:
Should prevent flakes when we merge fixes on release branches where they unintentionally
get tagged as `latest`.

Closes #956 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ CONFORM_VERSION ?= 57c9dbd
 
 SHA ?= $(shell $(BINDIR)/gitmeta git sha)
 TAG ?= $(shell $(BINDIR)/gitmeta image tag)
+BRANCH ?= $(shell $(BINDIR)/gitmeta git branch)
 
 COMMON_ARGS = --progress=plain
 COMMON_ARGS += --frontend=dockerfile.v0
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -332,12 +333,12 @@ login:
 push: gitmeta login
 	@docker tag autonomy/installer:$(TAG) autonomy/installer:latest
```
andrewrynhard: Should this be in the if statement too or do we depend on `latest` tag somewhere?

Makefile
```diff
@@ -332,12 +333,12 @@ login:
...
 	@docker push autonomy/installer:$(TAG)
-	# TODO: only push :latest if merge to master?
-	@docker push autonomy/installer:latest
 	@docker tag autonomy/talos:$(TAG) autonomy/talos:latest
```
andrewrynhard: Same ^

Makefile
bradbeam: Had the targets backwards in my head for docker tag.

andrewrynhard: 👍

PR #1047, opened 2019-08-18
chore(ci): Increase timeout for kube-bench readiness

bradbeam:
Maybe address some of the CI failures waiting for kube-bench to come up


andrewrynhard:
I didn't realize it was already at 5 minutes. In reality this thing should take less than 30 seconds to run. Wonder if your need to add another wait somewhere.

PR #1048, opened 2019-08-18
fix(machined): Remove host mounts for specific CNI providers

### LINKED ISSUE #1046: [calico]: /var/lib/calico is never created
During neither installation nor machined runtime is `/var/lib/calico` created.  However, when the kubelet is attempted to be run, it tries (and fails) to mount `/var/lib/calico` is the CNI is set to calico.

bradbeam:
We shouldnt need these anymore

Fixes #1046 


andrewrynhard:
Super confused how this has even been working for us.

Ulexus:
Works for me

PR #1057, opened 2019-08-23
chore: upgrade containerd to v1.2.8

andrewrynhard:
This pulls in the latest build containerd binaries and updates our
dependencies.


andrewrynhard:
1.2.8 seems to have a bug in how its building the image digest. Closing.

PR #1062, opened 2019-08-25
feat: add overlay task

andrewrynhard:
This adds a well defined task for handling all overlay mount points that
are required by the system.

### DIFF
--- a/internal/app/machined/internal/phase/rootfs/mount_overlays.go
+++ b/internal/app/machined/internal/phase/rootfs/mount_overlays.go
@@ -5,24 +5,13 @@
 package rootfs
 
 import (
-	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
-
-	"github.com/pkg/errors"
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/app/machined/internal/platform"
 	"github.com/talos-systems/talos/internal/app/machined/internal/runtime"
+	"github.com/talos-systems/talos/internal/pkg/mount"
+	"github.com/talos-systems/talos/internal/pkg/mount/manager"
+	"github.com/talos-systems/talos/internal/pkg/mount/manager/overlay"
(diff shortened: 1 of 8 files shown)


internal/app/machined/internal/phase/rootfs/mount_overlays.go
```diff
@@ -35,58 +24,19 @@ func NewMountOverlayTask() phase.Task {
...
-	for _, o := range overlays {
-		if err = overlay(o); err != nil {
-			return err
-		}
-	}
-
-	return nil
-}
-
-func (task *MountOverlay) container(platform platform.Platform, data *userdata.UserData) (err error) {
```
andrewrynhard: This is split into its' own task since its not overlay.

internal/pkg/mount/mount.go
```diff
@@ -18,7 +21,27 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// Point represents a linux mount point.
+// RetryFunc defines the requirements for retrying a mount point operation.
+type RetryFunc func(*Point) error
```
andrewrynhard: Adding overlay functionality triggered the `gocyclo` linter, and, in this case, I agreed with it. This function type was introduced to make retries more generic in order to make things more DRY, and thereby reducing the complexity.

internal/app/machined/internal/phase/rootfs/mount_shared.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/talos-systems/talos/internal/app/machined/internal/runtime"
+	"github.com/talos-systems/talos/pkg/userdata"
+	"golang.org/x/sys/unix"
+)
+
+// MountShared represents the MountShared task.
+type MountShared struct{}
+
+// NewMountSharedTask initializes and returns an MountShared task.
+func NewMountSharedTask() phase.Task {
```
andrewrynhard: This was worthy of its' own task.

bradbeam:
layout looks clean and i like the organization. makes sense

PR #1065, opened 2019-08-26
feat: add ability to pass data on event bus

andrewrynhard:
We need to support eventing with associated data. This moves the event
bus to an observer design pattern that allows observers to register for
specific events, and to receive the associated data.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -14,9 +14,15 @@ import (
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/event"
 	"github.com/talos-systems/talos/internal/app/machined/internal/sequencer"
+	"github.com/talos-systems/talos/internal/app/machined/proto"
 	"github.com/talos-systems/talos/pkg/constants"
 	"github.com/talos-systems/talos/pkg/startup"
 )
+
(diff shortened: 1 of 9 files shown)


internal/app/machined/main.go
```diff
@@ -75,9 +81,10 @@ func main() {
 	defer recovery()
 
 	// Subscribe to events.
-	events := make(chan event.Type, 5) // provide some buffer to avoid blocking the bus
-	event.Bus().Subscribe(events)
-	defer event.Bus().Unsubscribe(events)
+	init := EventBusObserver{&event.Embeddable{}}
+	defer close(init.Channel())
```
smira: this might be an issue: as events are published asynchronously, there's a chance that event might be published on a channel after `Unregister()` is called, and writing to closed channel is a panic. so probably better to avoid closing the channel

smira: I believe closing the channel has no semantic difference for the code, so it should be safe to skip that

andrewrynhard: In this yeah. Since it’s PID1 and if this returns then we are likely shutting down or rebooting. We can remove.

PR #1069, opened 2019-08-27
chore: Increase timers for healthchecks

bradbeam:
We've seen some instances where the initial delay is not long enough (containerd)
as well as a period of every second increases the log size for services like
proxyd which log incoming connections.

### DIFF
--- a/internal/app/machined/pkg/system/health/settings.go
+++ b/internal/app/machined/pkg/system/health/settings.go
@@ -17,7 +17,7 @@ type Settings struct {
 
 // DefaultSettings provides some default health check settings
 var DefaultSettings = Settings{
-	InitialDelay: 200 * time.Millisecond,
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/health/settings.go
```diff
@@ -17,7 +17,7 @@ type Settings struct {
 
 // DefaultSettings provides some default health check settings
 var DefaultSettings = Settings{
-	InitialDelay: 200 * time.Millisecond,
-	Period:       time.Second,
+	InitialDelay: time.Second,
```
smira: just fyi, this also slows down boot sequence potentially, as some services wait for others to be up & healthy. not saying this change is bad, but the longer `initialDelay` is, the longer it takes for the service to become healthy

PR #1071, opened 2019-08-27
feat(networkd): Add link stats

bradbeam:
Display link level stats like received/transmitted packets


andrewrynhard:
@bradbeam I actually think this would be good to model after `osctl services`. For example, `osctl services` lists all services, but if you pass in a service name (`osctl service osd`) you will get more detailed information on that particular service. So in this case we can have `osctl interfaces` to get a list and then `osctl interface eth0` which will print out all the details with metrics.

andrewrynhard:
The codebase has changed significantly since this so gonna close. We can reopen if we need to.

PR #1072, opened 2019-08-27
chore: rename v1 node configs to v1alpha1

rsmitty:
This PR moves to using v1alpha1 as the inital node config version, so
we can graduate these configs a little more cleanly later on.


andrewrynhard:
@rsmitty Can we get this in today?

PR #1082, opened 2019-08-28
feat: reset machine by performing an upgrade

andrewrynhard:
With the way that upgrades work now, we can utilize them to perform a
reset by executing and upgrade using the image specified in the node
config.


andrewrynhard:
thinking that a reset should essentially wipe the disk and reboot to allow for the installer specified in the machine config to be used.

andrewrynhard:
Configurable grace period would be useful and help get around k8s failures.

andrewrynhard:
I think it will be better if we have a dedicate set of phases/tasks specifically around resetting a node.

PR #1089, opened 2019-08-28
fix: leave etcd when upgrading control plane node

andrewrynhard:
We need to remove the current node from etcd when upgrading.

### DIFF
--- a/internal/app/machined/internal/phase/upgrade/leave_etcd.go
+++ b/internal/app/machined/internal/phase/upgrade/leave_etcd.go
@@ -7,17 +7,43 @@ package upgrade
 import (
 	"context"
 	"log"
+	"os"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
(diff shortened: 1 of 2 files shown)


internal/app/machined/internal/phase/upgrade/leave_etcd.go
```diff
@@ -7,17 +7,44 @@ package upgrade
...
 	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/userdata"
 
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/pkg/transport"
 )
 
-// LeaveEtcd removes a member of etcd.
-func LeaveEtcd(hostname string) (err error) {
+// LeaveEtcd represents the task for stop all containerd tasks in the
```
smira: looks like comment doesn't relate to the task?

andrewrynhard: Whoops. Thanks.

PR #1096, opened 2019-08-30
fix(init): Reap child processes

bradbeam:
Should take care of our issue with Zombies

### DIFF
--- a/hack/containerd.toml
+++ b/hack/containerd.toml
@@ -1,0 +1,77 @@
+root = "/var/lib/containerd"
+state = "/run/containerd"
+oom_score = 0
+subreaper = true
+
+[grpc]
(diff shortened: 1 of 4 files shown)


hack/containerd.toml
```diff
@@ -0,0 +1,77 @@
+root = "/var/lib/containerd"
+state = "/run/containerd"
+oom_score = 0
+subreaper = true
```
andrewrynhard: Think we should just have this?

andrewrynhard: I take it that flags take precedence over the config?

bradbeam: i believe so; the above is generated from `containerd config`

andrewrynhard: cool

PR #1100, opened 2019-08-31
feat: attempt DHCP discovery thrice

### LINKED ISSUE #1099: networkd: DHCP discovery should occur more than once
My (Intel, for what it's worth) network cards apparently report ready before they are _actually_ ready, which results in the `cli.Request()` to fail (no packet is actually emitted across the network).

In general, since a DHCP Request is a non-connection-oriented broadcast packet, it is not safe to assume that it was received (or sent, in this case), so it is a good idea to retransmit if there is no reply after the first one.

I would propose to retrying failed requests a limited number of times.  In a normal OS, this is typically forked into the background to retry a few times (or forever).

The problem I see presently is the DHCP fails after 10s, and the boot process moves on to download the userdata, which _does_ retry, but fails forever because DHCP is _not_ retried after the first failure.

Ulexus:
Fixes #1099

### DIFF
--- a/internal/app/networkd/pkg/address/dhcp.go
+++ b/internal/app/networkd/pkg/address/dhcp.go
@@ -13,10 +13,17 @@ import (
 
 	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/insomniacslk/dhcp/dhcpv4/nclient4"
+	"github.com/pkg/errors"
 	"github.com/talos-systems/talos/internal/pkg/kernel"
 	"github.com/talos-systems/talos/pkg/constants"
 	"golang.org/x/sys/unix"
 )
+
+// MaxRequests is the maximum number of DHCPRequests to make before giving up
+const MaxRequests = 3
+
+// MinRequestLoopTime is the minimum amount of time to wait between failed DHCPRequest attempts
+const MinRequestLoopTime = 3 * time.Second
 
 // DHCP implements the Addressing interface
 type DHCP struct {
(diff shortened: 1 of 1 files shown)


internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -158,12 +162,14 @@ func (d *DHCP) discover() (*dhcpv4.DHCPv4, error) {
...
-		// TODO: Make this a well defined error so we can make it not fatal
-		log.Println("failed dhcp request")
-		return nil, err
+	// TODO: this may be better handled by retrying indefinitely in the background
+	for i := 0; i < MaxRequests; i++ {
+		_, ack, err = cli.Request(context.Background(), mods...)
+		if err != nil {
+			// TODO: Make this a well defined error so we can make it not fatal
+			log.Println("failed dhcp request:", err)
+		}
```
andrewrynhard: Maybe we should do something like

`time.Sleep(i * time.Second)`

?

Ulexus: The dhcpv4 client internally waits a default of 10s for a timeout... but it's true that in the event of an immediate error, we shouldn't have a tight spin cycle.  I'll put an at-least timer in there.

andrewrynhard: Ahh wasn’t aware that the client already handled it to some degree. SGTM

Ulexus: Fixed

andrewrynhard:
This working for you? We can merge if so.

Ulexus:
Testing now

Ulexus:
Yes, this appears to fix my observed problem.

bradbeam:
Hold please. The library we use does three retries and is configurable with functional args.

Ulexus:
Does three retries already, you are saying?

Ulexus:
You appear to be correct:  the defaults are 5s timeout and 3 retries.  However, the process appears to timeout after on 10s, not 15, for me.  Perhaps it is triggering far earlier than expected.

Ulexus:
In any case, this is the wrong solution, since the library includes the required tunables internally.

PR #1103, opened 2019-09-02
chore: enable unit-tests-race

smira:
This is experiment to see how stable they are.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -308,6 +308,23 @@ steps:
   depends_on:
   - rootfs
 
+- name: unit-tests-race
+  image: autonomy/build-container:latest
+  commands:
+  - make unit-tests-race
+  environment:
+    BINDIR: /usr/local/bin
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -20,7 +20,7 @@ steps:
   - make machined
   environment:
     BINDIR: /usr/local/bin
-    BUILDKIT_HOST: ${BUILDKIT_HOST=tcp://buildkitd.ci.svc:1234}
+    BUILDKIT_HOST: "${BUILDKIT_HOST=tcp://buildkitd.ci.svc:1234}"
```
smira: I wonder if my version of `drone` CLI is different?

.drone.yml
```diff
@@ -2038,10 +2106,6 @@ volumes:
 node:
   node-role.kubernetes.io/ci: ""
 
-trigger:
```
smira: not sure about this change??

.drone.yml
smira: got it, wasn't using latest drone CLI

PR #1108, opened 2019-09-04
chore: align time command with output standards

### LINKED ISSUE #1091: osctl time is not in table format
We are standardizing on table format for now, `osctl time` should align with that.

andrewrynhard:
This changes the output to a table writer with all caps for headers.

Fixes #1091.

### DIFF
--- a/cmd/osctl/cmd/time.go
+++ b/cmd/osctl/cmd/time.go
@@ -6,6 +6,8 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"text/tabwriter"
 	"time"
 
 	"github.com/golang/protobuf/ptypes"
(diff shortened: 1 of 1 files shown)


cmd/osctl/cmd/time.go
```diff
@@ -52,9 +54,10 @@ var timeCmd = &cobra.Command{
 				helpers.Fatalf("error parsing remote time: %s", err)
 			}
 
-			fmt.Printf("NTP Server: %s\n", output.Server)
-			fmt.Printf("Local time: %s\n", localtime)
-			fmt.Printf("Remote time: %s\n", remotetime)
+			w := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+			fmt.Fprintln(w, "SERVER\tLOCAL-TIME\tREMOTE-TIME")
```
smira: should it be `NTP-SERVER` to make it less confusing with whatever 'server' might be?

andrewrynhard: Yes, good point

PR #1110, opened 2019-09-04
chore(machined): implement process reaper for PID 1 machined process

smira:
In UNIX, any zombies without parent process get re-parented to process
with PID 1 (usually running init), and PID 1 process should take care of
them (usually simply clean them up). Cleaning up zombies is important,
as they still take kerner resources, and having enormous amount of
zombie processes signifcantly degrades system performance.

For Talos, PID 1 process is machined, and machined itself forks to run
other processes in process runner and `pkg/cmd` one-time commands. Naive
solution of running `wait()` loop doesn't work as it might race with
`Process.Wait()` and clean up zombie which wasn't re-parented which
leads to process execution false failure.

After considering other solutions, we decided to go with the simple
approach: machined runs global zombie process reaper which publishes
information about reaped zombies. Any call to `Process.Wait()` (or
`Command.Wait()` which calls it) should be replaced with listening to
reaper's channel for notifications to catch info about the process which
was created in this call.

There are several changes in this PR:

1. Reaper implementation itself, started from machined.

2. Process runner and `pkg/cmd` can either use regular `Command.Wait()`
or use reaper notifications depending on reaper status (running/not
running). This allows using this code outside of machined.

3. Small bug fixes with process log which was affecting the tests.

### DIFF
--- a/internal/app/machined/pkg/system/log/log.go
+++ b/internal/app/machined/pkg/system/log/log.go
@@ -27,14 +27,15 @@ type Log struct {
 
 // New initializes and registers a log for a service.
 func New(name, rootPath string) (*Log, error) {
+	logpath := FormatLogPath(name, rootPath)
+
 	mu.Lock()
-	if l, ok := instance[name]; ok {
+	if l, ok := instance[logpath]; ok {
 		mu.Unlock()
 		return l, nil
 	}
 	mu.Unlock()
 
-	logpath := FormatLogPath(name, rootPath)
 	w, err := os.Create(logpath)
 	if err != nil {
 		return nil, fmt.Errorf("create log file: %s", err.Error())
@@ -47,7 +48,7 @@ func New(name, rootPath string) (*Log, error) {
 	}
 
 	mu.Lock()
-	instance[name] = l
+	instance[l.Path] = l
 	mu.Unlock()
 
 	return l, nil
@@ -61,7 +62,7 @@ func (l *Log) Write(p []byte) (n int, err error) {
 // Close implements io.WriteCloser.
 func (l *Log) Close() error {
 	mu.Lock()
-	delete(instance, l.Name)
+	delete(instance, l.Path)
 	mu.Unlock()
 
 	return l.source.Close()
--- a/internal/app/machined/pkg/system/runner/process/process.go
+++ b/internal/app/machined/pkg/system/runner/process/process.go
@@ -18,6 +18,7 @@ import (
 	processlogger "github.com/talos-systems/talos/internal/app/machined/pkg/system/log"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner"
 	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/proc/reaper"
 	"github.com/talos-systems/talos/pkg/userdata"
 )
 
@@ -108,6 +109,12 @@ func (p *processRunner) run(eventSink events.Recorder) error {
 		return errors.Wrap(err, "error building command")
 	}
 
+	notifyCh := make(chan reaper.ProcessInfo, 8)
+	usingReaper := reaper.Notify(notifyCh)
+	if usingReaper {
+		defer reaper.Stop(notifyCh)
+	}
+
 	if err = cmd.Start(); err != nil {
 		return errors.Wrap(err, "error starting process")
 	}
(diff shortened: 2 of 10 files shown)


internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -108,6 +109,12 @@ func (p *processRunner) run(eventSink events.Recorder) error {
 		return errors.Wrap(err, "error building command")
 	}
 
+	notifyCh := make(chan reaper.ProcessInfo, 8)
+	usingReaper := reaper.Notify(notifyCh)
```
andrewrynhard: It looks like `reaper.Notify` returns true only if it's running? I guess I don't understand  when/why we would want it to return `false`? The way I read this now it makes it seem like this is optional and up to the caller to decide if they want to use the reaper. I'm probably not understanding something fully, but this confuses me a little.

pkg/cmd/cmd.go
```diff
@@ -24,11 +26,19 @@ func Run(name string, args ...string) error {
...
+	if usingReaper {
+		defer reaper.Stop(notifyCh)
+	}
+
 	if err = cmd.Start(); err != nil {
 		return errors.Errorf("%s: %s", err, stderr.String())
 	}
 
-	if err = cmd.Wait(); err != nil {
+	err = reaper.WaitWrapper(usingReaper, notifyCh, cmd)
```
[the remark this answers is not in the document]
smira: yep, thanks, I missed that while code was going through too many changes

pkg/proc/reaper/hunter.go
```diff
@@ -0,0 +1,146 @@
...
+	for {
+		var (
+			wstatus syscall.WaitStatus
+			pid     int
+			err     error
+		)
+
+		for {
+			// retry EINTR on wait4()
+			pid, err = syscall.Wait4(-1, &wstatus, syscall.WNOHANG, nil)
```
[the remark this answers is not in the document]
smira: `-1` should be just "any PID".

`wait4` is the only one available in Go directly, and it's equivalent to `waitpid` (http://man7.org/linux/man-pages/man2/wait4.2.html)

`waitpid` (https://linux.die.net/man/2/waitpid) says that `pid == -1` is "any child process"

internal/app/machined/pkg/system/runner/process/process.go
smira: yep, exactly, it's for the caller to know whether reaper is running or not.

so the thing is that we need to subscribe for notifications before starting a process (as a process might exit before we get control back), so we don't know the PID of the child we're going to create.

if running with reaper, we just wait on the channel and filter for notification, without reaper we do regular `cmd.Wait()` (which does `waitpid(pid, ...)`).

I wanted the code in `pkg/cmd` and process runner to work with and without reaper, as it might be used outside of `machined` (PID 1 process), so I didn't want to have a hard dependency on reaper running, but at the same time flows with reaper and without reaper should be different

smira: so if we could wait for notifications once `cmd.Start()` returns, it could have been as easy as `reaper.Wait(cmd.Process.Pid)`, but as we don't know the PID before fork(), this has to be split into subscribe for notifications first, fork(), wait, unsubscribe flow, so code is a bit ugly

pkg/proc/reaper/hunter.go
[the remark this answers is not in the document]
andrewrynhard: Awesome thanks.

internal/app/machined/pkg/system/runner/process/process.go
andrewrynhard: Ahh ok. Having it work for things outside out machined explains it.

internal/app/machined/pkg/system/log/log.go
```diff
@@ -27,14 +27,15 @@ type Log struct {
 
 // New initializes and registers a log for a service.
 func New(name, rootPath string) (*Log, error) {
+	logpath := FormatLogPath(name, rootPath)
+
 	mu.Lock()
-	if l, ok := instance[name]; ok {
+	if l, ok := instance[logpath]; ok {
```
smira: the problem here was that log objects for the same service `name` but different `rootPath` were cached as the same object leading to surprising results in the tests when `logPath` is just `tmpDir`

andrewrynhard:
We good to merge this?

smira:
> We good to merge this?

I am good!

PR #1111, opened 2019-09-05
fix: run only essential services in container mode

andrewrynhard:
This adds NTP and Udev to the list of services designated only for
non-container modes. We should be able to rely on the host running
Docker for accurate time and Udev information.

### DIFF
--- a/internal/app/machined/internal/phase/services/start_services.go
+++ b/internal/app/machined/internal/phase/services/start_services.go
@@ -40,21 +40,19 @@ func (task *StartServices) loadSystemServices(args *phase.RuntimeArgs) {
 	svcs.Load(
 		&services.MachinedAPI{},
 		&services.Containerd{},
-		&services.Networkd{},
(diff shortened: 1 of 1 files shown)


internal/app/machined/internal/phase/services/start_services.go
```diff
@@ -44,15 +44,15 @@ func (task *StartServices) loadSystemServices(data *userdata.UserData, mode runt
 		&services.MachinedAPI{},
 		&services.Containerd{},
 		&services.Networkd{},
-		&services.Udevd{},
 		&services.OSD{},
-		&services.NTPd{},
 	)
 
 	if mode != runtime.Container {
 		// udevd-trigger is causing stalls/unresponsive stuff when running in local mode
```
smira: should we update comment with the reasons behind disabling these?

andrewrynhard: +1 We should.

PR #1112, opened 2019-09-05
feat: configure interfaces concurrently

andrewrynhard:
This uses a wait group to configure interfaces concurrently.

### DIFF
--- a/internal/app/networkd/pkg/networkd/networkd.go
+++ b/internal/app/networkd/pkg/networkd/networkd.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -80,33 +81,41 @@ func (n *Networkd) Configure(ifaces ...*nic.NetworkInterface) error {
...
-		for _, method := range iface.AddressMethod {
-			log.Printf("configuring %s addressing for %s\n", method.Name(), iface.Name)
+	wg.Add(len(ifaces))
 
-			if err = n.configureInterface(method); err != nil {
-				// Treat as non fatal error when failing to configure an interface
-				log.Println(err)
-				continue
+	for _, iface := range ifaces {
+		go func(i *nic.NetworkInterface) {
```
smira: no `defer wg.Done()` ?

I bet this should hang forever...

andrewrynhard: whoops

andrewrynhard: Fixed

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -80,33 +81,42 @@ func (n *Networkd) Configure(ifaces ...*nic.NetworkInterface) error {
...
+					log.Println(err)
+					return
+				}
+
+				if !method.Valid() {
+					return
+				}
+
+				// Aggregate a list of DNS servers/resolvers
+				resolvers = append(resolvers, method.Resolvers()...)
```
smira: this should have some locking as it's data race

andrewrynhard: good catch

smira: I'm not sure if that scope is right? should it be local variable? 

`resolvers` is a local variable in this function

andrewrynhard: 🤦‍♂ I'm tired today. Long night with the baby.

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -78,35 +79,47 @@ func (n *Networkd) Configure(ifaces ...*nic.NetworkInterface) error {
...
+					log.Println(err)
+					return
+				}
+
+				if !method.Valid() {
+					return
+				}
+
+				// Aggregate a list of DNS servers/resolvers
+				mu.Lock()
```
bradbeam: unsure if it's better to use a lock here or just kick off another loop afterwards
```
for _, iface := range ifaces {
  go func(i *nic.NetworkInterface) { ... }
}
wg.Wait()
var resolvers []string
for _, iface := range ifaces {
  resolvers = append(resolvers, method.Resolvers()...)
}

andrewrynhard: One for loop better than two maybe? IDK. We could probably go either way but int the end one doesn't have something over the other IMO.

PR #1113, opened 2019-09-05
refactor(ntpd): Improvements to the robustness of ntp

bradbeam:
- Use the Validate method to ensure we get an appropriate time back
- Hard set the clock initially, adjust clock by offsets afterwards
- Introduce functional opts to configure ntp client
- Add additional test coverage

### DIFF
--- a/internal/app/ntpd/pkg/ntp/ntp.go
+++ b/internal/app/ntpd/pkg/ntp/ntp.go
@@ -5,93 +5,110 @@
 package ntp
 
 import (
-	"errors"
+	"fmt"
 	"log"
 	"math/rand"
 	"syscall"
 	"time"
 
 	"github.com/beevik/ntp"
-)
-
-// https://access.redhat.com/solutions/39194
-// Using the above as reference for setting min/max
-const (
-	MaxPoll = 1000
-	MinPoll = 20
+	"github.com/hashicorp/go-multierror"
 )
 
 // NTP contains a server address
-// and the most recent response from a query
 type NTP struct {
-	Server   string
-	Response *ntp.Response
+	Server  string
+	MinPoll time.Duration
+	MaxPoll time.Duration
+	Retry   int
 }
 
 // NewNTPClient instantiates a new ntp client for the
 // specified server
-func NewNTPClient(server string) *NTP {
-	return &NTP{Server: server}
+func NewNTPClient(opts ...Option) (*NTP, error) {
+	ntp := defaultOptions()
+
+	var result *multierror.Error
+	for _, setter := range opts {
+		result = multierror.Append(setter(ntp))
+	}
+
+	return ntp, result.ErrorOrNil()
 }
 
 // Daemon runs the control loop for query and set time
 // We dont ever want the daemon to stop, so we only log
 // errors
 func (n *NTP) Daemon() (err error) {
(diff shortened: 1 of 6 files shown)


internal/app/ntpd/pkg/ntp/ntp.go
```diff
@@ -5,93 +5,110 @@
...
-	if resp == nil {
-		return errors.New("not a valid ntp response")
+	for i := 0; i < n.Retry; i++ {
+		resp, err := ntp.Query(n.Server)
+		if err != nil {
+			time.Sleep(time.Duration(i) * n.MinPoll)
+			continue
+		}
+
+		if err := resp.Validate(); err != nil {
```
bradbeam: This func handles the juicy details,

https://github.com/beevik/ntp/blob/master/ntp.go#L248-L252

internal/app/ntpd/pkg/ntp/ntp.go
```diff
@@ -5,93 +5,110 @@
...
+	var randSleep time.Duration
 	for {
-		<-ticker.C
-		// Set some variance with how frequently we poll ntp servers
-		resp, err = n.Query()
-		if err != nil {
+		// Set some variance with how frequently we poll ntp servers.
+		// This is based on rand(MaxPoll) + MinPoll so we wait at least
+		// MinPoll.
+		randSleep = time.Duration(rand.Intn(int(n.MaxPoll.Seconds()))) * time.Second
```
smira: I'm a big fan of exponential distribution to manage such intervals (https://godoc.org/math/rand#ExpFloat64). 

This could be different from MinPoll/MaxPoll setup, but should provide more natural distribution of poll intervals across cluster. 

Not really blocking this PR, I think what we have here with uniform distribution looks good.

internal/app/ntpd/pkg/ntp/ntp.go
```diff
@@ -5,93 +5,110 @@
...
-	resp, err = n.Query()
-	if err != nil {
+	if resp, err = n.Query(); err != nil {
 		log.Printf("error querying %s for time, %s", n.Server, err)
 		return err
 	}
-	n.Response = resp
 
-	if err = n.SetTime(resp); err != nil {
+	if err = setTime(resp.Time); err != nil {
```
smira: should we always use `ClockOffset` to set the time? (not sure here, but I don't see a difference between first and subsequent attempts, as they all just call settimeofday). 

If we used another advanced API to slowly adjust for clock offset, it would make more sense to have different implementation for the first iteration and subsequent iterations. But as in the end it boils down to `setTime` in both cases, the different is only how we calculate timestamp?

bradbeam: there's [syscall.Adjtimex](https://golang.org/pkg/syscall/#Adjtimex) // http://man7.org/linux/man-pages/man2/adjtimex.2.html which i'd probably need to read up more on.

internal/app/ntpd/pkg/ntp/ntp.go
andrewrynhard: Its probably worth us building out a library for generic backoff strategies. We use them in a number of places. Obviously not for this PR, but while we are on the topic of polling I figured I'd throw in my 2 cents.

andrewrynhard:
We good to merge this?

PR #1116, opened 2019-09-06
fix: prevent EBUSY when unmounting system disk

andrewrynhard:
Reading /proc/mounts while simultaneously unmounting mountpoints
prevents unmounting all submounts under /var. This is due to the fact
that /proc/mounts will change as we perform unmounts, and that causes a
read of the file to become inaccurate. We now read /proc/mounts into
memory to get a snapshot of all submounts under /var, and then we
proceed with unmounting them.

This also adds some additional logging that I found to be useful while
debugging this. It also adds logic to skip of DaemonSet managed pods.

### DIFF
--- a/internal/app/machined/internal/phase/rootfs/unmount_pod_mounts.go
+++ b/internal/app/machined/internal/phase/rootfs/unmount_pod_mounts.go
@@ -6,7 +6,9 @@ package rootfs
 
 import (
 	"bufio"
-	"os"
+	"bytes"
+	"io/ioutil"
+	"log"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -104,6 +104,12 @@ func (h *Helper) Drain(node string) error {
 	for _, pod := range pods.Items {
 		go func(p corev1.Pod) {
 			defer wg.Done()
+			for _, ref := range p.ObjectMeta.OwnerReferences {
+				if ref.Kind == "DaemonSet" {
```
andrewrynhard: DaemonSet managed pods are impossible to get rid of as long as the Kubelet is running. Skip them to speed up upgrades and prevent any thrashing that could potentially fail the upgrade.

internal/app/machined/internal/phase/rootfs/unmount_pod_mounts.go
```diff
@@ -37,14 +39,14 @@ func (task *UnmountPodMounts) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
 }
 
 func (task *UnmountPodMounts) standard(platform platform.Platform, data *userdata.UserData) (err error) {
-	file, err := os.Open("/proc/mounts")
-	if err != nil {
+	var b []byte
```
andrewrynhard: Read `/proc/mount` into memory since it will change as we unmount things.

PR #1122, opened 2019-09-07
test(ci): Add aws for e2e and conformance targets

bradbeam:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -813,6 +813,8 @@ steps:
   commands:
   - make capi
   environment:
+    AWS_SVC_ACCT:
+      from_secret: aws_svc_acct
     AZURE_SVC_ACCT:
       from_secret: azure_svc_acct
     BINDIR: /usr/local/bin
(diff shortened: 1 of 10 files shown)


.drone.yml
```diff
@@ -813,6 +813,10 @@ steps:
   commands:
   - make capi
   environment:
+    AWS_ACCESS_KEY_ID:
```
bradbeam: Most likely can drop these in favor of `AWS_SVC_ACCT`; still needs to be added

.drone.yml
```diff
@@ -865,11 +869,32 @@ steps:
   depends_on:
   - installer
 
+- name: image-snap-aws
```
bradbeam: adding a separate step for now to not clobber existing aws workflow.

andrewrynhard: Have you gotten the import to finish successfully and have you been able to boot a node? If we can get this working, let's drop the packet stuff entirely and replace it with this. This will align AWS with how we are doing GCP and Azure.

hack/test/aws-setup.sh
```diff
@@ -0,0 +1,43 @@
...
+REGION="us-east-1"
+BUCKET="talos-e2e"
+TMP=/tmp/e2e/aws
+
+## Setup svc account
+mkdir -p ${TMP}
+echo ${AWS_SVC_ACCT} | base64 -d > ${TMP}/svc-acct.ini
+
+# Ensure AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY env vars are available
+export AWS_ACCESS_KEY_ID=$(awk '/aws_access_key_id/ { print $NF }' ${TMP}/svc-acct.ini}
```
andrewrynhard: We should either go with credentials file or use the env vars as part of the step. the AWS CLI will look for `~/.aws/credentials` so we could either drop everything we need in there, or we can skip that entirely and just use env vars.

bradbeam:
closes #260

PR #1125, opened 2019-09-07
feat: allow network interface to be ignored

### LINKED ISSUE #1124: Add an Ignore option for a network interface
We should add the ability to ignore an interface.  Presently, the default operation for a network interface is to run a DHCP client on it.  This is good in most cases, but for servers with multiple NICs, this can result in unpleasant behaviour, either by having interfaces with IPs that should not or by taking a long time for the interfaces to time out their DHCP.

Ulexus:
Added a property to userdata to allow a network interface to be ignored,
such that Talos will perform no operations on it (including DHCP).

Also added kernel commandline parameter (talos.network.interface.ignore)
to specify a network interface should be ignored.

Fixes #1124

### DIFF
--- a/internal/pkg/kernel/kernel.go
+++ b/internal/pkg/kernel/kernel.go
@@ -82,6 +82,10 @@ func (v *Parameter) First() *string {
 
 // Get attempts to get a string from a value's internal representation.
 func (v *Parameter) Get(idx int) *string {
+	if v == nil {
+		return nil
+	}
+
 	if len(v.values) > idx {
 		return &v.values[idx]
 	}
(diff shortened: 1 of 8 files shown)


internal/pkg/kernel/kernel.go
```diff
@@ -91,6 +95,10 @@ func (v *Parameter) Get(idx int) *string {
 
 // Contains returns a boolean indicating the existence of a value.
 func (v *Parameter) Contains(s string) (ok bool) {
+	if v == nil {
+		return
```
andrewrynhard: nit: although golang allows it, we have been trying to avoid naked returns.

pkg/constants/constants.go
```diff
@@ -29,6 +29,9 @@ const (
 	// initial interface used to bootstrap the node
 	KernelParamDefaultInterface = "talos.interface"
 
+	// KernelParamNetworkInterfaceIgnore is the kernel parameter for specifying network interfaces which should be ignored by talos
+	KernelParamNetworkInterfaceIgnore = `talos.network.interface.ignore`
```
andrewrynhard: nit: to align with the other constants, can we use `"`?

PR #1129, opened 2019-09-08
fix: prepend custom options for kernel commandline

### LINKED ISSUE #1128: Cannot override serial console for bootloader install
Because the serial console is set in the defaults and because the kernel uses the _first_ serial console in the commandline options as the serial console, there is no way to override the serial console for systems which do not have their serial console on ttyS0.

Ulexus:
Added a decomposition option to the kernel.NewDefaultCmdline() so that
the Defaults can be added _after_ constructing a custom commandline.
This is then implemented for `osctl install`.

Fixes #1128

### DIFF
--- a/cmd/osctl/cmd/install_linux.go
+++ b/cmd/osctl/cmd/install_linux.go
@@ -42,13 +42,14 @@ var installCmd = &cobra.Command{
 			},
 		}
 
-		cmdline := kernel.NewDefaultCmdline()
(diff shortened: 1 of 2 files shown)


Ulexus:
Sorry, I wasn't comfortable making the console order change after reading some of those referenced docs.  Too much change of unintended consequences.  It's not apropos to the issue anyway.

PR #1132, opened 2019-09-08
feat: Allow spec of canonical controlplane addr

### LINKED ISSUE #1131: Allow specification of canonical ControlPlaneEndpoint
Presently, the kubeadm APIServer endpoint (as used in address binding and JoinConfiguration) and the canonical ControlPlaneEndpoint are the same thing.  However, in most production systems, we will want the canonical endpoint to be the IP address or DNS name of a load balancer (or just set of DNS A or AAAA records).

While we could hack this with the v0 userdata, this is no longer possible with the more-structured v1 userdata.

Ulexus:
Broke the binding between the discrete IP addresses of the control plane
elements and the ControlPlaneEndpoint.  This allows the specification of
a canonical controlplane address which may optionally be a DNS name.

Fixes #1131

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -125,7 +125,7 @@ var configAddCmd = &cobra.Command{
 
 // configGenerateCmd represents the config generate command.
 var configGenerateCmd = &cobra.Command{
-	Use:   "generate",
+	Use:   "generate <clusterName> <master-1-IP,master-2-IP,master-3-IP>",
 	Short: "Generate a set of configuration files",
 	Long:  ``,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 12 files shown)


cmd/osctl/cmd/config.go
```diff
@@ -282,6 +283,7 @@ func init() {
 	configAddCmd.Flags().StringVar(&crt, "crt", "", "the path to the certificate")
 	configAddCmd.Flags().StringVar(&key, "key", "", "the path to the key")
 	configGenerateCmd.Flags().StringSliceVar(&additionalSANs, "additionalSANs", []string{}, "additional Subject-Alt-Names for the APIServer certificate")
+	configGenerateCmd.Flags().StringVar(&canonicalControlplaneEndpoint, "controlplaneEndpoint", "", "the canonical controlplane endpoint (IP or DNS name) and optional port (defaults to 6443)")
```
andrewrynhard: nit: in other commands that have multi-word args we use something like `controlplane-endpoint`.

andrewrynhard:
@Ulexus I marked as ready for review since I can't merge in draft state. Sorry if this wasn't ready for that, but I assumed it was.

Issue #1090, opened 2019-08-28
Azure hostnames are set to default


andrewrynhard:
Azure nodes are getting the default `talos-$IP` hostname, and I don't recall this being the case before the new networkd implementation.

PR #1137, opened 2019-09-09
fix(machined): Fix hostname value when retrieving from cloud providers

bradbeam:
There was an issue where the hostname was getting set too early in the boot. This caused
the hostnam retrieved from platform.Hostname() to be ignored.


bradbeam:
fixes #1090

PR #1139, opened 2019-09-09
fix: make disk probe less prone to error

andrewrynhard:


andrewrynhard:
Still a WIP, I can't seem to reproduce locally so throwing this up now to try and catch it in drone.

PR #807, opened 2019-07-01
TLS renew

Ulexus:
Implement TLS certificate renewal for OSD.

This is definitely a WIP and needs architectural review.  As discussed, this is a temporary option until we can move to an ephemeral-by-default system.

PR #1147, opened 2019-09-10
feat: set expiry of certificates to 24 hours

andrewrynhard:
This defaults certificates to a 24 hour TTL.

### DIFF
--- a/pkg/userdata/userdata.go
+++ b/pkg/userdata/userdata.go
@@ -12,7 +12,6 @@ import (
 	stdlibnet "net"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/talos-systems/talos/pkg/crypto/x509"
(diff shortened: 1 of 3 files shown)


andrewrynhard:
> I'm all for the change, but not sure I understand why so many places got updated, as I expect it to be single one which generates cert from userdata, but I wasn't able to trace full flow in the code

Agree. I was surprised it was in so many places. We should simplify and make it in one place.

Ulexus:
Never mind; we do, and it was my own PR:  #807  :eyeroll:

andrewrynhard:
@smira I think I understand it now. The functional options for a new `CSR` and a new certificate are mixed when they should be two different sets of options. `NotAfter` is handled in `NewCertificateFromCSR`.

pkg/userdata/userdata.go
```diff
@@ -126,7 +125,6 @@ func (data *UserData) NewIdentityCSR() (csr *x509.CertificateSigningRequest, err
 	names := []string{hostname}
 	opts = append(opts, x509.DNSNames(names))
 	opts = append(opts, x509.IPAddresses(ips))
-	opts = append(opts, x509.NotAfter(time.Now().Add(time.Duration(8760)*time.Hour)))
```
andrewrynhard: @smira this was actually never used.

PR #1149, opened 2019-09-10
fix: move to per-platform console setup

rsmitty:
This PR will make sure that each platform gets the console settings it
needs by setting them as extra flags in the makefile. This should ensure
that we have console logs flowing properly for each cloud.


rsmitty:
Tested and working:
- [ ] AWS
- [x] Azure
- [x] GCP

PR #1153, opened 2019-09-10
chore: format code with gofumpt

andrewrynhard:
The gofumpt linter is a stricter drop-in replacement for gofmt. The
rules are ones that I strongly agree with and I think it would be better
if we added this linter instead of nit picking every PR.


andrewrynhard:
Yeah some are questionable, but like you said, better we have a standard than not.

PR #1157, opened 2019-09-11
fix: use /var/log for default log path

andrewrynhard:
This moves the default log path to /var/log. An expection is made for
machined-api and system-containerd since they must have zero
dependencies on the ephemeral disk. In the case of machined-api, we
cannot stop the service since it is required to perform an upgrade. As
for system-containerd, it starts before any ephemeral disk is mounted so
we will fail to start the service since /var/log is a read-only file system.

### DIFF
--- a/internal/app/machined/pkg/system/runner/runner.go
+++ b/internal/app/machined/pkg/system/runner/runner.go
@@ -61,7 +61,7 @@ func DefaultOptions() *Options {
 	return &Options{
 		Env:                     []string{},
 		Namespace:               "system",
-		LogPath:                 "/run",
+		LogPath:                 "/var/log",
 		GracefulShutdownTimeout: 10 * time.Second,
 		ContainerdAddress:       constants.ContainerdAddress,
 	}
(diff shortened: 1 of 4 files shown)


pkg/constants/constants.go
```diff
@@ -176,7 +176,7 @@ const (
 	RootfsAsset = "rootfs.sqsh"
 
 	// NodeCertFile is the filename where the current Talos Node Certificate may be found
-	NodeCertFile = "/var/talos-node.crt"
+	NodeCertFile = "/run/talos-node.crt"
```
andrewrynhard: This is one less file in `/var`, and we wanted to make node certs ephemeral so figured I could add it here. I will submit another PR so that we have it in the changelog, feels important enough.

andrewrynhard:
> I think we need also to fix API to build the log path by inspecting service's logPath instead of hardcoding '/var/log'

Good catch. As is this would still leave machined-api and system-containers logs broken.

andrewrynhard:
Hmm. Looking at the code, it doesn't seem like we have a good abstraction for getting the log path regardless of the runner type. Am I missing something?

smira:
> Hmm. Looking at the code, it doesn't seem like we have a good abstraction for getting the log path regardless of the runner type. Am I missing something?

yep, probably deserves a separate PR. would be great to push that as service property

smira:
anyway, your PR makes some logs working while today every log is broken :-D

andrewrynhard:
> anyway, your PR makes some logs working while today every log is broken :-D

Awesome, I will start looking at making the log path a service property.

PR #1158, opened 2019-09-11
feat: move node certificate to tmpfs

andrewrynhard:
This ensures that node certificates are ephemeral by storing them in a
tmpfs.


andrewrynhard:
@Ulexus pinging you so you are aware.

PR #1159, opened 2019-09-11
chore: add fmt target

andrewrynhard:
This provides a target that can be useful for developers. It will format
code according to our standards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,6 +262,10 @@ unit-tests-race: buildkitd
 		--opt build-arg:TESTPKGS=$(TESTPKGS) \
 		$(COMMON_ARGS)
 
+.PHONY: fmt
+fmt:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -262,6 +262,10 @@ unit-tests-race: buildkitd
 		--opt build-arg:TESTPKGS=$(TESTPKGS) \
 		$(COMMON_ARGS)
 
+.PHONY: fmt
+fmt:
+	@docker run --rm -it -v $(PWD):/src -w /src golang:1.12 bash -c "export GO111MODULE=on; export GOPROXY=https://proxy.golang.org; cd /tmp && go mod init tmp && go get mvdan.cc/gofumpt && gofumpt -s -w ."
```
smira: we have `$(GO_VERSION)`

PR #1160, opened 2019-09-11
chore: upgrade golancgi-lint to 1.18.0

smira:
New linter 'funlen' was disabled as too many functions break the default
limit, but might be considered for the future.

To limit peak memory usage, `GOGC=50` was added to the golangci-lint run
to make Go's garbage collector more aggressive. With this setting peak
seems to be around 8Gb.


smira:
On GOGC: https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint

smira:
On `lint` target runtime: looks like GOGC=50 doesn't have huge effect, but still around 26s vs. 38s

PR #1168, opened 2019-09-13
chore: randomize containerd namespace in tests

smira:
Looks like containerd creates shim file sockets in Linux abstract
namespace which are fixed (don't depend on containerd root directory)
and depend on container namespace and id. So if two containerd instances
on the same host run same namespace/id pair, that is going to create a
conflict on that shim filesocket.

Avoid that by randomizing namespace name. CRI tests should be fine as
namespace is fixed, but container ID is random.

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"math/rand"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd_test.go
```diff
@@ -93,7 +94,9 @@ func (suite *ContainerdSuite) SetupSuite() {
 	suite.client, err = containerd.New(suite.containerdAddress)
 	suite.Require().NoError(err)
 
-	ctx := namespaces.WithNamespace(context.Background(), containerdNamespace)
+	suite.containerdNamespace = fmt.Sprintf("talostest%d%d", rand.Int63(), time.Now().Unix())
```
smira: containerd has weird validation rules for namespace name... e.g. `uuid` doesn't work

it should start with letters, parts after `-` should start with letter, etc.

PR #1178, opened 2019-09-17
fix: conditionally set log path

andrewrynhard:
This is not the best solution to this, but it stops the bleeding. We can
conditionally build the log base path based on the service logs
requested.


andrewrynhard:
Adding a log path property to a service is going to be more work than I'd like for v0.2. Let's fix it in v0.3. For now, this should work.

PR #1180, opened 2019-09-17
fix: log system services to /run/system/log

andrewrynhard:
Writing system logs to /var/log breaks upgrades. The system disk unmount
fails with EBUSY. For now we can log to /run/system/log to avoid this.

### DIFF
--- a/internal/app/machined/pkg/system/services/osd.go
+++ b/internal/app/machined/pkg/system/services/osd.go
@@ -55,7 +55,7 @@ func (o *OSD) Condition(data *userdata.UserData) conditions.Condition {
 
 // DependsOn implements the Service interface.
 func (o *OSD) DependsOn(data *userdata.UserData) []string {
-	return []string{"system-containerd"}
+	return []string{"system-containerd", "containerd"}
 }
 
 func (o *OSD) Runner(data *userdata.UserData) (runner.Runner, error) {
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/system/services/osd.go
```diff
@@ -55,7 +55,7 @@ func (o *OSD) Condition(data *userdata.UserData) conditions.Condition {
 
 // DependsOn implements the Service interface.
 func (o *OSD) DependsOn(data *userdata.UserData) []string {
-	return []string{"system-containerd"}
+	return []string{"system-containerd", "containerd"}
```
andrewrynhard: Fixes:

```
[   82.736543] [talos] service[osd](Waiting): Error running Containerd(osd), going to restart forever: failed to create task: "osd": OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"rootfs_linux.go:58: mounting \\\"/run/containerd/containerd.sock\\\" to rootfs \\\"/run/system/containerd/io.containerd.runtime.v1.linux/system/osd/rootfs\\\" at \\\"/run/containerd/containerd.sock\\\" caused \\\"stat /run/containerd/containerd.sock: no such file or directory\\\"\"": unknown
```

PR #1184, opened 2019-09-18
chore: move gRPC API to public

andrewrynhard:
In order for other projects to make use of our APIs, they must not
reside underneath the internal directory. This moves the protobuf
definitions to a top-level "api" directory and scopes them according to
their domain. This change also removes generated code from the gitignore
file so that users don't have to generate the code themseleves.


smira:
I wonder if we should also promote `client` package from `osctl` to be more of top-level thing, as it provides nice Go wrappers for the API.

andrewrynhard:
> I wonder if we should also promote `client` package from `osctl` to be more of top-level thing, as it provides nice Go wrappers for the API.

I was thinking the same. It would be nice for when we build out the UI. For example, in the `osctl df` output we calculate a number of things client side. I'd like to be able to do that using a common code base.

andrewrynhard:
@smira ok if we follow up with another PR to make the client outside of `osctl`?

smira:
> @smira ok if we follow up with another PR to make the client outside of `osctl`?

sure thing!

PR #1188, opened 2019-09-19
feat: return a data structure in version RPC

andrewrynhard:
A byte slice is not very useful. Having a struct with fields makes for a
better experience.


pkg/version/version.go
```diff
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"runtime"
 	"text/template"
+
+	machineapi "github.com/talos-systems/talos/api/machine"
```
smira: I have mixed feelings about getting protobuf-generated structs into "core" packages

This avoids duplication, but looks a bit like reverse dependency?

andrewrynhard: Fair point. I'm not sure what to do TBH. I kinda glossed over this since I needed to get a good version API to build the operator. This is certainly something we should all discuss and come up with a plan.

andrewrynhard: @smira If you don't mind, I would like to merge this and then we can discuss our policy on using gRPC data structures in the core packages when we can tomorrow, and come to a consensus on how we want to handle these cases. For now this will unblock me. Super close to having a working operator :D.

smira: didn't mean to block you, I don't know better way, and in this case structs are exactly same, so duplication makes little sense

PR #1198, opened 2019-09-23
fix: set extra kernel args for all platforms

andrewrynhard:


cmd/osctl/cmd/install_linux.go
```diff
@@ -11,9 +11,9 @@ import (
```
andrewrynhard: The changes to this file is the fix. The rest are just renames.

PR #1205, opened 2019-09-24
refactor: decouple grpc client and userdata code

andrewrynhard:
This detangles the gRPC client code from the userdata code. The
motivation behind this is to make creating clients more simple and not
dependent on our configuration format.


osd.log
```diff
@@ -0,0 +1,3 @@
+2019/09/25 22:59:02.534498 main.go:46: failed to create new dynamic certificate provider: failed to load initial certificate: failed to parse cert and key into a TLS Certificate: tls: failed to find any PEM data in certificate input
+2019/09/25 22:59:07.869021 main.go:46: failed to create new dynamic certificate provider: failed to load initial certificate: failed to parse cert and key into a TLS Certificate: tls: failed to find any PEM data in certificate input
+2019/09/25 22:59:13.701818 main.go:46: failed to create new dynamic certificate provider: failed to load initial certificate: failed to parse cert and key into a TLS Certificate: tls: failed to find any PEM data in certificate input
```
bradbeam: delete me

andrewrynhard: thanks

pkg/crypto/x509/x509.go
```diff
@@ -385,6 +387,41 @@ func NewCertificateAndKeyFromFiles(crt, key string) (p *PEMEncodedCertificateAnd
...
+		return nil, nil, fmt.Errorf("failed to decode key")
+	}
+
+	keyEC, err := x509.ParseECPrivateKey(pemBlock.Bytes)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	opts := []Option{}
+	opts = append(opts, DNSNames([]string{hostname}))
```
bradbeam: I think it's fine for this pass, but I'm expecting this will need to be more flexible in the future.

pkg/grpc/tls/provider.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return nil
+}
+
+func (p *embeddableCertificateProvider) manageUpdates(ctx context.Context) (err error) {
+	nextRenewal := constants.DefaultCertificateValidityDuration
+
+	for ctx.Err() == nil {
+		if c, _ := p.GetCertificate(nil); c != nil { // nolint: errcheck
```
bradbeam: nit, can we move the comment above this section? i don't think we have the comment inline elsewhere.

pkg/grpc/tls/provider.go
```diff
@@ -0,0 +1,128 @@
...
+			if len(c.Certificate) > 0 {
```
bradbeam: not specific to this pr ( can be dealt with later ), but does it make sense to have GetCertificate return a slice or just a single cert?

pkg/grpc/tls/provider.go
andrewrynhard: Agree

pkg/crypto/x509/x509.go
andrewrynhard: Yep exactly. This was part of the existing setup.

pkg/grpc/tls/provider.go
andrewrynhard: We can't because its the standard library interface we must implement.

PR #1210, opened 2019-09-26
feat: Discover platform external addresses

bradbeam:
This introduces the functionality for discovering external addresses configured on an intance.
This allows us to automatically append these external addresses to our certificate SANs so we can
access the machines from these addresses without having to know about them ahead of time.

### DIFF
--- a/internal/pkg/platform/gcp/gcp.go
+++ b/internal/pkg/platform/gcp/gcp.go
@@ -5,14 +5,24 @@
 package gcp
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+
 	"github.com/talos-systems/talos/internal/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/userdata"
 	"github.com/talos-systems/talos/pkg/userdata/download"
 )
 
+// Ref: https://cloud.google.com/compute/docs/storing-retrieving-metadata
+// ex, curl -H "Metadata-Flavor: Google" 'http://169.254.169.254/computeMetadata/v1/instance/network-interfaces/?recursive=true'
 const (
 	// GCUserDataEndpoint is the local metadata endpoint inside of DO
 	GCUserDataEndpoint = "http://metadata.google.internal/computeMetadata/v1/instance/attributes/user-data"
+	// GCExternalIPEndpoint displays all external addresses associated with the instance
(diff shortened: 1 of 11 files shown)


internal/pkg/platform/platform.go
```diff
@@ -29,6 +30,7 @@ type Platform interface {
 	UserData() (*userdata.UserData, error)
 	Mode() runtime.Mode
 	Hostname() ([]byte, error)
+	InstanceIPs() ([]net.IP, error)
```
andrewrynhard: nit: Can we call this `ExternalIPs`? I feel like that is a better description here.

internal/app/machined/internal/sequencer/v1alpha1/types.go
```diff
@@ -62,6 +62,10 @@ func (d *Sequencer) Boot() error {
 			"mount extra devices",
 			userdatatask.NewExtraDevicesTask(),
 		),
+		phase.NewPhase(
```
andrewrynhard: This no longer needs to happen because of #1205. The PKI is generated when OSD starts which is the last phase.

internal/pkg/platform/gcp/gcp.go
```diff
@@ -35,5 +45,57 @@ func (gc *GCP) Mode() runtime.Mode {
 
 // Hostname implements the platform.Platform interface.
 func (gc *GCP) Hostname() (hostname []byte, err error) {
+	// curl -Lv -H "Metadata-Flavor: Google" http://169.254.169.254/computeMetadata/v1/instance/name
 	return nil, nil
 }
+
+// InstanceIPs provides any external addresses assigned to the instance
+func (gc *GCP) InstanceIPs() (addrs []net.IP, err error) {
+	// curl -H "Metadata-Flavor: Google" 'http://169.254.169.254/computeMetadata/v1/instance/network-interfaces/?recursive=true'
```
andrewrynhard: For GoDoc purposes, can we put the notes on the endpoint in the function docs?

andrewrynhard:
Seems like Azure, and Packet are missing, and we had a number of other things to figure out with this.

internal/pkg/platform/gcp/gcp_test.go
```diff
@@ -12,3 +12,73 @@ func TestEmpty(t *testing.T) {
 	// please remove it once any unit-test is added
 	// for this package
 }
+
+// TODO use this mock data for tests
```
andrewrynhard: Committing random curl/JSON output is not something I want to get into the practice of.

PR #1214, opened 2019-09-27
chore: update golangci-lint to 19.1

smira:
New linter `whitespace` checks for extra empty lines, I fixed quite a
few of those.

`godox` complains about `TODO:` in comments, but I had to disable it as
we have tons of these.


andrewrynhard:
@smira I'm going to close since we decided to wait for the 1.20 goodness.

PR #1215, opened 2019-09-27
feat: allow Kubernetes version to be configured

andrewrynhard:
This allows for users to specifify which version of Kubernetes to use.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -25,6 +25,7 @@ import (
 	"github.com/talos-systems/talos/cmd/osctl/cmd/cluster/pkg/node"
 	"github.com/talos-systems/talos/cmd/osctl/pkg/client/config"
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
+	"github.com/talos-systems/talos/pkg/constants"
 	"github.com/talos-systems/talos/pkg/userdata/v1alpha1/generate"
 	"github.com/talos-systems/talos/pkg/version"
 )
@@ -105,7 +106,7 @@ func create() (err error) {
 		ips[i] = fmt.Sprintf(baseNetwork, i+2)
 	}
 
-	input, err := generate.NewInput(clusterName, ips)
+	input, err := generate.NewInput(clusterName, ips, kubernetesVersion)
 	if err != nil {
 		return err
 	}
@@ -363,6 +364,7 @@ func init() {
 	clusterUpCmd.Flags().IntVar(&masters, "masters", 3, "the number of masters to create")
 	clusterUpCmd.Flags().StringVar(&clusterCpus, "cpus", "1.5", "the share of CPUs as fraction (each container)")
 	clusterUpCmd.Flags().IntVar(&clusterMemory, "memory", 1024, "the limit on memory usage in MB (each container)")
+	clusterUpCmd.Flags().StringVar(&kubernetesVersion, "kubernetes-version", constants.DefaultKubernetesVersion, "desired kubernetes version to run")
 	clusterCmd.PersistentFlags().StringVar(&clusterName, "name", "talos_default", "the name of the cluster")
 	clusterCmd.AddCommand(clusterUpCmd)
 	clusterCmd.AddCommand(clusterDownCmd)
--- a/pkg/userdata/translate/translate_v1alpha1.go
+++ b/pkg/userdata/translate/translate_v1alpha1.go
@@ -15,7 +15,6 @@ import (
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 	kubeadm "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2"
 
-	"github.com/talos-systems/talos/pkg/constants"
 	"github.com/talos-systems/talos/pkg/crypto/x509"
 	"github.com/talos-systems/talos/pkg/userdata"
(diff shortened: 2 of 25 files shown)


pkg/userdata/translate/translate_v1alpha1.go
```diff
@@ -218,7 +218,7 @@ func translateV1Alpha1Init(nc *v1alpha1.NodeConfig, ud *userdata.UserData) error
 			APIVersion: "kubeadm.k8s.io/v1beta2",
 		},
 		ClusterName:          nc.Cluster.ClusterName,
-		KubernetesVersion:    constants.KubernetesVersion,
+		KubernetesVersion:    ud.KubernetesVersion,
```
Ulexus: Since this is going to be user-supplied data, I would consider running it through a little bit of processing/validation.

At a minimum, I would `strings.TrimPrefix(ud.KubernetesVersion,"v")`.

cmd/osctl/cmd/cluster.go
```diff
@@ -105,7 +106,7 @@ func create() (err error) {
 		ips[i] = fmt.Sprintf(baseNetwork, i+2)
 	}
 
-	input, err := generate.NewInput(clusterName, ips)
+	input, err := generate.NewInput(clusterName, ips, kubernetesVersion)
```
bradbeam: /cc @rsmitty fyi for capi

bradbeam:
How are we/should we handle kubeadm versions?

andrewrynhard:
> How are we/should we handle kubeadm versions?

For this initial pass, we won't. But I believe we talked about building our own kubelet container and putting `kubeadm` into it.

pkg/userdata/translate/translate_v1alpha1.go
andrewrynhard: Yeah this is gonna be tricky. Upstream Kubernetes publishes images with the "v", and we can't always guarantee that if people use a custom registry (not currently possible) that the versions will be semantic with a "v" or no "v". For now we are pinned to the official GCR registry and they have the "v" in the tags. Any kind of validation you think we can/should do knowing this?

Ulexus: This is something we deal with a lot in telephony.  (Do we dial a "1" or not?  Do we dial a "+" or not?  Do we dial a "9" or not?)

The best solution is generally to break it down into a root structure on the parsing side (strip the v and validate semver), and then format it on the use side (add a "v" if required by the system).

I'm not sure that translates well here since, as you note, the tags are actually free-form.

Never-the-less, if we call it "kubernetes version" as opposed to "container tag," I think there _should_ be a canonical form that is used internally.

andrewrynhard: hmm... I think maybe requiring a valid semver would be nice? There is a library that has a `Parse` requiring strict semver.

bradbeam: > I think maybe requiring a valid semver would be nice?
I'm hesitant to do that because it'd limit any end user built kubernetes. Should be the rare case and we don't currently support it ( given we don't flex on `constants.KubernetesImage` ).

andrewrynhard: Yeah :/, its just if not semver then what is there to validate that we can depend on?

Ulexus: Yeah, that's why this is a little messy.  The actual versions should be semver, but the container tags may not be.  It could be useful to offer both.  Among other things, we really will need to know the actual version for other things.

Perhaps we could add an option (later?) which would allow the user to override the computed container tag which still retaining a formal kubernetes version.

andrewrynhard: ^ sounds like a good start for now ... we have the rest of 0.3 to figure this out, and this at least unblocks the upgrade operator.

andrewrynhard: is there harm in forcing users to use proper semantic versioning?

Ulexus: I definitely see issues with that for companies which have their own add-ons and iterations.  Might work if they are forced into a particular structure, like:
`v1.16.1-mycompany-r13`

andrewrynhard: fair point

Ulexus: Regardless, none of this should hold up this PR.  I was just mentioning the very common quandary of `v1.16.1` vs `1.16.1` when dealing with user-supplied values.

Ulexus: Which is to say that _I_ would generally strip the "v" on input and add it as necessary on output.

andrewrynhard: I'm adding semver check for now, until we can figure out a better way. The good news is that due to talos CAPI changes, we can start running the validation code when generating configs.

andrewrynhard:
@bradbeam @Ulexus I added simple semver check for now:

```
[I] ~/s/g/t/talos configurable-kubernetes-version* (3c861b29) $ ./build/osctl-linux-amd64 cluster create --masters=1 --workers=1 --kubernetes-version=v1.15.0
generating PKI and tokens
creating network talos_default
creating node master-1
failed to create node: 1 error occurred:
	* version must be semantic: Invalid character(s) found in major number "v1"
```

PR #1218, opened 2019-09-27
fix: set default node identity cert to expire in 10 years

andrewrynhard:
Since trustd does not rotate its' certificate in v0.2, we must generate
the default node identity certificate to 10 years. This has been fixed
in v0.3, but the implementation is too risky to cherry-pick.

### DIFF
--- a/internal/app/machined/internal/phase/userdata/pki.go
+++ b/internal/app/machined/internal/phase/userdata/pki.go
@@ -6,6 +6,7 @@ package userdata
 
 import (
 	"log"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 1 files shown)


pkg/userdata/userdata.go
```diff
@@ -126,6 +127,7 @@ func (data *UserData) NewIdentityCSR() (csr *x509.CertificateSigningRequest, err
 	names := []string{hostname}
 	opts = append(opts, x509.DNSNames(names))
 	opts = append(opts, x509.IPAddresses(ips))
+	opts = append(opts, x509.NotAfter(time.Now().Add(87600*time.Hour)))
```
andrewrynhard: Now that I'm looking at this, this option is only valid for the certificate creator.

PR #1221, opened 2019-09-29
feat: Allow env override of hack/qemu image location

### LINKED ISSUE #1220: feat: allow override of hack/qemu image location
For systems such as mine which have `/tmp` locked down, we cannot run qemu on a `/tmp`-sourced image.  It would be a minor quality-of-life improvement to be able to override the location.

Ulexus:
This fixes #1220

### DIFF
--- a/hack/dev/qemu-boot.sh
+++ b/hack/dev/qemu-boot.sh
@@ -2,6 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


Ulexus:
I'm pretty sure that drone failure doesn't have anything to do with this change.  It looks like contention on the os signal bus?

andrewrynhard:
👍

PR #1223, opened 2019-09-30
feat: Add etcd ca generation to userdata.Generate

bradbeam:
Allow us to generate/manage tls assets external to kubeadm.


pkg/userdata/generate/generate.go
```diff
@@ -282,6 +284,22 @@ func NewInput(clustername string, masterIPs []string, kubernetesVersion string)
 	if err != nil {
 		return nil, err
 	}
+	caPemBlock, _ := pem.Decode(osCert.CrtPEM)
```
bradbeam: Most of the red here is just grouping things up a little better

pkg/userdata/generate/controlplane.go
```diff
@@ -16,6 +16,9 @@ security:
     ca:
       crt: "{{ .Certs.K8sCert }}"
       key: "{{ .Certs.K8sKey }}"
+    etcd:
```
Ulexus: I know this is all kind of temporary, but etcd should not be under kubernetes.  Both topologically and from a PKI standpoint, it should be a separate top-level category.

Further, etcd has two PKI groups:  peering and client-server, and even if we use the same for both right now, the userdata should reflect these distinctions (even if it is just a subcategorization).

So, for instance, I would use security.etcd.peer.ca and security.etcd.service.ca and default the service PKI to peer, if it is not defined.

pkg/userdata/generate/generate.go
```diff
@@ -313,34 +331,31 @@ func NewInput(clustername string, masterIPs []string, kubernetesVersion string)
...
 	if err != nil {
 		return nil, err
 	}
 
 	certs := &Certs{
 		AdminCert: base64.StdEncoding.EncodeToString(adminCrt.X509CertificatePEM),
 		AdminKey:  base64.StdEncoding.EncodeToString(adminKey.KeyPEM),
-		OsCert:    base64.StdEncoding.EncodeToString(osCert.CrtPEM),
-		OsKey:     base64.StdEncoding.EncodeToString(osCert.KeyPEM),
+		EtcdCert:  base64.StdEncoding.EncodeToString(etcdCert.CrtPEM),
```
Ulexus: Same as above:  EtcdPeerCert, EtcdServiceCert

pkg/userdata/v1alpha1/cluster_config.go
```diff
@@ -55,6 +55,8 @@ type SchedulerConfig struct {
 // EtcdConfig represents etcd config vals
 type EtcdConfig struct {
 	Image string `yaml:"image,omitempty"`
+	Crt   string `yaml:"crt"`
```
Ulexus: It is unclear that this is the CA cert, not the peer, service, or client certs (assuming it _is_ the CA cert, of course).

pkg/userdata/generate/controlplane.go
andrewrynhard: +1

bradbeam: think this is addressed in the latest commit. 
```
security.etcd.ca.key // security.etcd.ca.crt
```

pkg/userdata/generate/generate.go
bradbeam: the above is just internal, I'm not sure it matters much ( ex, we don't really distinguish between the OS level certs ( osd )).

PR #1224, opened 2019-09-30
feat: allow for host level etcd

andrewrynhard:
This allows users to create an etcd service using the host init system.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -1,0 +1,331 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package services
+
+import (
+	"context"
+	stdlibx509 "crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"io/ioutil"
+	"log"
+	stdlibnet "net"
+	"os"
+	"strings"
+	"time"
+
+	containerdapi "github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
+	"github.com/containerd/containerd/oci"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/pkg/transport"
+
+	"github.com/talos-systems/talos/internal/app/machined/pkg/system/conditions"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/containerd"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/restart"
+	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/crypto/x509"
+	"github.com/talos-systems/talos/pkg/kubernetes"
+	"github.com/talos-systems/talos/pkg/net"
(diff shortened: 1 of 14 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+
+	// Set the process arguments.
+	args := runner.Args{
+		ID: e.ID(data),
+		ProcessArgs: []string{
+			"/usr/local/bin/etcd",
+			"--name=" + hostname,
+			"--listen-client-urls=https://0.0.0.0:2380",
```
Ulexus: We should probably make a package which takes an IP address (ips[0] in this case), checks whether it is IPv4 or IPv6, and offers AllAddress() (0.0.0.0 or ::) and LoopBackAddress() (127.0.0.1 or ::1) for each.

andrewrynhard: Good point 👍

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -53,6 +53,9 @@ func (k *Kubelet) Condition(data *userdata.UserData) conditions.Condition {
 
 // DependsOn implements the Service interface.
 func (k *Kubelet) DependsOn(data *userdata.UserData) []string {
+	if data.Services.Kubeadm.IsControlPlane() {
+		return []string{"containerd", "etcd", "kubeadm"}
```
andrewrynhard: I wonder if we should depend on etcd here. Do we gain anything? Probably faster bootstrap times.

pkg/constants/constants.go
```diff
@@ -115,17 +115,23 @@ const (
...
 	DefaultKubernetesVersion = "1.16.0"
 
 	// KubernetesImage is the enforced hyperkube image to use for the control plane.
 	KubernetesImage = "k8s.gcr.io/hyperkube"
 
+	// DefaultEtcdVersion is the default target version of etcd.
+	DefaultEtcdVersion = "3.3.15-0"
+
+	// EtcdImage is the reposistory for the etcd image.
+	EtcdImage = "k8s.gcr.io/etcd"
```
andrewrynhard: would love to use quay, but ran into an issue with the manifest version bing incompatible with containerd.

pkg/crypto/x509/x509.go
```diff
@@ -78,6 +85,13 @@ type Options struct {
 // Option is the functional option func.
 type Option func(*Options)
 
+// CommonName sets the common name of the certificate.
+func CommonName(o string) Option {
```
andrewrynhard: Kubernetes PKI auth doesn't work without a CN.

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Nope. Kubeadm still depends on etcd. Maybe when we move to bootkube we will get this benefit.

andrewrynhard:
Seeing the following in the etcd logs:

```
2019-10-03 03:05:40.578508 I | embed: rejected connection from "10.5.0.4:40210" (error "x509: certificate is valid for 10.5.0.2, 127.0.0.1, not 10.5.0.4", ServerName "", IPAddresses ["10.5.0.2" "127.0.0.1"], DNSNames [])
2019-10-03 03:05:42.936813 I | embed: rejected connection from "10.5.0.3:47532" (error "x509: certificate is valid for 10.5.0.2, 127.0.0.1, not 10.5.0.3", ServerName "", IPAddresses ["10.5.0.2" "127.0.0.1"], DNSNames [])
```

internal/app/machined/pkg/system/services/etcd.go
andrewrynhard: @Ulexus gonna follow up with a IPv6/IPV4 dual stack helper library.

PR #1225, opened 2019-09-30
feat: add configurator interface

andrewrynhard:
This moves from translating a config into an internal config
representation, to using an interface. The idea is that and interface
gives us stronger compile time checks, and an interface will prevent us
from having to copy from on struct to another. As long as a concrete
type implements the Configurator interface, it can be used to provide
instructions to Talos.


cmd/osctl/cmd/cluster.go
```diff
@@ -172,24 +173,16 @@ func createNodes(requests []*node.Request) (err error) {
...
-	for idx, req := range requests {
-		go func(idx int, req *node.Request) {
+	for _, req := range requests {
+		go func(req *node.Request) {
 			fmt.Println("creating node", req.Name)
-			// We'll use this to denote the previous node
-			// IP so we can coordinate how each control plane
-			// node comes up
-			// 1 <- 2 <- 3 <- 4 <- 5 ...
-			req.Input.Index = idx
```
andrewrynhard: We no longer depend on the index.

PR #1227, opened 2019-10-01
feat: Add kubeadm flex on etcd if service is enabled

bradbeam:
This allows us to dynamically set in the kubeadm configuration an external etcd instance.

### DIFF
--- a/internal/app/machined/pkg/system/services/kubeadm/kubeadm.go
+++ b/internal/app/machined/pkg/system/services/kubeadm/kubeadm.go
@@ -13,6 +13,7 @@ import (
 	"math"
 	"os"
 	"path"
+	"strconv"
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/system/services/kubeadm/kubeadm.go
```diff
@@ -103,6 +104,24 @@ func editClusterConfig(data *userdata.UserData) (err error) {
 		return err
 	}
 
+	// Flex on running etcd directly on Talos or via Kubeadm
+	if data.Services.Etcd.Enabled {
```
andrewrynhard: Need to check if `data.Services.Etcd` is `nil`. I just hit this in my PR.

PR #1234, opened 2019-10-04
chore: Remove unused code in kubeadm service

bradbeam:
We no longer rely on trustd to propagate cert info for kubernetes.

Not sure its worth merging since this service will be removed shortly. 


andrewrynhard:
Merged bootkube change.

PR #1235, opened 2019-10-04
feat: use bootkube for cluster creation

andrewrynhard:
This replaces kubeadm with bootkube.


MarkusTeufelberger:
...why? Bootkube seems like it is dying a slow death for years now and kubeadm sees all the action.

andrewrynhard:
> ...why? Bootkube seems like it is dying a slow death for years now and `kubeadm` sees all the action.

Good question :) . This wasn't an easy decision. I have personally contributed to `kubeadm` a number of times, and been a fan of it for a long time. We have been running it for some time now, and we found a number of reasons we felt it was better to move to `bootkube`.

One of the primary reasons is that we want to move to a self-hosted control plane. We feel that self-hosted complements Talos very well, and can provide better opportunities as we implement our upgrade operator. Also, with kubeadm, upgrades didn't quite work the way we had hoped. The decoupling that self-hosted gives us, makes for a very clean separation of concerns. Self-hosting doesn't seem to be the priority of `kubeadm` whereas with `bootkube` it is.

It felt like, to us at least, that `kubeadm` was very much designed for the typical workflow of creating a machine, ssh-ing into it, and performing post-boot provisioning. No surprise there as that is what 99.99% of all other Linux distros allow you to do. Things are different with Talos. As I'm sure you know, there is no SSH or console, so we found ourselves trying to fit a square peg into a round hole with `kubeadm`. Since `bootkube` is so lightweight and does not live in the main Kubernetes repo, we are able to have finer grained control of its' behavior.

At the end of the day, the aim with Talos is to abstract cluster creation and management away to a point where users don't really have to be concerned about what tool we use under the hood. We run full upstream conformance tests once a day, and if we pass that, we feel we are achieving a large portion our goals.

We intend on helping with `bootkube`, and hopefully, improving it for use outside of Talos. If we can help make it the defacto standard tool for _self-hosted_ Kubernetes, then we will. We see a lot of potential in `bootkube` and intend on being involved with it going forward.

MarkusTeufelberger:
Kubeadm also supports self hosting (https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-alpha/#cmd-selfhosting) and looking at bootkube I'd rather see this feature get towards beta in kubeadm than vendoring and fixing bootkube. But hey, it's your project, good luck with it, I hope it works out. Personally I'd rather have the choice to NOT self-host inside the cluster and am rather happy with the middle ground of static pods for control plane components on master nodes. Makes a scheduler or ETCD failure slightly less catastrophic.

andrewrynhard:
@MarkusTeufelberger Fair points. I have personally worked on self-hosting with `kubeadm`, but the issue, for us at least, is that vendoring `kubeadm` is extremely painful and imposes constraints on our design. I would also like to see it graduate in `kubeadm`, and I think it would make sense for the Talos team to even try and spearhead that, but the decision to move to `bootkube` isn't based on self-hosting alone. We would like to see `kubeadm` do the following:

- move outside of the k/k repo
- become more library like (with bootkube we can import the code and control it better)
- graduate self-hosting (we could help here)

As for the option to self host or not, I think that is a valid point and we will discuss this in our meeting next week. FWIW, we will not self-host Etcd. That will be on host and created and managed by Talos. I would like to invite you to our [slack](https://slack.dev.talos-systems.io), and/or our weekly meeting (see the README) to discuss this some more. The more opinions/feedback we can get the better 👍.

internal/app/machined/internal/phase/services/post_boot.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
+	"github.com/talos-systems/talos/internal/pkg/runtime"
+	"github.com/talos-systems/talos/pkg/config/machine"
+	"github.com/talos-systems/talos/pkg/kubernetes"
+)
+
+// LabelNodeAsMaster represents the LabelNodeAsMaster task.
+type LabelNodeAsMaster struct{}
```
Ulexus: nit:  LabelNodeAsMaster -> LabelNodeIfMaster? (since it is safe to be called on any node and will only label it a master _if_ it is a master.

internal/app/machined/internal/phase/services/post_boot.go
```diff
@@ -0,0 +1,57 @@
...
+	return &LabelNodeAsMaster{}
+}
+
+// RuntimeFunc returns the runtime function.
+func (task *LabelNodeAsMaster) RuntimeFunc(mode runtime.Mode) phase.RuntimeFunc {
+	return task.standard
+}
+
+func (task *LabelNodeAsMaster) standard(args *phase.RuntimeArgs) (err error) {
+	if args.Config().Machine().Type() == machine.Worker {
```
Ulexus: Personally, I would feel safer inverting this check, lest other node types eventually arise.

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Runner implements the Service interface.
+func (b *Bootkube) Runner(config config.Configurator) (runner.Runner, error) {
+	return goroutine.NewRunner(config, "bootkube", bootkube.NewService().Main), nil
+}
+
+// nolint: gocyclo
+func generateAssets(config config.Configurator) (err error) {
+	if err = os.MkdirAll("/etc/kubernetes/manifests", 0644); err != nil {
```
Ulexus: This strikes me as either an odd placement or a mismatch.  Nothing else here seems to use this directory, as declared.  There _is_ a write to `constants.AssetsDirectory`, however.

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		apiServers = append(apiServers, u)
+	}
+	u, err := url.Parse("https://127.0.0.1:6443")
+	if err != nil {
+		return err
+	}
+	apiServers = append(apiServers, u)
+
+	_, podCIDR, err := net.ParseCIDR("10.2.0.0/16")
```
Ulexus: 10.2.0.0/16<->DefaultPodCIDR
(Or do we expose this in config?)

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return err
+	}
+	_, serviceCIDR, err := net.ParseCIDR("10.3.0.0/24")
```
Ulexus: 10.3.0.0/24<->DefaultServiceCIDR
(Or is this exposed by config?)

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+	conf := asset.Config{
+		CACert:                 k8sCA,
+		CAPrivKey:              k8sKey,
+		EtcdCACert:             ca,
+		EtcdClientCert:         peer,
+		EtcdClientKey:          key,
+		EtcdServers:            []*url.URL{etcdServer},
+		EtcdUseTLS:             true,
+		APIServers:             apiServers,
+		APIServiceIP:           net.ParseIP("10.3.0.1"),
```
Ulexus: DefaultKubernetesServiceAddress?

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+		DNSServiceIP:           net.ParseIP("10.3.0.10"),
```
Ulexus: DefaultKubernetesDNSServiceAddress?

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -0,0 +1,210 @@
...
+		PodCIDR:                podCIDR,
+		ServiceCIDR:            serviceCIDR,
+		NetworkProvider:        "flannel",
```
Ulexus: DefaultCNI?

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -66,22 +133,22 @@ func (k *Kubelet) Runner(config config.Configurator) (runner.Runner, error) {
...
-			"--runtime-request-timeout=15m",
 			"--container-runtime-endpoint=unix://" + constants.ContainerdAddress,
+			"--anonymous-auth=false",
+			"--cert-dir=/var/lib/kubelet/pki",
+			"--client-ca-file=" + constants.KubernetesCACert,
+			"--cni-conf-dir=/etc/cni/net.d",
+			"--cluster-domain=cluster.local",
+			"--pod-manifest-path=/etc/kubernetes/manifests",
+			"--rotate-certificates",
+			"--cluster-dns=10.3.0.10",
```
Ulexus: DefaultKubernetesDNSServiceAddress?

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -66,22 +133,22 @@ func (k *Kubelet) Runner(config config.Configurator) (runner.Runner, error) {
...
 			"--container-runtime=remote",
-			"--runtime-request-timeout=15m",
 			"--container-runtime-endpoint=unix://" + constants.ContainerdAddress,
+			"--anonymous-auth=false",
+			"--cert-dir=/var/lib/kubelet/pki",
+			"--client-ca-file=" + constants.KubernetesCACert,
+			"--cni-conf-dir=/etc/cni/net.d",
+			"--cluster-domain=cluster.local",
+			"--pod-manifest-path=/etc/kubernetes/manifests",
+			"--rotate-certificates",
```
Ulexus: //TODO check to see that we have ca key before enabling auto-certificate rotation

pkg/config/types/v1alpha1/cluster_config.go
```diff
@@ -143,212 +117,25 @@ func (e *EtcdConfig) CA() *x509.PEMEncodedCertificateAndKey {
...
-			// KubeletExtraArgs: ,
-		},
-	}
-
-	encodedObjs := [][]byte{}
-	encoded, err := kubeadmutil.MarshalToYamlForCodecs(controlPlaneConfig, kubeadmv1beta2.SchemeGroupVersion, kubeadmscheme.Codecs)
-	if err != nil {
-		return "", err
+// ID implements the Configurator interface.
+func (c *ClusterConfig) ID() string {
```
Ulexus: It seems very odd to have `ClusterConfig.ID()` return the cluster's node bootstrap ID

pkg/config/types/v1alpha1/cluster_config.go
```diff
@@ -143,212 +117,25 @@ func (e *EtcdConfig) CA() *x509.PEMEncodedCertificateAndKey {
...
-		NodeRegistration: kubeadm.NodeRegistrationOptions{
-			CRISocket: constants.ContainerdAddress,
-			// KubeletExtraArgs: ,
-		},
-	}
-
-	if err := cis.EnforceWorkerRequirements(joinConfig); err != nil {
-		return "", err
+// Secret implements the Configurator interface.
+func (c *ClusterConfig) Secret() string {
```
Ulexus: It seems very odd that ClusterConfig.Secret() returns the cluster's node bootstrap secret

internal/app/machined/internal/phase/services/post_boot.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < 200; i++ {
```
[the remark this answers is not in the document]
andrewrynhard: We have a retry library pending that will fix this.

internal/app/machined/internal/phase/services/post_boot.go
andrewrynhard: Good point.

internal/app/machined/pkg/system/services/bootkube.go
andrewrynhard: Agree. Can fix in follow up.

internal/app/machined/pkg/system/services/bootkube.go
andrewrynhard: For now at least. As discussed today, we'd like to give the ability to point this somewhere remote.

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: 🤦‍♂ whoops

pkg/config/types/v1alpha1/cluster_config.go
andrewrynhard: This uses the nomenclature from Kubernetes. A bootstrap token has and id and a secret portion. Separated by the `.`.

pkg/config/types/v1alpha1/cluster_config.go
andrewrynhard: This uses the nomenclature from Kubernetes. A bootstrap token has and id and a secret portion. Separated by the `.`.

andrewrynhard:
@Ulexus sorry I merged before your feedback. I will address in follow ups this week.

pkg/config/types/v1alpha1/cluster_config.go
Ulexus: Right, I'm talking about the namespace:  bootstrap token aspects do not seem to be well-situated in the root of cluster config.  Put another way, one would expect `ClusterConfig.ID()` would return something like an ID of the _cluster_, not the bootstrap token _of_ the cluster.

andrewrynhard: ahh yes, this was just a struct where we could satisfy the interface.

PR #1238, opened 2019-10-05
chore: remove hyperkube dependency in CI

andrewrynhard:
The hyperkube image is bloated, and full of things we don't need. This
downloads kubectl and mounts it into an alpine image instead.


andrewrynhard:
We get zero benefit from this. The cluster is still bootstrapping by the time hypercube downloads so we might as well keep as is.

PR #1239, opened 2019-10-05
feat: add retry package

andrewrynhard:
This package provides a consistent way for us to retry arbitrary logic.
It provides the following backoff algorithms:

- exponential
- linear
- constant

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -10,7 +10,6 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
-	"log"
 	stdlibnet "net"
 	"os"
 	"strings"
@@ -34,6 +33,7 @@ import (
 	"github.com/talos-systems/talos/pkg/crypto/x509"
 	"github.com/talos-systems/talos/pkg/kubernetes"
 	"github.com/talos-systems/talos/pkg/net"
+	"github.com/talos-systems/talos/pkg/retry"
 )
 
 var etcdImage = fmt.Sprintf("%s:%s", constants.EtcdImage, constants.DefaultEtcdVersion)
@@ -304,13 +304,12 @@ func buildInitialCluster(config config.Configurator, name, ip string) (initial s
 		return "", err
 	}
 
-	for i := 0; i < 200; i++ {
-		endpoints, err := h.MasterIPs()
+	opts := []retry.Option{retry.WithUnits(3 * time.Second), retry.WithJitter(time.Second)}
+	err = retry.Constant(10*time.Minute, opts...).Retry(func() error {
+		var endpoints []string
+		endpoints, err = h.MasterIPs()
 		if err != nil {
-			log.Printf("failed to get client endpoints: %+v\n", err)
-			time.Sleep(3 * time.Second)
-
-			continue
+			return retry.ExpectedError(err)
 		}
 
 		// Etcd expects host:port format.
@@ -320,12 +319,10 @@ func buildInitialCluster(config config.Configurator, name, ip string) (initial s
 
 		peerAddrs := []string{"https://" + ip + ":2380"}
 
-		resp, err := addMember(endpoints, peerAddrs)
+		var resp *clientv3.MemberAddResponse
+		resp, err = addMember(endpoints, peerAddrs)
 		if err != nil {
-			log.Printf("failed to add etcd member: %+v\n", err)
-			time.Sleep(3 * time.Second)
-
-			continue
+			return retry.ExpectedError(err)
 		}
 
 		newID := resp.Member.ID
@@ -344,8 +341,12 @@ func buildInitialCluster(config config.Configurator, name, ip string) (initial s
 
 		initial = strings.Join(conf, ",")
 
-		return initial, nil
+		return nil
+	})
+
+	if err != nil {
+		return "", errors.New("failed to discover etcd cluster")
 	}
 
-	return "", errors.New("failed to discover etcd cluster")
+	return initial, nil
 }
--- a/internal/pkg/mount/mount.go
+++ b/internal/pkg/mount/mount.go
@@ -20,27 +20,27 @@ import (
 	gptpartition "github.com/talos-systems/talos/pkg/blockdevice/table/gpt/partition"
 	"github.com/talos-systems/talos/pkg/blockdevice/util"
 	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/retry"
 )
 
 // RetryFunc defines the requirements for retrying a mount point operation.
 type RetryFunc func(*Point) error
 
-func retry(f RetryFunc, p *Point) (err error) {
-	for i := 0; i < 50; i++ {
+func mountRetry(f RetryFunc, p *Point) (err error) {
+	err = retry.Constant(5*time.Second, retry.WithUnits(50*time.Millisecond)).Retry(func() error {
 		if err = f(p); err != nil {
 			switch err {
 			case unix.EBUSY:
-				time.Sleep(100 * time.Millisecond)
-				continue
+				return retry.ExpectedError(err)
 			default:
-				return err
+				return retry.UnexpectedError(err)
 			}
 		}
 
 		return nil
-	}
+	})
 
-	return errors.Errorf("timeout: %+v", err)
+	return err
 }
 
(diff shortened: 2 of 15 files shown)


andrewrynhard:
I didn't replace everywhere we are doing retry logic with this package. Up to you guys if we should, or we can follow up with another PR/multiple PRs. I'm good to either do it now, or later.

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -296,12 +296,11 @@ func buildInitialCluster(config config.Configurator, name, ip string) (initial s
 		return "", err
 	}
 
-	for i := 0; i < 200; i++ {
-		endpoints, err := h.MasterIPs()
+	err = retry.Constant(10*time.Minute, retry.WithUnits(3*time.Second)).Retry(func() error {
```
andrewrynhard: This says "retry every 3 seconds for 10 minutes."

internal/pkg/mount/mount.go
```diff
@@ -20,26 +20,27 @@ import (
...
+	"github.com/talos-systems/talos/pkg/retry"
 )
 
 // RetryFunc defines the requirements for retrying a mount point operation.
 type RetryFunc func(*Point) error
 
-func retry(f RetryFunc, p *Point) (err error) {
-	for i := 0; i < 50; i++ {
+func mountRetry(f RetryFunc, p *Point) (err error) {
+	err = retry.Constant(5*time.Second, retry.WithUnits(50*time.Millisecond)).Retry(func() error {
```
andrewrynhard: This says "retry every 50 milliseconds for 5 seconds"

pkg/blockdevice/probe/probe.go
```diff
@@ -60,15 +61,18 @@ func FileSystem(path string) (sb filesystem.SuperBlocker, err error) {
 	// Sleep for up to 5s to wait for kernel to create the necessary device files.
 	// If we dont sleep this becomes racy in that the device file does not exist
 	// and it will fail to open.
-	for i := 0; i <= 100; i++ {
+	err = retry.Constant(5 * time.Second).Retry(func() error {
```
andrewrynhard: This says "retry every second (default unit) for  seconds."

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -296,12 +296,12 @@ func buildInitialCluster(config config.Configurator, name, ip string) (initial s
 		return "", err
 	}
 
-	for i := 0; i < 200; i++ {
-		endpoints, err := h.MasterIPs()
+	opts := []retry.Option{retry.WithUnits(3 * time.Second), retry.WithJitter()}
+	err = retry.Constant(10*time.Minute, opts...).Retry(func() error {
```
andrewrynhard: This says "retry every 3 seconds, with [0, 3 seconds) jitter, for 10 minutes."

pkg/retry/linear.go
```diff
@@ -0,0 +1,72 @@
...
+		for {
+			select {
+			case <-l.stop:
+				return
+			default:
+				d := time.Duration(c)*l.options.Units + l.Jitter()
+
+				c++
+				l.C <- time.Now()
+				time.Sleep(d)
```
smira: not sure, but sounds like this code won't be stoppable while `Sleep` is running?

It might make sense to refactor this to `<-time.After` in `select` with `<-l.stop`?

andrewrynhard:
@smira I updated the code to address you comments. PTAL when you get a chance 👍.

andrewrynhard:
@Ulexus here is the retry package I mentioned.

pkg/retry/linear.go
andrewrynhard: Addressed in latest changes.

pkg/retry/retry.go
```diff
@@ -0,0 +1,112 @@
...
+
+func retry(f RetryableFunc, d time.Duration, t Ticker) error {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			return TimeoutError{}
+		case <-t.Stop():
```
andrewrynhard: @smira This should address the concern about returning on a ticker's `Stop()`.

pkg/retry/retry.go
```diff
@@ -0,0 +1,112 @@
...
+			return nil
+		default:
+			if err := f(); err != nil {
+				switch err.(type) {
+				case expectedError:
+					time.Sleep(t.Tick())
```
andrewrynhard: @smira I think this does what we discussed in slack? So the total time taken is now `time(f()) + time.Sleep(t.Tick())`.

smira: if we need it to be stoppable while sleeping, it makes sense to rewrite this a bit:

```suggestion
					select {
					case <-time.After(t.Tick()):
					case <-t.Stop():
					    return nil
					 }
```

smira: and I don't know how to actually handle all the cases if we want to:

* global timeout exceeded
* stop
* wait between calls
* in a call to `f()`

If we want it to be stoppable all the time, probably we should have a context which cancels both `f()` and any wait we have here, so on the surface this might be:

```
1. call f(ctx)
2. wait for either: ctx cancel, end of interval between retries, global timeout
```

if we want global timeout to cancel `f(ctx)` while running, we need to run `f` in goroutine and cancel its `ctx` if global timeout is reached...

so much here but not sure what kind of requirements we have

andrewrynhard: For now our needs are pretty basic. We just need to attempt something for a given duration. Fair to say we can follow up with a more complete implementation?

pkg/retry/retry.go
```diff
@@ -0,0 +1,112 @@
...
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-timer.C:
+			return TimeoutError{}
+		case <-t.Stop():
+			return nil
+		default:
```
Ulexus: If, instead of default, you put your between-tries timer here, as a case, you could eliminate the `time.Sleep()` below.

Ulexus: Then _after_ the select, you can call your function and handle its error.

smira: that's exactly what I was going after in the comment above, but Sean's version is more clear :)

pkg/retry/retry.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if t.rand == nil {
+		t.rand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	}
+
+	return time.Duration(t.rand.Int63n(int64(t.options.Jitter)))
+}
+
+func (t ticker) Stop() <-chan struct{} {
```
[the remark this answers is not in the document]
andrewrynhard: Moved to `StopChan` and exposed `Stop` publicly.

pkg/blockdevice/probe/probe.go
bradbeam: Given that we sleep 50ms before, we probably want to go with something lower than 1s? Probably insignificant in the long run, but not sure there's a need to change.

pkg/retry/options.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+// WithJitter is a functional option for setting the jitter flag.
+func WithJitter(o time.Duration) Option {
+	return func(args *Options) {
+		args.Jitter = o
+	}
+}
+
```
bradbeam: I feel like we should expose a max interval setting ( specifically for exponential backoff ). Ex, I want to wait a total 10min with a max of 30s intervals and exponentially grow from 1s.

pkg/retry/retry.go
```diff
@@ -0,0 +1,123 @@
...
+	return unexpectedError{err}
+}
+
+func retry(f RetryableFunc, d time.Duration, t Ticker) error {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+
+	// We run the func first to avoid having to wait for the next tick.
+	if err := f(); err != nil {
+		_, ok := err.(unexpectedError)
```
bradbeam: nit, worth squashing this into a single line?
```
if _, ok := err.(unexpectedError); ok {
  return err
}
```

pkg/retry/options.go
andrewrynhard: Potentially. Certainly something we can improve on.

PR #1246, opened 2019-10-08
docs: add the k8s certified logo to the intro page

tgerla:

### DIFF
--- a/docs/content/talos.md
+++ b/docs/content/talos.md
@@ -8,6 +8,14 @@ menu:
     weight: 1
 ---
 
+<!-- markdownlint-disable no-inline-html -->
+<a href="https://www.cncf.io/certification/software-conformance/">
+  <img class="object-contain h-32 float-right"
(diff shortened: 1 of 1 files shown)


docs/content/talos.md
```diff
@@ -8,6 +8,14 @@ menu:
     weight: 1
 ---
 
+<!-- markdownlint-disable no-inline-html -->
+<a href="https://www.cncf.io/certification/software-conformance/">
+  <img class="object-contain h-32 float-right"
+       src="/images/certified-kubernetes-color.svg"
```
andrewrynhard: @tgerla It looks like the `certified-kubernetes-color.svg` needs to be checked in too? Also, I think we can't technically add this to the master branch since we will eventually move to Kubernetes 1.17 and we won't be certified for that then. I think we need to make this PR against the `release-0.2` branch. 👍.

andrewrynhard:
@tgerla I'm going to close since we have the new docs now.

PR #1247, opened 2019-10-08
feat: add config validation task

andrewrynhard:
This should provide a better UX around misconfigured Talos nodes. It is
just the start of something we can expand on.

### DIFF
--- a/pkg/config/types/v1alpha1/config.go
+++ b/pkg/config/types/v1alpha1/config.go
@@ -5,8 +5,10 @@
 package v1alpha1
 
 import (
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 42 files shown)


pkg/config/types/v1alpha1/config.go
```diff
@@ -24,33 +26,47 @@ const (
...
 // Cluster implements the Configurator interface.
-func (n *Config) Cluster() cluster.Cluster {
-	return n.ClusterConfig
+func (c *Config) Cluster() cluster.Cluster {
+	return c.ClusterConfig
 }
 
 // Validate implements the Configurator interface.
-func (n *Config) Validate() error {
+func (c *Config) Validate(mode runtime.Mode) error {
```
andrewrynhard: This is a simple implementation for now. I plan to follow up with more robust checks and tests.

PR #1248, opened 2019-10-08
chore: re-enable end to end tests

rsmitty:
This PR will add the bits necessary to make use of changes to our
v1alpha1 cluster api provider for CI testing. This is needed since we've
had machine config changes.

### DIFF
--- a/hack/test/manifests/aws-cluster.yaml
+++ b/hack/test/manifests/aws-cluster.yaml
@@ -22,6 +22,7 @@ spec:
         type: aws
       controlplane:
         count: 3
+        k8sversion: "1.16.1"
 ---
 apiVersion: cluster.k8s.io/v1alpha1
 kind: Machine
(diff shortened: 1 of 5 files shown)


hack/test/manifests/gcp-cluster.yaml
```diff
@@ -23,6 +23,7 @@ spec:
         type: gce
       controlplane:
         count: 3
+      kubernetesVersion: "v1.16.1"
```
andrewrynhard: I think this needs to drop the `v`

andrewrynhard: Is this the version that get's passed to the Talos config?

rsmitty: yes. so I suspect you're right. no `v`.

hack/test/manifests/aws-cluster.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         type: aws
       controlplane:
         count: 3
+        k8sversion: "1.16.1"
```
andrewrynhard: Something for later, but maybe we should align this with what we do in talos and just call it `version` since its already namespaces under `controlplane`

PR #1250, opened 2019-10-08
fix: ignore case in install platform check

### LINKED ISSUE #1249: baremetal installer fails platform match
The platform detection of `osctl install` fails for baremetal installs because the platform name "Metal" does not match the case of the platform argument "metal".  This results in the configuration being thrown away.

Ulexus:
Because `platform.Name()` returns the capitalized name but the baremetal
platform kernel commandline option expects the lowercase 'metal', we
ignore the case of the platform when doing the platform match checking
in `ostctl install`.

Fixes #1249

### DIFF
--- a/cmd/osctl/cmd/install_linux.go
+++ b/cmd/osctl/cmd/install_linux.go
@@ -8,6 +8,7 @@ package cmd
 import (
 	"log"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


bradbeam:
I'm torn on this approach or updating all of the platforms under internal/pkg/platform/x/x.go to return back a lowercase name.

Ulexus:
Agreed; this was just the least-disruptive pathway.

andrewrynhard:
> Agreed; this was just the least-disruptive pathway.

Agree, lets stop the bleeding, and then followup with renames in the packages themselves.

PR #1252, opened 2019-10-08
fix: create etcd data directory

andrewrynhard:
This ensures that the data directory for etcd is exists before starting
etcd.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -49,6 +49,10 @@ func (e *Etcd) ID(config config.Configurator) string {
 
 // PreFunc implements the Service interface.
 func (e *Etcd) PreFunc(ctx context.Context, config config.Configurator) (err error) {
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -49,6 +49,10 @@ func (e *Etcd) ID(config config.Configurator) string {
 
 // PreFunc implements the Service interface.
 func (e *Etcd) PreFunc(ctx context.Context, config config.Configurator) (err error) {
+	if err = os.MkdirAll(constants.EtcdDataPath, 0755); err != nil {
+		return err
+	}
```
smira: not related to this change, but `fmt.Errorf` format should be `%w` for error wrapping in Go 1.13 (on line 66:70 below)

andrewrynhard: 👍 yeah I'd like to move to new error wrapping in 1.13 and remove `github.com/pkg/errors`.

PR #1253, opened 2019-10-08
fix: catch panics in boot go routine

andrewrynhard:
The builtin recover func is scoped to the current go routine, and since
our boot sequence is kicked off in its' own go routine, we were failing
to recover from panics.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -106,6 +106,8 @@ func main() {
 
 	// Start the boot sequence in a go routine so that we can list for events.
 	go func() {
+		defer recovery()
+		log.Println(err)
 		if err := seq.Boot(); err != nil {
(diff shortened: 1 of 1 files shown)


andrewrynhard:
This is a temporary workaround until we can revamp error handling.

internal/app/machined/main.go
```diff
@@ -106,6 +106,8 @@ func main() {
 
 	// Start the boot sequence in a go routine so that we can list for events.
 	go func() {
+		defer recovery()
+		log.Println(err)
```
smira: not sure what this is supposed to print? outer value of `err`? looks strange

andrewrynhard: ahh I think your right, I can fix

PR #1254, opened 2019-10-09
chore: Improve error messages if there is a network config overlap

### LINKED ISSUE #1251: Improve error messages if there's an networking config overlap
As you are experimenting with Talos, you might run into a situation where you already have a Docker cluster running, and if you run "cluster create" again, you get an error message like:

`Error response from daemon: Pool overlaps with other one on this address space
`

It's not clear to a new user what this means, so we should consider adding some context. Maybe something like:

`Error response from daemon: Pool overlaps with other one on this address space. A cluster might already exist, run "osctl cluster destroy" to permanently delete the existing cluster, and try again.
`

u5surf:
Fixes #1251 
It can be cleared the next steps for a new user.

Signed-off-by: u5surf <u5.horie@gmail.com>

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -117,7 +117,7 @@ func create() (err error) {
 	fmt.Println("creating network", clusterName)
 
 	if _, err = createNetwork(cli); err != nil {
-		return err
+		return errors.Wrap(err, " A cluster might already exist, run \"osctl cluster destroy\" to permanently delete the existing cluster, and try again.")
 	}
 
 	// Create the master nodes.


andrewrynhard:
@u5surf looks like rebase is needed.

cmd/osctl/cmd/cluster.go
```diff
@@ -117,7 +117,7 @@ func create() (err error) {
 	fmt.Println("creating network", clusterName)
 
 	if _, err = createNetwork(cli); err != nil {
-		return err
+		return errors.Wrap(err, ". A cluster might already exist, run \"osctl cluster destroy\" to permanently delete the existing cluster, and try again.")
```
andrewrynhard: One nit: there is and extra `.` at the beginning of the error message. `errors.Wrap` will automatically append a `:` to the `err`.

tgerla:
Thank you very much for the patch, @u5surf!

PR #1266, opened 2019-10-10
fix: update platform task to set hostname and cert SANs

andrewrynhard:
This adds a setter for the certificate SANs and sets the hostname based
on info from the platform.

### DIFF
--- a/internal/app/machined/internal/phase/platform/platform.go
+++ b/internal/app/machined/internal/phase/platform/platform.go
@@ -33,26 +33,27 @@ func (task *Platform) runtime(args *phase.RuntimeArgs) (err error) {
 		return err
 	}
 
-	_, err = args.Platform().Hostname()
+	hostname, err := args.Platform().Hostname()
 	if err != nil {
 		return err
 	}
 
-	// if hostname != nil {
-	// 	data.Networking.OS.Hostname = string(hostname)
-	// }
+	if hostname != nil {
+		args.Config().Machine().Network().SetHostname(string(hostname))
+	}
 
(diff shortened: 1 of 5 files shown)


internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -33,26 +33,27 @@ func (task *Platform) runtime(args *phase.RuntimeArgs) (err error) {
...
+	if hostname != nil {
+		args.Config().Machine().Network().SetHostname(string(hostname))
+	}
+
+	addrs, err := args.Platform().ExternalIPs()
+	if err != nil {
+		return err
+	}
+
+	sans := make([]string, len(addrs))
```
bradbeam: Does this need to be make([]string,0,len(addrs)) ?

internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -33,26 +33,27 @@ func (task *Platform) runtime(args *phase.RuntimeArgs) (err error) {
...
-	// if err != nil {
-	// 	return err
-	// }
-
-	// // And add them to our cert SANs for trustd
-	// for _, addr := range addrs {
-	// 	data.Services.Trustd.CertSANs = append(data.Services.Trustd.CertSANs, addr.String())
-	// }
+	if hostname != nil {
+		args.Config().Machine().Network().SetHostname(string(hostname))
```
bradbeam: Does that sethostname method exist already?

andrewrynhard: It does.

PR #1271, opened 2019-10-10
fix: Make updating cert sans an append operation

bradbeam:
This updates any discovered CertSANs to be appended to the list provided by userdata.

### DIFF
--- a/pkg/config/types/v1alpha1/machine_config.go
+++ b/pkg/config/types/v1alpha1/machine_config.go
@@ -107,7 +107,7 @@ func (m *MachineConfig) CertSANs() []string {
 
 // SetCertSANs implements the Configurator interface.
 func (m *MachineConfig) SetCertSANs(sans []string) {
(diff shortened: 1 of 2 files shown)


andrewrynhard:
@bradbeam Needs a rebase. Also, can we amend the commit and call it a fix?

pkg/config/types/v1alpha1/machine_config.go
```diff
@@ -107,7 +107,7 @@ func (m *MachineConfig) CertSANs() []string {
 
 // SetCertSANs implements the Configurator interface.
 func (m *MachineConfig) SetCertSANs(sans []string) {
-	m.MachineCertSANs = sans
+	m.MachineCertSANs = append(m.MachineCertSANs, sans...)
```
andrewrynhard: @bradbeam on second thought, I think we need to ensure that `m.MachineCertSANs` is not `nil`, and create an empty slice if it is.

PR #1276, opened 2019-10-11
refactor: add helper func to create cert and key

andrewrynhard:
It is common for us to create a certificate and key using a root CA.
This adds a helper function to make this process easier and
standardized. For now, the function is only used when generating
configs, but there are a number of places we should update in follow
ups.

### DIFF
--- a/pkg/config/types/v1alpha1/generate/generate.go
+++ b/pkg/config/types/v1alpha1/generate/generate.go
@@ -32,10 +32,48 @@ const DefaultIPv6PodNet = "fc00:db8:10::/56"
 // DefaultIPv6ServiceNet is the network to be used for kubernetes Services when using IPv6-based master nodes
 const DefaultIPv6ServiceNet = "fc00:db8:20::/112"
 
-// CertStrings holds the string representation of a certificate and key.
-type CertStrings struct {
-	Crt string
-	Key string
+// Type represents a config type.
+type Type int
+
+const (
+	// TypeInit indicates a config type should correspond to the kubeadm
+	// InitConfiguration type.
+	TypeInit Type = iota
+	// TypeControlPlane indicates a config type should correspond to the
+	// kubeadm JoinConfiguration type that has the ControlPlane field
+	// defined.
+	TypeControlPlane
+	// TypeJoin indicates a config type should correspond to the kubeadm
+	// JoinConfiguration type.
+	TypeJoin
+)
+
+// Sring returns the string representation of Type.
+func (t Type) String() string {
+	return [...]string{"Init", "ControlPlane", "Join"}[t]
+}
+
+// Config returns the talos config for a given node type.
+// nolint: gocyclo
+func Config(t Type, in *Input) (s string, err error) {
+	switch t {
+	case TypeInit:
+		if s, err = initUd(in); err != nil {
+			return "", err
+		}
+	case TypeControlPlane:
+		if s, err = controlPlaneUd(in); err != nil {
+			return "", err
+		}
+	case TypeJoin:
+		if s, err = workerUd(in); err != nil {
+			return "", err
+		}
+	default:
+		return "", errors.New("failed to determine config type to generate")
+	}
+
+	return s, nil
 }
 
 // Input holds info about certs, ips, and node type.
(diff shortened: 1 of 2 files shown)


pkg/config/types/v1alpha1/generate/generate.go
```diff
@@ -33,10 +33,48 @@ const DefaultIPv6PodNet = "fc00:db8:10::/56"
 // DefaultIPv6ServiceNet is the network to be used for kubernetes Services when using IPv6-based master nodes
 const DefaultIPv6ServiceNet = "fc00:db8:20::/112"
 
-// CertStrings holds the string representation of a certificate and key.
-type CertStrings struct {
-	Crt string
-	Key string
+// Type represents a config type.
```
andrewrynhard: This is just a move of the exported types to the top of the file.

pkg/config/types/v1alpha1/generate/generate.go
```diff
@@ -131,98 +169,76 @@ type TrustdInfo struct {
 	Token string
 }
 
-func generateCertificateKey() (string, error) {
-	key, err := randBytes(32)
-	if err != nil {
-		return "", err
+func NewEtcdCA() (ca *x509.CertificateAuthority, err error) {
```
andrewrynhard: I know there is a lot of commonality between the CA funcs, but they are explicit on purpose as a way to contain all related logic in a single function.

pkg/config/types/v1alpha1/generate/generate.go
```diff
@@ -131,98 +169,76 @@ type TrustdInfo struct {
...
+		x509.RSA(false),
+		x509.Organization("talos"),
+		x509.NotAfter(time.Now().Add(87600 * time.Hour)),
 	}
 
-	return string(token), err
+	return x509.NewSelfSignedCertificateAuthority(opts...)
 }
 
-// genToken will generate a token of the format abc.123 (like kubeadm/trustd), where the length of the first string (before the dot)
```
andrewrynhard: It is better to place unexpected functions at the bottom of the file.

pkg/config/types/v1alpha1/generate/generate.go
```diff
@@ -388,46 +329,82 @@ func NewInput(clustername string, masterIPs []string, kubernetesVersion string)
...
+
+	tokenTemp[1], err = randBytes(lenSecond)
+	if err != nil {
+		return "", err
+	}
+
+	return tokenTemp[0] + "." + tokenTemp[1], nil
+}
+
+func isIPv6(addrs ...string) bool {
```
andrewrynhard: This should be moved to our net package eventually

pkg/crypto/x509/x509.go
```diff
@@ -538,6 +539,78 @@ func (p *PEMEncodedCertificateAndKey) MarshalYAML() (interface{}, error) {
 	return aux, nil
 }
 
+// NewCertficateAndKey generates a new key and certificate signed by a CA.
+func NewCertficateAndKey(crt *x509.Certificate, key interface{}, setters ...Option) (p *PEMEncodedCertificateAndKey, err error) {
```
andrewrynhard: This is the motivation for the PR.

pkg/config/types/v1alpha1/generate/generate.go
smira: I hope these are `unexported` :)

pkg/crypto/x509/x509.go
```diff
@@ -538,6 +539,80 @@ func (p *PEMEncodedCertificateAndKey) MarshalYAML() (interface{}, error) {
...
+	} else {
+		priv, err = x509.ParseECPrivateKey(block.Bytes)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	csr, err := NewCertificateSigningRequest(priv, setters...)
+	if err != nil {
+		return nil, err
```
smira: errors in this function might benefit from any kind of wrapping (fmt.Errorf with %w or errors.Wrap), as sometimes it's hard to understand which part of the long flow returned an error

pkg/config/types/v1alpha1/generate/generate.go
andrewrynhard: lol yes

pkg/crypto/x509/x509.go
andrewrynhard: Agree 👍

PR #1277, opened 2019-10-11
fix: Use correct names for kubelet config

bradbeam:
With the change to bootkube, kubelet.conf has changed names and is now kubelet-kubeconfig.

### DIFF
--- a/internal/app/machined/pkg/system/services/osd.go
+++ b/internal/app/machined/pkg/system/services/osd.go
@@ -56,7 +56,7 @@ func (o *OSD) PostFunc(config config.Configurator) (err error) {
 // Condition implements the Service interface.
 func (o *OSD) Condition(config config.Configurator) conditions.Condition {
 	if config.Machine().Type() == machine.Worker {
-		return conditions.WaitForFileToExist("/etc/kubernetes/kubelet.conf")
+		return conditions.WaitForFileToExist(constants.KubeletKubeconfig)
 	}
 
 	return nil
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/system/services/osd.go
```diff
@@ -56,7 +56,7 @@ func (o *OSD) PostFunc(config config.Configurator) (err error) {
 // Condition implements the Service interface.
 func (o *OSD) Condition(config config.Configurator) conditions.Condition {
 	if config.Machine().Type() == machine.Worker {
-		return conditions.WaitForFileToExist("/etc/kubernetes/kubelet.conf")
+		return conditions.WaitForFileToExist("/etc/kubernetes/kubeconfig-kubelet")
```
andrewrynhard: I think we have this in `constants.go`, and if not, we should :)

internal/app/machined/pkg/system/services/osd.go
```diff
@@ -109,6 +109,8 @@ func (o *OSD) Runner(config config.Configurator) (runner.Runner, error) {
 	for key, val := range config.Machine().Env() {
 		env = append(env, fmt.Sprintf("%s=%s", key, val))
 	}
+	env = append(env, fmt.Sprintf("%s=%s", "GRPC_GO_LOG_VERBOSITY_LEVEL", "99"))
```
andrewrynhard: Cruft from previous PR?

bradbeam: yeah

PR #1278, opened 2019-10-11
feat: add Runtime interface

andrewrynhard:
Instead of passing around a struct, it is better if we pass around an
interface that describes the behavior we want. The Runtime interface
provides a common place to describe runtime specific parameters. This
initial implementation offers the runtime mode, the platform specifics,
and the config.


internal/app/machined/internal/phase/phase.go
```diff
@@ -52,22 +52,27 @@ func NewRunner(config config.Configurator) (*Runner, error) {
...
 
-	mode := platform.Mode()
-	switch mode {
-	case runtime.Metal:
-		fallthrough
-	case runtime.Cloud:
+	var mode runtime.Mode
+
+	switch platform.Name() {
+	case "Container":
```
andrewrynhard: Trying to find a better way to handle this.

andrewrynhard:
Follow up to this will be to allow for runtime errors so that we can avoid panics.

internal/pkg/runtime/runtime.go
```diff
@@ -22,3 +27,41 @@ const (
...
+func NewRuntime(m Mode, p platform.Platform, c config.Configurator) Runtime {
+	return &DefaultRuntime{
+		m: m,
+		p: p,
+		c: c,
+	}
+}
+
+// DefaultRuntime implements the Runtime interface.
+type DefaultRuntime struct {
```
bradbeam: feel like this is another instance of naming is hard. `Default` doesn't seem like the right fit, but at the same time i don't think these implementation details leak through at all, so not sure it really matters much.

internal/app/machined/internal/phase/phase.go
bradbeam: Why move it from `platform.Mode()`?

internal/pkg/runtime/runtime.go
andrewrynhard: Yep. Not sure what else to call it.

internal/app/machined/internal/phase/phase.go
andrewrynhard: Import cycle issue. Still chewing on this to find a better way. But the issue is that `runtime` imports platform, and platform imports `runtime`.

andrewrynhard:
This is the first PR in a series aimed at standardizing our boot process.

PR #1285, opened 2019-10-11
feat: use the unified pkgs repo artifacts

andrewrynhard:
This moves to using a single revision of pkgs. It includes a few
changes:

- kernel with KVM host support
- containerd v1.3.0

This change brings in a kernel with host KVM support. This will allow us
to use VMs within Talos for things like integrations tests. This also
allows users to do things with KVM as they see fit.

Signed-off-by: Andrew Rynhard <andrew@andrewrynhard.com>


andrewrynhard:
looks like we need to build some more packages since we don't publish `base` anymore:

```
COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libblkid.* /rootfs/lib
COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libuuid.* /rootfs/lib
COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libkmod.* /rootfs/lib
```

hack/containerd.toml
```diff
@@ -1,77 +1 @@
-root = "/var/lib/containerd"
-state = "/run/containerd"
-oom_score = 0
 subreaper = true
```
andrewrynhard: This is the only option we are interested in. To avoid having to keep track with the full config between any two versions I think we should only specify the things we are interested in. This PR brings in Containerd 1.3.0.

internal/pkg/containers/cri/cri_test.go
```diff
@@ -197,14 +197,14 @@ func (suite *CRISuite) TestPods() {
 	suite.Assert().Equal(pods[0].Name, pods[0].Containers[0].Display)
 	suite.Assert().Equal(pods[0].Name, pods[0].Containers[0].Name)
 	suite.Assert().Equal("SANDBOX_READY", pods[0].Containers[0].Status)
-	suite.Assert().Equal(pauseImageDigest, pods[0].Containers[0].Digest)
-	suite.Assert().Equal(pauseImage, pods[0].Containers[0].Image)
+	// suite.Assert().Equal(pauseImageDigest, pods[0].Containers[0].Digest)
```
andrewrynhard: I dug into this and can talk about it at our next meeting. I don't think this should block moving to container 1.3.

Dockerfile
```diff
@@ -201,30 +201,30 @@ COPY --from=osctl-darwin-build /osctl-darwin-amd64 /osctl-darwin-amd64
...
-COPY --from=docker.io/autonomy/musl:9bc7430 / /rootfs
-COPY --from=docker.io/autonomy/runc:c79f79d / /rootfs
-COPY --from=docker.io/autonomy/socat:032c783 / /rootfs
-COPY --from=docker.io/autonomy/syslinux:85e1f9c / /rootfs
-COPY --from=docker.io/autonomy/xfsprogs:5e50579 / /rootfs
-COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libblkid.* /rootfs/lib
-COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libuuid.* /rootfs/lib
-COPY --from=docker.io/autonomy/base:f9a4941 /toolchain/lib/libkmod.* /rootfs/lib
-COPY --from=docker.io/autonomy/kernel:66c8150 /lib/modules /rootfs/lib/modules
+COPY --from=docker.io/autonomy/fhs:1e2003e / /rootfs
```
bradbeam: Should we put this as an arg in the makefile like we do with the tools version? Unsure with the new pkgs thing if it's to be expected that all the components share the same sha.

internal/pkg/containers/cri/cri_test.go
```diff
@@ -27,10 +27,10 @@ import (
 )
 
 const (
-	busyboxImage       = "docker.io/library/busybox:1.30.1"
-	busyboxImageDigest = "sha256:64f5d945efcc0f39ab11b3cd4ba403cc9fefe1fa3613123ca016cf3708e8cafb"
-	pauseImage         = "k8s.gcr.io/pause:3.1"
-	pauseImageDigest   = "sha256:da86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e"
+	busyboxImage = "docker.io/library/busybox:1.30.1"
```
bradbeam: I'm not sure how much it matters here, but i was thinking through this the other day. Would it be mo better to use scratch as our base image for comparison?  Not sure if that'd have any speed impact.

internal/pkg/containers/cri/cri_test.go
bradbeam: Want to log an issue to follow up on the investigation?

Dockerfile
andrewrynhard: Unfortunately docker doesn't support using a var in `COPY --from`.

internal/pkg/containers/cri/cri_test.go
andrewrynhard: Negligible. This downloads in seconds.

internal/pkg/containers/cri/cri_test.go
andrewrynhard: Still looking into it and if I can't fix this week then for sure.

smira: Is it changing every time or just different?

hack/containerd.toml
smira: btw, in theory we don't need it anymore as we have reaper in machined

internal/pkg/containers/cri/cri_test.go
andrewrynhard: It is changing every time, and cycles through 2 or 3 different strings.

hack/containerd.toml
andrewrynhard: oh ... I wonder if we should just use our reaper? I'm good with that if you are.

internal/pkg/containers/cri/cri_test.go
smira: I think we could certainly remove these checks, they're more for completeness

but I wonder if CRI inspector doesn't show the containers as it should now?

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -168,7 +168,7 @@ func (k *Kubelet) Runner(config config.Configurator) (runner.Runner, error) {
 	mounts := []specs.Mount{
 		{Type: "bind", Destination: "/dev", Source: "/dev", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "sysfs", Destination: "/sys", Source: "/sys", Options: []string{"bind", "ro"}},
-		{Type: "cgroup", Destination: "/sys/fs/cgroup", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/sys/fs/cgroup", Source: "/sys/fs/cgroup", Options: []string{"rbind", "rshared", "rw"}},
```
andrewrynhard: Required now, as the `kubelet` will fail complaining about missing `cpu` cgroup otherwise.

hack/containerd.toml
smira: we should give it a try :)

PR #1292, opened 2019-10-15
chore: expand QEMU script to support HA setup

andrewrynhard:
This adds support for creating more than one node using the qemu-boot.sh
script.

### DIFF
--- a/docs/content/guides/kvm.md
+++ b/docs/content/guides/kvm.md
@@ -7,6 +7,19 @@ menu:
     parent: 'guides'
 ---
 
+## Generating the Configuration ISO
+
+```bash
+genisoimage  -output config.iso -volid cidata -joliet -rock <config>
+```
+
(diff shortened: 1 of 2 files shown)


pkg/constants/constants.go
```diff
@@ -158,9 +158,8 @@ const (
 	// ConfigPath is the path to the downloaded config.
 	ConfigPath = "/run/config.yaml"
 
-	// UserDataCIData is the volume label for NoCloud cloud-init.
-	// See https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html#datasource-nocloud.
-	UserDataCIData = "cidata"
+	// MetalConfigISOLabel is the volume label for ISO based configuration.
+	MetalConfigISOLabel = "metal-iso"
```
andrewrynhard: It's time to get away from the cloud-init nomenclature.

PR #1297, opened 2019-10-16
feat: allow specifcation of full url for endpoint

rsmitty:
This PR moves to using the full URL for endpoint instead of trying to
hardcode 6443 in various places like we were doing.

### DIFF
--- a/internal/app/machined/internal/phase/services/post_boot.go
+++ b/internal/app/machined/internal/phase/services/post_boot.go
@@ -6,7 +6,6 @@ package services
 
 import (
 	"fmt"
-	"net"
 	"os"
 	"time"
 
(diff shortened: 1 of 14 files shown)


pkg/config/types/v1alpha1/generate/generate.go
```diff
@@ -123,7 +124,15 @@ func (i *Input) GetControlPlaneEndpoint() string {
 // GetAPIServerSANs returns the formatted list of Subject Alt Name addresses for the API Server
 func (i *Input) GetAPIServerSANs() []string {
 	list := []string{"127.0.0.1", "::1"}
-	list = append(list, i.ControlPlaneEndpoint)
+
+	endpointURL, err := url.Parse(i.ControlPlaneEndpoint)
```
andrewrynhard: Not necessarily for this PR, but I think it makes sense to make `ControlPlaneEndpoint` a `*net.URL`.

internal/app/machined/internal/phase/services/post_boot.go
```diff
@@ -35,9 +34,11 @@ func (task *LabelNodeAsMaster) standard(r runtime.Runtime) (err error) {
...
 	}
 
-	endpoint := net.ParseIP(r.Config().Cluster().Endpoint())
-
-	h, err := kubernetes.NewTemporaryClientFromPKI(r.Config().Cluster().CA().Crt, r.Config().Cluster().CA().Key, endpoint.String(), "6443")
+	h, err := kubernetes.NewTemporaryClientFromPKI(
+		r.Config().Cluster().CA().Crt,
+		r.Config().Cluster().CA().Key,
+		r.Config().Cluster().Endpoint().Hostname(),
+		r.Config().Cluster().Endpoint().Port())
```
andrewrynhard: nit:
```suggestion
		r.Config().Cluster().Endpoint().Port(),
	)
```

rsmitty: Fixed in amended commit

pkg/config/types/v1alpha1/generate/generate.go
rsmitty: totally agree. I'll drop a backlog ticket for this

PR #1323, opened 2019-10-21
feat: Add node metadata wrapper to machine api

bradbeam:
- Added common.proto to host NodeMetadata
- go_package names were fixed up so imports are generated with the proper
  package names
- fixed up build work (dockerfile) to prevent copying the previously
  generated go proto files. This fixes a bug where we could incorrectly
  copy the previously generated protobuf instead of a new one generated
  at an incorrect location/name/etc. (fixes bug introduced with #1321)


bradbeam:
Need to update osctl components to use updated proto messages.

cmd/osctl/pkg/client/client.go
```diff
@@ -293,13 +293,8 @@ func (c *Client) CopyOut(ctx context.Context, rootPath string) (io.Reader, <-cha
```
bradbeam: this cleans up some of the processing we were doing at the client layer.

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -52,7 +52,9 @@ func (r *Registrator) Register(s *grpc.Server) {
```
bradbeam: Reformats the output messages to be in the new message format.

cmd/osctl/cmd/root.go
```diff
@@ -55,8 +56,14 @@ func Execute() {
 	var globalCtxCancel context.CancelFunc
 
 	globalCtx, globalCtxCancel = context.WithCancel(context.Background())
+
 	defer globalCtxCancel()
 
+	// Update context with grpc metadata for proxy/relay requests
```
bradbeam: Will make another PR to clean up the os api bits that explicitly set the metadata in the `cmd/osctl/cmd/x.go` files so we can just make use of this.

Dockerfile
```diff
@@ -35,30 +35,29 @@ WORKDIR /src
...
-COPY ./api/machine ./proto
-RUN protoc -I./proto --go_out=plugins=grpc:proto proto/machine.proto
-WORKDIR /ntpd
-COPY ./api/time ./proto
-RUN protoc -I./proto --go_out=plugins=grpc:proto proto/time.proto
-WORKDIR /networkd
-COPY ./api/network ./proto
-RUN protoc -I./proto --go_out=plugins=grpc:proto proto/network.proto
-
+COPY ./api/common/common.proto /api/common/common.proto
```
bradbeam: explicitly copy only the proto definition over. this prevents us from masking failures here because the generated proto files would still exist.

Dockerfile
```diff
@@ -35,30 +35,29 @@ WORKDIR /src
...
+RUN protoc -I/api --go_out=plugins=grpc:/api/common /api/common/common.proto
+#RUN protoc -I/api --plugin=proxy --proxy_out=plugins=grpc+proxy:proto api/api.proto
+COPY ./api/os/os.proto /api/os/os.proto
+# TODO: switch up os proxy generation with apid
```
bradbeam: I'll remove these comments in the next PR

PR #1327, opened 2019-10-22
fix: ensure control plane endpoint is set

andrewrynhard:
We were mistakenly overwriting the control plane endpoint in the
`generate` command. This fixes that and adds a simple validation of the
endpoint field in the config. We should expand the validation to ensure
that a valid IP or DNS name have been provided.

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -144,8 +144,6 @@ func genV1Alpha1Config(args []string) {
 	}
 
 	input.AdditionalSubjectAltNames = additionalSANs
-	input.ControlPlaneEndpoint = canonicalControlplaneEndpoint
-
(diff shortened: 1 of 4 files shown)


cmd/osctl/cmd/config.go
```diff
@@ -144,8 +144,6 @@ func genV1Alpha1Config(args []string) {
 	}
 
 	input.AdditionalSubjectAltNames = additionalSANs
-	input.ControlPlaneEndpoint = canonicalControlplaneEndpoint
```
bradbeam: do we need to set `input.ControlPlaneEndpoint` to anything?

andrewrynhard: Gets set in the generation code for `v1alpha1`.

PR #1335, opened 2019-10-23
feat: detect gzipped machine configs

rsmitty:
This PR will add the ability for talos to detect if the machine config
that it downloads from the platform is a gzipped file. If so, it will
unzip it and overwrite the byte slice that gets written to disk.

### DIFF
--- a/internal/app/machined/internal/phase/config/config.go
+++ b/internal/app/machined/internal/phase/config/config.go
@@ -5,7 +5,11 @@
 package config
 
 import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
 	"io/ioutil"
+	"net/http"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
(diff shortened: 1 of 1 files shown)


internal/app/machined/internal/phase/config/config.go
```diff
@@ -32,5 +35,20 @@ func (task *Task) standard(r runtime.Runtime) (err error) {
 		return err
 	}
 
+	// Detect if config is a gzip archive and unzip it if so
+	contentType := http.DetectContentType(b)
+	if contentType == "application/x-gzip" {
+		gzipReader, err := gzip.NewReader(bytes.NewBuffer(b))
```
smira: `bytes.NewReader` might be more appropriate here (instead of `bytes.NewBuffer`) as it's a bit more lightweight

rsmitty:
@smira fixed nits.

PR #1339, opened 2019-10-24
feat: Add APId

bradbeam:
This PR introduces APId. This service replaces the frontend functionality
previously provided by OSD. The main driver for this is two fold:

1. Create a single purpose application to expose the talos api

2. Make use of code generation to DRY api changes


Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN curl -sfL https://github.com/uber/prototool/releases/download/v1.8.0/prototo
 
 FROM scratch AS build
 COPY --from=tools / /
-COPY --from=autonomy/protoc-gen-proxy:a87401e /protoc-gen-proxy /toolchain/bin/protoc-gen-proxy
+COPY --from=bradbeam/protoc-gen-proxy:24088be /protoc-gen-proxy /toolchain/bin/protoc-gen-proxy
```
bradbeam: Depends on https://github.com/talos-systems/protoc-gen-proxy/pull/5

api/api/api.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package api;
+
+option go_package = "github.com/talos-systems/talos/api/api";
+
+import public "os/os.proto";
```
bradbeam: We use public here so this can be our 'meta' package

api/machine/machine.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package machineapi;
+package machine;
```
bradbeam: the name changes for the proto files are to align a bit better as well as help with identifying the socket name we need to use.

bradbeam:
Early preview. Still need to test and the like, but it compiles. Should help show the direction I'm going with this.

api/api/api.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
```
andrewrynhard: Let's move this up one directory. Importing `github.com/talos-systems/talos/api/api` can be cleaned up to `github.com/talos-systems/talos/api`

internal/app/apid/main.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	protoProxy := api.NewApiProxy(interceptorProvider)
+
+	err = factory.ListenAndServe(
+		&api.Registrator{
+			MachineClient: machineClient,
+			OSClient:      osClient,
+		},
+		factory.Port(constants.OsdPort),
```
andrewrynhard: I'm a bit confused about this. Is the idea that apid will run on every node?

pkg/constants/constants.go
```diff
@@ -194,12 +194,18 @@ const (
 	// InitSocketPath is the path to file socket of machine API.
 	InitSocketPath = SystemRunPath + "/init/init.sock"
 
+	// MachineSocketPath is the path to file socket of machine API.
+	MachineSocketPath = SystemRunPath + "/init/init.sock"
```
andrewrynhard: We should just rename `InitSocketPath`:

```
MachineSocketPath = SystemRunPath + "/machine/machine.sock"
```

andrewrynhard: We should just rename `InitSocketPath`:

```
MachineSocketPath = SystemRunPath + "/machine/machine.sock"
```

internal/app/apid/main.go
bradbeam: Yeah that's always been the plan -- to replace the frontend with this. I think we've talked about it a bit before and i don't think it's unreasonable to add in an option to control the proxy functionality ( we _should_ just be able to drop the unary interceptor portion and be back to normal ), but honestly I'm not sure I see any benefit to disabling that.

andrewrynhard: I have concerns about the long term effects of this, particularly when we get to implementing RBAC. I think we should run APId on master nodes only since there are already sensitive files on those machines and it’s best practice to not allow workloads on those machines. I’d like to see worker nodes have zero ability to reach out to any other node and use the API.

andrewrynhard: Let's chat about this at the community meeting today. I think this is worth a deeper discussion.

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -0,0 +1,155 @@
...
+		return conditions.WaitForFileToExist(constants.KubeletKubeconfig)
+	}
+
+	return nil
+}
+
+// DependsOn implements the Service interface.
+func (o *APID) DependsOn(config runtime.Configurator) []string {
+	if config.Machine().Type() == machine.Worker {
+		return []string{"system-containerd", "containerd", "kubelet"}
```
andrewrynhard: I don't think we need the dependency on `kubelet`.

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -0,0 +1,155 @@
...
+			"/apid",
+			"--config=" + constants.ConfigPath,
+			"--endpoints=" + strings.Join(endpoints, ","),
+		},
+	}
+
+	// Set the mounts.
+	mounts := []specs.Mount{
+		{Type: "bind", Destination: "/etc/ssl", Source: "/etc/ssl", Options: []string{"bind", "ro"}},
+		// config.yaml
```
andrewrynhard: Comment is kinda redundant here. The name of the constant is `ConfigPath`

internal/app/machined/pkg/system/services/osd.go
```diff
@@ -141,13 +116,7 @@ func (o *OSD) Runner(config runtime.Configurator) (runner.Runner, error) {
 // HealthFunc implements the HealthcheckedService interface
 func (o *OSD) HealthFunc(runtime.Configurator) health.Check {
 	return func(ctx context.Context) error {
-		var d net.Dialer
```
andrewrynhard: We should change this to use the local socket. I would hate to lose the health check.

internal/app/apid/main.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		log.Fatalf("machine client: %v", err)
+	}
+
+	osClient, err := api.NewLocalOSClient()
+	if err != nil {
+		log.Fatalf("networkd client: %v", err)
+	}
+
+	interceptorProvider, err := createProvider()
```
andrewrynhard: nit: `interceptorProvider` is a bad name IMO. I thought it was a new implementation. In OSD we just called it `provider`.

internal/app/machined/pkg/system/services/apid.go
bradbeam: osd depended on kubelet to make use of admin.conf to query for master nodes. i suppose we could drop the dependency for apid and not have os api available until osd comes up. i'll drop it.

internal/app/apid/main.go
bradbeam: we have a provider for the grpc server itself and then a secondary provider that we pass through to the unaryInterceptor ( `interceptorProvider` ). the provider we pass through to the interceptor is what gets used for auth on the proxied connections

Makefile
```diff
@@ -53,7 +53,7 @@ DOCKER_ARGS ?=
 
 TESTPKGS ?= ./...
 
-all: ci rootfs initramfs kernel osctl-linux osctl-darwin installer container
```
andrewrynhard: `osctl-darwin` needs to be added back.

api/api/api.proto
andrewrynhard: bump

internal/app/apid/main.go
andrewrynhard: Is there not a way to use one for both? We should be able to as we just use it to get the CA and get the certificate. I think I'd rather have this one daemon have one identity.

bradbeam: We _could_ use one for both, but it seemed worthwhile to split up the certificates used for apid as a service that clients interact with and certificates used for apid proxy/relay communication. ultimately the certificates are probably just about the same so i'm not sure if there is really any benefit to doing so. seemed more like the right approach.

andrewrynhard: Since they are essentially the same cert, and the idea is that the service should have a single "identity" I think we should go with one cert provider.

Makefile
andrewrynhard: @bradbeam I think this is the last thing to fix before 👍.

PR #1346, opened 2019-10-25
test: add integration test framework

smira:
This is just first steps and core foundation.

It can be used like:

```
make integration.test
osctl cluster create
build/integration.test -test.v
```

This should run the test against the Docker instance.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -336,6 +336,20 @@ ENV GO111MODULE on
 ARG TESTPKGS
 RUN --mount=type=cache,target=/root/.cache/go-build go test -v -count 1 -race ${TESTPKGS}
 
+# The integration-test target builds integration test binary.
+
+FROM base AS integration-test-build
+ARG SHA
+ARG TAG
+ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
+RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=amd64 go test -c \
+    -ldflags "-s -w -X ${VERSION_PKG}.Name=Client -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG}" \
+    -tags integration,integration_api \
+    ./internal/integration
+
+FROM scratch AS integration-test
+COPY --from=integration-test-build /src/integration.test /integration-test
+
 # The lint target performs linting on the source code.
 
 FROM base AS lint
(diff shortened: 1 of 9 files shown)


smira:
Not a finished solution of course, and build tags need to be fixed with the license header, but I would love to hear feedback whether this is right direction or not

smira:
Pushed this on top of #1348, please review only 2nd commit

smira:
as PoC, now tests are running as part of `basic-integration`

Dockerfile
```diff
@@ -313,6 +313,20 @@ ENV GO111MODULE on
 ARG TESTPKGS
 RUN --mount=type=cache,target=/root/.cache/go-build go test -v -count 1 -race ${TESTPKGS}
 
+# The integration.test target builds integration test binary.
+
+FROM base AS integration.test-build
```
andrewrynhard: nit: can we use `-` instead of `.` to align with the convention?

Makefile
```diff
@@ -262,6 +262,14 @@ unit-tests-race: buildkitd
 		--opt build-arg:TESTPKGS=$(TESTPKGS) \
 		$(COMMON_ARGS)
 
+.PHONY: integration.test
+integration.test: buildkitd
+	@$(BINDIR)/buildctl --addr $(BUILDKIT_HOST) \
+		build \
+		--output type=local,dest=build \
```
andrewrynhard: If this is output to `build`, then it will make it into our release. I think we should add this too `./bin` (its ignored by git).

hack/test/basic-integration.sh
```diff
@@ -10,6 +10,7 @@ export TALOSCONFIG="${TMP}/talosconfig"
 export KUBECONFIG="${TMP}/kubeconfig"
 export TIMEOUT=300
 export OSCTL="${PWD}/build/osctl-linux-amd64"
+export INTEGRATIONTEST="${PWD}/build/integration.test"
```
andrewrynhard: nit: `-` instead of `.`

andrewrynhard:
Awesome. This is gonna save us a lot of time in fixing regressions going forward.

Dockerfile
smira: yep, sure, I followed the default naming for `go test -c`, but we can certainly make it all dashes

Makefile
smira: thanks for the hint!

smira: in the long run, I feel our integration test might be part of the release (to make sure we have some level of backward or forward compatibility)

andrewrynhard: Yeah good idea. In that case should we leave it?

PR #1380, opened 2019-10-31
docs: various layout and responsiveness fixes

tgerla:
- adjust ul margin to keep the bullets inside the content area
- fix a few docs page responsiveness problems on small screens
- adjust the layout of the logo relative to the docs sidebar
- clean up some vestigial CSS classes

### DIFF
--- a/docs/website/components/DocumentationDropdown.vue
+++ b/docs/website/components/DocumentationDropdown.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="dropdown inline-block">
-    <button class="font-semibold py-2 px-4 rounded inline-flex items-center">
+    <button class="font-semibold py-2 pl-4 rounded inline-flex items-center">
       <span class="mr-1">{{ $store.state.sidebar.version }}</span>
       <svg
         id="dropdown-caret"
--- a/docs/website/components/Header.vue
+++ b/docs/website/components/Header.vue
@@ -1,24 +1,26 @@
 <template>
-  <header id="header" class="flex flex-wrap">
-    <div class="flex items-center justify-start max-w-6xl mx-auto py-2">
-      <Logo></Logo>
-    </div>
-    <div class="flex items-center justify-end max-w-6xl mx-auto py-6">
-      <ul class="flex flex-wrap items-center justify-end">
-        <li class="link min-w-full md:min-w-0">
-          <a href="/docs/v0.3">
-            <span class="font-semibold mr-1">Documentation</span>
-          </a>
-        </li>
-        <li class="link min-w-full md:min-w-0">
-          <CommunityDropdown></CommunityDropdown>
-        </li>
-        <li class="link min-w-full md:min-w-0">
-          <a href="/faqs">
-            <span class="font-semibold mr-1">FAQs</span>
-          </a>
-        </li>
-      </ul>
+  <header id="header">
+    <div class="flex flex-wrap w-4/5 mx-auto">
+      <div class="flex py-2">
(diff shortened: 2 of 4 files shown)


docs/website/components/Header.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <header id="header" class="flex flex-wrap">
-    <div class="flex items-center justify-start max-w-6xl mx-auto py-2">
+  <header id="header" class="flex flex-wrap w-4/5 mx-auto">
+    <div class="flex py-2">
       <Logo></Logo>
     </div>
-    <div class="flex items-center justify-end max-w-6xl mx-auto py-6">
+    <div class="flex py-6 ml-auto">
```
andrewrynhard: Not sure which change did it, but seeing this:

Screen Shot 2019-10-31 at 10 23 30 AM

Notice the bottom border doesn't extend the full width of the screen.

tgerla: Oh yeah, good catch. Let me try to fix this.

docs/website/components/Content.vue
```diff
@@ -37,3 +37,9 @@ export default {
   }
 }
 </script>
+
+<style>
+article.c-rich-text ul {
+  @apply ml-6;
```
andrewrynhard: This is a pretty nasty hack that I'd rather not have to maintain. Not a blocker, but I'd like to dig into why we need to use `outside`. Using `inside` fixes this everywhere. I will follow up and try to rid us of this.

docs/website/components/DocumentationDropdown.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="dropdown inline-block">
-    <button class="font-semibold py-2 px-4 rounded inline-flex items-center">
+    <button class="font-semibold py-2 pl-4 rounded inline-flex items-center">
```
andrewrynhard: I'm not sure I understand this one. `px` will place the button evenly in the center.

docs/website/components/Sidebar.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="sidenav sticky pt-4 pb-4">
+  <div class="sidenav sticky pb-4">
```
andrewrynhard: I'm not sure I agree with this. I have a TODO to make it such that the bottom of the current doc's title is aligned vertically with the first element of the sidebar. It make the main content pop better IMHO.

andrewrynhard: I've seen this strategy in a number of sites, but I have been using rook's as a reference:
Screen Shot 2019-10-31 at 2 35 56 PM

docs/website/components/Content.vue
tgerla: I think your fix is in now, so I'll remove this.

docs/website/components/DocumentationDropdown.vue
tgerla: When I reworked the header structure to line up the logo with the menu, I couldn't get it fully right-aligned to line up with the rightmost menu item. This isn't a complete fix. I can go back and rework it.

docs/website/components/Sidebar.vue
tgerla: Good call, I'll revert.

docs/website/components/DocumentationDropdown.vue
tgerla: I'll redo this as a separate fix, if that's OK.

PR #1388, opened 2019-10-31
chore: remove RAW disk

andrewrynhard:
We need to remove this so that it is not published in a release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,6 +240,7 @@ image-vmware:
 		--privileged $(DOCKER_ARGS) \
 		autonomy/installer:$(TAG) \
 		ova
+	@rm -rf $(PWD)/build/talos.raw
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -240,6 +240,7 @@ image-vmware:
 		--privileged $(DOCKER_ARGS) \
 		autonomy/installer:$(TAG) \
 		ova
+	@rm -rf $(PWD)/build/talos.raw
```
bradbeam: Here or in entrypoint?

andrewrynhard: We do this same thing for GCP in the Makefile. So for the sake of consistency here. I started reworking the entrypoint which should clean up our makefile a bit, but this PR gets us able to release at least.

PR #1389, opened 2019-11-01
feat: add installer command to installer container

andrewrynhard:
This replaces the entrypoint.sh shell script with a go binary.


hack/drone.jsonnet
```diff
@@ -183,7 +183,11 @@ local talos = Step("talos", depends_on=[rootfs]);
 local golint = Step("lint-go", depends_on=[setup_ci]);
 local protobuflint = Step("lint-protobuf", depends_on=[setup_ci]);
 local markdownlint = Step("lint-markdown", depends_on=[setup_ci]);
-local image_test = Step("image-test", depends_on=[installer]);
+local image_aws = Step("image-aws", depends_on=[installer]);
+local image_azure = Step("image-azure", depends_on=[installer]);
+local image_digital_ocean = Step("image-digital-ocean", depends_on=[installer]);
+local image_gcp = Step("image-gcp", depends_on=[installer]);
+local image_vmware = Step("image-vmware", depends_on=[installer]);
```
andrewrynhard: This is the best way to ensure that the image command doesn't break.

internal/pkg/runtime/platform/packet/packet.go
```diff
@@ -39,7 +40,12 @@ func (p *Packet) Hostname() (hostname []byte, err error) {
...
 
-// ExternalIPs provides any external addresses assigned to the instance
+// ExternalIPs implements the runtime.Platform interface.
 func (p *Packet) ExternalIPs() (addrs []net.IP, err error) {
 	return addrs, err
 }
+
+// KernelArgs implements the runtime.Platform interface.
+func (p *Packet) KernelArgs() kernel.Parameters {
+	return nil
```
andrewrynhard: We should probably return something here.

Dockerfile
```diff
@@ -285,16 +293,14 @@ RUN apk add --no-cache --update \
     syslinux \
     util-linux \
     xfsprogs
-COPY hack/installer/entrypoint.sh /bin/entrypoint.sh
-COPY hack/installer/template.ovf /template.ovf
 COPY --from=kernel /vmlinuz /usr/install/vmlinuz
 COPY --from=rootfs /usr/lib/syslinux/ /usr/lib/syslinux
 COPY --from=initramfs /initramfs.xz /usr/install/initramfs.xz
-COPY --from=osctl-linux-build /osctl-linux-amd64 /bin/osctl
+COPY --from=installer-build /installer /bin/installer
```
andrewrynhard: Maybe we should create a symlink to this at `/bin/osctl` for backwards compat.

Makefile
```diff
@@ -145,48 +145,8 @@ osctl-%:
...
-	@rm -rf $(PWD)/$(ARTIFACTS)/disk.raw
-
-.PHONY: image-vmware
-image-vmware: ## Builds a VM image suitable for use in VMware and outputs it to the artifact directory.
-	@$(MAKE) image-build-vmware CONFIG_SOURCE=guestinfo EXTRA_IMAGE_ARGS="-e console=tty0 -e earlyprintk=ttyS0,115200"
-	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out \
-		--privileged \
-		autonomy/installer:$(TAG) \
-		ova -n vmware
-	@rm -rf $(PWD)/$(ARTIFACTS)/talos.raw
```
andrewrynhard: All this logic is now contained in the `installer` binary.

cmd/installer/cmd/image.go
```diff
@@ -0,0 +1,391 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(imageCmd)
+
+	imageCmd.Flags().StringVar(&configArg, "config", "none", "The value of "+constants.KernelParamConfig)
+	imageCmd.Flags().BoolVar(&bootloaderArg, "bootloader", true, "Install a booloader to the specified disk")
+	imageCmd.Flags().StringVar(&diskArg, "disk", "", "The path to the disk to install to")
+	imageCmd.Flags().StringVar(&platformArg, "platform", "", "The value of "+constants.KernelParamPlatform)
+	imageCmd.Flags().StringArrayVar(&extraKernelArgs, "extra-kernel-arg", []string{}, "Extra argument to pass to the kernel")
```
andrewrynhard: This should probably be persistent flags on `root` command since `image` and `install` will both use them.

andrewrynhard: On second thought, maybe not... only `platform` makes sense here.

cmd/installer/cmd/install.go
```diff
@@ -0,0 +1,107 @@
...
+	rootCmd.AddCommand(installCmd)
+}
+
+func runInstallCmd() (err error) {
+	var config runtime.Configurator
+
+	platform, err := platform.CurrentPlatform()
+	if err == nil {
+		if !strings.EqualFold(platform.Name(), platformArg) {
+			log.Printf("platform mismatch (%s != %s)", platform.Name(), platformArg)
```
andrewrynhard: Maybe this should be an error?

cmd/installer/pkg/losetup.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/talos-systems/talos/pkg/cmd"
+)
+
+// Loattach attaches a loopback device by inoking the `losetup` command.
+func Loattach(img string) (dev string, err error) {
+	if dev, err = cmd.Run("losetup", "--find", "--partscan", "--nooverlap", "--show", img); err != nil {
+		return "", fmt.Errorf("failed to setup loopback device: %w", err)
+	}
+
+	return strings.TrimSuffix(dev, "\n"), nil
```
andrewrynhard: should `cmd.Run` do the trimming?

cmd/installer/cmd/image.go
```diff
@@ -0,0 +1,391 @@
...
+	"github.com/talos-systems/talos/pkg/version"
+)
+
+var (
+	configArg       string
+	bootloaderArg   bool
+	diskArg         string
+	platformArg     string
+	outputArg       string
+	extraKernelArgs []string
```
andrewrynhard: We should really start using a struct that represents a set of options.

internal/pkg/install/install.go
```diff
@@ -52,7 +51,7 @@ func Install(r runtime.Runtime) error {
 	}
 
 	args := []string{
-		"/bin/osctl",
+		"/bin/installer",
```
andrewrynhard: a symlink at `/bin/osctl` makes sure that we don't break anything.

PR #1391, opened 2019-11-01
feat: create cluster with default PSP

### LINKED ISSUE #1005: Create a default PSP
We have been running into issues with kube-proxy taking forever to start. Come to find out, this is due to the fact that we wait for all nodes to checkin before applying a PSP. If we apply the PSP as part of the bootstrap process of the cluster, this will fix this issue. It also provides a better user experience.

andrewrynhard:
This adds a default PSP that is applied upon bootstrapping the cluster.

Closes #1005.


andrewrynhard:
For now the PSP is wide open, but I'd like to lock it down for the v0.3 release.

PR #1392, opened 2019-11-01
fix: Avoid running bootkube on reboots

andrewrynhard:
Since bootkube should only be ran once, we need a way to determine if it
has already been ran. This makes use of etcd to store a key-value pair
indicating that the cluster has been initialized.

### DIFF
--- a/cmd/osctl/cmd/cluster/pkg/node/node.go
+++ b/cmd/osctl/cmd/cluster/pkg/node/node.go
@@ -17,6 +17,7 @@ import (
 	"github.com/docker/go-connections/nat"
 
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1/generate"
+	"github.com/talos-systems/talos/pkg/constants"
 )
 
 // Request represents the set of options available for configuring a node.
@@ -121,7 +122,7 @@ func NewNode(clusterName string, req *Request) (err error) {
 
 		fallthrough
 	case generate.TypeControlPlane:
-		containerConfig.Volumes["/var/lib/etcd"] = struct{}{}
+		containerConfig.Volumes[constants.EtcdDataPath] = struct{}{}
 
 		if req.IP == nil {
 			return errors.New("an IP address must be provided when creating a master node")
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -96,11 +103,93 @@ func (b *Bootkube) ID(config runtime.Configurator) string {
...
+	}
+
+	cli, err := clientv3.New(clientv3.Config{
+		Endpoints:   []string{"127.0.0.1:2379"},
+		DialTimeout: 5 * time.Second,
+		TLS:         tlsConfig,
+	})
+	if err != nil {
+		return err
+	}
```
andrewrynhard: This should all be part of an etcd package.

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -96,11 +103,93 @@ func (b *Bootkube) ID(config runtime.Configurator) string {
...
```
andrewrynhard: This should all be part of an etcd package.

internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -96,11 +103,93 @@ func (b *Bootkube) ID(config runtime.Configurator) string {
...
+
+	retry.Exponential(15*time.Second, retry.WithUnits(50*time.Millisecond), retry.WithJitter(25*time.Millisecond)).Retry(func() error {
+		var resp *clientv3.GetResponse
+		var err error
+
+		ctx := clientv3.WithRequireLeader(context.Background())
+		if resp, err = cli.Get(ctx, "initialized"); err != nil {
```
andrewrynhard: "initialized" should be in `constants.go`

cmd/osctl/cmd/cluster/pkg/node/node.go
```diff
@@ -121,8 +121,6 @@ func NewNode(clusterName string, req *Request) (err error) {
 
 		fallthrough
 	case generate.TypeControlPlane:
-		containerConfig.Volumes["/var/lib/etcd"] = struct{}{}
```
andrewrynhard: We don't need this anymore.

andrewrynhard:
I need this PR in order to start work on upgrades, and I'd like to get that in before KubeCon so gonna merge. This is simple enough that if there are any concerns I can fix in another PR.

PR #1393, opened 2019-11-01
feat: Add support for setting container output to stdout

bradbeam:
This allows the config.Debug setting to control container output to allow better troubleshooting.

We should probably expose a `Debug` setting per app/service, but that can be done later. 

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd.go
@@ -7,6 +7,7 @@ package containerd
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd.go
```diff
@@ -115,8 +116,17 @@ func (c *containerdRunner) Close() error {
...
+		err  error
+	)
+
 	// Create the task and start it.
-	task, err := c.container.NewTask(c.ctx, cio.LogFile(c.logPath()))
+	if c.debug {
+		task, err = c.container.NewTask(c.ctx, cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr)))
+	} else {
+		task, err = c.container.NewTask(c.ctx, cio.LogFile(c.logPath()))
+	}
```
andrewrynhard: This could be simplified a bit:

```
	var (
		task containerd.Task
		err  error
	)

	creator := cio.LogFile(c.logPath())
	if c.debug {
		creator = cio.NewCreator(cio.WithStreams(os.Stdin, os.Stdout, os.Stderr))
	}

	// Create the task and start it.
	task, err = c.container.NewTask(c.ctx, creator)
	if err != nil {
		return fmt.Errorf("failed to create task: %q: %w", c.args.ID, err)
	}
```

internal/app/machined/pkg/system/runner/containerd/containerd.go
```diff
@@ -115,8 +116,17 @@ func (c *containerdRunner) Close() error {
 func (c *containerdRunner) Run(eventSink events.Recorder) error {
```
andrewrynhard: nit: we can have a named return: `(err error)` instead.

bradbeam: we could, but I'm not sure there's any benefit to that since most all of the returns are wrapping the error ( `fmt.Errorf(xxx, err)` ) and not actually returning `err`

PR #1396, opened 2019-11-02
refactor: use retry package in ntpd

andrewrynhard:
This moves to using the retry package for retrying NTP queries. It also
adds some additional logging that is useful when NTP queries fail.

### DIFF
--- a/internal/app/ntpd/pkg/ntp/ntp.go
+++ b/internal/app/ntpd/pkg/ntp/ntp.go
@@ -7,12 +7,13 @@ package ntp
 import (
 	"fmt"
 	"log"
-	"math/rand"
 	"syscall"
 	"time"
 
 	"github.com/beevik/ntp"
 	"github.com/hashicorp/go-multierror"
+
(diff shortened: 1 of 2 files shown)


internal/app/ntpd/pkg/ntp/ntp.go
```diff
@@ -76,23 +77,26 @@ func (n *NTP) Daemon() (err error) {
 }
 
 // Query polls the ntp server and verifies a successful response.
-func (n *NTP) Query() (*ntp.Response, error) {
-	for i := 0; i < n.Retry; i++ {
-		resp, err := ntp.Query(n.Server)
+func (n *NTP) Query() (resp *ntp.Response, err error) {
+	err = retry.Exponential(1*time.Minute, retry.WithUnits(time.Second), retry.WithJitter(250*time.Millisecond)).Retry(func() error {
+		resp, err = ntp.Query(n.Server)
```
bradbeam: We need to honor minpoll for the retries

PR #1400, opened 2019-11-02
refactor: unify service stop on upgrade

andrewrynhard:


cmd/osctl/cmd/install_linux.go
```diff
@@ -40,46 +37,20 @@ var installCmd = &cobra.Command{
...
-				if err != nil {
-					log.Fatal(err)
-				}
-				config, err = machineconfig.New(content)
-				if err != nil {
-					log.Fatal(err)
-				}
-				extraKernelArgs = append(extraKernelArgs, config.Machine().Install().ExtraKernelArgs()...)
-			}
-		}
```
andrewrynhard: We shouldn't try to be smart here. We can handle this logic in `machined` and run this command with the proper args instead.

internal/pkg/runtime/platform/aws/aws.go
```diff
@@ -117,7 +117,7 @@ func IsEC2() (b bool) {
 
 // Name implements the platform.Platform interface.
 func (a *AWS) Name() string {
-	return "AWS"
+	return "aws"
```
andrewrynhard: We've talked about making these consistent with the kernel args. I ran into issues with mismatches and figured now is as good a time as any to fix this.

PR #1424, opened 2019-11-04
refactor: Move kubeconfig to machined

bradbeam:
This moves the Kubeconfig api endpoint to machined and consolidates the
"read a file" code into machined. This also changes Kubeconfig to
use the CopyOut method which changes Kubeconfig to a streaming grpc call.


api/machine/machine.proto
```diff
@@ -14,16 +14,17 @@ import "common/common.proto";
 // The machine service definition.
 service Machine {
```
bradbeam: sorted methods

cmd/osctl/cmd/cp.go
```diff
@@ -81,59 +79,7 @@ captures ownership and permission bits.`,
 				}
 			}
```
bradbeam: moved this to a common func in `cmd/osctl/cmd/root.go`

cmd/osctl/cmd/kubeconfig.go
```diff
@@ -8,41 +8,55 @@ package cmd
...
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
 )
 
 // kubeconfigCmd represents the kubeconfig command
 var kubeconfigCmd = &cobra.Command{
-	Use:   "kubeconfig",
+	Use:   "kubeconfig [local-path]",
 	Short: "Download the admin kubeconfig from the node",
-	Long:  ``,
+	Long: `Download the admin kubeconfig from the node.
```
andrewrynhard: nit: I think I'd prefer to be explicit here and skip the default to give it the same feel as `osctl cp`. But not gonna block. My OCD will get me enough later to where I can change in a follow up.

PR #1431, opened 2019-11-05
fix: verify that all etcd members are running before upgrading

andrewrynhard:
This verifies that all etcd members are running before performing an
upgrade. Without this we run the risk of destroying the etcd cluster.

### DIFF
--- a/internal/pkg/etcd/etcd.go
+++ b/internal/pkg/etcd/etcd.go
@@ -5,12 +5,16 @@
 package etcd
 
 import (
+	"context"
+	"fmt"
 	"net/url"
 	"time"
 
 	"go.etcd.io/etcd/clientv3"
 	"go.etcd.io/etcd/pkg/transport"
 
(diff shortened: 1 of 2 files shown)


internal/pkg/etcd/etcd.go
```diff
@@ -68,3 +72,41 @@ func NewClientFromControlPlaneIPs(creds *x509.PEMEncodedCertificateAndKey, endpo
...
+		resp, err := client.MemberList(context.Background())
+		if err != nil {
+			return err
+		}
+
+		for _, member := range resp.Members {
+			// If the member is not started, the name will be an empty string.
+			if len(member.Name) == 0 {
+				return fmt.Errorf("etcd member %d is not started, all members must be running to perform an upgrade", member.ID)
+			}
```
andrewrynhard: There is certainly more we can check but this is a good start for now.

PR #1432, opened 2019-11-05
refactor: rename Helper to Client

andrewrynhard:
The name helper isn't very good. This renames it to Client. A new func
was also added, NewForConfig, that will allow for the creation of the helper
client from an arbitrary Kubernetes REST config.


andrewrynhard:
The `NewForConfig` will be used by the operator.

PR #1435, opened 2019-11-05
fix: Disable support for proxy variables for apid.

bradbeam:
Since APId/gRPC connections should never go through a proxy, we will explicitly exclude
these environment variables from apid.

### DIFF
--- a/internal/app/machined/pkg/system/services/apid.go
+++ b/internal/app/machined/pkg/system/services/apid.go
@@ -112,8 +112,19 @@ func (o *APID) Runner(config runtime.Configurator) (runner.Runner, error) {
 	}
 
 	env := []string{}
+
 	for key, val := range config.Machine().Env() {
-		env = append(env, fmt.Sprintf("%s=%s", key, val))
+		switch strings.ToLower(key) {
+		// explicitly exclude proxy variables from apid since this will
+		// negatively impact grpc connections.
+		// ref: https://github.com/grpc/grpc-go/blob/0f32486dd3c9bc29705535bd7e2e43801824cbc4/clientconn.go#L199-L206
+		// ref: https://github.com/grpc/grpc-go/blob/63ae68c9686cc0dd26c4f7476d66bb2f5c31789f/proxy.go#L118-L144
+		case "no_proxy":
+		case "http_proxy":
+		case "https_proxy":
+		default:
+			env = append(env, fmt.Sprintf("%s=%s", key, val))
+		}
 	}
 
 	return restart.New(containerd.NewRunner(


internal/app/machined/pkg/system/services/apid.go
```diff
@@ -113,7 +113,17 @@ func (o *APID) Runner(config runtime.Configurator) (runner.Runner, error) {
 
 	env := []string{}
 	for key, val := range config.Machine().Env() {
-		env = append(env, fmt.Sprintf("%s=%s", key, val))
+		switch key {
+		// explicitly exclude proxy variables from apid since this will
+		// negatively impact grpc connections.
+		// ref: https://github.com/grpc/grpc-go/blob/0f32486dd3c9bc29705535bd7e2e43801824cbc4/clientconn.go#L199-L206
+		// ref: https://github.com/grpc/grpc-go/blob/63ae68c9686cc0dd26c4f7476d66bb2f5c31789f/proxy.go#L118-L144
+		case strings.ToLower("no_proxy"):
```
smira: should it be `switch strings.ToLower(key)` ?

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -113,7 +113,17 @@ func (o *APID) Runner(config runtime.Configurator) (runner.Runner, error) {
 
 	env := []string{}
 	for key, val := range config.Machine().Env() {
-		env = append(env, fmt.Sprintf("%s=%s", key, val))
+		switch key {
```
smira: qq, how could those variables get into environment, as we control the environment when we run the service?

internal/app/machined/pkg/system/services/apid.go
andrewrynhard: 👍

bradbeam: worked for me :trollface:

internal/app/machined/pkg/system/services/apid.go
smira: ah, got it, it could be user-supplied

bradbeam: We support setting environment variables via machine config ( a single set that gets used everywhere ):
Can set them like this:
```
[18:23:34] $ head -20 proxyconfigs/init.yaml 
version: v1alpha1
machine:
  type: init
  token: igdquc.xhmz8aackuvi3x0h
  env:
    http_proxy: http://10.0.0.1:3128
    https_proxy: http://10.0.0.1:3128
```
And then in all of our services we consume `config.Machine().Env()`
```
internal/app/machined/internal/phase/config/extra_env_vars.go:	for key, val := range r.Config().Machine().Env() {
internal/app/machined/pkg/system/services/udevd_trigger.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/osd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/udevd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/containerd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/ntpd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/etcd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/trustd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/kubelet.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/system-containerd.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/apid.go:	for key, val := range config.Machine().Env() {
internal/app/machined/pkg/system/services/networkd.go:	for key, val := range config.Machine().Env() {
```

bradbeam: Since APId is the only service that connects to other gRPC endpoints, it's the only one we need to filter atm.

PR #1439, opened 2019-11-05
test: add CLI integration test

smira:
This starts with a very simple test for `osctl version` using regexps as
output of the command depends a lot on current version.

We might use more of 'gold' matches for other commands potentially.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -344,7 +344,7 @@ ARG TAG
 ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
 RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=amd64 go test -c \
     -ldflags "-s -w -X ${VERSION_PKG}.Name=Client -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG}" \
-    -tags integration,integration_api \
+    -tags integration,integration_api,integration_cli \
     ./internal/integration
 
 FROM scratch AS integration-test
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -344,7 +344,7 @@ ARG TAG
 ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
 RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=amd64 go test -c \
```
bradbeam: should this be go build?

bradbeam:
Nice, starting to see how this is going to come together/be used... Just to make sure, this will require a talos cluster to be up correct?

Dockerfile
```diff
@@ -344,7 +344,7 @@ ARG TAG
 ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
```
andrewrynhard: nit: remove

andrewrynhard: whoops... I'm wrong

Dockerfile
smira: we can't use `go build` afaik to build unit-tests, the trick is `go test -c` compiles the test binary, but doesn't run it. so we can export that binary back and use it as a test runner every time we need to have an integration test

smira:
> Nice, starting to see how this is going to come together/be used... Just to make sure, this will require a talos cluster to be up correct?

yep, exactly, cluster should be up, but it can be literally any Talos cluster. All we need is basically `talosconfig`. I feel we should add more of the tests for multi-node APIs, but I started with something very simple for now.

PR #1440, opened 2019-11-05
feat: add metadata file to boot partition

andrewrynhard:
This introduces the notion of metadata for a node. In this initial pass
there are only two fields. A timestamp to indicate when the install was
performed, and a field to indicate if the install was performed as part
of an upgrade.

### DIFF
--- a/internal/app/machined/internal/phase/rootfs/check_install.go
+++ b/internal/app/machined/internal/phase/rootfs/check_install.go
@@ -5,10 +5,10 @@
 package rootfs
 
 import (
-	"os"
 	"path/filepath"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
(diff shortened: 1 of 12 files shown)


internal/app/machined/internal/phase/rootfs/check_install.go
```diff
@@ -32,6 +32,10 @@ func (task *CheckInstall) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 }
 
 func (task *CheckInstall) standard(r runtime.Runtime) (err error) {
-	_, err = os.Stat(filepath.Join(constants.BootMountPoint, "installed"))
+	_, err = metadata.Open(filepath.Join(constants.BootMountPoint, constants.MetadataFile))
```
andrewrynhard: The `metadata.Open` function has an additional check to ensure that the file is not empty.

PR #1443, opened 2019-11-06
test: add k8s integration tests

smira:
Once again, mostly groundwork and one simple test for node versions.

### DIFF
--- a/internal/integration/base/k8s.go
+++ b/internal/integration/base/k8s.go
@@ -1,0 +1,38 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// +build integration_k8s
+
+package base
+
+import (
+	"context"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+// K8sSuite is a base suite for K8s tests
+type K8sSuite struct {
+	APISuite
+
+	Clientset *kubernetes.Clientset
+}
+
+// SetupSuite initializes Kubernetes client
+func (k8sSuite *K8sSuite) SetupSuite() {
+	k8sSuite.APISuite.SetupSuite()
+
+	kubeconfig, err := k8sSuite.Client.Kubeconfig(context.Background())
+	k8sSuite.Require().NoError(err)
+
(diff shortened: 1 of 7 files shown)


internal/integration/base/k8s.go
```diff
@@ -0,0 +1,65 @@
...
+	APISuite
+
+	Clientset *kubernetes.Clientset
+}
+
+// SetupSuite initializes Kubernetes client
+func (k8sSuite *K8sSuite) SetupSuite() {
+	k8sSuite.APISuite.SetupSuite()
+
+	r, errCh, err := k8sSuite.Client.Kubeconfig(context.Background())
```
smira: I feel our `Kubeconfig` method is not really easy to use, as it requires reading .tar.gz and extracting files back. I thought of adding another method which return simple `[]byte` (based on code below), thoughts?

internal/integration/k8s/version.go
```diff
@@ -0,0 +1,49 @@
...
+	return "k8s.VersionSuite"
+}
+
+// TestExpectedVersion verifies that node versions matches expected
+func (suite *VersionSuite) TestExpectedVersion() {
+	nodes, err := suite.Clientset.CoreV1().Nodes().List(metav1.ListOptions{})
+	suite.Require().NoError(err)
+
+	expectedTalosVersion := fmt.Sprintf("Talos (%s)", suite.Version)
+	expectedContainerRuntimeVersion := fmt.Sprintf("containerd://%s", "1.3.0") // TODO: expose as constant
```
smira: how about exposing those versions in `constants`?

it would require bumping each time we get new software version, but at the same time I feel it's critical to verify at least some of these versions to make sure we get what is expected

internal/integration/base/k8s.go
andrewrynhard: Oof. Yeah this is pretty nasty for a simple API. Agree we should revert to just reading the file.

internal/integration/k8s/version.go
andrewrynhard: SGTM

internal/integration/base/k8s.go
smira: I'll update the client code to wrap back .tar.gz stream to a single file so that we could more easily get what we need

internal/integration/k8s/version.go
smira: I'll get it updated for the next version of this PR

internal/integration/base/k8s.go
smira: Added `client.Kubeconfig` as easy wrapper around `client.KubeconfigRaw` which provides streaming

PR #1451, opened 2019-11-07
feat(networkd): Add support for custom nameservers

bradbeam:
This adds support for specify nameservers in the config.

### DIFF
--- a/pkg/config/types/v1alpha1/network_config.go
+++ b/pkg/config/types/v1alpha1/network_config.go
@@ -20,6 +20,7 @@ import (
 type NetworkConfig struct {
 	NetworkHostname   string           `yaml:"hostname,omitempty"`
 	NetworkInterfaces []machine.Device `yaml:"interfaces,omitempty"`
+	NameServers       []string         `yaml:"nameservers,omitempty"`
 }
 
 // NetworkDeviceCheck defines the function type for checks.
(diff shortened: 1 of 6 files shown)


pkg/config/types/v1alpha1/network_config.go
```diff
@@ -20,6 +20,7 @@ import (
 type NetworkConfig struct {
 	NetworkHostname   string           `yaml:"hostname,omitempty"`
 	NetworkInterfaces []machine.Device `yaml:"interfaces,omitempty"`
+	NameServers       []string         `yaml:"resolvers,omitempty"`
```
andrewrynhard: Is `nameservers` a better public name here? I feel like it hints at what it is within `/etc/resolv.conf` better.

pkg/config/types/v1alpha1/network_config.go
```diff
@@ -41,6 +42,11 @@ func (n *NetworkConfig) Devices() []machine.Device {
 	return n.NetworkInterfaces
 }
 
+// Resolvers implements the Configurator interface.
+func (n *NetworkConfig) Resolvers() []string {
+	return n.NameServers
```
andrewrynhard: The mismatch in naming here is a little confusing. If we decide on resolvers or nameservers, I think I'd prefer to stick with it all the way through.

andrewrynhard:
Can we add to the docs too por favor?

pkg/config/types/v1alpha1/network_config.go
bradbeam: The mismatch is because we can't have a field and a method be named the same

andrewrynhard: I know its redundant, and I'm hoping to fix eventually, but in other cases we just add something before what we really wanna call it. I care less about this right now, as I intend on clean the whole interface implementation up.

pkg/config/types/v1alpha1/network_config.go
andrewrynhard: Just to be clear I meant the yaml tag: `yaml:”nameservers”`

Issue #1397, opened 2019-11-02
Target flag is broken for logs


andrewrynhard:
Something like `osctl logs -t $IP ntpd` does not get the logs from $IP, but from the node that is in the talosconfig.

bradbeam:
We only have support for non-streaming endpoints atm

andrewrynhard:
> We only have support for non-streaming endpoints atm

The `--target` flag should still work for every API IMO. I think we can just gate streaming APIs with a check on the value of `target` and enforce that it be a single IP.

PR #1454, opened 2019-11-07
chore: Add warning about stream support

bradbeam:
We dont currently support an interceptor for stream calls, so the proxy passes
through the connection to the talosconfig.target ( like it previously used to ).

```
[22:09:07] $ ./build/osctl-linux-amd64 logs networkd
2019/11/07 22:03:19.198989 main.go:36: discovering local network interfaces
2019/11/07 22:03:19.201181 main.go:54: overlaying config network configuration
2019/11/07 22:03:19.201201 main.go:66: creating interface lo
...

[22:09:17] $ ./build/osctl-linux-amd64 logs networkd --target 10.5.0.3
2019/11/07 22:09:20 Warning: `--target` is not supported for streaming methods. 127.0.0.1 will be used
2019/11/07 22:03:19.198989 main.go:36: discovering local network interfaces
2019/11/07 22:03:19.201181 main.go:54: overlaying config network configuration
2019/11/07 22:03:19.201201 main.go:66: creating interface lo
```

### DIFF
--- a/cmd/osctl/cmd/root.go
+++ b/cmd/osctl/cmd/root.go
@@ -110,15 +110,22 @@ func Execute() {
 
 // setupClient wraps common code to initialize osd client
 func setupClient(action func(*client.Client)) {
-	// Update context with grpc metadata for proxy/relay requests
-	md := metadata.New(make(map[string]string))
-	md.Set("targets", target...)
-	globalCtx = metadata.NewOutgoingContext(globalCtx, md)
-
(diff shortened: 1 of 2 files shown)


bradbeam:
closes #1397

andrewrynhard:
The goal of #1397 is to have `--targets` works as it used to. What is the technical blocker here? is there not a way to conditionally handle it the old way?

smira:
I wonder if we should have non-streaming version of the `logs` as well (as `tail -100` or something) which would play nice with multiple nodes?

PR #1455, opened 2019-11-08
feat: Add support for streaming apis in apid

bradbeam:
This brings in the recent updates to protoc-gen-proxy to allow support
for proxying streaming api requests. We artificially limit it to only the first
target specified in the list while we work through what multi target stream
support looks like.


bradbeam:
I goofed on my rebase and accidentally dropped the changes to apid/main. whoops.

PR #1459, opened 2019-11-09
docs: add public roadmap

andrewrynhard:
This adds the first pass out our public roadmap. It is intended to be a
living document.

### DIFF
--- a/docs/website/components/Header.vue
+++ b/docs/website/components/Header.vue
@@ -15,6 +15,11 @@
             <CommunityDropdown></CommunityDropdown>
           </li>
           <li class="link min-w-full md:min-w-0">
+            <a href="/roadmap">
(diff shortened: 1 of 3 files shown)


andrewrynhard:
Still a WIP.

docs/website/components/Header.vue
```diff
@@ -14,6 +14,11 @@
           <li class="link min-w-full md:min-w-0">
             <CommunityDropdown></CommunityDropdown>
           </li>
+          <li class="link min-w-full md:min-w-0">
+            <a href="/roadmap">
```
andrewrynhard: I'd like to eventually have a "Project" dropdown, but in case we can't get to that by kubecon, this should hold us over.

PR #1460, opened 2019-11-09
docs: add customization guide for running behing a proxy

andrewrynhard:
This adds docs on how to run Talos behind a proxy.

### DIFF
--- a/docs/website/content/v0.3/en/customization/proxy.md
+++ b/docs/website/content/v0.3/en/customization/proxy.md
@@ -1,0 +1,78 @@
+---
+title: 'Running Behind a Corporate Proxy'
+---
+
+## Creating an Image for MITM Proxies
+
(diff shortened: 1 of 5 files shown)


andrewrynhard:
Closes talos-systems/talos#1452.

docs/website/content/v0.3/en/customization/proxy.md
```diff
@@ -0,0 +1,72 @@
...
+    no_proxy: <no proxy>
+```
+
+Additionally, configure the `nameservers`:
+
+```yaml
+machine:
+  env:
+  ...
+  network:
```
bradbeam: we should probably add in the ntp server config in here as well. tbh i'm not sure how it translates to yaml keys ( or if that part is currently broken ), but we consume it via `config.Machine().Time().Server()`.

PR #1462, opened 2019-11-09
docs: add autogenerated config reference

andrewrynhard:
This adds a small program to parse our config structs and generate
markdown from them. This will allow us to enforce a standard and require
documentation for fields as they get added.

Closes talos-systems/talos#1377


docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+This will generate a machine config for each node type, and a talosconfig.
+The following is an example of an `init.yaml`:
+
+```yaml
+version: v1alpha1
+machine:
+  type: init
+  token: 5dt69c.npg6duv71zwqhzbg
+  ca:
+    crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJIekNCMHFBREFnRUNBaEIwL2grQ2hocGhWUXhXQ2x4dSsxSjVNQVVHQXl0bGNEQVFNUTR3REFZRFZRUUsKRXdWMFlXeHZjekFlRncweE9URXhNRGt4T0RRM016bGFGdzB5T1RFeE1EWXhPRFEzTXpsYU1CQXhEakFNQmdOVgpCQW9UQlhSaGJHOXpNQ293QlFZREsyVndBeUVBWXRKQ2FkWnhOK1A2eUVlNHNEM2dLQXJ2Y1ZmU3JqdlJ5YzNmCnRhTHorcFNqUWpCQU1BNEdBMVVkRHdFQi93UUVBd0lDaERBZEJnTlZIU1VFRmpBVUJnZ3JCZ0VGQlFjREFRWUkKS3dZQkJRVUhBd0l3RHdZRFZSMFRBUUgvQkFVd0F3RUIvekFGQmdNclpYQURRUUNrczNuMUFNeDhBcGluaEk1YgpaQU9IVFJ0dEZiSXQ3RmU1VFJ2aEdvN21rR0FnWmI5M21ZbGx4NFZjL2VKc0p0aVJURHN6cVp5VnlUVWVCaU45CnF5b0wKLS0tLS1FTkQgQ0VSVElGSUNBVEUtLS0tLQo=
```
bradbeam: i feel like we should try to mask or abbreviate the certificates here. It's not super helpful to see these giant blocks.

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+nolint: lll
```
bradbeam: shouldn't be here :)

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Reference
+
+### type
```
bradbeam: i feel like these keys need to be fully qualified `machine.type` unless we split things up into `machine` config reference and `cluster` config reference docs.

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+
+Type: `string`
```
bradbeam: 👍

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+The `token` is used by a machine to join the PKI of the cluster.
+Using this token, a machine will create a certificate signing request (CSR), and request a certificate that will be used as its' identity.
+
+Examples:
+
+```yaml
+token: 328hom.uqjzh6jnn2eie9oi
+```
+
+> Warning: It is important to ensure that this token is correct sinc a machine's certificate has a short TTL by default
```
bradbeam: sp  s/sinc/since/

bradbeam: think this warning could use additional context / detail. 
since the token gets generated, id assume most people politely ignore it. so there should be low chance of the token being incorrect. but, in the case that the token does get truncated during editing, what does that actually translate to? i don't think the certificate ttl comes in to play at this point because most likely the <broken> machine won't be able to talk to trustd to request a certificate.

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+
+### ca
+
+Type: `PEMEncodedCertificateAndKey`
```
bradbeam: the type here is technically correct but not useful to end users. i guess we could try to get fancy and link to godoc here to articulate the type definition, but not sure i care for that either.

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+```yaml
+ca:
+  crt: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJIekNCMHF...
+  key: LS0tLS1CRUdJTiBFRDI1NTE5IFBSSVZBVEUgS0VZLS0tLS0KTUM...
+
+```
+
+### certSANs
+
+Type: `array`
```
bradbeam: probably need to qualify it as a string array / string list.. or maybe just list? nit on the wording i guess, not sure what the correct phrasing would be

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+certSANs:
+  - 10.0.0.10
+  - 172.16.0.10
+  - 192.168.0.10
+
+```
+
+### kubelet
+
+Type: `KubeletConfig`
```
bradbeam: similar comment here with PEMEncoded... `kubeletconfig` isn't helpful on it's own

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -0,0 +1,276 @@
...
+  wipe:
+  force:
+
+```
+
+### files
+
+Type: `array`
+
+The `files` field allows the addition of user specified files.
```
bradbeam: probably want to clarify the contents should be a string/yaml block ( be explicit that contents shouldn't be b64 encoded ). i don't know why but in my head i assumed it was.. probably after seeing all the b64 encoded cert/key blobs

hack/go2md/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+go build .
+
+./go2md /root/code/github.com/talos-systems/talos/pkg/config/types/v1alpha1/machine_config.go /root/code/github.com/talos-systems/talos/docs/website/content/v0.3/en/configuration/v1alpha1.md
```
bradbeam: I wonder if we can add this to a go generate block.. maybe under `pkg/config/types/v1alpha1/docs.go`?

not sure if this is entirely right, but im sure you get the gist
```
//go:generate -command go run ../../../../hack/go2md/main.go ../../../../docs/website/content/v0.3/en/configuration/v1alpha1.md
```

pkg/config/types/v1alpha1/v1alpha1_configurator.go
```diff
@@ -0,0 +1,348 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: This is just a lift of all the methods that were split across multiple files and dropping them into a single file.

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -0,0 +1,474 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: This is a lift of all the different types spread across multiple files and dropped into a single file. It is also adds the  `docgen` annotations.

pkg/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -7,45 +7,49 @@
 package v1alpha1
```
andrewrynhard: This moves all validation related code into a single file.

docs/website/content/v0.3/en/configuration/v1alpha1.md
andrewrynhard: The best I could come up with is to have all options related to a struct under a common header. Checkout the new web preview. The Golang AST stuff is pretty hairy. Maybe something we can look at later, but with the sections it is actually pretty nice IMHO. We don't need to have a fully qualified since the entire section is about the specific struct.

docs/website/content/v0.3/en/configuration/v1alpha1.md
andrewrynhard: It comes into play if/when someone wants to or accidentally changes the token, say for rotation. Its important to get makes sure the nodes have the correct token because they renew so often. Since we don't have a way, yet, to update a machine's token, this means bouncing the node, but still important to point out that they could potentially lose node access fairly quickly if they don't update worker tokens after updating master tokens.

docs/website/content/v0.3/en/configuration/v1alpha1.md
andrewrynhard: Another limitation of AST. I intend on linking to the local sections that contains the info about the type, but things get weird if the type is from another package. We may have to live with this until we can spend a lot more time digging into the AST library. At the very least, we now have sections dedicated to non-builtin-in types that are part of the current package.

docs/website/content/v0.3/en/configuration/v1alpha1.md
andrewrynhard: I'm working on getting the element type using the AST lib, still a WIP.

hack/go2md/run.sh
andrewrynhard: done

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -0,0 +1,515 @@
...
+	//   description: |
+	//     The service subnet CIDR.
+	//   examples:
+	//     -  |
+	//       serviceSubnets:
+	//         - 10.96.0.0/12
+	ServiceSubnet []string `yaml:"serviceSubnets"`
+}
+
+// Bond contains the various options for configuring a bonded interface.
```
bradbeam: We probably should drop this for now while we don't actually support bonding.

andrewrynhard: I think I want to add functionality to `docgen` that would allow skipping a field so we can keep the definition, but not render it.

Issue #1453, opened 2019-11-07
Explore ways to handle entirely static network configuration


bradbeam:
The current bootup/networkd implementation follows the pattern of:

```
- DHCP all interfaces
- Hope we get an address
- Attempt to download userdata
```

This works on most cloud providers, but for on prem situations, dhcp may not be an option at all. We should figure out some ways/methods/etc to work around this.

Only thought that comes to mind: 

- kernel args -- while this is probably the 'simplest' to add on, it gets complex quickly. We'd need to expose not only the addressing configuration but routing as well. I'm concerned we'd end up with a whole hodge podge of json.. ex,  `talos.network = '{ device: eth0, cidr: 10.0.0.2, routes: [ { src: x, dst: y }, ... ] }'`

andrewrynhard:
@Ulexus Pinging you here just in case you have thought about this before, and I know it may apply to you.

Ulexus:
We do have mechanisms for disabling DHCP now.  It just has to be declared for _each_ network interface in both the kernel commandline _and_ the talos config yaml.  It works, but the user experience is not great.

Two features which would make for greater convenience:
  - automatic population of the kernel commandline parameters disabling DHCP on installation referenced from the talos network configuration in the YAML (so these don't have to be managed separately by the user)
  - ability to globally disable DHCP on all interfaces (reduce boilerplate for nodes with large numbers of network interfaces)

All of that said, I do agree that the default SHOULD be to have DHCP _enabled_.  I wouldn't want to change that.  It would just be nice to make it more convenient to kill it when appropriate.

Ulexus:
Ah, but I neglected the meat of the problem:  how do you specify the network config in the chicken-and-egg of talos config fetchment over the network.

The JSON-encoded talos.network is not a horrible idea.  I do wonder about the spaces in the parameter, though, such that additionally base64-encoding might be good?

The kernel itself does already have a pretty complete [network configuration](https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt) with `ip=XXX` (including, for instance, DNS and even NTP).  I would definitely rather re-use that than invent something new.  On the other hand, the _intention_ of that option is for configuring minimal networking to bootstrap from NFS (which is why the docs are bundled in with the nfsroot stuff), so there _might_ be some side-effects I don't know about.

[...]

PR #1463, opened 2019-11-10
feat: Disable networkd configuration if `ip` kernel parameter is specified

bradbeam:
This allows the kernel argument `ip` to take precedence over networking configuration. Documentation for
this parameter can be found here https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt


bradbeam:
closes #1453

PR #1471, opened 2019-11-11
fix: mount extra disks after system disk

andrewrynhard:
The extra disks functionality was completely broken. One fundamental
issue was that we were attempting to create and mount the partitions
before the system disk was created. This moves the extra disks tasks to
the correct part of the boot sequnce. This also adds a simple check that
refuses to operate on a disk if any partitions are found.


internal/pkg/installer/installer.go
```diff
@@ -60,7 +60,7 @@ func (i *Installer) Install(sequence runtime.Sequence) (err error) {
 
 	// Partition and format the block device(s).
 
-	if err = i.manifest.ExecuteManifest(i.manifest); err != nil {
```
andrewrynhard: Removed this inception.

internal/pkg/installer/manifest/manifest.go
```diff
@@ -112,29 +111,12 @@ func NewManifest(install machine.Install) (manifest *Manifest, err error) {
...
-				Device: extra.Device,
-				Size:   part.Size,
-				Force:  true,
-				Test:   false,
-			}
-
-			manifest.Targets[extra.Device] = append(manifest.Targets[extra.Device], extraTarget)
-		}
-	}
-
```
andrewrynhard: For some reason, we had extra disks as part of the installation, but it shouldn't be. The extra disks should stand completely independent of the installation instructions.

internal/pkg/installer/manifest/manifest.go
```diff
@@ -112,29 +111,12 @@ func NewManifest(install machine.Install) (manifest *Manifest, err error) {
...
 	return manifest, nil
 }
 
 // ExecuteManifest partitions and formats all disks in a manifest.
-func (m *Manifest) ExecuteManifest(manifest *Manifest) (err error) {
-	for dev, targets := range manifest.Targets {
+func (m *Manifest) ExecuteManifest() (err error) {
+	for dev, targets := range m.Targets {
```
andrewrynhard: Inception.

internal/pkg/installer/manifest/manifest.go
```diff
@@ -196,17 +178,7 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
...
-	// multiple places.
-	switch dev := t.Device; {
-	case strings.HasPrefix(dev, "/dev/nvme"):
-		fallthrough
-	case strings.HasPrefix(dev, "/dev/loop"):
-		t.PartitionName = t.Device + "p" + strconv.Itoa(int(part.No()))
-	default:
-		t.PartitionName = t.Device + strconv.Itoa(int(part.No()))
-	}
+	t.PartitionName = util.Partpath(t.Device, int(part.No()))
```
andrewrynhard: Finally addressed the `TODO` here. This DRYs things up a tiny bit.

pkg/blockdevice/util/util.go
```diff
@@ -54,8 +55,27 @@ func DevnameFromPartname(partname string) (devname string, err error) {
 	case strings.HasPrefix(p, "vd"):
 		fallthrough
 	case strings.HasPrefix(p, "xvd"):
-		return strings.TrimRight(partname, partno), nil
+		return strings.TrimRight(p, partno), nil
 	default:
-		return "", fmt.Errorf("could not determine dev name from partition name: %s", partname)
+		return "", fmt.Errorf("could not determine dev name from partition name: %s", p)
```
andrewrynhard: Minor nit fixes.

pkg/blockdevice/util/util.go
```diff
@@ -54,8 +55,27 @@ func DevnameFromPartname(partname string) (devname string, err error) {
...
+	default:
+		partname = fmt.Sprintf("%s%d", p, n)
+	}
+
+	return partname
+}
+
+func Partpath(d string, n int) string {
+	return filepath.Join("/dev", Partname(d, n))
+}
```
andrewrynhard: This is the answer to the `TODO`.

pkg/config/machine/machine.go
```diff
@@ -96,7 +97,6 @@ type Route struct {
 type Install interface {
 	Image() string
 	Disk() string
-	ExtraDisks() []Disk
```
andrewrynhard: Non-system disks should not be part of the install block.

pkg/config/types/v1alpha1/machine_config.go
```diff
@@ -19,6 +19,7 @@ type MachineConfig struct {
 	MachineCertSANs []string                          `yaml:"certSANs"`
 	MachineKubelet  *KubeletConfig                    `yaml:"kubelet,omitempty"`
 	MachineNetwork  *NetworkConfig                    `yaml:"network,omitempty"`
+	MachineDisks    []machine.Disk                    `yaml:"disks,omitempty"`
```
andrewrynhard: Disks are now a top-level item right up there with networking.

internal/app/machined/internal/phase/config/extra_disks.go
```diff
@@ -0,0 +1,106 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: Renamed the file to `extra_disks.go` since `disks` is the word of choice in the rest of the code when talking about block devices.

internal/app/machined/internal/phase/config/extra_disks.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// TODO(andrewrynhard): This is disgusting, but it works. We should revisit
+	// this at a later time.
+	for _, p := range probed {
+		for _, disk := range r.Config().Machine().Disks() {
+			for i := range disk.Partitions {
+				partname := util.PartPath(disk.Device, i+1)
+				if p.Path == partname {
+					log.Printf(("found existing partitions for %q"), disk.Device)
+					return nil
```
bradbeam: todo for later:
probably don't want return in the case we only have 1 of X partitions created

internal/app/machined/internal/phase/config/extra_disks.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, extra := range r.Config().Machine().Disks() {
+		for i, part := range extra.Partitions {
+			partname := util.PartPath(extra.Device, i+1)
+			mountpoints.Set(partname, mount.NewMountPoint(partname, part.MountPoint, "xfs", unix.MS_NOATIME, ""))
+		}
+	}
+
+	extras := manager.NewManager(mountpoints)
+	if err = extras.MountAll(); err != nil {
+		log.Println("WARNING:", err)
```
bradbeam: let's return err here ( ex, if external etcd dir fails to mount this would be no good )

PR #1473, opened 2019-11-11
feat: Add context key to osctl

### LINKED ISSUE #1171: osctl context flag
We should support one off context changes 

```
osctl ps --context=cluster-a
```

sekerin:
Closes #1171 Added context switch for specific osctl command

### DIFF
--- a/cmd/osctl/cmd/root.go
+++ b/cmd/osctl/cmd/root.go
@@ -39,6 +39,7 @@ var (
 	rsa            bool
 	talosconfig    string
 	target         []string
+	cmdcontext     string
 )
 
 // rootCmd represents the base command when called without any subcommands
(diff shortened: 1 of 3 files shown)


cmd/osctl/cmd/root.go
```diff
@@ -115,7 +117,7 @@ func setupClient(action func(*client.Client)) {
 	md.Set("targets", target...)
 	globalCtx = metadata.NewOutgoingContext(globalCtx, md)
 
-	t, creds, err := client.NewClientTargetAndCredentialsFromConfig(talosconfig)
+	t, creds, err := client.NewClientTargetAndCredentialsFromConfig(talosconfig, cmdcontext)
```
andrewrynhard: Instead of this I think it would be better to follow the same pattern we use for `--targets` and set this in https://github.com/talos-systems/talos/blob/a4e1479b07c3951a9f4f5e38a9d5bec4eebc8549/cmd/osctl/cmd/root.go#L112

andrewrynhard:
@sekerin Thanks again for the PR! Left some feedback. Feel free to ping me if you have questions.

cmd/osctl/cmd/root.go
sekerin: Thank you for review! But i don't understand pattern for `--targets`. Can you explain me?

andrewrynhard: Sorry I wasn't very clear. And looking closer at my suggest I was just wrong :)

andrewrynhard:
Thanks @sekerin!

PR #1478, opened 2019-11-11
feat: implement grpc request loggging

smira:
Logging is pretty simple and bare minimum is being logged. I believe
better logging can be provided for apid when it does fan-out, but that
is beyond the scope for the first PR.

Sample logs:

```
$ osctl-linux-amd64 logs machined-api
machined 2019/11/11 21:16:43 OK [/machine.Machine/ServiceList] 0.000ms unary Success (:authority=unix:/run/system/machined/machine.sock;content-type=application/grpc;user-agent=grpc-go/1.23.0)
machined 2019/11/11 21:17:09 Unknown [/machine.Machine/Logs] 0.000ms stream open /run/system/log/machined.log: no such file or directory (:authority=unix:/run/system/machined/machine.sock;content-type=application/grpc;user-agent=grpc-go/1.23.0)
```

### DIFF
--- a/pkg/grpc/middleware/log/log.go
+++ b/pkg/grpc/middleware/log/log.go
@@ -2,4 +2,90 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// Package log provides simple grpc logging middleware
 package log
+
+import (
+	"context"
+	"log"
+	"sort"
+	"strings"
+	"time"
+
+	"google.golang.org/grpc"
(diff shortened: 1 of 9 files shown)


pkg/grpc/middleware/log/log.go
```diff
@@ -2,4 +2,90 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// Package log provides simple grpc logging middleware
```
andrewrynhard: ❤️ We need to do this more often.

PR #1491, opened 2019-11-12
feat: Add support for resetting the network during bootup

bradbeam:
This introduces the ability to reset the network interface during the bootup sequence.
This allows for user defined static networking to be the only configuration on the
network interface instead of potentially dhcp+static.

### DIFF
--- a/internal/app/machined/internal/phase/network/reset.go
+++ b/internal/app/machined/internal/phase/network/reset.go
@@ -7,8 +7,6 @@ package network
 import (
 	"log"
 
-	"golang.org/x/sys/unix"
-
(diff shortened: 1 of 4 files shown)


internal/app/machined/internal/phase/network/network.go
```diff
@@ -56,6 +56,12 @@ func (task *InitialNetworkSetup) runtime(r runtime.Runtime) (err error) {
 		return err
 	}
 
+	log.Println("overlaying config network configuration")
```
andrewrynhard: nit: Can we say something like `"overriding discovered network configuration with use defined options"`? I get `overlay`, but something like the above is super explicit, which I find to be helpful in the early boot.

PR #1492, opened 2019-11-12
fix: create resolv.conf when ip kernel arg is set

andrewrynhard:
This symlinks /etc/resolv.conf to /proc/net/pnp.

### DIFF
--- a/internal/app/machined/internal/phase/rootfs/etc/etc.go
+++ b/internal/app/machined/internal/phase/rootfs/etc/etc.go
@@ -6,6 +6,7 @@ package etc
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
@@ -13,6 +14,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/talos-systems/talos/internal/pkg/kernel"
 	"github.com/talos-systems/talos/pkg/version"
 
 	"golang.org/x/sys/unix"
@@ -85,19 +87,31 @@ func Hosts(hostname string) (err error) {
 // ResolvConf copies the resolv.conf generated in the early boot to the new
 // root.
 func ResolvConf() (err error) {
-	target := "/run/system/etc/resolv.conf"
+	option := kernel.ProcCmdline().Get("ip").First()
+	switch option {
+	case nil:
+		target := "/run/system/etc/resolv.conf"
 
-	var f *os.File
+		var f *os.File
 
-	if f, err = os.OpenFile(target, os.O_WRONLY|os.O_CREATE, 0644); err != nil {
-		return err
-	}
+		if f, err = os.OpenFile(target, os.O_WRONLY|os.O_CREATE, 0644); err != nil {
+			return err
+		}
 
-	// nolint: errcheck
-	defer f.Close()
+		// nolint: errcheck
+		defer f.Close()
 
-	if err = unix.Mount("/run/system/etc/resolv.conf", "/etc/resolv.conf", "", unix.MS_BIND, ""); err != nil {
-		return fmt.Errorf("failed to create bind mount for /etc/resolv.conf: %w", err)
+		if err = unix.Mount("/run/system/etc/resolv.conf", "/etc/resolv.conf", "", unix.MS_BIND, ""); err != nil {
+			return fmt.Errorf("failed to create bind mount for /etc/resolv.conf: %w", err)
+		}
+	default:
+		if _, err = os.Stat("/proc/net/pnp"); err != nil {
+			return errors.New("failed to symlink /etc/resolv.conf to /proc/net/pnp")
+		}
+
+		if err = os.Symlink("/etc/resolv.conf", "/proc/net/pnp"); err != nil {
+			return err
+		}
 	}
 
 	return nil


internal/app/machined/internal/phase/rootfs/etc/etc.go
```diff
@@ -85,19 +86,27 @@ func Hosts(hostname string) (err error) {
...
+		// nolint: errcheck
+		defer f.Close()
 
-	if err = unix.Mount("/run/system/etc/resolv.conf", "/etc/resolv.conf", "", unix.MS_BIND, ""); err != nil {
-		return fmt.Errorf("failed to create bind mount for /etc/resolv.conf: %w", err)
+		if err = unix.Mount("/run/system/etc/resolv.conf", "/etc/resolv.conf", "", unix.MS_BIND, ""); err != nil {
+			return fmt.Errorf("failed to create bind mount for /etc/resolv.conf: %w", err)
+		}
+	default:
+		if err = os.Symlink("/etc/resolv.conf", "/proc/net/pnp"); err != nil {
```
Ulexus: This is probably fine, but I would actually check for the existence of `/proc/net/pnp` (and maybe even that it contains a line with "nameserver") instead of using the somewhat surrogate `ip` cmdline check.

andrewrynhard: 👍

Ulexus: Actually, that file exists whether or not `ip=` was issued, so the existence check would be meaningless.

bradbeam:
While I understand this is appropriate, it does feel weird having resolv.conf being RW in some cases. If resolv.conf is linked to /proc, we can't write to it. Can we instead read in pnp and write out the actual file?

andrewrynhard:
> While I understand this is appropriate, it does feel weird having resolv.conf being RW in some cases. If resolv.conf is linked to /proc, we can't write to it. Can we instead read in pnp and write out the actual file?

The kernel is responsible for keep that file up to date. So in the case of `ip=dhcp` it could change. We could setup a look to look periodically, but that is gross. I think symlink is the most maintainable way.

internal/app/machined/internal/phase/rootfs/etc/etc.go
```diff
@@ -85,19 +87,31 @@ func Hosts(hostname string) (err error) {
...
+		defer f.Close()
 
-	// nolint: errcheck
-	defer f.Close()
+		if err = unix.Mount("/run/system/etc/resolv.conf", "/etc/resolv.conf", "", unix.MS_BIND, ""); err != nil {
+			return fmt.Errorf("failed to create bind mount for /etc/resolv.conf: %w", err)
+		}
+	default:
+		if _, err = os.Stat("/proc/net/pnp"); os.IsNotExist(err) {
+			return errors.New("failed to setup /etc/resolv.conf, /proc/net/pnp does not exist")
```
smira: this looks wrong as it ignores errors which are not `os.IsNotExist()`

smira: I would probably simply return any `err != nil` via `fmt.Errorf()` context

andrewrynhard: Good point. It gets a little ugly because`!os.IsNotExist()` doesn't guarantee that it does exist because the error _could_ have been something else. I can try to rework it.

Ulexus:
Generally speaking, I wonder if a better choice wouldn't be to simply _always_ make a first pass creation of `/run/system/resolv.conf` have the contents of `/proc/net/pnp`.  The networkd will do its thing if it is run, overriding it as appropriate.

Ulexus:
```
  11.327630] [talos] [phase]: basic system configuration
[   11.333607] [talos] [phase]: basic system configuration error running task: failed to link kernel DNS autoconfig to resolv.conf: symlink /etc/resolv.conf /proc/net/pnp: file exists
[   11.362502] [talos] [phase]: basic system configuration done, 34.871708ms
[   11.369936] [talos] error running phase "basic system configuration": 1 error occurred:
[   11.369936]  * failed to link kernel DNS autoconfig to resolv.conf: symlink /etc/resolv.conf /proc/net/pnp: file exists
```

andrewrynhard:
> Generally speaking, I wonder if a better choice wouldn't be to simply _always_ make a first pass creation of `/run/system/resolv.conf` have the contents of (or symlink to) `/proc/net/pnp`. Then networkd will do its thing if it is run, overwriting it.
> 
> That would result in less internal branching, if nothing else. This phase would _always_ just symlink in, as it does now.

I don’t think pnp exists unless the kernel arg is supplied.

Ulexus:
It does on my machine:

```
$ cat /proc/net/pnp
#MANUAL
$ cat /proc/cmdline 
\EFI\gentoo\bzImage ro dobtrfs root=/dev/nvme0n1p2 real_root=/dev/nvme0n1p2 rootflags=subvol=@root,device=/dev/nvme0n1p2,device=/dev/nvme2n1p2 rootfstype=btrfs init=/usr/lib/systemd/systemd radeon.dpm=1 amdgpu.dc=0 initrd=\EFI\gentoo\initrd
```

Ulexus:
Even if it doesn't, creating a symlink to a non-existent file doesn't make anything worse.  In the normal course of events, networkd would just remove the link and create its own version.

Ulexus:
Though you are correct; spot-checking a number of other machines does indicate it is not universally present on Linux.  Simple enough with the os.Stat() check, though.

I'm still partial to a layered approach which doesn't internally branch, but that's just personal preference.

internal/app/machined/internal/phase/rootfs/etc/etc.go
smira: I think it looks better now, but I would rather still keep original `err`, as it might be anything, with `fmt.Errorf`

andrewrynhard: Whoops sorry, was rushing this as I was running out the door.

andrewrynhard:
We decided to go a different route for this. @bradbeam Will handle it.

PR #1493, opened 2019-11-13
feat: allow extra arguments to be passed to etcd

andrewrynhard:
This allows for an arbitrary set of args to be passed to etcd. Due to
the fact the we do the discovery of the current cluster state, we
blacklist initial-cluster. Additionally, we black list args that would
makes sense to modify, like the path to the certs.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -30,6 +30,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/etcd"
 	"github.com/talos-systems/talos/internal/pkg/metadata"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
+	"github.com/talos-systems/talos/pkg/argsbuilder"
 	"github.com/talos-systems/talos/pkg/config/machine"
 	"github.com/talos-systems/talos/pkg/constants"
 	"github.com/talos-systems/talos/pkg/crypto/x509"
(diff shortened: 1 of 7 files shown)


andrewrynhard:
Still a WIP.

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -359,3 +317,85 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial
...
+		if extraArgs.Contains(k) {
+			return nil, blacklistError(k)
+		}
+	}
+
+	if extraArgs.Contains("initial-cluster-state") {
+		return nil, blacklistError("initial-cluster-state")
+	}
+
+	a.Set("initial-cluster-state", "new")
```
bradbeam: Why not have `"initial-cluster-state": "new"` added to the args list above?

It'd allow us to remove this whole bit because it should get caught via blacklist in the previous check (362-366).

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -359,3 +317,85 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial
...
+	ips, err := net.IPAddrs()
+	if err != nil {
+		return nil, fmt.Errorf("failed to discover IP addresses: %w", err)
+	}
+
+	if len(ips) == 0 {
+		return nil, errors.New("failed to discover local IP")
+	}
+
+	a := argumentor.Args{
```
bradbeam: could we rename `a` to `defaultArgs` or something like that?

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -359,3 +317,85 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial
...
+	}
+
+	if extraArgs.Contains("initial-cluster-state") {
+		return nil, blacklistError("initial-cluster-state")
+	}
+
+	a.Set("initial-cluster-state", "new")
+
+	if !extraArgs.Contains("initial-cluster") {
+		existing := config.Machine().Type() == machine.ControlPlane || metadata.Upgraded
```
bradbeam: Does it make more sense to set these by default and then let the merge below overwrite them if present?

Having these blocks of statements is nice for grouping purposes, but feels like a bit more work than is needed.

pkg/argumentor/argumentor_interface.go
```diff
@@ -0,0 +1,62 @@
...
+	for key, val := range a {
+		args = append(args, fmt.Sprintf("--%s=%s", key, val))
+	}
+
+	return args
+}
+
+// Get returns an args value.
+func (a Args) Get(k Key) Value {
+	return a[k]
```
bradbeam: do we need to do this?
```
if v, ok := a[k]; ok {
  return v
}
```

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -359,3 +317,85 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial
...
+
+	if !extraArgs.Contains("initial-advertise-peer-urls") {
+		a.Set("initial-advertise-peer-urls", fmt.Sprintf("https://%s:2380", ips[0].String()))
+	}
+
+	if !extraArgs.Contains("advertise-client-urls") {
+		a.Set("advertise-client-urls", fmt.Sprintf("https://%s:2379", ips[0].String()))
+	}
+
+	args := a.Merge(extraArgs).Args()
```
bradbeam: nit, we could 
```
return a.Merge(extraArgs).Args(), nil
```

internal/app/machined/pkg/system/services/etcd.go
andrewrynhard: Still a WIP

internal/app/machined/pkg/system/services/etcd.go
andrewrynhard: That means we would go through the whole discovery process when we don't need to.

pkg/argumentor/argumentor_interface.go
andrewrynhard: No, it will return the zero value for the map element type.

internal/app/machined/pkg/system/services/etcd.go
andrewrynhard: They're not default in the sense that they can be overwritten. Will come up with something more descriptive.

andrewrynhard:
Still needs tests

PR #1494, opened 2019-11-13
feat: Add support for kernel network config (ip=)

bradbeam:
This introduces a kernel addressing method. Given that the kernel handles this configuration
this mode is largely noop ( address should already exist/be defined ), however, this adds
the ability to handle the addressing method as a first class citizen. TLDR; we can support
configuring resolv.conf via kernel ip= argument.

### DIFF
--- a/internal/app/machined/internal/phase/network/network.go
+++ b/internal/app/machined/internal/phase/network/network.go
@@ -12,7 +12,6 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/app/networkd/pkg/networkd"
 	"github.com/talos-systems/talos/internal/app/networkd/pkg/nic"
-	"github.com/talos-systems/talos/internal/pkg/kernel"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
 )
 
@@ -36,13 +35,6 @@ func (task *InitialNetworkSetup) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 
 // nolint: gocyclo
 func (task *InitialNetworkSetup) runtime(r runtime.Runtime) (err error) {
-	// Check to see if a static IP was set via kernel args;
-	// if so, we'll skip the initial dhcp discovery
-	if option := kernel.ProcCmdline().Get("ip").First(); option != nil {
-		log.Println("skipping initial network setup, found kernel arg 'ip'")
-		return nil
-	}
-
 	nwd, err := networkd.New()
 	if err != nil {
 		return err
@@ -60,9 +52,13 @@ func (task *InitialNetworkSetup) runtime(r runtime.Runtime) (err error) {
 	if r.Config() != nil {
 		log.Println("merging user defined network configuration")
 
-		if err = netconf.BuildOptions(r.Config()); err != nil {
+		if err = netconf.BuildOptionsFromConfig(r.Config()); err != nil {
 			log.Fatal(err)
 		}
+	}
+
+	if err = netconf.BuildOptionsFromKernel(); err != nil {
+		log.Fatal(err)
 	}
 
 	// Configure specified interface
--- a/internal/app/networkd/pkg/address/kernel.go
+++ b/internal/app/networkd/pkg/address/kernel.go
@@ -1,0 +1,181 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package address
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"golang.org/x/sys/unix"
+
+	"github.com/talos-systems/talos/internal/pkg/kernel"
+)
+
+// Kernel implements the Addressing interface
+type Kernel struct {
+	Addr              string
+	ServerAddress     string
+	Gateway           string
+	Netmask           string
+	FQDN              string
+	Device            string
+	Auto              string
+	PrimaryResolver   string
+	SecondaryResolver string
+	NTPServer         string
+}
+
+// Discover doesnt do anything in the static configuration since all
+// the necessary configuration data is supplied via config.
+// nolint: gocyclo
+func (k *Kernel) Discover(ctx context.Context) error {
+	var (
+		option *string
+		err    error
+	)
+
+	if option = kernel.ProcCmdline().Get("ip").First(); option == nil {
+		return fmt.Errorf("%s", "no kernel.ip argument supplied, skipping")
+	}
+
+	// TODO may need to move this earlier in networkd; if ip.dhcp is specified
+	// we shouldnt do anything because we dont know which (all?) interfaces the kernel
+	// will autoconfigure
+	if *option == "dhcp" {
+		return fmt.Errorf("unsupported kernel.ip configuration method: %s", *option)
+	}
+
+	// https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt
+	// ip=<client-ip>:<server-ip>:<gw-ip>:<netmask>:<hostname>:<device>:<autoconf>:<dns0-ip>:<dns1-ip>:<ntp0-ip>
+	fields := strings.Split(*option, ":")
+
+	if len(fields) < 4 {
+		return fmt.Errorf("invalid kernel option syntax: found %d fields but there must be at least 4", len(fields))
+	}
+
+	for idx, field := range fields {
+		switch idx {
+		case 0:
+			k.Addr = field
+		case 1:
+			k.ServerAddress = field
+		case 2:
+			k.Gateway = field
+		case 3:
+			k.Netmask = field
+		case 4:
+			k.FQDN = field
+		case 5:
+			k.Device = field
+		case 6:
+			k.Auto = field
+		case 7:
+			k.PrimaryResolver = field
+		case 8:
+			k.SecondaryResolver = field
+		case 9:
+			k.NTPServer = field
+		}
+	}
+
+	return err
+}
+
+// Name returns back the name of the address method.
+func (k *Kernel) Name() string {
+	return "kernel"
+}
+
+// Address returns the IP address
+func (k *Kernel) Address() *net.IPNet {
+	return &net.IPNet{
+		IP:   net.ParseIP(k.Addr),
+		Mask: k.Mask(),
+	}
+}
+
+// Mask returns the netmask.
+func (k *Kernel) Mask() net.IPMask {
+	netmask := net.ParseIP(k.Netmask).To4()
+	return net.IPv4Mask(netmask[0], netmask[1], netmask[2], netmask[3])
+}
+
+// MTU returns the specified MTU.
+func (k *Kernel) MTU() uint32 {
+	return 1500
+}
+
+// TTL returns the address lifetime. Since this is static, there is
+// no TTL (0).
+func (k *Kernel) TTL() time.Duration {
+	return 0
+}
+
+// Family qualifies the address as ipv4 or ipv6
+func (k *Kernel) Family() int {
+	if k.Address().IP.To4() != nil {
+		return unix.AF_INET
+	}
+
+	return unix.AF_INET6
+}
+
+// Scope sets the address scope
+func (k *Kernel) Scope() uint8 {
+	return unix.RT_SCOPE_UNIVERSE
+}
+
+// Routes aggregates the specified routes for a given device configuration
+func (k *Kernel) Routes() (routes []*Route) {
+	return []*Route{}
+}
+
+// Resolvers returns the DNS resolvers
+func (k *Kernel) Resolvers() []net.IP {
+	resolvers := []net.IP{}
+
+	for _, resolver := range []string{k.PrimaryResolver, k.SecondaryResolver} {
+		if addr := net.ParseIP(resolver); addr != nil {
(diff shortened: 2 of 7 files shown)


internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if option == "dhcp" {
+		return fmt.Errorf("unsupported kernel.ip configuration method: %s", option)
+	}
+
+	// https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt
+	// ip=<client-ip>:<server-ip>:<gw-ip>:<netmask>:<hostname>:<device>:<autoconf>:<dns0-ip>:<dns1-ip>:<ntp0-ip>
+	r := strings.NewReader(strings.Join(strings.Split(option, ":"), " "))
+	_, err := fmt.Fscan(r,
```
andrewrynhard: `Fscan` will require that `ip` has the full set of options specified, otherwise we will get `EOF`.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	Device            string
+	Auto              string
+	PrimaryResolver   string
+	SecondaryResolver string
+	NTPServer         string
+}
+
+// Discover doesnt do anything in the static configuration since all
+// the necessary configuration data is supplied via config.
+func (s *Kernel) Discover(ctx context.Context) error {
```
andrewrynhard: Receiver name convention would call for `(k *kernel)`

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	var option *string
+	if option = kernel.ProcCmdline().Get("ip").First(); option == nil {
+		return fmt.Error("no kernel.ip argument supplied, skipping")
```
andrewrynhard: ` kernel.ip` is confusing. Something like `no ip argument supplied to kernel, skipping` is better IMO.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Address returns the IP address
+func (s *Kernel) Address() *net.IPNet {
+	return s.Address
+}
+
+// Mask returns the netmask.
+func (s *Kernel) Mask() net.IPMask {
+	netmask := net.ParseIP(s.Netmask).To4()
```
andrewrynhard: We support IPv6 `Family()`, we should support it here too.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	return resolvers
+}
+
+// Hostname returns the hostname
+func (s *Kernel) Hostname() string {
+	return strings.Split(s.Hostname, ".")[0]
+}
+
+// Link returns the underlying net.Interface that this address
+// method is configured for
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+func (s *Kernel) Scope() uint8 {
+	return unix.RT_SCOPE_UNIVERSE
+}
+
+// Routes aggregates the specified routes for a given device configuration
+func (s *Kernel) Routes() (routes []*Route) {
+	return routes
+}
+
+// Resolvers returns the DNS resolvers
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return unix.AF_INET6
+}
+
+// Scope sets the address scope
+func (s *Kernel) Scope() uint8 {
+	return unix.RT_SCOPE_UNIVERSE
+}
+
+// Routes aggregates the specified routes for a given device configuration
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return err
+}
+
+// Name returns back the name of the address method.
+func (s *Kernel) Name() string {
+	return "kernel"
+}
+
+// Address returns the IP address
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, resolver := range []string{s.PrimaryResolver, s.SecondaryResolver} {
+		if addr := net.ParseIP(resolver); addr != nil {
+			resolvers = append(resolvers, addr)
+		}
+	}
+
+	return resolvers
+}
+
+// Hostname returns the hostname
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/talos-systems/talos/internal/pkg/kernel"
+	"golang.org/x/sys/unix"
+)
+
+// Kernel implements the Addressing interface
```
andrewrynhard: nit: Sentence needs period. These get published to godocs.

internal/app/networkd/pkg/address/kernel.go
Ulexus: Significantly, though, if the kernel supports IPv6 spec via `ip=`, it will also require bracket notation, since the delimiter for both IPv6 addresses _and_ the `ip=` arguments is `:`.  However, the documentation for `ip=` is completely silent as regards IPv6.

Ulexus: According to this [Stack Overflow](https://serverfault.com/questions/445296/is-there-a-linux-kernel-boot-parameter-to-configure-an-ipv6-address), IPv6 is supported with the usual brackets.  Not exactly authoritative, but probably correct.

bradbeam: I don't think ipv6 is supported on mainline linux. 
`CONFIG_IP_PNP*` is what enables this support and it gets parsed here for ipv4 -- https://elixir.bootlin.com/linux/v5.3.11/source/net/ipv4/ipconfig.c#L1668 
I don't see anything similar for ipv6 ( ex, no PNP like variables in the kernel config -- https://elixir.bootlin.com/linux/v5.3.11/source/net/ipv6/Kconfig .

We could still choose to support this, but it'd be on us for the entire implementation.

internal/app/networkd/pkg/address/kernel.go
andrewrynhard: I guess it's not required, but we will need to check if err is `EOF`.

bradbeam:
need to address the punctuation still i guess, but i think from a technical perspective this is done. only support for ipv4 for this pr, can follow up later for ipv6

internal/pkg/kernel/kernel.go
```diff
@@ -243,10 +243,13 @@ func parse(parameters string) (parsed Parameters) {
 	return parsed
 }
 
+// CmdLine exposed for testing purposes
```
andrewrynhard: Not a big fan of leaking package vars into tests. An arg to the `func read` would be cleaner IMO.

internal/app/networkd/pkg/address/kernel_test.go
```diff
@@ -0,0 +1,99 @@
...
+	suite.Assert().Equal(kern.Family(), unix.AF_INET)
+	suite.Assert().Equal(kern.Address().IP, net.ParseIP("1.1.1.1"))
+	suite.Assert().Equal(kern.Mask().String(), "ffffff00")
+	suite.Assert().Equal(len(kern.Routes()), 0)
+	suite.Assert().Equal(kern.Hostname(), "hostname")
+	suite.Assert().Equal(kern.Resolvers()[0], net.ParseIP("4.4.4.4"))
+	suite.Assert().Equal(kern.Resolvers()[1], net.ParseIP("5.5.5.5"))
+}
+
+/*
```
andrewrynhard: Is this on purpose? We should remove it.

bradbeam:
Commented tests are there for discussion point around the kernel package. Because of the singleton approach, we can only ever parse this once during a run. What other options can we explore around the kernel package or these tests to be able to do multiple cases?

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,181 @@
...
+		case 6:
+			k.Auto = field
+		case 7:
+			k.PrimaryResolver = field
+		case 8:
+			k.SecondaryResolver = field
+		case 9:
+			k.NTPServer = field
+		}
+	}
```
andrewrynhard: If we pull this out as a separate function that returns `*option` (`option`) we can use `kerrnel.NewCmdline` in the tests instead of exposing `var Cmdline` at the package level. We don't have the linter on now, but public package level env vars would ding us because they are generally considered bad. See https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables for a good read on this.

internal/pkg/kernel/kernel.go
andrewrynhard: On second thought: https://github.com/talos-systems/talos/pull/1494/files#r346676068

internal/app/machined/internal/phase/network/network.go
```diff
@@ -60,11 +52,15 @@ func (task *InitialNetworkSetup) runtime(r runtime.Runtime) (err error) {
 	if r.Config() != nil {
 		log.Println("merging user defined network configuration")
 
-		if err = netconf.BuildOptions(r.Config()); err != nil {
+		if err = netconf.BuildOptionsFromConfig(r.Config()); err != nil {
 			log.Fatal(err)
 		}
 	}
 
+	if err = netconf.BuildOptionsFromKernel(); err != nil {
```
Ulexus: This order still strikes me as wrong:  options from config should override options from kernel.

internal/app/networkd/pkg/address/kernel.go
```diff
@@ -0,0 +1,181 @@
...
+	)
+
+	if option = kernel.ProcCmdline().Get("ip").First(); option == nil {
+		return fmt.Errorf("%s", "no kernel.ip argument supplied, skipping")
+	}
+
+	// TODO may need to move this earlier in networkd; if ip.dhcp is specified
+	// we shouldnt do anything because we dont know which (all?) interfaces the kernel
+	// will autoconfigure
+	if *option == "dhcp" {
```
Ulexus: If the option is DHCP, then we really should just fall back to our default behaviour, which is DHCP anyway.  I don't think that constitutes an error.

internal/app/networkd/pkg/networkd/netconf.go
```diff
@@ -45,3 +46,22 @@ func (n *NetConf) BuildOptions(config runtime.Configurator) error {
 
 	return nil
 }
+
+// BuildOptionsFromKernel translates the supplied config to functional options.
+func (n *NetConf) BuildOptionsFromKernel() error {
+	// Check to see if a kernel supplied configuration option was specified
+	kern := &address.Kernel{}
+	if err := kern.Discover(context.Background()); err != nil {
+		return nil
```
Ulexus: While I agree that failing to build options from the kernel should not constitute an error, I think the error determination should be processed inside the discovery, since we should have situations where we genuinely encounter an error which should be handled (or at least logged), such as for a parse failure of the `ip=` clause, and those situations which should not constitute an error, such as the non-existence of an `ip=` clause, or for where `ip=dhcp`, perhaps.

Thus, it would seem to me a better path would be to use a sentinel error (e.g. `ErrNoop`) or return the parsed object or even a boolean flag to indicate whether to continue building the options.

Ulexus:
Overall, I can't help but think that this is making things more complicated than necessary.  If the kernel gets an `ip=` argument, it will mostly do what it is supposed to do, and all we should do is stay out of its way (if kernelCmdline has ` ip=`, skip our own interface configuration).

The real issue is the DHCP server, since the kernel cannot do anything with that.  Instead of tying this into the basic network setup, it seems like a reasonable option would be to separately handle the generation of resolv.conf with a routine common to all scenarios:
  1. If we have DNS servers defined in the userdata, use that
  2. Else if we have DNS servers supplied in `/proc/net/pnp`, use that
  3. Otherwise, use reasonable defaults: (1.1.1.1, 8.8.8.8, and 2001:4860:4860::8888)

While @andrewrynhard had a good point with respect to DHCP-based items changing over time, we should _only_ be talking about bootstrapping here, not deployment (after which, we should have a real config to tell us what to do), so I don't think that really matters here.

Only in the case that we don't have a config do we care at all what the kernel `ip=` flag says, so it seems reasonable to fork at that point, to isolate the bootstrap condition.

internal/app/networkd/pkg/address/kernel.go
andrewrynhard: +1

andrewrynhard:
> Commented tests are there for discussion point around the kernel package. Because of the singleton approach, we can only ever parse this once during a run. What other options can we explore around the kernel package or these tests to be able to do multiple cases?

I left a comment on how we can do the testing better that should allow for multiple tests at once.

bradbeam:
Opened up a new PR to replace this - #1550 
Got a bit too heavy to rebase on recent changes.

PR #1499, opened 2019-11-13
docs: add docs command to osctl

andrewrynhard:
This allows osctl to generate documentation for itself.


andrewrynhard:
I can’t think of a good way to integrate this so that we enforce the docs to be up to date. For now we will have to generate them manually. Was thinking maybe we run the command and then check to see if the tree is dirty? In any case not something we have time for right now, as is should add immediate value.

PR #1502, opened 2019-11-14
test: add integration test for full boot sequence

andrewrynhard:
This adds an integration test that can be ran on a KVM enabled Linux
machine. It makes use of docker, matchbox, dnsmasq, libvirt, and HAproxy
to create an HA cluster.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ join.yaml
 docgen
 talosconfig
 kubeconfig
+hack/test/integration/matchbox/assets/*
+!hack/test/integration/matchbox/assets/.gitkeep
 
 # vim Swap
 [._]*.s[a-v][a-z]
--- a/hack/test/integration/README.md
+++ b/hack/test/integration/README.md
@@ -1,0 +1,93 @@
+# Integration Testing
+
+## Setup
+
+### Prerequisites
+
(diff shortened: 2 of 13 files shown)


cmd/osctl/cmd/install_linux.go
```diff
@@ -69,9 +69,9 @@ var installCmd = &cobra.Command{
 		}
 
 		sequence := runtime.None
-		if upgradeArg {
-			sequence = runtime.Upgrade
-		}
+		// if upgradeArg {
+		// 	sequence = runtime.Upgrade
+		// }
```
andrewrynhard: There is a bug that I haven't found yet that is making the initial install and upgrade sequence. We have an if statement in the etcd service that assumes if an upgrade has been performed, then it attempts to discover current control plane node IPs using the Kubernetes API. The problem is that it thinks an upgrade has been performed, so when attempting to discover the other members it fails.

smira:
:heart:  I'm super excited to have that!

cmd/osctl/cmd/install_linux.go
andrewrynhard: I plan on fixing in a follow up to keep this scoped as a test.

PR #1504, opened 2019-11-14
fix: honor the extraArgs option for the kubelet

andrewrynhard:
This allows users to supply extra arguments for the kubelet.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -92,7 +92,7 @@ func (e *Etcd) DependsOn(config runtime.Configurator) []string {
 
 // Runner implements the Service interface.
 func (e *Etcd) Runner(config runtime.Configurator) (runner.Runner, error) {
-	a, err := args(config)
+	a, err := e.args(config)
 	if err != nil {
 		return nil, err
 	}
@@ -318,12 +318,8 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial 
 	return initial, nil
 }
 
-func blacklistError(s string) error {
-	return fmt.Errorf("extra etcd arg %q is not allowed", s)
-}
-
 // nolint: gocyclo
-func args(config runtime.Configurator) ([]string, error) {
+func (e *Etcd) args(config runtime.Configurator) ([]string, error) {
 	hostname, err := os.Hostname()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -358,11 +354,11 @@ func args(config runtime.Configurator) ([]string, error) {
 		"peer-key-file":         constants.KubernetesEtcdPeerKey,
 	}
 
-	extraArgs := argsbuilder.Args(config.Cluster().Etcd().ExtraArgs())
+	extraArgs := argsbuilder.Args(config.Machine().Kubelet().ExtraArgs())
```
andrewrynhard: whoops

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -247,3 +226,43 @@ func (k *Kubelet) HealthSettings(runtime.Configurator) *health.Settings {
...
+	blackListArgs := argsbuilder.Args{
+		"bootstrap-kubeconfig":       constants.KubeletBootstrapKubeconfig,
+		"kubeconfig":                 constants.KubeletKubeconfig,
+		"container-runtime":          "remote",
+		"container-runtime-endpoint": "unix://" + constants.ContainerdAddress,
+		"anonymous-auth":             "false",
+		"cert-dir":                   "/var/lib/kubelet/pki",
+		"client-ca-file":             constants.KubernetesCACert,
+		"cni-conf-dir":               "/etc/cni/net.d",
+		"cluster-domain":             "cluster.local",
```
bradbeam: Why is cluster domain blacklisted?

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -318,12 +318,8 @@ func buildInitialCluster(config runtime.Configurator, name, ip string) (initial
 	return initial, nil
 }
 
-func blacklistError(s string) error {
-	return fmt.Errorf("extra etcd arg %q is not allowed", s)
-}
-
 // nolint: gocyclo
-func args(config runtime.Configurator) ([]string, error) {
+func etcdArgs(config runtime.Configurator) ([]string, error) {
```
smira: nit, but would it be better to make them more of methods on `Etcd` type to get them scoped under the service?

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Shouldn't be.

internal/app/machined/pkg/system/services/etcd.go
andrewrynhard: Yep!

PR #1505, opened 2019-11-14
fix: update documentation url

lexfrei:
old url looks like 404 now

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 ## Documentation
 
-For instructions on deploying and managing Talos, see the [Documentation](https://www.talos-systems.com/docs/talos).
+For instructions on deploying and managing Talos, see the [Documentation](https://www.talos.dev/docs/v0.3/).
 
 ## Community
 


andrewrynhard:
Thanks @lexfrei!

andrewrynhard:
@lexfrei can you amend the commit as `fix: update documentation url` please? We enforce conventional commits :)

andrewrynhard:
@lexfrei Try this 
```
git commit -s --amend -m 'fix: update documentation url' -m 'old url looks like 404 now'
```

lexfrei:
@andrewrynhard it's not easy to do this on phone :)
Is it done?

andrewrynhard:
@lexfrei It's not :) ... when your in front of a computer, there will be two things you need to do. We require 1 commit to keep the tree linear, and we require conventional commit format so that we can generate our CHANGELOG automatically. So you'll need to rebase and squash everything down to 1 commit that is in conventional commit format.

andrewrynhard:
@lexfrei I know its a bit of a pain, but do you mind amending and squashing down to 1 conventional commit?

andrewrynhard:
See https://github.com/talos-systems/talos/pull/1558.

PR #1506, opened 2019-11-14
test: implement node discovery for integration tests

smira:
This adds support for node discovery for API-based tests, but discovery
is based on k8s state. Discovery can be overridden if we provide a list
of node IPs as a flag.

Also adds a test for K8s API server version.

### DIFF
--- a/internal/integration/base/api.go
+++ b/internal/integration/base/api.go
@@ -34,6 +34,32 @@ func (apiSuite *APISuite) SetupSuite() {
 	apiSuite.Require().NoError(err)
 }
 
+// DiscoverNodes provides list of Talos nodes in the cluster.
+//
+// As there's no way to provide this functionality via Talos API, it works the following way:
+// 1. If there's a provided list of nodes, it's used.
+// 2. If integration test was compiled with k8s support, k8s is used.
+func (apiSuite *APISuite) DiscoverNodes() []string {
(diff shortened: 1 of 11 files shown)


internal/integration/base/api.go
```diff
@@ -34,6 +34,32 @@ func (apiSuite *APISuite) SetupSuite() {
...
+	discoveredNodes := apiSuite.TalosSuite.DiscoverNodes()
+	if discoveredNodes != nil {
+		return discoveredNodes
+	}
+
+	var err error
+
+	apiSuite.discoveredNodes, err = discoverNodesK8s(apiSuite.Client)
+	if err != nil {
+		apiSuite.Require().Error(err)
```
smira: :man_facepalming: should be `NoError()`, I'll get it fixed

smira:
e2e has errors, looking into that, believe it's not the test

smira:
I really didn't want it to be merged, as it might break the e2e tests

andrewrynhard:
Whoops. Sorry about that.

andrewrynhard:
@smira should I revert?

smira:
probably we can wait for the e2e run, I haven't found any recent one... or was I looking for the wrong label in Drone. If it keeps failing, I can send a PR to disable that test, other changes in the PR should be good. Let me anyway send a PR to skip this test for now.

PR #1507, opened 2019-11-14
chore: reverse order of events in `osctl service`

smira:
Now it shows events from last one down the oldest one.

Andrew, you wanted that :)


andrewrynhard:
:D Thanks @smira !

PR #1508, opened 2019-11-14
fix: make logging middleware first in the list, fix duration

smira:
Logging middleware should be the first one to log the request properly
including logging before proxy goes into action.

I had sec -> msec convertion wrong, but in the end I thought I should
replace it simply with `duration.String()` which is nicer.

### DIFF
--- a/pkg/grpc/middleware/log/log.go
+++ b/pkg/grpc/middleware/log/log.go
@@ -63,7 +63,7 @@ func (m *Middleware) UnaryInterceptor() grpc.UnaryServerInterceptor {
 			msg = err.Error()
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/grpc/middleware/log/log.go
```diff
@@ -63,7 +63,7 @@ func (m *Middleware) UnaryInterceptor() grpc.UnaryServerInterceptor {
 			msg = err.Error()
 		}
 
-		m.logger.Printf("%s [%s] %.3fms unary %s (%s)", code, info.FullMethod, duration.Seconds()/1000.0, msg, extractMetadata(ctx))
```
smira: should have been `* 1000.0` :man_facepalming:

PR #1518, opened 2019-11-16
fix: require arg length of 1 for kubeconfig command

andrewrynhard:
The kubeconfig command needs only one argument.


andrewrynhard:
> Should we default to . If args = 0?

I'm torn. I kinda like explicit, but 🤷‍♂ not sure I care too much either way.

PR #1519, opened 2019-11-16
refactor: rename version label

andrewrynhard:
Since we are standardizing on talos.dev everywhere else, we should align
version label with this.


andrewrynhard:
> All the cool kids use io

Once we own `talos.io` I would love to use it. But if people go to talos.io they may get confused.

PR #1522, opened 2019-11-18
feat(networkd): Add support for bonding

bradbeam:
This brings in support for creating bonded interfaces via networkd.


bradbeam:
Think `make fmt` may have pulled in some updates. Lots of random formatting changes. This should take care of most of the technical bits, will need to follow up with the actual orchestration bits.

andrewrynhard:
@bradbeam can we close this in favor of #1527?

PR #1528, opened 2019-11-24
feat: add read API

andrewrynhard:
This adds an API for reading arbitrary files.


api/machine/machine.proto
```diff
@@ -262,3 +263,7 @@ message LogsRequest {
   // driver might be default "containerd" or "cri"
   common.ContainerDriver driver = 3;
 }
+
+message ReadRequest {
+  string path = 1;
```
Ulexus: Not for the first-pass feature, but properties of `Offset` and `Limit` would be useful from an API perspective.

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -507,3 +508,32 @@ func getContainerInspector(ctx context.Context, namespace string, driver common.
...
+			Response: []*common.DataResponse{
+				{
+					Bytes: &common.Data{Bytes: b},
+				},
+			},
+		}
+
+		return reply, nil
+	default:
+		return nil, fmt.Errorf("path must be a file")
```
Ulexus: "regular file", perhaps?  As it stands, things like device nodes, sockets, etc what _are_ files will still get caught by this filter.

All of these are caught by this filter:
```
ModeType = ModeDir | ModeSymlink | ModeNamedPipe | ModeSocket | ModeDevice | ModeCharDevice | ModeIrregular
```

api/machine/machine.proto
andrewrynhard: Agree. Was debating if this should actually be a stream but yeah we can take a closer look in the future.

Ulexus: +1 to streaming _receive_, but this is the request side, anyway.  I suppose there is an argument to offering flow control via a streaming request side interface, but that's even further out, I think.

internal/app/machined/internal/api/reg/reg.go
andrewrynhard: Fixed

andrewrynhard:
> So I guess the question, then, is whether you want to make the receive be a streamed interface now or later, with consideration that this is in the public API. Without streaming, you run a bit of a risk with open-endedly-large files.

Proper thing probably is to stream. I'll do that.

smira:
:+1: With streaming API, one can probably read also some non-regular files as well.

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -507,3 +507,38 @@ func getContainerInspector(ctx context.Context, namespace string, driver common.
...
+		f, err := os.Open(in.Path)
+		if err != nil {
+			return err
+		}
+
+		ctx, cancel := context.WithCancel(context.Background())
+
+		defer cancel()
+
+		chunker := stream.NewChunker(f)
```
smira: that is not obvious, but `f` is going to be closed by `chunker`

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -507,3 +507,38 @@ func getContainerInspector(ctx context.Context, namespace string, driver common.
...
+
+		chunkCh := chunker.Read(ctx)
```
smira: that is more of a question for `stream.Chunker`, but there's no way to get read errors out of it, probably worth an issue to fix later, like providing extra `errCh` back

andrewrynhard: Yeah we can and should improve this later. We're doing a lot more streaming now.

internal/app/machined/internal/api/reg/reg.go
andrewrynhard: Hmm yeah, I'm not seeing _where_ though.

smira: should be https://github.com/talos-systems/talos/blob/master/pkg/chunker/stream/stream.go#L65

andrewrynhard: Ahh I'm using the file chunker. That one doesn't seem to close the file.

andrewrynhard: Should I use the file one? Been a while since I looked at these.

PR #1530, opened 2019-11-25
feat: upgrade packages

andrewrynhard:
This brings in the following changes:

- Linux 5.3.13
- Containerd 1.3.1


smira:
Please update https://github.com/talos-systems/talos/blob/master/pkg/constants/constants.go#L15 to match integration test

andrewrynhard:
Passed E2E: https://ci.dev.talos-systems.io/talos-systems/talos/5239

PR #1531, opened 2019-11-25
chore: disable all azure e2e temporarily

rsmitty:
This PR removes all refs to the azure e2e/integration/conformance tests
for now, since we need to wait on some upstream CAPI fixes and the test
is currently broken.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -912,24 +912,6 @@ steps:
   depends_on:
   - installer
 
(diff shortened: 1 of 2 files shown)


rsmitty:
This should be g2g now. Verified that we made it through an e2e run and all the way to sonobuoy on a conformance run. Should pass conformance when some upstream bootkube changes are landed.

PR #1532, opened 2019-11-25
feat: allow sysctl writes

andrewrynhard:
This adds the ability to write sysctls on boot by specifying them in the
machine config.


andrewrynhard:
Forced push dismissed Andrey's approval ... I only added better docs so gonna force merge.

PR #1539, opened 2019-11-26
feat: support force flag for osctl kubeconfig

rsmitty:
This PR allows you to specify the kubeconfig command like `osctl
kubeconfig . --force` or `osctl kubeconfig . -f` to delete the
kubeconfig file before we write it out.

### DIFF
--- a/cmd/osctl/cmd/kubeconfig.go
+++ b/cmd/osctl/cmd/kubeconfig.go
@@ -17,6 +17,8 @@ import (
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
 )
 
+var force bool
+
 // kubeconfigCmd represents the kubeconfig command
 var kubeconfigCmd = &cobra.Command{
 	Use:   "kubeconfig [local-path]",
@@ -55,12 +57,21 @@ Kubeconfig will be written to PWD/kubeconfig or [local-path]/kubeconfig if speci
 			if len(args) == 1 {
 				localPath = args[0]
 			}
+			localPath = filepath.Clean(localPath)
 
-			extractTarGz(filepath.Clean(localPath), r)
+			// Drop the existing kubeconfig before writing the new one if force flag is specified.
+			if force {
+				err = os.Remove(filepath.Join(localPath, "kubeconfig"))
+				if err != nil && !os.IsNotExist(err) {
+					helpers.Fatalf("error deleting existing kubeconfig: %s", err)
+				}
+			}
+			extractTarGz(localPath, r)
 		})
 	},
 }
 
 func init() {
+	kubeconfigCmd.Flags().BoolVarP(&force, "force", "f", false, "Force overwrite of kubeconfig if already present")
 	rootCmd.AddCommand(kubeconfigCmd)
 }


cmd/osctl/cmd/kubeconfig.go
```diff
@@ -56,11 +60,19 @@ Kubeconfig will be written to PWD/kubeconfig or [local-path]/kubeconfig if speci
 				localPath = args[0]
 			}
 
+			//Drop the existing kubeconfig before writing the new one if force flag is specified.
+			if force {
+				err = os.Remove(filepath.Clean(localPath + "/kubeconfig"))
```
smira: better to use `filepath.Join`

cmd/osctl/cmd/kubeconfig.go
```diff
@@ -56,11 +60,19 @@ Kubeconfig will be written to PWD/kubeconfig or [local-path]/kubeconfig if speci
...
+				if err != nil && !os.IsNotExist(err) {
+					helpers.Fatalf("error deleting existing kubeconfig: %s", err)
+				}
+			}
 			extractTarGz(filepath.Clean(localPath), r)
```
smira: I feel this whole thing better to be rewritten without `extractTarGz`, as this a bit of overkill for a single file

smira: we have https://github.com/talos-systems/talos/blob/master/cmd/osctl/pkg/client/client.go#L166 wrapper method which can be used to simply write down `kubeconfig` where appropriate (and hopefully one day merge it into user's kubeconfig)

cmd/osctl/cmd/kubeconfig.go
rsmitty: fixed this up a bit to clean up the localpath and then use filepath.Join if force is supplied.

cmd/osctl/cmd/kubeconfig.go
rsmitty: Seems fair, but sounds like it's a bit larger overhaul of this kubeconfig command. I'll throw it in the backlog.

cmd/osctl/cmd/kubeconfig.go
```diff
@@ -55,12 +57,21 @@ Kubeconfig will be written to PWD/kubeconfig or [local-path]/kubeconfig if speci
 			if len(args) == 1 {
 				localPath = args[0]
 			}
+			localPath = filepath.Clean(localPath)
 
-			extractTarGz(filepath.Clean(localPath), r)
+			// Drop the existing kubeconfig before writing the new one if force flag is specified.
+			if force {
+				err = os.Remove(filepath.Join(localPath, "/kubeconfig"))
```
andrewrynhard: super nit: the extra `/` in `filepath.Join(localPath, "/kubeconfig")` isn't needed.

cmd/osctl/cmd/kubeconfig.go
andrewrynhard: Wonder if the new read api would be better here? It supports streaming. But agree something for a follow-up.

cmd/osctl/cmd/kubeconfig.go
rsmitty: fixed, hopefully ready for merge now.

PR #1552, opened 2019-11-27
feat: use grpc-proxy in apid

smira:
This replaces codegen version of apid proxying with
talos-systems/grpc-proxy based version. Proxying is transparent, it
doesn't require exact information about methods and response types. It
requires some common layout response to enhance it properly with node
metadata or errors.

There should be no signifcant changes to the API with the previous
version, but it's worth mentioning a few changes:

1. grpc.ClientConn is established just once per upstream (either local
service or remote apid instance).

2. When called without `-t` (`targets`), apid proxies immediately down
to local service skipping proxying to itself (as before), which results
in empty node metadata in response (before it had local node IP). Might
revert this later to proxy to itself (?).

3. Streaming APIs are now fully supported with multiple targets, but
message definition doesn't contain `ResponseMetadata`, so streaming APIs
are broken now with targets (needs a fix).

4. Errors are now returned as responses with `Error` field set in
`ResponseMetadata`, this requires client library update and `osctl` to
handle it properly.


api/common/common.proto
```diff
@@ -5,8 +5,11 @@ package common;
 option go_package = "github.com/talos-systems/talos/api/common";
 
 // Common metadata message nested in all reply message types
-message NodeMetadata {
+message ResponseMetadata {
+  // hostname of the server response comes from (injected by proxy)
   string hostname = 1;
+  // error is set if request failed to the upstream (rest of response is undefined)
+  string error = 2;
```
smira: error from each upstream is embedded here (doesn't fail whole request)

api/common/common.proto
```diff
@@ -15,14 +18,22 @@ message Data {
...
-  NodeMetadata metadata = 1;
+  ResponseMetadata metadata = 1;
   Data bytes = 2;
 }
 
 message DataReply {
   repeated DataResponse response = 1;
 }
 
+message EmptyResponse {
```
smira: required for cut & paste on response proto message

cmd/osctl/cmd/version.go
```diff
@@ -46,6 +46,9 @@ var versionCmd = &cobra.Command{
 
 				if resp.Metadata != nil {
 					node = resp.Metadata.Hostname
+					if resp.Metadata.Error != "" {
```
smira: this is just for debugging to see actual error from backend, should be implemented on top level for real debugging

internal/app/apid/local.go
```diff
@@ -0,0 +1,77 @@
...
+	if lb.conn != nil {
+		return ctx, lb.conn, nil
+	}
+
+	var err error
+	lb.conn, err = grpc.DialContext(
+		ctx,
+		"unix:"+lb.SocketPath,
+		grpc.WithInsecure(),
+		grpc.WithCodec(proxy.Codec()), //nolint: staticcheck
```
smira: `.WithCodec()` is deprecated, but offered replacement is not equivalent

internal/app/apid/main.go
```diff
@@ -61,43 +61,35 @@ func main() {
...
-	timeClient, err := api.NewLocalTimeClient()
-	if err != nil {
-		log.Fatalf("time client: %v", err)
-	}
-
-	networkClient, err := api.NewLocalNetworkClient()
+	factoryTLSConfig, err = tls.New(
+		tls.WithClientAuthType(tls.Mutual),
+		tls.WithCACertPEM(ca),
+		tls.WithKeypair(*certs),
```
smira: this should be creds which are refreshing, but didn't have time to dig into it

internal/app/apid/proxy.go
```diff
@@ -0,0 +1,72 @@
...
+		backends[i] = backendFactory(target)
+	}
+
+	log.Printf("aggregateDirector: targets = %v", targets)
+
+	return proxy.One2Many, backends, nil
+}
+
+func streamedDetector(fullName string) bool {
+	return false
```
smira: TODO: implement me

smira:
I want to refactor it into sub-packages with some cleaner structure, without global variables, so this is just PoC for now -- it works for our requests.

Big changes with previous implementation:

1. `apid` can proxy APIs which it doesn't know about (with the exception it needs to know if api is streaming)
2. no protobuf marshal/unmarshal when proxying (performance)
3. `gprc.ClientConn`s are cached (this of course was possible with previous implementation), no handshake overhead, no lingering connections which are not closed (performance)

internal/app/apid/backend.go
```diff
@@ -0,0 +1,116 @@
...
+		grpc.WithTransportCredentials(ab.Creds),
+		grpc.WithCodec(proxy.Codec()), //nolint: staticcheck
+	)
+
+	log.Printf("got conn, %v %v", ab.conn, err)
+
+	return outCtx, ab.conn, err
+}
+
+// AppendInfo is called to enhance response from the backend with additional data.
```
smira: need better docs on what is going here

bradbeam: i _think_ i get what's going on here but it definitely feels like its in the realm of black wizardry,

internal/app/apid/factory.go
```diff
@@ -0,0 +1,34 @@
...
+	backendCache     sync.Map
+	factoryTLSConfig *tls.Config
+)
+
+func backendFactory(target string) proxy.Backend {
+	b, ok := backendCache.Load(target)
+	if ok {
+		return b.(proxy.Backend)
+	}
+
```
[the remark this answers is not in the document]
smira: it is being used now, but just in a hacky way, `factoryTLSConfig` above is initialized in main.go from the same provider, but I need to clean things up

internal/app/apid/backend.go
smira: New version has docstring & tests

smira:
New version: refactoring, split into packages, unit-tests for almost every piece.

TLS provider is left as it was, but it needs some :heart: , plus streaming APIs need a bit of love to make them work with targets.

cmd/osctl/cmd/cluster/pkg/node/node.go
```diff
@@ -86,8 +86,8 @@ func NewNode(clusterName string, req *Request) (err error) {
 
 	switch req.Type {
 	case generate.TypeInit:
-		var osdPort nat.Port
-		osdPort, err = nat.NewPort("tcp", "50000")
+		var apidPort nat.Port
+		apidPort, err = nat.NewPort("tcp", "50000")
```
smira: renamed this one, as `osdPort` was used for `apiPort`...

internal/app/apid/pkg/backend/apid.go
```diff
@@ -0,0 +1,197 @@
...
+func (a *APID) GetConnection(ctx context.Context) (context.Context, *grpc.ClientConn, error) {
+	origMd, ok := metadata.FromIncomingContext(ctx)
+
+	md := origMd.Copy()
+	delete(md, "targets")
+	delete(md, ":authority")
+
+	if ok {
+		md.Set("proxyfrom", origMd[":authority"]...)
+	} else {
```
bradbeam: Just to make sure i'm reading this right, the `else` condition would only be met if we had a context passed through that didn't include metadata?

internal/app/apid/pkg/backend/apid.go
```diff
@@ -0,0 +1,197 @@
...
+//
+// If we look at binary representation of any 'SomeReply' message, it will always contain one
+// protobuf field with field ID 1 (see above) and type 2 (embedded message SomeResponse is encoded
+// as string with length). So if we want to add fields to 'SomeResponse', we can simply read field
+// header, adjust length for new 'SomeResponse' representation, and prepend new field header.
+//
+// At the same time, we can add 'common.ResponseMetadata' structure to 'SomeResponse' by simply
+// appending or prepending 'common.ResponseMetadata' as a single field. This requires 'metadata'
+// field to be not defined in original response. (This is due to the fact that protobuf message
+// representation is concatenation of each field representation).
```
bradbeam: <3 thank you

internal/app/apid/pkg/backend/apid.go
```diff
@@ -0,0 +1,197 @@
...
+	if a.conn != nil {
+		return outCtx, a.conn, nil
+	}
+
+	var err error
+	a.conn, err = grpc.DialContext(
+		ctx,
+		fmt.Sprintf("%s:%d", a.target, constants.ApidPort),
+		grpc.WithTransportCredentials(a.creds),
+		grpc.WithCodec(proxy.Codec()), //nolint: staticcheck
```
bradbeam: not sure if you saw, 
```
// WithCodec returns a DialOption which sets a codec for message marshaling and
// unmarshaling.
//
// Deprecated: use WithDefaultCallOptions(ForceCodec(_)) instead.  Will be
// supported throughout 1.x.
```
https://godoc.org/google.golang.org/grpc#ForceCodec

internal/app/apid/pkg/backend/apid.go
smira: yep, I thought we should include something anyways. I know `:authority` should always be there, but just to be 100% sure to make proxied request.

internal/app/apid/pkg/backend/apid.go
smira: I tried that, but it doesn't work, this seems to affect outgoing calls iirc, while `WithCodec` affects data decoding on input to the server.

internal/app/apid/pkg/director/director.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if _, exists := md["proxyfrom"]; exists {
+		return r.localDirector(fullMethodName)
+	}
+
+	var targets []string
+
+	if targets, ok = md["targets"]; !ok {
+		// send directly to local node, skips another layer of proxying
```
smira: this is something I would like to highlight.

In previous implementation, if `targets` MD field was missing, request was routed to another `apid` with target set as `md[:authority]`. I changed this behavior because `:authority` is kind of `Host:` HTTP header, it is set by the caller, so it should LB endpoint. So request goes like: `client -> LB -> apid [host1] -> LB -> apid [host2]`. In this example `host1` != `host2`. This seems to be a bit confusing path. And `apid` might not be able to talk back to the LB (why not?). Plus it requires valid TLS client cert to be present on `apid` instance (to talk back to itself), which might be tricky in failure modes (cert not renewed, etc). So it might block ability to access one node directly.

But proxying though once again when `targets` was not set allowed to set host metadata for the response, but it was set to `:authority` value, which once again might not be the node (it might be LB endpoint). So after this change if we do `osctl` without `-t`, `NODE:` comes back blank. If we do `osctl -t <targets>`, `NODE:` is set properly.

tl;dr

Now `apid` if `targets` is not set behaves like `osd` before - proxies directly to instances on the same node. If `targets` key is set, it forwards to `apid` instances.

Same goes with response wrapping, error handling, if `targets` is not set, errors are returned directly, response is passed as is, etc. (like `osd` before).

Thoughts?

smira:
update: fixed streaming methods, and while I was at it, made all the responses consistent (except for `LS` which is obviously different). 

now all the APIs work correctly, but there's a lot of cleanups on CLI/client side to handle errors from multi-target responses, I would like to take that as separate PR as this seems to be already big

internal/app/apid/pkg/director/director.go
andrewrynhard: As we discussed on slack. I think this is a good change.

PR #1554, opened 2019-11-29
fix: provide a way for client TLS config to use Provider

smira:
In `tls.Config`, there are two hooks for getting certificate for client
and server config. So we need separate configuration methods to
configure them both.

Required in apid to provide refreshing TLS client cert to
grpc.ClientConn.


smira:
Updated to use refreshing provider for apid client certificates

PR #1555, opened 2019-12-01
chore: fix conformance

andrewrynhard:
The `--e2e-parallel` flag seems to skip all tests when running in
certified-conformance mode. This reverts that change, and also adds a
check that fails if the conformance tests do not pass. This ensures that
we are not publishing broken versions of our edge release.

### DIFF
--- a/hack/test/conformance.sh
+++ b/hack/test/conformance.sh
@@ -25,8 +25,8 @@ e2e_run "set -eou pipefail
             --wait \
             --skip-preflight \
             --plugin e2e \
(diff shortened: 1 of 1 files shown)


hack/test/conformance.sh
```diff
@@ -25,8 +25,8 @@ e2e_run "set -eou pipefail
             --wait \
             --skip-preflight \
             --plugin e2e \
-            --plugin-env e2e.E2E_USE_GO_RUNNER=true \
```
andrewrynhard: This is the default as of 1.16.1

PR #1571, opened 2019-12-03
chore: rewrite basic integration in go instead of bash

rsmitty:
This PR will be the start of several. It rewrites the basic integration
in go. We'll do these one at a time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,8 +280,8 @@ container: buildkitd
 	@docker load < build/$@.tar
 
 .PHONY: basic-integration
-basic-integration:
-	@TAG=$(TAG) ./hack/test/$@.sh
+basic-integration: gitmeta
+	@TAG=$(TAG) SHA=$(SHA) go run ./internal/test-framework/main.go basic-integration
 
 .PHONY: capi
 capi:
--- a/.drone.yml
+++ b/.drone.yml
@@ -368,7 +368,7 @@ steps:
 
 - name: basic-integration
   pull: always
-  image: autonomy/build-container:latest
+  image: golang:1.13
   commands:
   - make basic-integration
   environment:
@@ -823,7 +823,7 @@ steps:
 
 - name: basic-integration
   pull: always
-  image: autonomy/build-container:latest
+  image: golang:1.13
   commands:
   - make basic-integration
   environment:
@@ -1427,7 +1427,7 @@ steps:
 
 - name: basic-integration
   pull: always
-  image: autonomy/build-container:latest
+  image: golang:1.13
   commands:
   - make basic-integration
   environment:
@@ -2061,7 +2061,7 @@ steps:
 
 - name: basic-integration
   pull: always
-  image: autonomy/build-container:latest
+  image: golang:1.13
   commands:
   - make basic-integration
   environment:
@@ -2695,7 +2695,7 @@ steps:
 
 - name: basic-integration
   pull: always
-  image: autonomy/build-container:latest
+  image: golang:1.13
   commands:
   - make basic-integration
   environment:
--- a/hack/drone.jsonnet
+++ b/hack/drone.jsonnet
@@ -152,7 +152,7 @@ local markdownlint = Step("markdownlint");
 local image_test = Step("image-test", depends_on=[installer]);
 local unit_tests = Step("unit-tests", depends_on=[rootfs]);
 local unit_tests_race = Step("unit-tests-race", depends_on=[lint]);
-local basic_integration = Step("basic-integration", depends_on=[container, osctl_linux, integration_test]);
+local basic_integration = Step("basic-integration", image="golang:1.13", depends_on=[container, osctl_linux, integration_test]);
 
 local coverage = {
   name: 'coverage',
--- a/internal/test-framework/cmd/basic-integration.go
+++ b/internal/test-framework/cmd/basic-integration.go
@@ -1,0 +1,247 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package cmd
+
+import (
+	"context"
+	"log"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strconv"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/mount"
+	"github.com/docker/docker/client"
+	"github.com/spf13/cobra"
+
+	"github.com/talos-systems/talos/internal/test-framework/internal/pkg/runner"
+	"github.com/talos-systems/talos/pkg/constants"
+)
+
+const (
+	timeoutSeconds = 300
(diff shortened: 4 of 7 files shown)


rsmitty:
Should be ready for nit picking now. Seems to be working as expected for basic-integration and e2e.

Makefile
```diff
@@ -280,8 +280,8 @@ container: buildkitd
 	@docker load < build/$@.tar
 
 .PHONY: basic-integration
-basic-integration:
-	@TAG=$(TAG) ./hack/test/$@.sh
+basic-integration: talos-test
+	@TAG=$(TAG) $(BINDIR)/talos-test basic-integration
```
andrewrynhard: Instead of building it, I maybe it might be better to just `go run`? That way we don't have to add the build to drone at all.

hack/test/cmd/basic-integration.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/spf13/cobra"
+	"github.com/talos-systems/talos/hack/test/internal/pkg/constants"
+	"github.com/talos-systems/talos/hack/test/internal/pkg/runner"
+)
+
+var cleanup bool
+var talosImage = "docker.io/autonomy/talos:"
+
+// Add basic-integration command
+var bi = &cobra.Command{
```
andrewrynhard: The naming convention used by cobra would call for this to be `basicIntegrationCmd`. Was this created with `cobra add`?

hack/test/cmd/root.go
```diff
@@ -0,0 +1,21 @@
...
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	rootCmd = &cobra.Command{
+		Use:   "talos-test",
+		Short: "A small cli for talos testing",
```
andrewrynhard: nit: `A CLI for performing tests`. 1 We already know its Talos, 2 "small" isn't that useful :) ... nits I know.

hack/test/internal/pkg/constants/constants.go
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package constants
+
+const KubeVersion = "1.16.2"
+const KubeImage = "k8s.gcr.io/hyperkube:v" + KubeVersion
```
andrewrynhard: I think we should source these from Talos constants. One source of truth for these values will be nice to have. Today when we need to update Kubernetes we have a number of different places we need to change. I'd like to get to a point where we only need to change 1 variable.

Dockerfile
```diff
@@ -356,6 +356,20 @@ RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=amd64 go test -
...
+ARG TAG
+ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
+COPY ./hack/test /src/hack/test
+WORKDIR /src/hack/test
+RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=amd64 go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Client -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG}" -o talos-test .
+RUN chmod +x ./talos-test
+
+FROM scratch AS talos-test
+COPY --from=talos-test-build /src/hack/test/talos-test /talos-test
+
```
andrewrynhard: I think for these tests (basic integration, e2e, and conformance) we should just use `go run`. The integration test is a bit different since we will publish it.

hack/test/cmd/basic-integration.go
rsmitty: nah, I just threw this together manually. but I'll rename for sure.

Makefile
rsmitty: had the same thought and it makes my life easier anyways. will move this over.

rsmitty:
on the linting, do you think we should move this out of hack? into a /test dir or something?

hack/test/internal/pkg/constants/constants.go
rsmitty: Fixed in amended commit.

Makefile
rsmitty: fixed in amended commit.

Dockerfile
rsmitty: fixed in amended commit.

internal/talos-test/internal/pkg/runner/runner.go
```diff
@@ -0,0 +1,121 @@
...
+	if inspect.State.ExitCode != 0 {
+		return errors.New("container exited with non-zero code")
+	}
+
+	return nil
+}
+
+// CommandInContainerWithTimeout is a wrapper around commandInContainer that retries until we hit the timeout value
+func CommandInContainerWithTimeout(ctx context.Context, client *client.Client, runnerConfig *ContainerConfigs, command string, timeout int) error {
+	for i := 0; i < timeout; i += 10 {
```
andrewrynhard: Maybe we should use the `retry` package for this?

internal/talos-test/cmd/root.go
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: I have one last nit on this. I think the `talos` portion in the name here is redundant since the code is already part of the Talos repo. But we can change in a follow up.

internal/talos-test/talos-test.go
```diff
@@ -0,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: nit: This should be `main.go`

internal/talos-test/internal/pkg/runner/runner.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// CommandInContainer simply runs a bash command in the hyperkube containers
+// nolint: gocyclo
+func CommandInContainer(ctx context.Context, client *client.Client, runnerConfig *ContainerConfigs, command string) error {
```
smira: it's not about this specific PR, but rather a general question.

We do equivalent of `docker run` many times, which has some good overhead. 

Does it make more sense to do `docker run` once with container which sleeps forever, and use `docker exec` to run each command?

andrewrynhard: ^ probably a good idea

internal/talos-test/talos-test.go
rsmitty: done.

internal/talos-test/cmd/root.go
rsmitty: renamed to test-framework

internal/talos-test/internal/pkg/runner/runner.go
rsmitty: 100% something we should do. throwing that into the backlog.

PR #1575, opened 2019-12-03
feat: support output directory for osctl config generate

### LINKED ISSUE #1509: Add support for `osctl config generate` output path
It would be useful if we could output all generated files into `--output X`

rsmitty:
This PR adds a `--output-dir/-o` flag to osctl config generate and will
attempt to create the directory path and write the generated files to
that location. Will close #1509.

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -28,6 +29,7 @@ var (
 	kubernetesVersion string
 	installDisk       string
 	installImage      string
+	outputDir         string
 )
 
 // configCmd represents the config command.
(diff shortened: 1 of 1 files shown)


cmd/osctl/cmd/config.go
```diff
@@ -138,6 +140,20 @@ var configGenerateCmd = &cobra.Command{
...
+	var err error
+	if outputDir == "" {
+		outputDir, err = os.Getwd()
+		if err != nil {
+			helpers.Fatalf("failed to get working dir: %w", err)
+		}
+	}
+
+	// Create dir path, ignoring "already exists" messages
+	if err = os.MkdirAll(outputDir, os.ModePerm); err != nil && !os.IsExist(err) {
```
andrewrynhard: We may need to invert this. `!os.IsExist(err)` doesn't guarantee that it does indeed exist. This is one of the caveats with `os.IsExist`. There could be a valid error, maybe something like permission denied or something like that.

andrewrynhard: Whoops. I messed up. I was turned upside down.

PR #1577, opened 2019-12-04
fix: return a unique set of errors on retry failure

### LINKED ISSUE #1496: Improve retry library error message
When the retry function fails due to timeout we only get `"timeout"` in the error message. We need to do better than this. Knowing _why_ the logic in the retry function is failing is super useful.

andrewrynhard:
This makes errors more descriptive in the case that a timeout occurs
when using a retry func. Today the error is simply "timeout", but that
doesn't provide anything useful to the user. It now returns a unique set of
errors that are collected through the duration of the retry.

Closes #1496.

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -5,7 +5,9 @@
 package retry
 
 import (
+	"fmt"
 	"math/rand"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/retry/retry.go
```diff
@@ -24,6 +26,42 @@ type Ticker interface {
...
+	if e.errs == nil {
+		e.errs = make(map[string]error)
+	}
+
+	if _, ok := e.errs[err.Error()]; !ok {
+		e.errs[err.Error()] = err
+	}
+
+	return e
+}
```
andrewrynhard: I wonder if we should use a slice to preserve order.

pkg/retry/retry.go
```diff
@@ -24,6 +26,42 @@ type Ticker interface {
...
+
+func (e *ErrorSet) Error() string {
+	if len(e.errs) == 0 {
+		return ""
+	}
+
+	errString := fmt.Sprintf("%d error(s) occurred:", len(e.errs))
+	for _, err := range e.errs {
+		errString = fmt.Sprintf("%s\n%s", errString, err)
```
smira: should be `+=` ?

pkg/retry/retry.go
```diff
@@ -24,6 +26,42 @@ type Ticker interface {
...
+
+	mu sync.Mutex
+}
+
+func (e *ErrorSet) Error() string {
+	if len(e.errs) == 0 {
+		return ""
+	}
+
+	errString := fmt.Sprintf("%d error(s) occurred:", len(e.errs))
```
smira: missing `\n` ?

pkg/retry/retry.go
smira: > should be `+=` ?

please disregard, being blind

pkg/retry/retry.go
smira: > missing `\n` ?

missed that one as well below

pkg/retry/retry.go
smira: I'm not sure if order matters after deduplication

andrewrynhard: SGTM

PR #1578, opened 2019-12-04
fix: Add hostname setting to networkd

bradbeam:
This moves all the hostname aggregation and setting into networkd so we can
get a correct response.


bradbeam:
Looks like I still have some work to do on this.

internal/app/machined/internal/phase/rootfs/etc/etc.go
```diff
@@ -136,33 +73,28 @@ func OSRelease() (err error) {
...
 }
 
-func ip() string {
-	addrs, err := net.InterfaceAddrs()
-	if err != nil {
-		return ""
+// etcFiles creates a common way to create a writable source file with a bind
+// mounted destination. This is most commonly used for well known files under
+// /etc that need to be adjusted during startup.
+func etcFiles(src, dst string) (err error) {
```
andrewrynhard: Can we rename this to something more descriptive? Maybe `createBindMountedFile`?

internal/app/machined/internal/phase/rootfs/network_configuration.go
```diff
@@ -34,5 +34,10 @@ func (task *NetworkConfiguration) runtime(r runtime.Runtime) (err error) {
 		return err
 	}
 
+	// Create /etc/hosts
+	if err = etc.Hosts(); err != nil {
```
andrewrynhard: Do we need to create this twice? Just above here we do it in `internal/app/machined/internal/phase/rootfs/hostname.go`.

internal/app/machined/internal/phase/rootfs/etc/etc.go
```diff
@@ -127,7 +64,7 @@ func OSRelease() (err error) {
 
 	tmpl, err := template.New("").Parse(osReleaseTemplate)
 	if err != nil {
-		return
+		return err
```
smira: that's where it becomes confusing imho: if we have function result declared as `(err error)`, what does `tmpl, err := ` do ? I believe it creates another `err` which shadows previous `err`, so bare `return` returns wrong `err`

internal/app/machined/internal/phase/rootfs/etc/etc.go
```diff
@@ -136,33 +73,28 @@ func OSRelease() (err error) {
...
 	}
 
-	for _, address := range addrs {
-		if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-			if ipnet.IP.To4() != nil {
-				return ipnet.IP.String()
-			}
-		}
+	// nolint: errcheck
+	defer f.Close()
```
smira: I believe we don't really need to `defer` here, as we can close immediately?
```suggestion
	if err = f.Close(); err != nil {
	  return
	}
```

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -54,3 +53,39 @@ func writeResolvConf(resolvers []string) error {
...
+		Alias    string
+	}{
+		IP:       address.String(),
+		Hostname: hostname,
+		Alias:    strings.Split(hostname, ".")[0],
+	}
+
+	tmpl, err := template.New("").Parse(hostsTemplate)
+	if err != nil {
+		return
```
smira: same comment as above on var shadowing, idk, I would probably just remove named result variable

internal/app/machined/internal/phase/rootfs/etc/etc.go
bradbeam: should be easy enough to change. i just did a blind copy paste of that bit, whoops.

PR #1579, opened 2019-12-04
chore: support image specification in drone step function

rsmitty:
This PR adds the ability to override the image in our default build step
function in jsonnet. Needed so we can override the image easily for
basic integration tests.


rsmitty:
note that the .drone.yml file isn't changed. that's to be expected, since I haven't made use of this function in this PR. It'll change with the basic-integration PR.

PR #1581, opened 2019-12-04
feat: use containerd-shim-runc-v2

### LINKED ISSUE #1529: Use containerd-shim-runc-v2

andrewrynhard:
This configures the CRI containerd to use containerd-shim-runc-v2.

Closes #1529.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -206,29 +206,29 @@ COPY --from=osctl-darwin-build /osctl-darwin-amd64 /osctl-darwin-amd64
 # The kernel target is the linux kernel.
 
 FROM scratch AS kernel
-COPY --from=docker.io/autonomy/kernel:ed3b1fd /boot/vmlinuz /vmlinuz
-COPY --from=docker.io/autonomy/kernel:ed3b1fd /boot/vmlinux /vmlinux
+COPY --from=docker.io/autonomy/kernel:a70af2a /boot/vmlinuz /vmlinuz
+COPY --from=docker.io/autonomy/kernel:a70af2a /boot/vmlinux /vmlinux
 
 # The rootfs target provides the Talos rootfs.
 
 FROM build AS rootfs-base
-COPY --from=docker.io/autonomy/fhs:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/ca-certificates:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/containerd:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/dosfstools:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/eudev:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/iptables:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/libressl:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/libseccomp:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/musl:ed3b1fd / /rootfs
-COPY --from=docker.io/autonomy/runc:ed3b1fd / /rootfs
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -239,8 +239,7 @@ COPY images/networkd.tar /rootfs/usr/images/
 # symlinks to avoid accidentally cleaning them up.
 COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
 RUN cleanup.sh /rootfs
-COPY hack/containerd.toml /etc/containerd.toml
-COPY hack/containerd.toml /etc/containerd-system.toml
+COPY hack/containerd.toml /rootfs/etc/containerd/cri.toml
```
andrewrynhard: Come to find out, we haven't been writing this file to the correct path so the config was never used. Things have been working fine without it so I don't think we need to configure the system containerd in any special way. This changes it so that we only configure the CRI containerd differently.

internal/app/machined/pkg/system/services/containerd.go
```diff
@@ -61,7 +61,7 @@ func (c *Containerd) Runner(config runtime.Configurator) (runner.Runner, error)
 			"--address",
 			constants.ContainerdAddress,
 			"--config",
-			"/etc/containerd.toml",
+			"/etc/containerd/cri.toml",
```
andrewrynhard: The default is `/etc/containerd/config.toml`, so I decided to drop it into `/etc/containerd` and call it "cri" because that is its' role in Talos. It is the containerd used as the CRI. In v0.4 I'd like to change our system services names to `cri` and `containerd`.

internal/app/machined/pkg/system/services/system-containerd.go
```diff
@@ -59,8 +59,6 @@ func (c *SystemContainerd) Runner(config runtime.Configurator) (runner.Runner, e
 			"--address", constants.SystemContainerdAddress,
 			"--state", "/run/system/containerd",
 			"--root", "/run/system/lib/containerd",
-			"--config",
-			"/etc/containerd-system.toml",
```
andrewrynhard: We have unknowingly been using the defaults, so no need for the config anymore. If we ever find a need for it we can always bring it back in pretty easily.

internal/app/machined/pkg/system/services/containerd.go
bradbeam: Does this still make sense when considering the use case for exposing `containerd` configuration options via user data? We've had some requests about adding in things like insecure registries and whatnot to this config. 
I guess the name itself doesn't matter too much, but might be a bit confusing for someone wanting to inspect the configuration ( via osctl ls/read/cp ) and seeing something other than the default.

hack/containerd.toml
```diff
@@ -1 +1,2 @@
-subreaper = true
```
bradbeam: why remove `subreaper=true`?

andrewrynhard: We have been running without it for a long time now, which shows that our reaper in `machined` is working. This config was never really used. We were copying it into the wrong location.

internal/app/machined/pkg/system/services/containerd.go
andrewrynhard: I absolutely think it makes sense. :) It is more confusing IMO to have containerd and system-containerd. We really only use container as the CRI when it comes to Kubernetes.

Dockerfile
smira: I think because we had reaper before system/cri containerd split

andrewrynhard: Exactly

andrewrynhard:
I know there were concerns in slack on the naming of `/etc/containerd/cri.toml` but since there is no longer a system specific config, we will go with `cri.toml` to avoid any ambiguity on whether or not the config is for `system-containerd` or `containerd`. The config is for the containerd instance who's purpose it is to serve as the CRI within Talos.

PR #1583, opened 2019-12-04
fix: provide peer remote address for 'NODE': as default in osctl

smira:
This change is pretty mechanical, just wrap every API so that remote
peer address is used as default for `resp.Metadata.Hostname`.

This makes `NODE:` non-empty in all the API calls.

### DIFF
--- a/cmd/osctl/cmd/containers.go
+++ b/cmd/osctl/cmd/containers.go
@@ -14,7 +14,9 @@ import (
 
 	criconstants "github.com/containerd/cri/pkg/constants"
 	"github.com/spf13/cobra"
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 
 	"github.com/talos-systems/talos/api/common"
 	osapi "github.com/talos-systems/talos/api/os"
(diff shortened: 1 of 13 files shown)


smira:
After I made all the changes, I thought that there might be better way to change the client API. Instead of returning `remotePeer` from each API call, we could allow passing `options... grpc.CallOption` to all the `client` package methods and this way caller (`osctl`) might use it to pass `grpc.Peer()` option. On one hand, `osctl` needs peer info for every call, on other hand allowing options is more generic. Thoughts?

smira:
Forgot to mention: while I was it, I fixed `osctl service <id>` command to display service info, as it had some broken append() logic in the loops

smira:
I skipped `Dmesg` as I'm going to redo it as streaming API, so skipping work which is going to be wasted (streaming multi-node version will be different anyways).

smira:
> After I made all the changes, I thought that there might be better way to change the client API. Instead of returning `remotePeer` from each API call, we could allow passing `options... grpc.CallOption` to all the `client` package methods and this way caller (`osctl`) might use it to pass `grpc.Peer()` option. On one hand, `osctl` needs peer info for every call, on other hand allowing options is more generic. Thoughts?

decision in the channel: use generic `callOptions ...grpc.Option`

PR updated

PR #1587, opened 2019-12-04
fix: extract errors from API response

smira:
This PR only touches `Version` method, but I will expand it to other
methods in the next PR.

When proxying to many upstreams, errors are wrapped as responses as we
can't return error and response from grpc call. Reflect-based function
was introduced to filter out responses which contain errors as
multierror. Reflection was used, as each response is a different Go
type, and we can't write a generic function for it.

osctl was updated to support having both resp & err not nil. One failed
response shouldn't result in error.

Re-enabled integration test for multiple targets and version
consistency, need e2e validation.


smira:
I thought I should also handle edge case when all the responses are errors better, but I will leave it for tomorrow

smira:
e2e pass with version tests for all cluster nodes enabled: https://ci.dev.talos-systems.io/talos-systems/talos/5557/1/30

PR #1589, opened 2019-12-05
fix: reverse preference order of network config

### LINKED ISSUE #1588: Bonding breaks when `ip=` kernel cmdline present
Two things are causing bonding to break when the kernel is supplied with an `ip=` flag:
  - kernel config overrides file config
  - `bond0` is the first non-local interface in the device enumeration

Since the first of those should never be the case, I would propose reversing that.

Ulexus:
Kernel config should always play second to a file-based config.

Fixes #1588


Ulexus:
It is perhaps overzealous to say this _fixes_ 1588, since the way things are, if there is no file-based configuration of the network but there _is_ a kernel `ip=` flag, things will not work.  But that would be a pretty silly setup.

PR #1592, opened 2019-12-05
refactor: extract TLS bits from apid main.go

smira:
No functional changes, just moving code around.

### DIFF
--- a/internal/app/apid/pkg/provider/tls.go
+++ b/internal/app/apid/pkg/provider/tls.go
@@ -1,0 +1,85 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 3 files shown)


internal/app/apid/pkg/provider/tls.go
```diff
@@ -0,0 +1,85 @@
...
+	tlsConfig.certificateProvider, err = tls.NewRemoteRenewingFileCertificateProvider(
+		config.Machine().Security().Token(),
+		endpoints,
+		constants.TrustdPort,
+		hostname,
+		ips,
+	)
+	if err != nil {
+		return nil, err
+	}
```
bradbeam: nit, but is it worth dropping this check and only doing `return tlsConfig, err`  below?

smira: idk, I always feel uncomfortable returning a structure (`tlsConfig`) which won't work as it's not initialized properly.

not a big deal either way, but tried to stand on either return value or error

PR #1594, opened 2019-12-05
feat: add ability to append to existing files with extrafiles

### LINKED ISSUE #1467: Add ability to add self-signed CAs to /etc/ssl/certs/ca-certificates.crt

rsmitty:
This PR introduces "op" to the extra files options. This allows for a
user to specify "append" as the op, which will create a copy of the file
specified, add the extra data provided, and bind mount over the existing
file if the file is not in the /var path. Will close #1467

### DIFF
--- a/internal/app/machined/internal/phase/config/extra_files.go
+++ b/internal/app/machined/internal/phase/config/extra_files.go
@@ -5,11 +5,14 @@
 package config
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
+	"golang.org/x/sys/unix"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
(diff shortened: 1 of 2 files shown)


internal/app/machined/internal/phase/config/extra_files.go
```diff
@@ -32,15 +35,59 @@ func (task *ExtraFiles) runtime(r runtime.Runtime) (err error) {
...
+			}
 		}
 	}
 
 	return result.ErrorOrNil()
 }
+
+// slurpFile simplyu reads in a file and returns a byte slice
+func slurpFile(path string) ([]byte, error) {
+	file, err := ioutil.ReadFile(path)
```
smira: isn't this function just `ioutil.ReadFile()` - it does nothing else?

rsmitty: yep - that's fair. I thought it was going to be more complex than it was :)

will pull that up into the other code

PR #1597, opened 2019-12-05
feat: add support for `osctl logs -f`

smira:
Now default is not to follow the logs (which is similar to `kubectl logs`).

Integration test was added for `Logs()` API and `osctl logs` command.


internal/app/machined/pkg/system/log/log.go
```diff
@@ -70,12 +69,6 @@ func (l *Log) Close() error {
 	return l.source.Close()
 }
 
-// Read implements chunker.Chunker.
-func (l *Log) Read(ctx context.Context) <-chan []byte {
```
smira: gc'ed unused method, and it was dangerous, as it used same `*os.File` as for writing to the logs, while `chunker` might still Seek the stream

pkg/chunker/file/file.go
```diff
@@ -31,6 +32,13 @@ func Size(s int) Option {
 	}
 }
 
+// Follow file updates using inotify().
+func Follow() Option {
```
andrewrynhard: nit: The convention for our functional opts has been to prepend `With` (`WithFollow`). Not a blocker for me.

pkg/chunker/file/file.go
```diff
@@ -68,36 +76,39 @@ func (c *File) Read(ctx context.Context) <-chan []byte {
...
+				return
+			}
 		}
 
 		buf := make([]byte, c.options.Size)
 
 		for {
 			for {
-				n, err := c.source.ReadAt(buf, offset)
+				n, err := c.source.Read(buf)
```
smira: figured out `ReadAt()` was not that necessary, as there was a `.Seek()` which anyway resets file offset.

So now chunker doesn't `.Seek()` and uses regular `.Read()`, this way it will be easier to combine with any kind of "tail" implementation which sets seek offset before passing file to this method.

The way we call chunker is that we always separate `*os.File` instance for the duration of the API call.

pkg/chunker/file/file.go
```diff
@@ -107,7 +118,7 @@ func (c *File) Read(ctx context.Context) <-chan []byte {
 					select {
 					case <-ctx.Done():
 						return
-					case event := <-watcher.Events:
+					case event := <-watcherEvents:
```
smira: read from `nil` channel blocks forever, so it works as if watcher is not enabled

pkg/chunker/file/file.go
smira: I can rename both options, as above there's `Size()` as well

andrewrynhard: Ahh missed that one.

PR #1601, opened 2019-12-05
feat: add create and overwrite file operations

andrewrynhard:
Explicit over implicit is the reasoning here. I think it will be more
clear to users this way.

### DIFF
--- a/docs/website/content/v0.3/en/configuration/v1alpha1.md
+++ b/docs/website/content/v0.3/en/configuration/v1alpha1.md
@@ -242,6 +242,10 @@ install:
 #### files
 
 Allows the addition of user specified files.
+The value of `op` can be `create`, `overwrite`, or `append`.
+In the case of `create`, `path` must not exist.
+In the case of `overwrite`, and `append`, `path` must be a valid file.
+If an `op` value of `append` is used, the existing file will be appended.
 Note that the file contents are not required to be base64 encoded.
 
 Type: `array`
@@ -249,11 +253,12 @@ Type: `array`
 Examples:
 
 ```yaml
-kubelet:
-  contents: |
-    ...
-  permissions: 0666
-  path: /tmp/file.txt
+files:
+  - content: |
+      ...
+    permissions: 0666
+    path: /tmp/file.txt
+    op: append
 
 ```
 
--- a/internal/app/machined/internal/phase/config/extra_files.go
+++ b/internal/app/machined/internal/phase/config/extra_files.go
@@ -31,12 +31,30 @@ func (task *ExtraFiles) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 	return task.runtime
 }
 
+// nolint: gocyclo
 func (task *ExtraFiles) runtime(r runtime.Runtime) (err error) {
 	var result *multierror.Error
 
 	for _, f := range r.Config().Machine().Files() {
-		// Slurp existing file if append is our op and add contents to it
-		if f.Op == "append" {
+		content := f.Content
+
+		switch f.Op {
+		case "create":
+			if err = doesNotExists(f.Path); err != nil {
+				result = multierror.Append(result, fmt.Errorf("file must not exist: %q", f.Path))
+				continue
+			}
+		case "overwrite":
+			if err = existsAndIsFile(f.Path); err != nil {
+				result = multierror.Append(result, err)
+				continue
+			}
+		case "append":
+			if err = existsAndIsFile(f.Path); err != nil {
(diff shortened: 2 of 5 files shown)


docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -242,18 +242,20 @@ install:
...
-  contents: |
-    ...
-  permissions: 0666
-  path: /tmp/file.txt
+files:
+  - contents: |
+      ...
+    permissions: 0666
+    path: /tmp/file.txt
+    op: [write,append]
```
andrewrynhard: Originally started with this, and then found out that some error handling was needed around the value of `op`, then realized explicit is probably better than implicit.

smira: is this a list, or this is supposed to be single value?

andrewrynhard: Single value ... I see the confusion ... I was trying to make the example use the "optional" notation.

andrewrynhard: Fixed

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -125,14 +125,16 @@ type MachineConfig struct {
 	MachineInstall *InstallConfig `yaml:"install,omitempty"`
 	//   description: |
 	//     Allows the addition of user specified files.
+	//     If an `op` value of `append` is used, the existing file will be appended.
```
smira: I guess we need to document that `op` can be one of: `write`, `append`, as `write` is not in the docs?

andrewrynhard: Done

internal/app/machined/internal/phase/config/extra_files.go
```diff
@@ -31,12 +31,34 @@ func (task *ExtraFiles) TaskFunc(mode runtime.Mode) phase.TaskFunc {
...
+		case "create":
+			if err = existsAndIsFile(f.Path); err == nil {
+				return fmt.Errorf("file must not exist: %q", f.Path)
+			}
+		case "overwrite":
+			if err = existsAndIsFile(f.Path); err != nil {
+				return err
+			}
+
+			if err = os.Remove(f.Path); err != nil {
```
smira: can we remove from readonly rootfs? or we can just bind mount on top of that?

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -125,14 +125,19 @@ type MachineConfig struct {
...
 	//     Note that the file contents are not required to be base64 encoded.
 	//   examples:
 	//     - |
-	//       kubelet:
-	//         contents: |
-	//           ...
-	//         permissions: 0666
-	//         path: /tmp/file.txt
+	//       files:
+	//         - contents: |
```
smira: should be `content` now?

internal/app/machined/internal/phase/config/extra_files.go
andrewrynhard: We cannot. Was hoping that when we attempt to remove that the error would bubble up and it would be clear to the user. Should we verify the path?

pkg/config/types/v1alpha1/v1alpha1_types.go
andrewrynhard: yes, thanks

internal/app/machined/internal/phase/config/extra_files.go
andrewrynhard: Oh I see. Your saying don't remove, and just let the logic after the `switch` to its' thing.

smira: yep, as it will bind mount overwriting it effectively I guess in any case

internal/app/machined/internal/phase/config/extra_files.go
```diff
@@ -82,3 +103,22 @@ func (task *ExtraFiles) runtime(r runtime.Runtime) (err error) {
 
 	return result.ErrorOrNil()
 }
+
+func existsAndIsFile(p string) (err error) {
+	var info os.FileInfo
+
+	info, err = os.Stat(p)
```
bradbeam: this logic is a bit off to satisfy both create and overwrite requests

andrewrynhard: fixed

PR #1604, opened 2019-12-06
fix: error reporting in `osctl kubeconfig`

smira:
Problem seems to be on multiple levels, and there are a bit of changes
which got mixed in from another PR (just same file changed).

Core of the issue is that `helpers.Fatalf()` calls `os.Exit()` which
terminates execution and doesn't let the `defer` and other handlers to
run. This uses Cobra feature of error propagation to pop errors through
the stack back to root command.

### DIFF
--- a/cmd/osctl/cmd/cp.go
+++ b/cmd/osctl/cmd/cp.go
@@ -79,7 +79,9 @@ captures ownership and permission bits.`,
 				}
 			}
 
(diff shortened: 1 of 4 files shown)


smira:
found a bug, please don't merge...

smira:
> found a bug, please don't merge...

bug fixed, should be good now

PR #1605, opened 2019-12-06
fix: make retry errors ordered

andrewrynhard:
We need deterministic ordering for retry errorss

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -28,7 +28,7 @@ type Ticker interface {
 
 // ErrorSet represents a set of unique errors.
 type ErrorSet struct {
-	errs map[string]error
+	errs []error
 
 	mu sync.Mutex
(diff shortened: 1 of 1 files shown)


pkg/retry/retry.go
```diff
@@ -52,11 +52,20 @@ func (e *ErrorSet) Append(err error) error {
...
-		e.errs = make(map[string]error)
+		e.errs = []error{}
 	}
 
-	if _, ok := e.errs[err.Error()]; !ok {
-		e.errs[err.Error()] = err
+	ok := false
+
+	for _, existingErr := range e.errs {
+		if err == existingErr {
```
smira: should we compare `err.Error() == existingErr.Error()` ?

as errors might different by pointer, but same by value? (if say, generated with `fmt.Errrof()`)

PR #1607, opened 2019-12-06
feat: allow ability to specify custom CNIs

### LINKED ISSUE #1593: Add remote option for CNI

rsmitty:
This PR will allow users to specify one or many URLs for CNI so that
they can bypass bootkube deploying flannel and bring their own. Will
close #1593


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -327,6 +330,35 @@ func generateAssets(config runtime.Configurator) (err error) {
 		return err
 	}
 
+	// If "none" is the CNI, we expect the user to supply one or more urls that point to CNI yamls
+	if config.Cluster().Network().CNI().Name() == "none" {
+		ctx := context.Background()
+
+		for _, url := range config.Cluster().Network().CNI().URLs() {
+			urlExploded := strings.Split(url, "/")
+			fileName := urlExploded[len(urlExploded)-1]
```
smira: there's a nice chance it ends up empty if URL looks like `http://host/some/dir/`

idk if `path.Base()` might do better job here

smira: plus it might be nice to filter out `.` and `..` and other friends... or once `filepath.Clean()`, fail if `fileName` becomes empty

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -510,9 +510,17 @@ type EtcdConfig struct {
 type ClusterNetworkConfig struct {
 	//   description: |
 	//     The CNI used.
-	//   values:
-	//     - flannel
-	CNI string `yaml:"cni"`
+	//     Composed of "name" and "url".
+	//     URL is only used if name == "remote".
```
andrewrynhard: `"custom"`

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -510,9 +510,17 @@ type EtcdConfig struct {
...
+	//     URL should point to a single yaml file that will get deployed.
```
andrewrynhard: I think this needs to be updated since its a slice now, and not a single string.

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -510,9 +510,17 @@ type EtcdConfig struct {
...
+	//     Empty struct or any other name will default to bootkube's flannel.
+	//   examples:
+	//     - |
+	//       cni:
+	//         name: "none" //only supports upstream bootkube options [flannel, calico, none]
```
andrewrynhard: I think this should be spelled out in the main docs block.

PR #1611, opened 2019-12-06
test: add retries to the test which verifies cluster version

smira:
It fails on AWS, need to figure out if it's transient failure or not.

While I was there, found lots of small bugs when endpoint is
unresponsive, or target nodes are unresponsive and fixed them.

In retry formatting added `\t` so that embedded errors are better
aligned in the output (same as multierror).

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -40,7 +40,7 @@ func (e *ErrorSet) Error() string {
 
 	errString := fmt.Sprintf("%d error(s) occurred:", len(e.errs))
 	for _, err := range e.errs {
-		errString = fmt.Sprintf("%s\n%s", errString, err)
+		errString = fmt.Sprintf("%s\n\t%s", errString, err)
 	}
 
 	return errString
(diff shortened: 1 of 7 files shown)


pkg/retry/retry.go
```diff
@@ -40,7 +40,7 @@ func (e *ErrorSet) Error() string {
 
 	errString := fmt.Sprintf("%d error(s) occurred:", len(e.errs))
 	for _, err := range e.errs {
-		errString = fmt.Sprintf("%s\n%s", errString, err)
+		errString = fmt.Sprintf("%s\n\t%s", errString, err)
```
andrewrynhard: This may need an update to the tests.

smira: oh yeah, totally forgot, thanks, I'll get it updated

pkg/retry/retry.go
```diff
@@ -124,12 +124,20 @@ func (t ticker) Stop() {
 // ExpectedError error represents an error that is expected by the retrying
 // function. This error is ignored.
 func ExpectedError(err error) error {
+	if err == nil {
```
smira: added this as shortcut, so that we could do `return ExpectedError(someFunc())` without caring about return status in `RetryableFunc`

PR #1620, opened 2019-12-07
chore: make the CNI configurable in local KVM test

andrewrynhard:
This changes the local KVM setup to use a configurable CNI URL.


andrewrynhard:
> Because the cilium version changes fairly frequently, I would _consider_ pulling the version or URL out as a top-level variable.

Better yet, I made it configurable.

PR #1621, opened 2019-12-08
feat: allow configurable SANs for API

### LINKED ISSUE #1512: Support DNS in Talos API path
Everything from `osctl` to `apid` should be able to use DNS names.

andrewrynhard:
This adds the ability to supply additional SANs for apid and trustd.

Closes #1512.

### DIFF
--- a/internal/app/apid/pkg/provider/tls.go
+++ b/internal/app/apid/pkg/provider/tls.go
@@ -9,7 +9,6 @@ import (
 	stdlibtls "crypto/tls"
 	"fmt"
 	stdlibnet "net"
-	"os"
 
 	"github.com/talos-systems/talos/internal/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/constants"
@@ -28,16 +27,18 @@ func NewTLSConfig(config runtime.Configurator, endpoints []string) (*TLSConfig, 
 	if err != nil {
 		return nil, fmt.Errorf("failed to discover IP addresses: %w", err)
 	}
-	// TODO(andrewrynhard): Allow for DNS names.
+
+	dnsNames, err := net.DNSNames()
+	if err != nil {
+		return nil, err
+	}
+
 	for _, san := range config.Machine().Security().CertSANs() {
 		if ip := stdlibnet.ParseIP(san); ip != nil {
 			ips = append(ips, ip)
+		} else {
+			dnsNames = append(dnsNames, san)
 		}
-	}
-
-	hostname, err := os.Hostname()
-	if err != nil {
-		return nil, fmt.Errorf("failed to discover hostname: %w", err)
 	}
 
 	tlsConfig := &TLSConfig{}
@@ -46,7 +47,7 @@ func NewTLSConfig(config runtime.Configurator, endpoints []string) (*TLSConfig, 
 		config.Machine().Security().Token(),
 		endpoints,
 		constants.TrustdPort,
-		hostname,
+		dnsNames,
 		ips,
 	)
 	if err != nil {
--- a/internal/app/trustd/main.go
+++ b/internal/app/trustd/main.go
@@ -8,7 +8,6 @@ import (
 	"flag"
 	"log"
 	stdlibnet "net"
-	"os"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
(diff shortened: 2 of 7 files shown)


internal/app/trustd/main.go
```diff
@@ -56,20 +58,34 @@ func main() {
 		log.Fatal(err)
 	}
 
+	domainName, err := ioutil.ReadFile("/proc/sys/kernel/domainname")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	dnsNames := []string{}
```
Ulexus: Should _also_ include the short hostname in the set.

andrewrynhard: 👍 done

internal/app/apid/pkg/provider/tls.go
```diff
@@ -28,25 +30,39 @@ func NewTLSConfig(config runtime.Configurator, endpoints []string) (*TLSConfig,
...
+	if strings.Contains(string(domainName), "(none)") {
+		var hostname string
+
+		if hostname, err = os.Hostname(); err != nil {
+			return nil, err
+		}
+
+		dnsNames = append(dnsNames, hostname, fmt.Sprintf("%s.%s", hostname, domainName))
+	}
+
```
andrewrynhard: I think this should be in a helper func, but not quite sure _where_ to put it.

Ulexus: It does kind of seem like something that might want to be reused, like say for the kubernetes side.  I'd classify it as network.

andrewrynhard: Yeah, I mean I'm already using it twice in this PR.

Ulexus: Not only that, but you might one day want to hook it up to a cloud provider's host metadata lookup for the hostname and/or IP.

andrewrynhard: good point

pkg/net/net.go
```diff
@@ -66,3 +70,24 @@ func NthIPInNetwork(network *net.IPNet, n int) (net.IP, error) {
...
+	if strings.Contains(string(domainName), "(none)") {
+		var hostname string
+
+		if hostname, err = os.Hostname(); err != nil {
+			return nil, err
+		}
+
+		dnsNames = append(dnsNames, hostname, fmt.Sprintf("%s.%s", hostname, domainName))
+	}
+
```
bradbeam: don't we need to handle the case where domainName is not `none`?

bradbeam: Eh I guess I misread the logic // comment.

andrewrynhard: Not sure I follow. Although this could be a stale review since I just pushed changes. Shouldn't `if string(domainName) != "(none)\n"` cover what your saying?

bradbeam: Want to change the conditional to be
```
if string(domainName) == "(none)\n" {
  return dnsNames, nil
}

var hostname ...
```

andrewrynhard: Yep, much better.

pkg/net/net.go
```diff
@@ -66,3 +69,26 @@ func NthIPInNetwork(network *net.IPNet, n int) (net.IP, error) {
 
 	return nil, errors.New("network does not contain enough IPs")
 }
+
+// DNSNames returns a default set of machine names. It includes the hostname,
+// and FQDN. If the kernel domain name is not set, no names will be returned.
+func DNSNames() ([]string, error) {
```
bradbeam: super nit, feel free to ignore

should we change to named return types here and clean up the variable declaration?

```
func DNSNames()  (dnsNames []string, err error) {
  var (
    domainName string
    hostname string
  )
 ...
```

bradbeam: feel like it helps make the function a bit more concise.

andrewrynhard: 👍 works for me.

pkg/net/net.go
```diff
@@ -66,3 +69,46 @@ func NthIPInNetwork(network *net.IPNet, n int) (net.IP, error) {
 
 	return nil, errors.New("network does not contain enough IPs")
 }
+
+// DNSNames returns a default set of machine names. It includes the hostname,
+// and FQDN. If the kernel domain name is not set, no names will be returned.
```
andrewrynhard: I'm wondering if we should do this now that I'm looking at it. If a domain name is not found, I don't think that we should ignore hostname as this code is now. Thoughts?

andrewrynhard: I went with sticking the hostname in there even if the domain name isn't set. I can change back if we decide we don't want to do this.

pkg/net/net.go
```diff
@@ -66,3 +69,51 @@ func NthIPInNetwork(network *net.IPNet, n int) (net.IP, error) {
...
+func DomainName() (domainname string, err error) {
+	var b []byte
+
+	if b, err = ioutil.ReadFile("/proc/sys/kernel/domainname"); err != nil {
+		return "", err
+	}
+
+	domainname = string(b)
+
+	if domainname == "(none)\n" {
```
Ulexus: I don't have a better idea, unfortunately, but this feels awfully brittle.

andrewrynhard: yeah its ugly as hell, but apparentlythis is the default value from the kernel.

PR #1627, opened 2019-12-09
feat: add domain line to resolv.conf

### LINKED ISSUE #1626: use domainname for target suffix
If we have a domainname on a node and we are given a target (node) which is an unresolvable name, attempt to append the domainname before giving up.

This is hardly a high-priority request, but merely a convenience to have Talos's apid proxy act a little more like standard domain resolution, by attempting to resolve the hostname by appending the domainname (exactly as a `domain` line in `resolv.conf` would do).

In fact, it would probably be sufficient to add a `domain` line to the local `resolv.conf`, and it is likely Go's `net.Dial` will just do the right thing.

Ulexus:
If we have a domainname, add it to resolv.conf.

Fixes #1626

### DIFF
--- a/internal/app/networkd/pkg/networkd/misc.go
+++ b/internal/app/networkd/pkg/networkd/misc.go
@@ -12,6 +12,8 @@ import (
 	"net"
 	"strings"
 	"text/template"
+
+	talosnet "github.com/talos-systems/talos/pkg/net"
 )
 
 // filterInterfaceByName filters network links by name so we only mange links
@@ -46,6 +48,14 @@ func writeResolvConf(resolvers []string) (err error) {
 		if _, err = resolvconf.WriteString(fmt.Sprintf("nameserver %s\n", resolver)); err != nil {
 			log.Println("failed to add some resolver to resolvconf:", resolver)
 			return err
+		}
+	}
+
+	if domain, err := talosnet.DomainName(); err == nil {
+		if domain != "" {
+			if _, err = resolvconf.WriteString(fmt.Sprintf("search %s\n", domain)); err != nil {
+				return fmt.Errorf("failed to add domain search line to resolvconf: %s", err)
+			}
 		}
 	}
 


bradbeam:
Would you mind using the search key instead of domain name?

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -12,6 +12,9 @@ import (
 	"net"
 	"strings"
 	"text/template"
+
+	"github.com/pkg/errors"
```
andrewrynhard: We moved to standard library wrapping.

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -49,6 +52,14 @@ func writeResolvConf(resolvers []string) (err error) {
 		}
 	}
 
+	if domain, err := talosnet.DomainName(); err == nil {
+		if domain != "" {
+			if _, err = resolvconf.WriteString(fmt.Sprintf("domain %s\n", domain)); err != nil {
+				return errors.Wrap(err, "failed to add domain line to resolvconf")
```
andrewrynhard: Can we make this `fmt.Errorf("failed to add domain line to resolvconf: %w", err)`

Ulexus:
Well, we could certainly use the search key, but this _is_ what the domain key is for.  Is there a reason for that preference?

internal/app/networkd/pkg/networkd/misc.go
Ulexus: Ah, yeah.  Wishing he would update his library to use stdlib with the same API.  Oh well.

andrewrynhard:
I'm kinda impartial but my 2 cents: Since `domain` essentially gives us the same thing as `search`, and it aligns with the kernel domain name, I would lean towards using `domain`.

bradbeam:
I was under the impression domain was mutually exclusive with search, but according to this man page domain is obsolete 
http://man7.org/linux/man-pages/man5/resolv.conf.5.html

andrewrynhard:
> I was under the impression domain was mutually exclusive with search, but according to this man page domain is obsolete
> http://man7.org/linux/man-pages/man5/resolv.conf.5.html

Good find. That changes my mind :).

Ulexus:
Likewise...  My local man page doesn't have the same wording, so I guess _it_ is obsolete.

andrewrynhard:
@Ulexus Wanna move the PR out of draft. @bradbeam You LGTM in this?

Ulexus:
Yes, just testing first

Ulexus:
Yep, works perfectly to fix #1626

PR #1629, opened 2019-12-09
fix: append domainname to DHCP-sourced hostname

### LINKED ISSUE #1628: DHCP client should tie in domain name
If we receive a domain name from the dhcp server, we should append it to the hostname received.

Ulexus:
If we receive a domain name from the DHCP server, append it to the
received hostname.

Fixes #1628

### DIFF
--- a/internal/app/networkd/pkg/address/dhcp.go
+++ b/internal/app/networkd/pkg/address/dhcp.go
@@ -6,6 +6,7 @@ package address
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -129,7 +130,14 @@ func (d *DHCP) Resolvers() []net.IP {
 func (d *DHCP) Hostname() string {
 	// Truncate the returned hostname to only return
 	// the actual host entry
-	return strings.Split(d.Ack.HostName(), ".")[0]
+	hostname := strings.Split(d.Ack.HostName(), ".")[0]
```
bradbeam: since we do this split later, should we return d.ack.hostname in full if d.ack.domainname == ""?

Ulexus: Good question.  That sounds reasonable.

PR #1633, opened 2019-12-09
docs: simplify corporate proxy

alex1989hu:
It is possible to append to file since 84354c59414, hence no need to create custom Docker image


andrewrynhard:
Thanks @alex1989hu !

PR #1634, opened 2019-12-09
fix: kill POD network mode pods first on upgrades

andrewrynhard:
When we upgrade a node, we kill off all pods before performing a fresh
install. The issue with this is that we run the risk of killing the CNI
pod before we finish killing all other pods, leaving the CRI unable to
teardown the pod's networking. This works around that by first killing
any pods running without host networking so that the CNI can do its'
job, and then removing the remaining pods.

### DIFF
--- a/internal/app/machined/internal/phase/kubernetes/remove_all_pods.go
+++ b/internal/app/machined/internal/phase/kubernetes/remove_all_pods.go
@@ -6,11 +6,9 @@ package kubernetes
 
 import (
 	"context"
-	"log"
 	"time"
 
-	"golang.org/x/sync/errgroup"
-	"k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system"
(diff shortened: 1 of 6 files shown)


internal/app/machined/internal/phase/kubernetes/remove_all_pods.go
```diff
@@ -47,6 +47,25 @@ func (task *RemoveAllPods) standard() (err error) {
...
+	// CNI, preventing the CRI from cleaning up the pod's netwokring.
+
+	if err = RemovePodsWithNetworkMode(client, runtimeapi.NamespaceMode_POD); err != nil {
+		return err
+	}
+
+	// With the POD network mode pods out of the way, we kill the remaining
+	// pods.
+
+	if err = RemovePodsWithNetworkMode(client, runtimeapi.NamespaceMode_NODE); err != nil {
```
smira: should we kill really all the remaining pods here? (ignoring network namespace altogether?)

smira: I mean in the future there might be more modes available, or something else...

andrewrynhard: We should, just didn't know a good way to indicate that all modes should be killed. Maybe `-1` or something?

smira: One easy way imho is to use pointer receiver and skip the check if value is `nil`, but probably passing pointer to value would be challenging

internal/pkg/cri/pods.go
```diff
@@ -79,3 +81,92 @@ func (c *Client) PodSandboxStatus(ctx context.Context, podSandBoxID string) (*ru
...
+
+		networkMode := status.GetLinux().GetNamespaces().GetOptions().GetNetwork()
+
+		if networkMode != mode {
+			continue
+		}
+
+		g.Go(func() error {
+			if err := remove(ctx, c, pod); err != nil {
+				return err
```
smira: looks like could be `return remove(ctx, c, pod)`

plus it needs to wrap variables, as `pod` is `for` loop variable

andrewrynhard: A bit further up we use `pod := pod // https://golang.org/doc/faq#closures_and_goroutines`. A little trick I picked up not too long ago.

smira: oh, missed that, sorry for the noise!

internal/app/machined/internal/phase/kubernetes/remove_all_pods.go
andrewrynhard: Went with making `mode` variadic. If there are no modes then all modes are valid.

internal/pkg/cri/pods.go
andrewrynhard: Moved to using `return remove(ctx, c, pod)`.

PR #1639, opened 2019-12-09
chore: Remove increased timeouts for dhcp addressing.

bradbeam:
These timeouts were initially increased to handle long times for links to be ready. I think
with the updated link ready check in networkd these timers are unnecessary.


bradbeam:
As a note for followup when there's another use case for this. We may want to expose these timers as kernel options in the off chance they're needed ( since we'll want them available for the initial network configuration to pull user data ).

PR #1640, opened 2019-12-09
feat: rename confusing target options, --endpoints, etc.

### LINKED ISSUE #1610: Rename target/targets in talosconfig/osctl
In `talosconfig`, get rid of `Target` field and command `config target`. In general, it's not safe to change target to any node, as some commands (e.g. `kubeconfig`) work only against master nodes. At the same time it's easy to forget that `target` was changed in config and try to debug failures coming from it.

Plus, `config target` is confusing with `--target` flag which in fact do two completely different things.

*Proposal*: drop `talosconfig.Target`, introduce `talosconfig.Endpoints []string` which list all the master endpoints. Talos client should put them all to `grpc.ClientConn` for round-robin/failover.

Add flag `-e/--endpoint <endpoint>` to `osctl` to change the endpoint for the call (if we want to hit a specific node).

Rename `--target` flag to `--nodes`. (Probably makes sense to rename grpc metadata to `nodes` as well to stay consistent).

smira:
Fixes #1610

1. In `talosconfig`, deprecate `Target` in favor of `Endpoints`
(client-side LB to come next).

2. In `osctl`, use `--nodes` in place of `--target`.

3. In `osctl` add option `--endpoints` to override `Endpoints` for the
call.

Other changes are just updates to catch up with the changes. Most
probably I missed something... And CAPI provider needs update.


cmd/osctl/pkg/client/config/config.go
```diff
@@ -18,12 +18,26 @@ type Config struct {
...
+}
+
 // Context represents the set of credentials required to talk to a target.
 type Context struct {
-	Target string `yaml:"target"`
-	CA     string `yaml:"ca"`
-	Crt    string `yaml:"crt"`
-	Key    string `yaml:"key"`
+	DeprecatedTarget string   `yaml:"target,omitempty"` // Field deprecated in favor of Endpoints
+	Endpoints        []string `yaml:"endpoints"`
```
andrewrynhard: How difficult would it be to add a `Nodes []string` field here? It would be awesome if we didn't have to type out which nodes we want to target every time.

smira: default to empty? and command to set them?

smira: added, acts as default value for `--nodes`

PR #1643, opened 2019-12-09
refactor(networkd): Update bond config parameters to align with kernel

bradbeam:
This updates the yaml keys for the bonding configuration to align with the kernel documentation.
This mostly adds in underscores to the keys.

### DIFF
--- a/pkg/config/machine/machine.go
+++ b/pkg/config/machine/machine.go
@@ -83,32 +83,32 @@ type Device struct {
 // bonded interface.
 type Bond struct {
 	Interfaces      []string `yaml:"interfaces"`
-	ARPIPTarget     []string `yaml:"arpiptarget"`
+	ARPIPTarget     []string `yaml:"arpIPTarget"`
 	Mode            string   `yaml:"mode"`
-	HashPolicy      string   `yaml:"hashpolicy"`
-	LACPRate        string   `yaml:"lacprate"`
-	ADActorSystem   string   `yaml:"adactorsystem"`
-	ARPValidate     string   `yaml:"arpvalidate"`
-	ARPAllTargets   string   `yaml:"arpalltargets"`
+	HashPolicy      string   `yaml:"xmitHashPolicy"`
+	LACPRate        string   `yaml:"lacpRate"`
+	ADActorSystem   string   `yaml:"adActorSystem"`
+	ARPValidate     string   `yaml:"arpValidate"`
+	ARPAllTargets   string   `yaml:"arpAllTargets"`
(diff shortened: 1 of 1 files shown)


Ulexus:
Lgtm

andrewrynhard:
I only have one nit on this, but just wanted it to be clear. Our convention everywhere else has been to use camel case. These would be the only fields not following that convention. I'll leave it up to you guys what we wanna do. I can go either way, but since beta is being dropped either today or Wednesday, I want to make sure we are certain about this.

Also, @bradbeam will a `make docs` be needed? Or will the examples in the docs need updating?

bradbeam:
I'm not sure I have a strong feeling one way or another. 

Re: docs, no docs get generated for changes to the data types in machine config, so nothing to do here. It's something we should document, but it's not automated.

Ulexus:
I would be fine with camel case, I think, so long as the casing changes match with the underscore changed of the kernel.  That would result in expectable names which map well.  My problem was mainly with a few slightly-differently-named items and items which were all run together multiple words.

bradbeam:
whatyoudontlikewhenallthewordsaretogetherandyourenotsurewhereonewordbeginsandtheotherends?ithinkitsjustswell

Ulexus:
I suppose if I were a little more ancient, I would have been used to it, but I'm just used to modern conveniences, I suppose.

andrewrynhard:
Sounds good. @bradbeam can we make these camel case then? Gracias.

pkg/config/machine/machine.go
```diff
@@ -83,32 +83,32 @@ type Device struct {
...
 	Primary         string   `yaml:"primary"`
-	PrimaryReselect string   `yaml:"primaryreselect"`
-	FailOverMac     string   `yaml:"failovermac"`
-	ADSelect        string   `yaml:"adselect"`
+	PrimaryReselect string   `yaml:"primaryReselect"`
+	FailOverMac     string   `yaml:"failOverMac"`
+	ADSelect        string   `yaml:"adSelect"`
 	MIIMon          uint32   `yaml:"miimon"`
 	UpDelay         uint32   `yaml:"updelay"`
 	DownDelay       uint32   `yaml:"downdelay"`
```
andrewrynhard: Not sure what the kernel version is, but should these be `upDelay` and `downDelay`?

andrewrynhard: Looks like these are good 👍

Ulexus: Yeah, for whatever reason, the kernel options run those together as single words.

PR #1644, opened 2019-12-09
feat: add security hardening settings

andrewrynhard:
This pulls in an update from our bootkube fork that adds security
hardening to the control plane. The following was changed:

- API server now uses an EncryptionConfig for encrypting secrets
- API server now has an audit policy
- Profiling was disabled on all control plane components
- PodSecurityPolicy is enabled
- API server TLS cipher suites were set to the recommended ciphers by CIS


smira:
I guess it's https://github.com/talos-systems/bootkube/pull/2

PR #1649, opened 2019-12-11
feat: add config nodes command

andrewrynhard:
This adds a `osctl config nodes` command that updates the talosconfig
with the specified nodes.

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -41,9 +41,10 @@ var configCmd = &cobra.Command{
 
 // configEndpointCmd represents the config endpoint command.
 var configEndpointCmd = &cobra.Command{
-	Use:   "endpoint <endpoint>...",
-	Short: "Set the endpoint(s) for the current context",
-	Long:  ``,
+	Use:     "endpoint <endpoint>...",
+	Aliases: []string{"endpoints"},
+	Short:   "Set the endpoint(s) for the current context",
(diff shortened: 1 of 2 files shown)


cmd/osctl/cmd/config.go
```diff
@@ -60,6 +61,28 @@ var configEndpointCmd = &cobra.Command{
 	},
 }
 
+// configNodeCmd represents the config node command.
+var configNodeCmd = &cobra.Command{
+	Use:     "node <endpoint>...",
```
smira: should be "<node>" or "<address>"? as we don't allow port there?

cmd/osctl/cmd/root.go
```diff
@@ -107,7 +107,7 @@ func Execute() {
 
 	rootCmd.PersistentFlags().StringVar(&talosconfig, "talosconfig", defaultTalosConfig, "The path to the Talos configuration file")
 	rootCmd.PersistentFlags().StringVar(&cmdcontext, "context", "", "Context to be used in command")
-	rootCmd.PersistentFlags().StringSliceVarP(&nodes, "nodes", "", []string{}, "target the specified nodes")
+	rootCmd.PersistentFlags().StringSliceVarP(&nodes, "nodes", "n", []string{}, "target the specified nodes")
```
smira: I skipped that as I thought `-n` is usually shorthand for `--dry-run`, but probably that is not important

PR #1652, opened 2019-12-11
refactor: rename protobuf services, RPCs, and messages

andrewrynhard:
This PR brings our protobuf files into conformance with the protobuf
style guide, and community conventions. It is purely renames, along with
generated docs.


internal/app/apid/main.go
```diff
@@ -73,9 +73,9 @@ func main() {
 
 	// all existing streaming methods
 	for _, methodName := range []string{
-		"/machine.Machine/CopyOut",
+		"/machine.Machine/Copy",
```
smira: :+1:

PR #1657, opened 2019-12-12
refactor: Make basic-integration more modular

bradbeam:
This breaks up the checks and runners into something that we can
extend a bit easier.


cmd/osctl/cmd/cluster.go
```diff
@@ -78,11 +78,14 @@ var clusterDownCmd = &cobra.Command{
...
 
-	cli, err := client.NewEnvClient()
+	cli, err := client.NewClientWithOpts(client.FromEnv)
 	if err != nil {
 		return err
 	}
 
+	// Ensure docker client version works with server
+	cli.NegotiateAPIVersion(ctx)
+
```
bradbeam: This is necessary to bring in the updated docker libs.

`NewEnvClient()` is deprecated in favor of `NewClientWithOpts(...)`
`cli.NegotiateAPIVersion()` handles discovering which version of the api it needs to use to talk with the server.

internal/test-framework/cmd/basic_integration_run.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return false
+}
+
+func checkFile(filename string) bool {
+	f, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+
```
smira: if `err != nil` here, I believe `f` will be `nil`, so check for `f.IsDir()` is going to fail

internal/test-framework/cmd/basic_integration_run.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func clusterExists(ctx context.Context, cRunner runner.Runner) bool {
+	filters := filters.NewArgs()
+	filters.Add("label", fmt.Sprintf("%s=%s", "talos.owned", "true"))
+	filters.Add("label", fmt.Sprintf("%s=%s", "talos.cluster.name", clusterName))
+
+	containers, err := cRunner.(*container.ContainerRunner).Client.ContainerList(ctx, types.ContainerListOptions{Filters: filters})
+	if err != nil {
+		log.Fatal(err)
```
smira: idk if we want to return error back, but I imagine having some context here is going to help a lot, otherwise whole process crashes with some error like "context canceled", and we have zero idea on the context of the failure

.gitignore
```diff
@@ -10,6 +10,7 @@ talosconfig
 kubeconfig
 hack/test/integration/matchbox/assets/*
 !hack/test/integration/matchbox/assets/.gitkeep
+internal/test-framework/test-framework
```
smira: what is this directory? do we get it created while running?

internal/test-framework/pkg/basicintegration/options.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+}
+
+// WithTalosConfig specifies an existing location to a talosconfig.
+func WithTalosConfig(o string) Option {
+	return func(b *BasicIntegration) (err error) {
+		// Not sure if we want to do this yet
+		// since the cluster may or may not exist
+		/*
```
smira: should we remove this code if we're not going to use it?

smira: it should be a path where we store talosconfig once the cluster is provisioned? or the existing one which should be used?

internal/test-framework/pkg/basicintegration/basicintegration.go
```diff
@@ -0,0 +1,248 @@
...
+	if err = b.createCluster(ctx, cRunner); err != nil {
+		log.Fatal(err)
+	}
+
+	// nolint: errcheck
+	defer cRunner.Cleanup(ctx)
+
+	// TODO readd
+	/*
+		if b.cleanup {
```
smira: should we remove this code or uncomment it?

internal/test-framework/pkg/runner/container/container.go
```diff
@@ -0,0 +1,285 @@
...
+	)
+
+	cr := defaultOptions()
+
+	for _, setter := range setters {
+		result = multierror.Append(result, setter(cr))
+	}
+
+	if reader, err = cr.Client.ImagePull(context.Background(), cr.Container.Image, types.ImagePullOptions{}); err != nil {
+		result = multierror.Append(result, err)
```
smira: I think there's a flow problem here:

1. if method returns an error, `reader` is undefined, so we should abort execution here, we can't really continue
2. in fact, same above for setting options: if options can't be set, we shouldn't keep running 

so I believe instead of `multierror` we should immediately fail on error

internal/test-framework/pkg/runner/container/container.go
```diff
@@ -0,0 +1,285 @@
...
+			inspect types.ContainerExecInspect
+			resp    types.HijackedResponse
+		)
+
+		execConfig := types.ExecConfig{
+			Tty:          true,
+			AttachStdin:  true,
+			AttachStdout: true,
+			AttachStderr: true,
+			Cmd:          []string{"/bin/bash", "-c", check.Command.String()},
```
smira: as a suggestion, we can prepend safety defaults: https://github.com/talos-systems/bldr/blob/master/internal/pkg/types/v1alpha2/instruction.go#L15

andrewrynhard:
Closing in favor of the direction we are going with #1687.

PR #1661, opened 2019-12-16
feat: add support for tailing logs

### LINKED ISSUE #1564: Tailing logs
`osctl logs` & `osctl dmesg` should support `--tail=<lines>` flag.

smira:
Fixes #1564


api/machine/machine.proto
```diff
@@ -28,63 +28,41 @@ service MachineService {
   rpc Shutdown(google.protobuf.Empty) returns (ShutdownResponse);
   rpc Upgrade(UpgradeRequest) returns (UpgradeResponse);
 
-  rpc Start(StartRequest) returns (StartResponse) {
```
smira: looks like my editor, will revert

internal/pkg/tail/tail.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+)
+
+// Window is the read window size for tail scanning.
+const Window = 4096
+
+// SeekLines seeks the passed io.ReadSeeker so that it's -N lines from the tail.
```
smira: couldn't find a better name for it, suggestions?

smira:
TODO: 

[ ] add integration test for tail

internal/pkg/tail/tail.go
andrewrynhard: I think this sounds fine. It describes what it does.

api/machine/machine.proto
```diff
@@ -258,6 +258,7 @@ message LogsRequest {
   // driver might be default "containerd" or "cri"
   common.ContainerDriver driver = 3;
   bool follow = 4;
+  int32 tailLines = 5;
```
andrewrynhard: nit: the convention in protobuf is to use snake case.

smira: thanks, fixed!

smira:
_technically_ this can go into 0.3 as it's backwards compatible change

PR #1664, opened 2019-12-17
feat: support specifying CIDR for docker network

rsmitty:
This PR adds a `--cidr` flag to the custer create command that allows
for specification of the cidr block users want to use for their docker
network.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -28,20 +28,20 @@ import (
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1/generate"
 	"github.com/talos-systems/talos/pkg/constants"
+	talosnet "github.com/talos-systems/talos/pkg/net"
 	"github.com/talos-systems/talos/pkg/version"
 )
 
 var (
 	clusterName   string
 	nodeImage     string
+	networkCIDR   string
 	networkMTU    string
 	workers       int
 	masters       int
 	clusterCpus   string
 	clusterMemory int
 )
-
-const baseNetwork = "10.5.0.%d"
 
 // clusterCmd represents the cluster command
 var clusterCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/osctl/cmd/cluster.go
```diff
@@ -289,7 +305,7 @@ func createNetwork(cli *client.Client) (types.NetworkCreateResponse, error) {
 		IPAM: &network.IPAM{
 			Config: []network.IPAMConfig{
 				{
-					Subnet: fmt.Sprintf(baseNetwork, 0) + "/24",
+					Subnet: networkCIDR,
```
rsmitty: Originally thought we'd need to validate the cidr here, but I think we can actually just pass the string directly, since we ensured, and would have bailed, if the string wasn't valid on line 105 above.

rsmitty:
This isn't required for the configurable docker clusters necessarily, but it felt like it'd be plenty useful.

cmd/osctl/cmd/cluster.go
```diff
@@ -100,15 +99,32 @@ func create() (err error) {
...
 	ips := make([]string, masters)
 	for i := range ips {
-		ips[i] = fmt.Sprintf(baseNetwork, i+2)
+		ips[i] = masterIP.String()
+
+		// Make sure we don't overflow b/c some jabroni has requested a hilarious amount of masters
+		if masterIP[3] == 254 {
+			return errors.New("can't allocate any new IPs out of CIDR")
+		}
+		masterIP[3]++
```
smira: worker nodes are going to consume IPs as well, are we going to leave it up to docker to resolve?

rsmitty: Yep, that was my thought. That was the behavior we had previously.

cmd/osctl/cmd/cluster.go
jsimonetti: > This isn't required for the configurable docker clusters necessarily, but it felt like it'd be plenty useful. Tested on default cidr, `10.10.0.0/24`, and `10.10.0.128/25`

It is useful for running multiple talos clusters on one docker server.

cmd/osctl/cmd/cluster.go
```diff
@@ -100,15 +99,32 @@ func create() (err error) {
...
 
-	fmt.Println("generating PKI and tokens")
+	masterIP, _, err := net.ParseCIDR(networkCIDR)
+	if err != nil {
+		return fmt.Errorf("error validating cidr block: %w", err)
+	}
+
+	// Set starting ip at 2nd ip in range, ex: 192.168.0.2
+	masterIP = masterIP.To4()
+	masterIP[3] += 2
```
andrewrynhard: We have https://github.com/talos-systems/talos/blob/master/pkg/net/net.go#L52 to do this for us. Has all kinds of checks built into it.

cmd/osctl/cmd/cluster.go
andrewrynhard: We can remove this if we use  https://github.com/talos-systems/talos/blob/master/pkg/net/net.go#L52.

cmd/osctl/cmd/cluster.go
rsmitty: updated with using the nth function. thx for the pointer to that. also updated the names of the containers that get created to prepend the clustername. one step closer to doing multi-clusters on a single node.

cmd/osctl/cmd/cluster.go
```diff
@@ -129,7 +144,7 @@ func create() (err error) {
 		requests[i] = &node.Request{
 			Input:    *input,
 			Image:    nodeImage,
-			Name:     fmt.Sprintf("master-%d", i+1),
+			Name:     fmt.Sprintf("%s_master_%d", clusterName, i+1),
```
andrewrynhard: We should use `-` instead of `_` since `_` isn't allowed in DNS. 
nit: should we do `master-%s-%d` instead?

rsmitty: fixed

PR #1665, opened 2019-12-17
feat: add support for tftp download

andrewrynhard:
This adds supportr for downloading the machine config over TFTP. This
will allows users to avoid having to setup an HTTP server, and use
whatever they are using for PXE.

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -67,8 +67,9 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
 		opt(dlOpts)
 	}
 
-	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
-	if err != nil {
+	var req *http.Request
+
+	if req, err = http.NewRequest(http.MethodGet, u.String(), nil); err != nil {
 		return b, err
 	}
 
@@ -82,27 +83,26 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
 			return retry.ExpectedError(err)
 		}
 
-		if dlOpts.Format == b64 {
-			var b64 []byte
-
-			b64, err = base64.StdEncoding.DecodeString(string(b))
-			if err != nil {
-				return err
-			}
-
-			b = b64
-		}
-
 		return nil
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to download config from %q: %w", u.String(), err)
 	}
 
+	if dlOpts.Format == b64 {
+		var b64 []byte
+
+		b64, err = base64.StdEncoding.DecodeString(string(b))
+		if err != nil {
+			return nil, err
+		}
+
+		b = b64
+	}
+
 	return b, nil
 }
 
-// download handles the actual http request
 func download(req *http.Request) (data []byte, err error) {
 	client := &http.Client{}
 
@@ -123,4 +123,9 @@ func download(req *http.Request) (data []byte, err error) {
 	}
 
 	return data, err
+}
+
+func init() {
+	transport := (http.DefaultTransport.(*http.Transport))
+	transport.RegisterProtocol("tftp", NewTFTPTransport())
 }
--- a/pkg/download/tftp.go
+++ b/pkg/download/tftp.go
@@ -1,0 +1,66 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package download
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/pin/tftp"
+)
+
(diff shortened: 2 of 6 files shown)


pkg/download/download.go
```diff
@@ -67,43 +71,94 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
...
+	}
+
+	w, err := c.Receive(filename, "octet")
+	if err != nil {
+		return nil, err
+	}
+
+	var b bytes.Buffer
+
+	written, err := w.WriteTo(bufio.NewWriter(&b))
```
smira: I guess wrapping `bytes.Buffer` into `bufio` is kind of overkill, as it's buffering writes to a buffer ;)

afaik `io.Copy` can detect availability of `WriteTo` and `ReadFrom` methods, so `io.Copy(&b, w)` should work as well

`w` should be `r` probably? as it's more of a `Reader`?

andrewrynhard: Awesome. Will update.

andrewrynhard: hmm looks like `io.Copy(&b, w)` won't work because `c.Receive` returns an `io.WriterTo`.

andrewrynhard: Ok fixed. PTAL.

pkg/download/download.go
```diff
@@ -67,43 +70,94 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
 		opt(dlOpts)
 	}
 
-	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
-	if err != nil {
-		return b, err
-	}
+	switch u.Scheme {
+	case "http", "https":
```
smira: one more option which I liked in the past is to register `tftp` as a new protocol and keep the code clean from the difference in the protocol: https://godoc.org/net/http#Transport.RegisterProtocol

smira: I had some for the ftp written a long time ago, it might be that `tftp` would be easier for a case of a single download (no need to keep connections around, etc): https://github.com/smira/go-ftp-protocol/blob/master/protocol/protocol.go

andrewrynhard: 👍 update on the way

pkg/download/download.go
```diff
@@ -82,29 +83,32 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
...
+		b = b64
+	}
+
 	return b, nil
 }
 
-// download handles the actual http request
 func download(req *http.Request) (data []byte, err error) {
-	client := &http.Client{}
+	transport := &http.Transport{}
```
smira: I really hate that in Go's stdlib, but empty structure is a bad initializer as http's DefaultTransport is not empty: https://godoc.org/net/http#RoundTripper

Most important is missing `Proxy` function iirc, which allows to parse proxy from the env.

One option is to copy directly by value and use that as initial value:

`transport := *(http.DefaultTransport.(*http.Transport))`

pkg/download/download.go
```diff
@@ -82,29 +83,32 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
...
+
+	transport.RegisterProtocol("tftp", NewTFTPTransport())
```
smira: :heart:

pkg/download/download.go
smira: not sure if I got the syntax correctly, but I hope the idea is clear

andrewrynhard: Good point! Fixed.

andrewrynhard:
in a follow up I want to remove matchbox since we can just use `tftp` now. Less overhead.

pkg/download/download.go
```diff
@@ -82,29 +83,32 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
...
+		b = b64
+	}
+
 	return b, nil
 }
 
-// download handles the actual http request
 func download(req *http.Request) (data []byte, err error) {
-	client := &http.Client{}
+	transport := (http.DefaultTransport.(*http.Transport))
```
smira: this is not a copy, it's a reference to default transport, so it has a side effect of installing `tftp` to the default transport

pkg/download/tftp.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+	}
+
+	return &http.Response{
+		Status:        "200 OK",
+		StatusCode:    200,
+		Proto:         "TFTP/1.0",
+		ProtoMajor:    1,
+		ProtoMinor:    0,
+		Body:          ioutil.NopCloser(buf),
```
andrewrynhard: @smira This is the only thing I wasn't too sure about, but as far as I can tell, the TFTP library doesn't return anything I need to close. Gonna merge, can follow up if you think this isn't good.

smira: I haven't looked at tftp library closely, but you return the buffer, so you can close tftp client by the time request is done

smira: and as it's all UDP, I believe there's nothing to close really

andrewrynhard: yep... ok so I think this is good

PR #1667, opened 2019-12-17
fix: fail on muliple nodes for commands which don't support it

### LINKED ISSUE #1663: [osctl] fail fast on multiple nodes if command is not ready
All the `apid` requests can successfully be passed to multiple nodes, but not all the command implementations are ready to handle the response properly.

Until we get a proper fix, makes sense to do quick check/fail mode in the following commands:

* `copy`
* `list`
* `read`
* `kubeconfig`

smira:
Fixes #1663

(I believe it's 0.3 backport strong candidate).

### DIFF
--- a/cmd/osctl/cmd/copy.go
+++ b/cmd/osctl/cmd/copy.go
@@ -37,6 +37,10 @@ captures ownership and permission bits.`,
 		}
 
 		setupClient(func(c *client.Client) {
+			if err := failIfMultiNodes(globalCtx, "copy"); err != nil {
+				helpers.Fatalf("%s", err)
+			}
+
 			r, errCh, err := c.Copy(globalCtx, args[0])
 			if err != nil {
 				helpers.Fatalf("error copying: %s", err)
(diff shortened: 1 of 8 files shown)


andrewrynhard:
> (I believe it's 0.3 backport strong candidate).

andrewrynhard:
I agree with all of these except `list`. FWIW, it works fine today, with the exception that the node name is missing.

smira:
> I agree with all of these except `list`. FWIW, it works fine today, with the exception that the node name is missing.

up to you, I agree it doesn't break badly, but I thought adding new column (I was mostly worried about mode without `-l` which just prints the names) might be breaking change for 0.3? 

so I think there are three options:

1. keep list failing on multi-node
2. keep it as it is today (no failures, but no node)
3. fix list to print node names

smira:
> I agree with all of these except `list`. FWIW, it works fine today, with the exception that the node name is missing.

removed `list` from this PR, will fix `list` on multi-nodes in the follow-up PR

PR #1671, opened 2019-12-18
refactor: osctl code cleanup, no functional changes

### LINKED ISSUE #1666: [osctl] cleanup common comand infrastructure
1. Stop showing errors on our own, disable `SilenceUsage` and `SilenceErrors`
2. Use `Args:` validators instead of our own hand-rolled versions.
3. Use `RunE` consistently everywhere, get rid of `helpers.Fatalf` (should provide a chance for `defer` to run).
4. Use `setupClientE` everywhere, rename as `setupClient` (with `error` return).
5. Hide internal commands (docs generation sounds like first candidate).
6. Pass `ctx` explicitly via `setupClient`, get rid of `globalCtx` variable.

smira:
Fixes #1666

1. Remove custom validation of Args, use cobra-provided validators.
2. Always use errors to stop the execution flow, don't rely on
`log.Fatal` and `panic` for normal flows. This makes sure `defer` always
has a chance to run, connection is shut down in a clean way.
3. Command `docs` is hidden, as it's not for users.
4. Global variable `globalCtx` is removed, `WithClient` is used to pass
context to the command.
5. `setupClientE` renamed to `WithClient`, `setupClient` removed.
6. Code from `cmd/root.go` moved to `pkg/helpers` when possible.


cmd/osctl/cmd/list.go
```diff
@@ -23,45 +24,38 @@ import (
 
 const sixMonths = 6 * time.Hour * 24 * 30
 
+var (
+	long           bool
```
smira: reworked these flags to use variables to align other commands.

would like to push those global variables to be small internal structures passed to the command in the next round.

cmd/osctl/cmd/list.go
```diff
@@ -23,45 +24,38 @@ import (
...
+
 // lsCmd represents the ls command
 var lsCmd = &cobra.Command{
 	Use:     "list [path]",
 	Aliases: []string{"ls"},
 	Short:   "Retrieve a directory listing",
 	Long:    ``,
-	Run: func(cmd *cobra.Command, args []string) {
-		setupClient(func(c *client.Client) {
+	Args:    cobra.MaximumNArgs(1),
```
andrewrynhard: ❤️

cmd/osctl/cmd/list.go
```diff
@@ -173,9 +167,9 @@ var lsCmd = &cobra.Command{
...
 
 func init() {
-	lsCmd.Flags().BoolP("long", "l", false, "display additional file details")
-	lsCmd.Flags().BoolP("recurse", "r", false, "recurse into subdirectories")
-	lsCmd.Flags().BoolP("humanize", "H", false, "humanize size and time in the output")
-	lsCmd.Flags().Int32P("depth", "d", 0, "maximum recursion depth")
+	lsCmd.Flags().BoolVarP(&long, "long", "l", false, "display additional file details")
+	lsCmd.Flags().BoolVarP(&recurse, "recurse", "r", false, "recurse into subdirectories")
+	lsCmd.Flags().BoolVarP(&humanizeFlag, "humanize", "H", false, "humanize size and time in the output")
+	lsCmd.Flags().Int32VarP(&recursionDepth, "depth", "d", 0, "maximum recursion depth")
```
andrewrynhard: I like this style of args much better 👍.

smira:
> **new alerts:**
> 
> * 1 for Arbitrary file write during zip extraction ("zip slip")
> 
> **fixed alerts:**
> 
> * 1 for Arbitrary file write during zip extraction ("zip slip")

:man_facepalming:

PR #1674, opened 2019-12-19
feat: prepend cluster name to config generate

rsmitty:
This PR will take the cluster name and add it to the node type that is
generated by osctl config generate (ex: "mycluster-init.yaml"). Driver
for this is to help make it clearer which cluster a given yaml applies
to. Will also help with the ux of being able to config docker-based
talos clusters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,11 +2,15 @@ bin
 build
 images
 .vscode
+*-init.yaml
+*-controlplane.yaml
+*-join.yaml
+*-talosconfig
 init.yaml
 controlplane.yaml
 join.yaml
-docgen
 talosconfig
+docgen
 kubeconfig
 hack/test/integration/matchbox/assets/*
 !hack/test/integration/matchbox/assets/.gitkeep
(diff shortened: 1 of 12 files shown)


andrewrynhard:
There may be some documentation fixes to make because of this if you feel like getting those out of the way too.

rsmitty:
@andrewrynhard updated the doc mentions and tried to update the kvm integration. may want to double check me on that last one.

andrewrynhard:
I’m kinda torn on this now that I see what it looks like. Kinda thinking that people can just use the output dir if they need to generate more than one set of configs.

rsmitty:
I still think there's some value since, by default, we dump the yamls right in the working directory. So you could easily overwrite another cluster's configs. But yeah, we can chat about it when you're online.

andrewrynhard:
Gonna close this. I think the output dir offers enough flexibility here. To address to issue of overwriting, we could refuse to write if the file already exists.

PR #1677, opened 2019-12-20
test: add integration tests for (most) CLI commands

smira:
I added tests for all the commands which work reliably in container mode.

Some tests are naive, some are more sophisticated. While going
through the tests, I think I found a small bug in `osctl gen keypair`.

When we get reliable KVM tests, I can revisit and add missing
tests for time, reboot, shutdown and friends.

### DIFF
--- a/cmd/osctl/cmd/gen.go
+++ b/cmd/osctl/cmd/gen.go
@@ -244,8 +244,8 @@ func init() {
 	crtCmd.Flags().IntVar(&hours, "hours", 24, "the hours from now on which the certificate validity period ends")
 	// Keypairs
 	keypairCmd.Flags().StringVar(&ip, "ip", "", "generate the certificate for this IP address")
-	keypairCmd.Flags().StringVar(&ca, "ca", "", "path to the PEM encoded CERTIFICATE")
-	helpers.Should(cobra.MarkFlagRequired(keypairCmd.Flags(), "ca"))
+	keypairCmd.Flags().StringVar(&organization, "organization", "", "X.509 distinguished name for the Organization")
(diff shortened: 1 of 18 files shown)


cmd/osctl/cmd/gen.go
```diff
@@ -244,8 +244,8 @@ func init() {
 	crtCmd.Flags().IntVar(&hours, "hours", 24, "the hours from now on which the certificate validity period ends")
 	// Keypairs
 	keypairCmd.Flags().StringVar(&ip, "ip", "", "generate the certificate for this IP address")
-	keypairCmd.Flags().StringVar(&ca, "ca", "", "path to the PEM encoded CERTIFICATE")
-	helpers.Should(cobra.MarkFlagRequired(keypairCmd.Flags(), "ca"))
+	keypairCmd.Flags().StringVar(&organization, "organization", "", "X.509 distinguished name for the Organization")
```
smira: I believe this was a bug, as `ca` flag was never used

cmd/osctl/cmd/kubeconfig.go
```diff
@@ -25,7 +25,7 @@ var kubeconfigCmd = &cobra.Command{
 	Short: "Download the admin kubeconfig from the node",
 	Long: `Download the admin kubeconfig from the node.
 Kubeconfig will be written to PWD/kubeconfig or [local-path]/kubeconfig if specified.`,
-	Args: cobra.ExactArgs(1),
+	Args: cobra.MaximumNArgs(1),
```
smira: bug I introduced in previous PR, as now it requires an arg, while zero args should work (now with tests!)

PR #1679, opened 2019-12-20
fix: use proxy env vars when attempting an install

andrewrynhard:
Long story short, golang does not update its' internal proxy config
after it has been initialized. This means that if there are no proxy env
vars set, and the http client is used in any way, it will forever be
configured that way. So even if we export proxy env vars later, they
will never be respected. This addresses that by skipping the caching
mechanism that golang uses (once.Do), and forces the http client to
check for proxy env vars, for every request.


andrewrynhard:
See https://github.com/golang/go/issues/22079.

PR #1680, opened 2019-12-21
fix: Reset default http client to work around proxyEnv

bradbeam:
This fixes an issue where the default http client transport only reads the
environment variables once. This allows us to use a default/native network
configuration via dhcp to fetch userdata and then update the environment
after applying the configuration from userdata.

### DIFF
--- a/internal/app/machined/internal/phase/phase.go
+++ b/internal/app/machined/internal/phase/phase.go
@@ -7,10 +7,13 @@ package phase
 import (
 	"fmt"
 	"log"
+	"net/http"
+	"net/url"
 	goruntime "runtime"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	"golang.org/x/net/http/httpproxy"
 
 	"github.com/talos-systems/talos/internal/pkg/kmsg"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
@@ -59,6 +62,16 @@ func NewRunner(config runtime.Configurator, sequence runtime.Sequence) (*Runner,
 		if err = kmsg.Setup("[talos]", true); err != nil {
 			return nil, fmt.Errorf("failed to setup logging: %w", err)
 		}
+	}
+
+	// Re-define the default http client
+	// Work around our fun proxy.Do once bug
+	http.DefaultClient = &http.Client{
+		Transport: &http.Transport{
+			Proxy: func(req *http.Request) (*url.URL, error) {
+				return httpproxy.FromEnvironment().ProxyFunc()(req.URL)
+			},
+		},
 	}
 
 	runner := &Runner{
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	github.com/vmware/vmw-guestinfo v0.0.0-20170707015358-25eff159a728
 	go.etcd.io/etcd v3.3.13+incompatible
 	golang.org/x/crypto v0.0.0-20191108234033-bd318be0434a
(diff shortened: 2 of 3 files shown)


internal/app/machined/internal/phase/phase.go
```diff
@@ -61,6 +64,16 @@ func NewRunner(config runtime.Configurator, sequence runtime.Sequence) (*Runner,
 		}
 	}
 
+	// Re-define the default http client
+	// Work around our fun proxy.Do once bug
+	http.DefaultClient = &http.Client{
```
smira: I'm not sure I understand why we need to reset `http.DefaultClient`, as it's effectively zero value?

while `http.DefaultTransport` might need to be "reset", but I guess only `Proxy` func?

could there be a race if we have something using http is running in the background? Should we just define our own `Proxy` func which does lookup every time?

and `http.DefaulTransport` is _not_ zero value, so by setting it this way we lose a lot of specific settings:

```
var DefaultTransport RoundTripper = &Transport{
    Proxy: ProxyFromEnvironment,
    DialContext: (&net.Dialer{
        Timeout:   30 * time.Second,
        KeepAlive: 30 * time.Second,
        DualStack: true,
    }).DialContext,
    ForceAttemptHTTP2:     true,
    MaxIdleConns:          100,
    IdleConnTimeout:       90 * time.Second,
    TLSHandshakeTimeout:   10 * time.Second,
    ExpectContinueTimeout: 1 * time.Second,
}
```

in fact my 2 cents are that we should never use `DefaultClient`/`DefaultTransport` as they never set request timeouts, and that leads to hanging "forever" request, we should be much better at having our own `Client`/`Transport` instances shared across our codebase but with timeouts set.

bradbeam: > I'm not sure I understand why we need to reset http.DefaultClient, as it's effectively zero value?
> 
> while http.DefaultTransport might need to be "reset", but I guess only Proxy func?

Correct, really it's just the proxyFunc we need to reset to work around the sync.Once bit. 

> could there be a race if we have something using http is running in the background? Should we just define our own Proxy func which does lookup every time?

We could probably set it a single time earlier in the boot sequence, but this felt like a clean place to implement it. At the point in the boot sequence where this code runs this should be the only code that runs and it's serial by nature ( to ensure we start things up in the proper sequence ).

> and http.DefaulTransport is not zero value, so by setting it this way we lose a lot of specific settings:

Whoops. We should clean that up. 

> in fact my 2 cents are that we should never use DefaultClient/DefaultTransport

I think internally we use the default client a few times, but the issue isn't only limited to our use of the default client. Libraries we import also make use of the default client ( and this was the issue we ran into -- containerd client makes use of the default client to fetch images ). We thought of a few ways to try to mitigate the problem, and this seemed to be the only one that would catch most if not all of these instances. 
If we were to implement our own http client and pass that around, we'd most likely need to look a linter (or something) to catch the use of a default http client and warn to avoid these situations.

smira: > I think internally we use the default client a few times, but the issue isn't only limited to our use of the default client. Libraries we import also make use of the default client ( and this was the issue we ran into -- containerd client makes use of the default client to fetch images ).

oh, this is really not nice that these libraries rely on default HTTP client to behave the way we need...

bradbeam: I don't want to throw too much shade at containerd because we do have the option to override the default ( via WithResolver ) but I wouldn't be surprised if we get caught somewhere else that we aren't aware of.

PR #1683, opened 2019-12-23
fix(networkd): Make better route scoping decisions

bradbeam:
This brings in an updated library along with some tweaks on our side to allow for
better decision making when it comes to the scope of routes. This also fixes an
issue where multiple configuration definitions for an interface were not properly
merged and instead were overwritten.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.13
 
 replace (
 	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-	github.com/jsimonetti/rtnetlink => github.com/bradbeam/rtnetlink v0.0.0-20191116224831-33eac9dd89f6
 	github.com/kubernetes-sigs/bootkube => github.com/talos-systems/bootkube v0.14.1-0.20191210195608-7606a2efb62d
 	github.com/opencontainers/runtime-spec v1.0.1 => github.com/opencontainers/runtime-spec v0.1.2-0.20180301181910-fa4b36aa9c99
 )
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 replace (
 	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-	github.com/jsimonetti/rtnetlink => github.com/bradbeam/rtnetlink v0.0.0-20191116224831-33eac9dd89f6
+	github.com/jsimonetti/rtnetlink => github.com/bradbeam/rtnetlink v0.0.0-20191223025716-956ac0494b04
```
bradbeam: related: https://github.com/jsimonetti/rtnetlink/pull/67

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -74,7 +74,11 @@ func New(config runtime.Configurator) (*Networkd, error) {
 				continue
 			}
 
-			netconf[name] = opts
```
bradbeam: this allows for merging of the config options if an interface is defined multiple times.

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -187,11 +188,7 @@ func (n *NetworkInterface) Addressing() error {
 	for _, method := range n.AddressMethod {
 		if err := n.configureInterface(method); err != nil {
 			// Treat as non fatal error when failing to configure an interface
-			return nil
-		}
-
-		if !method.Valid() {
-			return nil
```
bradbeam: continue instead of return early. This function could probably use a little bit of a refactor ( not return a value since we don't ever return an error ), but that can wait till later.

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -273,9 +271,15 @@ func (n *NetworkInterface) configureInterface(method address.Addressing) error {
 
 	// Add any routes
 	for _, r := range method.Routes() {
```
bradbeam: Explicitly set gateway to nil if it is set to `0.0.0.0`

bradbeam:
e2e seemed to work just fine against this change. neat. was worried about gcp with the static classless routes, but seems like it's doing the right thing :tm:

https://ci.dev.talos-systems.io/talos-systems/talos/6003/1/30

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -273,9 +271,15 @@ func (n *NetworkInterface) configureInterface(method address.Addressing) error {
...
+		gw := r.Router
+		if net.IPv4zero.Equal(gw) {
+			gw = nil
+		}
+
 		// Any errors here would be non-fatal, so we'll just ignore them
-		// nolint: errcheck
-		n.rtnlConn.RouteAddSrc(method.Link(), *r.Dest, method.Address(), r.Router)
+		//nolint: errcheck
+		n.rtnlConn.RouteAddSrc(method.Link(), *r.Dest, method.Address(), gw)
```
smira: should we log the errors in case it helps with the debugging? or that is something we don't really need?

bradbeam: once we get log levels implemented i'd very much like to spit out errors

PR #1687, opened 2019-12-23
refactor: extract Talos cluster provisioner as common code

smira:
This extracts Docker Talos cluster provisioner as common code
which might be shared between `osctl cluster` and integration-test.

There should be almost no functional changes.

As proof of concept, abstract cluster readiness checks were implemented
based on provisioned cluster state. It implements same checks as
`basic-integration.sh` in pure Go via Talos/K8s clients.

`conditions` package was promoted from machined-internal to
`internal/pkg` as it is used to run the checks.


internal/pkg/provision/check/default.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"time"
+
+	"github.com/talos-systems/talos/internal/pkg/conditions"
+	"github.com/talos-systems/talos/internal/pkg/provision"
+)
+
+// DefaultClusterChecks returns a set of default Talos cluster readiness checks.
+func DefaultClusterChecks() []ClusterCheck {
+	return []ClusterCheck{
```
smira: this is the core of the new checks: list of all the checks to execute against the cluster

cmd/osctl/cmd/cluster.go
```diff
@@ -118,238 +100,109 @@ func create() (err error) {
...
 
-	var wg sync.WaitGroup
-
-	wg.Add(len(containers))
-
-	for _, container := range containers {
-		go func(container types.Container) {
-			fmt.Println("destroying node", container.Names[0][1:])
+	// Run cluster readiness checks
+	checkCtx, checkCtxCancel := context.WithTimeout(ctx, 20*time.Minute)
```
smira: this timeout should probably be a flag for `osctl cluster create`, it limits overall time for cluster readiness

cmd/osctl/cmd/cluster.go
```diff
@@ -382,13 +235,13 @@ func parseCPUShare() (int64, error) {
 }
 
 func init() {
-	clusterUpCmd.Flags().StringVar(&nodeImage, "image", defaultImage(constants.DefaultTalosImageRepository), "the image to use")
```
smira: oops mismerge, will fix it

smira:
Not connected to the build (yet), want to put that as a separate PR, but now we can do:

```
$ osctl cluster create --masters 3 --workers 1 --wait
validating CIDR and reserving master IPs
generating PKI and tokens
creating network talos-default
creating master nodes
creating worker nodes
waiting for bootkube to finish: ...
waiting for bootkube to finish: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: EOF"
waiting for bootkube to finish: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: read tcp 127.0.0.1:47972->127.0.0.1:50000: read: connection reset by peer"
waiting for bootkube to finish: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: EOF"
waiting for bootkube to finish: service "bootkube" not in expected state "Finished": current state [Waiting] Waiting for service "etcd" to be "up"
waiting for bootkube to finish: service "bootkube" not in expected state "Finished": current state [Preparing] Running pre state
waiting for bootkube to finish: service "bootkube" not in expected state "Finished": current state [Running] Service started as goroutine
waiting for bootkube to finish: OK
waiting for apid ready: ...
waiting for apid ready: OK
waiting for all k8s nodes to report: ...
waiting for all k8s nodes to report: Get https://10.5.0.2:6443/api/v1/nodes: dial tcp 10.5.0.2:6443: connect: connection refused
waiting for all k8s nodes to report: OK
waiting for all k8s nodes to report ready: ...
waiting for all k8s nodes to report ready: OK
waiting for all master nodes to be part of k8s control plane: ...
waiting for all master nodes to be part of k8s control plane: OK
waiting for kube-proxy to report ready: ...
waiting for kube-proxy to report ready: OK
waiting for kube-dns to report ready: ...
waiting for kube-dns to report ready: OK
```

cmd/osctl/cmd/cluster.go
```diff
@@ -118,238 +100,109 @@ func create() (err error) {
...
-		// nolint: errcheck
-		defer reader.Close()
+	return provisioner.Destroy(ctx, fakeCluster{clusterName})
+}
 
-		if _, err = io.Copy(ioutil.Discard, reader); err != nil {
-			return err
-		}
-	}
+type fakeCluster struct {
```
andrewrynhard: A little confused by this. Is this temporary? What makes it "fake", and why can't we have a "real" one?

cmd/osctl/cmd/cluster.go
```diff
@@ -383,12 +236,13 @@ func parseCPUShare() (int64, error) {
 
 func init() {
 	clusterUpCmd.Flags().StringVar(&nodeImage, "image", defaultImage(constants.DefaultTalosImageRepository), "the image to use")
-	clusterUpCmd.Flags().StringVar(&networkMTU, "mtu", "1500", "MTU of the docker bridge network")
+	clusterUpCmd.Flags().IntVar(&networkMTU, "mtu", 1500, "MTU of the docker bridge network")
```
andrewrynhard: 👍

cmd/osctl/pkg/client/client.go
```diff
@@ -49,14 +49,20 @@ type Client struct {
...
 	if err != nil {
 		return
 	}
 
+	context, creds, err = NewClientContextAndCredentialsFromParsedConfig(c, ctx)
+
+	return
+}
+
+// NewClientContextAndCredentialsFromParsedConfig initializes Credentials from config file.
```
andrewrynhard: Should this doc string be updated? What makes it different than `NewClientContextAndCredentialsFromConfig`. I can tell by looking at the code, but just pointing out that maybe the doc should be updated?

internal/pkg/conditions/poll.go
```diff
@@ -0,0 +1,86 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewrynhard: Should we look at expanding the retry package for this? Feels similar, but not sure.

internal/pkg/conditions/poll.go
```diff
@@ -0,0 +1,86 @@
...
+		} else {
+			lastErr = "OK"
+		}
+	}
+	p.lastErrMu.Unlock()
+
+	return fmt.Sprintf("%s: %s", p.description, lastErr)
+}
+
+func (p *pollingCondition) Wait(ctx context.Context) error {
```
andrewrynhard: Should we look at expanding the retry package for this? Feels similar, but not sure. Or maybe we use the retry package under the hood?

internal/pkg/provision/access/access.go
```diff
@@ -0,0 +1,6 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package access provides methods to access provisioned Talos cluster.
```
andrewrynhard: Not something for this PR, but we need to do this a lot more. Also, a convention I have seen is to call this `docs.go`. Or was it `doc.go` 🤔? Something for later.

internal/pkg/provision/access/adapter.go
```diff
@@ -0,0 +1,104 @@
...
+
+	client, err := client.NewClient(creds, endpoints, constants.ApidPort)
+	if err == nil {
+		a.clients[key] = client
+	}
+
+	return client, err
+}
+
+func (a *adapter) K8sClient(ctx context.Context) (*kubernetes.Clientset, error) {
```
andrewrynhard: We have a kubernetes package we maintain, should we move some of this logic in there? I could go either way, just pointing out the package (`pkg/kubernetes`).

internal/pkg/provision/check/kubernetes.go
```diff
@@ -0,0 +1,172 @@
...
+					if nodeAddress.Type == v1.NodeInternalIP {
+						actualNodes = append(actualNodes, nodeAddress.Address)
+						break
+					}
+				}
+
+				break
+			}
+		}
+	}
```
andrewrynhard: This is fun 😄.

cmd/osctl/cmd/cluster.go
smira: this is something I don't have elegant solution for so far. In order to destroy a cluster, I want to have a `Cluster` description, which makes sense for many types of clusters I think (e.g. for firecracker). with Docker our flow relies only on cluster name, as Docker has rich metadata API and we can find nodes, network only by cluster name. So to satisfy the interface, here I build a structure which implements `Cluster` interface, but provides only `ClusterName` property, probably I should make it less cryptic and move it some package, as it might be only Docker-specific.

cmd/osctl/pkg/client/client.go
smira: yep, thanks, will do!

internal/pkg/provision/access/access.go
smira: I tried to follow the convention of having `<package_name>/<package_name>.go` file with this documentation line and most common types.

internal/pkg/conditions/poll.go
smira: This looks very similar, but `retry` has two issues:

* it doesn't support passing `ctx` down to the function
* it doesn't support canceling retry loop on `ctx`

`Condition` seemed to be more natural choice, as it encapsulates both waiting for something and polling under the hood, and it passes `ctx` and it has description for console output.

So if we expand `retry` to handle `ctx`, we can rewrite this to unify it with `retry`, and plus this code limits execution of one attempt to `interval`, so that we can't have 1 attempt taking all the `timeout`.

internal/pkg/provision/access/adapter.go
smira: I missed that, looks like I can move lots of stuff there

internal/pkg/conditions/poll.go
andrewrynhard: Awesome, sounds like something we can do in the future.

cmd/osctl/cmd/cluster.go
smira: I removed `fakeCluster` and replaced with `Provisioner.Reflect` method which allows to reflect back `Cluster` object from running configuration. It's not completely implemented yet, but enough for `cluster destroy` to work properly. My future plan is to have more complete support for reflection, so that we could potentially scale existing cluster, or wait for readiness of already existing cluster. Either reflection, or dumping/loading `Cluster` to `yaml` (might be better to see which ones works better after firecracker).

internal/pkg/provision/access/adapter.go
smira: Looked into this more, and I couldn't find good way here. Either move all the assertions to `pkg/kubernetes` which looked strange as they depend on `ClusterAccess` object (we could first derive k8s client, and then call `pkg/kubernetes`), and these assertions seems to be specific to cluster readiness, not really generic.

Another option is to extract common code of assertions, but they seem to be trivial (getting list of nodes, pods), which is just 2-3 lines of code.

andrewrynhard: SGTM, let's leave as is then.

PR #1690, opened 2019-12-25
fix(machined): Add additional defaults for http transport

bradbeam:
Followup from #1680.

This also moves the setting from phases to machine.init to set it earlier in
the boot sequence to ensure that we get the defaults set properly from the
start and set it only once.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -8,9 +8,13 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"time"
 
+	"golang.org/x/net/http/httpproxy"
 	"golang.org/x/sys/unix"
 
 	machineapi "github.com/talos-systems/talos/api/machine"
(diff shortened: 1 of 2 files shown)


internal/app/machined/main.go
```diff
@@ -68,6 +72,30 @@ func sync() {
...
+	http.DefaultClient.Transport = &http.Transport{
+		Proxy: func(req *http.Request) (*url.URL, error) {
+			return httpproxy.FromEnvironment().ProxyFunc()(req.URL)
+		},
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		ForceAttemptHTTP2:     true,
```
smira: one more option is to copy from `http.DefaultTransport`, but I guess it's  same in the end (as it's unlikely they're going to change defaults ever)

https://github.com/aptly-dev/aptly/blob/master/http/download.go#L43

andrewrynhard: Probably true, but then again if defaults ever do change we won't have to worry if we copy. Not a blocker for me.

PR #1696, opened 2019-12-27
chore: use osctl cluster --wait in basic-integration

smira:
Added flag to workaround CI/CD issue to use direct connection to docker
container (init node) instead of 127.0.0.1 passthrough (default is still
using 127.0.0.1).

### DIFF
--- a/hack/test/basic-integration.sh
+++ b/hack/test/basic-integration.sh
@@ -3,93 +3,35 @@
 set -eou pipefail
 
 
-export KUBERNETES_VERSION=v1.17.0
-export TALOS_IMG="docker.io/autonomy/talos:${TAG}"
-export TMP="/tmp/e2e"
+TALOS_IMG="docker.io/autonomy/talos:${TAG}"
+OSCTL="${PWD}/${ARTIFACTS}/osctl-linux-amd64"
+INTEGRATIONTEST="${PWD}/bin/integration-test"
+TMP="/tmp/e2e"
 export TALOSCONFIG="${TMP}/talosconfig"
-export KUBECONFIG="${TMP}/kubeconfig"
-export TIMEOUT=300
-export OSCTL="${PWD}/${ARTIFACTS}/osctl-linux-amd64"
-export INTEGRATIONTEST="${PWD}/bin/integration-test"
+
+case "${CI:-false}" in
+  true)
+    ENDPOINT="docker"
+    ;;
+  *)
+    ENDPOINT="127.0.0.1"
+    ;;
+esac
 
 case $(uname -s) in
   Linux*)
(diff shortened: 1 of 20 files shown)


pkg/config/types/v1alpha1/generate/controlplane.go
```diff
@@ -17,7 +17,7 @@ func controlPlaneUd(in *Input) (string, error) {
 		MachineType:     "controlplane",
 		MachineToken:    in.TrustdInfo.Token,
 		MachineCA:       in.Certs.OS,
-		MachineCertSANs: []string{},
+		MachineCertSANs: in.AdditionalSubjectAltNames,
```
smira: I'm not sure if `AdditionalSubjectAltNames` can be applied here for `apid` cert? Or we need a different setting in `generate.Input`?

andrewrynhard: We have two sets of SANs in our config... we have one for k8s and one for Talos. I don't think we have a way to set the Talos SANs in the `Input` though.

smira: So should I introduce separate field for Talos SANs or fine to re-use?

andrewrynhard: I think we will have to introduce another flag.

andrewrynhard: Can't think of a better way at the moment... maybe once @rsmitty implements the ability to specify files for the user data then maybe we could generate the configs, modify them, and then create the cluster.

hack/test/basic-integration.sh
```diff
@@ -3,93 +3,32 @@
...
-export KUBECONFIG="${TMP}/kubeconfig"
-export TIMEOUT=300
-export OSCTL="${PWD}/${ARTIFACTS}/osctl-linux-amd64"
-export INTEGRATIONTEST="${PWD}/bin/integration-test"
+TALOS_IMG="docker.io/autonomy/talos:${TAG}"
+OSCTL="${PWD}/${ARTIFACTS}/osctl-linux-amd64"
+INTEGRATIONTEST="${PWD}/bin/integration-test"
+
+
+case "${CI:-false}" in
```
andrewrynhard: I don't see where CI is getting set. Does Drone set it?

smira: yep, it's one of the Drone environment variables

pkg/config/types/v1alpha1/generate/controlplane.go
smira: Added `AdditionalMachineCertSANs` :)

PR #1697, opened 2019-12-27
feat: enable DynamicKubeletConfiguration

andrewrynhard:
This moves to using the KubeletConfiguration instead of flags to the
kubelet. It also enables DynamicKubeletConfiguration, which allows users
to configure kubelets using a ConfigMap.


internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -34,6 +37,7 @@ import (
 	"github.com/talos-systems/talos/pkg/argsbuilder"
 	"github.com/talos-systems/talos/pkg/constants"
 	tnet "github.com/talos-systems/talos/pkg/net"
+	// tnet "github.com/talos-systems/talos/pkg/net"
```
rsmitty: nit: drop this guy

andrewrynhard: ahh good catch

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -225,47 +232,99 @@ func (k *Kubelet) HealthSettings(runtime.Configurator) *health.Settings {
...
+
+	return &kubeletconfig.KubeletConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "kubelet.config.k8s.io/v1beta1",
+			Kind:       "KubeletConfiguration",
+		},
+		StaticPodPath:      "/etc/kubernetes/manifests",
+		Address:            "0.0.0.0",
+		Port:               10250,
+		ReadOnlyPort:       10255, // TODO(andrewrynhard): Disable this.
```
rsmitty: should this be disabled now or is that for future andrew?

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -225,47 +232,99 @@ func (k *Kubelet) HealthSettings(runtime.Configurator) *health.Settings {
...
+		},
+		Authorization: kubeletconfig.KubeletAuthorization{
+			Mode: kubeletconfig.KubeletAuthorizationModeWebhook,
+		},
+		ClusterDomain:       "cluster.local",
+		ClusterDNS:          clusterDNS,
+		SerializeImagePulls: &f,
+		FailSwapOn:          &f,
+	}
+}
```
andrewrynhard: This is basically a translation of old flags into their respective config field.

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Fixed 👍

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Future Andrew :D ... this is something we can do to improve security, but nothing we did prior to this PR.

andrewrynhard: I should probably make a ticket.

PR #1699, opened 2019-12-28
test: ensure etcd is healthy on all control plane nodes

andrewrynhard:
This adds a cluster check that requires etcd to be healthy.

### DIFF
--- a/internal/pkg/provision/check/service.go
+++ b/internal/pkg/provision/check/service.go
@@ -9,17 +9,18 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/talos-systems/talos/cmd/osctl/pkg/client"
 	"github.com/talos-systems/talos/internal/pkg/provision"
 )
 
 // ServiceStateAssertion checks whether service reached some specified state.
 func ServiceStateAssertion(ctx context.Context, cluster provision.ClusterAccess, service, state string) error {
-	client, err := cluster.Client()
+	cli, err := cluster.Client()
 	if err != nil {
 		return err
 	}
 
-	servicesInfo, err := client.ServiceInfo(ctx, service)
+	servicesInfo, err := cli.ServiceInfo(ctx, service)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/provision/check/service.go
```diff
@@ -45,3 +46,65 @@ func ServiceStateAssertion(ctx context.Context, cluster provision.ClusterAccess,
...
+		lastEvent := serviceInfo.Service.Events.Events[len(serviceInfo.Service.Events.Events)-1]
+		if lastEvent.State != "Running" {
+			return fmt.Errorf("service %q not in expected state %q: current state [%s] %s", service, "Running", lastEvent.State, lastEvent.Msg)
+		}
+
+		if !serviceInfo.Service.GetHealth().GetHealthy() {
+			return fmt.Errorf("service is not health: %s", service)
+		}
+
+		serviceOk = true
```
smira: should we check it's healthy on _all_ the nodes? this seems to succeed if just one node service is healthy?

internal/pkg/provision/check/service.go
```diff
@@ -45,3 +46,65 @@ func ServiceStateAssertion(ctx context.Context, cluster provision.ClusterAccess,
...
+			return fmt.Errorf("no events recorded yet for service %q", service)
+		}
+
+		lastEvent := serviceInfo.Service.Events.Events[len(serviceInfo.Service.Events.Events)-1]
+		if lastEvent.State != "Running" {
+			return fmt.Errorf("service %q not in expected state %q: current state [%s] %s", service, "Running", lastEvent.State, lastEvent.Msg)
+		}
+
+		if !serviceInfo.Service.GetHealth().GetHealthy() {
+			return fmt.Errorf("service is not health: %s", service)
```
smira: nit: healthY

internal/pkg/provision/check/service.go
andrewrynhard: Updated. I think it checks all now?

PR #1702, opened 2019-12-30
feat(machined): Introduce limits api

bradbeam:
This enables a user to get the current process limits or specify
a new value to set the limit to.


bradbeam:
This compiles, but haven't tested yet. I'm not too fond of the names here but should at least give a good idea of where I was going.

PR #1707, opened 2019-12-31
fix(image): Use correct args for vmware

bradbeam:
This fixes the console and config options for vmware.

### DIFF
--- a/internal/pkg/runtime/platform/vmware/vmware.go
+++ b/internal/pkg/runtime/platform/vmware/vmware.go
@@ -28,41 +28,32 @@ func (v *VMware) Name() string {
 
 // Configuration implements the platform.Platform interface.
 func (v *VMware) Configuration() ([]byte, error) {
-	var option *string
-	if option = kernel.ProcCmdline().Get(constants.KernelParamConfig).First(); option == nil {
-		return nil, fmt.Errorf("no config option was found")
+	ok, err := vmcheck.IsVirtualWorld()
+	if err != nil {
+		return nil, err
 	}
 
-	if *option == constants.ConfigGuestInfo {
-		ok, err := vmcheck.IsVirtualWorld()
-		if err != nil {
-			return nil, err
-		}
-
-		if !ok {
-			return nil, errors.New("not a virtual world")
-		}
-
-		config := rpcvmx.NewConfig()
-
-		val, err := config.String(constants.VMwareGuestInfoConfigKey, "")
-		if err != nil {
-			return nil, fmt.Errorf("failed to get guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
-		}
-
-		if val == "" {
-			return nil, fmt.Errorf("config is required, no value found for guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
-		}
-
-		b, err := base64.StdEncoding.DecodeString(val)
-		if err != nil {
-			return nil, fmt.Errorf("failed to decode guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
-		}
-
-		return b, nil
+	if !ok {
+		return nil, errors.New("not a virtual world")
 	}
 
-	return nil, nil
+	config := rpcvmx.NewConfig()
+
+	val, err := config.String(constants.VMwareGuestInfoConfigKey, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
+	}
+
+	if val == "" {
+		return nil, fmt.Errorf("config is required, no value found for guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
+	}
+
+	b, err := base64.StdEncoding.DecodeString(val)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode guestinfo.%s: %w", constants.VMwareGuestInfoConfigKey, err)
+	}
+
+	return b, nil
 }
 
 // Hostname implements the platform.Platform interface.
@@ -83,7 +74,7 @@ func (v *VMware) ExternalIPs() (addrs []net.IP, err error) {
 // KernelArgs implements the runtime.Platform interface.
 func (v *VMware) KernelArgs() kernel.Parameters {
 	return []*kernel.Parameter{
-		kernel.NewParameter("console").Append("ttyS0"),
+		kernel.NewParameter("console").Append("tty0"),
 		kernel.NewParameter("earlyprintk").Append("ttyS0,115200"),
 	}
 }
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -166,6 +166,8 @@ const (
 	MetalConfigISOLabel = "metal-iso"
 
 	// ConfigGuestInfo is the name of the VMware guestinfo config strategy.
+	// Deprecation notice:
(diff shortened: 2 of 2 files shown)


internal/pkg/runtime/platform/vmware/vmware.go
```diff
@@ -83,7 +83,8 @@ func (v *VMware) ExternalIPs() (addrs []net.IP, err error) {
 // KernelArgs implements the runtime.Platform interface.
 func (v *VMware) KernelArgs() kernel.Parameters {
 	return []*kernel.Parameter{
-		kernel.NewParameter("console").Append("ttyS0"),
+		kernel.NewParameter("console").Append("tty0"),
 		kernel.NewParameter("earlyprintk").Append("ttyS0,115200"),
+		kernel.NewParameter(constants.KernelParamConfig).Append(constants.ConfigGuestInfo),
```
andrewrynhard: I think we should set this in our official OVA using the Makefile. It is possible that a user could not want to use guest info for the config. VMware is sort of one of the special cases, like KVM, where there is more than one way you can get the config to the machine.

andrewrynhard: Maybe the Makefile is a bad place for it. I like the idea of not setting this if the user supplies something different.

cmd/installer/pkg/install.go
```diff
@@ -29,12 +30,21 @@ func Install(p runtime.Platform, config runtime.Configurator, sequence runtime.S
...
+	option := cmdline.Get(constants.KernelParamConfig).First()
+
+	switch {
+	case option == nil:
+		cmdline.Append(constants.KernelParamConfig, opts.ConfigSource)
+	case option != nil && opts.ConfigSource != "none":
+		fmt.Printf("warning: no kernel arguments supplied for %s\n", constants.KernelParamConfig)
+		fmt.Println("this may cause your machine to fail to boot")
+	}
+
```
andrewrynhard: This should be in `cmd/installer/cmd/install.go`. The idea was that this package is a simple point and fire function, and that all the logic for figuring out the installation parameters would happen in `cmd/installer/cmd/install.go`.

cmd/installer/pkg/install.go
```diff
@@ -29,12 +30,21 @@ func Install(p runtime.Platform, config runtime.Configurator, sequence runtime.S
...
 	if err = cmdline.AppendAll(p.KernelArgs().Strings()); err != nil {
 		return err
 	}
 
+	option := cmdline.Get(constants.KernelParamConfig).First()
+
+	switch {
+	case option == nil:
+		cmdline.Append(constants.KernelParamConfig, opts.ConfigSource)
+	case option != nil && opts.ConfigSource != "none":
```
andrewrynhard: Not sure I understand this. The `option` variable is not `nil`, yet the warning says "no kernel arguments supplied".

cmd/installer/pkg/install.go
```diff
@@ -29,12 +30,22 @@ func Install(p runtime.Platform, config runtime.Configurator, sequence runtime.S
 	cmdline := kernel.NewCmdline("")
 	cmdline.Append("initrd", filepath.Join("/", "default", constants.InitramfsAsset))
 	cmdline.Append(constants.KernelParamPlatform, p.Name())
-	cmdline.Append(constants.KernelParamConfig, opts.ConfigSource)
 
 	if err = cmdline.AppendAll(p.KernelArgs().Strings()); err != nil {
 		return err
 	}
 
+	if option := cmdline.Get(constants.KernelParamConfig).First(); option == nil {
```
andrewrynhard: This will always be true since we create an empty command line with `kernel.NewCmdline("")`.

bradbeam: Negative, see L34 -- 
```
if err = cmdline.AppendAll(p.KernelArgs().Strings()); err != nil {
		return err
	}
```
This should only be the case when a platform does not provide a location / configuration. It's fine if the platform provides `talos.config=none` since the platform is being explicit.

internal/pkg/runtime/platform/vmware/vmware.go
bradbeam: since this is specific to the VMWare platform i don't think there's any harm here.

andrewrynhard: The harm is that `guestinfo` is optional, and that VMware allows for more than one way to configure it.

andrewrynhard: Also, VMware becomes the snowflake here. No other provider is supplying this arg, because it is meant to be supplied on the frontend via the installer.

internal/pkg/runtime/platform/vmware/vmware.go
```diff
@@ -83,7 +83,8 @@ func (v *VMware) ExternalIPs() (addrs []net.IP, err error) {
 // KernelArgs implements the runtime.Platform interface.
 func (v *VMware) KernelArgs() kernel.Parameters {
 	return []*kernel.Parameter{
-		kernel.NewParameter("console").Append("ttyS0"),
+		kernel.NewParameter("console").Append("tty0"),
```
andrewrynhard: Let's scope this PR to just this. I think this needs a bit more thought on how to best get the config arg.

bradbeam: We can't scope it to just this. Without config=guestinfo we can't boot.

bradbeam: if we keep the old way in place, then when we parse  / query kernel args, we pull the first arg which is config=none. 
there's definitely room for the vmware platform userdata bit to be enhanced, but i think that's a bit out of scope. eh maybe not. we can drop `guestinfo` as an option for talos.config and just go based on platform. part of the buy-in when you select a platform ( ex, if i pick aws i'm going to use metadata service for aws )

andrewrynhard: We can scope it to fixing the console output. For cloud providers I think it makes sense that you buy in to the platform’s metadata service, but VMware is different. It should be more flexible. Plus, setting this here flips everything we have done with the installer. The installer allows for the configuration of the config option. It should not be baked into the platform. Our official images are configured with same defaults, but the idea with the installer has always been to allow for configuring this option.

bradbeam: i disagree. 
vmware isn't any different. if you want to run on vmware but don't want to use the baked in cloud provider bits, then set the platform to bare metal and customize until you are content. otherwise it should behave like all other platforms and use the default discovery method for userdata. 
ill push up another commit to revert the changes to cmd/install and loosen up the provider/vmware bits / deprecate the talos.config=guestinfo part.

andrewrynhard: Allowing for more than one way to configure VMware was an intentional design decision made when the VMware provider was first implemented. I would like to keep it that way. The reason being is that even in places like AWS, you can boot bare metal machines, and you may want to supply the config another way, but still use the other platform features like discovering the external IPs and all the other magic we do. That is a bit of a contrived example, but when I first implemented the VMware platform, I was using an http server to serve the configs, and found it very useful to _not_ have to use the `guestinfo` option. The installer has the option to explicitly provide the config source, and I want to keep that. The idea has always been to provide sane images out of the box, but to not _force_ users into a particular workflow. If they want to generate images, it is a simple docker command, and it has all the flags to configure the image to your liking. Let's hold this PR until we can come up with a better solution.

PR #1708, opened 2019-12-31
chore: allow re-use of docker network for local clusters

rsmitty:
This PR will allow users to use an existing docker network for their
talos cluster. Hoping this will be useful for those wanting further
control and configuration of their local docker clusters, as well as
possibly useful for us during CI. The docker networks can be pre-created
with something like: `docker network create my-cluster --subnet
192.168.0.0/24 --label talos.owned=true --label
talos.cluster.name=my-cluster`. Note that the labels are pre-reqs for our discovery and re-use of these networks.

### DIFF
--- a/internal/pkg/provision/providers/docker/network.go
+++ b/internal/pkg/provision/providers/docker/network.go
@@ -6,6 +6,7 @@ package docker
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 
 	"github.com/docker/docker/api/types"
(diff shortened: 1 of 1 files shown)


internal/pkg/provision/providers/docker/network.go
```diff
@@ -16,7 +17,24 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/provision"
 )
 
+// createNetwork will take a network request and check if a network with the same name + cidr exists.
+// If so, it simply returns without error and assumes we will re-use that network. Otherwise it will create a new one.
 func (p *provisioner) createNetwork(ctx context.Context, req provision.NetworkRequest) error {
+	existingNet, err := p.listNetworks(ctx, req.Name)
```
smira: should we update or remove the error message in https://github.com/talos-systems/talos/blob/master/internal/pkg/provision/providers/docker/create.go#L52-L54, as right now it will be triggered in the wrong case?

rsmitty: yeah, I'm going to update that error message with PR #1716

PR #1709, opened 2020-01-01
fix: set the correct kernel args for VMware

andrewrynhard:
This enusres that we default to using `guestinfo` for VMware's config
source, and that we use tty0 instead of ttyS0 for the console.

### DIFF
--- a/cmd/installer/cmd/image.go
+++ b/cmd/installer/cmd/image.go
@@ -18,6 +18,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/runtime/platform"
 	"github.com/talos-systems/talos/pkg/cmd"
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1"
+	"github.com/talos-systems/talos/pkg/constants"
 )
 
 var outputArg string
@@ -39,6 +40,7 @@ func init() {
 	rootCmd.AddCommand(imageCmd)
 }
 
+//nolint: gocyclo
 func runImageCmd() (err error) {
 	p, err := platform.NewPlatform(options.Platform)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/installer/cmd/image.go
```diff
@@ -83,6 +84,16 @@ func runImageCmd() (err error) {
 		},
 	}
 
+	if options.ConfigSource == "" {
+		switch p.Name() {
+		case "aws", "azure", "digital-ocean", "gcp":
```
andrewrynhard: We eventually need a better way to get platform names, but this is nothing new compared to what we are doing in other places in the code.

bradbeam: This allows the user to misconfigure how userdata is retrieved. Each platform has a specific way and only a specific way to fetch userdata

Aws: https://github.com/talos-systems/talos/blob/master/internal/pkg/runtime/platform/aws/aws.go#L126
Gcp: https://github.com/talos-systems/talos/blob/master/internal/pkg/runtime/platform/gcp/gcp.go#L38
Vmware: https://github.com/talos-systems/talos/blob/master/internal/pkg/runtime/platform/vmware/vmware.go#L36

Metal is the only one that offers the flexibility to download userdata from an arbitrary location, but we don't create images for metal.

andrewrynhard: Yeah that is a known bug for a while now. I can follow up with a PR to fix that.

cmd/installer/cmd/root.go
```diff
@@ -33,7 +33,8 @@ func Execute() {
 var options = &pkg.InstallOptions{}
 
 func init() {
-	rootCmd.PersistentFlags().StringVar(&options.ConfigSource, "config", "none", "The value of "+constants.KernelParamConfig)
+	rootCmd.PersistentFlags().StringVar(&options.ConfigSource, "config", "", "The value of "+constants.KernelParamConfig)
+	rootCmd.PersistentFlags().MarkHidden("config")
```
andrewrynhard: We can hide the flag as we sort this out.

cmd/installer/cmd/image.go
```diff
@@ -83,6 +85,16 @@ func runImageCmd() (err error) {
...
+			options.ConfigSource = "none"
```
andrewrynhard: We should really rename this. Maybe `default` instead. Something for the future.

PR #1716, opened 2020-01-02
feat: support configurable docker-based clusters

rsmitty:
This PR will allow users to issue `osctl config generate`, tweak the
configs to their liking, then use those configs to call `osctl cluster
create`.

Example workflow:

```
osctl config generate my-cluster https://10.5.0.2:6443 -o ./my-cluster

** tweaky tweak **

osctl cluster create --name my-cluster --input-dir "$PWD/my-cluster"
```

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -6,13 +6,18 @@ package cmd
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"net"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
+	yaml "gopkg.in/yaml.v2"
 
 	"github.com/talos-systems/talos/cmd/osctl/pkg/client/config"
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
@@ -38,6 +43,7 @@ var (
 	clusterWaitTimeout      time.Duration
 	forceInitNodeAsEndpoint bool
 	forceEndpoint           string
+	inputDir                string
 )
 
 // clusterCmd represents the cluster command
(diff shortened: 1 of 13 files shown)


internal/pkg/provision/request.go
```diff
@@ -30,10 +30,10 @@ type NetworkRequest struct {
 }
 
 // NodeRequests is a list of NodeRequest.
-type NodeRequests []NodeRequest
+type NodeRequests []*NodeRequest
```
smira: is there a strong case for using pointer here?

Go is really efficient at copying data structures around, and I guess it might skip copying if it can prove it's safe.

Having copy of the data eliminates lots of small cases here and there. 

Of course there's a probably a lot of cases when having a pointer makes sense (like Mutex being a field), but is there anything like this here?

internal/pkg/provision/providers/docker/create.go
```diff
@@ -35,16 +40,14 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
...
-	}
+	b64Configs := map[string]string{"Init": "", "ControlPlane": "", "Join": ""}
+	talosConfig := &config.Config{}
 
+	endpoints := []string{"127.0.0.1"}
 	if options.ForceEndpoint != "" {
-		input.AdditionalSubjectAltNames = append(input.AdditionalSubjectAltNames, options.ForceEndpoint)
-		input.AdditionalMachineCertSANs = append(input.AdditionalMachineCertSANs, options.ForceEndpoint)
+		endpoints = []string{options.ForceEndpoint}
+	} else if options.ForceInitNodeAsEndpoint {
```
smira: Feels like this whole piece of code which deals with config generation (or reading existing config) can be moved out of the provisioner, as `Create()` function expects `ClusterRequest` which now contains config data (in `NodeRequest`), so Provisioner should get it pre-filled I guess. 

If there's any kind of dependency on config I think we can pass it in a parsed format in `NodeRequest`.

internal/pkg/provision/request.go
rsmitty: I was just having trouble tweaking the configdata value in the node requests without it being a pointer. Let me take a second look.

cmd/osctl/cmd/cluster.go
```diff
@@ -110,6 +116,78 @@ func create(ctx context.Context) (err error) {
 
 	defer provisioner.Close() //nolint: errcheck
 
+	b64Configs := map[string]string{"Init": "", "ControlPlane": "", "Join": ""}
```
andrewrynhard: I don't think a map is the best way for this. Feels a little fragile. Can we maybe have custom type that encapsulates all this? A simple struct with methods to either populate itself using generated data, or data read from files on disk would be nice.

andrewrynhard:
@rsmitty Can you rebase? We can merge this when you do.

jsimonetti:
This PR works as expected in single master setups (default in docker).
As discussed in slack, this however fails for multiple masters.

PR #1720, opened 2020-01-04
chore: update all target in Makefile

andrewrynhard:
We should build the most common things by default.

### DIFF
--- a/hack/drone.jsonnet
+++ b/hack/drone.jsonnet
@@ -179,11 +179,9 @@ local apid = Step("apid", depends_on=[setup_ci]);
 local osctl_linux = Step("osctl-linux", depends_on=[setup_ci]);
 local osctl_darwin = Step("osctl-darwin", depends_on=[setup_ci]);
 local docs = Step("docs", depends_on=[osctl_linux]);
-local integration_test = Step("integration-test", depends_on=[setup_ci]);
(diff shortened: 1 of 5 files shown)


jsimonetti:
Since everything is now outputted to `_out` directory, is it possible to add automatic target dependencies again?

So one would not have to remember to `make apps` before `make talos`, etc.

andrewrynhard:
> Since everything is now outputted to `_out` directory, is it possible to add automatic target dependencies again? (In other words, getting rid of most of the .PHONY targets)
> 
> So one would not have to remember to `make apps` before `make talos`, etc.

Yeah. Pushed and update. The `talos` and `installer` targets now depend on `apps`. Buildkit will automatically rebuild the apps if they need to be.

hack/drone.jsonnet
```diff
@@ -179,11 +179,9 @@ local apid = Step("apid", depends_on=[setup_ci]);
 local osctl_linux = Step("osctl-linux", depends_on=[setup_ci]);
 local osctl_darwin = Step("osctl-darwin", depends_on=[setup_ci]);
 local docs = Step("docs", depends_on=[osctl_linux]);
-local integration_test = Step("integration-test", depends_on=[setup_ci]);
```
andrewrynhard: These changes are necessary to prevent the app tarballs from being corrupted. Each on of these steps now depends on `apps`, and the `apps` target outputs tarballs. If all run in parallel then they start to compete in the write and subsequent steps end up getting corrupted tarballs

Issue #1722, opened 2020-01-06
Improve ISO installer workflow


rsmitty:
We talked about exploring bi-directional grpc to create a small service for ISO installer that would allow osctl to interact and inject configs for the iso install.

PR #1723, opened 2020-01-06
chore: disable iso artifact creation

### LINKED ISSUE #1442: Remove ISO installer from build artifacts

rsmitty:
This PR will disable iso creation for now. We plan to reincorporate the
ability to use ISOs once we've researched #1722.

Will close #1442

### DIFF
--- a/hack/drone.jsonnet
+++ b/hack/drone.jsonnet
@@ -422,7 +422,6 @@ local release = {
       '_out/installer.tar',
       '_out/osctl-darwin-amd64',
       '_out/osctl-linux-amd64',
-      '_out/talos.iso',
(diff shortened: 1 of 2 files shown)


hack/drone.jsonnet
```diff
@@ -426,7 +425,6 @@ local release = {
       '_out/installer.tar',
       '_out/osctl-darwin-amd64',
       '_out/osctl-linux-amd64',
-      '_out/talos.iso',
```
andrewrynhard: I think we should keep everything else. This line should be enough to not publish it.

rsmitty: k - amended.

PR #1724, opened 2020-01-07
fix: catch SIGINT and SIGTERM

andrewrynhard:
We need to catch these signals and shutdown machined gracefully.


andrewrynhard:
We may not need this after all. Looks like we already handle this in a task.

PR #1730, opened 2020-01-08
refactor: use an interface for config data

rsmitty:
This PR will move to using a `ConfiguratorBundle` interface for our
various config data files. This will allow us to easily abstract away
various versions and easily get the data with functions.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -9,22 +9,19 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"math/big"
 	"net"
-	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
-	yaml "gopkg.in/yaml.v2"
 
-	"github.com/talos-systems/talos/cmd/osctl/pkg/client/config"
+	clientconfig "github.com/talos-systems/talos/cmd/osctl/pkg/client/config"
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
 	"github.com/talos-systems/talos/internal/pkg/provision"
 	"github.com/talos-systems/talos/internal/pkg/provision/access"
 	"github.com/talos-systems/talos/internal/pkg/provision/check"
 	"github.com/talos-systems/talos/internal/pkg/provision/providers/docker"
+	"github.com/talos-systems/talos/pkg/config"
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1"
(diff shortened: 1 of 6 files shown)


cmd/osctl/cmd/cluster.go
```diff
@@ -140,29 +141,22 @@ func create(ctx context.Context) (err error) {
...
+			switch configType {
+			case generate.TypeInit:
+				configBundle.InitCfg = generatedConfig
+			case generate.TypeControlPlane:
+				configBundle.ControlPlaneCfg = generatedConfig
+			case generate.TypeJoin:
+				configBundle.JoinCfg = generatedConfig
 			}
-
-			b64Configs[key] = base64.StdEncoding.EncodeToString([]byte(data))
```
andrewrynhard: I think we should have the v1alpha1.ConfigBundle encapsulate all this logic in `NewConfigBundle()`. That way we only need to initialize it with one line here, and if there is ever the case in the future where we may want to generate a new bundle, we don't have it all hidden away in this cobra command.

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -9,10 +9,20 @@ package v1alpha1
...
 
+// ConfigBundle defines the group of v1alpha1 config files.
+//docgen: nodoc
+type ConfigBundle struct {
+	InitCfg         *Config
+	ControlPlaneCfg *Config
+	JoinCfg         *Config
+	TalosCfg        *config.Config
+}
+
```
andrewrynhard: super nit: should this be in `v1alpha1_configurator_bundle.go`?

cmd/osctl/cmd/cluster.go
```diff
@@ -212,10 +165,22 @@ func create(ctx context.Context) (err error) {
...
 			typ = generate.TypeInit
+			configDataStruct = configBundle.InitCfg
 		} else {
 			typ = generate.TypeControlPlane
+			configDataStruct = configBundle.ControlPlaneCfg
+		}
+
+		configDataString, err = configDataStruct.String()
+		if err != nil {
+			return err
```
andrewrynhard: Not something for this PR, but I think we can get this to use the interface methods. I would like to tackle that just to get a feel for things if you don't mind?

rsmitty: oh yeah, we should be getting those with the methods for sure. whoops. I'll let you fix after we merge.

pkg/config/types/v1alpha1/v1alpha1_types.go
rsmitty: moved in amended commit.

PR #1738, opened 2020-01-09
feat: add a basic architectural diagram and a call to action

tgerla:
- add an architecture diagram
- add a call-to-action button on the front page
- clean up some lint issues
- adjust the way the "chips" image is displayed
- move the K8s certified logo to the "Features" section


andrewrynhard:
A few things stick out to me. The new chips layout feels like a 1990's background because you can see that it is just an image repeated multiple times. The certified k8s logo feels out of place in the features section, although I don't have an opinion on where it should go otherwise. Maybe that is worthy of its' own little section?

tgerla:
Good points on both. I wasn't sure where to put the CNCF thing. Let me tackle fixing the repeating "chips" and I will experiment with the CNCF logo placement.

PR #1745, opened 2020-01-09
 test: provision Talos clusters via Firecracker VMs

smira:
This is initial PR to push the initial code, it has several known
problems which are going to be addressed in follow-up PRs:

1. there's no "cluster destroy", so the only way to stop the VMs is to
`pkill firecracker`

2. provisioner creates state in `/tmp` and never deletes it, that is
required to keep cluster running when `osctl cluster create` finishes

3. doesn't run any controller process around firecracker to support
reboots/CNI cleanup (vethxyz interfaces are lingering on the host as
they're never cleaned up)

The plan is to create some structure in `~/.talos` to manage cluster
state, e.g. `~/.talos/clusters/<name>` which will contain all the
required files (disk images, file sockets, VM logs, etc.). This
directory structure will also work as a way to detect running clusters
and clean them up.

For point number 3, `osctl cluster create` is going to exec lightweight
process to control the firecracker VM process and to simulate VM reboots
if firecracker finishes cleanly (when VM reboots).


smira:
Example command: 

```sudo ~/.local/bin/osctl cluster create --provisioner firecracker --cidr 172.19.0.0/24 --install-image docker.io/autonomy/installer:v0.4.0-alpha.1  --memory 1536 --masters 3 --workers 3 --init-node-as-endpoint --wait```

More info in README on the prerequisites.

File: cmd/osctl/cmd/cluster.go [hunk 1]
```diff
@@ -84,29 +93,40 @@ func create(ctx context.Context) (err error) {
...
+	var hostIP net.IP
+
+	hostIP, err = talosnet.NthIPInNetwork(cidr, 1)
+	if err != nil {
+		return err
+	}
+
 	// Set starting ip at 2nd ip in range, ex: 192.168.0.2
-	ips := make([]string, masters)
+	ips := make([]net.IP, masters+workers)
```
smira: firecracker provisioner doesn't have any way to allocate IPs on its own, so I went ahead and allocated all the IPs in the `osctl` for both docker & firecracker

cmd/osctl/cmd/cluster.go
```diff
@@ -84,29 +93,40 @@ func create(ctx context.Context) (err error) {
...
-	fmt.Println("validating CIDR and reserving master IPs")
+	fmt.Println("validating CIDR and reserving IPs")
 
 	_, cidr, err := net.ParseCIDR(networkCIDR)
 	if err != nil {
 		return fmt.Errorf("error validating cidr block: %w", err)
 	}
 
+	// Host addr at 1st IP in range, ex. 192.168.0.1
+	var hostIP net.IP
```
smira: probably should rename to `gatewayIP` ?

cmd/osctl/cmd/cluster.go
```diff
@@ -119,20 +139,23 @@ func create(ctx context.Context) (err error) {
 	if inputDir != "" {
 		configBundleOpts = append(configBundleOpts, config.WithExistingConfigs(inputDir))
 	} else {
-		var genOptions []generate.GenOption
+		genOptions := []generate.GenOption{
+			generate.WithInstallImage(nodeInstallImage),
+			generate.WithInstallDisk("/dev/vda"), // TODO: fixme, firecracker specific
```
smira: not sure how to go with this one, probably provisioner-specific GenOptions

Dockerfile
```diff
@@ -178,9 +178,10 @@ ENTRYPOINT ["/networkd"]
 FROM base AS osctl-linux-build
 ARG SHA
 ARG TAG
+ARG ARTIFACTS
```
smira: this is to provide default path to `vmlinux` & `initramfs.xz` in `osctl`

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 replace (
 	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+	github.com/firecracker-microvm/firecracker-go-sdk v0.19.0 => github.com/smira/firecracker-go-sdk v0.19.1-0.20200110185541-4fce8cba9f84
```
smira: this is my fork with fixes, I've got one merged and two are outstanding:

* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/167
* https://github.com/firecracker-microvm/firecracker-go-sdk/issues/173

internal/pkg/provision/providers/firecracker/node.go
```diff
@@ -0,0 +1,195 @@
...
+	cmdline.Append("ip", fmt.Sprintf(
+		"%s::%s:%s:%s:eth0:off",
+		nodeReq.IP,
+		clusterReq.Network.GatewayAddr,
+		net.IP(clusterReq.Network.CIDR.Mask),
+		nodeReq.Name))
+
+	ones, bits := clusterReq.Network.CIDR.IP.DefaultMask().Size()
+
+	cfg := firecracker.Config{
```
smira: from this line down I want to extract into separate process which receives `firecracker.Config` serialized and uses it to run the firecracker VM

internal/pkg/provision/providers/firecracker/node.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	defer os.Remove(cfg.SocketPath) //nolint: errcheck
+
+	logFile, err := os.OpenFile(filepath.Join(state.tempDir, fmt.Sprintf("%s.log", nodeReq.Name)), os.O_APPEND|os.O_CREATE|os.O_RDWR, 0666)
+	if err != nil {
+		return provision.NodeInfo{}, err
+	}
+
+	// defer logFile.Close() //nolint: errcheck
```
smira: need to get it back once separate control process is implemented

internal/pkg/provision/providers/firecracker/node.go
```diff
@@ -0,0 +1,195 @@
...
+
+	for i := 0; i < 2; i++ {
```
smira: add a comment

PR #1747, opened 2020-01-09
fix: check for installer image before proceeding with upgrade

### LINKED ISSUE #1606: Fail fast on upgrade if image is unavailable
Currently we get pretty far into an upgrade before we attempt to pull the upgrade container. With the need to specify the image with `docker.io` or 'registry', this can inadvertently break a node. 
We should attempt to pull the image / verify the image prior to initiating the upgrade.

rsmitty:
This PR will add in some code to pre-pull the installer image before we
run an upgrade of a given talos node. Additionally, this will add some
functional args to the install package to allow for specifying whether
or not to pull the installer image. This was needed since there was no
sense in pulling the installer again once we made it that far into the
upgrade process.

Will close #1606 

### DIFF
--- a/internal/app/machined/internal/api/reg/reg.go
+++ b/internal/app/machined/internal/api/reg/reg.go
@@ -20,6 +20,8 @@ import (
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
 
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
 	criconstants "github.com/containerd/cri/pkg/constants"
 
 	"github.com/talos-systems/talos/api/common"
@@ -27,7 +29,7 @@ import (
 	machineapi "github.com/talos-systems/talos/api/machine"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system"
 	"github.com/talos-systems/talos/internal/pkg/containers"
-	"github.com/talos-systems/talos/internal/pkg/containers/containerd"
+	taloscontainerd "github.com/talos-systems/talos/internal/pkg/containers/containerd"
 	"github.com/talos-systems/talos/internal/pkg/containers/cri"
 	"github.com/talos-systems/talos/internal/pkg/etcd"
 	"github.com/talos-systems/talos/internal/pkg/event"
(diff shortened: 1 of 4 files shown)


internal/pkg/install/options.go
```diff
@@ -0,0 +1,29 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package install
+
+// Option controls generate options specific to input generation.
+type Option func(o *Options) error
+
+// WithExistingImage specifies whether to pull the installer image during an install
+func WithExistingImage() Option {
```
andrewrynhard: nit: Should we make this `WithPullImage` if we want to use `PullImage` as the field? If we want `WithExistingImage`, maybe we should make the field `ExistingImage` or something like that. At the end of the day, I don't care either way, but I feel like there is some double negation type thinking this forces me to do since `WithExistingImage` sort of inverts what the field `PullImage` makes you think. Can you tell I'm confused? :D

internal/pkg/install/install.go
```diff
@@ -21,17 +21,34 @@ import (
...
-		return err
+	var image containerd.Image
+
+	if options.PullImage {
+		image, err = client.Pull(ctx, r.Config().Machine().Install().Image(), []containerd.RemoteOpt{containerd.WithPullUnpack}...)
+		if err != nil {
+			return err
+		}
+	} else {
+		image, err = client.GetImage(ctx, r.Config().Machine().Install().Image())
```
andrewrynhard: Good idea. Brings some assurance that the image exists.

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -97,7 +98,18 @@ func (r *Registrator) Shutdown(ctx context.Context, in *empty.Empty) (reply *mac
 
 // Upgrade initiates an upgrade.
 func (r *Registrator) Upgrade(ctx context.Context, in *machineapi.UpgradeRequest) (data *machineapi.UpgradeResponse, err error) {
-	if err = etcd.ValidateForUpgrade(); err != nil {
```
andrewrynhard: I think this should still be here.

rsmitty: yep, thx.

PR #1750, opened 2020-01-10
fix: shutdown on button/power ACPI event

### LINKED ISSUE #1748: ACPI event needs to be verified for power off
Right now we just power off a machine on any api event. We should verify that the event is a power off event.

andrewrynhard:
This adds a check to ensure that only a button/power ACPI event can
cause a shutdown of a node.

Fixes #1748.

### DIFF
--- a/internal/app/machined/internal/phase/acpi/acpi.go
+++ b/internal/app/machined/internal/phase/acpi/acpi.go
@@ -9,12 +9,21 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/mdlayher/genetlink"
+	"github.com/mdlayher/netlink"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/pkg/event"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
+)
+
+const (
+	// PowerButtonEvent is the ACPI event name associated with the power off
+	// button.
+	PowerButtonEvent = "button/power"
 )
 
 // Handler represents the ACPI handler task.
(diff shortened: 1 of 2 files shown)


internal/app/machined/internal/phase/acpi/acpi.go
```diff
@@ -87,13 +96,42 @@ func listenForPowerButton() (err error) {
...
 	}()
 
 	return nil
 }
+
+func parse(msgs []genetlink.Message, event string) (bool, error) {
+	for _, msg := range msgs {
+		ad, err := netlink.NewAttributeDecoder(msg.Data)
+		if err != nil {
+			return false, fmt.Errorf("failed to create attribute decoder: %w", err)
```
smira: should we continue here, as other messages won't be parsed and they might be the message we're looking for?

andrewrynhard: good point ... probably need multi error too then.

internal/app/machined/internal/phase/acpi/acpi.go
```diff
@@ -9,14 +9,23 @@ import (
...
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/pkg/event"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
 )
 
+const (
+	// PowerButtonEvent is the ACPI event name associated with the power off
+	// button.
+	PowerButtonEvent = "button/power"
```
bradbeam: Is button/power the only one we need to handle?

andrewrynhard: As far as I could tell. The docs on ACPI events are practically non-existent, or just hard to find.

PR #1755, opened 2020-01-10
fix(networkd): Fix incorrect resolver settings

bradbeam:
This  modifies the way the hostname gets set. Previously, we would run
through the entire addressing and resolver configuration and then set the
hostname. This is problematic because the resolver depends on the functionality
of Hostname() ( resolver configuration relies on the domainname of the host ).

### DIFF
--- a/internal/app/networkd/pkg/networkd/networkd.go
+++ b/internal/app/networkd/pkg/networkd/networkd.go
@@ -243,11 +243,22 @@ func (n *Networkd) Configure() (err error) {
 		}
 	}
 
+	// Set hostname must be before the resolv configuration
+	// so we can ensure the hosts domainname is set properly
+	// before we write the search stanza
+	if err = n.Hostname(); err != nil {
+		return err
+	}
+
 	if len(resolvers) == 0 {
 		resolvers = n.resolvers
 	}
 
-	return writeResolvConf(resolvers)
+	if err = writeResolvConf(resolvers); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 // Renew sets up a long running loop to refresh a network interfaces
@@ -284,14 +295,24 @@ func (n *Networkd) Hostname() (err error) {
 		return err
 	}
 
+	var p runtime.Platform
+
+	p, err = platform.CurrentPlatform()
+	if err != nil {
+		return err
+	}
+
+	// Skip hostname/domainname setting when running in container mode
+	if p.Mode() == runtime.Container {
+		return nil
+	}
+
 	if err = unix.Sethostname([]byte(hostname)); err != nil {
 		return err
 	}
 
-	if domainname != "" {
-		if err = unix.Setdomainname([]byte(domainname)); err != nil {
-			return err
-		}
+	if err = unix.Setdomainname([]byte(domainname)); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -264,6 +276,10 @@ func (n *Networkd) Reset() {
 	for _, iface := range n.Interfaces {
 		iface.Reset()
 	}
+
+	// Empty out resolv.conf
+	//nolint: errcheck
+	ioutil.WriteFile("/etc/resolv.conf", []byte{}, 0644)
```
andrewrynhard: Maybe we should log the error here?

andrewrynhard: To be consistent with how we are doing it in other places ... something like:

```go
log.Printf("WARNING: failed to ....: %v", err)
```

bradbeam: id rather address it when we get better logging in place altogether. `iface.Reset()` can throw errors as well, but we silently ignore them for now

andrewrynhard: A simple log here wouldn't hurt, and I'm not sure what "better logging" buys us other than maybe a convenience `log.Warn` function. If this fails to write, we will banging our heads against the wall trying to figure out why `/etc/resolv.conf` isn't correct.

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -285,11 +297,15 @@ func (n *Networkd) Hostname() (err error) {
 	}
 
 	if err = unix.Sethostname([]byte(hostname)); err != nil {
-		return err
+		// Workaround for container based setups
+		if err != syscall.EPERM {
+			return err
+		}
```
andrewrynhard: I know we have done work to fine tune permissions, and this is unlikely to be an issue as long as we don't change permissions, but I feel strongly that a better route would be to skip setting the hostname and domain name if we are running in container mode. I can think of two ways to do this:

1. We add a flag to networkd that we can use to determine the runtime mode
2. We load the config and parse it to get the `runtime.Configurator`

internal/app/networkd/pkg/networkd/networkd.go
bradbeam: I ended up dropping this since it's not necessary.

internal/app/networkd/pkg/networkd/networkd.go
bradbeam: If we ignored EPERM for the entire networkd stack I'd agree it's too wide of a net. But we are ignoring it only for setting the domain name and hostname, two specific syscalls. I think this is reasonably tight. 

Having some notion of modes for networkd to run in feels like the wrong approach because it introduces another code branch which would introduce unnecessary complexity. Really the only thing we want from networkd in 'container' mode is setting up resolv.conf to prevent a loop error with coredns. So from that perspective I'd rather just drop networkd from the list of container services and bring back out patch for coredns to explicitly set the resolvers.

andrewrynhard: Setting the hostname, and domain name, as trivial as they sound, are _extremely_ important. PKI is configured this way. Ignoring any kind of error on these should never happen.

I do like the idea of not running `networkd` in container mode. Especially with the work @smira is doing with firecracker. Remind me, how did we patch coredns? Feels like that was pre `osctl cluster create`. We would have to build that in somewhere in `osctl`. That makes it a little more ugly 🤔.

PR #1757, opened 2020-01-10
refactor: use ConfiguratorBundle interface for config generate

rsmitty:
This PR overhauls osctl's config generate command to make use of the new
ConfigBundle implementation of the ConfiguratorBundle interface

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -121,11 +121,6 @@ func create(ctx context.Context) (err error) {
 		}
 	}
 
-	masterIPs := make([]string, 0, masters)
-	for _, ip := range ips[:masters] {
-		masterIPs = append(masterIPs, ip.String())
-	}
-
 	provisioner, err := providers.Factory(ctx, provisioner)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


internal/app/machined/internal/phase/platform/platform.go
```diff
@@ -56,9 +56,6 @@ func (task *Platform) runtime(r runtime.Runtime) (err error) {
 	}
 
 	if r.Platform().Mode() == runtime.Container {
-		// TODO: add ::1 back once I figure out why bootkube barfs
```
rsmitty: We handle injecting this in cluster.go now. Was causing duplicates in machine sans.

cmd/osctl/cmd/cluster.go
```diff
@@ -139,24 +134,25 @@ func create(ctx context.Context) (err error) {
...
 		if forceEndpoint != "" {
-			genOptions = append(genOptions, generate.WithEndpointList([]string{forceEndpoint}))
+			endpointList = append(endpointList, forceEndpoint)
 			provisionOptions = append(provisionOptions, provision.WithEndpoint(forceEndpoint))
 		} else if forceInitNodeAsEndpoint {
-			genOptions = append(genOptions, generate.WithEndpointList([]string{ips[0].String()}))
+			endpointList = append(endpointList, ips[0].String())
 		}
 
+		endpointList = append(endpointList, "127.0.0.1")
```
andrewrynhard: nit: We should have a comment here prefixed with `NB:` to indicate that it is crucial to set this _last_: https://en.wikipedia.org/wiki/Nota_bene .

smira: I wonder if it should be more of `switch`-like logic here? As right now it appends to the list, while it was supposed to be replacing the list depending on the config? 

Once we do client LB, having 127.0.0.1 while it's not running current cluster might lead to fun results.

andrewrynhard: I _think_ we append only for the purposes of getting a valid cert SANs? @rsmitty is that right?

rsmitty: fixed in amend.

PR #1759, opened 2020-01-11
refactor: change machine type names

andrewrynhard:
BREAKING CHANGE: A machine type of "join" is no longer valid. Use
"worker" instead.

### DIFF
--- a/pkg/config/machine/machine.go
+++ b/pkg/config/machine/machine.go
@@ -17,27 +17,36 @@ import (
 type Type int
 
 const (
-	// TypeInit represents an init node.
-	TypeInit Type = iota
+	// TypeBootstrap represents a bootstrap node.
+	TypeBootstrap Type = iota
 	// TypeControlPlane represents a control plane node.
 	TypeControlPlane
 	// TypeWorker represents a worker node.
 	TypeWorker
 )
 
+const (
+	// TypeBootstrapString is the string representation of TypeBootstrap.
+	TypeBootstrapString = "bootstrap"
+	// TypeControlPlaneString is the string representation of TypeControlPlane.
+	TypeControlPlaneString = "controlplane"
+	// TypeWorkerString is the string representation of TypeWorker.
+	TypeWorkerString = "worker"
+)
+
 // String returns the string representation of Type.
 func (t Type) String() string {
-	return [...]string{"Init", "ControlPlane", "Join"}[t]
+	return [...]string{TypeBootstrapString, TypeControlPlaneString, TypeWorkerString}[t]
 }
 
 // ParseType parses string constant as Type
 func ParseType(t string) (Type, error) {
 	switch t {
-	case "Init":
-		return TypeInit, nil
-	case "ControlPlane":
+	case TypeBootstrap.String():
+		return TypeBootstrap, nil
+	case TypeControlPlane.String():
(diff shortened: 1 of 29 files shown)


pkg/config/machine/machine.go
```diff
@@ -27,17 +27,17 @@ const (
...
 func ParseType(t string) (Type, error) {
 	switch t {
-	case "Init":
+	case "init":
 		return TypeInit, nil
-	case "ControlPlane":
+	case "controlplane":
 		return TypeControlPlane, nil
-	case "Join":
+	case "worker":
```
andrewrynhard: This is now the source of truth for all things that use a machine's type in any way.

pkg/config/types/v1alpha1/v1alpha1_configurator.go
```diff
@@ -108,14 +108,12 @@ func (m *MachineConfig) Files() []machine.File {
...
-		return machine.TypeInit
-	case "controlplane":
-		return machine.TypeControlPlane
-	default:
+	t, err := machine.ParseType(m.MachineType)
+	if err != nil {
 		return machine.TypeWorker
 	}
+
+	return t
```
andrewrynhard: Today we do no validation of `machine.type`, we just default to `TypeWorker` if `"init", or `"controlplane"` are not specified. That means that, today, users can throw just about anything they want in this field and it is still a worker. This fixes that.

pkg/config/types/v1alpha1/v1alpha1_configurator.go
```diff
@@ -108,14 +108,12 @@ func (m *MachineConfig) Files() []machine.File {
 
 // Type implements the Configurator interface.
 func (m *MachineConfig) Type() machine.Type {
-	switch m.MachineType {
-	case "init":
-		return machine.TypeInit
-	case "controlplane":
-		return machine.TypeControlPlane
-	default:
+	t, err := machine.ParseType(m.MachineType)
```
andrewrynhard: I don't know of a good way to prevent having to return an error here. If we change the signature, things get super ugly in places we need to know the machine type. I was thinking that we could do `machine.ParseType` in the validation function, that way we know that we will not get an error here, and we can ignore it.

Ulexus:
Generally :+1: Also "bootstrap" > "init" to my mind, though I wonder if tags might be a better approach, so that you could have a node which was both "controlplane" _and_ "bootstrap".  Or, more generally "etcd", "controlplane", "bootstrap", "worker" because, depending on the system, one node might very well have all or any of these in fairly orthogonal fashion.  (Not as Talos stands now, of course, but a priori.)

andrewrynhard:
> Generally 👍 Also "bootstrap" > "init" to my mind, though I wonder if tags might be a better approach, so that you could have a node which was both "controlplane" _and_ "bootstrap". Or, more generally "etcd", "controlplane", "bootstrap", "worker" because, depending on the system, one node might very well have all or any of these in fairly orthogonal fashion. (Not as Talos stands now, of course, but a priori.)

I do like bootstrap better than init. Much more clear about what it does. And we often use this word in our meetings.

rsmitty:
I don't particularly mind the rename, but this feels like prime territory for moving to v1alpha2 if it's gonna cause breaking changes to current v1alpha1 configs that users may have.

PR #1761, opened 2020-01-11
chore: run sonobuoy in quick mode

andrewrynhard:
This adds sonobuoy's quick mode test to basic integration.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,8 +44,6 @@ steps:
   image: autonomy/build-container:latest
   commands:
   - make machined
-  environment:
-    BINDIR: /usr/local/bin
   volumes:
   - name: dockersock
     path: /var/run
@@ -65,8 +63,6 @@ steps:
   image: autonomy/build-container:latest
   commands:
   - make osd
-  environment:
-    BINDIR: /usr/local/bin
   volumes:
   - name: dockersock
     path: /var/run
@@ -86,8 +82,6 @@ steps:
   image: autonomy/build-container:latest
   commands:
   - make apid
-  environment:
-    BINDIR: /usr/local/bin
   volumes:
   - name: dockersock
     path: /var/run
@@ -107,8 +101,6 @@ steps:
   image: autonomy/build-container:latest
   commands:
   - make trustd
-  environment:
-    BINDIR: /usr/local/bin
   volumes:
   - name: dockersock
     path: /var/run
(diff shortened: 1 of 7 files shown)


.drone.yml
```diff
@@ -526,6 +526,7 @@ steps:
   - make basic-integration
   environment:
     BINDIR: /usr/local/bin
+    TALOS_PLATFORM: docker
```
andrewrynhard: Had to change to this, otherwise I got:

```
+ make basic-integration
2	make[1]: Entering directory '/drone/src'
3	make[2]: Entering directory '/drone/src'
4	"docker": unknown operating system or architecture: invalid argument
5	make[2]: *** [Makefile:78: target-integration-test-linux] Error 1
6	make[2]: Leaving directory '/drone/src'
7	make[1]: Leaving directory '/drone/src'
8	make[1]: *** [Makefile:84: local-integration-test-linux] Error 2
9	make: *** [Makefile:189: integration-test-linux] Error 2
```

Not sure why, but it looks related to containerd. My suspicion is that containerd respects `PLATFORM` env var because it is used in buildkit.

hack/test/capi.sh
```diff
@@ -24,10 +24,6 @@ cat ${PWD}/hack/test/manifests/capg-components.yaml| envsubst > ${TMP}/capg-comp
 ##They've got an incorrect image pull policy.
 ##curl -L ${CAPG_COMPONENTS} | envsubst > ${TMP}/capg-components.yaml
 
-## Clean up osctl endpoint, fetch kubeconfig
-e2e_run "osctl config endpoint 10.5.0.2"
```
smira: this line is required I believe, as it switches endpoint from 127.0.0.1 (when running outside of docker network) to 10.5.0.2 (when running inside docker network)

andrewrynhard: This seemed to work without this.

andrewrynhard: Do you know when we needed this? For CI? Or local?

smira: For CI. `basic-integration` should leave the address in config in a way it works for pod-to-pod communication (from build step pod to docker service pod), while e2e uses`docker run`, so it runs inside a docker service pod and inside a docker network, and pod IP won't be accessible there iirc.

andrewrynhard: Ok. Doesn't hurt to put it back. I will do that 👍.

PR #1766, opened 2020-01-12
fix: allow only one sequence at a time

### LINKED ISSUE #1448: Add local upgrade lock
A node needs a way to know that it is being upgraded. If a user makes an upgrade request more than once, who knows what could happen?

andrewrynhard:
This adds a lock to the v1alpha1 sequencer that only alows one sequence
to run at a time. This fixes a number of edge cases:

- running multiple upgrades at once
- upgrading before the machine is fully booted
- prevent shutdowns before the machine is fully booted

Fixes #1448.

### DIFF
--- a/internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
+++ b/internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
@@ -6,6 +6,7 @@ package v1alpha1
 
 import (
 	"fmt"
+	"sync/atomic"
 
 	machineapi "github.com/talos-systems/talos/api/machine"
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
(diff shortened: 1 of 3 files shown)


internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -28,221 +28,261 @@ import (
...
 
-	var dev *probe.ProbedBlockDevice
+	return s.run(f)
+}
 
-	dev, err = probe.GetDevWithFileSystemLabel(constants.EphemeralPartitionLabel)
-	if err != nil {
-		return err
+func (s *Sequencer) run(f func() error) error {
+	if s.locked {
```
smira: I guess we need to protect `locked` with a mutex to be on a safe side

andrewrynhard: Ahh. Yep. Good point.

andrewrynhard:
So in practice this is what happens:

- issue upgrade command, and can see upgrade being performed
- issue another upgrade command, can't see any log about upgrade only being allowed once, because of the mutex (it blocks)
- upgrade completes and then you can see `failed to run upgrade sequence: sequence can only be performed once`

```
147.75.64.133: [talos] performing upgrade via "docker.io/andrewrynhard/installer:latest"
147.75.64.133: [talos] [phase]: upgrade done, 1.456731477s
147.75.64.133: [talos] failed to run upgrade sequence: sequence can only be performed once
147.75.64.133: [talos] [phase]: stop services
```

smira:
sorry, @andrewrynhard I meant to keep `.locked` variable, but to add mutex around reading/writing it (only when it's touched), main sequencer code should still run without a mutex

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -28,221 +30,266 @@ import (
...
 	}
 
-	var dev *probe.ProbedBlockDevice
+	return s.run(f)
+}
 
-	dev, err = probe.GetDevWithFileSystemLabel(constants.EphemeralPartitionLabel)
-	if err != nil {
-		return err
+func (s *Sequencer) Locked() bool {
```
smira: :+1: :zap: :cool: 

idk, if method name is self-descriptive, or should it be `TryLock` or something? (nit)

andrewrynhard: yeah that sounds way better ... I'm all for good naming

internal/app/machined/main.go
```diff
@@ -157,31 +158,48 @@ func main() {
...
+				if unix.Reboot(rebootFlag) == nil {
+					select {}
+				}
+			}()
 		case event.Upgrade:
-			var (
-				req *machineapi.UpgradeRequest
-				ok  bool
-			)
+			go func() {
```
andrewrynhard: Prevents upgrade requests from queuing up.

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	machineapi "github.com/talos-systems/talos/api/machine"
+)
+
+type MockSequencer struct {
+	Sequencer
```
andrewrynhard: Embed the default sequencer so we can override the methods and still be able to test the `run` func.

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -28,221 +29,269 @@ import (
...
+func (s *Sequencer) run(f func() error) error {
+	if s.TryLock() {
+		return ErrLocked{}
 	}
 
-	devname := dev.Device().Name()
+	defer s.Unlock()
 
-	if err := dev.Close(); err != nil {
+	if err := f(); err != nil {
```
smira:
```suggestion
	return f()
```

internal/app/machined/main.go
smira: on other hand, if we pass all the events which launch sequences via the event bus, doesn't it mean we effectively achieve one-at-a-time lock?

we could do something like

```go
func (bus) TryPublish(timeout, event) error {
   select {
   case bus <- event:
      return nil
   case <-time.After(timeout):
     return fmt.Errorf("bus is locked by another activity, please try later)
   }
}
```

internal/app/machined/main.go
```diff
@@ -157,31 +158,48 @@ func main() {
 			rebootFlag = unix.LINUX_REBOOT_CMD_POWER_OFF
 			fallthrough
 		case event.Reboot:
-			if err := seq.Shutdown(); err != nil {
-				panic(fmt.Errorf("shutdown failed: %w", err))
-			}
+			go func() {
+				if err := seq.Shutdown(); err != nil {
```
smira: this flow still allows for (slightly) concurrently sequences, as when `seq.Shutdown()` returns, another event might launch another sequence while this goroutine progresses through the `sync()` and other steps

while what I proposed below might be a more robust solution (if we push startup as event I guess)

andrewrynhard:
We have some designing discussions to work through for this, and the main goal of allowing only one sequence at a time made it in in another PR. Closing this.

PR #1772, opened 2020-01-13
fix: refuse to upgrade if single master

### LINKED ISSUE #1770: Check number of etcd members on upgrade
We should prevent single master setup from upgrading by checking the number of etcd members.

```
if len(members) == 1 {
    return fmt.Errorf("asdf")
}
```

rsmitty:
This PR adds some simple logic to bail early in the upgrade process if
there only seems to be a single etcd node present in the cluster. This
allows us to keep from blowing up non-HA clusters if users issue an
upgrade command.

Will close #1770.

### DIFF
--- a/internal/pkg/etcd/etcd.go
+++ b/internal/pkg/etcd/etcd.go
@@ -95,6 +95,10 @@ func ValidateForUpgrade() error {
 			return err
 		}
 
+		if len(resp.Members) == 1 {
+			return fmt.Errorf("only 1 etcd member found. assuming this is not an HA setup and refusing to upgrade")
+		}
+
 		for _, member := range resp.Members {
 			// If the member is not started, the name will be an empty string.
 			if len(member.Name) == 0 {


internal/app/machined/internal/phase/upgrade/leave_etcd.go
```diff
@@ -55,6 +55,10 @@ func (task *LeaveEtcd) standard(r runtime.Runtime) (err error) {
 		return err
 	}
 
+	if len(resp.Members) == 1 {
+		return fmt.Errorf("only 1 etcd member found. assuming this is not an HA setup and refusing to leave cluster")
+	}
+
```
andrewrynhard: We have a `ValidateForUpgrade` function in `internal/pkg/etcd` that is responsible for all the safety check for etcd. We should probably throw this in there.

rsmitty: ah, good call. fixed in amended commit.

PR #1775, opened 2020-01-14
fix: add Close func in remote generator

andrewrynhard:
Consumers of RemoteGenerator need a way to close the grpc.ClientConn.

### DIFF
--- a/pkg/grpc/gen/remote.go
+++ b/pkg/grpc/gen/remote.go
@@ -21,6 +21,8 @@ import (
 // RemoteGenerator represents the OS identity generator.
 type RemoteGenerator struct {
 	client securityapi.SecurityServiceClient
+	conn   *grpc.ClientConn
+	done   chan struct{}
 }
 
 // NewRemoteGenerator initializes a RemoteGenerator with a preconfigured grpc.ClientConn.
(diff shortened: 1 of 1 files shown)


smira:
or... should they just use the same connection? we could even add load-balancing to it, I have that in the back-burner for osctl, but approach should be same

pkg/grpc/gen/remote.go
```diff
@@ -81,6 +82,11 @@ func (g *RemoteGenerator) Identity(csr *x509.CertificateSigningRequest) (ca, crt
 	return ca, crt, nil
 }
 
+// Close closes the gRPC client connection.
+func (g *RemoteGenerator) Close() error {
+	return g.conn.Close()
```
smira: I don't see it being called anywhere, am I missing something?

andrewrynhard: This has less of a use case for Talos, but in the controller we do: https://github.com/talos-systems/talos-controller-manager/blob/master/pkg/upgrader/upgrader_v1alpha1.go#L105. In Talos a go routine is kicked of for the lifetime of the machine, that should never need to be closed, but in the controller we create a new certificate per upgrade. Now that I'm thinking about it, the better way would be to have the controller also use a go routine. Regardless, I think maybe it is still good to have a way to close the connection. This will be temporary to stop any bleeding, because I like the idea of using the same connection as you mentioned. I feel like that will take a little bit of design and warrants another PR? How do you feel about us getting this in and then working on a design for client-side load balancing that will use the same connection?

smira: got it, thanks about controller actually using it

yeah, certainly, a way to close a connection is actually required, no matter how we're using it. just learned that grpc connection is so complicated and handles failures, that sometimes it's better not to re-establish it if possible

PR #1776, opened 2020-01-14
fix: Ensure assets directory does not exist

bradbeam:
This fixes an issue where the bootkube assets directory was left over from a failed
installation.

### DIFF
--- a/internal/app/machined/pkg/system/services/bootkube.go
+++ b/internal/app/machined/pkg/system/services/bootkube.go
@@ -199,6 +199,14 @@ func generateAssets(config runtime.Configurator) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -199,6 +199,14 @@ func generateAssets(config runtime.Configurator) (err error) {
 		return err
 	}
 
+	// Ensure assets directory does not exist / is left over from a failed install
+	if err = os.RemoveAll(constants.AssetsDirectory); err != nil {
+		// Ignore if the directory does not exist
+		if _, ok := err.(*os.PathError); !ok {
```
andrewrynhard: nit: We should use `errors.Is`.

PR #1778, opened 2020-01-15
refactor: use tls.Config as client credentials

andrewrynhard:
The `client.Creds` struct was not used very often, and made using the
`client.NewClient` function impossible to use in combination with the
`RemoteRenewingFileCertificateProvider`. This modifies
`client.NewClient` to accept a `tls.Config` instead of `client.Creds`,
allowing for the use of `RemoteRenewingFileCertificateProvider` with
`client.NewClient`.

### DIFF
--- a/cmd/osctl/cmd/root.go
+++ b/cmd/osctl/cmd/root.go
@@ -16,6 +16,7 @@ import (
 	"github.com/talos-systems/talos/cmd/osctl/pkg/client"
 	"github.com/talos-systems/talos/cmd/osctl/pkg/helpers"
 	"github.com/talos-systems/talos/pkg/constants"
+	"github.com/talos-systems/talos/pkg/grpc/tls"
 	"github.com/talos-systems/talos/pkg/version"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -34,7 +34,7 @@ import (
 // Client represents a set of helper methods for interacting with the
 // Kubernetes API.
 type Client struct {
-	client *kubernetes.Clientset
```
bradbeam: Is this just random cleanup?

smira: looks like it's being used in controller-manager PR, and seems to be a good change, as I stumbled upon it a while ago: hard to use "regular" APIs if you wrap your k8s client with `Client`

smira:
And looks like lgtm has good points :)

pkg/kubernetes/kubernetes.go
andrewrynhard: This allows users of this package to get all the convenience functions we have added plus all the underlying Kubernetes functions.

cmd/osctl/cmd/root.go
```diff
@@ -110,7 +111,16 @@ func WithClient(action func(context.Context, *client.Client) error) error {
 		// Update context with grpc metadata for proxy/relay requests
 		ctx = client.WithNodes(ctx, targetNodes...)
 
-		c, err := client.NewClient(creds, configEndpoints, constants.ApidPort)
+		tlsconfig, err := tls.New(
+			tls.WithKeypair(creds.Crt),
+			tls.WithClientAuthType(tls.Mutual),
```
andrewrynhard: I'm not sure we need to actually set this for the client TLS config.

PR #1782, opened 2020-01-16
chore: validate installer image before upgrade

rsmitty:
This PR will add a very basic check to make sure that the installer
image specified during something like `osctl upgrade` indeed seems to be
a talos installer. We do this by simply calling the help flag of a known
binary (`/bin/installer --help`). If this doesn't exist, we exit with
the assumption that the user has made a mistake.


rsmitty:
This feels like a lot of extra code just to run the container and check the exit code. Kind of wondering if we already have that written somewhere that I can make use of?

PR #1788, opened 2020-01-17
fix: block when handling bus event

andrewrynhard:
If we don't block, there is the potential for multiple shutdown,
reboot, and upgrade requests to be processed.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -138,34 +138,23 @@ func main() {
 	// Boot the machine.
 	seq := sequencer.New(sequencer.V1Alpha1)
 
-	// Start the boot sequence in a go routine so that we can list for events.
-	go func() {
-		defer recovery()
-		if err := seq.Boot(); err != nil {
-			log.Println(err)
-			panic(fmt.Errorf("boot failed: %w", err))
-		}
-	}()
-
-	rebootFlag := unix.LINUX_REBOOT_CMD_RESTART
+	if err := seq.Boot(); err != nil {
+		log.Println(err)
+		panic(fmt.Errorf("boot failed: %w", err))
+	}
 
 	// Wait for an event.
 
+	flag := unix.LINUX_REBOOT_CMD_RESTART
+
+L:
 	for {
 		switch e := <-init.Channel(); e.Type {
 		case event.Shutdown:
-			rebootFlag = unix.LINUX_REBOOT_CMD_POWER_OFF
+			flag = unix.LINUX_REBOOT_CMD_POWER_OFF
 			fallthrough
 		case event.Reboot:
-			if err := seq.Shutdown(); err != nil {
-				panic(fmt.Errorf("shutdown failed: %w", err))
-			}
-
-			sync()
-
-			if unix.Reboot(rebootFlag) == nil {
-				select {}
-			}
+			break L
 		case event.Upgrade:
 			var (
 				req *machineapi.UpgradeRequest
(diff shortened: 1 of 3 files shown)


internal/app/machined/main.go
```diff
@@ -72,6 +72,37 @@ func sync() {
...
 
+func shutdown(seq sequencer.Sequencer, flag int) {
+	if err := seq.Shutdown(); err != nil {
+		panic(fmt.Errorf("shutdown failed: %w", err))
+	}
+
+	sync()
+
+	if unix.Reboot(flag) == nil {
+		select {}
```
andrewrynhard: This is the magic ... every event eventually lands here.

internal/pkg/event/event.go
```diff
@@ -52,17 +52,17 @@ type ObserveNotifier interface {
...
-	types   []Type
+	Chan  Channel
+	types []Type
 }
 
 // Channel implements the Observer interface.
 func (e *Embeddable) Channel() Channel {
-	if cap(e.channel) == 0 {
-		e.channel = make(Channel, 20)
+	if cap(e.Chan) == 0 {
```
smira: should it be `e.Chan == nil`?

internal/app/machined/main.go
```diff
@@ -114,7 +145,7 @@ func main() {
 	defer recovery()
 
 	// Subscribe to events.
-	init := EventBusObserver{&event.Embeddable{}}
+	init := EventBusObserver{&event.Embeddable{Chan: make(event.Channel, 1)}}
```
smira: I guess buffer of one still allows channel to "swallow" one event without feedback

internal/app/machined/main.go
smira: one potential way to make it easier to read the code is to put this code after the `for` loop in `main` and break out of the loop when we are ready to reboot/shutdown :) could have used return value of those upgrade/shutdown funcs to decide whether to halt the machine (just an idea)

internal/pkg/event/event.go
andrewrynhard: 🤔 not sure. I can check.

internal/app/machined/main.go
andrewrynhard: This isn't really needed I think. I can remove this since it will be addressed in the bus improvements we discussed.

internal/pkg/event/event.go
andrewrynhard: `cap` seems to work: https://play.golang.org/p/IHJz1I74p0j

internal/app/machined/main.go
andrewrynhard: We reboot regardless of success or not. I can look at `break`ing out of the for loop.

internal/pkg/event/event.go
```diff
@@ -52,17 +52,13 @@ type ObserveNotifier interface {
 
 // Embeddable is a type that implements sane defaults as an observer.
 type Embeddable struct {
-	channel Channel
-	types   []Type
+	Chan  Channel
```
andrewrynhard: This is so that if we ever want to make a non-blocking bus, we could.

internal/pkg/event/event.go
smira: what I meant is that unbuffered not-nil `channel` object will have `cap(e.Chan) == 0`, but this code will replace it with buffered channel. so `cap(nil) == 0` as well I guess, but this covers both. probably it's not important for this PR if we get a "bus" fix coming up.

internal/pkg/event/event.go
smira: now I'm confused, it looks to be uninitialized in `machined` main.go?

PR #1789, opened 2020-01-17
test: firecracker provisioner fixes, implement cluster destroy 

smira:
This implements `osctl cluster destroy` for Firecracker, adds
new utility command `osctl cluser show`.

Firecracker mode now has control process for firecracker VMs, allowing
clean reboots and background operations.

Lots of small fixes to Firecracker mode, clean CNI shutdown, cleaning up
netns, etc.

### DIFF
--- a/cmd/osctl/cmd/cluster.go
+++ b/cmd/osctl/cmd/cluster.go
@@ -10,8 +10,13 @@ import (
 	"fmt"
 	"math/big"
 	"net"
+	"os"
+	"path/filepath"
+	"sort"
+	"text/tabwriter"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
 
 	clientconfig "github.com/talos-systems/talos/cmd/osctl/pkg/client/config"
@@ -20,7 +25,6 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/provision/access"
 	"github.com/talos-systems/talos/internal/pkg/provision/check"
 	"github.com/talos-systems/talos/internal/pkg/provision/providers"
-	"github.com/talos-systems/talos/internal/pkg/provision/providers/docker"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/config"
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1/generate"
@@ -50,19 +54,20 @@ var (
 	cniBinPath              []string
 	cniConfDir              string
 	cniCacheDir             string
+	stateDir                string
 )
 
 // clusterCmd represents the cluster command
 var clusterCmd = &cobra.Command{
 	Use:   "cluster",
-	Short: "A collection of commands for managing local docker-based clusters",
+	Short: "A collection of commands for managing local docker-based or firecracker-based clusters",
 	Long:  ``,
 }
 
 // clusterUpCmd represents the cluster up command
 var clusterUpCmd = &cobra.Command{
 	Use:   "create",
-	Short: "Creates a local docker-based kubernetes cluster",
+	Short: "Creates a local docker-based or firecracker-based kubernetes cluster",
 	Long:  ``,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -73,11 +78,22 @@ var clusterUpCmd = &cobra.Command{
 // clusterDownCmd represents the cluster up command
 var clusterDownCmd = &cobra.Command{
 	Use:   "destroy",
-	Short: "Destroys a local docker-based kubernetes cluster",
+	Short: "Destroys a local docker-based or firecracker-based kubernetes cluster",
 	Long:  ``,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return helpers.WithCLIContext(context.Background(), destroy)
+	},
+}
+
+// clusterShowCmd represents the cluster show command
+var clusterShowCmd = &cobra.Command{
+	Use:   "show",
+	Short: "Shows info about a local provisioned kubernetes cluster",
+	Long:  ``,
+	Args:  cobra.NoArgs,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return helpers.WithCLIContext(context.Background(), show)
 	},
 }
 
@@ -134,7 +150,12 @@ func create(ctx context.Context) (err error) {
 	if inputDir != "" {
 		configBundleOpts = append(configBundleOpts, config.WithExistingConfigs(inputDir))
 	} else {
-		var genOptions []generate.GenOption
+		genOptions := []generate.GenOption{
+			generate.WithInstallImage(nodeInstallImage),
+		}
+
+		genOptions = append(genOptions, provisioner.GenOptions()...)
+
 		endpointList := []string{}
 
 		if forceEndpoint != "" {
@@ -189,6 +210,9 @@ func create(ctx context.Context) (err error) {
 		KernelPath:        nodeVmlinuxPath,
 		InitramfsPath:     nodeInitramfsPath,
 		KubernetesVersion: kubernetesVersion,
+
+		SelfExecutable: os.Args[0],
+		StateDirectory: stateDir,
 	}
 
 	// Create the master nodes.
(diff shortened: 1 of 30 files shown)


smira:
One thing left out is cleaning up bridge interface and CNI config file, I will defer to the next PR.

internal/pkg/provision/providers/firecracker/launch.go
```diff
@@ -0,0 +1,124 @@
...
+			waitCh <- m.Wait(ctx)
+		}()
+
+		select {
+		case err := <-waitCh:
+			if err != nil {
+				return fmt.Errorf("failed running VM: %w", err)
+			}
+
+			// restart vm by proceeding with the for loop
```
smira: need to add a short sleep here to prevent fast VM crash restart loop

smira:
How state is kept:

1. `docker` provisioner doesn't keep any state, it relies on Docker metadata to figure out cluster state, nodes, network, etc. (no changes in this PR)
2. `firecracker` provisioner simply dumps internal state `state.yaml` into clusters directory (by default `~/.talos/clusters/<name>/`) which contains list of nodes, cluster name, network info, etc. 

State information is used both in `osctl cluster show` and `osctl cluster destroy`.

smira:
Example:

```
$ time sudo ~/.local/bin/osctl cluster create --provisioner firecracker --name integration --cidr 172.20.0.0/24 --install-image docker.io/autonomy/installer:v0.4.0-alpha.1  --memory 1536 --masters 1 --workers 3 --init-node-as-endpoint 
generating PKI and tokens
creating state directory in "/home/smira/.talos/clusters/integration"
creating network integration
creating master nodes
creating worker nodes

real	0m2.870s
user	0m2.703s
sys	0m0.092s
```

```
$ osctl cluster show --name integration --provisioner firecracker
PROVISIONER       firecracker
NAME              integration
NETWORK NAME      integration
NETWORK CIDR      172.20.0.0/24
NETWORK GATEWAY   172.20.0.1
NETWORK MTU       1500

NODES:

NAME                   TYPE   IP           CPU    RAM      DISK
integration-master-1   Init   172.20.0.2   1.50   1.6 GB   4.3 GB
integration-worker-1   Join   172.20.0.3   1.50   1.6 GB   4.3 GB
integration-worker-2   Join   172.20.0.4   1.50   1.6 GB   4.3 GB
integration-worker-3   Join   172.20.0.5   1.50   1.6 GB   4.3 GB
```

```
$ sudo ~/.local/bin/osctl cluster destroy --name integration --provisioner firecracker
stopping VMs...
stopping VM integration-worker-2
stopping VM integration-worker-3
stopping VM integration-worker-1
stopping VM integration-master-1
removing state directory...
```

(using `sudo` here due to CNI requirements)

internal/pkg/provision/request.go
```diff
@@ -23,6 +23,12 @@ type ClusterRequest struct {
 	KernelPath        string
 	InitramfsPath     string
 	KubernetesVersion string
+
+	// Path to osctl executable to re-execute itself as needed.
+	SelfExecutable string
```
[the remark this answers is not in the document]
smira: If we want to take out `provisioner` out of `osctl`, then I guess we would need to have this field point to `osctl`. 

For example, if `integration-test` is spinning up Firecracker cluster for upgrade test (older version Talos cluster to upgrade to the current version).

So might be bad name here, but I guess we need it.

Another way cross-platform is `os.Args[0]`, but still prefer it to be explicit here.

cmd/osctl/pkg/client/config/path.go
```diff
@@ -0,0 +1,40 @@
...
+	if talosDir, ok := os.LookupEnv(constants.TalosDirectoryEnvVar); ok {
+		return talosDir, nil
+	}
+
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(home, ".talos"), nil
```
andrewrynhard: I'm torn on the `TALOSDIR` env var. I would prefer an explicit flag. Kinda like how `containerd` functions (it has a flag for specifying where to store state). Configurability is nice, but I think we should keep this particular option as simple and opinionated as possible. Knowing that `~/.talos` is a directory we "own" is easier to communicate and makes for less moving parts. Configuring this via a flag would be clear and explicit still.

docs/osctl/osctl_cluster_show.md
```diff
@@ -0,0 +1,35 @@
...
+### Options
+
+```
+  -h, --help   help for show
+```
+
+### Options inherited from parent commands
+
+```
+      --clusters-dir string   directory path to store cluster state (default "/home/user/.talos/clusters")
```
andrewrynhard: nit: `--state` (like `containerd`)?

internal/integration/integration_test.go
```diff
@@ -65,17 +63,7 @@ func TestIntegration(t *testing.T) {
...
-		ok                 bool
-	)
-
-	if defaultTalosConfig, ok = os.LookupEnv(constants.TalosConfigEnvVar); !ok {
-		home, err := os.UserHomeDir()
-		if err == nil {
-			defaultTalosConfig = path.Join(home, ".talos", "config")
-		}
-	}
+	defaultTalosConfig, _ := config.GetDefaultPath() //nolint: errcheck
```
andrewrynhard: I know we have no other way other than a `panic` here, but it would be nice if we didn't have to check for an error. If we didn't have the `TALOSDIR` env var, we would have to do this I think.

internal/pkg/provision/providers/docker/reflect.go
```diff
@@ -7,12 +7,13 @@ package docker
...
 	"context"
 	"net"
+	"strconv"
 
 	"github.com/talos-systems/talos/internal/pkg/provision"
 	"github.com/talos-systems/talos/pkg/config/machine"
 )
 
-func (p *provisioner) Reflect(ctx context.Context, clusterName string) (provision.Cluster, error) {
+func (p *provisioner) Reflect(ctx context.Context, clusterName, clustersDirectory string) (provision.Cluster, error) {
```
andrewrynhard: I think we should refer to `clustersDirectory` as `stateDirectory`. It is a little bit more clear IMO (aligns with `--state` flag as well).

internal/pkg/provision/providers/docker/result.go
```diff
@@ -5,13 +5,23 @@
...
 
+func (res *result) Provisioner() string {
+	return "docker"
+}
+
 func (res *result) Info() provision.ClusterInfo {
 	return res.clusterInfo
 }
+
+func (res *result) StateDirectoryPath() (string, error) {
```
andrewrynhard: Just pointing this out to support the argument for changing `--cluster-dir` to `--state`. Aligns with the interface, which is nice in this case.

internal/pkg/provision/providers/firecracker/create.go
```diff
@@ -28,80 +27,77 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
...
-
-		cfg, err = node.Config.String()
-		if err != nil {
-			return nil, err
-		}
-
-		if err = httpServer.AddFile(fmt.Sprintf("%s.yaml", node.Name), []byte(cfg)); err != nil {
-			return nil, err
-		}
+	err = os.MkdirAll(state.StateDirectory, os.ModePerm)
```
andrewrynhard: style (feel free to ignore): `if err = ...; err != nil`

internal/pkg/provision/providers/firecracker/create.go
```diff
@@ -28,80 +27,77 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
...
-	return state, nil
+	// save state
+	stateFile, err := os.Create(filepath.Join(state.StateDirectory, stateFileName))
+	if err != nil {
+		return nil, err
+	}
+
+	defer stateFile.Close() //nolint: errcheck
+
+	if err = yaml.NewEncoder(stateFile).Encode(&state); err != nil {
```
andrewrynhard: What is the benefit of `Encode` vs `Unmarshal`?

internal/pkg/provision/providers/firecracker/launch.go
```diff
@@ -0,0 +1,133 @@
...
+// Launch restarts VM forever until control process is stopped itself with a signal.
+//
+// Process is expected to receive configuration on stdin. Current working directory
+// should be cluster state directory, process output should be redirected to the
+// logfile in state directory.
+//
+// When signals SIGINT, SIGTERM are received, control process stops firecracker and exits.
+//
+//nolint: gocyclo
+func Launch() error {
```
andrewrynhard: 👍

internal/pkg/provision/providers/firecracker/node.go
```diff
@@ -149,50 +155,137 @@ func (p *provisioner) createNode(ctx context.Context, state *state, clusterReq p
...
-			}
-		}
+	defer launchConfigFile.Close() //nolint: errcheck
+
+	cmd := exec.Command(clusterReq.SelfExecutable, "firecracker-launch")
+	cmd.Stdout = logFile
+	cmd.Stderr = logFile
+	cmd.Stdin = launchConfigFile
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		Setsid: true, // daemonize
```
andrewrynhard: 👍

cmd/osctl/pkg/client/config/path.go
smira: Explicit flag is there for `~/.talos/clusters`, but this provides default for `~/.talos` itself (and influences config location). I can hardcode this to `~/.talos` though.

internal/integration/integration_test.go
smira: this is just building a default value, if we can't build it for whatever reason, it's not fatal, integration-test can still proceed if flag is provided explicitly, so I think panic is overkill here?

internal/pkg/provision/providers/docker/result.go
smira: this is a bit different, this is path to `~/.talos/clusters/<name>` (state for specific cluster), but I think overall moving to `StateDirectory` instead of `ClustersDirectory` sounds good to me

internal/pkg/provision/providers/firecracker/create.go
smira: functionally equivalent, but `Encode` works on `io.Writer` directly without buffering in memory. no big deal for a small structure here, but a big difference for huge objects.

cmd/osctl/cmd/cluster.go
```diff
@@ -134,7 +150,12 @@ func create(ctx context.Context) (err error) {
 	if inputDir != "" {
 		configBundleOpts = append(configBundleOpts, config.WithExistingConfigs(inputDir))
 	} else {
-		var genOptions []generate.GenOption
+		genOptions := []generate.GenOption{
```
smira: i was recovering here from mis-merge

cmd/osctl/cmd/config.go
```diff
@@ -203,12 +204,14 @@ func genV1Alpha1Config(args []string) error {
...
-				Endpoint:                  args[1],
-				KubeVersion:               kubernetesVersion,
-				AdditionalSubjectAltNames: additionalSANs,
-				InstallDisk:               installDisk,
-				InstallImage:              installImage,
+				ClusterName: args[0],
+				Endpoint:    args[1],
+				KubeVersion: kubernetesVersion,
+				GenOptions: []generate.GenOption{
+					generate.WithInstallDisk(installDisk),
```
smira: using existing `GenOption`s, and added one for SANs

pkg/constants/constants.go
```diff
@@ -203,6 +203,9 @@ const (
 	// CRIContainerdConfig is the path to the config for the containerd instance that provides the CRI.
 	CRIContainerdConfig = "/etc/cri/containerd.toml"
 
+	// TalosDirectoryEnvVar is the environment variable for setting the Talos directory path.
+	TalosDirectoryEnvVar = "TALOSDIR"
```
andrewrynhard: I think we can remove this.

smira: oh yeah, my bad, will remove in the next PR, thanks!

PR #1796, opened 2020-01-21
fix: leave etcd after draining node

andrewrynhard:
This fixes an edge case where etcd can potentially become unresponsive,
and prevent a draining of a node. We drain the node, and then leave etcd
since all subsequent tasks do not depend on etcd in any way, and will
reduce the chances of an upgrade failure.


internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -211,6 +211,9 @@ func (d *Sequencer) Upgrade(req *machineapi.UpgradeRequest) error {
 		phase.NewPhase(
 			"cordon and drain node",
 			kubernetes.NewCordonAndDrainTask(),
+		),
+		phase.NewPhase(
+			"etcd",
```
andrewrynhard: needs a better name

PR #1800, opened 2020-01-21
fix: install sequence stuck on event bus

smira:
machined's main.go waits for boot sequence to finish, while metal
platform initializer tries to send a message to the event bus without
any listeners, so this is pure deadlock.

Resolve that by panicking from initializer, this aborts phase and
sequence, and leads to reboot on panic. Not really clean as it leaves
scary stacktraces in the dmesg, but it works. Cleanup might be done by
introducing error value for reboot, and ignoring it when printing the
errors.

### DIFF
--- a/internal/app/machined/internal/phase/phase.go
+++ b/internal/app/machined/internal/phase/phase.go
@@ -5,6 +5,7 @@
 package phase
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	goruntime "runtime"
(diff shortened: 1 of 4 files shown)


internal/app/machined/internal/phase/phase.go
```diff
@@ -128,7 +128,7 @@ func (r *Runner) runTask(task Task, errCh chan<- error) {
 
 	defer func() {
 		if r := recover(); r != nil {
-			buf := make([]byte, 8192)
+			buf := make([]byte, 512) // using small buffer here, as kmsg has its limits
```
smira: kmsg has limit slightly less than 1 KB: https://github.com/torvalds/linux/blob/master/kernel/printk/printk.c#L450

smira: this fix is in fact good on its own: it was preventing panic messages to be printed in any phase is traceback was long enough

internal/app/machined/internal/phase/phase.go
```diff
@@ -103,19 +104,30 @@ func (r *Runner) runPhase(phase *Phase) error {
...
+			}
 		}
 
 		result = multierror.Append(result, err)
 	}
 
 	log.Printf("[phase]: %s done, %s", phase.description, time.Since(start))
 
+	if reboot {
+		return runtime.ErrReboot
```
smira: this is required because what is returned by `multierror` doesn't support `errors.Is` protocol, so we can't check if error is `runtime.ErrReboot` later on in `main.go`

PR #1803, opened 2020-01-22
chore: use firecracker in basic-integration

andrewrynhard:


hack/drone.jsonnet
```diff
@@ -142,6 +157,7 @@ local Step(name, image='', target='', depends_on=[], environment={}) = {
   image: if image == '' then build_container else image,
   pull: "always",
   commands: [make],
+  privileged: true,
```
smira: :heart: I missed that it should enabled explicitly

hack/test/basic-integration.sh
```diff
@@ -32,9 +32,41 @@ esac
...
+      --install-image docker.io/autonomy/installer:latest
+    ;;
+
+  *)
+    echo "unknown provisioner: ${PROVISIONER}"
+    exit 1
+    ;;
+esac
 
-"${INTEGRATION_TEST}" -test.v -talos.osctlpath "${OSCTL}" -talos.k8sendpoint "${ENDPOINT}:6443"
```
smira: iirc, `talos.k8sendpoint` is still required for Docker, because internal endpoint (10.5.0.2) is different from external (`docker` in our case)

smira: for Firecracker we don't need, as init node is routable both externally and internally

smira: but if we define `ENDPOINT=172.20.0.2` for Firecracker, it will be no-op

andrewrynhard: ahh thanks, was banging my head against the wall

Issue #1721, opened 2020-01-05
Look into gRPC  health check rpc


bradbeam:
grpc gives an example/rough outline of what a healthcheck endpoint should look like. this seems like it'd be beneficial to enhance our init/service health check api.
https://github.com/grpc/grpc/blob/master/doc/health-checking.md

PR #1807, opened 2020-01-22
feat(networkd): Add health api

bradbeam:
This introduces a health/ready api for networkd. This
will allow us to better determine the state of networkd
and allow for some level of monitoring.

implementation of #1721 


internal/app/machined/pkg/system/services/networkd.go
```diff
@@ -120,11 +123,27 @@ func (n *Networkd) Runner(config runtime.Configurator) (runner.Runner, error) {
...
+			hcResp *healthapi.HealthCheckResponse
+		)
+
+		if conn, err = grpc.Dial("unix:"+constants.NetworkSocketPath, grpc.WithInsecure()); err != nil {
+			return err
+		}
+		defer conn.Close() //nolint: errcheck
+
+		nClient := healthapi.NewHealthClient(conn)
+		if hcResp, err = nClient.Check(context.Background(), &empty.Empty{}); err != nil {
```
smira: please use `ctx` here, as otherwise health check might take more time than health check interval (`ctx` has proper deadline set)

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -149,3 +151,49 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+}
+
+// Ready implements the Health api and provides visibility to the state of networkd.
+// Ready signifies the initial network configuration ( interfaces, routes, hostname, resolv.conf )
+// settings have been applied.
+// Not Ready signifies that the initial network configuration still needs to happen.
+func (r *Registrator) Ready(ctx context.Context, in *empty.Empty) (reply *healthapi.ReadyCheckResponse, err error) {
+	rdy := &healthapi.ReadyCheck{Status: healthapi.ReadyCheck_NOT_READY}
+
+	if r.Networkd.Ready {
```
smira: this is racy with variable write, so need mutex/atomic to synchronize

bradbeam:
I'm curious about how we can better integrate the notion of a service being ready into our dependency chain. 
As an example, for services that depend on networkd we return healthy pretty quick ( api is up and available ), but we should really wait for networkd to be ready ( as in all interfaces and such configured ).

smira:
> I'm curious about how we can better integrate the notion of a service being ready into our dependency chain.
> As an example, for services that depend on networkd we return healthy pretty quick ( api is up and available ), but we should really wait for networkd to be ready ( as in all interfaces and such configured ).

Quick & dirty is to make service `networkd` depend on Ready health API. Any service that depends on `networkd` won't start until `networkd` is healthy (Ready in our definition).

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -41,6 +41,9 @@ type Networkd struct {
 
 	hostname  string
 	resolvers []string
+
+	readyLock sync.Mutex
```
andrewrynhard: We can also embed `sync.Mutex`, then we can just do `n.Lock()` and `n.Unlock`. That is a common convention I have along with calling this "`mu`", similar to the convention of using `wg` for `sync.WaitGroup`.

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -389,3 +394,18 @@ func (n *Networkd) decideHostname() (hostname string, domainname string, address
...
+	defer n.readyLock.Unlock()
+
+	return n.ready
+}
+
+// SetReady sets the readiness state of networkd.
+func (n *Networkd) SetReady() {
+	n.readyLock.Lock()
+	n.ready = true
+	n.readyLock.Unlock()
```
andrewrynhard: nit: The `Lock()` and `defer Unlock()` strategy is used in `Ready`, but not here. That is preferred, but either way the strategy/method (whatever you want to cal it) should align/match.

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -149,3 +151,49 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+// Ready signifies the daemon (api) is healthy and ready to serve requests.
+func (r *Registrator) Watch(in *empty.Empty, srv healthapi.Health_WatchServer) (err error) {
+	return srv.Send(&healthapi.HealthCheckResponse{
+		Messages: []*healthapi.HealthCheck{
+			{
+				Status: healthapi.HealthCheck_SERVING,
+			},
+		},
+	})
+}
```
andrewrynhard: I like the idea of watching this. Not sure if its just due to the way networkd works or its just not implemented, but I'm curious about the vision for the real world use case for watching.

bradbeam: i don't know what the actual health check for networkd looks like (yet?).

internal/app/networkd/pkg/reg/reg_test.go
```diff
@@ -28,13 +30,13 @@ type NetworkdSuite struct {
 
 func TestNetworkdSuite(t *testing.T) {
 	// Hide all our state transition messages
-	// log.SetOutput(ioutil.Discard)
+	log.SetOutput(ioutil.Discard)
```
smira: this is going to affect all the tests running after this one, so I think better way should be to restore previous output device when test finishes; tests for different packages are launched as separate binaries, but I guess still better hygiene

internal/app/machined/pkg/system/services/networkd.go
```diff
@@ -120,11 +123,27 @@ func (n *Networkd) Runner(config runtime.Configurator) (runner.Runner, error) {
...
+			hcResp *healthapi.HealthCheckResponse
+		)
+
+		if conn, err = grpc.Dial("unix:"+constants.NetworkSocketPath, grpc.WithInsecure()); err != nil {
+			return err
+		}
+		defer conn.Close() //nolint: errcheck
+
+		nClient := healthapi.NewHealthClient(conn)
+		if hcResp, err = nClient.Check(ctx, &empty.Empty{}); err != nil {
```
smira: should we depend on `Ready()` here if we want to wait for networkd to complete all the initial config?

internal/app/networkd/pkg/reg/reg_test.go
andrewrynhard: +1

internal/app/networkd/pkg/reg/reg.go
bradbeam: I think the simplest and most effective health check for now would be a simple ping to the default gateway. There's a reference implementation of a pinger here https://godoc.org/golang.org/x/net/icmp#example-PacketConn--NonPrivilegedPing

```
Check() => ping default gateway
Watch(int) => ping default gateway with frequency in seconds
```

internal/app/machined/pkg/system/services/networkd.go
bradbeam: I think we will want to do a combination of both.

If !networkd.Ready() then there's no use in trying a health check.

If networkd.Ready() && networkd.Check() then healthy.

internal/app/networkd/pkg/reg/reg.go
andrewrynhard: Yeah that sounds reasonable. Would be a good indicator that things are working as expected.

bradbeam:
I'll follow up with a separate pr to add in the pinger stuff.

internal/app/machined/pkg/system/services/networkd.go
andrewrynhard: Is this in the follow up PR?

api/health/health.proto
```diff
@@ -0,0 +1,44 @@
...
+
+service Health {
+  rpc Check(google.protobuf.Empty) returns (HealthCheckResponse);
+  rpc Watch(HealthWatchRequest)  returns (stream HealthCheckResponse);
+  rpc Ready(google.protobuf.Empty) returns (ReadyCheckResponse);
+}
+
+message HealthWatchRequest {
+	int64 count = 1;
+  int64 interval_seconds = 2;
```
andrewrynhard: nit: these are misaligned

andrewrynhard: I don't think we should have a count, but rather the client should decide when its done and should watch forever.

internal/app/machined/pkg/system/services/networkd.go
bradbeam: negative, this is here.
L139-145 - ready check
L147-153 - health check

the health check is 'dumb' right now in that it always returns that it's up/available. This should maintain current functionality.

api/health/health.proto
bradbeam: by default it works like that. a 0 value indicates forever, >0 indicates a limit to the retries. 

ping vs ping -c4 for example.

andrewrynhard: ahh cool.

api/health/health.proto
```diff
@@ -0,0 +1,43 @@
...
+import "google/protobuf/empty.proto";
+
+service Health {
+  rpc Check(google.protobuf.Empty) returns (HealthCheckResponse);
+  rpc Watch(HealthWatchRequest)  returns (stream HealthCheckResponse);
+  rpc Ready(google.protobuf.Empty) returns (ReadyCheckResponse);
+}
+
+message HealthWatchRequest {
+  int64 interval_seconds = 1;
```
andrewrynhard: No more count?

bradbeam: not necessary as you pointed out, this will work. if we need add count later it'd be an additive change which should be easy enough to implement after the fact.

andrewrynhard: Might even be something we leave up to client side implementations.

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -149,3 +153,67 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+	if in == nil {
+		return errors.New("an input interval is required")
+	}
+
+	var (
+		resp   *healthapi.HealthCheckResponse
+		ticker = time.NewTicker(time.Duration(in.IntervalSeconds) * time.Second)
+	)
+
+	for range ticker.C {
```
smira: this loop never finishes, not sure why lines below are commented out as they seem to make perfect sense to me

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -149,3 +153,67 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+		// select {
+		// case <-srv.Context().Done():
+		//	return nil
```
smira: I guess it should return `srv.Context().Error()`

internal/app/networkd/pkg/reg/reg.go
andrewrynhard: Yeah I missed this :(

PR #1811, opened 2020-01-23
fix: retry system disk busy check

andrewrynhard:
This adds a retry to checking if the system disk is in use. An extra
unix.Sync() was thrown in as an extra measure to maybe help unblock any
IO threads.

### DIFF
--- a/internal/app/machined/internal/phase/disk/verify_availability.go
+++ b/internal/app/machined/internal/phase/disk/verify_availability.go
@@ -6,11 +6,13 @@ package disk
 
 import (
 	"errors"
+	"time"
 
 	"golang.org/x/sys/unix"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
+	"github.com/talos-systems/talos/pkg/retry"
 )
 
 // VerifyDiskAvailability represents the task for verifying that the system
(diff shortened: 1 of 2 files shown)


internal/app/machined/internal/phase/disk/verify_availability.go
```diff
@@ -35,8 +37,18 @@ func (task *VerifyDiskAvailability) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 }
 
 func (task *VerifyDiskAvailability) standard() (err error) {
-	if isBusy(task.devname) {
-		return errors.New("system disk in use")
+	unix.Sync()
+
+	err = retry.Constant(1*time.Minute, retry.WithUnits(500*time.Millisecond), retry.WithJitter(50*time.Millisecond)).Retry(func() error {
```
smira: looks like you can `return retry.Constant(...` here

jsimonetti:
I would suggest to add an additional log entry displaying what is keeping the fs busy.

`lsof +f -- /` should show you more info

andrewrynhard:
@jsimonetti unfortunately we don't have `lsof` the binary, but we could certainly follow up with the go version of that. There is a little bit of a deep hole here though. We chatted about it in the meeting today, but the gist is that we can get al the way to stopping every k8s container, and unmounting the disk, but even though the disk is unmounted, the kerne still thinks it is busy.

PR #1814, opened 2020-01-23
docs(apid): Add apid docs

bradbeam:
Describes apid and introduces some workflows to illustrate what apid does.

### DIFF
--- a/.conform.yaml
+++ b/.conform.yaml
@@ -17,6 +17,7 @@ policies:
           - test
         scopes:
           - ami
+          - apid
           - ci
           - conformance
           - gpt
--- a/docs/website/content/v0.3.en.json
+++ b/docs/website/content/v0.3.en.json
@@ -88,6 +88,10 @@
     "path": "v0.3/en/components",
     "items": [
       {
+        "title": "apid",
+        "path": "v0.3/en/components/apid"
+      },
+      {
         "title": "kernel",
         "path": "v0.3/en/components/kernel"
       },
--- a/docs/website/content/v0.3/en/components/apid.md
+++ b/docs/website/content/v0.3/en/components/apid.md
@@ -1,0 +1,49 @@
+---
+title: 'apid'
+---
+
+When interacting with Talos, the gRPC api endpoint you will interact with directly is `apid`.
+Apid acts as the gateway for all component interactions.
+Apid provides a mechanism to route requests to the appropriate destination when running on a control plane node.
+
+We'll use some examples below to illustrate what `apid` is doing.
+
+When a user wants to interact with a Talos component via `osctl`, there are two flags that control the interaction with `apid`.
+The `-e | --endpoints` flag is used to denote which Talos node ( via `apid` ) should handle the connection.
+Typically this is a public facing server.
+The `-n | --nodes` flag is used to denote which Talos node(s) should respond to the request.
+If `--nodes` is not specified, the first endpoint will be used.
+> Note: Typically there will be an `endpoint` already defined in the Talos config file.
+Optionally, `nodes` can be included here as well.
+
+For example, if a user wants to interact with `osd`, a command like `osctl -e cluster.talos.dev memory` may be used.
+
+```bash
+$ osctl -e cluster.talos.dev memory
+NODE                TOTAL   USED   FREE   SHARED   BUFFERS   CACHE   AVAILABLE
+cluster.talos.dev   7938    1768   2390   145      53        3724    6571
+```
+
+In this case, `osctl` is interacting with `apid` running on `cluster.talos.dev` and forwarding the request to the `osd` api.
+
+If we wanted to extend our example to retrieve `memory` from another node in our cluster, we could use the command `osctl -e cluster.talos.dev -n node02 memory`.
+
+```bash
+$ osctl -e cluster.talos.dev -n node02 memory
+NODE    TOTAL   USED   FREE   SHARED   BUFFERS   CACHE   AVAILABLE
+node02  7938    1768   2390   145      53        3724    6571
+```
+
+The `apid` instance on `cluster.talos.dev` receives the request and forwards it to `apid` running on `node02` which forwards the request to the `osd` api.
+
+We can further extend our example to retrieve `memory` for all nodes in our cluster by appending additional `-n node` flags or using a comma separated list of nodes ( `-n node01,node02,node03` ):
+
+```bash
+$ osctl -e cluster.talos.dev -n node01 -n node02 -n node03 memory
+NODE     TOTAL    USED    FREE     SHARED   BUFFERS   CACHE   AVAILABLE
+node01   7938     871     4071     137      49        2945    7042
+node02   257844   14408   190796   18138    49        52589   227492
+node03   257844   1830    255186   125      49        777     254556
+```
+
+The `apid` instance on `cluster.talos.dev` receives the request and forwards is to `node01`, `node02`, and `node03` which then forwards the request to their local `osd` api.


docs/website/content/v0.3.en.json
```diff
@@ -99,6 +99,10 @@
         "title": "machined",
         "path": "v0.3/en/components/machined"
       },
+      {
+        "title": "apid",
+        "path": "v0.3/en/components/apid"
+      },
```
andrewrynhard: I think what we went for in a previous PR is to keep these in alphabetical order.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
+---
+title: 'apid'
+---
+
+When interacting with Talos, the gRPC api endpoint you will interact with directly is APId.
+APId acts as the gateway for all component interactions.
+APId provides a mechanism to route requests to the appropriate destination.
```
andrewrynhard: In this paragraph the service is referred to as "APId", and in the next it is "apid". Not sure I care too much, but these should probably align. Adding that a bit, in other component docs we use the reformatted lowercase version (e.g.`apid`) of the name. To give a consistent feel, we should do that in this doc too.

.conform.yaml
```diff
@@ -39,6 +40,7 @@ policies:
       skipPaths:
         - .git/
         - .buildkit/
+        - vendor/
```
andrewrynhard: We don't vendor anything so we can remove this.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
+---
+title: 'apid'
+---
+
+When interacting with Talos, the gRPC api endpoint you will interact with directly is apid.
+Apid acts as the gateway for all component interactions.
+Apid provides a mechanism to route requests to the appropriate destination.
```
andrewrynhard: In other component docs we use the pre-formatted lowercase version (e.g. `apid`) of the name. To give a consistent feel, we should do that in this doc as well (Except for sentences that start with the service name, "Apid" , like you have it is fine).

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
...
+
+We'll use some examples below to illustrate what apid is doing.
+
+When a user wants to interact with a Talos component via `osctl`, there are two flags that control the interaction with apid.
+The `-e | --endpoints` flag is used to denote which Talos node ( via apid ) should handle the connection.
+Typically this is a public facing server.
+*Note* Typically the endpoints are configured in the Talos config file
```
andrewrynhard: Notes should use `> Note: ...`. We have some CSS for `>` specifically. In docs we also put notes after the paragraph. We might also want to do something like:

```
> Note: The endpoints, and nodes can be specified in the Talos config file (e.g. `osctl config endpoints asdf`).
```

Or maybe each command is worthy of an example?

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
...
+
+For example, if a user wants to interact with `osd`, a command like `osctl -e cluster.talos.dev memory` may be used.
+
+```bash
+$ osctl -e cluster.talos.dev memory
+NODE                TOTAL   USED   FREE   SHARED   BUFFERS   CACHE   AVAILABLE
+cluster.talos.dev   7938    1768   2390   145      53        3724    6571
+```
+
+In this case, `osctl` is interacting with `apid` running on `cluster.talos.dev` and forwarding the request to the `osd` api.
```
andrewrynhard: Another thing worth pointing out is that only instances of `apid` running control plane nodes can be used as an endpoint.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
...
+
+```bash
+$ osctl -e cluster.talos.dev -n node01 -n node02 -n node03 memory
+NODE     TOTAL    USED    FREE     SHARED   BUFFERS   CACHE   AVAILABLE
+node01   7938     871     4071     137      49        2945    7042
+node02   257844   14408   190796   18138    49        52589   227492
+node03   257844   1830    255186   125      49        777     254556
+```
+
+The `apid` instance on `cluster.talos.dev` receives the request and forwards is to `node01`, `node02`, and `node03` which them forward the request to their local `osd` api.
```
andrewrynhard: `s/which them/which then`

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
...
+NODE    TOTAL   USED   FREE   SHARED   BUFFERS   CACHE   AVAILABLE
+node02  7938    1768   2390   145      53        3724    6571
+```
+
+The `apid` instance on `cluster.talos.dev` receives the request and forwards it to `apid` running on `node02` which forwards the request to the `osd` api.
+
+We can further extend out example to retrieve `memory` for all nodes in our cluster by appending additional `-n node` flags:
+
+```bash
+$ osctl -e cluster.talos.dev -n node01 -n node02 -n node03 memory
```
andrewrynhard: Not sure what a better first impression is:

```
osctl -e cluster.talos.dev -n node01 -n node02 -n node03 memory
```
or
```
osctl -e cluster.talos.dev -n node01,node02,node03 memory 
```

andrewrynhard: Maybe the latter since I can already hear people complaining about having to type too many characters 😄.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,47 @@
...
+NODE     TOTAL    USED    FREE     SHARED   BUFFERS   CACHE   AVAILABLE
+node01   7938     871     4071     137      49        2945    7042
+node02   257844   14408   190796   18138    49        52589   227492
+node03   257844   1830    255186   125      49        777     254556
+```
```
andrewrynhard: It would be cool if we could expand on this example a bit more and show how someone could do the same request, but without having to pass `-n`. Basically a working example of `osctl config nodes ...`.

docs/website/content/v0.3/en/components/apid.md
bradbeam: That's not true. Any apid can be used as an endpoint.

docs/website/content/v0.3/en/components/apid.md
bradbeam: I think that's covered by the note above. This is really just trying to be super verbose to explain what is going on.

docs/website/content/v0.3/en/components/apid.md
andrewrynhard: Then we have a bug 🙂. IIRC that was the original design, but we decided to only have the instances of apid that are in masters do any kind of proxying.

bradbeam: I'm not really sure what benefit we'd gain from restricting this, but either way it's not relevant to this pr. We can update this when(if?) it gets fixed.

andrewrynhard: I'd rather not document what is a bug as a feature. Documenting the proper behavior is what we should do, and follow up with a fix. We get a fair amount of traffic to our docs every day, and I'd rather paint a picture that is accurate to the design in this case.

andrewrynhard: I'm gonna put something on the agenda for tomorrow to walk through whether or not we want to proxy only on master nodes.

smira: I guess it might be important if that proxying depends more on PKI infra, and we do machine-level certificates for cross-apid communication. Then apid might refuse proxied connections from non-masters.

I agree that technically today we can proxy either way, but at the same time this seems to be overkill in general.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,48 @@
...
+When interacting with Talos, the gRPC api endpoint you will interact with directly is `apid`.
+Apid acts as the gateway for all component interactions.
+Apid provides a mechanism to route requests to the appropriate destination.
+
+We'll use some examples below to illustrate what `apid` is doing.
+
+When a user wants to interact with a Talos component via `osctl`, there are two flags that control the interaction with `apid`.
+The `-e | --endpoints` flag is used to denote which Talos node ( via `apid` ) should handle the connection.
+Typically this is a public facing server.
+The `-n | --nodes` flag is used to denote which Talos node(s) should respond to the request.
```
smira: I think we should say that if `--nodes` is not set, `osctl` talks directly to the endpoint without proxying. So default might be `--nodes = --endpoints[0]`?

docs/website/content/v0.3/en/components/apid.md
andrewrynhard: Exactly this ^. I think running `apid` in proxy mode has security implications, and we should protect it when we get to RBAC.

docs/website/content/v0.3/en/components/apid.md
```diff
@@ -0,0 +1,48 @@
...
+
+```bash
+$ osctl -e cluster.talos.dev -n node02 memory
+NODE    TOTAL   USED   FREE   SHARED   BUFFERS   CACHE   AVAILABLE
+node02  7938    1768   2390   145      53        3724    6571
+```
+
+The `apid` instance on `cluster.talos.dev` receives the request and forwards it to `apid` running on `node02` which forwards the request to the `osd` api.
+
+We can further extend out example to retrieve `memory` for all nodes in our cluster by appending additional `-n node` flags or using a comma separated list of nodes ( `-n node01,node02,node03` ):
```
andrewrynhard: `s/We can further extend out example/We can further extend our example`

PR #1819, opened 2020-01-24
fix: correctly split lines with /dev/kmsg output

smira:
This fixes cases when large `log.Printf()` was simply lost as it exceeds
max line length (despite being multi-line string with each line below
1KB). Now console logging should be much more reliable.

### DIFF
--- a/internal/pkg/install/install.go
+++ b/internal/pkg/install/install.go
@@ -7,6 +7,7 @@ package install
 import (
 	"context"
 	"fmt"
+	"log"
 
 	"github.com/containerd/containerd"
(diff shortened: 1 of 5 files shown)


internal/pkg/install/install.go
```diff
@@ -38,6 +39,8 @@ func RunInstallerContainer(r runtime.Runtime, opts ...Option) error {
 		return err
 	}
 
+	log.Printf("pulling installer container image: %q", r.Config().Machine().Install().Image())
```
smira: I added this as when image pull hangs on retries, it's hard to understand why boot hangs

PR #1827, opened 2020-01-27
chore: rework firecracker code around upstream Go SDK + PRs

smira:
This removes use of private fork with custom `ip=` kernel argument
handling and switches fully to upstream version of it.

Firecracker Go SDK version is `master` + following PRs:

* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/167
* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/177
* https://github.com/firecracker-microvm/firecracker-go-sdk/pull/178

MTU handling support was implemented as well.

Changes:

* hostname to each node is passed via `talos.hostname=` kernel arg
* IP configuration is generated by SDK from CNI result
* fixed bugs with wrong netmask
* nameservers & MTU is passed via Talos config
* network CNI config is now not written to disk (no need to clean it up)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.13
 
 replace (
 	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-	github.com/firecracker-microvm/firecracker-go-sdk v0.19.0 => github.com/smira/firecracker-go-sdk v0.19.1-0.20200110185541-4fce8cba9f84
+	github.com/firecracker-microvm/firecracker-go-sdk v0.19.0 => github.com/smira/firecracker-go-sdk v0.19.1-0.20200124123725-fb8b25794e34
 	github.com/kubernetes-sigs/bootkube => github.com/talos-systems/bootkube v0.14.1-0.20200123150754-82cbbbe2c4de
 	github.com/opencontainers/runtime-spec v1.0.1 => github.com/opencontainers/runtime-spec v0.1.2-0.20180301181910-fa4b36aa9c99
 )
--- a/internal/pkg/provision/providers/firecracker/network.go
+++ b/internal/pkg/provision/providers/firecracker/network.go
@@ -11,8 +11,6 @@ import (
 	"encoding/hex"
 	"fmt"
 	"net"
-	"os"
-	"path/filepath"
 	"strconv"
 	"text/template"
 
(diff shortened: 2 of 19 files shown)


pkg/config/types/v1alpha1/generate/options.go
```diff
@@ -43,12 +45,22 @@ func WithInstallImage(imageRef string) GenOption {
 	}
 }
 
+// WithNetworkConfig allows to pass network config to be used.
+func WithNetworkConfig(network *v1alpha1.NetworkConfig) GenOption {
```
smira: I don't like passing whole `NetworkConfig`, but I don't know better way? pass individual fields? `machine.Device`?

internal/pkg/provision/providers/firecracker/network.go
```diff
@@ -140,23 +137,23 @@ const networkTemplate = `
...
 	"plugins": [
-	  {
-		"type": "bridge",
-		"bridge": "{{ .InterfaceName }}",
-		"ipMasq": true,
-		"isGateway": true,
-		"isDefaultGateway": true,
-		"ipam": {
-		  "type": "static"
+		{
```
smira: just formatted properly without changes

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.13
 
 replace (
 	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-	github.com/firecracker-microvm/firecracker-go-sdk v0.19.0 => github.com/smira/firecracker-go-sdk v0.19.1-0.20200110185541-4fce8cba9f84
+	github.com/firecracker-microvm/firecracker-go-sdk v0.19.0 => github.com/smira/firecracker-go-sdk v0.19.1-0.20200124123725-fb8b25794e34
```
andrewrynhard: Maybe I am misunderstanding but the PR comment says this is supposed to move to the official SDK?

pkg/config/types/v1alpha1/generate/options.go
andrewrynhard: This works for now I think. We can figure out a better way in the future IMO.

go.mod
smira: yeah, it's still a branch in my fork with three PRs mentioned above merged together, previous branch contained some hacks I did to pass `ip=` manually and validation was disabled, so this is much cleaner version, if/when those PRs are merged, we can just switch to master now

andrewrynhard: Ahh gotcha.

pkg/net/net.go
```diff
@@ -24,7 +24,7 @@ func IPAddrs() (ips []net.IP, err error) {
 	}
 
 	for _, a := range addrs {
-		if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+		if ipnet, ok := a.(*net.IPNet); ok && ipnet.IP.IsGlobalUnicast() {
```
smira: this is a major change here, the problem was that once we assign `link-local` address 169.254.x.x to the VM in network config, it gets picked up as node address, and this kills the etcd cluster obviously.

`IsGlobalUnicast` is defined as the following: 

```
	return (len(ip) == IPv4len || len(ip) == IPv6len) &&
		!ip.Equal(IPv4bcast) &&
		!ip.IsUnspecified() &&
		!ip.IsLoopback() &&
		!ip.IsMulticast() &&
		!ip.IsLinkLocalUnicast()
```

it fiters out multicast, link-local, loopback, unspecified (zeroes), but it allows for private and any public IPs. this seems to be a reasonable change, but still might be a major change

PR #1838, opened 2020-01-30
feat(networkd): Make healthcheck perform a check

bradbeam:
This implements an actual health check for networkd. We use the arp table ( ip neighbors )
to determine if the machine is actively sending traffic. We should see at least one entry
with a REACHABLE state during normal operating conditions.

### DIFF
--- a/internal/app/networkd/pkg/reg/reg.go
+++ b/internal/app/networkd/pkg/reg/reg.go
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+// Package reg provides the gRPC network service implementation.
 package reg
 
 import (
@@ -155,14 +156,52 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
 }
 
 // Check implements the Health api and provides visibilty into the state of networkd.
-// Ready signifies the daemon (api) is healthy and ready to serve requests.
+// We determine networkd health based on neighbor (arp table) entries.
+// Under normal operating circumstances, there should be a steady state of reachable neighbor entries.
+// If we get in a situation where all of our neighbor entries are stale, then we're in trouble.
 func (r *Registrator) Check(ctx context.Context, in *empty.Empty) (reply *healthapi.HealthCheckResponse, err error) {
+	// Set initial state to unknown
 	reply = &healthapi.HealthCheckResponse{
 		Messages: []*healthapi.HealthCheck{
 			{
-				Status: healthapi.HealthCheck_SERVING,
+				Status: healthapi.HealthCheck_UNKNOWN,
 			},
 		},
+	}
+
+	var neighbors []rtnetlink.NeighMessage
+
+	neighbors, err = r.Conn.Neigh.List()
+	if err != nil {
+		return reply, err
+	}
+
+	// After getting a list of neighbors we can upgrade to not serving
+	reply.Messages[0].Status = healthapi.HealthCheck_NOT_SERVING
+
+	// Find at least one neighbor in a reachable state
+	for _, neighbor := range neighbors {
+		// Verify neighbor is associated with a link
+		// managed by networkd ( skip cni links )
+		link, err := r.Conn.Link.Get(neighbor.Index)
+		if err != nil {
+			continue
+		}
+
+		if _, ok := r.Networkd.Interfaces[link.Attributes.Name]; !ok {
+			continue
+		}
+
+		// Verify neighbor state
+		switch neighbor.State {
+		case unix.NUD_REACHABLE:
+			fallthrough
+		case unix.NUD_STALE:
+			fallthrough
+		case unix.NUD_DELAY:
+			reply.Messages[0].Status = healthapi.HealthCheck_SERVING
+			return reply, nil
+		}
 	}
 
 	return reply, nil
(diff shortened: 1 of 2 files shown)


internal/app/networkd/pkg/reg/reg_test.go
```diff
@@ -118,7 +118,7 @@ func (suite *NetworkdSuite) TestHealthAPI() {
 	nClient := healthapi.NewHealthClient(conn)
 	hcResp, err := nClient.Check(context.Background(), &empty.Empty{})
 	suite.Assert().NoError(err)
-	suite.Assert().Equal(healthapi.HealthCheck_SERVING, hcResp.Messages[0].Status)
+	suite.Assert().Equal(healthapi.HealthCheck_NOT_SERVING, hcResp.Messages[0].Status)
```
bradbeam: Feels janky setting the tests to this, but we don't have a good way to ensure that there is a valid arp entry during tests. So we'll just leave it here to verify basic ( netlink ) functionality.

bradbeam:
would like to run this through conformance or something before merging so we can ensure things are stable for a longer run.

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -155,16 +155,39 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+	}
+
+	// After getting a list of neighbors we can upgrade to not serving
+	reply.Messages[0].Status = healthapi.HealthCheck_NOT_SERVING
+
+	// Find at least one neighbor in a reachable state
+	for _, neighbor := range neighbors {
+		// State = 2 == reachable
+		// State = 4 == stale
+		if neighbor.State == 2 {
```
bradbeam: Wonder if we should include `neighbor.State == 2 || neighbor.State == 4`. I think the transition time between the two states is ~5-10s of inactivity, so looking for reachable only might be too tight.

smira: can we also use some constants for these so that is more readable?

bradbeam: yeah, couldn't find them last night but looks like these map up:
https://github.com/golang/sys/blob/63cb32ae39b28d6bb8e7e215c1fc39dd80dcdb02/unix/linux/types.go#L593-L601

jsimonetti: Would it be smart to test that the interface index matches interfaces managed by networkd?

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -155,16 +156,54 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
+	// Find at least one neighbor in a reachable state
+	for _, neighbor := range neighbors {
+		// Verify neighbor is associated with a link
+		// managed by networkd ( skip cni links )
+		link, err := r.Conn.Link.Get(neighbor.Index)
+		if err != nil {
+			continue
+		}
+
+		if _, ok := r.Networkd.Interfaces[link.Attributes.Name]; !ok {
```
smira: I guess this map might be mutated while API is running and at the same time networkd is discovering new interfaces?

so potential data race here, with map it's even worse potentially

internal/app/networkd/pkg/reg/reg.go
```diff
@@ -155,16 +156,54 @@ func toCIDR(family uint8, prefix net.IP, prefixLen int) string {
...
 			{
-				Status: healthapi.HealthCheck_SERVING,
+				Status: healthapi.HealthCheck_UNKNOWN,
 			},
 		},
 	}
 
+	var neighbors []rtnetlink.NeighMessage
+
+	neighbors, err = r.Conn.Neigh.List()
```
smira: never checked that, but if this socket-like thing, would it be bad if we do concurrent requests to it (from API and networkd doing its job)

bradbeam: https://github.com/mdlayher/netlink/blob/master/conn.go#L18-L20
https://github.com/mdlayher/netlink/blob/master/conn.go#L116-L118

(rtnetlink uses netlink under the covers )
From what I gather each netlink request acquires a lock in the kernel, so I don't believe this is a concern. 
We _may_ want to do some sort of pooling for these checks, but I'm not sure we're at a high enough threshold per connection. I'm not sure what the value of the threshold would be ( multiple a second?), but we're currently only at 1 per healthcheck interval + whatever an end user would add on top of that.

internal/app/networkd/pkg/reg/reg.go
bradbeam: I'm pretty sure this should be fine with the given workflow. networkd.Interfaces only gets populated during the `Configure()` phase and since we're only doing key lookups ( we don't care about the value ) I'm not sure I see there being any harm this way. 
There's probably another ticket worth to potentially refactor it from a map to something more useful, so maybe something for a follow up pr.

internal/app/networkd/pkg/reg/reg.go
smira: actually `Execute()` (method behind all the higher-level functions) does locking: https://github.com/mdlayher/netlink/blob/master/conn.go#L122, so we should be good

internal/app/networkd/pkg/reg/reg.go
smira: yep, we start actually API server when initial configuration is done: https://github.com/talos-systems/talos/blob/master/internal/app/networkd/main.go#L46-L60

so this shoudl be safe.

PR #1844, opened 2020-01-30
chore: implement reboot test

smira:
Reboot test does node-by-node reboots followed by cluster health checks
(same as done by provisioner).

Fixed bug with `Read()` returning `Reader` instead of `ReadCloser`
(minor).

Allowed `bootkube` to be `Skipped` (for rebooted node).

Added support for doing checks via provided client instance.

Implemented generic capabilities to skip tests based on cluster
platform.

### DIFF
--- a/internal/app/machined/internal/api/reg/reg.go
+++ b/internal/app/machined/internal/api/reg/reg.go
@@ -78,7 +78,9 @@ func (r *Registrator) Register(s *grpc.Server) {
 // Reboot implements the machineapi.MachineServer interface.
 func (r *Registrator) Reboot(ctx context.Context, in *empty.Empty) (reply *machineapi.RebootResponse, err error) {
 	reply = &machineapi.RebootResponse{
-		Messages: []*machineapi.Reboot{},
+		Messages: []*machineapi.Reboot{
+			{},
+		},
 	}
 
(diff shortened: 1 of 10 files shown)


smira:
On top of #1840 until that one is merged...

internal/app/machined/internal/api/reg/reg.go
```diff
@@ -78,7 +78,9 @@ func (r *Registrator) Register(s *grpc.Server) {
 // Reboot implements the machineapi.MachineServer interface.
 func (r *Registrator) Reboot(ctx context.Context, in *empty.Empty) (reply *machineapi.RebootResponse, err error) {
 	reply = &machineapi.RebootResponse{
-		Messages: []*machineapi.Reboot{},
+		Messages: []*machineapi.Reboot{
+			{},
```
smira: looks like the problem with broken API response was this thing: we actually returned empty response instead of a single message (as expected).

smira:
AWS:

```
--- PASS: TestIntegration/api.RebootSuite (265.81s)
```

~ 44 sec per node

GCP:

```
--- PASS: TestIntegration/api.RebootSuite (793.51s)
```

~ 132 sec per node

PR #1848, opened 2020-02-01
chore(networkd): Report on errors during interface configuration

bradbeam:
This DRYs up the interface configuration and adds in an error channel to capture
any issues that come up from interface configuration. These errors are still
treated as non-fatal, but should provide some additional insight.

Signed-off-by: Brad Beam <brad.beam@b-rad.info>

### DIFF
--- a/internal/app/networkd/pkg/networkd/networkd.go
+++ b/internal/app/networkd/pkg/networkd/networkd.go
@@ -160,77 +160,23 @@ func New(config runtime.Configurator) (*Networkd, error) {
 // configuration, and any addressing that is needed. We care about ordering
 // here so that we can ensure any links that make up a bond will be in
 // the correct state when we get to bonding configuration.
-// nolint: gocyclo
+//
+//nolint: gocyclo
 func (n *Networkd) Configure() (err error) {
-	var wg sync.WaitGroup
-
-	log.Println("configuring non-bonded interfaces")
-
-	for _, iface := range n.Interfaces {
-		// Skip bonded interfaces so we can ensure a proper base state
-		if iface.Bonded {
-			continue
+	// Configure non-bonded interfaces first so we can ensure basic
+	// interfaces exist prior to bonding
+	for _, bonded := range []bool{false, true} {
+		if bonded {
+			log.Println("configuring bonded interfaces")
+		} else {
+			log.Println("configuring non-bonded interfaces")
 		}
 
-		wg.Add(1)
-
-		go func(netif *nic.NetworkInterface, wg *sync.WaitGroup) {
-			defer wg.Done()
-
-			// Ensure link exists
-			if err = netif.Create(); err != nil {
-				log.Printf("error creating nic: %v", err)
-				return
-			}
-
-			if err = netif.Configure(); err != nil {
-				log.Printf("error configuring nic: %v", err)
-				return
-			}
-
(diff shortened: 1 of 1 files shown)


internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -410,3 +348,52 @@ func (n *Networkd) SetReady() {
...
+		if iface.Bonded != bonded {
+			continue
+		}
+
+		wg.Add(1)
+
+		go func(netif *nic.NetworkInterface, wg *sync.WaitGroup) {
+			defer wg.Done()
+
+			log.Printf("setting up %s", netif.Name)
```
smira: one potential way to DRY up this function a bit is to do something like:

```go
errCh <- func() error {
   			if err := netif.Create(); err != nil {
				return fmt.Errorf("error creating nic: %v", err)
			}
                        
                        if err = ...
        }
}
```

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -410,3 +348,52 @@ func (n *Networkd) SetReady() {
...
+
+			// Ensure link exists
+			if err = netif.Create(); err != nil {
```
smira: this write is a data race, variable `err` is used from multiple goroutines

should be `if err := netif.Create()`

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -410,3 +348,52 @@ func (n *Networkd) SetReady() {
...
+	)
+
+	for _, iface := range n.Interfaces {
+		if iface.Bonded != bonded {
+			continue
+		}
+
+		wg.Add(1)
+
+		go func(netif *nic.NetworkInterface, wg *sync.WaitGroup) {
```
smira: usually `wg` is not passed to the closure, as it's not a for loop variable, implicit variable access from outer scope is just fine

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -160,78 +160,16 @@ func New(config runtime.Configurator) (*Networkd, error) {
 // configuration, and any addressing that is needed. We care about ordering
 // here so that we can ensure any links that make up a bond will be in
 // the correct state when we get to bonding configuration.
-// nolint: gocyclo
 func (n *Networkd) Configure() (err error) {
-	var wg sync.WaitGroup
-
-	log.Println("configuring non-bonded interfaces")
```
smira: should we keep those log message so it's easy to see what's going on?

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -410,3 +356,45 @@ func (n *Networkd) SetReady() {
...
+
+		count++
+
+		go func(netif *nic.NetworkInterface) {
+			log.Printf("setting up %s", netif.Name)
+
+			errCh <- func() error {
+				// Ensure link exists
+				if err := netif.Create(); err != nil {
+					return fmt.Errorf("error creating nic: %v", err)
```
smira: I guess we need to add NIC name to the error message to make it easy to match error & config

andrewrynhard: Good point

PR #1850, opened 2020-02-01
fix: bind etcd to IPv6 if available

### LINKED ISSUE #1842: etcd does not bind to IPv6 interface
etcd does not appear to be binding to the IPv6 address(es) of the node, listening only on IPv4.  This will be a problem for IPv6-only installations.

### LINKED ISSUE #1843: etcd PKI certs do not make SAN for full hostname
etcd certs are only valid for the short hostname, not the FQDN.

Ulexus:
If an IPv6 address is available, etcd should bind to `[::]` instead of
`0.0.0.0`.  This will cause etcd to listen on both IPv4 and IPv6
interfaces.

Additionally, this fixes the SAN list for the etcd certificate
generation to include the FQDN of the host.

Fixes #1842
Fixes #1843

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -86,7 +86,7 @@ func (e *Etcd) Condition(config runtime.Configurator) conditions.Condition {
 
 // DependsOn implements the Service interface.
 func (e *Etcd) DependsOn(config runtime.Configurator) []string {
-	return []string{"containerd"}
+	return []string{"containerd", "networkd"}
 }
 
 // Runner implements the Service interface.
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -164,15 +164,24 @@ func generatePKI(config runtime.Configurator) (err error) {
...
 	hostname, err := os.Hostname()
 	if err != nil {
 		return fmt.Errorf("failed to get hostname: %w", err)
 	}
 
+	dnsNames, err := net.DNSNames()
+	if err != nil {
+		return fmt.Errorf("failed to get host DNS names: %w", err)
+	}
+	dnsNames = append(dnsNames, "localhost")
```
andrewrynhard: Looks like linter is mad about this and want a space above this line.

Ulexus: Yeah, just fixed.

Ulexus:
re: testing, yes, definitely... but IPv6 overall in Talos doesn't really work yet.  I'm gradually working through it, though, and most of the problems are on the kubernetes side.

Ulexus:
testing etcd itself on IPv6 should be fine, though

PR #1868, opened 2020-02-09
fix: don't proxy gRPC unix connections

andrewrynhard:
The default gRPC dialer honors proxy environment variables, which causes
local unix socket connections to attempt to go through the proxy. This
fixes that by using a custom dialer.

### DIFF
--- a/pkg/grpc/dialer/dialer_test.go
+++ b/pkg/grpc/dialer/dialer_test.go
@@ -1,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package dialer_test
+
+import "testing"
+
+func TestEmpty(t *testing.T) {
+	// added for accurate coverage estimation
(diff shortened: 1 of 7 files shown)


internal/app/machined/main.go
```diff
@@ -82,6 +82,10 @@ func init() {
 	// is made.
 	http.DefaultClient.Transport = &http.Transport{
 		Proxy: func(req *http.Request) (*url.URL, error) {
+			if req.URL.Scheme == "unix" {
+				return req.URL, nil
+			}
+
```
andrewrynhard: Don't think we need this.

pkg/grpc/dialer/dialer_test.go
```diff
@@ -0,0 +1,14 @@
...
+package dialer_test
+
+import "testing"
+
+func TestEmpty(t *testing.T) {
+	// added for accurate coverage estimation
+	//
+	// please remove it once any unit-test is added
+	// for this package
+}
```
andrewrynhard: Still looking for a good way to test this.

alex1989hu:
Image made from 1345f3fff works great 👍

PR #1874, opened 2020-02-12
chore: bring in newer kernel build

rsmitty:
This PR will bring in the kernel build that enables USB + USB network
adapters. This allows talos to be booted on dev boxes that don't have a
dedicated ethernet port.


andrewrynhard:
@rsmitty I think we can close this now... I pulled in newer packages in another PR that has these changes.

PR #1875, opened 2020-02-12
chore: implement loadbalancer for firecracker provisioner

smira:
This PR contains generic simple TCP loadbalancer code, and glue code for
firecracker provisioner to use this loadbalancer.

K8s control plane is passed through the load balancer, and Talos API is
passed only to the init node (for now, as some APIs, including
kubeconfig, don't work with non-init node).


smira:
This PR also re-enables reboot tests.

PR #1876, opened 2020-02-13
chore: push installer & talos images to the CI registry on every build

smira:
This enables a way to run the matching installer image in firecracker
tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SHA ?= $(shell git describe --match=none --always --abbrev=8 --dirty)
 TAG ?= $(shell git describe --tag --always --dirty)
 BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 REGISTRY_AND_USERNAME := $(REGISTRY)/$(USERNAME)
+DOCKER_LOGIN_ENABLED ?= true
 
 ARTIFACTS := _out
 TOOLS ?= autonomy/tools:v0.1.0-1-g1c0644f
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -218,17 +219,19 @@ e2e-%: $(ARTIFACTS)/$(INTEGRATION_TEST_DEFAULT_TARGET)-amd64 $(ARTIFACTS)/sonobu
 
 .PHONY: login
 login: ## Logs in to the configured container registry.
+ifeq ($(DOCKER_LOGIN_ENABLED), true)
```
smira: disable login for local registry (it's not authenticated)

hack/test/e2e-firecracker.sh
```diff
@@ -7,6 +7,18 @@ source ./hack/test/e2e.sh
 PROVISIONER=firecracker
 CLUSTER_NAME=e2e-${PROVISIONER}
 
+case "${REGISTRY:-false}" in
+  registry.ci.svc:5000)
+    REGISTRY_ADDR=`python -c "import socket; print socket.gethostbyname('registry.ci.svc')"`
```
smira: resolve hostname before passing it to the firecracker; firecracker runs with default DNS servers (8.8.8.8), so it can't resolve cluster-local addresses

hack/test/e2e-firecracker.sh
```diff
@@ -16,9 +28,9 @@ function create_cluster {
     --memory 2048 \
     --cpus 2.0 \
     --cidr 172.20.0.0/24 \
-    --init-node-as-endpoint \
```
smira: this isn't required anymore, as firecracker LB running on the bridge provides default endpoint

PR #1884, opened 2020-02-17
fix: PodCIDR, ServiceCIDR should be comma sets

### LINKED ISSUE #1883: Need to support two Pod and Service CIDRs for dual stack IPv6
Right now, the machine config allows the specification of multiple PodCIDRs and ServiceCIDRs, but only the first is sent to the bootkube template.  The set should be consolidated to a comma-delimited string of the two.

Ulexus:
PodCIDRs and ServiceCIDRs are now returned as comma-delimited set of
their slices, as per the documenation for kube-api-server and
kube-controller-manager in dual-stack configurations.

Ref: https://kubernetes.io/docs/concepts/services-networking/dual-stack/#enable-ipv4-ipv6-dual-stack

Fixes #1883

### DIFF
--- a/pkg/config/types/v1alpha1/v1alpha1_configurator.go
+++ b/pkg/config/types/v1alpha1/v1alpha1_configurator.go
@@ -379,7 +379,7 @@ func (c *ClusterConfig) PodCIDR() string {
 		return constants.DefaultPodCIDR
 	}
 
-	return c.ClusterNetwork.PodSubnet[0]
+	return strings.Join(c.ClusterNetwork.PodSubnet, ",")
 }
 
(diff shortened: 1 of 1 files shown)


Ulexus:
Note that the referenced documentation is not wholly correct.  The flag for kube-controller-manager remains `--cluster-cidr` not `--cluster-cidrs`, and modifications to kube-api-server are also required, specifically its `--service-cluster-ip-range` flag.

This PR correctly consolidates the lists, regardless, and while dual stack support does still additionally require the `IPv6DualStack` feature gate to be set for all the controlplane and kubelet components, that is beyond the scope of this change.

PR #1889, opened 2020-02-18
fix: allow kublet to handle multiple service CIDRs

### LINKED ISSUE #1888: kubelet dns calculation fails in dual stack
In a dual stack configuration, the ServiceCIDR is a set of CIDRs, which confuses the DNS service IP calculation of our kubelet service runner.

Ulexus:
Handle multiple service CIDRs (such as for dual-stack configurations)
cleanly, calculating the DNS service IP for each.

Fixes #1888


Ulexus:
Step by tiny step... but we'll get there. :)

PR #1891, opened 2020-02-18
fix: fix reset command

rsmitty:
This PR will fix the reset command to actually wipe the system disk as
expected. Under the hood, it creates a whole new reset sequence which does some of the same things as the upgrade sequence, but just shuts down at the end after wiping disks.


cmd/osctl/pkg/client/client.go
```diff
@@ -240,8 +240,8 @@ func (c *Client) Restart(ctx context.Context, namespace string, driver common.Co
 }
 
 // Reset implements the proto.OSClient interface.
-func (c *Client) Reset(ctx context.Context) (err error) {
-	_, err = c.MachineClient.Reset(ctx, &empty.Empty{})
+func (c *Client) Reset(ctx context.Context, graceful bool) (err error) {
+	_, err = c.MachineClient.Reset(ctx, &machineapi.ResetRequest{})
```
smira: `graceful` is lost here?

smira:
I think it's possible to implement integration test for this feature?

one caveat might be that Firecracker doesn't support "POWEROFF" afaik, so we need to make it reboot probably to come up with wiped disk and go through reinstall? 

idk if we can detect reinstall in a meaningful way? 

not sure if I understand `reset` completely, i.e. should it come back clean or just sit powered off?

rsmitty:
Definitely expecting this to be a destructive command. It should wipe the disks and power the machine off. You're right about losing the graceful bit though, thanks for that!

andrewrynhard:
@smira The way Talos is designed is that a reset sort of implies a re-install. On boot we check for a partition with the `EPHEMERAL` label (as I'm sure you know) and perform an install if not found. So in this case it will be gone, and an install will be performed again. Now the debate is, reboot vs poweroff. I think maybe reset should have another flag (`reboot`) that we can use to decide that because I can see it being useful to either shutdown or reboot. @rsmitty thoughts?

PR #1894, opened 2020-02-19
fix: skip links without a carrier

andrewrynhard:
This change ensures that we avoid setting up interfaces without a
carrier.

### DIFF
--- a/internal/app/networkd/pkg/networkd/misc.go
+++ b/internal/app/networkd/pkg/networkd/misc.go
@@ -13,26 +13,61 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/jsimonetti/rtnetlink"
+	"golang.org/x/sys/unix"
+
 	talosnet "github.com/talos-systems/talos/pkg/net"
 )
 
-// filterInterfaceByName filters network links by name so we only mange links
+// filterInterfaces filters network links by name so we only mange links
 // we need to.
-func filterInterfaceByName(links []net.Interface) (filteredLinks []net.Interface) {
-	for _, link := range links {
+//
+// nolint: gocyclo
+func filterInterfaces(interfaces []net.Interface) (filtered []net.Interface, err error) {
+	var conn *rtnetlink.Conn
+
+	for _, iface := range interfaces {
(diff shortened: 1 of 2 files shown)


andrewrynhard:
@jsimonetti ... pinging you here to see if this PR makes sense.

jsimonetti:
lgtm

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -13,26 +13,47 @@ import (
...
+			filtered = append(filtered, iface)
+		case strings.HasPrefix(iface.Name, "lo"):
+			filtered = append(filtered, iface)
+		case strings.HasPrefix(iface.Name, "bond"):
+			filtered = append(filtered, iface)
 		}
 	}
 
-	return filteredLinks
+	return filtered
```
jsimonetti: Missing return value

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -13,26 +13,47 @@ import (
...
 )
 
-// filterInterfaceByName filters network links by name so we only mange links
+// filterInterfaces filters network links by name so we only mange links
 // we need to.
-func filterInterfaceByName(links []net.Interface) (filteredLinks []net.Interface) {
-	for _, link := range links {
+func filterInterfaces(interfaces []net.Interface) (filtered []net.Interface, err error) {
+	var conn *rtnetlink.Conn
+	conn, err = rtnetlink.Dial(nil)
```
smira: I guess we need to close this?

internal/app/networkd/pkg/networkd/misc.go
```diff
@@ -13,26 +13,47 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, iface := range interfaces {
+		link, err := conn.Link.Get(uint32(iface.Index))
+		if err != nil {
+			return nil, err
```
smira: should we skip those links which give back an error? as otherwise this fails the whole flow?

PR #1895, opened 2020-02-19
fix: add reboot flag to reset command

andrewrynhard:
This exposes the reboot option for thee reset API by adding a `--reboot`
flag to the CLI.


cmd/osctl/cmd/reset.go
```diff
@@ -37,5 +37,6 @@ var resetCmd = &cobra.Command{
 
 func init() {
 	resetCmd.Flags().BoolVar(&graceful, "graceful", true, "if true, attempt to cordon/drain node and leave etcd (if applicable)")
+	resetCmd.Flags().BoolVar(&graceful, "reboot", true, "if true, reboot the node after resetting (default is to shutdown)")
```
smira: `graceful` once again and no code to send to the request? :question:

andrewrynhard: woops

PR #1897, opened 2020-02-19
feat: support sending machine info

andrewrynhard:
This allows users to specify well known query parameters in `talos.config`.
The only supported parameter in this change is `uuid`. This will send
the node's UUID determined from SMBIOS along with the request for the
config.

### DIFF
--- a/internal/pkg/runtime/platform/metal/metal.go
+++ b/internal/pkg/runtime/platform/metal/metal.go
@@ -7,10 +7,14 @@ package metal
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net"
+	"net/url"
 	"path/filepath"
 
 	"golang.org/x/sys/unix"
+
+	"github.com/talos-systems/go-smbios/smbios"
 
 	"github.com/talos-systems/talos/internal/pkg/kernel"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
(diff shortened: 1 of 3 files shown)


internal/pkg/runtime/platform/metal/metal.go
```diff
@@ -38,6 +40,34 @@ func (m *Metal) Configuration() ([]byte, error) {
...
+				if err != nil {
+					return nil, err
+				}
+
+				uuid, err := s.SystemInformation().UUID()
+				if err != nil {
+					return nil, err
+				}
+
+				values.Set("uuid", uuid.String())
```
smira: does this actually work? my understanding was that it won't update `u` ?

Ulexus: This value is not guaranteed to be filled, so I would caution against using it directly.

internal/pkg/runtime/platform/metal/metal.go
```diff
@@ -38,6 +40,34 @@ func (m *Metal) Configuration() ([]byte, error) {
...
+	if len(values) > 0 {
+		for key := range values {
+			switch key {
+			case "uuid":
+				s, err := smbios.New()
+				if err != nil {
+					return nil, err
+				}
+
+				uuid, err := s.SystemInformation().UUID()
```
Ulexus: This value is not guaranteed to be filled, so I would caution against using it directly.

internal/pkg/runtime/platform/metal/metal.go
smira: I guess you need `u.RawQuery = values.Encode()` before you do `u.String()`

internal/pkg/runtime/platform/metal/metal.go
andrewrynhard: Yeah this is something we have seen. The good news is that this is completely optional, and really only here for our metadata server we are building (for now). It should work _most_ of the time, but should be enough for now to get started on some of the bare metal stuff we are doing.

PR #1898, opened 2020-02-19
fix: ignore link-local and other non routable addresses

smira:
This is backport of
https://github.com/talos-systems/talos/commit/fae5e6915d32c41b464978baf7ba335264963e58#diff-da3509c3f67d46ea31daf53392216e68.

This fixes `etcd` bootstrap with firecracker provisioner, as firecracker
provisions link-local addresses on each node and `etcd` code might pick
them up instead of real interface address.

### DIFF
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -24,7 +24,7 @@ func IPAddrs() (ips []net.IP, err error) {
 	}
 
 	for _, a := range addrs {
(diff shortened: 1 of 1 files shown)


smira:
Tested new `installer` image locally, and this allows upgrade tests to build 3-master 0.3.2-xyz cluster

PR #1904, opened 2020-02-21
fix: unmount bind mounts for system (fixes upgrade stuck on disk busy)

smira:
Bind mounts created by `machine.File` section of the config (e.g. when
appending to config outside of `/var`) are not unmounted with other
flows and keep ephemeral disk busy.

New task processes such mounts and unmounts them. Also added debugging
dump of `/proc/mounts` on system disk busy error.


smira:
I'm not enabling upgrade tests 0.4 -> latest, as we need first 0.4 which is safe to upgrade from (this PR merged should be good base)

PR #1905, opened 2020-02-21
test: enable upgrade tests 0.4.x -> latest

smira:
With the fix #1904, it's now possible to upgrade 0.4.x with
`machine.File` extra files (caused by registry mirror for
registry.ci.svc).

Bump resources for upgrade tests in attempt to speed it up.


internal/integration/provision/upgrade.go
```diff
@@ -54,7 +54,7 @@ type upgradeSpec struct {
 
 const (
 	talos03Version = "v0.3.2-1-g71ac6696"
-	talos04Version = "v0.4.0-alpha.5"
+	talos04Version = "v0.4.0-alpha.5-19-g8913d9df"
```
andrewrynhard: Should we use `latest` here?

smira: this is source version to upgrade from, so I think we should use the earliest version upgrade works from?

smira: well, actually it's both destination and source, it's source for 0.4 -> latest and destination for 0.3 -> 0.4 probably I should split them ?

smira: for 0.4 -> latest upgrade, it makes sense to use the earliest version which works for upgrades, and for 0.3 -> 0.4 it probably makes sense to use the latest.

the problem with :latest is that we don't know the exact version so it's hard to assert on upgrade correctness, we assert that Talos version after upgrade matches expected

andrewrynhard: 👍

go.mod
```diff
@@ -59,7 +59,7 @@ require (
 	github.com/ryanuber/columnize v2.1.0+incompatible
 	github.com/smira/go-xz v0.0.0-20150414201226-0c531f070014
 	github.com/spf13/cobra v0.0.5
-	github.com/stretchr/testify v1.5.0
+	github.com/stretchr/testify v1.5.1
```
smira: had to update because of https://github.com/stretchr/testify/pull/894

1.5.0 narrowed `suite.T()` to fake interface which doesn't expose full of `testing.T`, while I wanted to access `.Failed()` method

smira:
I added 'crashdump' feature to dump logs of VMs, containers on test failure. This should further help to debug the issues with tests.

PR #1907, opened 2020-02-24
fix: ensure proxy is used when fetching additional manifests for bootkube

salkin:
Http getter library is by default using it's own client. This ensure the default transport is used for fetching additional manifests.


andrewrynhard:
@salkin do you mind rebasing against master?

salkin:
ok, will do.

PR #1909, opened 2020-02-26
chore: update pkgs & tools for Go 1.14

smira:
See also:

* https://github.com/talos-systems/tools/pull/89
* https://github.com/talos-systems/pkgs/pull/103

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ENV PATH /toolchain/bin:/toolchain/go/bin
 RUN ["/toolchain/bin/mkdir", "/bin", "/tmp"]
 RUN ["/toolchain/bin/ln", "-svf", "/toolchain/bin/bash", "/bin/sh"]
 RUN ["/toolchain/bin/ln", "-svf", "/toolchain/etc/ssl", "/etc/ssl"]
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.23.3
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.23.6
 RUN cd $(mktemp -d) \
     && go mod init tmp \
     && go get mvdan.cc/gofumpt/gofumports \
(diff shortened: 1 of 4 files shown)


pkg/config/machine/machine.go
```diff
@@ -255,8 +255,6 @@ func (cfg *RegistryTLSConfig) GetTLSConfig() (*tls.Config, error) {
 		tlsConfig.InsecureSkipVerify = true
 	}
 
-	tlsConfig.BuildNameToCertificate()
```
smira:
```
 > [lint-go 2/4] RUN --mount=type=cache,target=/.cache/go-build golangci-lint run --config .golangci.yml:
#51 25.86 pkg/config/machine/machine.go:258:2: SA1019: tlsConfig.BuildNameToCertificate is deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.  (staticcheck)
#51 25.86 	tlsConfig.BuildNameToCertificate()
#51 25.86 	^
```

Dockerfile
```diff
@@ -12,7 +12,7 @@ ENV PATH /toolchain/bin:/toolchain/go/bin
 RUN ["/toolchain/bin/mkdir", "/bin", "/tmp"]
 RUN ["/toolchain/bin/ln", "-svf", "/toolchain/bin/bash", "/bin/sh"]
 RUN ["/toolchain/bin/ln", "-svf", "/toolchain/etc/ssl", "/etc/ssl"]
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.23.3
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.23.6
```
smira: it seems that 1.23.3 is not compatible with Go 1.14

pkg/config/machine/machine.go
smira: more context: https://godoc.org/crypto/tls#Config.BuildNameToCertificate

pkg/constants/constants.go
```diff
@@ -192,7 +192,7 @@ const (
 	TrustdPort = 50001
 
 	// DefaultContainerdVersion is the default container runtime version
-	DefaultContainerdVersion = "1.3.2"
+	DefaultContainerdVersion = "1.3.3"
```
smira: looks like we missed some `pkgs` update to Talos

smira:
rebased without changes to give it another test run

Makefile
```diff
@@ -7,8 +7,8 @@ REGISTRY_AND_USERNAME := $(REGISTRY)/$(USERNAME)
 DOCKER_LOGIN_ENABLED ?= true
 
 ARTIFACTS := _out
-TOOLS ?= autonomy/tools:v0.1.0-1-g1c0644f
-GO_VERSION ?= 1.13
+TOOLS ?= autonomy/tools:v0.1.0-2-g4d7ce7d
+GO_VERSION ?= 1.14
```
smira: one more push to bump Go version here (used in `unit-tests-race`)

PR #1913, opened 2020-02-27
feat: generate kubeconfig on the fly on request

smira:
This extracts admin kubeconfig generation out of bootkube, now based on
Talos x509 library. On each API request for `kubeconfig`, config is
generated on the fly and sent back on the wire.

This fixes two issues:

* any master node can now generate `kubeconfig` (worker nodes can do
that too, but that should probably change in the future)
* after upgrade-and-wipe the disk scenario, `osctl kubeconfig` still
works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ controlplane.yaml
 join.yaml
 docgen
 talosconfig
-kubeconfig
+/kubeconfig
 hack/test/libvirt/matchbox/assets/*
 !hack/test/libvirt/matchbox/assets/.gitkeep
 
--- a/cmd/osctl/cmd/loadbalancer_launch.go
+++ b/cmd/osctl/cmd/loadbalancer_launch.go
@@ -14,8 +14,9 @@ import (
 )
 
 var loadbalancerLaunchCmdFlags struct {
-	addr      string
-	upstreams []string
+	addr             string
+	upstreams        []string
+	apidOnlyInitNode bool
 }
 
 // loadbalancerLaunchCmd represents the loadbalancer-launch command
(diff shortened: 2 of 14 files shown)


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -271,22 +271,12 @@ func generateAssets(config runtime.Configurator) (err error) {
 	urls = append(urls, config.Cluster().CertSANs()...)
 	altNames := altNamesFromURLs(urls)
 
-	block, _ = pem.Decode(config.Cluster().CA().Crt)
-	if block == nil {
-		return errors.New("failed to Kubernetes CA certificate")
-	}
-
-	k8sCA, err := x509.ParseCertificate(block.Bytes)
+	k8sCA, err := config.Cluster().CA().GetCert()
```
smira: replaced with common method to decode to cert/key

pkg/config/cluster/cluster.go
```diff
@@ -11,18 +11,33 @@ import (
 	"github.com/talos-systems/talos/pkg/crypto/x509"
 )
 
+// Name defines subset of Cluster to provide cluster name.
+type Name interface {
```
smira: I did this split-the-big-interface trick here to provide a way to have smaller mock in the test for Cluster config (in fact part of Cluster config required for cert generation)

cmd/osctl/cmd/loadbalancer_launch.go
```diff
@@ -34,11 +34,6 @@ var loadbalancerLaunchCmd = &cobra.Command{
 				upstreams[i] = fmt.Sprintf("%s:%d", loadbalancerLaunchCmdFlags.upstreams[i], port)
 			}
 
-			// for apid, add only init node for now (first item)
```
smira: load-balancing to apid master nodes should work now hopefully

.gitignore
```diff
@@ -6,7 +6,7 @@ controlplane.yaml
 join.yaml
 docgen
 talosconfig
-kubeconfig
+/kubeconfig
```
andrewrynhard: The leading slash is needed?

smira: yep, as I think intention was to ignore `kubeconfig` at the root (from `osctl kubeconfig`), but in this form it ignores any file/directory named `kubeconfig`, as I added `internal/pkg/kubeconfig`, I need to make sure it's not ignored

cmd/osctl/cmd/loadbalancer_launch.go
smira: updated this to be configurable for upgrading from older Talos

PR #1917, opened 2020-02-28
feat: allow for persistence of config data

rsmitty:
This PR will allow users to set the `persist: true` value in their
config data to tell talos not to re-pull the config data at each reboot.
The default will still remain as a "pull every time" methodology in order
to encourage immutability by default.

### DIFF
--- a/docs/website/content/v0.3/en/configuration/v1alpha1.md
+++ b/docs/website/content/v0.3/en/configuration/v1alpha1.md
@@ -72,6 +72,19 @@ Valid Values:
 
 - ``v1alpha1``
 
+#### persist
+
+Indicates whether to pull the machine config upon every boot.
+
+Type: `bool`
+
+Valid Values:
+
+- `true`
+- `yes`
+- `false`
+- `no`
+
 #### machine
 
 Provides machine specific configuration options.
--- a/internal/app/machined/internal/phase/config/config.go
+++ b/internal/app/machined/internal/phase/config/config.go
@@ -9,19 +9,25 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 
 	"github.com/talos-systems/talos/internal/app/machined/internal/phase"
 	"github.com/talos-systems/talos/internal/pkg/runtime"
+	"github.com/talos-systems/talos/pkg/config"
 	"github.com/talos-systems/talos/pkg/constants"
 )
 
 // Task represents the Task task.
-type Task struct{}
+type Task struct {
+	cfgBytes *[]byte
+}
 
 // NewConfigTask initializes and returns a Task task.
-func NewConfigTask() phase.Task {
-	return &Task{}
+func NewConfigTask(cfgBytes *[]byte) phase.Task {
+	return &Task{
+		cfgBytes: cfgBytes,
+	}
 }
 
 // TaskFunc returns the runtime function.
@@ -30,30 +36,63 @@ func (task *Task) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 }
 
 func (task *Task) standard(r runtime.Runtime) (err error) {
+	cfg, err := config.NewFromFile(constants.ConfigPath)
+	if err != nil || !cfg.Persist() {
+		log.Printf("failed to read config from file or persistence disabled. re-pulling config")
+
+		var cfgBytes []byte
+
+		cfgBytes, err = fetchConfig(r)
+		if err != nil {
+			return err
+		}
+
+		*task.cfgBytes = cfgBytes
+
+		return nil
+	}
+
+	log.Printf("using existing config on disk")
+
+	cfgBytes, err := cfg.Bytes()
+	if err != nil {
+		return err
+	}
+
+	*task.cfgBytes = cfgBytes
+
+	return nil
+}
+
+func fetchConfig(r runtime.Runtime) (out []byte, err error) {
 	var b []byte
 
 	if b, err = r.Platform().Configuration(); err != nil {
-		return err
+		return nil, err
 	}
 
 	// Detect if config is a gzip archive and unzip it if so
 	contentType := http.DetectContentType(b)
 	if contentType == "application/x-gzip" {
-		gzipReader, err := gzip.NewReader(bytes.NewReader(b))
+		var gzipReader *gzip.Reader
+
+		gzipReader, err = gzip.NewReader(bytes.NewReader(b))
 		if err != nil {
-			return fmt.Errorf("error creating gzip reader: %w", err)
+			return nil, fmt.Errorf("error creating gzip reader: %w", err)
 		}
 
 		// nolint: errcheck
 		defer gzipReader.Close()
 
-		unzippedData, err := ioutil.ReadAll(gzipReader)
+		var unzippedData []byte
+
+		unzippedData, err = ioutil.ReadAll(gzipReader)
 		if err != nil {
-			return fmt.Errorf("error unzipping machine config: %w", err)
+			return nil, fmt.Errorf("error unzipping machine config: %w", err)
 		}
 
 		b = unzippedData
 	}
 
-	return ioutil.WriteFile(constants.ConfigPath, b, 0600)
+	return b, nil
 }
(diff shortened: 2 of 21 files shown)


smira:
I wonder if we want to umount `/boot` at some early point in boot to make sure it's "clean" on reboot? or mount/re-mount it as ro/rw as needed, basically my point is to make it either mounted only when we need it or make it rw only for the time we need to write config to minimize the risk of corruption, as it's FAT.

smira:
And if we want to test it out, we can make it (an option? default?) for `osctl cluster create` which will go to integration tests... we can even stop config HTTP service after some reboots (or after successful install?). This should be optional to keep upgrade tests happy (as they need to be able to provision older clusters)

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -135,6 +154,10 @@ func (d *Sequencer) Boot() error {
 			configtask.NewExtraFilesTask(),
 			configtask.NewSysctlsTask(),
 		),
+		phase.NewPhase(
```
rsmitty: Leaving this mounted as RO at the end here because other sequences like Upgrade want to read this file immediately when they start running.

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -152,17 +175,7 @@ func (d *Sequencer) Boot() error {
...
-		return err
-	}
-
-	config, err := config.New(content)
-	if err != nil {
-		return err
-	}
-
-	phaserunner, err := phase.NewRunner(config, runtime.Shutdown)
+	phaserunner, err := phase.NewRunner(nil, runtime.Shutdown)
```
rsmitty: There was no need for shutdown to read the config file from what I could tell.

smira: what if we need to access it in the shutdown sequence some time in the future? if we get to shutdown phase, we should have got through the boot phase successfully, so config should be in `/boot`?

internal/app/machined/internal/phase/config/config.go
```diff
@@ -30,30 +37,95 @@ func (task *Task) TaskFunc(mode runtime.Mode) phase.TaskFunc {
...
+
+			return nil
+		}
+
+		return err
+	}
+
+	log.Printf("found an existing config.yaml. determining if we should pull anyways")
+
+	content, err := config.FromFile(constants.ConfigPath)
```
smira: this might be _theoretical_, but still: what if config gets corrupted/it is unreadable for some reason, or due to some bug we got it saved in a wrong way?

before this PR, as config was fetched every time, this wasn't noticeable

after this PR:

* if persist is enabled, probably there's not much we can do (if original config source is gone)
* but if persist is not enabled, it means that node won't boot anymore, while technically there's no good reason to fail the boot

so proposal:

if config can't be read, treat it as non-fatal error and fetch config, so in other words unify case above then config file doesn't exist with case below when config file can't be read/parsed or config has !.Persist. 

Basically, try reading config, if that fails, log an error, and treat that as !Persist.

rsmitty: @smira reworked this to accomplish what you were talking about. if we have any issue either reading in the config from disk or parsing it, we'll force the re-fetch.

internal/app/machined/internal/phase/config/config.go
```diff
@@ -30,30 +37,101 @@ func (task *Task) TaskFunc(mode runtime.Mode) phase.TaskFunc {
...
+	forceCfgPull := false
+
+	content, err := config.FromFile(constants.ConfigPath)
+	if err != nil {
+		forceCfgPull = true
+
+		log.Printf("failed to read config, forcing a re-pull of the config: %s", err)
+	}
+
+	cfg, err := config.New(content)
```
smira: I think this should go under `else` for the `if err != nil` above, as `content` will be `nil` if we can't read the file

internal/app/machined/internal/phase/config/config.go
smira: thanks, this looks much better, but I feel we could try to unify it a bit more potentially simplifying a code.

if we had a function `readConfig() (runtime.Configurator, error)` which tries to read config, but returns an error if it's missing, we can call it, and if it returns an error or `!cfg.Persist()` try to fetch it. this would unify the code above (which checks if file exists and below which checks for readability). I think this will be less code, and we could have used this function in the sequencer as well as it seems to be doing same thing (reading config from file) over and over again?

internal/app/machined/internal/phase/config/config.go
rsmitty: yep, good spot. fixed in amended commit.

internal/app/machined/internal/phase/config/config.go
smira: so my point is that the case when config doesn't exist or when it's not readable is the same case

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
rsmitty: fixed this in amended commit

pkg/config/config.go
```diff
@@ -122,8 +122,8 @@ func NewConfigBundle(opts ...BundleOption) (*v1alpha1.ConfigBundle, error) {
 	return bundle, nil
 }
 
-// New initializes and returns a Configurator.
-func New(c Content) (config runtime.Configurator, err error) {
+// newConfig initializes and returns a Configurator.
```
rsmitty: went ahead and made this function, `fromFile`, and `fromBytes` private since I removed all references to them and replaced with the proper `NewConfigFrom*` function.

docs/website/content/v0.3/en/configuration/v1alpha1.md
```diff
@@ -72,6 +72,19 @@ Valid Values:
 
 - ``v1alpha1``
 
+#### persist
+
+Indicates whether to pull the config data upon every boot.
```
andrewrynhard: nit: We refer to this config as "machine config" in the rest of the docs.

pkg/config/config.go
andrewrynhard: nit: The package is already called "`config`". The extra "`Config`" in the `NewConfigFrom*` functions is redundant.

pkg/constants/constants.go
```diff
@@ -168,7 +168,7 @@ const (
 	EtcdDataPath = "/var/lib/etcd"
 
 	// ConfigPath is the path to the downloaded config.
-	ConfigPath = "/run/config.yaml"
+	ConfigPath = "/boot/config.yaml"
```
andrewrynhard: A note for my future self: I wonder if we should call this `machine.yaml` since we always refer to it as "machine config".

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -122,10 +124,26 @@ func (d *Sequencer) Boot() error {
...
 			rootfs.NewVarDirectoriesTask(),
 		),
+		phase.NewPhase(
+			"unmount /boot",
+			rootfs.NewUnmountSystemDisksTask(constants.BootPartitionLabel),
+		),
+		phase.NewPhase(
+			"mount /boot as rw",
+			rootfs.NewMountSystemDisksTask(constants.BootPartitionLabel),
+		),
```
andrewrynhard: Should we maybe mount as R/W to begin with earlier in boot? Then we won't have to do these.

rsmitty: Andrey's concern around that was that because it's a FAT file system there's a higher chance of corruption. So we should mount as rw only when we needed it.

rsmitty:
@andrewrynhard fixed your nits in amended commit.

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
smira: my concern was close to that, but basically not to leave it r/w longer than we need, as if we screw `/boot`, we screw whole node

PR #1920, opened 2020-03-04
feat: split routerd from apid

smira:
New service `routerd` performs exactly single task: based on incoming
API call service name, it routes the requests to the appropriate Talos
service (`networkd`, `osd`, etc.) Service `routerd` listens of file
socket and routes requests to file sockets.

Service `apid` now does single task as well:

* it either fans out request to other `apid` services running on other
nodes and aggregates responses
* or it forwards requests to local `routerd` (when request destination
is local node)

Cons:

* one more proxying layer on request path

Pros:

* more clear service roles
* `routerd` is part of core Talos, services should register with it to
expose their API; no auth in the service (not exposed to the world)
* `apid` might be replaced with other implementation, it depends on TLS infra,
auth, etc.
* `apid` is better segregated from other Talos services (can only access
`routerd`, can't talk to other Talos services directly, so less exposure
in case of a bug)

This change is no-op to the end users.


internal/app/routerd/main.go
```diff
@@ -0,0 +1,52 @@
...
+func main() {
+	log.SetFlags(log.Lshortfile | log.Ldate | log.Lmicroseconds | log.Ltime)
+
+	if err := startup.RandSeed(); err != nil {
+		log.Fatalf("failed to seed RNG: %v", err)
+	}
+
+	router := director.NewRouter()
+
+	// TODO: this should be dynamic based on plugin registration
```
andrewrynhard: ❤️

PR #1921, opened 2020-03-04
chore: cleanup assets dir after bootkube is done

rsmitty:
This PR will clean up bootkube assets regardless of whether bootkube
succeeds. This will allow for a failed bootkube deployment to retry on
reboot.

### DIFF
--- a/internal/app/machined/internal/bootkube/bootkube.go
+++ b/internal/app/machined/internal/bootkube/bootkube.go
@@ -7,6 +7,9 @@ package bootkube
 import (
 	"context"
 	"io"
+	"log"
+	"os"
+	"path/filepath"
 
 	"github.com/kubernetes-sigs/bootkube/pkg/bootkube"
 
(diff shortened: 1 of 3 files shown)


internal/app/machined/internal/bootkube/bootkube.go
```diff
@@ -45,5 +45,16 @@ func (s *Service) Main(ctx context.Context, config runtime.Configurator, logWrit
...
 	}
 
-	return bk.Run()
+	defer func() {
+		if err = os.RemoveAll(constants.AssetsDirectory); err != nil {
+			log.Printf("failed to cleanup bootkube assets dir %s", constants.AssetsDirectory)
+		}
+	}()
+
+	err = bk.Run()
```
smira: could still do `return bk.Run()` here

internal/app/machined/internal/bootkube/bootkube.go
```diff
@@ -45,5 +46,24 @@ func (s *Service) Main(ctx context.Context, config runtime.Configurator, logWrit
...
 	}
 
+	defer func() {
+		if err = os.RemoveAll(constants.AssetsDirectory); err != nil {
+			log.Printf("failed to cleanup bootkube assets dir %s", constants.AssetsDirectory)
+		}
+
+		bootstrapWildcard := filepath.Join(constants.ManifestsDirectory, "bootstrap-*")
+
+		bootstrapFiles, err := filepath.Glob(bootstrapWildcard)
```
andrewrynhard: Nice 👍

smira:
you can hate me, but that should be testable in integration test pretty easily: just verify that those dirs/files are gone with `ls` or `read` APIs

PR #1923, opened 2020-03-05
chore: improve handling of etcd responses in bootkube pre-func

smira:
Try more attempts, wait for the response. Treat empty response as no
error (as this is what to expect when key is not set yet).


smira:
Not really sure about this PR, it passes the tests, but this is so much edge case...

smira:
Also added debugging for failed integration test run:

1. Stop on the first failure to avoid messing logs with unrelated failures from follow-up tests.
2. Dump provisioner logs (crashdump).

PR #1924, opened 2020-03-06
chore: add service state to postfunc

rsmitty:
This PR will allow us to take conditional actions in the postfunc of our
services by passing the state of the service into the postfunc call. We
can use this to do conditional cleanups and finalizers if success.

### DIFF
--- a/internal/app/machined/pkg/system/service_runner.go
+++ b/internal/app/machined/pkg/system/service_runner.go
@@ -60,6 +60,14 @@ func NewServiceRunner(service Service, config runtime.Configurator) *ServiceRunn
 	}
 }
 
+// GetState implements events.Recorder
+func (svcrunner *ServiceRunner) GetState() events.ServiceState {
+	svcrunner.mu.Lock()
+	defer svcrunner.mu.Unlock()
+
+	return svcrunner.state
+}
+
 // UpdateState implements events.Recorder
 func (svcrunner *ServiceRunner) UpdateState(newstate events.ServiceState, message string, args ...interface{}) {
 	svcrunner.mu.Lock()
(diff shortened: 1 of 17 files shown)


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -146,7 +147,11 @@ func (b *Bootkube) PreFunc(ctx context.Context, config runtime.Configurator) (er
 }
 
 // PostFunc implements the Service interface.
-func (b *Bootkube) PostFunc(config runtime.Configurator) (err error) {
+func (b *Bootkube) PostFunc(config runtime.Configurator, state events.ServiceState) (err error) {
```
rsmitty: An example of us making use of this. Needed it for skipping the etcd update anyways.

internal/app/machined/pkg/system/service_runner.go
```diff
@@ -224,7 +224,8 @@ func (svcrunner *ServiceRunner) Start() {
 		svcrunner.UpdateState(events.StateFinished, "Service finished successfully")
 	}
 
-	if err := svcrunner.service.PostFunc(svcrunner.config); err != nil {
+	// PostFunc passes in the state so that we can take actions that depend on the outcome of the run
+	if err := svcrunner.service.PostFunc(svcrunner.config, svcrunner.state); err != nil {
```
smira: `svcrunner.state` is protected with `svcrunner.mu` mutex (see `UpdateState` as an example)

so I think we might need `.GetState()` function with proper locking, and we should be all good

internal/app/machined/pkg/system/service_runner.go
```diff
@@ -60,6 +60,15 @@ func NewServiceRunner(service Service, config runtime.Configurator) *ServiceRunn
 	}
 }
 
+// GetState implements events.Recorder
+func (svcrunner *ServiceRunner) GetState() events.ServiceState {
+	svcrunner.mu.Lock()
+	state := svcrunner.state
+	svcrunner.mu.Unlock()
+
+	return state
```
smira: nitpick:

```suggestion
	svcrunner.mu.Lock()
	defer svcrunner.mu.Unlock()

	return svcrunner.state
```

rsmitty: fixed in amended commit.

PR #1925, opened 2020-03-07
feat: Ensure /opt/cni/bin is mounted in kubelet.

salkin:
When using multi network CNI like Multus,Genie etc. kubelet needs to access CNI binaries to add additional network interfaces. This ensures the /opt/cni/bin folder is created before kubelet is started and any CNI provider have created the directory.

### DIFF
--- a/internal/app/machined/pkg/system/services/kubelet.go
+++ b/internal/app/machined/pkg/system/services/kubelet.go
@@ -98,6 +98,10 @@ func (k *Kubelet) PreFunc(ctx context.Context, config runtime.Configurator) erro
 		return err
 	}
 
+	if err := os.MkdirAll(cni.DefaultCNIDir, 0700); err != nil {
+		return err
+	}
+
 	if err := ioutil.WriteFile(constants.KubernetesCACert, config.Cluster().CA().Crt, 0500); err != nil {
 		return err
 	}
@@ -164,6 +168,7 @@ func (k *Kubelet) Runner(config runtime.Configurator) (runner.Runner, error) {
 		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/log/pods", Source: "/var/log/pods", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/opt/cni/bin", Source: "/opt/cni/bin", Options: []string{"rbind", "rshared", "rw"}},
 	}
 
 	// Add in the additional CNI mounts.


internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -164,6 +168,7 @@ func (k *Kubelet) Runner(config runtime.Configurator) (runner.Runner, error) {
 		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/log/pods", Source: "/var/log/pods", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/opt/cni/bin", Source: "/opt/cni/bin", Options: []string{"rbind", "rshared", "rw"}},
```
andrewrynhard: My only concern here is that hyperkube comes with CNI bundled in `/opt/cni/bin`, and I don't think there are strong guarantees that each CNI installs all the plugins it needs. The standard set of plugins might be assumed by flannel IIRC.

andrewrynhard: There is a solution today, however. In the `kubelet.extraMounts` section you can tell talos to mount `/opt/cni/bin` to the kubelet container. Something like:

```yaml
kubelet:
  extraMounts:
    - source: /opt/cni/bin
      destination: /opt/cni/bin
      type: bind
      options:
        - rshared
        - rw
```

salkin: I think I tried this but could not get that to work properly as the /opt/cni/bin folder is not existing. No idea though why it did not work

salkin: I think I tried this but could not get that to work properly as the /opt/cni/bin folder is not existing. No idea though why it did not work

salkin: Eq. I think the mount is not ensuring the directory is first created on host side

salkin: @andrewrynhard would it be better if we add some option to the extra mounts to create the directory if not exists then that option would work fine I think

andrewrynhard: Yeah I think so. Maybe even just attempt to create the directory if it doesn't exist? Only issue there is that the rootfs is mostly read only, so we won't be able to create the directory.

A simple fix is that we can create `/opt/cni/bin` by default in the rootfs. That is definitely in the 80% of use cases side of things. So we would need to create an empty overlay mount as well. Probably will take a little more thinking.

andrewrynhard: Maybe a mix of what you said (add an option to the extra mounts to create the directory if not exists), _and_ an optiton to create it as overlay.

salkin: So would something similar as for kubernetes volume mounts be the format for the extra mounts:

Like: 
```
kubelet:
  extraMounts:
    - source: /opt/cni/bin
         type: DirOrCreate
      destination: /opt/cni/bin
      type: bind
      options:
        - rshared
        - rw
```

andrewrynhard: 🤔 Sounds reasonable. Although, I don't think we can be quite as flexible because the creation of an arbitrary directory is not possible. The directory have to exist.

- `Directory`
- `DirectoryWithOverlay`

Might be the only types we could support. The former would be an existing directory that we want to mount into the kubelet for some reason. The latter would create an overlay mount on top of an existing directory, making it `rw`.

salkin:
Closing based on discussion.

maxpain:
How I can make Multus work with Talos?
I'm getting this error:

```
error adding container to network "macvlan": failed to find plugin "macvlan" in path [/opt/cni/bin /opt/cni/bin]
```

PR #1930, opened 2020-03-10
feat: respect panic kernel flag

rsmitty:
This PR allows Talos to respect the panic=0 flag if users pass that in
their kernel args. Doing this makes it easier to catch kernel panics in
debug scenarios and allows the user to manually trigger a restart with
ctrl+alt+del when they're ready.

### DIFF
--- a/internal/app/init/main.go
+++ b/internal/app/init/main.go
@@ -8,9 +8,14 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/app/init/main.go
```diff
@@ -63,6 +67,18 @@ func run() (err error) {
 }
 
 func recovery() {
+	// If panic is set in the kernel flags, we'll hang instead of rebooting.
+	// But we still allow users to hit CTRL+ALT+DEL to try and restart when they're ready.
```
andrewrynhard: Nice.

PR #1937, opened 2020-03-13
fix: respect dns domain from machine config

rsmitty:
BREAKING CHANGE: This PR fixes a bug where we were only passing `cluster.local` to the
kubelet configuration. It will also pull in a new version of the
bootkube fork to ensure that custom domains got propogated down to the
API Server certs, as well as the CoreDNS configuration for a cluster.

Existing users should be aware that, if they were previously trying to
use this option in machine configs, that an upgrade will may break
their cluster. It will update a kubelet flag with the new domain, but
CoreDNS and API Server certs will not change since bootkube has already
run. One option may be to change these values manually inside the
Kubernetes cluster. However, it may prove easier to rebuild the cluster
if necessary.

Additionally, this PR also exposes a flag to `osctl config generate`
to allow tweaking this domain value as well.

### DIFF
--- a/cmd/osctl/cmd/config.go
+++ b/cmd/osctl/cmd/config.go
@@ -26,6 +26,7 @@ import (
 
 var (
 	configVersion     string
+	dnsDomain         string
 	kubernetesVersion string
 	installDisk       string
 	installImage      string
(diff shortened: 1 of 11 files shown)


andrewrynhard:
Looks like `cluster.local` is hardcoded in few places in bootkube:

- https://github.com/talos-systems/bootkube/blob/talos/pkg/asset/tls.go#L133
- https://github.com/talos-systems/bootkube/blob/talos/pkg/asset/internal/templates.go#L894

I might be missing others.

PR #1939, opened 2020-03-17
fix: refactor the upgrade process

andrewrynhard:
This PR introduces a new strategy for upgrades. Instead of attempting to
zap the partition table, create a new one, and then format the
partitions, this change will only update the `vmlinuz`, and
`initramfs.xz` being used to boot. It introduces an A/B style upgrade
process, which will allow for easy rollbacks. One deviation from our
original intention with upgrades is that this change does not completely
reset a node. It falls just short of that and does not reset the
partition table. This forces us to keep the current partition scheme in
mind as we make changes in the future, because an upgrade assumes a
specific partition scheme. We can improve upgrades further in the
future, but this will at least make them more dependable. Finally, one
more feature in this PR is the ability to keep state. This enables
single node clusters to upgrade since we keep the etcd data around.


cmd/installer/pkg/installer.go
```diff
@@ -117,20 +204,48 @@ func (i *Installer) Install(sequence runtime.Sequence) (err error) {
...
-	if err = syslinux.Install(filepath.Join(constants.BootMountPoint), syslinuxcfg); err != nil {
-		return err
+		if err = syslinux.Install(filepath.Join(constants.BootMountPoint), i.Current, syslinuxcfg, sequence, i.bootPartitionFound); err != nil {
+			return err
+		}
+	} else {
+		i.cmdline.Append("initrd", filepath.Join("/", i.Next, constants.InitramfsAsset))
+
+		syslinuxcfg := &syslinux.Cfg{
+			Default: i.Next,
```
andrewrynhard: We should use `i.Current`

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -124,25 +123,13 @@ func (d *Sequencer) Boot() error {
...
 		),
 		phase.NewPhase(
-			"unmount /boot",
-			rootfs.NewUnmountSystemDisksTask(constants.BootPartitionLabel),
-		),
-		phase.NewPhase(
-			"mount /boot as ro",
-			rootfs.NewMountSystemDisksTask(constants.BootPartitionLabel, mount.WithReadOnly(true)),
+			"selinux",
+			rootfs.NewClearOnceTask(),
```
andrewrynhard: This should be moved to post startup tasks.

pkg/blockdevice/filesystem/vfat/filesystem.go
```diff
@@ -84,6 +85,8 @@ func (fs *FileSystem) Open(path string) (io.Reader, error) {
 		found := false
 
 		for _, entry := range entries {
+			log.Printf("entry.filename: %s", entry.filename)
```
andrewrynhard: Get rid of this.

internal/pkg/provision/providers/firecracker/crashdump.go
```diff
@@ -47,7 +47,7 @@ func (p *provisioner) CrashDump(ctx context.Context, cluster provision.Cluster,
 			continue
 		}
 
-		if err = tail.SeekLines(f, 1000); err != nil {
+		if err = tail.SeekLines(f, 15000); err != nil {
```
andrewrynhard: Will something like 5000 work?

andrewrynhard:
Where is the best place to document the high level way upgrades work?

cmd/installer/pkg/bootloader/syslinux/adv.go
```diff
@@ -0,0 +1,195 @@
...
+
+const (
+	// AdvEnd is the noop tag.
+	AdvEnd = iota
+	// AdvBootonce is the bootonce tag.
+	AdvBootonce
+	// AdvMenusave is the menusave tag.
+	AdvMenusave
+	// AdvUpgrade is the upgrade tag.
+	AdvUpgrade
```
andrewrynhard: We should add a buffer of maybe 2 or 3 "reserved" slots here, just in case syslinux decides to add moer flags in the furture.

cmd/installer/pkg/bootloader/syslinux/adv.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+// ADV represents the Syslinux Auxiliary Data Vector.
+type ADV []byte
+
+// NewADV returns the Auxiliary Data Vector.
+func NewADV(r io.ReadSeeker) (adv ADV, err error) {
+	_, err = r.Seek(-1024, 2)
```
andrewrynhard: `s/1024/2*AdvSize/g`

cmd/installer/pkg/bootloader/syslinux/adv_test.go
```diff
@@ -0,0 +1,299 @@
...
+import (
+	"bytes"
+	"io"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestNewADV(t *testing.T) {
+	f, err := os.Open("ldlinux.sys")
```
andrewrynhard: We should move this to the "testdata" directory.

internal/app/machined/internal/phase/rootfs/syslinux.go
```diff
@@ -0,0 +1,154 @@
...
+
+	var (
+		wg     sync.WaitGroup
+		result *multierror.Error
+		errCh  = make(chan error, len(svcs))
+	)
+
+	wg.Add(len(svcs))
+
+	for _, svc := range svcs {
```
andrewrynhard: We can use use the `conditions.Wait` instead of all this instead.

internal/app/machined/internal/sequencer/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -238,21 +220,14 @@ func (d *Sequencer) Upgrade(req *machineapi.UpgradeRequest) error {
...
+			rootfs.NewUnmountSystemDisksTask(constants.BootPartitionLabel),
 			rootfs.NewUnmountSystemDisksTask(constants.EphemeralPartitionLabel),
 		),
 		phase.NewPhase(
 			"unmount system disk bind mounts",
 			rootfs.NewUnmountSystemDiskBindMountsTask(devname),
 		),
-		phase.NewPhase(
-			"verify system disk not in use",
-			disk.NewVerifyDiskAvailabilityTask(devname),
```
andrewrynhard: Update this to run this test against the ephemeral partition (e.g. `/dev/sda2` _NOT_ `/dev/sda as we do today)

cmd/installer/pkg/bootloader/syslinux/adv.go
```diff
@@ -0,0 +1,201 @@
...
+		if tag != t {
+			i++
+			continue
+		}
+
+		b := make([]byte, a[i+1])
+
+		length := uint8(len(b))
+
+		a[i] = 0
```
smira: this doesn't seem to be correct if tag being deleted is not the last one, as zeroing out tag area is going to turn it into `AdvEnd`, so tags following it up won't be accessible

instead, we should move the area following the tag N bytes to the left (2 + tag value len)

cmd/installer/pkg/bootloader/syslinux/syslinux.go
```diff
@@ -110,21 +144,73 @@ func Install(base string, config interface{}) (err error) {
...
+			return fmt.Errorf("failed to set upgrade tag: %q", label)
+		}
+
+		if _, err = f.Write(adv); err != nil {
+			return err
+		}
+
+		log.Println("set upgrade tag in ADV")
+
+		return nil
```
smira: could be removed?

internal/app/machined/internal/phase/services/start_services.go
```diff
@@ -81,3 +85,20 @@ func (task *StartServices) loadKubernetesServices(r runtime.Runtime) {
 		)
 	}
 }
+
+func (task *StartServices) wait(r runtime.Runtime) (err error) {
+	svcs := system.Services(r.Config()).List()
```
smira: waiting for all the services to start might break some stuff a bit for those 5 minutes. if some service doesn't start for any reason, boot sequence will never finish, so `machined` will never get into the state when it accepts other sequences (reboot for example)

not a big deal I guess, but I thought it might be important

andrewrynhard: Yeah, but unfortunately this is the only surefire way to know that the upgrade was successful. It is a tradeoff :/

cmd/installer/pkg/bootloader/syslinux/adv.go
andrewrynhard: Good catch.

internal/app/machined/internal/phase/services/start_services.go
smira: other option I thought about is not that clean and breaks out of sequencer, but basically spin up goroutine here, wait for services to be healthy, and only after that do upgrade cleanup

andrewrynhard: Yeah that is a possibility. I feel like we should stick with the sequencer paradigm as close as possible, and in the next version of the sequencer we can address the limitations in the current implementation, then fix this (and a number of other things). And, actually, today the `post startup tasks` continuously tries to mark control plane nodes with the master label, effectively blocking APIs, even today without this change. So nothing really changes from how things work today.

PR #1941, opened 2020-03-17
fix: ensure printing of panic message

rsmitty:
This PR reworks the ordering of our recovery function. It will make sure
we actually show the user the recovery message prior to looking into
whether to auto-reboot.


internal/app/init/main.go
```diff
@@ -71,19 +71,20 @@ func recovery() {
...
-			signal.Notify(exitSignal, syscall.SIGINT, syscall.SIGTERM)
-			<-exitSignal
-		}
-	}
-
 	if r := recover(); r != nil {
 		log.Printf("recovered from: %+v\n", r)
 
+		p := procfs.ProcCmdline().Get(constants.KernelParamPanic).First()
+		if p != nil {
```
smira: nitpick, we could do `if p != nil && *p == "0" {`

rsmitty: yep, good call. fixed in amended commit.

PR #1945, opened 2020-03-18
feat: Add support for VLAN devices on Talos

salkin:
- Creates VLAN devices that are configured under a certain master device
- Address and routes are added to VLAN devices
- Allows a physical device to be created without Addressing if VLANs are defined for a network interface.

### DIFF
--- a/internal/app/networkd/pkg/address/static.go
+++ b/internal/app/networkd/pkg/address/static.go
@@ -16,8 +16,10 @@ import (
 
 // Static implements the Addressing interface
 type Static struct {
-	Device      *machine.Device
+	CIDR        string
+	Mtu         int
 	FQDN        string
+	RouteList   []machine.Route
 	NetIf       *net.Interface
 	NameServers []net.IP
 }
@@ -37,7 +39,7 @@ func (s *Static) Name() string {
 // Address returns the IP address
 func (s *Static) Address() *net.IPNet {
 	// nolint: errcheck
-	ip, ipn, _ := net.ParseCIDR(s.Device.CIDR)
+	ip, ipn, _ := net.ParseCIDR(s.CIDR)
 	ipn.IP = ip
 
 	return ipn
@@ -46,13 +48,13 @@ func (s *Static) Address() *net.IPNet {
 // Mask returns the netmask.
 func (s *Static) Mask() net.IPMask {
 	// nolint: errcheck
-	_, ipnet, _ := net.ParseCIDR(s.Device.CIDR)
+	_, ipnet, _ := net.ParseCIDR(s.CIDR)
 	return ipnet.Mask
 }
 
 // MTU returns the specified MTU.
 func (s *Static) MTU() uint32 {
-	mtu := uint32(s.Device.MTU)
+	mtu := uint32(s.Mtu)
 	if mtu == 0 {
 		mtu = uint32(s.NetIf.MTU)
 	}
@@ -83,7 +85,7 @@ func (s *Static) Scope() uint8 {
 // Routes aggregates the specified routes for a given device configuration
 // TODO: do we need to be explicit on route vs gateway?
 func (s *Static) Routes() (routes []*Route) {
-	for _, route := range s.Device.Routes {
+	for _, route := range s.RouteList {
 		// nolint: errcheck
 		_, ipnet, _ := net.ParseCIDR(route.Network)
 
--- a/internal/app/networkd/pkg/networkd/netconf.go
+++ b/internal/app/networkd/pkg/networkd/netconf.go
@@ -6,6 +6,7 @@ package networkd
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"strings"
 
@@ -31,7 +32,7 @@ func buildOptions(device machine.Device, hostname string) (name string, opts []n
 	// Configure Addressing
 	switch {
 	case device.CIDR != "":
-		s := &address.Static{Device: &device}
+		s := &address.Static{CIDR: device.CIDR, RouteList: device.Routes, Mtu: device.MTU}
 
 		// Set a default for the hostname to ensure we always have a valid
 		// ip + hostname pair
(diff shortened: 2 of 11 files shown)


andrewrynhard:
Thanks @salkin! This looks good to me, but we will need to address the linting error, and make sure this is squashed down to 1 commit before we can merge.

internal/app/networkd/pkg/networkd/netconf.go
```diff
@@ -43,11 +44,31 @@ func buildOptions(device machine.Device, hostname string) (name string, opts []n
 		}
 
 		opts = append(opts, nic.WithAddressing(s))
+	case device.DHCP == false:
+		//Allow master interface without any addressing if VLANs exist
+		if device.CIDR == "" && len(device.Vlans) > 0 {
```
smira: looks like this check is not needed, as `device.CIDR == ""` is processed in the `case` above

internal/app/networkd/pkg/networkd/netconf.go
```diff
@@ -43,11 +44,31 @@ func buildOptions(device machine.Device, hostname string) (name string, opts []n
 		}
 
 		opts = append(opts, nic.WithAddressing(s))
+	case device.DHCP == false:
```
smira: I feel we can simplify this switch by structuring it as:

```
switch{
case CIDR != "":
   // assign static address
case DHCP:
   // use dhcp
default:
  // use dhcp is no vlans, otherwise keep without an address
}
```

internal/app/networkd/pkg/nic/netlink.go
```diff
@@ -85,6 +96,37 @@ func (n *NetworkInterface) configureBond(idx int, attrs *netlink.AttributeEncode
...
+		Change: 0,
+		Attributes: &rtnetlink.LinkAttributes{
+			Info: &rtnetlink.LinkInfo{
+				// https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/if_link.h#L612
+				Kind: "vlan",
+				Data: nlAttrBytes,
+			},
+		},
+	})
+	if err != nil {
```
smira: this could be simplified to `return n.rtConn.Link.Set(...)`

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -196,6 +254,26 @@ func (n *NetworkInterface) Addressing() error {
...
+// This is inclusive of the address itself as well as any defined routes.
+func (n *NetworkInterface) AddressingSub() error {
+	if n.IsIgnored() {
+		return nil
+	}
+
+	for _, vlan := range n.Vlans {
+		for _, method := range vlan.AddressMethod {
+			if err := n.configureInterface(method, vlan.Link); err != nil {
+				// Treat as non fatal error when failing to configure an interface
```
smira: should we still log the errors for debugging?

internal/app/networkd/pkg/address/static.go
```diff
@@ -9,15 +9,16 @@ import (
 	"net"
 	"time"
 
-	"golang.org/x/sys/unix"
-
 	"github.com/talos-systems/talos/pkg/config/machine"
+	"golang.org/x/sys/unix"
```
smira: our convention is to keep talos imports as a group

andrewrynhard: FYI. `make fmt` should fix this

internal/app/networkd/pkg/networkd/netconf.go
salkin: Agree, refactored to this approach

internal/app/networkd/pkg/networkd/netconf.go
salkin: True, it came there by mistake during debugging why DHCP was still coming after network reset before I realized nic.go have a force DHCP if addressing methods missing.

internal/app/networkd/pkg/nic/netlink.go
salkin: Function removed as it is not used

andrewrynhard:
@salkin looksnlike the longer is failing. See https://ci.dev.talos-systems.io/talos-systems/talos/7825/1/14

internal/app/networkd/pkg/nic/nic.go
salkin: Added a log for this

andrewrynhard:
@salkin you want to run `make lint-go` locally? Looks like you got hit by the linter again: https://ci.dev.talos-systems.io/talos-systems/talos/7826

salkin:
I fixed the lint problems but it still fails, and I have no idea what it is about. It say nic.go is not formated but if I run "make fmt" no change happens to the file.

@andrewrynhard any help?

andrewrynhard:
@salkin i will checkout the pr locally in the morning and take a look.

andrewrynhard:
@salkin looks like formatting doesn't fix comments:
```
$ gofumports -d -local github.com/talos-systems/talos .
diff -u internal/app/networkd/pkg/nic/nic.go.orig internal/app/networkd/pkg/nic/nic.go
--- internal/app/networkd/pkg/nic/nic.go.orig	2020-03-23 08:45:53.000000000 -0700
+++ internal/app/networkd/pkg/nic/nic.go	2020-03-23 08:45:53.000000000 -0700
@@ -75,7 +75,7 @@
 	}

 	// If no addressing methods have been configured, default to DHCP.
-	//If VLANs exist do not force DHCP on master device
+	// If VLANs exist do not force DHCP on master device
 	if len(iface.AddressMethod) == 0 && len(iface.Vlans) == 0 {
 		iface.AddressMethod = append(iface.AddressMethod, &address.DHCP{})
 	}
```

andrewrynhard:
@salkin one more thing :) ... sorry

The commit should start will `feat:` instead of `Feat:`. The way we generate our CHANGELOG depends on the format..

salkin:
thanks @andrewrynhard  now it looks better!

andrewrynhard:
@salkin no problem. Just unit-tests failing now, but you should be good to get past all the linting.

internal/app/networkd/pkg/address/static.go
```diff
@@ -16,8 +16,10 @@ import (
 
 // Static implements the Addressing interface
 type Static struct {
-	Device      *machine.Device
+	CIDR        string
+	Mtu         int
```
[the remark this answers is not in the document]
salkin: This is done as when you have VLANs under the machine.Device that has their own parameters we cannot use the machine.Device to set the CIDR. 

This could probably be refactored to something else but the base concept of the machine.Device need to change in order to support the idea of VLANs as subdevices.

PR #1959, opened 2020-03-21
feat: build talosctl for ARM64

andrewrynhard:
This adds an ARM64 build of `talosctl`.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -3458,6 +3458,7 @@ steps:
     - _out/installer.tar
     - _out/talosctl-darwin-amd64
     - _out/talosctl-linux-amd64
+    - _out/talosctl-linux-arm64
     - _out/vmware.ova
     - _out/vmlinux
     - _out/vmlinuz
@@ -3572,6 +3573,6 @@ depends_on:
 
 ---
 kind: signature
-hmac: de88d1e74755b71883afb67e12dea17205b75e9ec462837f035257292ffbe805
+hmac: 16c7e32ef2ca8687136040290ea9f3e1b74c54d3ec103c908fc1cea2d38858c0
 
 ...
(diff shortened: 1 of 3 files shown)


dave93cab:
Hi Andrew,

I managed to build from your repo. Turns out this box is ARMv7 but was trivial to get it to build for that too. Let me know if you still think this is reasonable for inclusion and if you need a PR. I'm not too hot on Dockerfile/Makefiles. Is there a way of looping over a dictionary of arches and GO parameters to remove the redundancy? I can see how you would do it inside the FROM base section but not sure how you would have something that works inside as well as outside that for the COPY part.

```
FROM base AS talosctl-linux-armv7-build
ARG SHA
ARG TAG
ARG ARTIFACTS
ARG MGMT_HELPERS_PKG="github.com/talos-systems/talos/cmd/talosctl/pkg/mgmt/helpers"
WORKDIR /src/cmd/talosctl
RUN --mount=type=cache,target=/.cache/go-build GOOS=linux GOARCH=arm GOARM=7 go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Client -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${MGMT_HELPERS_PKG}.ArtifactsPath=${ARTIFACTS}" -o /talosctl-linux-armv7
RUN chmod +x /talosctl-linux-armv7

FROM scratch AS talosctl-linux
COPY --from=talosctl-linux-amd64-build /talosctl-linux-amd64 /talosctl-linux-amd64
COPY --from=talosctl-linux-arm64-build /talosctl-linux-arm64 /talosctl-linux-arm64
COPY --from=talosctl-linux-armv7-build /talosctl-linux-armv7 /talosctl-linux-armv7
```

andrewrynhard:
@dave93cab I'm not too concerned just yet about repeating docker statements for each arch. I can throw something up for v7. Will piing you in the PR.

PR #1962, opened 2020-03-23
fix: update rtnetlink checks for bit masks

smira:
According to `rtnetlink(7)` field `state` is a bitmask of states, so
update the check to test accordingly.

### DIFF
--- a/internal/app/networkd/pkg/reg/reg.go
+++ b/internal/app/networkd/pkg/reg/reg.go
@@ -193,12 +193,12 @@ func (r *Registrator) Check(ctx context.Context, in *empty.Empty) (reply *health
 		}
 
 		// Verify neighbor state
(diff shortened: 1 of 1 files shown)


smira:
```

              ndm_state is a bit mask of the following states:

              NUD_INCOMPLETE   a currently resolving cache entry
              NUD_REACHABLE    a confirmed working cache entry
              NUD_STALE        an expired cache entry

              NUD_DELAY        an entry waiting for a timer
              NUD_PROBE        a cache entry that is currently reprobed
              NUD_FAILED       an invalid cache entry

              NUD_NOARP        a device with no destination cache
              NUD_PERMANENT    a static entry
```

(http://man7.org/linux/man-pages/man7/rtnetlink.7.html)

PR #1963, opened 2020-03-23
feat: upgrade kubernetes to 1.18

rsmitty:
This PR will pull in the latest release of k8s 1.18 so we can start
validating it through our test suite.


andrewrynhard:
We will have to update the go modules too.

PR #1964, opened 2020-03-23
docs: add bare-metal install example yaml

dave93cab:
I know there is an example at the top of the page but the example mid-page was empty yaml

### DIFF
--- a/pkg/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/config/types/v1alpha1/v1alpha1_types.go
@@ -134,12 +134,13 @@ type MachineConfig struct {
 	//   examples:
 	//     - |
 	//       install:
-	//         disk:
+	//         disk: /dev/sda
 	//         extraKernelArgs:
-	//         image:
-	//         bootloader:
-	//         wipe:
-	//         force:
+	//           - option=value
+	//         image: docker.io/autonomy/installer:latest
+	//         bootloader: true
+	//         wipe: false
+	//         force: false
 	MachineInstall *InstallConfig `yaml:"install,omitempty"`
 	//   description: |
 	//     Allows the addition of user specified files.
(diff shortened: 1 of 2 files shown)


andrewrynhard:
Thanks @dave93cab! A few things to note so we can get this in.

1. There PR must have only 1 commit (a squash is needed)
2. The commit must be in conventional commit format: https://www.conventionalcommits.org/en/v1.0.0/. So in this case something like `docs: add bare-metal install example yaml` would work.
3. The edits you made are actually generated from go code. I think you might actually want to edit https://github.com/talos-systems/talos/blob/master/pkg/config/types/v1alpha1/v1alpha1_types.go#L132-L142 instead, and then run `make docs`.

Hope that is clear, and helpful. Feel free to ping me if you have any questions.

dave93cab:
That only update 0.4 docs

andrewrynhard:
@dave93cab Yeah, I can make a similar PR for v0.3 (the `release-0.3`) docs. We're still working on our tooling to simplify this process, so unfortunately I will have to make a PR against that branch as well.

pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -134,12 +134,13 @@ type MachineConfig struct {
...
 	//     - |
 	//       install:
-	//         disk:
+	//         disk: /dev/sda
 	//         extraKernelArgs:
-	//         image:
-	//         bootloader:
-	//         wipe:
-	//         force:
+        //           - option=value
```
andrewrynhard: looks like an extra indent here that makes `gofmt` unhappy.

andrewrynhard:
@dave93cab looks like I missed something too ... the commit needs a DCO. You can amend the commit with the `-s` flag to include the DCO. `git commit -s --amend --no-edit`.

dave93cab:
@andrewrynhard hopefully that is it. Thanks for your patience

PR #1974, opened 2020-03-24
chore: update upgrade tests for new version, split into two tracks

smira:
This updates upgrade tests to run two flows with 3+1 clusters:

1. 0.3 -> current (testing upgrade with partition wiping)
2. 0.4-alpha.7 -> current (testing upgrade without partition wiping,
boot-a/boot-b)

And small upgrade with preserve enabled for single-node cluster.

Provision tests are now split into two parallel tracks in Drone.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -496,11 +496,11 @@ steps:
   - kernel
   - push-local
 
-- name: provision-tests
+- name: provision-tests-prepare
   pull: always
   image: autonomy/build-container:latest
   commands:
-  - make provision-tests
+  - make provision-tests-prepare
   environment:
     REGISTRY: registry.ci.svc:5000
   privileged: true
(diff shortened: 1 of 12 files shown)


andrewrynhard:
Would it be possible to have the tracks names be something like `upgrade-v0.3-to-v0.4` and `upgrade-v0.4-to-latest`? Would be helpful to have the name be descriptive for new contributors, and ourselves as the tracks change over time.

smira:
we can certainly name them anything, but there are 3 tests today split over two tracks, if we add more tests in the future, is there a way to keep consistent naming? I don't think so. 

I went with generic numbers, so we can easily have two track or three tracks or anything else depending on our resources/desired parallelism

smira:
track just allows us to run tests in parallel, test names are still descriptive as they were before

PR #1977, opened 2020-03-26
fix: make upgrades work with UEFI

andrewrynhard:
Since the `--once` option of `extlinux` seems to only work with BIOS, we
needed to change to remove any reliance on this option. Instead of
booting the upgraded version once, and then making it the default after
a successful boot, we now make it the default, and then revert on any
boot error.

### DIFF
--- a/cmd/installer/pkg/bootloader/syslinux/syslinux.go
+++ b/cmd/installer/pkg/bootloader/syslinux/syslinux.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"os"
@@ -23,7 +24,9 @@ import (
 )
 
 const syslinuxCfgTpl = `DEFAULT {{ .Default }}
-  SAY Talos
+PROMPT 1
+TIMEOUT 50
+
 {{- range .Labels }}
 INCLUDE /{{ .Root }}/include.cfg
 {{- end }}`
@@ -33,12 +36,6 @@ const syslinuxLabelTpl = `LABEL {{ .Root }}
   INITRD {{ .Initrd }}
   APPEND {{ .Append }}
 `
-
-const (
-	gptmbrbin   = "/usr/lib/syslinux/gptmbr.bin"
-	syslinuxefi = "/usr/lib/syslinux/syslinux.efi"
-	ldlinuxe64  = "/usr/lib/syslinux/ldlinux.e64"
-)
 
 // Cfg reprsents the syslinux.cfg file.
 type Cfg struct {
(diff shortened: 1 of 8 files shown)


cmd/installer/pkg/bootloader/syslinux/syslinux.go
```diff
@@ -23,7 +23,9 @@ import (
 )
 
 const syslinuxCfgTpl = `DEFAULT {{ .Default }}
-  SAY Talos
+PROMPT 1
+TIMEOUT 50
```
andrewrynhard: Units if 1/10 sec. So this is a 5 second wait.

cmd/installer/pkg/bootloader/syslinux/syslinux.go
```diff
@@ -35,6 +37,18 @@ const syslinuxLabelTpl = `LABEL {{ .Root }}
...
+
+	// BootB is a syslinux label.
+	BootB = "boot-b"
+
+	// SyslinuxLdlinux is the path to ldlinux.sys.
+	SyslinuxLdlinux = constants.BootMountPoint + "/syslinux/ldlinux.sys"
+
+	// SyslinuxConfig is the path to the Syslinux config.
+	SyslinuxConfig = constants.BootMountPoint + "/syslinux/syslinux.cfg"
+
```
andrewrynhard: As we start to modularize, it makes sense to have the syslinux specific constants live in this package.

Ulexus: How about a `constants.go` file?  I don't have a strong opinion of pulling out constants to a separate file, but since we _currently_ pull them to a separate _package_, I wonder if we couldn't retain the value in that way.

andrewrynhard: 👍 works for me.

cmd/installer/pkg/bootloader/syslinux/syslinux.go
```diff
@@ -253,3 +214,82 @@ func WriteSyslinuxCfg(base, path string, syslinuxcfg *Cfg) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(efiDir+"/BOOTX64.EFI", input, 0600)
+	if err != nil {
+		return err
+	}
+
+	input, err = ioutil.ReadFile(ldlinuxe64)
```
Ulexus: nit:  maybe use `io.Copy()` to avoid reading whole files into RAM?

andrewrynhard: Yeah I opted for simplicity here since these files are kbs, but it still bugs me that we read the whole thing into memory. I'll clean up.

andrewrynhard: Fixed.

PR #1980, opened 2020-03-27
feat: move bootkube out as full service

smira:
This moves `bootkube start` into container which runs pretty much like
any other Talos service.

This resolves issues with bootkube log and provides better isolation.


internal/app/bootkube/assets.go
```diff
@@ -0,0 +1,311 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smira: contents of this file were just literally moved from the service file, no changes

PR #1981, opened 2020-03-27
test: mark long tests as !short

smira:
This skips long-running integration tests if `-test.short` mode is
enabled.


smira:
when short mode is enabled, all the tests run in 10s:

```
real	0m9,992s
user	0m4,103s
sys	0m0,594s
```

This doesn't enable short tests in Drone yet.

PR #1984, opened 2020-03-28
refactor: improve upgrade image validation logging

andrewrynhard:
This prevents the validation test from printing to stdout. An
additional log message was added to the beginning of the upgrade API
method to indicate that the image wil be validated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ KUBECTL_URL ?= https://storage.googleapis.com/kubernetes-release/release/v1.18.0
 SONOBUOY_VERSION ?= 0.17.1
 SONOBUOY_URL ?= https://github.com/heptio/sonobuoy/releases/download/v$(SONOBUOY_VERSION)/sonobuoy_$(SONOBUOY_VERSION)_$(OPERATING_SYSTEM)_amd64.tar.gz
 TESTPKGS ?= ./...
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ KUBECTL_URL ?= https://storage.googleapis.com/kubernetes-release/release/v1.18.0
 SONOBUOY_VERSION ?= 0.17.1
 SONOBUOY_URL ?= https://github.com/heptio/sonobuoy/releases/download/v$(SONOBUOY_VERSION)/sonobuoy_$(SONOBUOY_VERSION)_$(OPERATING_SYSTEM)_amd64.tar.gz
 TESTPKGS ?= ./...
-RELEASES ?= v0.3.2 v0.4.0-alpha.7
+RELEASES ?= v0.3.2 v0.4.0-alpha.8-9-g3b189c8c
```
andrewrynhard: Snuck this in so that we can use it to debug flakes in CI. It has an improved log when unmount errors occur.

smira: this should be a release tag, as it tries to download the assets from GitHub here

smira: it's fine to use `v0.4.0-alpha.8` here and `v0.4.0-alpha.8-9-g3b189c8c` down in the test; initramfs & kernel are only used for the first boot, after that installer installs correct initramfs/kernel and next reboot should have correct version of assets

PR #1990, opened 2020-03-30
Generate host configs

dave93cab:
@andrewrynhard not expecting this to be merged as-is but wondering your thoughts on something similar

usage: talos config gen --host-configs MachineConfigs.yaml

will output the init/controlplane/join as well as host-specific files. Left the files in there for an example.


dave93cab:
There's actually no need to use or modify the bundle, that's a remnant of the few other ways I tried to do it before settling on this, I will update tomorrow

andrewrynhard:
Hey @dave93cab ... If I'm understanding this PR, its an attempt to merge a set of base configs with an overlay to produce configs that are a little more specialized?

dave93cab:
@andrewrynhard  correct. examples were a bit contrived when in reality it would 90% of the time be a list of nodes and network configs

andrewrynhard:
@dave93cab Are you free on Thursday at 16:00 UTC? We have a zoom meeting that I'd like to use to dig into this some more. I have wanted to solve this problem for a while now, and the team has chatted about similar ideas in the past. Would be great if we could all collaborate on a design here.

dave93cab:
@andrewrynhard potentially, yeah. I'm working from home just depends on if I am in a call or something

andrewrynhard:
@dave93cab With internal discussions, and some research, I think we will have to wait for `v1alpha2` config (some time in august) because we are looking at changing the format to be compatible with tooling that can handle what this PR does. Appreciate the time you spent on this, but I will be closing. If you want to help us with `v1alpha2`, please join our community meetings. Thanks!

PR #1991, opened 2020-03-31
feat: automatically uncordon node after upgrade

andrewrynhard:
This adds a task that will automatically uncordon a node if we detect
that the node is recovering from an upgrade.

### DIFF
--- a/internal/app/machined/internal/phase/kubernetes/uncordon.go
+++ b/internal/app/machined/internal/phase/kubernetes/uncordon.go
@@ -1,0 +1,71 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package kubernetes
+
+import (
+	"os"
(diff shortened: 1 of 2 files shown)


internal/app/machined/internal/phase/kubernetes/uncordon.go
```diff
@@ -0,0 +1,68 @@
...
+
+// TaskFunc returns the runtime function.
+func (task *Uncordon) TaskFunc(mode runtime.Mode) phase.TaskFunc {
+	return func(r runtime.Runtime) error {
+		return task.standard()
+	}
+}
+
+func (task *Uncordon) standard() (err error) {
+	f, err := os.OpenFile(syslinux.SyslinuxLdlinux, os.O_RDWR, 0700)
```
andrewrynhard: We should open read only.

andrewrynhard:
I wonder if we should wait for the kubelet to report `READY`? We do this in the upgrade controller, and we can rip that code from there.

andrewrynhard:
We think that we need a more robust set of health checks before we can safely uncordon. Closing.

PR #1993, opened 2020-04-01
feat: introduce ability to specify extra hosts in /etc/hosts

rsmitty:
This PR will allow users to configure /etc/hosts through the network
config section, as opposed to having to use a file append operation.

Example usage might look something like:

```
...
...
machine:
  ...
  ...
  network:
    extraHostEntries:
      - ip: 192.168.1.100
        aliases:
          - test
          - test.wtf.bbq
...
...
```

### DIFF
--- a/pkg/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/config/types/v1alpha1/v1alpha1_types.go
@@ -430,6 +430,16 @@ type NetworkConfig struct {
 	//     Used to statically set the nameservers for the host.
 	//     Defaults to `1.1.1.1` and `8.8.8.8`
 	NameServers []string `yaml:"nameservers,omitempty"`
+	//   description: |
+	//     Allows for extra entries to be added to /etc/hosts file
+	//   examples:
+	//     - |
+	//       extraHostEntries:
+	//         - ip: 192.168.1.100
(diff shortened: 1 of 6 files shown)


pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -430,6 +430,16 @@ type NetworkConfig struct {
...
 	NameServers []string `yaml:"nameservers,omitempty"`
+	//   description: |
+	//     Allows for extra entries to be added to /etc/hosts file
+	//   examples:
+	//     - |
+	//       extraHostEntries:
+	//         - ip: 192.168.1.100
+	//           aliases:
+	//             - test
+	//             - test.wtf.bbq
```
andrewrynhard, line 441: We probably shouldn't have `test.wtf.bbq` in here. Silly, I know, but I think we should keep examples free of anything that could be offensive, and use very neutral and generic things like `example.com`, `example`, `domain.tld`, `domain`, etc.

rsmitty: fixed

andrewrynhard: 👍 Thank you.

PR #1997, opened 2020-04-03
chore: update timeout values for e2e tests

rsmitty:
This PR will update the values for timeout when testing e2e. We were
hitting issues in GCP on the reboot test, as the nodes seemed to be
taking a few minutes to become responsive again. I also moved the
"cluster health" check in the node-by-node reboot test to use the
default suite context, so it'll have a timeout of 30m instead of the 5
that it had initially. This seems to solve the node-by-node bailing as
well.

### DIFF
--- a/internal/integration/api/reboot.go
+++ b/internal/integration/api/reboot.go
@@ -59,7 +59,7 @@ func (suite *RebootSuite) TestRebootNodeByNode() {
 
 		func(node string) {
 			// timeout for single node reboot
-			ctx, ctxCancel := context.WithTimeout(suite.ctx, 5*time.Minute)
+			ctx, ctxCancel := context.WithTimeout(suite.ctx, 10*time.Minute)
 			defer ctxCancel()
 
 			nodeCtx := client.WithNodes(ctx, node)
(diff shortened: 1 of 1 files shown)


internal/integration/api/reboot.go
```diff
@@ -72,7 +72,7 @@ func (suite *RebootSuite) TestRebootNodeByNode() {
 
 			var uptimeAfter float64
 
-			suite.Require().NoError(retry.Constant(3 * time.Minute).Retry(func() error {
+			suite.Require().NoError(retry.Constant(10 * time.Minute).Retry(func() error {
```
smira, line 75: `ctx` timeout above should be bumped to be be at least 10 minutes to make this reasonable, otherwise `ctx`/`nodeCtx` is going be canceled before retry stops retrying

internal/integration/api/reboot.go
```diff
@@ -90,7 +90,7 @@ func (suite *RebootSuite) TestRebootNodeByNode() {
 			if suite.Cluster != nil {
 				// without cluster state we can't do deep checks, but basic reboot test still works
 				// NB: using `ctx` here to have client talking to init node by default
-				suite.AssertClusterHealthy(ctx)
+				suite.AssertClusterHealthy(suite.ctx)
```
smira, line 93: this drops health assertion out of `ctx` timeout, but probably that's fine

rsmitty: moved back to ctx. with the increased timeout it should be fine anyways.

internal/integration/api/reboot.go
rsmitty: bumped ctx to 10 mins.

PR #2002, opened 2020-04-07
fix: fixes proxy for bootkube download of additional assets.

salkin:

### DIFF
--- a/internal/app/bootkube/main.go
+++ b/internal/app/bootkube/main.go
@@ -7,11 +7,16 @@ package main
 import (
 	"flag"
 	"log"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/kubernetes-sigs/bootkube/pkg/bootkube"
 	"github.com/kubernetes-sigs/bootkube/pkg/util"
+	"golang.org/x/net/http/httpproxy"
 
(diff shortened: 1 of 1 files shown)


internal/app/bootkube/main.go
```diff
@@ -80,3 +85,28 @@ func main() {
...
 }
+
+func init() {
+	// Explicitly set the default http client transport
+	// to work around our fun proxy.Do once bug.
+	// This is the http.DefaultTransport with the Proxy
+	// func overridden so that the environment variables
+	// with be reread/initialized each time the http call
+	// is made.
+	http.DefaultClient.Transport = &http.Transport{
```
smira, line 96: this shouldn't be required afaik, as `bootkube` runs a service and environment variables don't change over time

salkin: Ok, maybe I need to add some debug logging then as it did not work for me without this.

smira: @salkin what was the error? 

In recent versions, `bootkube` logs are properly handled, so you can pull them with `talosctl logs bootkube`.

Also yesterday another fix went in: #2001, if you are trying to download something from https://

salkin: ok then it was probably the https issue as yes, trying from https.

PR #2008, opened 2020-04-08
chore: add PR template for contributors

rsmitty:
This PR will add the ability for us to show new contributors some
protips when they open a new PR against Talos.


rsmitty:
@smira yes! thanks. updated.

PR #2010, opened 2020-04-09
feat: add extra headers to fetch of extraManifests

salkin:
Provides capability to add extra headers in cases where files can only be fetched with token based authenction.

### DIFF
--- a/pkg/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/config/types/v1alpha1/v1alpha1_types.go
@@ -343,6 +343,14 @@ type ClusterConfig struct {
 	//         - "https://www.mysweethttpserver.com/manifest2.yaml"
 	ExtraManifests []string `yaml:"extraManifests,omitempty"`
 	//   description: |
+	//     A map of key value pairs that will be added while fetching the ExtraManifests.
+	//   examples:
+	//     - |
+	//       extraManifestHeaders:
+	//         Token: "1234567"
+	//         X-ExtraInfo: info
+	ExtraManifestHeaders map[string]string `yaml:"extraManifestHeaders,omitempty"`
+	//   description: |
 	//     Settings for admin kubeconfig generation.
 	//     Certificate lifetime can be configured.
 	//   examples:
--- a/docs/website/content/v0.4/en/configuration/v1alpha1.md
+++ b/docs/website/content/v0.4/en/configuration/v1alpha1.md
@@ -597,6 +597,21 @@ extraManifests:
 
 ```
 
+#### extraManifestHeaders
+
+A map of key value pairs that will be added while fetching the ExtraManifests.
+
+Type: `map`
+
+Examples:
+
+```yaml
+extraManifestHeaders:
+  Token: "1234567"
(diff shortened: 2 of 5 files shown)


pkg/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -343,6 +343,14 @@ type ClusterConfig struct {
 	//         - "https://www.mysweethttpserver.com/manifest2.yaml"
 	ExtraManifests []string `yaml:"extraManifests,omitempty"`
 	//   description: |
+	//     A list of key:value pairs that will be added while fetching the ExtraManifests.
```
rsmitty, line 346: Would it make more sense to just have `Headers` be an optional part of a larger ExtraManifests struct? It seems likely that I would want headers for some, but not necessarily all, of the manifests I wish to deploy. I'm thinking something like this in the yaml:

```yaml
extraManifests:
  - url: http://test.com/file.yaml
    headers:
      - foo: bar
      - baz: test
  - url: http://test2.com/file2.yaml
```

rsmitty:
Hey @salkin! Thanks for this. Looks like a good start, but I've got a question about the implementation in hopes we can make it a bit easier to use.

Also, you'll probably need to run `make docs` any time you make changes to the `v1alpha1_types.go` comments, since those become autogenerated docs and will fail CI if they don't match.

pkg/config/types/v1alpha1/v1alpha1_types.go
andrewrynhard: Yeah I like this. It does make things a bit more verbose if each URL has the same headers, but we don't lose the ability to have different headers if we need to.

salkin: Are we ok with this as it would break v1alpha1 config?

andrewrynhard: Oh, I guess I didn't realize that it broke `v1alpha1` 🤔 . In that case, then I suppose the current implementation would have to work.

andrewrynhard: My only nit in that case, is that I think `ExtraManifestHeaders` should be a map. Then we don't have to split based on `:`.

andrewrynhard:
@salkin Can you rebase? We had a chat about this today and decided that this should be fine for `v1alpha1`. Also, what do you think about making `ExtraManifestHeaders` a map?

salkin:
Sure the map makes more change will redo to that

salkin:
@andrewrynhard  updated to map and rebased, please check if ok.

andrewrynhard:
@salkin Thanks! This looks good to me. I willl merge once we have 2 approvals.

andrewrynhard:
@salkin Looks like you need to run `make docs`.

andrewrynhard:
@salkin looks like linters got you. Try running `make lint-go`. We recently added a PR template that has guidance on how to test a PR locally, but this PR wasn’t created when we had it. Take a look in `.github`. It might help.

PR #2016, opened 2020-04-10
chore: run conform with GitHub actions

andrewrynhard:
This moves the conform execution to GitHub actions.


andrewrynhard:
There is no way to update the PR's status checks using actions :(

PR #2019, opened 2020-04-11
chore: address random CI nits

andrewrynhard:
This PR does the following:

- updates the conform config
- cleans up conform scopes
- moves slash commands to the talos-bot
- adds a check list to the pull request template
- disables codecov comments
- uses `BOT_TOKEN` so all actions are performed as the talos-bot user
- adds a `make conformance` target to make it easy for contributors to
check their commit before creating a PR
- bumps golangci-lint to v1.24.0


.github/workflows/label-accepted.yml
```diff
@@ -5,27 +5,27 @@ on:
...
-      - name: Check for test slash command
+      - name: Check for accept slash command
         uses: xt0rted/slash-command-action@v1
         with:
-          repo-token: ${{ secrets.GITHUB_TOKEN }}
-          command: "test"
-          permission-level: "write"
-          reaction: "true"
-          reaction-type: "rocket"
+          repo-token: ${{ secrets.BOT_TOKEN }}
```
andrewrynhard, line 15: We now have an official `talos-bot` account, and this is the personal access token for that.

internal/app/machined/main.go
```diff
@@ -104,7 +104,7 @@ func init() {
 	}
 }
 
-// nolint: gogcylo
+// nolint: gocylo
```
smira, line 107: looks like still a misprint? :)

andrewrynhard: 👍

andrewrynhard:
We should change `/accept` to `/approve` and use the k8s convention of `/approve` and `/lgtm`.

andrewrynhard:
> so now flows are triggered without `/test` ?

@smira by accepting a PR, CI is allowed to run and is triggered by `talos-bot`. In practice we just use `/accept` now, instead of `/test`. Its the convention used in prow/k8s. To merge we do `/lgtm`, again just like you do in k8s projects.

PR #2022, opened 2020-04-14
chore: update pkgs tag to v0.2.0

andrewrynhard:
This change updates the pkgs tag to v0.2.0. There is no difference in
the content of the packages.


smira:
yay, no more cryptic hashes!

PR #2024, opened 2020-04-14
feat: make machine config persist by default

### LINKED ISSUE #2023: Persist config by default
We should persist the config by default. We should also add an API to tell Talos to download the config again.

rsmitty:
## What? (description) 

feat: make machine config persist by default

## Why? (reasoning)
This PR will change the default behavior for machine configs to
`persist: true`. This seems to be expected behavior from our users so
we'll move to this method for v0.5

Will close #2023

### DIFF
--- a/internal/pkg/provision/providers/docker/docker.go
+++ b/internal/pkg/provision/providers/docker/docker.go
@@ -43,7 +43,9 @@ func (p *provisioner) Close() error {
 
 // GenOptions provides a list of additional config generate options.
 func (p *provisioner) GenOptions(networkReq provision.NetworkRequest) []generate.GenOption {
-	return nil
+	return []generate.GenOption{
+		generate.WithPersist(false),
+	}
 }
 
 // GetLoadBalancers returns internal/external loadbalancer endpoints.
--- a/pkg/config/types/v1alpha1/generate/controlplane.go
+++ b/pkg/config/types/v1alpha1/generate/controlplane.go
@@ -14,6 +14,7 @@ func controlPlaneUd(in *Input) (*v1alpha1.Config, error) {
 	config := &v1alpha1.Config{
 		ConfigVersion: "v1alpha1",
 		ConfigDebug:   in.Debug,
+		ConfigPersist: in.Persist,
 	}
 
 	machine := &v1alpha1.MachineConfig{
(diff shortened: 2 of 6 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -174,6 +174,7 @@ func create(ctx context.Context) (err error) {
 		genOptions := []generate.GenOption{
 			generate.WithInstallImage(nodeInstallImage),
 			generate.WithDebug(configDebug),
+			generate.WithPersist(false),
```
smira, line 177: should we use defaults here? as otherwise we're not testing most common (default case)

it might make sense to disable config persistence for upgrade tests when going from 0.3/0.4 which doesn't do config persist by default

rsmitty: What do you think about [this](https://github.com/talos-systems/talos/pull/2024/files#diff-4f6e60bfe3f88bb8b5699b02b63e29d0R179)? I wanted the docker creation behavior to not persist since there's no use in doing so. But this will also allow the default case to be tested w/ firecracker.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -176,6 +176,11 @@ func create(ctx context.Context) (err error) {
 			generate.WithDebug(configDebug),
 		}
 
+		// Do not persist config files when using docker-based provisioner
+		if provisionerName == "docker" {
```
smira, line 180: you can use call `provisioner.GenOptions()` call below to push `generate.WithPersist(false)` from docker provisioner without doing special case here

cmd/talosctl/cmd/mgmt/cluster/create.go
smira: sgtm, I just left comment how to make it more generic via other call :+1:

internal/pkg/provision/providers/docker/docker.go
```diff
@@ -43,7 +43,9 @@ func (p *provisioner) Close() error {
 
 // GenOptions provides a list of additional config generate options.
 func (p *provisioner) GenOptions(networkReq provision.NetworkRequest) []generate.GenOption {
-	return nil
+	return []generate.GenOption{
+		generate.WithPersist(false),
```
smira, line 47: :+1:

PR #2027, opened 2020-04-14
chore: use a single CHANGELOG

andrewrynhard:
Instead of keeping a CHANGELOG for each release in the master branch, a
single CHANGELOG should be used since it will move into release branches
anyways. This prevents us from having to keep the files in sync across
master and the release branch. This also adds better tooling for
generating the CHANGELOG.md.


hack/release.sh
```diff
@@ -8,7 +8,7 @@ function changelog {
     exit 1
   fi
 
-  conventional-changelog -p angular -i $1 -s
+  git-chglog --output CHANGELOG-0.4.md --tag-filter-pattern "^${1}" -c ./hack/chglog/config.yml "${1}-alpha.1..master"
```
smira, line 11: should this generate `CHANGELOG.md` ?

hack/chglog/CHANGELOG.tpl.md
```diff
@@ -0,0 +1,22 @@
+{{ range .Versions }}
```
smira, line 1: not sure how to comment on that moved file, but should `CHANGELOG.md` contain info for all the releases now?

andrewrynhard: We can I think for now. But github has a max size it will render for markdown. So one day we will have to come up with a way to fix that if we do all changes.

hack/release.sh
andrewrynhard: 👍 thanks

hack/chglog/CHANGELOG.tpl.md
andrewrynhard: In thinking this through some more if we use one global file, then we still need to keep it in sync between release branches and master. If we have `CHANGELOG.md` only be relevant for the current release we won't have to do that since we will generate the file directly from the release branch.

PR #2028, opened 2020-04-14
[wip] chore: remove install from output when running gen config

### LINKED ISSUE #1729: Exclude install section in config on generate

rsmitty:
## What? (description)
chore: remove install from output when running gen config

## Why? (reasoning)
This PR will remove the output of all the various keys under
`machine.install{}` in the output when `talosctl gen config` is issued.
Install section only applies for bare metal, so this should clean up the
config some for the average use case. Users that are doing bare metal
will still have these options available if needed.

Will close #1729

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2028)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/config.go
+++ b/cmd/talosctl/cmd/mgmt/config.go
@@ -15,7 +15,6 @@ import (
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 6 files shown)


andrewrynhard:
Due to reasons I found while refactoring machined, I don't think we can actually do this. I think we can change the default in `v1alpha2` config.

andrewrynhard:
@rsmitty I'm gonna close. As I went through this machined refactor, I can definitely confirm that we can't leave this out.

PR #2029, opened 2020-04-15
fix: set ephemeral partition to max size

andrewrynhard:
This sets the size of the ephemeral partition to the maximum
allowed size at installation time. We have reports of `xfs_growfs` causing
extremely slow boot times when the disk is 1TB or more. In our research
we found evidence that `xfs_growfs` is an expensive operation when
growing to a size of 10 times or more of the base. Instead, users should
create the disk close to the max disk size at install time. The
difference being that `mkfs.xfs` will handle larger disks better.


cmd/installer/cmd/image.go
```diff
@@ -66,8 +66,8 @@ func runImageCmd() (err error) {
 	defer func() {
 		log.Println("detaching loopback device")
 
-		if err = pkg.Lodetach(options.Disk); err != nil {
-			log.Println(err)
+		if e := pkg.Lodetach(options.Disk); e != nil {
+			log.Println(e)
```
andrewrynhard, line 70: This was preventing any errors in the parent func from bubbling up.

internal/pkg/mount/manager/manager.go
```diff
@@ -35,12 +35,12 @@ func (m *Manager) MountAll() (err error) {
...
 		if mountpoint.Resize {
 			if err = mountpoint.ResizePartition(); err != nil {
-				return fmt.Errorf("resize: %w", err)
+				return fmt.Errorf("error resizing %q: %w", iter.Value().Source(), err)
 			}
 		}
 
 		if err = mountpoint.Mount(); err != nil {
-			return fmt.Errorf("mount: %w", err)
+			return fmt.Errorf("error mounting %q: %w", iter.Value().Source(), err)
```
andrewrynhard, line 43: This was useful when debugging.

hack/drone.jsonnet
```diff
@@ -189,10 +189,10 @@ local golint = Step("lint-go", depends_on=[check_dirty]);
...
 local protobuflint = Step("lint-protobuf", depends_on=[check_dirty]);
 local image_aws = Step("image-aws", depends_on=[installer]);
-local image_azure = Step("image-azure", depends_on=[installer]);
-local image_digital_ocean = Step("image-digital-ocean", depends_on=[installer]);
-local image_gcp = Step("image-gcp", depends_on=[installer]);
-local image_vmware = Step("image-vmware", depends_on=[installer]);
+local image_azure = Step("image-azure", depends_on=[image_aws]);
+local image_digital_ocean = Step("image-digital-ocean", depends_on=[image_azure]);
+local image_gcp = Step("image-gcp", depends_on=[image_digital_ocean]);
+local image_vmware = Step("image-vmware", depends_on=[image_gcp]);
```
andrewrynhard, line 195: This prevents `losetup` race condition. Shouldn't effect build times at all.

pkg/blockdevice/table/gpt/gpt.go
```diff
@@ -286,9 +286,9 @@ func (gpt *GPT) Add(size uint64, setters ...interface{}) (table.Partition, error
 	end = start + size/gpt.lba.LogicalBlockSize
 
 	if end > gpt.header.LastUsableLBA {
-		// TODO(andrewrynhard): This calculation is wrong, fix it.
```
andrewrynhard, line 289: Come to find out, this wasn't wrong 😄 .

PR #2036, opened 2020-04-16
fix: ensure ordering of interfaces when deciding hostname

### LINKED ISSUE #1927: [bug] /etc/hosts / IP address inconsistencies due to unordered map
(as discussed with @andrewrynhard)
since `Interfaces` is an unordered map, it causes unpredictable behaviour when picking the hostname

https://github.com/talos-systems/talos/blob/64b5b32732d4023576321111f1cbe79516404758/internal/app/networkd/pkg/networkd/networkd.go#L282

rsmitty:
## What? (description)
fix: ensure ordering of interfaces when deciding hostname

## Why? (reasoning)
This PR will fix some unexpected behavior where we were using Interfaces
as an unordered map, which led to varying behavior with hostname
determination. We will now go through the list of interfaces
alphabetically by name.

Will close #1927

### DIFF
--- a/internal/app/networkd/pkg/networkd/networkd.go
+++ b/internal/app/networkd/pkg/networkd/networkd.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -280,9 +280,17 @@ func (n *Networkd) decideHostname() (hostname string, domainname string, address
 	address = net.ParseIP("127.0.1.1")
 	hostname = fmt.Sprintf("%s-%s", "talos", strings.ReplaceAll(address.String(), ".", "-"))
 
+	// Sort interface names alphabetically so we can ensure parsing order
+	interfaceNames := make([]string, len(n.Interfaces))
```
smira, line 285: this should be `make([]string, 0, len(n.Interfaces))`, as otherwise it creates list of length N and appends N items

internal/app/networkd/pkg/networkd/networkd.go
```diff
@@ -280,9 +280,17 @@ func (n *Networkd) decideHostname() (hostname string, domainname string, address
...
+	for intName := range n.Interfaces {
+		interfaceNames = append(interfaceNames, intName)
+	}
+
```
smira, line 289: I guess `sort.Strings(interfaceNames)` is missing here, as otherwise it's equivalent to the code it was before

rsmitty: yep, thanks. just realized myself. give it a refresh.

internal/app/networkd/pkg/networkd/networkd.go
rsmitty: ah yep. fixed in amended commit. thx.

PR #2037, opened 2020-04-17
docs: add channels branching model proposal

andrewrynhard:
This adds a proposal for a branching model that better facilitates channels.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2037)

### DIFF
--- a/docs/proposals/20200416-ChannelsBranchingModel.md
+++ b/docs/proposals/20200416-ChannelsBranchingModel.md
@@ -1,0 +1,82 @@
+# Proposal: Branching Model for Channels
+
+Author(s): Andrew Rynhard
+
+## Abstract
+
+[A short summary of the proposal.]
+
+## Background
+
+[An introduction of the necessary background and the problem being solved by the proposed change.]
+
+## Proposal
+
+### Alpha to Beta
+
+- Branch creation indicates beta timeline start/reset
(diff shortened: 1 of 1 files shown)


docs/proposals/20200416-RollingReleasesBranchingStrategy.md
```diff
@@ -0,0 +1,79 @@
+# Proposal: RollingReleasesBranchingModel
```
andrewrynhard, line 1:
```suggestion
# Proposal: Rolling Releases Branching Model
```

docs/proposals/20200416-RollingReleasesBranchingStrategy.md
```diff
@@ -0,0 +1,79 @@
...
+
+Author(s): Andrew Rynhard
+
+## Abstract
+
+[A short summary of the proposal.]
+
+## Background
+
+[An introduction of the necessary background and the problem being solved by the proposed change.]
```
smira, line 11: some goals we captured yesterday during the call?

I would also add that no commit might be tagged with two release tags, and that `alpha` tags are all on `master` branch commits, `beta` and `stable` only on release branches.

andrewrynhard: Ahh yes, good one.

PR #2053, opened 2020-04-21
fix: pass dev path to mkfs

andrewrynhard:
This fixes a bug caused by a missing device argument to `mkfs.xfs`.
Without a device, `mkfs.xfs` will error out. Additionally, this ensures
that the installer container is started with the `kmsg` writer that
ensures logs are formatted correctly for `/dev/kmsg`. Without this we
lose a lot of the logs output by the container, one of them being any
error from `mkfs.xfs`

### DIFF
--- a/cmd/installer/pkg/manifest/manifest.go
+++ b/cmd/installer/pkg/manifest/manifest.go
@@ -184,11 +184,11 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
 // Format creates a filesystem on the device/partition.
 func (t *Target) Format() error {
 	if t.Label == constants.BootPartitionLabel {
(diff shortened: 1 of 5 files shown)


cmd/installer/pkg/manifest/manifest.go
```diff
@@ -184,11 +184,11 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
 // Format creates a filesystem on the device/partition.
 func (t *Target) Format() error {
 	if t.Label == constants.BootPartitionLabel {
-		log.Printf("formatting partition %s - %s as %s\n", t.PartitionName, t.Label, "fat")
+		log.Printf("formatting partition %q - %q as %q\n", t.PartitionName, t.Label, "fat")
```
andrewrynhard, line 187: This makes it clear in the logs that the variable is `""`

PR #2057, opened 2020-04-22
fix: ensure disk is not busy

andrewrynhard:
This fixes a few bugs that collectively caused `EBUSY` error when trying
to format the system disk. It ensures that we use the correct path to
the partition, and requires the path to exist. It also removes the
`/etc/kubernetes` mount from `trustd` as it was discovered that this
mount was causing the `EBUSY`.

### DIFF
--- a/internal/app/machined/internal/phase/disk/verify_availability.go
+++ b/internal/app/machined/internal/phase/disk/verify_availability.go
@@ -6,6 +6,7 @@ package disk
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/app/machined/internal/phase/disk/verify_availability.go
```diff
@@ -35,36 +36,50 @@ func NewVerifyDiskAvailabilityTask(devname string) phase.Task {
 // TaskFunc returns the runtime function.
 func (task *VerifyDiskAvailability) TaskFunc(mode runtime.Mode) phase.TaskFunc {
 	return func(r runtime.Runtime) error {
+		//  We only need to verify system disk availability if we are going to
+		// reformat the ephemeral partition.
+		if r.Config().Machine().Install().Force() {
```
smira, line 41: should this check be inverted?

internal/app/machined/pkg/system/services/trustd.go
```diff
@@ -76,7 +76,6 @@ func (t *Trustd) Runner(config runtime.Configurator) (runner.Runner, error) {
 	mounts := []specs.Mount{
 		{Type: "bind", Destination: "/tmp", Source: "/tmp", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: constants.ConfigPath, Source: constants.ConfigPath, Options: []string{"rbind", "ro"}},
-		{Type: "bind", Destination: "/etc/kubernetes", Source: "/etc/kubernetes", Options: []string{"rbind", "ro"}},
```
smira, line 79: :+1:

internal/app/machined/internal/phase/disk/verify_availability.go
andrewrynhard: 👍

PR #2067, opened 2020-04-29
feat: add recovery API

andrewrynhard:
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2067)


andrewrynhard:
I have the initial version done. Ready for some reviews. I still need integration tests.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1455,3 +1459,76 @@ func Install(seq runtime.Sequence, data interface{}) runtime.TaskExecutionFunc {
...
+			return err
+		}
+
+		if err = as.WriteFiles(constants.AssetsDirectory); err != nil {
+			return fmt.Errorf("failed to write recovered assets: %w", err)
+		}
+
+		svc := &services.Bootkube{Recover: true}
+
+		system.Services(r).ReloadAndStart(svc)
```
andrewrynhard, line 1527: This will effectively limit recovery API to the init node since that is the only node that runs bootkube. This needs to be fixed.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -181,6 +181,36 @@ func (s *Server) Reset(ctx context.Context, in *machine.ResetRequest) (reply *ma
...
 }
 
+// Recover recovers the control plane.
+//
+// nolint: dupl
+func (s *Server) Recover(ctx context.Context, in *machine.RecoverRequest) (reply *machine.RecoverResponse, err error) {
+	log.Printf("recover request received")
+
+	if s.Controller.Runtime().Config().Machine().Type() == runtime.MachineTypeJoin {
+		return nil, fmt.Errorf("recover can only be performed on a control plane node")
```
andrewrynhard, line 191: I can't get this to actually return back to the client for some reason. This follows the same pattern as other RPCs so I don't really get it.

internal/integration/api/recover.go
```diff
@@ -0,0 +1,107 @@
...
+		err := suite.k8s.Clientset.CoreV1().Pods(pod.GetNamespace()).Delete(suite.ctx, pod.GetName(), metav1.DeleteOptions{})
+		suite.api.Assert().NoError(err)
+	}
+
+	for i, node := range nodes {
+		if i != 0 {
+			continue
+		}
+
+		// first node should be init node, due to bug with etcd cluster build for init node
```
smira, line 80: seems like this comment shouldn't be here

internal/app/machined/pkg/system/system.go
```diff
@@ -83,6 +83,34 @@ func (s *singleton) Load(services ...Service) []string {
...
+	}
+
+	ids := make([]string, 0, len(services))
+
+	for _, service := range services {
+		id := service.ID(s.runtime)
+		ids = append(ids, id)
+
+		if _, exists := s.state[id]; exists {
+			svcrunner := NewServiceRunner(service, s.runtime)
```
smira, line 104: idk, if we should do some check that service runner being replaced (contents of `s.state[id]`) is in some terminal state? as otherwise there might be a chance of duplicate service runners for the same service

internal/integration/api/recover.go
andrewrynhard: Yep, thanks :+1:

internal/app/machined/pkg/system/system.go
andrewrynhard: Good point.

PR #2078, opened 2020-05-01
feat: add bootstrap API

andrewrynhard:
This adds the ability to bootstrap a cluster using the API.
The API simply starts the bootkube service.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2078)


internal/app/machined/pkg/system/services/bootkube.go
```diff
@@ -101,33 +100,6 @@ func (b *Bootkube) PreFunc(ctx context.Context, r runtime.Runtime) (err error) {
...
-		}
-
-		return nil
-	})
-	if err != nil {
-		return fmt.Errorf("failed to put state into etcd: %w", err)
-	}
-
-	log.Println("updated initialization status in etcd")
-
```
andrewrynhard, line 130: This is now a task.

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -439,7 +424,142 @@ func (e *Etcd) args(r runtime.Runtime) ([]string, error) {
...
+		err = e.argsForInit(r)
+		if err != nil {
+			errCh <- err
+		}
+	case runtime.MachineTypeControlPlane:
+		err = e.argsForControlPlane(r)
+		if err != nil {
+			errCh <- err
+		}
+	}
```
andrewrynhard, line 473: We need this still because we have to handle the init node differently. Goal is to get rid of the init node, but we have to keep it around for some time.

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -439,7 +424,143 @@ func (e *Etcd) args(r runtime.Runtime) ([]string, error) {
...
 	}
 
-	return blackListArgs.Merge(extraArgs).Args(), nil
+	e.args = blackListArgs.Merge(extraArgs).Args()
+
+	return nil
+}
+
+func (e *Etcd) setup(ctx context.Context, r runtime.Runtime, errCh chan error) {
+	var err error
```
smira, line 433: this can be simplified by wrapping a function which returns error:

```go
  errCh <- func() error {
    if ...; err != nil {
       return err
    }
   
   return nil
  }
}
```

this way we can't miss pushing value to `errCh`

andrewrynhard: Nice :+1:

PR #2083, opened 2020-05-06
test: stabilize tests by bumping timeouts

smira:
Bump timeouts for reset API test as K8s control plane teardown might
take 3 minutes on its own.

Bump Go Firecracker SDK timeout when talking to firecracker process.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2083)

### DIFF
--- a/internal/integration/api/reset.go
+++ b/internal/integration/api/reset.go
@@ -86,18 +86,14 @@ func (suite *ResetSuite) TestResetNodeByNode() {
 
 			var uptimeAfter float64
 
-			start := time.Now()
-
-			suite.Require().NoError(retry.Constant(3 * time.Minute).Retry(func() error {
-				since := time.Since(start)
-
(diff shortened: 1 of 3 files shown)


internal/integration/api/reset.go
```diff
@@ -86,18 +86,14 @@ func (suite *ResetSuite) TestResetNodeByNode() {
...
-				since := time.Since(start)
-
+			suite.Require().NoError(retry.Constant(10 * time.Minute).Retry(func() error {
 				uptimeAfter, err = suite.ReadUptime(nodeCtx)
 				if err != nil {
 					// API might be unresponsive during reboot
 					return retry.ExpectedError(err)
 				}
 
-				if uptimeAfter >= uptimeBefore+since.Seconds() {
```
smira, line 100: removed that as it actually doesn't test for node reboot this way

PR #2087, opened 2020-05-07
chore: serialize firecracker e2e tests

rsmitty:
This PR will ensure that the firecracker provision tests will only run
after a successful e2e_firecracker run. This is being added in hopes of
freeing up some resources during CI testing and making things more
stable.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2087)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -574,6 +574,7 @@ steps:
   - name: tmp
     path: /tmp
   depends_on:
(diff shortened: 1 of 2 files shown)


andrewrynhard:
Do we still need this? I think the errors we were seeing were due to bad hardware.

PR #2088, opened 2020-05-07
fix: write machined RPC logs to file

andrewrynhard:
This ensures that the machined RPC logs are written to disk so that users
can retrieve them via the log API.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2088)


internal/app/machined/main.go
```diff
@@ -143,7 +144,19 @@ func main() {
 			Controller: c,
 		}
 
-		e := factory.ListenAndServe(server, factory.Network("unix"), factory.SocketPath(constants.MachineSocketPath))
+		f, e := os.OpenFile(filepath.Join(constants.SystemRunPath, "log", "machined.log"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
```
smira, line 147: you also could potentially use https://github.com/talos-systems/talos/blob/master/internal/app/machined/pkg/system/log/log.go to abstract away a bit this piece

PR #2103, opened 2020-05-13
feat: add support for file scheme

andrewrynhard:
This adds support for reading the config from a local file.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2103)

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net/url"
 	"os"
 	"path/filepath"
 	"unsafe"
@@ -36,6 +37,7 @@ type Options struct {
 	Upgrade         bool
 	Force           bool
 	Zero            bool
+	Save            bool
 }
 
 // Install installs Talos.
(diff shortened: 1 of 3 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -276,6 +278,38 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 		}
 	}
 
+	if i.options.Save {
```
andrewrynhard, line 281: i think this should actually be implicit ... no need for an extra flag

pkg/download/download.go
```diff
@@ -61,6 +61,10 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
 		return b, err
 	}
 
+	if u.Scheme == "file" {
```
andrewrynhard, line 64: i don't think we need this actually ... if persistence is required, then we will never download the config

andrewrynhard: This is probably safe to have.

cmd/installer/pkg/install/install.go
andrewrynhard: if we make this implicit, i don't like that we still attempt to parse the url ... due to parsing ambiguities, parsing non-urls like 'none' works but still doesnt make me feel good about it

PR #2106, opened 2020-05-14
fix: run machined API as a service

andrewrynhard:
In recent refactoring the machined API service was changed to run outside
of the service framework. This brings it back as a service.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2106)

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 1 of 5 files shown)


internal/app/machined/main.go
```diff
@@ -138,38 +140,39 @@ func main() {
...
 	if err = c.Run(runtime.SequenceBoot, nil); err != nil {
 		handle(err)
 	}
 
+	// Watch and handle runtime events.
+	c.Runtime().Events().Watch(func(events <-chan machine.Event) {
+		for {
+			select {
+			case event := <-events:
+				if event.Data.TypeUrl == "talos/runtime/"+proto.MessageName(&machine.SequenceEvent{}) {
```
andrewrynhard, line 161: I would like to eventually come up with some helpers that could do this kind of checking but this should get us by for now.

smira: my preference will be towards passing some interface as event and allow to typecast directly to Go type

protobuf internally requires too much handling imho

andrewrynhard: i agree ... the hard part then becomes having to maintain internal structs and protuf defined ones... i also wonder how structs would work with plugings. Protobuf felt like the only common denominator.

andrewrynhard:
This seems to be hanging when shutting down all services on a reboot. Still digging into it.

andrewrynhard:
> This seems to be hanging when shutting down all services on a reboot. Still digging into it.

I see. I need to stop the gRPC server on a context cancel.

PR #2107, opened 2020-05-14
chore: fix nits in the events code

smira:
1. `sync.Mutex` zero value is fine as initial value, and it's part of
the `Events` structured which is passed by pointer, it's fine to embed
it. (simplify)

2. Add safety net around potentially blocked event subscribers so that
it doesn't block event propagation. This is not a perfect solution, but
my feeling that it's better than blocking.

3. Fix delete sequence - first remove the channel, then close it, as
otherwise send might panic to closed channel.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2107)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
@@ -20,14 +20,13 @@ import (
 type Events struct {
 	subscribers []chan machine.Event
 
-	*sync.Mutex
+	sync.Mutex
 }
 
 // NewEvents initializes and returns the v1alpha1 runtime event stream.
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -38,8 +37,8 @@ func (e *Events) Watch(f runtime.WatchFunc) {
 	ch := e.add()
 
 	go func() {
-		defer e.delete(ch)
 		defer close(ch)
+		defer e.delete(ch)
```
smira, line 41: send to closed channel panics, so we should first delete channel from subscribers, then close it

PR #2113, opened 2020-05-15
chore: replace underlying event implementation with single slice

smira:
The idea here is to use single slice of events for all the consumers.
Each consumer keeps its own position within the stream, and stream is
structured as circular buffer to avoid using too much memory.

This implementation allows for one more future: looking "back" into the
event history and returning past event starting with some offset (e.g.
timestamp, event ID, etc.). This feature is not implemented yet.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2113)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
@@ -5,97 +5,186 @@
 package v1alpha1
 
 import (
+	"context"
 	"fmt"
-	"log"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes/any"
 
-	"github.com/talos-systems/talos/api/machine"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 )
 
 // Events represents the runtime event stream.
+//
+// Events internally is implemented as circular buffer of `runtime.Event`.
+// `e.stream` slice is allocated to the initial capacity and slice size doesn't change
+// throughout the lifetime of Events.
+//
+// To explain the internals, let's call `Publish()` method 'Publisher' (there might be
+// multiple callers for it), and each `Watch()` handler as 'Consumer'.
+//
+// For Publisher, `Events` keeps `e.writePos` and `e.gen`, `e.writePos` is index into
+// `e.stream` to write the next event to. After the write `e.writePos` is incremented.
+// As `e.writePos` goes above capacity-1, it wraps around  to zero and `e.gen` is incremented.
+//
+// So at any time `0 <= e.writePos < e.cap`, but `e.gen` indicates how many times `e.stream` slice
+// was overwritten.
+//
+// Each Consumer captures initial position it starts consumption from as `pos` and `gen` which are
+// local to each Consumers, as Consumers are free to work on their own pace. Following diagram shows
+// Publisher and three Consumers:
+//
+//                                                 Consumer 3                         Consumer 2
+//                                                 pos = 9                            pos = 16
+//                                                 gen = 1                            gen = 1
+//  e.stream []Event                               |                                  |
+//                                                 |                                  |
+//  +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+//  | 0  | 1  | 2  | 3  | 4  | 5  | 6  | 7  | 8  | 9  | 10 | 11 | 12 | 13 | 14 | 15 | 16 |17  |
+//  +----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+----+
+//                                       |                                  |
+//                                       |                                  |
+//                                       Consumer 1                         Publisher
+//                                       pos = 7                            e.writePos = 14
+//                                       gen = 2                            e.gen = 2
+//
+// Capacity of Events in this diagram is 18, Publisher published already 14 + 2 * 18 = 40 events, so it
+// already overwrote `e.stream` twice fully (e.gen = 2).
+//
+// Consumer1 is trying to keep up with the publisher, it's on the same `gen`, and it has 14-7 = 7 events
+// to catch up.
+//
+// Consumer2 is on `gen` 1, so it is reading events which were published before the Publisher did last
+// wraparound for `e.writePos` at `e.gen == 1`. Consumer 2 has a lot of events to catch up, but as it stays
+// on track, it can still do that.
+//
+// Consumer3 is doing bad - it's on `gen` 1, but Publisher already overwrote this element while going on `gen` 2,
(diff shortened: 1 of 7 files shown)


smira:
this is still WIP, I would like to see if I can compare with previous implementation over a set of benchmarks.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events_test.go
```diff
@@ -58,137 +31,126 @@ func BenchmarkWatch(b *testing.B) {
...
-			for i := 0; i < tt.count; i++ {
-				e.Watch(func(events <-chan machine.Event) {
-					<-events
+			l := rate.NewLimiter(1000, tt.cap/2)
 
-					mu.Lock()
-					got++
-					mu.Unlock()
+			for i := 0; i < tt.messages; i++ {
+				_ = l.Wait(context.Background())
```
smira, line 112: this is required to make publishing rate-limited, so that it doesn't overrun the buffer while watchers are still starting their goroutines

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -5,97 +5,122 @@
...
 
-	for _, sub := range e.subscribers {
-		// drop the event if some subscriber is stuck
-		// dropping is bad, but better than blocking event propagation
-		select {
-		case sub <- event:
-		default:
+		for {
+			e.mu.Lock()
+			for pos == e.writePos {
```
andrewrynhard, line 70: I'm struggling to understand when this would be false or true and why we need this.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -5,97 +5,122 @@
...
+
+				select {
+				case <-ctx.Done():
+					e.mu.Unlock()
+					return
+				default:
+				}
+			}
+
+			if e.gen > gen && pos < e.writePos {
```
andrewrynhard, line 81: I don't understand how this specific condition is a buffer overrun.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -5,97 +5,122 @@
...
 
-	for i, sub := range e.subscribers {
-		if sub == ch {
-			l := len(e.subscribers)
-			e.subscribers[i] = e.subscribers[l-1]
-			e.subscribers[l-1] = nil
-			e.subscribers = e.subscribers[:l-1]
-		}
+	if e.writePos == 0 {
+		e.gen++
```
andrewrynhard, line 122: IIUC we should never have a generation of zero?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -5,97 +5,122 @@
...
+
+			if e.gen > gen && pos < e.writePos {
+				// buffer overrun, there's no way to signal error in this case,
+				// so for now just return
+				e.mu.Unlock()
+				return
+			}
+
+			event := e.stream[pos]
+			pos = (pos + 1) % e.cap
```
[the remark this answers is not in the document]
smira: because `e.stream` is a circular buffer, so when pos reaches the end (`e.cap-1`), it should go back to the first item?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -5,97 +5,122 @@
...
-			// In the future, we can publish `talos/runtime`, and
-			// `talos/plugin/<plugin>` (or something along those lines) events.
-			TypeUrl: fmt.Sprintf("talos/runtime/%s", proto.MessageName(msg)),
-			Value:   value,
-		},
-	}
+	e.mu.Lock()
+	pos := e.writePos
+	gen := e.gen
+	e.mu.Unlock()
```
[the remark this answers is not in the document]
smira: we capture the `gen` and `pos`, so if we overrun happens (publishing faster than consuming), we can detect that. 

as an edge case, if something publishes `e.cap` items so quickly that consumer never had a chance to catch up, `e.pos == pos` (as position wraps around and it's same), while `gen` will be different from `e.gen`, as `e.gen` increments every time `e.pos` wraps around zero

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
smira: I tried to explain this above, but this means that consumer is behind the publisher so much that `e.writePos` overwrote the current position consumer was supposed to read next

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
smira: after first publish, `e.writePos == 1`, so initial `e.gen` stays at zero until first time `e.writePos` wraps around zero

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
smira: we start with `pos == e.writePos`, so consumer waits for the entry to be published by the publisher.

as publisher writes another entry to `e.stream`, it increases `e.writePos` (with mod, so it might go back to zero as well), and consumer picks up element from `pos` and increments `pos` until it once again hits `e.writePos`

smira:
```
 go test -v -benchmem -timeout 10s ./internal/app/machined/pkg/runtime/v1alpha1 -run '^$' -bench '^(BenchmarkWatch|BenchmarkPublish)$'
goos: linux
goarch: amd64
pkg: github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1
BenchmarkWatch
BenchmarkWatch-12         658449              2436 ns/op             731 B/op          7 allocs/op
BenchmarkPublish
BenchmarkPublish/Watchers-0
BenchmarkPublish/Watchers-0-12           4390780               240 ns/op              64 B/op          2 allocs/op
BenchmarkPublish/Watchers-1
BenchmarkPublish/Watchers-1-12           5691968               235 ns/op              64 B/op          2 allocs/op
BenchmarkPublish/Watchers-10
BenchmarkPublish/Watchers-10-12          5638048               228 ns/op              64 B/op          2 allocs/op
PASS
ok      github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1       12.563s
```

andrewrynhard:
Nice performance gains!

smira:
> Nice performance gains!

Hard to compare, as previously benchmark was sending 1 message to ~200k watchers :)

most of the savings are due to not serializing message in Publish path.

PR #2134, opened 2020-05-22
feat: implement simplified client method to consume events

smira:
This provides interface to consume (watch) events in the same way (with
the same API) as server code does. So some processing might be moved
back and forth between client and server without much code change.

Also moved peer methods out of helpers to the main client package, they
might be useful for any client, not only talosctl.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2134)


smira:
potential fix for test failure #2136 (unrelated to this PR)

PR #2145, opened 2020-05-25
feat: adjust time properly in timed via adjtime()

smira:
This should be proper way to adjust time incrementally without causing
jumps one in +/- direction. Time-sensitive services might be confused by
huge jumps.

Next steps: insert/delete leap seconds, provide timed health status.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2145)

### DIFF
--- a/internal/app/timed/pkg/ntp/options.go
+++ b/internal/app/timed/pkg/ntp/options.go
@@ -17,6 +17,8 @@ const (
 	MaxAllowablePoll = 1024
 	// MinAllowablePoll is the minimum time allowed for a client to query a time server
 	MinAllowablePoll = 4
+	// AdjustTimeLimit is a maximum time drift to compensate via adjtimex()
+	AdjustTimeLimit = 128 * time.Millisecond
 )
 
 func defaultOptions() *NTP {
--- a/internal/app/timed/pkg/reg/reg.go
+++ b/internal/app/timed/pkg/reg/reg.go
@@ -6,12 +6,14 @@ package reg
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/empty"
 	"google.golang.org/grpc"
 
+	healthapi "github.com/talos-systems/talos/api/health"
 	timeapi "github.com/talos-systems/talos/api/time"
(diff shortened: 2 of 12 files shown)


smira:
I want to provide more output decoding the response of `adjtime()`, as it acts in a weird way on firecracker for me.

internal/app/timed/pkg/ntp/options.go
```diff
@@ -23,9 +23,9 @@ func defaultOptions() *NTP {
 	// defaults for minpoll + maxpoll
 	// http://www.ntp.org/ntpfaq/NTP-s-algo.htm#AEN2082
 	return &NTP{
-		Server:  "pool.ntp.org",
-		MaxPoll: MaxAllowablePoll * time.Second,
-		MinPoll: 64 * time.Second,
+		Server:  "ru.pool.ntp.org",
```
smira, line 26: temporary changes for debugging, will remove that

andrewrynhard: I wonder if we can pick this default intelligently. Based on location of the machine.

andrewrynhard: Not for this PR of course

smira: I read in the FAQ that `pool.ntp.org` is actually supposed to give out best server based on location, so probably we don't need much for defaults here.

andrewrynhard: Oh nice.

smira:
`timed` doesn't quite work on Firecracker, because Firecracker doesn't implement any adjustable clock (just uses host clock), so I tested service on the host.

internal/app/timed/pkg/reg/reg.go
```diff
@@ -88,3 +91,62 @@ func genProtobufTimeResponse(local, remote time.Time, server string) (*timeapi.T
...
+		select {
+		case <-srv.Context().Done():
+			return srv.Context().Err()
+		case <-ticker.C:
+			resp, err := r.Check(srv.Context(), &empty.Empty{})
+			if err != nil {
+				return err
+			}
+
+			if err := srv.Send(resp); err != nil {
```
andrewrynhard, line 128: On a related note, these would be good events one day.

PR #2156, opened 2020-05-29
feat: add arm64 support

andrewrynhard:
This adds support for building arm64 images.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2156)


andrewrynhard:
@smira I think we can close this. This was more experimental.

PR #2157, opened 2020-05-29
fix: make services depend on timed

andrewrynhard:
This adds a dependency on timed to services that depend on time.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2157)


andrewrynhard:
> I can do proper HC for timed once I'm done with other stuff so that we can wait for the first sync, but I wonder if we should allow to boot if timed can't sync? (in airgapped or other requirements)?

Since time is so important I think that makes sense. With the way the boot sequence works now, if timed never becomes healthy (once we have that) there is a ten minute timeout where we will reboot.

PR #2164, opened 2020-06-03
fix: allow node names

### LINKED ISSUE #2163: talosctl nodes no longer accepts names
## Bug Report

As of 67b31029a7838045cb819766a1a53e63addd776c, names are no longer allowed for nodes.  This change was done to _allow_ IPv6 addresses, but in the process blocked names.

### Environment

- Platform:

Ulexus:
This allows node names for APID, but checks for empty or port-appended
addresses in a best-effort manner.

Fixes #2163

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2164)

### DIFF
--- a/pkg/net/net.go
+++ b/pkg/net/net.go
@@ -50,6 +50,36 @@ func FormatAddress(addr string) string {
 	return addr
 }
 
+// AddressContainsPort checks to see if the supplied address contains both an address and a port.
+// This will not catch every possible permutation, but it is a best-effort routine suitable for prechecking human-interactive parameters.
+func AddressContainsPort(addr string) bool {
+	if !strings.Contains(addr, ":") {
+		return false
+	}
+
+	pieces := strings.Split(addr, ":")
+
+	if ip := net.ParseIP(strings.Trim(addr, "[]")); ip != nil {
+		return false
+	}
+
+	// Check to see if it parses as an IP _without_ the last (presumed) `:port`
+	trimmedAddr := strings.TrimSuffix(addr, ":"+pieces[len(pieces)-1])
+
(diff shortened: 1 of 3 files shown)


Ulexus:
Working on a better check system...

Ulexus:
Okay, much better now.

pkg/net/net.go
```diff
@@ -50,6 +50,36 @@ func FormatAddress(addr string) string {
 	return addr
 }
 
+// AddressContainsPort checks to see if the supplied address contains both an address and a port.
+// This will not catch every possible permutation, but it is a best-effort routine suitable for prechecking human-interactive parameters.
+func AddressContainsPort(addr string) bool {
+	pieces := strings.Split(addr, ":")
```
smira, line 56: can we simplify it to:

* if `addr` doesn't contain ':', we let it through
* if `addr` contains `:`, but it parses as valid IP, we let it through?

Ulexus: For the case we are looking at right now, that would be fine.  I was making the check a bit more elaborate to be useful in the other direction, too:  checking in cases where an address _should_ contain a port.

I'm not really sure; maybe it _is_ overly-elaborate for what it needs to be.

Ulexus: Oh, sorry, no.  That still isn't sufficient:  the problem is that we need to be able to accept DNS _names_ as well as IPs.

Ulexus: However, we _can_ branch early in the absence of any `:`, which seems like a reasonable simplification.

smira: yep, sounds like I thought about `AddressDoesntContainPort` which might be simpler

PR #2165, opened 2020-06-04
refactor: implement LoggingManager as central log flow processor

smira:
At the moment only write path is covered, read path still needs to be
updated.


internal/app/machined/pkg/system/runner/goroutine/goroutine.go
```diff
@@ -80,9 +79,9 @@ func (r *goroutineRunner) wrappedMain() (err error) {
 		}
 	}()
 
-	var w *log.Log
+	var w io.WriteCloser
 
-	w, err = log.New(r.id, r.opts.LogPath)
+	w, err = r.opts.LoggingManager.ServiceLog(r.id).Writer()
```
smira, line 84: need to add `w.Close()` here

internal/app/machined/pkg/system/services/osd.go
```diff
@@ -88,7 +88,6 @@ func (o *OSD) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: "/tmp", Source: "/tmp", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: constants.ConfigPath, Source: constants.ConfigPath, Options: []string{"rbind", "ro"}},
 		{Type: "bind", Destination: path.Dir(constants.ContainerdAddress), Source: path.Dir(constants.ContainerdAddress), Options: []string{"bind", "ro"}},
-		{Type: "bind", Destination: constants.DefaultLogPath, Source: constants.DefaultLogPath, Options: []string{"bind", "ro"}},
```
smira, line 91: I believe we don't need it as logs are served by machined

andrewrynhard: :+1:

Ulexus:
I do find it a little weird to write to a file and then use fsnotify to watch for changes to that file.  So long as we are _always_ file-oriented, that is not really a problem, but it does seem like it would be easier to just implement a follow hook in the common log handler.

To be clear:  I don't have a problem with how this is done for now.

smira:
> I do find it a little weird to write to a file and then use fsnotify to watch for changes to that file. So long as we are _always_ file-oriented, that is not really a problem, but it does seem like it would be easier to just implement a follow hook in the common log handler.
> 
> To be clear: I don't have a problem with how this is done for now.

yes, we could try to shortcut things and connect `Write()` to a file as a way notify `Read()` that there's new data available, but that means tight coupling of both parts being part of the same process. Today it's in the same `machined` process, but I think this might not always be the case going forward. Probably we could move log reading to `osd`, or we want to spawn services from `serviced`. In this case OS-based notifications is the only way to go.

PR #2169, opened 2020-06-09
feat: update k8s and sonobuoy versions

rsmitty:
This PR will update k8s to the lastest 1.18 release and bump sonobuoy to
help resolve some e2e flakes. Also adds some retry logic around the
sonobuoy run.


hack/test/e2e.sh
```diff
@@ -102,12 +102,17 @@ function run_talos_integration_test_docker {
...
     --kubeconfig ${KUBECONFIG} \
     --wait \
     --skip-preflight \
     --plugin e2e \
-    --mode ${SONOBUOY_MODE}
+    --mode ${SONOBUOY_MODE}; do
+    [[ $(date +%s) -gt $timeout ]] && exit 1
+    echo "attempting to run sonobuoy"
+    sleep 10
+  done
```
andrewrynhard, line 115: In the case of running in the certification mode, will this run through the whole thing again if it fails? It looks it actually covers the case when running the sonobuoy command fails initially at all for some reason. Correct me if I am wrong.

rsmitty: Yeah, it'll re-kick if sonobuoy fails at all initially. But the way I'm thinking it will work is, since sonobuoy takes forever to run, if it fails deep in the run for some reason, it'd be past the timeout and would exit w/ error code 1.

PR #2171, opened 2020-06-09
Set the firecracker seccomp-level

dmolik:
On more secure host machines the firecracker seccomp-level needs
to be set to 0 to allow VMs to start

## What? (description)
set the seccomp-level to not filter InstanceStart

## Why? (reasoning)
for local cluster testing using firecracker


andrewrynhard:
@dmolik I can fix other issues with the commit message, but you will have to do something like:

```
git commit -s --amend --no-edit && git push -f
```
for legal reasons. The `-s` signs the commit with the DCO.

dmolik:
oh my bad, signed.

dmolik:
I re-worked the commit message.

internal/pkg/provision/providers/firecracker/launch.go
```diff
@@ -118,6 +118,7 @@ func Launch() error {
 			cmd := firecracker.VMCommandBuilder{}.
 				WithBin("firecracker").
 				WithSocketPath(config.FirecrackerConfig.SocketPath).
+				WithArgs([]string{"--seccomp-level", "0"}).
```
smira, line 121: I'm not sure this is good change as it seems to disable secccomp filters altogether for firecracker

so this potentially makes it less secure/more vulnerable

smira: could it be that the problem is missing seccomp support? so that firecracker fails applying seccomp filters?

dmolik: I'll double check with a different version of firecracker

smira: ah, I see about `InstanceStart`, do you have the error message?

dmolik: So this is an issue with the Gentoo build of Firecracker, your intuition was spot on.

dmolik:
So as is, I'm going to drop this PR, due to this being a Gentoo build issue of firecracker, there is an opportunity to bubble the seccomp-level setting to the CLI, in-case other folks have a similar issue to myself. Thoughts?

PR #2185, opened 2020-06-11
fix: isolate kubelet /run directory

smira:
For better security, kubelet shouldn't have access to contents of `/run`
directory which contains Talos services sockets and other system
information.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2185)

### DIFF
--- a/internal/app/machined/pkg/system/services/kubelet.go
+++ b/internal/app/machined/pkg/system/services/kubelet.go
@@ -105,6 +105,10 @@ func (k *Kubelet) PreFunc(ctx context.Context, r runtime.Runtime) error {
 		return err
 	}
 
+	if err := os.MkdirAll("/run/kubelet", 0700); err != nil {
+		return err
+	}
+
 	client, err := containerdapi.New(constants.ContainerdAddress)
 	if err != nil {
 		return err
@@ -164,7 +168,8 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: "/etc/os-release", Source: "/etc/os-release", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/etc/cni", Source: "/etc/cni", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"rbind", "rshared", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/run", Source: "/run/kubelet", Options: []string{"rbind", "rprivate", "rw"}},
+		{Type: "bind", Destination: "/run/containerd", Source: "/run/containerd", Options: []string{"rbind", "rprivate", "ro"}},
 		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/log/pods", Source: "/var/log/pods", Options: []string{"rbind", "rshared", "rw"}},


smira:
WIP to see if it works for all the tests, need to make it `/run/kubelet`, not `/run/system/kubelet`.

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -164,7 +168,8 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: "/etc/os-release", Source: "/etc/os-release", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/etc/cni", Source: "/etc/cni", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"rbind", "rshared", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/var/run", Source: "/run/kubelet", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/var/containerd", Source: "/run/containerd", Options: []string{"rbind", "rshared", "rw"}},
```
smira, line 172: might need to change this to `ro`, I think `ro` is enough here

andrewrynhard, line 172: I _think_ we might be able to get away with keeping the original socket location if we do something like:

```
{Type: "bind", Destination: constants.ContainerdAddress, Source: constants.ContainerdAddress, Options: []string{"bind", "ro"}},
```

andrewrynhard: The error about multiple overlay filesystems might be due to the fact that we are trying to mount a dir, vs a file.

smira: thanks, I'll try that!

smira: looks like same error when both mounts end up in the same tree:

```
 service[kubelet](Waiting): Runner Containerd(kubelet) exited without error, going to restart it
[  141.519972] [talos] service[kubelet](Waiting): Error running Containerd(kubelet), going to restart forever: failed to create task: "kubelet": OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/run/containerd\\\" to rootfs \\\"/run/containerd/io.containerd.runtime.v2.task/k8s.io/kubelet/rootfs\\\" at \\\"/run/containerd/io.containerd.runtime.v2.task/k8s.io/kubelet/rootfs/run/containerd\\\" caused \\\"no such file or directory\\\"\"": unknown
[  146.547418] overlayfs: upperdir is in-use as upperdir/workdir of another mount, accessing files from both mounts will result in undefined behavior.
[  146.549122] overlayfs: workdir is in-use as upperdir/workdir of another mount, accessing files from both mounts will result in undefined behavior.
```

smira: trying to debug that, I wonder if that is some error with our containerd runner

andrewrynhard: Strange. Ok so I was wrong then. I guess overlay is used regardless. I'm trying to think of a way we can secure this but keep the containerd socket the default path. Will keep thinking.

smira: figured it out, `rshared` was a problem, as this was creating a loop

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -164,7 +168,8 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: "/etc/os-release", Source: "/etc/os-release", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/etc/cni", Source: "/etc/cni", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"rbind", "rshared", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"rbind", "rshared", "rw"}},
+		{Type: "bind", Destination: "/run", Source: "/run/kubelet", Options: []string{"rbind", "rprivate", "rw"}},
```
smira, line 171: `/var/run` is a symlink to `/run` inside container, so using `/run` directly

internal/app/machined/pkg/system/services/kubelet.go
smira: with `rshared`, on the host side we had `/run/containerd` mounted under `/run/kubelet`, so on kubelet restart two mounts were layered one on top of another (kind of duplicate mount?)

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -105,6 +105,10 @@ func (k *Kubelet) PreFunc(ctx context.Context, r runtime.Runtime) error {
 		return err
 	}
 
+	if err := os.MkdirAll("/run/kubelet", 0700); err != nil {
```
andrewrynhard, line 108: nit: Can we do `/run/system/kubelet`? That is where we have other service directories.

smira: copying from Slack for history:

> I thought /run/system is for system services vs. k8s stuff which lives in /run ?
> e.g. CRI containerd is /run/containerd/, but system containerd is /run/system/containerd

andrewrynhard: :+1:

PR #2193, opened 2020-06-13
test: default to using the bootstrap API

andrewrynhard:
This moves our test scripts to using the bootstrap API. Some
automation around invoking the bootstrap API was also added
to give the same ease of use when creating clusters with the
CLI.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2193)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -11,6 +11,7 @@ import (
 	"math/big"
 	"net"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
@@ -23,11 +24,13 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/provision/access"
 	"github.com/talos-systems/talos/internal/pkg/provision/providers"
 	"github.com/talos-systems/talos/pkg/cli"
+	"github.com/talos-systems/talos/pkg/client"
 	clientconfig "github.com/talos-systems/talos/pkg/client/config"
 	"github.com/talos-systems/talos/pkg/config"
 	"github.com/talos-systems/talos/pkg/config/types/v1alpha1/generate"
 	"github.com/talos-systems/talos/pkg/constants"
 	talosnet "github.com/talos-systems/talos/pkg/net"
+	"github.com/talos-systems/talos/pkg/retry"
 )
 
 var (
@@ -208,7 +211,6 @@ func create(ctx context.Context) (err error) {
 		}
 
 		genOptions = append(genOptions, generate.WithEndpointList(endpointList))
-
 		configBundleOpts = append(configBundleOpts,
 			config.WithInputOptions(
 				&config.InputOptions{
@@ -232,6 +234,18 @@ func create(ctx context.Context) (err error) {
 	for i := 0; i < masters; i++ {
 		var cfg runtime.Configurator
 
+		nodeReq := provision.NodeRequest{
+			Name:     fmt.Sprintf("%s-master-%d", clusterName, i+1),
+			IP:       ips[i],
+			Memory:   memory,
+			NanoCPUs: nanoCPUs,
+			DiskSize: diskSize,
+		}
+
+		if i == 0 {
+			nodeReq.Ports = []string{"50000:50000/tcp", "6443:6443/tcp"}
+		}
+
 		if withInitNode {
 			if i == 0 {
 				cfg = configBundle.Init()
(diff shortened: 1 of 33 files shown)


internal/pkg/cluster/check/service.go
```diff
@@ -22,14 +22,13 @@ func ServiceStateAssertion(ctx context.Context, cluster ClusterInfo, service str
...
+	nodes := cluster.NodesByType(runtime.MachineTypeControlPlane)
 
-	if len(initNodes) != 1 {
-		return fmt.Errorf("init node not found, len(initNodes) = %d", len(initNodes))
+	if len(nodes) == 0 {
+		return fmt.Errorf("no control plane nodes found")
 	}
 
-	nodeCtx := client.WithNodes(ctx, initNodes[0])
+	nodeCtx := client.WithNodes(ctx, nodes[0])
```
smira, line 31: we need "bootstrap" node here iirc, as we're checking for `bootkube`, and bootkube only runs on bootstrap/init node. should we also enforce some way we pick up bootstrap node? 

and as init node is technically still supported, we could do the following:

1. if init node is defined, use it
2. otherwise, pick the first control plane node with the first IP in alphabetical order (and make sure we call bootstrap on that very same node).

smira:
looks awesome to me, just some comments

internal/pkg/cluster/check/service.go
andrewrynhard: Yeah I wasn't quite sure how to handle this. The problem I seem to be having now is that `nodes[0] could be any one of the control plane nodes. Alphabetical order should fix this.

andrewrynhard: I made these changes, and that gets through the initial bootstrapping but running cluster health checks after reboot/reset tests complains about not being able to find the bootkube service. Oh, now that I am saying it out loud, the bootkube service isn't created ever again. That is the whole point of the bootstrap API. Any ideas on what we should do here? I am thinking that maybe the bootkube service check is optional in some way.

internal/pkg/cluster/check/default.go
```diff
@@ -24,7 +25,16 @@ func DefaultClusterChecks() []ClusterCheck {
...
+				err := ServiceStateAssertion(ctx, cluster, "bootkube", "Finished", "Skipped")
+				if err != nil {
+					if errors.Is(err, ErrServiceNotFound) {
+						return nil
+					}
+
+					return err
+				}
+
+				return nil
```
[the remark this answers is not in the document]
smira: makes sense :+1:

internal/integration/base/cli.go
```diff
@@ -38,22 +38,22 @@ func (cliSuite *CLISuite) DiscoverNodes() []string {
...
 
 	args = append([]string{"--talosconfig", cliSuite.TalosConfig}, args...)
 
 	return exec.Command(cliSuite.TalosctlPath, args...)
 }
 
-// RunOsctl runs talosctl binary with the options provided
-func (cliSuite *CLISuite) RunOsctl(args []string, options ...RunOption) {
-	Run(&cliSuite.Suite, cliSuite.buildOsctlCmd(args), options...)
+// RunCLI runs talosctl binary with the options provided
```
andrewrynhard, line 49: More generic naming so that _if_ we ever change the cli name again, we don't have to update everywhere.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -232,25 +235,34 @@ func create(ctx context.Context) (err error) {
...
 			if i == 0 {
 				cfg = configBundle.Init()
 			} else {
 				cfg = configBundle.ControlPlane()
 			}
 		} else {
-			cfg = configBundle.ControlPlaneCfg
+			// Any one of the control plane nodes can be the init node, so we use the
+			// init config's content, but change the type to a control plane.
+			configBundle.InitCfg.MachineConfig.MachineType = runtime.MachineTypeControlPlane.String()
```
smira, line 259: just a nit, but seems like this potentially might look better if we update MachineType for `cfg`, as here we multiple times update bundle, while other changes in the loop are no-op.

Or even we could have a configBundle method/param to return init bundle with control plane type

internal/pkg/provision/providers/docker/node.go
```diff
@@ -109,9 +109,8 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
 
 	// Mutate the container configurations based on the node type.
 
-	switch nodeReq.Config.Machine().Type() {
-	case runtime.MachineTypeInit:
-		portsToOpen := []string{"50000:50000/tcp", "6443:6443/tcp"}
+	if nodeReq.Config.Machine().Type() == runtime.MachineTypeControlPlane {
```
smira, line 112: should it be both MachineControlPlane and MachineTypeInit here to maintain backwards compatibility with init-node style?

andrewrynhard: :+1:

cmd/talosctl/cmd/mgmt/cluster/create.go
andrewrynhard: Good point! :thinking: Most of this is a hack until we can remove the init node type. I am tempted to go the easier path here (updating `cfg`) since will all hopefully go away sooner rather than later. But if we feel that we have a need to make the bundle generator more robust, I can do that.

smira: agreed, let's take shortcut, we probably don't need to support this for too long if we say that official way is to use bootstrap API

andrewrynhard: I found a better solution. Since control plane type will have to be what init type is today (it is almost the same even today), we can just generate the controlplane type with the init generator and then change the type string. This has the benefit that we can deprecate the init node config type at a later date and all controlplane configs should be compatible.

PR #2203, opened 2020-06-16
fix: allow for using /dev/disk/* symlinks

andrewrynhard:
This change ensures that any symlinks under `/dev/disks` are created
by udevd early enough so that users can specify disks identified
by these symlinks in `machine.disks` entries. This change also moves
the udevd-trigger from a service to a post-startup task for udevd.
This is a one time command that doesn't need to be a service. Some
other minor changes were made around parsing the partition name that
makes use of the symlinks created by udevd. This also ensures that
target mount points in `machine.disks` entries are created, fixing a
preexisting bug

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2203)


cmd/installer/pkg/install/manifest.go
```diff
@@ -176,19 +176,22 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
...
+	}
 
 	return nil
 }
 
 // Format creates a filesystem on the device/partition.
 func (t *Target) Format() error {
 	if t.Label == constants.BootPartitionLabel {
-		log.Printf("formatting partition %q - %q as %q\n", t.PartitionName, t.Label, "fat")
+		log.Printf("formatting partition %q as %q with label %q\n", t.PartitionName, "fat", t.Label)
```
andrewrynhard, line 190: Minor formatting changes that makes debugging easier.

cmd/installer/pkg/install/manifest.go
```diff
@@ -176,19 +176,22 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
...
 		return vfat.MakeFS(t.PartitionName, vfat.WithLabel(t.Label))
 	}
 
-	log.Printf("formatting partition %q - %q as %q\n", t.PartitionName, t.Label, "xfs")
+	log.Printf("formatting partition %q as %q with label %q\n", t.PartitionName, "xfs", t.Label)
 	opts := []xfs.Option{xfs.WithForce(t.Force)}
 
 	if t.Label != "" {
```
andrewrynhard, line 197: Minor formatting changes that makes debugging easier.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -575,6 +576,32 @@ func StartContainerd(seq runtime.Sequence, data interface{}) runtime.TaskExecuti
...
+
+		if err = system.WaitForService(system.StateEventUp, svc.ID(r)).Wait(ctx); err != nil {
+			return err
+		}
+
+		if _, err = cmd.Run("/sbin/udevadm", "trigger"); err != nil {
+			return err
+		}
+
+		if _, err = cmd.Run("/sbin/udevadm", "settle"); err != nil {
```
andrewrynhard, line 597: This ensures that `udevd` finishes processing kernel events triggered by `udevd trigger`. This prevents a race condition between `udevd` and Talos when a user wants to use something under `/dev/disk/*` in `machine.disks`.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -737,6 +765,28 @@ func partitionAndFormatDisks(logger *log.Logger, r runtime.Runtime) (err error)
...
+		pt, err = bd.PartitionTable(true)
+		if err != nil {
+			return err
+		}
+
+		if len(pt.Partitions()) > 0 {
+			logger.Printf(("skipping setup of %q, found existing partitions"), disk.Device)
+			continue
+		}
+
```
andrewrynhard, line 789: Improved detection of existing partitions, which is critical so that we don't repartition/reformat any extra disks.

freswa:
tested lgtm

PR #2215, opened 2020-06-17
fix: detect if partition table is missing

andrewrynhard:
This adds a sentinel error for a missing partition table. This error
is used to detect if a partition table already exists when setting
up user defined disks.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2215)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -767,7 +767,7 @@ func partitionAndFormatDisks(logger *log.Logger, r runtime.Runtime) (err error)
 	for _, disk := range r.Config().Machine().Disks() {
 		var bd *blockdevice.BlockDevice
 
-		bd, err = blockdevice.Open(disk.Device, blockdevice.WithNewGPT(false))
```
andrewrynhard, line 770: Default is to not create a GPT.

PR #2223, opened 2020-06-17
fix: delete manifests dir on bootkube failure

andrewrynhard:
Bootkube expects a clean manifests directory. This ensures that we
clean it up on a failure.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2223)

### DIFF
--- a/internal/app/bootkube/main.go
+++ b/internal/app/bootkube/main.go
@@ -52,6 +52,13 @@ func run() error {
 	}
 
 	defer func() {
+		// We want to cleanup the manifests directory only if bootkube fails.
+		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/app/bootkube/main.go
```diff
@@ -52,6 +52,13 @@ func run() error {
 	}
 
 	defer func() {
+		// We want to cleanup the manifests directory only if bootkube fails.
+		if err != nil {
+			if err = os.RemoveAll(constants.ManifestsDirectory); err != nil {
```
smira, line 57: I don't think this works as `err` is set on line 49 only, and `defer` block is later than `return err` on line 51

PR #2230, opened 2020-06-19
feat: implement circular buffer for system logs

smira:
This replaces logging to files with inotify following to pure in-memory
circular buffer which grows on demand capped at specified maximum
capacity.

The concern with previous approach was that logs on tmpfs were growing
without any bound potentially consuming all the node memory.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2230)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
@@ -26,7 +26,6 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/acpi"
 	"github.com/talos-systems/talos/internal/pkg/kmsg"
 	"github.com/talos-systems/talos/pkg/config"
-	"github.com/talos-systems/talos/pkg/constants"
 )
 
 // Controller represents the controller responsible for managing the execution
@@ -65,7 +64,7 @@ func NewController(b []byte) (*Controller, error) {
 	// TODO: this should be streaming capacity and probably some constant
 	e := NewEvents(1000)
 
-	l := logging.NewFileLoggingManager(constants.DefaultLogPath)
+	l := logging.NewCircularBufferLoggingManager()
 
 	ctlr := &Controller{
 		r: NewRuntime(cfg, s, e, l),
@@ -302,7 +301,14 @@ func (c *Controller) runTask(n int, f runtime.TaskSetupFunc, seq runtime.Sequenc
 
 	logger := &log.Logger{}
 
-	if err := kmsg.SetupLogger(logger, fmt.Sprintf("[talos] task %d:", n), true); err != nil {
+	machinedLog, err := c.Runtime().Logging().ServiceLog("machined").Writer()
+	if err != nil {
+		return err
+	}
+
+	defer machinedLog.Close() //nolint: errcheck
+
+	if err := kmsg.SetupLogger(logger, fmt.Sprintf("[talos] task %d:", n), machinedLog); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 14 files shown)


smira:
Sending this as draft, as I only have "follow" reader implemented, not regular one.

So `talosctl logs` works as if `-f` was given all the time.

smira:
```
ok  	github.com/talos-systems/talos/internal/pkg/circular	4.429s	coverage: 94.7% of statements
```

andrewrynhard:
Quick question... I still see machined logs:

```
[N] talosctl ls /run/system/log
.
machined.log
```
Is that intentional?

smira:
> Quick question... I still see machined logs:
> 
> ```
> [N] talosctl ls /run/system/log
> .
> machined.log
> ```
> 
> Is that intentional?

nice catch, I think I need to look into this one. It doesn't go via regular flow

internal/pkg/kmsg/kmsg.go
```diff
@@ -12,73 +12,37 @@ import (
...
-}
-
 // SetupLogger configures the logger to write to the kernel ring buffer via
 // /dev/kmsg.
-func SetupLogger(logger *log.Logger, prefix string, withLogFile bool) error {
+//
+// If logger is nil, default `log` logger is redirectred.
+//
+// If extraWriter is not nil, logs will be copied to it as well.
+func SetupLogger(logger *log.Logger, prefix string, extraWriter io.Writer) error {
```
smira, line 25: merged two methods together to reduce code duplication. 

now it's abstract enough (just accepts extraWriter), so it can deal with any logging subsystem

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -302,7 +301,14 @@ func (c *Controller) runTask(n int, f runtime.TaskSetupFunc, seq runtime.Sequenc
...
 
-	if err := kmsg.SetupLogger(logger, fmt.Sprintf("[talos] task %d:", n), true); err != nil {
+	machinedLog, err := c.Runtime().Logging().ServiceLog("machined").Writer()
+	if err != nil {
+		return err
+	}
+
+	defer machinedLog.Close() //nolint: errcheck
+
+	if err := kmsg.SetupLogger(logger, fmt.Sprintf("[talos] task %d:", n), machinedLog); err != nil {
```
smira, line 311: new kmsg logger setup with runtime logging interface (supports anything we pick up for the log implementation)

smira:
> > Quick question... I still see machined logs:
> > ```
> > [N] talosctl ls /run/system/log
> > .
> > machined.log
> > ```
> > 
> > 
> > Is that intentional?
> 
> nice catch, I think I need to look into this one. It doesn't go via regular flow

@andrewrynhard I fixed that by using common logging infrastructure, now it's consistent

internal/pkg/kmsg/kmsg.go
andrewrynhard: Not gonna block on this, but does is `log.Logger` an `io.Writer`? If so, then what do you think about making this a variadic function:

```
func SetupLogger(prefix string, loggers ...io.Writer) error
```

I probably am way off, still need to look into this function some more.

smira: there are two issues with that:

* default logger (`log.Printf`) doesn't expose itself as a value we can pass in
* what we do is that we configure `logger`, and make it write to both `kmsg` and passed in `extraWriter`. we can make `extraWriters` variadic, but we can't mix `logger` and `extraWriters` as they are used in different ways

what we could do instead is that we could split this function into two functions:

1. function which returns writer to `/dev/kmsg` in this package
2. in some other package function which sets up logging the way we want for a list of writers (which will include in runtime kmsg and logging framework writers)

andrewrynhard: The current implementation works. I will leave it up to you how for you wanna go with it. Something we could always circle back to.

PR #2231, opened 2020-06-23
fix: attempt to pull machine config from mounted disk in azure

rsmitty:
This PR fixes a bug caused by azure disabling their metadata endpoint
for some headscratching reason. The config data is mounted only at first
boot (from what I recall) as a cdrom device essentially. This adds
support for trying to hit the metadata endpoint and, if it doesn't work,
falling back to the cdrom.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2231)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
@@ -5,14 +5,20 @@
 package azure
 
 import (
+	"encoding/base64"
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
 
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
```diff
@@ -158,3 +189,56 @@ func (a *Azure) KernelArgs() procfs.Parameters {
...
+			}
+
+			ovfEnvFile, err := ioutil.ReadFile(filepath.Join(mnt, "ovf-env.xml"))
+			if err != nil {
+				// Device mount worked, but it wasn't the "CD" that contains the xml file
+				if os.IsNotExist(err) {
+					continue
+				}
+
+				return nil, fmt.Errorf("failed to read config: %s", err.Error())
```
smira, line 219: nit, 
```suggestion
				return nil, fmt.Errorf("failed to read config: %w", err)
```

PR #2236, opened 2020-06-24
chore: upgrade golangci-lint to 1.27

smira:
This version is built with newer Go which fixes
`fatal error: mlock failed` on Ubuntu kernels which are actually
patched, but their minor version doesn't match Go expectations.

New linters were disabled to minimize the changes, and plan is to fix
the linting errors in subsequent PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2236)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -112,6 +112,11 @@ linters:
     - godox
     - gocognit
     - gomnd
+    - nolintlint # TODO: fix & enable me
+    - testpackage # TODO: fix & enable me
+    - goerr113
+    - nestif
+    - godot # TODO: fix & enable me
   disable-all: false
   fast: false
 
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -112,6 +112,11 @@ linters:
     - godox
     - gocognit
     - gomnd
+    - nolintlint # TODO: fix & enable me
+    - testpackage # TODO: fix & enable me
+    - goerr113
+    - nestif
```
andrewrynhard, line 118: Are these useful?

smira: doesn't looks like that, or they're far away from our coding style.

`goerr113` prohibits `fmt.Errrof`

`nestif` seems to complain a lot around our code being too complicated

PR #2238, opened 2020-06-25
fix: make installer re-read partition table before formatting

smira:
This hopefully should fix errors like:

```
2020/06/25 18:23:22 attaching loopback device
2020/06/25 18:23:22 partitioning /dev/loop2 - ESP
2020/06/25 18:23:22 partitioning /dev/loop2 - EPHEMERAL
2020/06/25 18:23:22 formatting partition "/dev/loop2p1" as "fat" with label "ESP"
2020/06/25 18:23:22 detaching loopback device
2020/06/25 18:23:22 failed to format device: exit status 1: mkfs.vfat: can't open '/dev/loop2p1': No such file or directory
```

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2238)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,13 +139,13 @@ talosctl-%:
 talosctl: $(TALOSCTL_DEFAULT_TARGET) ## Builds the talosctl binary for the local machine.
 
 image-%: ## Builds the specified image. Valid options are aws, azure, digital-ocean, gcp, and vmware (e.g. image-aws)
-	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged autonomy/installer:$(TAG) image --platform $*
+	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged $(USERNAME)/installer:$(TAG) image --platform $*
 
 images: image-aws image-azure image-digital-ocean image-gcp image-vmware ## Builds all known images (AWS, Azure, Digital Ocean, GCP, and VMware).
 
 .PHONY: iso
 iso: ## Builds the ISO and outputs it to the artifact directory.
-	@docker run --rm -i -v $(PWD)/$(ARTIFACTS):/out autonomy/installer:$(TAG) iso
+	@docker run --rm -i -v $(PWD)/$(ARTIFACTS):/out $(USERNAME)/installer:$(TAG) iso
 
 .PHONY: boot
 boot: ## Creates a compressed tarball that includes vmlinuz and initramfs.xz. Note that these files must already be present in the artifacts directory.
--- a/cmd/installer/pkg/install/manifest.go
+++ b/cmd/installer/pkg/install/manifest.go
@@ -10,6 +10,8 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -139,13 +139,13 @@ talosctl-%:
...
 image-%: ## Builds the specified image. Valid options are aws, azure, digital-ocean, gcp, and vmware (e.g. image-aws)
-	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged autonomy/installer:$(TAG) image --platform $*
+	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged $(USERNAME)/installer:$(TAG) image --platform $*
 
 images: image-aws image-azure image-digital-ocean image-gcp image-vmware ## Builds all known images (AWS, Azure, Digital Ocean, GCP, and VMware).
 
 .PHONY: iso
 iso: ## Builds the ISO and outputs it to the artifact directory.
-	@docker run --rm -i -v $(PWD)/$(ARTIFACTS):/out autonomy/installer:$(TAG) iso
+	@docker run --rm -i -v $(PWD)/$(ARTIFACTS):/out $(USERNAME)/installer:$(TAG) iso
```
smira, line 148: this way I can do `make installer image-aws` to test the flow

cmd/installer/pkg/install/manifest.go
```diff
@@ -130,6 +130,10 @@ func (m *Manifest) ExecuteManifest() (err error) {
 			}
 		}
 
+		if err = bd.RereadPartitionTable(); err != nil {
+			return fmt.Errorf("failed to re-read partition table on %q: %w", dev, err)
+		}
+
```
andrewrynhard, line 136: There is an implicit reread (slightly different syscall) when adding a partition. This error, I think, is due to a race condition between parallel builds. They all share the same loopback pool.

smira: I checked it multiple times and saw the failure with no other builds running, that's why I thought it's something else. 

On other hand, it always fails straight after partitioning before formatting, so it should be something there

andrewrynhard: In that case, then yeah that is strange. Just for reference, here is the other syscall we use to inform the kernel of the newly added partition: https://github.com/talos-systems/talos/blob/master/pkg/blockdevice/table/gpt/gpt.go#L315-L317

smira: for example: https://ci.dev.talos-systems.io/talos-systems/talos/9619/1/18

all the failures seem to be exactly the same

smira: and yep, `InformKernelOfAdd` should have took care of that :shrug:

smira: do you think it might cause issues in real installs? I can make it ignore the error and just print the warning if it fails to re-read the partition table?

smira: updated failure to be a warning, so it should be harmless to add

smira:
build failed with the exactly same error I was trying to fix :(

PR #2244, opened 2020-06-29
chore: bring back tmp volume shared from e2e-docker to CAPI steps

smira:
CAPI-based steps are using docker Talos cluster built at `e2e-docker`
stage as a bootstrap cluster. Share the config via volume which is
attached to specific steps. Volume is temporary, but no longer resides
on `tmpfs`.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2244)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -450,6 +450,8 @@ steps:
     path: /root/.kube
   - name: dev
     path: /dev
+  - name: tmp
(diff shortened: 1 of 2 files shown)


smira:
e2e promotion: https://ci.dev.talos-systems.io/talos-systems/talos/9666

PR #2245, opened 2020-06-29
test: fix and improve reboot/reset tests

smira:
These tests rely on node uptime checks. These checks are quite flaky.

Following fixes were applied:

* code was refactored as common method shared between reset/reboot tests
(reboot all nodes does checks in a different way, so it wasn't updated)

* each request to read uptime times out in 5 seconds, so that checks
don't wait forever when node is down (or connection is aborted)

* to account for node availability vs. lower uptime in the beginning of
test, add extra elapsed time to the check condition

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2245)

### DIFF
--- a/internal/integration/api/reboot.go
+++ b/internal/integration/api/reboot.go
@@ -57,43 +57,9 @@ func (suite *RebootSuite) TestRebootNodeByNode() {
 	for _, node := range nodes {
 		suite.T().Log("rebooting node", node)
 
-		func(node string) {
-			// timeout for single node reboot
-			ctx, ctxCancel := context.WithTimeout(suite.ctx, 10*time.Minute)
-			defer ctxCancel()
-
-			nodeCtx := client.WithNodes(ctx, node)
-
-			// read uptime before reboot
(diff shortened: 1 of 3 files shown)


internal/integration/api/reboot.go
```diff
@@ -143,20 +114,27 @@ func (suite *RebootSuite) TestRebootAllNodes() {
...
+					uptimeAfter, err := suite.ReadUptime(requestCtx)
 					if err != nil {
 						// API might be unresponsive during reboot
 						return retry.ExpectedError(fmt.Errorf("error reading uptime for node %q: %w", node, err))
 					}
 
-					if uptimeAfter >= uptimeBefore {
+					// uptime of the node before it actually reboots still goes up linearly
+					// so we can safely add elapsed time here
+					if uptimeAfter >= uptimeBefore+elapsed {
```
andrewrynhard, line 135: :+1:

smira:
re-pushed with debugging `log.Printf` removed, no other changes

PR #2259, opened 2020-07-01
test: run integration pipeline nightly

smira:
Just a copy of `integration` pipeline with the same trigger as `nightly`
pipeline, so we can have two separate pipelines and two notifications
for better visibility.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2259)

### DIFF
--- a/hack/drone.jsonnet
+++ b/hack/drone.jsonnet
@@ -335,7 +335,17 @@ local integration_trigger = {
   },
 };
 
+local integration_nightly_trigger = {
+  trigger: {
+    cron: {
+      include: ['nightly'],
+    },
+  },
+};
+
+
 local integration_pipeline = Pipeline('integration', integration_steps) + integration_trigger;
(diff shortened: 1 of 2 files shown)


hack/drone.jsonnet
```diff
@@ -335,7 +335,17 @@ local integration_trigger = {
   },
 };
 
+local integration_nightly_trigger = {
```
smira, line 338: unfortunately trigger expressions only support `and` conditions, so we can't make pipeline run both as promote target and cron

andrewrynhard: Ahh that's right. :(

smira:
> Will this run both nightlies at the same time?

I guess so, do you think that might be a problem?

andrewrynhard:
> > Will this run both nightlies at the same time?
> 
> I guess so, do you think that might be a problem?

Well, hopefully we should be able to tolerate more than one build running at a time, but I have seen our CI struggle to do so. I plan on fixing that soon, so we can deal with it if it is a problem.

PR #2260, opened 2020-07-02
feat: upgrade runc to v1.0.0-rc90

andrewrynhard:
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2260)


cmd/talosctl/cmd/talos/list.go
```diff
@@ -45,7 +45,7 @@ var lsCmd = &cobra.Command{
 				rootDir = args[0]
 			}
 
-			stream, err := c.LS(ctx, machineapi.ListRequest{
+			stream, err := c.LS(ctx, &machineapi.ListRequest{
```
andrewrynhard, line 48: Linter complained about a context being passed by value that was part of the `ListRequest`. Passing by reference fixes this.

internal/app/apid/pkg/backend/apid.go
```diff
@@ -157,7 +165,7 @@ func (a *APID) AppendInfo(streaming bool, resp []byte) ([]byte, error) {
 	// cut off embedded message header
 	resp = resp[n1+n2:]
 	// build new embedded message header
-	prefix := append(proto.EncodeVarint((metadataField<<3)|metadataType), proto.EncodeVarint(uint64(len(resp)+len(payload)))...)
+	prefix := append(protowire.AppendVarint(nil, (metadataField<<3)|metadataType), protowire.AppendVarint(nil, uint64(len(resp)+len(payload)))...)
```
andrewrynhard, line 168: The `DecodeVarint` and `EncodeVarint` are not implemented anywhere in `google.golang.org/protobuf`.

smira: just a nit, I guess we could do the following: 
```suggestion
	prefix := protowire.AppendVarint(protowire.AppendVarint(nil, (metadataField<<3)|metadataType)), uint64(len(resp)+len(payload)))
```
```

PR #2265, opened 2020-07-02
feat: implement service events

smira:
This implements service events, adds test for events API based on
service events as they're the easiest to generate on demand.

Disabled validate test for 'metal' as it validates disk device against
local system which doesn't make much sense.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2265)


internal/app/machined/pkg/system/events/events.go
```diff
@@ -60,6 +60,15 @@ type ServiceEvent struct {
 	Timestamp time.Time
 }
 
+// AsEventProto returns protobuf representation of respective machined evet.
+func (event *ServiceEvent) AsEventProto(service string) *machineapi.ServiceStateEvent {
```
andrewrynhard, line 64: I really like this. We should have an abstracted type that embeds `proto.Message` and augments it with an interface that requires every event to implement something like this.

```
t.AsProto()
```

Not sure what it would take, but that would clean things up a lot.

smira: misprint there, and probably I need to change that... it's a bit confusing, as we have `ServiceEvent` which is used in services API and `ServiceStateEvent` in events API, but essentially same thing

andrewrynhard: ahh I see. Think I got excited too soon :D ... this is useful for when we have internal types, and we want to convert to event types.

PR #2274, opened 2020-07-06
feat: implement API access to event history

smira:
1. Add [xid-based](https://github.com/rs/xid) event IDs. Xids
are sortable and unique enough. Xids also encode event publishing
time with a second precision.

2. Add three ways to look back into event history: based on number of
events, on time and ID. Lookup via ID might be used to restart event
polling in case of broken API connection from the same moment.

3. Reimplement core event buffer with positions which are always
incremented instead of generation+index, this implementation is much
more simple (idea from circular buffer).

4. By default, Events API works the same - it shows no history and
starts streaming new events only.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2274)


pkg/client/client.go
```diff
@@ -753,93 +751,3 @@ func ReadStream(stream MachineStream) (io.ReadCloser, <-chan error, error) {
 
 	return pr, errCh, stream.CloseSend()
 }
-
-// Events implements the proto.OSClient interface.
-func (c *Client) Events(ctx context.Context) (stream machineapi.MachineService_EventsClient, err error) {
```
smira, line 758: moved to `events.go`, as there're too many functions related to Events API

internal/app/machined/pkg/runtime/events.go
```diff
@@ -14,15 +18,74 @@ import (
...
 type WatchFunc func(<-chan Event)
 
+// WatchOptions defines options for the watch call.
+//
+// Only one of TailEvents, TailID or TailDuration should be non-zero.
+type WatchOptions struct {
+	// Return that many past events.
+	//
+	// If TailEvents is negative, return all the events available.
+	TailEvents int
```
smira, line 35: any suggestions on better names?

andrewrynhard: I always like super descriptive names for public facing things. This works. Just throwing out more ideas since you asked:

```
NumberOfHistoricalEvents
EventsCount
EventsInHistory
```

cmd/talosctl/cmd/talos/events.go
```diff
@@ -9,13 +9,20 @@ import (
...
 
 	"github.com/talos-systems/talos/api/machine"
 	"github.com/talos-systems/talos/pkg/client"
 )
 
+var eventsCmdFlags struct {
+	tailEvents   int32
+	tailDuration time.Duration
+	tailID       string
+}
```
andrewrynhard, line 24: I like this. We need more of this. I hate that we have so many global variables.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
```diff
@@ -45,48 +43,56 @@ import (
...
+// Maximum available event history is (cap-gap).
+func NewEvents(cap, gap int) *Events {
 	e := &Events{
 		stream: make([]runtime.Event, cap),
 		cap:    cap,
+		gap:    gap,
+	}
+
+	if gap >= cap {
+		panic("NewEvents: gap >= cap")
```
andrewrynhard, line 95: Do we want to panic here?

cmd/talosctl/cmd/talos/events.go
```diff
@@ -24,7 +31,21 @@ var eventsCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		return WithClient(func(ctx context.Context, c *client.Client) error {
 			w := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
-			fmt.Fprintln(w, "NODE\tEVENT\tMESSAGE")
+			fmt.Fprintln(w, "NODE\tID\tEVENT\tOBJECT\tMESSAGE")
```
andrewrynhard, line 34: I can't figure out what `OBJECT` is here.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
smira: I thought we should never get to this condition, as `Events` is internal and not configurable directly, but if `gap >= cap`, things will go completely nuts in this file, so I thought there should be some protection.

cmd/talosctl/cmd/talos/events.go
smira: not sure if `OBJECT` is the best name, but looks like all the events have some kind of name/identitifer:

* phase/sequence/task name
* service name

So we could have provided some filtering in the future on it.

cmd/talosctl/cmd/talos/events.go
smira: yep, I try to do that at least for new commands 

the worst thing is that sometimes such variable is shared across commands :confused:

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_events.go
andrewrynhard: :+1:

cmd/talosctl/cmd/talos/events.go
andrewrynhard: I thought that that was what `EVENT` is. I believe its the name.

smira: `EVENT` right now is event type ID. 

Example output with this PR:

```
 talosctl events --tail=15
NODE         EVENT                                     ID                     OBJECT     MESSAGE
172.20.0.1   talos/runtime/machine.ServiceStateEvent   bs1o9kq2qblc738ovma0   bootkube   PREPARING: Running pre state
172.20.0.1   talos/runtime/machine.ServiceStateEvent   bs1o9kq2qblc738ovmag   bootkube   PREPARING: Creating service runner
172.20.0.1   talos/runtime/machine.ServiceStateEvent   bs1o9kq2qblc738ovmb0   bootkube   RUNNING: Started task bootkube (PID 2231) for container bootkube
172.20.0.1   talos/runtime/machine.SequenceEvent   bs1o9uq2qblc738ovmgg   boot   STOP
172.20.0.1   talos/runtime/machine.ServiceStateEvent   bs1oa822qblc738ovmh0   bootkube   FINISHED: Service finished successfully
```

andrewrynhard: Ahh ok that makes sense. Pretty cool. What about `SOURCE`?

smira: makes perfect sense to me, I will update!

PR #2281, opened 2020-07-07
test: update init node check in reset API tests

### LINKED ISSUE #2277: enable reset API test on init node
## Feature Request

As we moved towards using bootstrap API in the tests, we no longer need to keep the notion of `init` node, so we should be able to `reset()` all the nodes.

### Description

Remove the check which skips init node.

smira:
Previously we assumed that node 0 is the init node, and it can't be
reset. With new bootstrap API approach, there's no init node, and all
the nodes can be reset. This corrects the check to skip only the init
node, and with bootstrap API there's no init node (so no nodes are
skipped).

As part of test debugging following updates were done:

* context propagation/timeouts in kubernetes and etcd client

* fixed check for control plane components to check for number of pods
as when controller-manager & schduler are not running, daemonset stats
are not updated

* fixed bootkube start sequence to work always by implementing 'Unload'
method for services

Fixes #2277

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2281)

### DIFF
--- a/internal/app/machined/pkg/system/services/apid.go
+++ b/internal/app/machined/pkg/system/services/apid.go
@@ -81,13 +81,16 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 	if r.Config().Machine().Type() == runtime.MachineTypeJoin {
 		opts := []retry.Option{retry.WithUnits(3 * time.Second), retry.WithJitter(time.Second)}
 
-		err := retry.Constant(10*time.Minute, opts...).Retry(func() error {
+		err := retry.Constant(4*time.Minute, opts...).Retry(func() error {
+			ctx, ctxCancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer ctxCancel()
+
 			h, err := kubernetes.NewClientFromKubeletKubeconfig()
 			if err != nil {
 				return retry.ExpectedError(fmt.Errorf("failed to create client: %w", err))
 			}
 
-			endpoints, err = h.MasterIPs()
+			endpoints, err = h.MasterIPs(ctx)
 			if err != nil {
 				return retry.ExpectedError(err)
 			}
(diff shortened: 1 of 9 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -269,28 +269,31 @@ func addMember(r runtime.Runtime, addrs []string, name string) (*clientv3.Member
...
+func buildInitialCluster(ctx context.Context, r runtime.Runtime, name, ip string) (initial string, err error) {
 	err = retry.Constant(10*time.Minute, retry.WithUnits(3*time.Second), retry.WithJitter(time.Second)).Retry(func() error {
 		var (
 			peerAddrs = []string{"https://" + net.FormatAddress(ip) + ":2380"}
 			resp      *clientv3.MemberListResponse
 			id        uint64
 		)
 
-		resp, id, err = addMember(r, peerAddrs, name)
+		attemptCtx, attemptCtxCancel := context.WithTimeout(ctx, 30*time.Second)
```
smira, line 293: added context propagation and limiting each attempt duration to make sure we do more attempts during those 10 minutes

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -81,13 +81,16 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 	if r.Config().Machine().Type() == runtime.MachineTypeJoin {
 		opts := []retry.Option{retry.WithUnits(3 * time.Second), retry.WithJitter(time.Second)}
 
-		err := retry.Constant(10*time.Minute, opts...).Retry(func() error {
+		err := retry.Constant(4*time.Minute, opts...).Retry(func() error {
```
smira, line 84: adjusted retry limit to be less than 5 minute default timeout

pkg/kubernetes/kubernetes.go
```diff
@@ -48,6 +48,10 @@ func NewClientFromKubeletKubeconfig() (client *Client, err error) {
 		return nil, err
 	}
 
+	if config.Timeout == 0 {
+		config.Timeout = 30 * time.Second
```
smira, line 52: enforce default timeout to be on the safe side

smira:
I'll split the PR into smaller PRs.

PR #2283, opened 2020-07-08
fix: update the control plane cluster health check

smira:
Include kube-apiserver in the list of daemon sets to be checked, and
for each daemon set verify number of pods running and ready, as when
control plane is damaged daemon set properties are not updated properly.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2283)


smira:
:facepalm: pushed to the wrong branch, I'll get it fixed

PR #2287, opened 2020-07-08
test: update init node check in reset API tests

### LINKED ISSUE #2277: enable reset API test on init node

smira:
Previously we assumed that node 0 is the init node, and it can't be
reset. With new bootstrap API approach, there's no init node, and all
the nodes can be reset. This corrects the check to skip only the init
node, and with bootstrap API there's no init node (so no nodes are
skipped).

Fixes #2277

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2287)

### DIFF
--- a/internal/integration/api/reset.go
+++ b/internal/integration/api/reset.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/internal/integration/base"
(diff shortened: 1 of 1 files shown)


smira:
passed integration test: https://ci.dev.talos-systems.io/talos-systems/talos/9945

PR #2291, opened 2020-07-09
feat: uncordon nodes automatically on boot

smira:
Talos will mark node as schedulable if it was previously cordoned by
Talos (for upgrade, reset, etc.)

If user marked node as not schedulable, Talos won't change it on boot.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2291)


smira:
Based on #1991

PR #2292, opened 2020-07-09
docs: update firecracker for new home of tc-redirect-tap plugin

smira:
See https://github.com/firecracker-microvm/firecracker-go-sdk/issues/174#issuecomment-655798205

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2292)


smira:
added a section on `make kernel initrafms`, as by default kernel is not built

docs/website/content/v0.6/en/guides/local/firecracker.md
```diff
@@ -75,15 +75,25 @@ sudo cp cni-plugins-linux/{bridge,firewall,static} /opt/cni/bin
...
+> Note: if `$GOPATH` is not set, it defaults to `~/go`.
+
+## Build Talos kernel and initramfs
+
+Firecracker provisioner depends on Talos kernel and initramfs:
+
+```bash
+make kernel initramfs
+```
+
```
andrewrynhard, line 96: An alternative is that they could also download these from a release.

smira: good point, let me add that, guide doesn't assume you have to be a developer, thanks!

PR #2296, opened 2020-07-10
fix: improve node uncordon tasks

smira:
1. Increase retry timeout.

2. Use timeout per attempt.

3. Check for node readiness as a gate to succeed.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2296)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -1123,6 +1123,10 @@ func UncordonNode(seq runtime.Sequence, data interface{}) runtime.TaskExecutionF
 			return err
 		}
 
+		if err = kubeHelper.WaitUntilReady(hostname); err != nil {
+			return err
+		}
+
 		if err = kubeHelper.Uncordon(hostname); err != nil {
 			return err
 		}
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -202,6 +202,33 @@ func (h *Client) LabelNodeAsMaster(name string) (err error) {
 	return nil
 }
 
+// WaitUntilReady waits for a node to be ready.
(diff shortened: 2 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1123,6 +1123,10 @@ func UncordonNode(seq runtime.Sequence, data interface{}) runtime.TaskExecutionF
 			return err
 		}
 
+		if err = kubeHelper.WaitUntilReady(hostname); err != nil {
+			return err
+		}
+
```
andrewrynhard, line 1129: Good idea. We could use this in the future for other use cases.

pkg/kubernetes/kubernetes.go
```diff
@@ -202,6 +202,29 @@ func (h *Client) LabelNodeAsMaster(name string) (err error) {
...
 
+// WaitUntilReady waits for a node to be ready.
+func (h *Client) WaitUntilReady(name string) error {
+	return retry.Exponential(3*time.Minute, retry.WithUnits(250*time.Millisecond), retry.WithJitter(50*time.Millisecond)).Retry(func() error {
+		attemptCtx, attemptCtxCancel := context.WithTimeout(context.TODO(), 30*time.Second)
+		defer attemptCtxCancel()
+
+		node, err := h.CoreV1().Nodes().Get(attemptCtx, name, metav1.GetOptions{})
+		if err != nil {
+			return retry.UnexpectedError(err)
```
andrewrynhard, line 213: I think we need to check if the error is a "Not found" (there is a Kubernetes function for this somewhere) error and make that expected. This was the original error we saw in CI:

```
2020/07/10 00:09:09 error running phase 9 in boot sequence: task 1/1: failed, failed to uncordon node e2e-docker-worker-1: 1 error(s) occurred:
	nodes "e2e-docker-worker-1" not found
```

PR #2315, opened 2020-07-16
chore: update golangci-lint to 1.28.3

### LINKED ISSUE #2272: upgrade golangci-lint to 1.28.3
golangci-lint now has integrated gofumpt, so we can get rid of extra invocation.

smira:
Fixes #2272

`gofumpt` is now included into `golangci-lint`, but not the
`gofumports`, so we keep it using it as separate binary, but we keep
versions in sync with `golangci-lint`.

This contains fixes from:

* `gofumpt` (automated, mostly around octal constants)
* `exhaustive` in `switch` statements
* `noctx` (adding context with default timeout to http requests)

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2315)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
@@ -5,12 +5,14 @@
 package gcp
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
+	"time"
 
 	"github.com/talos-systems/go-procfs/procfs"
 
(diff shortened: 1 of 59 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
```diff
@@ -60,7 +62,10 @@ func (g *GCP) ExternalIPs() (addrs []net.IP, err error) {
 		resp *http.Response
 	)
 
-	if req, err = http.NewRequest("GET", GCExternalIPEndpoint, nil); err != nil {
+	ctx, ctxCancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer ctxCancel()
+
+	if req, err = http.NewRequestWithContext(ctx, "GET", GCExternalIPEndpoint, nil); err != nil {
```
andrewrynhard, line 68: nit: There is `http.MethodGet` that we could use instead of `"GET"`.

smira: yep, I used that it many places, not sure how I missed this one!

proper fix is to add context to all those platform methods, but this should come as another PR

PR #2317, opened 2020-07-16
chore: update module dependencies

### LINKED ISSUE #2316: update go.mod dependencies
## Feature Request

There are many libs we depend upon which are not the libraries we track versions precisely (e.g. etcd/Kubernetes), but we could have upgraded those to the latest releases.

### Description

Keeping up with upstream changes, potential fixes.

smira:
Fixes #2316

Simply update dependencies we don't track on version level to be
compatible with Talos components (like etcd or k8s).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2317)


smira:
updated even more dependencies, and fixed a bug in tests: now `TearDownTest()` is called even when `SetupTest()` exits (e.g. with `SkipNow()`).

PR #2327, opened 2020-07-20
feat: force nodes to be set in `talosctl` commands using the API

smira:
With load-balancing enabled by default running `talosctl` without
`--nodes` is risky, as it might hit any control plane by default without
`--nodes`.

Only two commands do not enforce this check, as they do their own node
contexts: `crashdump` and `health` (client-side).

Integration tests were updated to always supply `--nodes` cli argument,
while doing that I refactored the storage for discovered nodes to use
existing `cluster.Info` interface.

The downside is that with e2e CAPI tests CLI tests will be mostly
skipped as we don't support discovery in CLI tests at the momemnt. This
can be fixed by using `talosctl kubeconfig` + `kubectl get nodes` for
node discovery.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2327)

### DIFF
--- a/cmd/talosctl/cmd/talos/crashdump.go
+++ b/cmd/talosctl/cmd/talos/crashdump.go
@@ -25,7 +25,7 @@ var crashdumpCmd = &cobra.Command{
 	Long:  ``,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return WithClient(func(ctx context.Context, c *client.Client) error {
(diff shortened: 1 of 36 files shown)


smira:
:facepalm: I need to fix the scripts

smira:
e2e-gcp: passed

e2e-aws: failed, but I believe failure is not related: https://ci.dev.talos-systems.io/talos-systems/talos/10107/1/28

PR #2331, opened 2020-07-20
feat: make partitions on additional disk without size occupy full disk

### LINKED ISSUE #2214: Allow to use full disks without explicitly giving the size
## Feature Request
```
machine:
  disks:
    - device: /dev/disk/by-id/scsi-0HC_Volume_5878184
      partitions:
        - mountpoint: /var/data
          fullDisk: true
```
Given this option should probably lead to `partition.WithMaximumSize(true)` being passed to `pt.Add()` - see manifest.go:164 ff

It results in the partition using the maximum available size.

smira:
Fixes #2214

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2331)

### DIFF
--- a/cmd/installer/pkg/install/manifest.go
+++ b/cmd/installer/pkg/install/manifest.go
@@ -188,6 +188,10 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
 	default:
 		typeID := "AF3DC60F-8384-7247-8E79-3D69D8477DE4"
 		opts = append(opts, partition.WithPartitionType(typeID))
+
+		if t.Size == 0 {
+			opts = append(opts, partition.WithMaximumSize(true))
+		}
 	}
 
 	part, err := pt.Add(uint64(t.Size), opts...)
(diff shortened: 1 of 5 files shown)


cmd/installer/pkg/install/manifest.go
```diff
@@ -188,6 +188,10 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
 	default:
 		typeID := "AF3DC60F-8384-7247-8E79-3D69D8477DE4"
 		opts = append(opts, partition.WithPartitionType(typeID))
+
+		if t.Size == 0 {
+			opts = append(opts, partition.WithMaximumSize(true))
+		}
```
andrewrynhard, line 194: I'm not sure how to best enforce/check, but I was thinking that this option makes sense only if it is the last partition. What happens if a user specifies 2 partitions, and omits `size` for both?

smira: probably two sides of that:

1. make sure partitioning code can handle that (return meaningful error that no space is available)
2. config validation to make sure only the last partition can have size of zero

andrewrynhard: Makes sense. We can leave this as-is, and make it a lesser known feature for 0.6. Up to you.

smira: added both validation checks

PR #2333, opened 2020-07-20
fix: log messages properly when sequence/phase/task fails

smira:
Previously log said `done` for the failed items which seemed not obvious
when combined with the error message later on.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2333)


smira:
This might lead to too many error messages repeated, maybe we should just say "failed" and omit the error message?

andrewrynhard:
> This might lead to too many error messages repeated, maybe we should just say "failed" and omit the error message?

Works for me.

smira:
> This might lead to too many error messages repeated, maybe we should just say "failed" and omit the error message?

re-pushed without extra error messages for the log

PR #2334, opened 2020-07-21
feat: qemu provisioner

Unix4ever:
Starts and stops qemu VMs, has some initial configuration subset.
Sets up networking through CNI tools, sets up DHCP server which gives IP
addresses to nodes.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2334)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -41,6 +41,7 @@ var (
 	registryMirrors         []string
 	kubernetesVersion       string
 	nodeVmlinuxPath         string
+	nodeVmlinuzPath         string
 	nodeInitramfsPath       string
 	bootloaderEmulation     bool
 	configDebug             bool
@@ -153,9 +154,10 @@ func create(ctx context.Context) (err error) {
 			},
 		},
 
-		Image:         nodeImage,
-		KernelPath:    nodeVmlinuxPath,
-		InitramfsPath: nodeInitramfsPath,
+		Image:                  nodeImage,
+		UncompressedKernelPath: nodeVmlinuxPath,
+		CompressedKernelPath:   nodeVmlinuzPath,
+		InitramfsPath:          nodeInitramfsPath,
 
 		SelfExecutable: os.Args[0],
 		StateDirectory: stateDir,
@@ -412,6 +414,7 @@ func init() {
 	createCmd.Flags().StringVar(&nodeImage, "image", helpers.DefaultImage(constants.DefaultTalosImageRepository), "the image to use")
 	createCmd.Flags().StringVar(&nodeInstallImage, "install-image", helpers.DefaultImage(constants.DefaultInstallerImageRepository), "the installer image to use")
 	createCmd.Flags().StringVar(&nodeVmlinuxPath, "vmlinux-path", helpers.ArtifactPath(constants.KernelUncompressedAsset), "the uncompressed kernel image to use")
+	createCmd.Flags().StringVar(&nodeVmlinuzPath, "vmlinuz-path", helpers.ArtifactPath(constants.KernelAsset), "the compressed kernel image to use")
 	createCmd.Flags().StringVar(&nodeInitramfsPath, "initrd-path", helpers.ArtifactPath(constants.InitramfsAsset), "the uncompressed kernel image to use")
 	createCmd.Flags().BoolVar(&bootloaderEmulation, "with-bootloader-emulation", false, "enable bootloader emulation to load kernel and initramfs from disk image")
 	createCmd.Flags().StringSliceVar(&registryMirrors, "registry-mirror", []string{}, "list of registry mirrors to use in format: <registry host>=<mirror URL>")
--- a/hack/test/e2e-qemu.sh
+++ b/hack/test/e2e-qemu.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -eou pipefail
+
+source ./hack/test/e2e.sh
+
+PROVISIONER=qemu
+CLUSTER_NAME=e2e-${PROVISIONER}
+
+case "${CI:-false}" in
+  true)
+    REGISTRY="127.0.0.1:5000"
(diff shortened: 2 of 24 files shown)


internal/pkg/provision/providers/qemu/launch.go
```diff
@@ -0,0 +1,86 @@
...
+func Launch() error {
+	var config LaunchConfig
+
+	if err := vm.ReadConfig(&config); err != nil {
+		return err
+	}
+
+	c := vm.ConfigureSignals()
+
+	cmd := exec.Command(
```
smira, line 47: :+1:

internal/pkg/provision/providers/qemu/launch.go
```diff
@@ -0,0 +1,86 @@
...
+		config.QemuExecutable,
+		"-m", strconv.FormatInt(config.MemSize, 10),
+		"-drive", fmt.Sprintf("format=raw,file=%s", config.DiskPath),
+		"-smp", fmt.Sprintf("cpus=%d", config.VCPUCount),
+		"-accel",
+		"kvm",
+		"-nographic",
```
smira, line 54: probably we need `-kernel`, `-initrd` and `-append` for the first boot as disk image is empty

also not 100% sure how to do that, but we need to tell qemu to map in disk as virtio device (should be faster and it will have the same name `/dev/vda` as firecracker)

smira: looks like `,if=virtio` should be added to the `-drive` line

internal/pkg/provision/providers/qemu/qemu.go
```diff
@@ -0,0 +1,75 @@
...
+			"console=ttyS0",
+			// reboot configuration
+			"reboot=k",
+			"panic=1",
+			// disable stuff we don't need
+			"pci=off",
+			"acpi=off",
+			"i8042.noaux=",
+			// Talos-specific
+			"talos.platform=metal",
```
smira, line 56: this is something we'll fix later, but with firecracker we had each node loading its own config by kernel command line args, with qemu we probably need a bit of a different way, e.g. we could deliver different config from http server based on node IP so that kernel arg is same for all the nodes (otherwise need a different config per node, which is also doable)

but that is a thought for the future

Unix4ever: yeah, I guess we can think about it when we get network in place.

internal/pkg/provision/providers/qemu/launch.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read disk file %w", err)
+	}
+
+	if bytes.Equal(buf, make([]byte, checkSize)) {
+		args = append(args,
+			"-kernel", config.KernelImagePath,
+			"-initrd", config.InitrdPath,
+			"-append", config.KernelArgs,
+		)
```
smira, line 80: we might want to add `-no-reboot` here so that `qemu` runs with our kernel until the reboot (as installer should have run and installed the Talos, so next reboot it should go with the bootloader)

smira: if we add `-no-reboot`, we might also need to make a big `for` loop here like with firecracker so that qemu is restarted if it exits cleanly

Unix4ever: gotcha, will make a change in a moment

internal/pkg/cniutils/cniutils.go
```diff
@@ -0,0 +1,74 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package cniutils provides helper functions to parse CNI results.
+package cniutils
+
+import (
+	"fmt"
+
+	"github.com/containernetworking/cni/pkg/types/current"
+)
```
smira, line 12: TODO: add copyright header here

Unix4ever: which one?

hack/test/e2e-qemu.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
smira, line 1: we probably want to update this with the latest fixes we have in `e2e-firecracker.sh` (for the registry at least)...

internal/pkg/cniutils/cniutils.go
smira: I took the code here: https://github.com/firecracker-microvm/firecracker-go-sdk/blob/master/cni/internal/cniutil.go

As it's internal package, I can't import it. So we might need to add copyright notice somewhere after `import` so we don't break the license rules for the original code.

Unix4ever: ahh, gotcha.

smira:
Pushed simple fixes:

* `make docs` to generate CLI docs
* fixed `make talosctl-darwin` (qemu can't be built on OS X because of CNI dependency with network namespaces)

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -412,6 +414,7 @@ func init() {
 	createCmd.Flags().StringVar(&nodeImage, "image", helpers.DefaultImage(constants.DefaultTalosImageRepository), "the image to use")
 	createCmd.Flags().StringVar(&nodeInstallImage, "install-image", helpers.DefaultImage(constants.DefaultInstallerImageRepository), "the installer image to use")
 	createCmd.Flags().StringVar(&nodeVmlinuxPath, "vmlinux-path", helpers.ArtifactPath(constants.KernelUncompressedAsset), "the uncompressed kernel image to use")
+	createCmd.Flags().StringVar(&nodeVmlinuzPath, "vmlinuz-path", helpers.ArtifactPath(constants.KernelAsset), "the compressed kernel image to use")
```
andrewrynhard, line 417: We might be able to collapse these into a single "`--kernel-`" flag in the future.

smira: yep, in fact we could do that in follow-up PR, we can uncompress vmlinuz for firecracker now

PR #2342, opened 2020-07-22
fix: wait for bootkube to finish before trying to uncordon the node

smira:
This fixes potential race issue when temporary control plane is torn
down by bootkube and `uncordonNode` task is running.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2342)


smira:
We need a better solution here.

PR #2347, opened 2020-07-23
chore: use Kubernetes pipelines

andrewrynhard:
# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2347)


Dockerfile
```diff
@@ -482,9 +482,9 @@ RUN prototool lint --protoc-bin-path=/toolchain/bin/protoc --protoc-wkt-path=/to
 
 # The markdownlint target performs linting on Markdown files.
 
-FROM node:8.16.1-alpine AS lint-markdown
-RUN npm i -g markdownlint-cli@0.22.0
-RUN npm i -g textlint@11.6.3
+FROM node:14.5.0-alpine AS lint-markdown
+RUN npm i -g markdownlint-cli@0.23.2
+RUN npm i -g textlint@11.7.6
```
andrewrynhard, line 487: For some reason, I was seeing some errors that we never saw in the old setup. Had to bump these to address them.

Makefile
```diff
@@ -42,6 +42,8 @@ COMMON_ARGS += --build-arg=USERNAME=$(USERNAME)
 COMMON_ARGS += --build-arg=http_proxy=$(http_proxy)
 COMMON_ARGS += --build-arg=https_proxy=$(https_proxy)
 
+CI_ARGS ?=
+
```
andrewrynhard, line 46: Just in case we need to pass an extra args when running in CI.

Makefile
```diff
@@ -168,7 +171,7 @@ lint: ## Runs linters on go, protobuf, and markdown file types.
 	@$(MAKE) lint-go lint-protobuf lint-markdown
 
 check-dirty: ## Verifies that source tree is not dirty
-	@case "$(TAG)" in *-dirty) echo "Source tree is dirty"; exit 1 ;; esac
+	@case "$(TAG)" in *-dirty) echo "Source tree is dirty"; git status; exit 1 ;; esac
```
andrewrynhard, line 174: Found this useful when I was debugging why the tree was dirty.

hack/test/e2e-docker.sh
```diff
@@ -25,15 +16,19 @@ function create_cluster {
...
     --with-init-node=false \
     --crashdump
 
   "${TALOSCTL}" config node 10.5.0.2
 }
 
+function destroy_cluster() {
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}"
+}
+
```
andrewrynhard, line 28: This was required due to some port binding collisions, but I think we can/will fix that outside of this, but still good to destroy so that we can test it.

hack/test/e2e-firecracker.sh
```diff
@@ -7,9 +7,10 @@ source ./hack/test/e2e.sh
 PROVISIONER=firecracker
 CLUSTER_NAME=e2e-${PROVISIONER}
 
-case "${REGISTRY:-false}" in
-  registry.ci.svc:5000)
-    REGISTRY_ADDR=`python -c "import socket; print socket.gethostbyname('registry.ci.svc')"`
+case "${CI:-false}" in
```
andrewrynhard, line 10: Simplify things a bit with less environment variables. Just use the `CI` env var instead since we know that we have a regsitry on `127.0.0.1`.

hack/test/e2e-firecracker.sh
```diff
@@ -7,11 +7,14 @@ source ./hack/test/e2e.sh
...
+  true)
+    REGISTRY="127.0.0.1:5000"
+    REGISTRY_ADDR=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' registry`
     FIRECRACKER_FLAGS="--registry-mirror ${REGISTRY}=http://${REGISTRY_ADDR}:5000 --with-bootloader-emulation"
     INSTALLER_TAG="${TAG}"
+    docker tag ${IMAGE} 127.0.0.1:5000/autonomy/talos:"${TAG}"
+    docker push 127.0.0.1:5000/autonomy/installer:"${TAG}"
```
andrewrynhard, line 17: Allows us to avoid having to have "local" builds.

PR #2368, opened 2020-07-28
refactor: make `pkg/config` not rely on `machined/../internal/runtime`

smira:
This makes `pkg/config` directly importable from other projects.

There should be no functional changes.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2368)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/validate.go
+++ b/cmd/talosctl/cmd/mgmt/validate.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/cli"
-	"github.com/talos-systems/talos/pkg/config"
+	"github.com/talos-systems/talos/pkg/config/configloader"
 )
 
 var (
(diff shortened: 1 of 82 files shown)


pkg/config/types.go
```diff
@@ -2,110 +2,17 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
```
andrewrynhard, line 2: Should we move the remaining types into `provider.go`?

pkg/config/types/v1alpha1/v1alpha1_configurator.go
```diff
@@ -11,13 +11,14 @@ import (
 	"strings"
```
andrewrynhard, line 11: Should we rename this file to `v1alpha1_provider.go`?

pkg/config/types.go
smira: probably yep, I split them first by `interface`/`struct`, but it makes little sense, they are both kind of interface

cmd/talosctl/cmd/mgmt/validate.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/cli"
-	"github.com/talos-systems/talos/pkg/config"
+	"github.com/talos-systems/talos/pkg/config/configloader"
```
andrewrynhard, line 14: super duper nit: Since this is already namespaced with "`config`", it might be better to call this package just "`loader`", for the same reasons that (some linter I can't remember) doesn't allow types to have the same name as the package name (e.g. `thing.Thing`).

PR #2374, opened 2020-07-29
fix: update status when adjusting the time

smira:
Status big `STA_UNSYNC` should be cleared otherwise kernel assumes that
time is not sync. `STA_PLL` is set to notify kernel that we apply offset
adjustments periodically.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2374)


smira:
with the fix:

```
172.20.0.2: 2020/07/29 17:20:06.578174 ntp.go:152: adjusting time by 8.07696ms, state TIME_OK, status STA_NANO | STA_PLL
```

PR #2375, opened 2020-07-29
feat: taint master nodes with `NoSchedule` taint

### LINKED ISSUE #2350: Taint control plane nodes
### Description
Control plane nodes should be tainted with `NoSchedule`.

smira:
Fixes #2350

This also brings in a fix for `coredns` tolerations from
https://github.com/talos-systems/bootkube-plugin/pull/19.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2375)

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -185,7 +185,9 @@ func (h *Client) WorkerIPs(ctx context.Context) (addrs []string, err error) {
 	return addrs, nil
 }
 
-// LabelNodeAsMaster labels a node with the required master label.
+// LabelNodeAsMaster labels a node with the required master label and NoSchedule taint.
+//
+//nolint: gocyclo
 func (h *Client) LabelNodeAsMaster(name string) (err error) {
 	n, err := h.CoreV1().Nodes().Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -205,6 +207,23 @@ func (h *Client) LabelNodeAsMaster(name string) (err error) {
...
+	}
+
+	if !taintFound {
+		n.Spec.Taints = append(n.Spec.Taints, corev1.Taint{
+			Key:    constants.LabelNodeRoleMaster,
+			Value:  "true",
+			Effect: corev1.TaintEffectNoSchedule,
+		})
+	}
+
```
andrewrynhard, line 226: As an alternative, we could pass in the `--register-with-taints` flag to the kubelet: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/. We would just need to be mindful that user's could set that flag in the `extraArgs` field, and we would have to merge them. Not sure I care either way, but wanted to point this option out to see what you think.

smira: you confused me now, now I'm stuck in [Buridan's ass](https://en.wikipedia.org/wiki/Buridan%27s_ass) paradox

andrewrynhard: :laughing: First time seeing that one. Love it. Let's stick with this. It is easier to merge this toleration in with any user supplied ones this way.

PR #2376, opened 2020-07-29
feat: add dynamic config decoder

andrewrynhard:
This adds the ability to dynamically decode mult-doc YAML files.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2376)


pkg/config/configloader/configloader.go
```diff
@@ -9,76 +9,50 @@ import (
...
-// NewFromBytes will take a byteslice and attempt to parse a config file from it.
-func NewFromBytes(in []byte) (config.Provider, error) {
-	c, err := fromBytes(in)
-	if err != nil {
-		return nil, err
+	// Look for the older flat v1alpha1 file first, since we have to handle it in
+	// a special way.
+	for _, manifest := range manifests {
+		if talosconfig, ok := manifest.(*v1alpha1.Config); ok {
+			log.L.Println("WARNING: This version of the machine config will be deprecated in a future release. Please prepare to migrate to the next version of the config as soon as possible.")
```
smira: should we not include this warning yet for 0.6?

pkg/config/decoder/decoder.go
```diff
@@ -0,0 +1,150 @@
...
+	var (
+		version string
+		kind    string
+		spec    *yaml.Node
+	)
+
+	for i, node := range manifest.Content {
+		switch node.Value {
+		case ManifestVersionKey:
+			if err = manifest.Content[i+1].Decode(&version); err != nil {
```
smira, line 107: should we check that `i+1` is available? not sure, but what if document looks like `spec:` without any content on the right?

andrewrynhard: Probably a good test case. I will add it.

PR #2381, opened 2020-07-30
chore: use qemu instead of firecracker in CI

smira:
qemu opens up a bunch of possibilities, including the bootloader
testing.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2381)


smira:
> Is the plan to drop firecracker?

I feel we can consider that in 0.7/0.8. At least we should ship 0.6 with firecracker

PR #2389, opened 2020-07-30
docs: add config valpha2 proposal

andrewrynhard:
This adds a proposal for the v1alpha2 config.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2389)

### DIFF
--- a/docs/proposals/20200730-ConfigV1Alpha2.md
+++ b/docs/proposals/20200730-ConfigV1Alpha2.md
@@ -1,0 +1,203 @@
+# Proposal: Config v1alpha2
+
+Author(s): Andrew Rynhard
+
+## Abstract
+
+The goal in this proposal is to outline the v1alpha2 config manifests and to detail how multi-doc YAML enables API and config driven management of Talos without introducing accidental complexity.
+
+## Background
+
+The `v1alpha1` config has limitations, and layout flaws that we would like to fix.
+In our experience with `v1alpha1` we found that a single configuration file made it difficult to evlove configuration options without breaking things.
+
+## Proposal
+
+```yaml
+---
+# Controlplane contains configuration options for the Talos and Kubernetes
+# control planes. The absence of this manifest enables automatic retrieval from
+# the Talos control plane.
+#
+# Dynamic: true
+# Required: false
+kind: controlplane
+version: v1alpha1
+spec:
+  talos:
+    ca:
+      key: ""
+---
+# Join contains the parameters required to join the Talos and Kubernetes control
+# planes.
+#
+# Required: true
+# Dynamic: false
+kind: join
+version: v1alpha1
+spec:
+  kubernetes:
+    endpoint: ""
+    token: ""
+    ca:
+      crt: ""
+  talos:
+    token: ""
+    ca: # Client needs this to validate trustd connection.
+      crt: ""
+---
+# Etcd contains configuration options for the etcd service. The presence of
+# this manifest enables the etcd service.
+#
+# Required: true
+# Dynamic: true
+kind: etcd
+version: v1alpha1
+spec:
+  image: ""
+  args: {}
+---
+# Kubelet contains configuration options for the kubelet service. In the absence
+# of this manifest, a node will pull this manifest from the Talos control plane.
+#
+# Required: true
+# Dynamic: true
+kind: kubelet
+version: v1alpha1
+spec:
+  image: ""
+  args: {}
+---
+# Machine contains machine configuration options.
+#
+# Required: true
+# Dynamic: false
+kind: machine
+version: v1alpha1
+spec:
+  identity:
+    sans: []
+  bootloader:
+    cmdline: []
+  kernel:
+    parameters: {}
+  storage: {}
+  time:
+    ntp:
+      servers: {}
+  env: {}
+  registries: {}
+---
+# Network contains network configuration options.
+#
+# Required: false
+# Dynamic: false
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
+      dhcp: bool
+      cidrs: []
+  routes:
+    - destination: ""
+      gateway: ""
+  bonds:
+    - name: ""
+      interfaces: []
+      cidrs: []
+      mode: ""
+---
+# Install contains installation options.
+#
+# Proposal: Use this for upgrades. The cached local version will be updated
+#           upon an upgrade.
+#
+# Required: false
+# Dynamic: false
+kind: install
+version: v1alpha1
+spec:
+  disk: ""
+  image: ""
+  zero: bool
+  args: [] # Overrides the defaults determined by Talos.
+---
+# Kubernetes contains options for the cluster creation.
+#
+# Required: true
+# Dynamic: true (Can be submitted with the bootstrap API)
+kind: kubernetes
+version: v1alpha1
+spec:
+  name: ""
+  auth:
+    token: ""
+    pki:
+      ca:
+        crt: ""
+        key: ""
+  encryption:
+    aescbc: ""
+  network:
+    cni: ""
+    dns:
+      domain: ""
+    subnets:
+      pods: []
+      services: []
+  components:
+    controlPlane:
+      apiServer:
+        image: ""
+        args: []
+      controllerManager:
+        image: ""
+        args: []
+      scheduler:
+        image: ""
+        args: []
+    addOns:
+      coreDNS:
+        enabled: bool
+        image: ""
+        args: []
+      kubeProxy:
+        enabled: bool
+        image: ""
+        args: []
+    extraManifests:
+      - source: {}
+```
+
+## Rationale
+
+One benefit of the multi-doc approach is that we can iterate on finely scoped configuration changes.
+If there is a fundamental design flaw in a manifest, we can safely release a new version, and do so gradually.
+Additionally, multi-doc allows for different services to have a manigest.
+If this becomes the case we could create a service such that internal services will get their configuration using this new service's API.
+
+Over time we can start to align manifest kinds with APIs, thereby allowing us users to manage Talos via the API, or by configuration.
+This also opens up opportunities for Talos to become more dynamic. We can make decisions based on the presence/absence of a manifest kind, and, in some cases, either generate or retrieve from the Talos control plane.
+
+In this version, all manifests will be persisted locally, and never updated from the `talos.config` endpoint.
+An API for config management will be introduced that allows for updating the config.
+
+## Compatibility
+
+We can introduce this without any changes to the way that `v1alpha1` behaves.
+However, it is an opportunity for us to omit options that are in `v1alpha1`.
+In other words, there is the possibility that there will not be feature parity.
+
+## Implementation
+
+- preparation
+  - change provider concrete types to interfaces
+  - proposal
+- implementation
+- generate functionality
+- documentation
+
+## Open issues (if applicable)
+
+[A discussion of issues relating to this proposal for which the author does not know the solution.][this section may be omitted if there are none.]


docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+---
+# Machine contains machine configuration options.
+#
+# Dynamic: false
+# Required: true
+kind: machine
+version: v1alpha1
+spec:
+  identity:
+    sans: []
```
smira, line 84: that is for the Talos API, right?

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
+      cidrs:
+        1.2.3.4/8:
+          routes: []
+  bonds:
+    - name: ""
```
smira, line 111: do we set up cidrs for bonds in `interfaces:`?

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: Correct

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: That is the idea. Not sure if its best, still trying to understand things better.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+# Required: false
+# Dynamic: false
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
+      cidrs:
+        1.2.3.4/8:
+          routes: []
```
Ulexus, line 109: Routes are independent of cidrs.  They should be top-level artifacts of the interface.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+#
+# Required: false
+# Dynamic: false
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
+      cidrs:
+        1.2.3.4/8:
```
Ulexus, line 108: It makes more sense to be for the cidrs to be a simple array

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+          routes: []
+  bonds:
+    - name: ""
+      interfaces: []
```
Ulexus, line 112: I wound _consider_ making `bondInterfaces` a different property from `interfaces`, since it has a different, more complicated structure which shouldn't be used or available for standard "native" interfaces.

Same idea for things like `vlanInterfaces` or `tunnelInterfaces` or what have you.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+spec: {}
+---
+# Kubelet contains configuration options for the kubelet service. In the absence
+# of this manifest, a node will pull this manifest from the Talos control plane.
+#
+# Required: True
+# Dynamic: true
+kind: kubelet
+version: v1alpha1
+spec:
```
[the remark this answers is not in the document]
Ulexus: Mount options for the kubelet?  Can you elaborate (or is it just me who doesn't follow this)?

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+  args: [] # Overrides the defaults determined by Talos.
+---
+# Kubernetes contains options for the cluster creation.
+#
+# Requred: true
+# Dynamic: true (Can be submitted with the bootstrap API)
+kind: kubernetes
+version: v1alpha1
+spec:
+  name: "test"
```
Ulexus, line 137: What is this used for?  Is this the cluster name used in the client-side kubernetes configuration?

docs/proposals/20200730-ConfigV1Alpha1.md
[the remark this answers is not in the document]
szucsitg: Kubelet currently doesn't have much access to host paths. We currently mount up /tmp to store some Unix domain socket,  but others could be useful for more persistent storage. As hostPath mounts are handled solely by kubelet it needs to have access to folder to expose as such, even for traditional deployments, like mounting up /var/log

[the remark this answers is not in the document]
Ulexus: Ah, okay, so more `hostMountPoints` or `hostBindMounts` than "mount options"?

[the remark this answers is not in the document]
szucsitg: It's probably better wording, but currently the bind mount propagation can be freely set

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+      services:
+        - 10.96.0.0/12
+  components:
+    controlPlane:
+      apiServer: {}
+      controllerManager: {}
+      scheduler: {}
+    addOns:
+      coreDNS: {}
+      kubeProxy: {}
```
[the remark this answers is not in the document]
Ulexus: There's aren't _many_ CNIs for which this is true (kube-router is the only other one I know of), and Cilium only supports replacing kube-proxy in very specific cases and with special configuration of its own.

Having a `disable` property for it (and probably other addOns) is probably a good idea, though.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+      podCheckpointer: {}
+    extraManifests:
```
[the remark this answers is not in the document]
Ulexus: The entirety of this yaml should only be used during bootstrap, so far as I am aware.  Install manifest is for the _machine_, not the _cluster_.  These extraManifests relate to the _cluster_, so they belong here, IMO.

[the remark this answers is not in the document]
Ulexus: That's not to say, however, that there couldn't be additional static manifests available to the machine, as well.  If we add such a thing, I think that would better live in the kubelet manifest, though.

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: But arguably, there could be some parameters associated with a CIDR, even if I cannot think of anything which would be appropriate at this time.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+---
+# Network contains network configuration options.
+#
+# Required: false
+# Dynamic: false
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
```
Ulexus, line 106: Name is potentially problematic.  Is this the name which the interface should be assigned, given other selectors (which don't exist here), or is this intended to match the natural interface name as generated by the kernel (using one of at least _three_ different naming systems)?  

Systemd-networkd handles this gracefully:  name is the name which the interface should be given after configuration.  Then, you can select the interface via any number of other methods (MAC address, wildcards, kernel name, etc).  Admittedly, that would be complicated to reimplement here, but I do find the name `name` a bit ambiguous here in the grander scheme.

Redhat always does it wrong (general axiom).  Debian always references the kernel interface name, and if you have to do anything fancy, woe be unto you.  Gentoo did a good job with `conf.d/net`, but even it was a bit complicated.  To my mind, systemd-networkd did it the best.

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: Yes, bonds (or `bondInterfaces` as I would prefer) should definitely have CIDRs.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+
+---
+# Network contains network configuration options.
+#
+# Required: false
+# Dynamic: false
+kind: network
+version: v1alpha1
+spec:
+  interfaces:
```
Ulexus, line 105: Remember that we also need a `dhcp: false` or `disableDHCP: true` kind of option.  Interfaces should probably default to DHCP being enabled, but we definitely need a way to disable DHCP on a per-interface basis to eliminate startup delays on systems with many network interfaces.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+  image: docker.io/autonomy/kubelet:latest
+  args: {}
+---
+# Machine contains machine configuration options.
+#
+# Dynamic: false
+# Required: true
+kind: machine
+version: v1alpha1
+spec:
```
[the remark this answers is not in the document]
andrewrynhard: @szucsitg We are thinking about removing the ability to create/overwrite/append arbitrary files (for security reasons). Is this a feature that your team uses still?

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: If we wanted to implement this ideally, it sounds like we need a selector field that can us the MAC address, kernel name, etc., and then name can stay as is. Any proposal on the layout of the selector (not necessarily the field name we need to use)?

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: :+1:

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: Yeah, I think when I wrote this I was thinking that we could do easier lookups on the backend if we wanted to, but UX > DX for the config IMHO. So, agree :).

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: What do you guys think about breaking this up even more and have a config for a single bond, a single interface, and a single route configuration?

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: Any proposal on what that would look like?

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: Looks like this is used by bootkube. They call it `ClusterName`, and we followed suit.

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: Meaning separate yaml?  I don't see a compelling reason to do so.  I don't see those as being likely targets for composition.  I also think a strong binding between routes and the interfaces through which they flow makes for better management.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+  auth:
+    token: wlzjyw.bei2zfylhs2by0wd
+    pki:
+      ca:
+        crt: ""
+        key: ""
+  encryption:
+    aescbc: z01mye6j16bspJYtTB/5SFX8j7Ph4JXxM2Xuu4vsBPM=
+  network:
+    cni: ""
```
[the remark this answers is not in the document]
andrewrynhard: I have been thinking that we may want to change the way this works. Basically, ALL options must be a remote URL, or some embedded definition here. The default just becomes a link to upstream flannel.

docs/proposals/20200730-ConfigV1Alpha1.md
andrewrynhard: Works for me :+1:

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: So I don't think using the systemd-network [Match doc](https://www.freedesktop.org/software/systemd/man/systemd.network.html#%5BMatch%5D%20Section%20Options) would be a bad place to start.

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: I think it would be good to start with:
  - `simpleInterfaces` or just `interfaces`, since that's likely all most people will use
  - `bondInterfaces`
  - `vlanInterfaces`

~~Many (most, actually) of the fields of a `simpleInterface` would extend to `bondInterface` and `vlanInterface` such that in the backend, you may simply inherit `simpleInterface` into the fancier ones.  However, it makes sense to separate them, both from the UX side and the processing side.~~ (see below instead, but leaving this here for background thoughts)

Bond interfaces have the list of component interfaces for the bond, as well as the various parameters for the bond itself.  VLAN interfaces are similar, with a single physical interface on which they are built and a number of vlan parameters.

Down the road, there may be virtual interfaces, loopback interfaces, tunnel interfaces, etc.  There are a whole bunch of different types, which is part of the reason I think they deserve to be separate.

Rather than `simpleInterface` being the root, actually, I think it makes more sense to have an `ipInterface`, which specifies IP-related characteristics common to all network interfaces used for IP(v4+v6).  This would include things like `cidrs` and `routes` which are common to all of them.  For physical interfaces, you would extend that to include the match/selector description to define _which_ physical interface.

Another reasonably good structured network configuration system is that used by Vyatta.  It is organized in a similar way (though they break static routes into an entirely different tree, under "protocols", which I don't especially care for).

docs/proposals/20200730-ConfigV1Alpha1.md
Ulexus: ClusterName, I'd be fine with.  I think Name is less descriptive in this context unless we make it `kind: KubernetesCluster`.

docs/proposals/20200730-ConfigV1Alpha2.md
```diff
@@ -0,0 +1,194 @@
...
+version: v1alpha1
+spec:
+  interfaces:
+    - name: ""
+      dhcp: true
+      cidrs:
+        - 1.2.3.4/8
+  routes:
+    - destination:
+      gateway:
```
andrewrynhard, line 103: @Ulexus In looking at how systemd does it, a route definition is required of dhcp is not enabled. Which makes it seem like `routes` should be part of the `interface` struct. Alternatively, I guess what I am missing from this `route` struct is a reference to the interface.

Ulexus: No, I route definition is never required, unless I'm really missing something.  However, yes, routes should be bound to the interface.

andrewrynhard: I very well could be misreading it. I am looking at https://wiki.archlinux.org/index.php/Systemd-networkd#Configuration_examples. And it says 
> Gateway= this option is mandatory unless DHCP is used
Under `[Route]`.

Ulexus: Meaning that for each route statement, you have to have a gateway for that route.  That doesn't mean that you must have a route statement.

```
ip route add 192.168.2.0/24 via 192.168.1.1
```

In the iproute2 statement above, the `Destination` is `192.168.2.0/24` and the `Gateway` is `192.168.1.1`.

andrewrynhard: :+1: I see now. Not sure how I missed that.

docs/proposals/20200730-ConfigV1Alpha1.md
```diff
@@ -0,0 +1,166 @@
...
+    dns:
+      domain: cluster.local
+    subnets:
+      pods:
+        - 10.244.0.0/16
+      services:
+        - 10.96.0.0/12
+  components:
+    controlPlane:
+      apiServer: {}
```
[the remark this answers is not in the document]
andrewrynhard: Can you elaborate more?

docs/proposals/20200730-ConfigV1Alpha2.md
Ulexus: The official documentation is better for a complete reference.  The arch documentation is excellent for getting your bearings, but the details are in the official freedesktop documentation.

docs/proposals/20200730-ConfigV1Alpha1.md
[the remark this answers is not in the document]
szucsitg: That currently the docs/draft missing the image references for these components

docs/proposals/20200730-ConfigV1Alpha1.md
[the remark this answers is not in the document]
szucsitg: Absolutely, we need to our proxy's public ca cert

docs/proposals/20200730-ConfigV1Alpha1.md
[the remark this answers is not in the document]
andrewrynhard: Ah, yes. 👍

docs/proposals/20200730-ConfigV1Alpha2.md
```diff
@@ -0,0 +1,203 @@
...
+version: v1alpha1
+spec:
+  image: ""
+  args: {}
+---
+# Machine contains machine configuration options.
+#
+# Required: true
+# Dynamic: false
+kind: machine
```
andrewrynhard, line 75: I feel like we should just commit to multi-doc and break this up into dedicate configs. Every time I come back to this, it still feels like too much. Most of this could be their own config. They may be small to begin with, but if feels like it leaves room for expansion, and that we aren't in this weird halfway commit place to multi-doc. A narrowly scoped config feels better IMHO.

smira: I agree with that fully. And if we do that, we don't have to deal with patching when updating configs - we can simply replace those small pieces of the config. 

I think we could go as small as network interface being a config block.

PR #2406, opened 2020-08-07
test: PoC for Sidero integration test

smira:
Code is a mess, just PoC for now.

Changes:

1. Added "PXE" nodes to provisioner, qemu support for it.

2. dhcpd in qemu can hand over TFTP-related options, in fact I guess we
need even less for qemu (it already has UEFI iPXE built-in).

3. Script based on Spencer's notes which works up to server registration
in Sidero.

Lots of stuff to fix, e.g. we need to generate node UUID and put that to
SMBIOS for qemu, additional data we might want to set, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2406)

### DIFF
--- a/internal/pkg/provision/providers/vm/dhcpd.go
+++ b/internal/pkg/provision/providers/vm/dhcpd.go
@@ -21,6 +21,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/provision"
 )
 
+//nolint: gocyclo
 func handler(serverIP net.IP, statePath string) server4.Handler {
 	return func(conn net.PacketConn, peer net.Addr, m *dhcpv4.DHCPv4) {
 		if m.OpCode != dhcpv4.OpcodeBootRequest {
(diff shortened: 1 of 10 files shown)


internal/pkg/provision/providers/vm/dhcpd.go
```diff
@@ -58,6 +58,26 @@ func handler(serverIP net.IP, statePath string) server4.Handler {
 			return
 		}
 
+		if m.IsOptionRequested(dhcpv4.OptionBootfileName) {
+			log.Printf("received PXE boot request from %s", m.ClientHWAddr)
+
+			if match.TFTPServer != "" {
+				bootFilename := match.BootFilename
+
+				for _, userClass := range m.UserClass() {
```
smira, line 67: looks like this is always true for QEMU, so we can simplify this to use single IPXEBootFilename

smira:
Superseded by #2444

PR #2418, opened 2020-08-12
refactor: add `go.mod` for some public packages

smira:
The idea is to provide a way for other projects to import pieces of
Talos without pulling all the top-level dependencies that Talos has.

E.g. if someone needs to generate the config, `pkg/config` is all they
need to import, but root `go.mod` file contains tons of other unrelated
dependencies.

More info on multi-module in a single repo:
https://github.com/golang/go/wiki/Modules#what-are-multi-module-repositories

One caveat is that `go list ./...` (and any use of `./...`) no longer
descends into nested modules, while
`go list github.com/talos-systems/talos/...` does it properly).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2418)


smira:
superseded by #2424

PR #2420, opened 2020-08-13
fix: ignore eth0 interface in docker provisioner

smira:
This avoids pause on container startup when `networkd` tries to do DHCP
over `eth0` (which fails for obvious reasons). Interfaces are
pre-configured in Docker.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2420)

### DIFF
--- a/pkg/provision/providers/docker/docker.go
+++ b/pkg/provision/providers/docker/docker.go
@@ -48,18 +48,25 @@ func (p *provisioner) GenOptions(networkReq provision.NetworkRequest) []generate
 		generate.WithPersist(false),
 	}
 
+	networkConfig := &v1alpha1.NetworkConfig{
+		NetworkInterfaces: []*v1alpha1.Device{
+			{
+				DeviceInterface: "eth0",
+				DeviceIgnore:    true,
+			},
+		},
+	}
+
 	if len(networkReq.Nameservers) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/provision/providers/docker/docker.go
```diff
@@ -48,19 +48,26 @@ func (p *provisioner) GenOptions(networkReq provision.NetworkRequest) []generate
...
 	}
 
+	networkConfig := &v1alpha1.NetworkConfig{
+		NetworkInterfaces: []*v1alpha1.Device{
+			{
+				DeviceInterface: "eth0",
+				DeviceIgnore:    true,
+			},
+		},
+	}
```
andrewrynhard, line 58: I am thinking that skipping DHCP might be the better approach. Reason is that `eth0` isn't guaranteed to be the interface name. Also, if a user has more than one interface, we still get the DHCP delay.

smira: this is docker, I guess there's nothing more than `eth0`

and skipping DHCP is not an option, as config doesn't validate ("invalid addressing method")

andrewrynhard: Just realized that my comment would be true only if the container was running with host net enable. Ignore me.

PR #2427, opened 2020-08-14
refactor: rework `pkg/grpc/tls` to break dependency on `pkg/grpc/gen`

smira:
The goal of `pkg/grpc/tls` is to generate `*tls.Config` based on some
input parameters, but it had dependency on `pkg/grpc/gen` for 'remote'
certificate provider (it uses trustd client to sign CSRs).

Package `pkg/client` which is a part of future `machinery/` module
depends on `pkg/grpc/tls` for TLS config generation, so this pulls
`pkg/grpc` into `machinery/`, while it's not really good idea as most of
`pkg/grpc` is about server-side gRPC handling.

So the idea is to move `pkg/grpc/tls` (which has nothing to do with
gRPC), to `github.com/talos-systems/crypto/tls`, so we need to make sure
it has no dependencies on other Talos code.

The idea of this refactoring is to squash local & remote certificate
renewing providers as they had common part extracted, but even after
that almost all the code was identical except for different generators
beind used.

There should be no functional changes with this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2427)


pkg/grpc/tls/provider.go
```diff
@@ -27,25 +30,79 @@ type CertificateProvider interface {
...
+	)
+
+	if ca, cert, err = provider.update(); err != nil {
+		return nil, fmt.Errorf("failed to create initial certificate: %w", err)
+	}
+
+	provider.updateCertificates(ca, &cert)
+
+	// nolint: errcheck
+	go provider.manageUpdates(context.Background())
```
Ulexus, line 74: This is exactly the kind of place `context.TODO()` was intended for.

Ulexus: Because ultimately, NewRenewingCertificateProvider should accept a context.  Probably out of scope for this PR, but important to have static code analysis cue off this error.  Goroutine spawned with no termination mechanism.

smira: agree, I'll get it updated, but this code was copy-pasted as I squashed two providers together.

I'm not sure if that's the best way as context doesn't allow to wait for provider to stop, we can only cancel it. But that might be something for the future PR.

Ulexus: I'm not sure I follow the second part.  Isn't this what you are talking about?

```go
for {
  select {
    case <-ctx.Done():
      return
    default:
  }
}
```

smira: I mean the caller (whoever creates the context) can cancel the context, but it can't wait for the provider to actually stop. So sometimes the goal is to make sure the provider fully stopped doing whatever it was busy with, and that is not possible with the context alone, we need some kind of `Wait()` mechanism. 

Right now we have providers used only in `main.go` of two services (`apid` and `trustd` iirc), so this isn't much of an issue.

Ulexus: Ah, yeah, okay.  True, that mechanism would be _more_ useful.  I'm just mainly concerned about an un-cleaned-up goroutine... even if it should generally have the lifecycle of the process itself.

But not enough to block.  I think that changing it to `context.TODO()` is more than sufficient at the moment.

PR #2435, opened 2020-08-17
fix: handle bootkube recover correctly, support recovery from etcd 

### LINKED ISSUE #2428: fix bootkube recovery via etcd backend
## Feature Request

This _might_ allow us to get rid of pod-checkpointer.

This should be fixed to provide better resiliency to control plane issues.

### Description

Some pointer and links to the code, rough path forward.

[High-level overview of K8s encryption at rest in etcd](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/)

Today Talos is initializing etcd encryption at rest with AES CBC encryption and a single key. We should add more generic support for different encryption providers which might be used in the cluster.

The way encryption works is the following:

1. If value is encrypted, it is stored with special prefix like `k8s:enc:....` which makes it possible to figure out encryption method from the prefix. Current prefixes: https://github.com/kubernetes/apiserver/blob/56503e793f21bfebf0a0cb8a9b18061eb06fd6c8/pkg/server/options/encryptionconfig/config.go#L47

2. K8s high level from etcd to value flow is: https://github.com/kubernetes/apiserver/blob/master/pkg/storage/value/transformer.go#L132-L194. API server tries to execute any matching (by prefix) transformer on the raw data.Transformer might fail (if that was old key which was rotated out), so it tries next matching transformer.

4. Bootkube tries to decode value verbatim: https://github.com/kubernetes-sigs/bootkube/blob/master/pkg/recovery/etcd.go#L64-L93. We should add intermediate processing which attempts decryption based on config which is a list of prefixes and transformers to apply. We shouldn't use function `NewPrefixTransformers` directly, as it tries to bump Prometheus counters, not something we want to be part of machined. We should use though transformers, e.g. https://godoc.org/k8s.io/apiserver/pkg/storage/value/encrypt/aes 

So the proposed change to bootkube etcd recovery plugin is:

1. `NewEtcdBackend` supports additional parameter:

```go
type TransformerFromStorage func(value []byte) ([]byte, error) 

func NewEtcdBackend(..., transformer TransformerFromStorage)
```
   We could also add another method `NewEtcdBackendWithTransformer` to keep public API intact.

2. `get` and `getRaw` methods of `Backend` attempt to decrypt data via the `transformer` callback.

On Talos side:

1. We implement `TransformerFromStorage` with prefix list + list of transformers based on cluster config (AES CBC key), or we can read kube-apiserver config?

smira:
Bootkube recover process (and `talosctl recover`) was actually
regenerating assets each time `recover` runs forcing control plane to be
at the state when cluster got created. This PR fixes that by running
recover process correctly.

Recovery via etcd was fixed to handle encrypted etcd data:
it follows the way `apiserver` handles encryption at rest, and as at
the moment AES CBC is the only supported encryption method, code simply
follows the same path.

Fixes #2428

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2435)


smira:
I would like to clean up the code, but this should work

PR #2437, opened 2020-08-18
feat: add grub bootloader

andrewrynhard:
This moves to using grub instead of syslinux.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2437)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/constants.go
```diff
@@ -0,0 +1,18 @@
...
+package grub
+
+import "github.com/talos-systems/talos/pkg/machinery/constants"
+
+const (
+	// BootA is a bootloader label.
+	BootA = "boot-a"
+
+	// BootB is a bootloader label.
+	BootB = "boot-b"
```
andrewrynhard, line 14: These should be top-level bootloader constants.

cmd/installer/pkg/install/install.go
```diff
@@ -73,11 +74,13 @@ func Install(p runtime.Platform, seq runtime.Sequence, opts *Options) (err error
...
 type Installer struct {
-	cmdline  *procfs.Cmdline
-	options  *Options
-	manifest *Manifest
-	Current  string
-	Next     string
+	cmdline    *procfs.Cmdline
+	options    *Options
+	manifest   *Manifest
+	bootloader bootloader.Bootloader
```
andrewrynhard, line 80: Only added this. We now use an interface instead of an explicit bootloader (still a WIP).

cmd/installer/pkg/install/install.go
```diff
@@ -240,9 +244,9 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 	if seq != runtime.SequenceUpgrade || !i.bootPartitionFound {
 		i.cmdline.Append("initrd", filepath.Join("/", i.Current, constants.InitramfsAsset))
 
-		syslinuxcfg := &syslinux.Cfg{
+		grubcfg := &grub.Cfg{
```
andrewrynhard, line 247: Need to create this via the bootloader interface (method is still needed in the interface definition).

cmd/installer/pkg/install/manifest.go
```diff
@@ -178,15 +178,16 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
...
+		typeID := "21686148-6449-6E6F-744E-656564454649"
 		opts = append(opts, partition.WithPartitionType(typeID), partition.WithPartitionName(t.Label), partition.WithLegacyBIOSBootableAttribute(true))
 	case constants.EphemeralPartitionLabel:
 		// Ephemeral Partition
-		typeID := "AF3DC60F-8384-7247-8E79-3D69D8477DE4"
+		typeID := "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
 		opts = append(opts, partition.WithPartitionType(typeID), partition.WithPartitionName(t.Label), partition.WithMaximumSize(true))
 	default:
-		typeID := "AF3DC60F-8384-7247-8E79-3D69D8477DE4"
+		typeID := "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
```
andrewrynhard, line 190: Turns out that we have been doing the UUIDs wrong in the GPT package. 3 fields are actually middle endian (or mixed endiain). Which means the first three groups of the UUID should be encoded as Little Endian, while the rest should be encoded as Big Endian. When moving to grub we are required to have a specific partition type for BIOS.

```
Welcome to fdisk (util-linux 2.36).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.


Command (m for help): p
Disk /dev/loop11: 4 GiB, 4294967296 bytes, 8388608 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 125A2162-E1D2-11EA-961D-025D91EBDA2C

Device          Start     End Sectors  Size Type
/dev/loop11p1      34 1048610 1048577  512M BIOS boot
/dev/loop11p2 1048611 8388574 7339964  3.5G Linux filesystem
```

Notice `BIOS boot` for the first partition's (boot partition) `Type`. Without the middle-endian UUID, this field is completely wrong. The `"AF3DC60F-8384-7247-8E79-3D69D8477DE4"` happened to be "correct" because it was set as such here in a hacky way that I can't remember, but the gist was a double negative. Best way I can explain it.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
```diff
@@ -4,8 +4,11 @@
...
 package bootloader
 
+import "github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
+
 // Bootloader describes a bootloader.
 type Bootloader interface {
+	Labels() (string, string, error)
 	Prepare(string) error
-	Install(string, interface{}) error
+	Install(string, interface{}, runtime.Sequence, bool) error
```
andrewrynhard, line 13: We were skipping the interface before this PR. This just aligns it with what syslinux does so that the interface works for both bootloaders.

pkg/blockdevice/table/gpt/partition/options.go
```diff
@@ -47,17 +47,16 @@ func WithMaximumSize(o bool) Option {
 func WithLegacyBIOSBootableAttribute(o bool) Option {
 	return func(args *Options) {
 		if o {
-			args.Flags = 4
+			args.Flags |= (1 << 2)
```
andrewrynhard, line 50: Proper way to set bit 2 of the flags.

cmd/installer/pkg/install/manifest.go
```diff
@@ -74,13 +74,29 @@ func NewManifest(label string, sequence runtime.Sequence, opts *Options) (manife
...
+		Test:   false,
+	}
+
+	biosTarget := &Target{
+		Device: opts.Disk,
+		Label:  constants.BIOSGrubPartitionLabel,
+		Size:   1 * 1024 * 1024,
+		Force:  true,
+		Test:   false,
+	}
```
andrewrynhard, line 91: New partitions to satisfy grub requirements.

cmd/installer/pkg/install/manifest.go
```diff
@@ -176,18 +192,23 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
 
 	opts := []interface{}{}
 
+	const (
+		EFISystemPartition  = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
+		BIOSBootPartition   = "21686148-6449-6E6F-744E-656564454649"
+		LinuxFilesystemData = "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
+	)
```
andrewrynhard, line 199: These should be part of the gpt package since they are specific to GPT.

cmd/installer/pkg/install/manifest.go
```diff
@@ -213,19 +234,33 @@ func (t *Target) Partition(bd *blockdevice.BlockDevice) (err error) {
...
-	log.Printf("formatting partition %q as %q with label %q\n", t.PartitionName, "xfs", t.Label)
-	opts := []xfs.Option{xfs.WithForce(t.Force)}
+		if t.Label != "" {
+			opts = append(opts, xfs.WithLabel(t.Label))
+		}
 
-	if t.Label != "" {
-		opts = append(opts, xfs.WithLabel(t.Label))
-	}
+		return xfs.MakeFS(t.PartitionName, opts...)
```
andrewrynhard, line 251: Boot partition can now be XFS.

pkg/blockdevice/probe/probe.go
```diff
@@ -178,6 +180,49 @@ func probe(devpath string) (devpaths []string) {
 	return devpaths
 }
 
+// GetBlockDeviceWithPartitonName probes all known block device's partition
+// table for a parition with the specified name.
+func GetBlockDeviceWithPartitonName(name string) (bd *blockdevice.BlockDevice, err error) {
```
andrewrynhard, line 185: This is required to do a lookup up the partitions that we never create a filesystem on.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -0,0 +1,192 @@
...
+	blk, err := util.DevnameFromPartname(dev.Path)
+	if err != nil {
+		return err
+	}
+
+	loopDevice := strings.HasPrefix(blk, "loop")
+
+	blk = fmt.Sprintf("/dev/%s", blk)
+
+	// TODO: run twice
```
smira, line 129: I forgot to remove this `TODO:`, we can remove it now (this was to run twice on x86 for bios/efi)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -131,14 +131,22 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
 				"install",
 				Install,
 			).Append(
-				"mountBoot",
+				"mountSystem",
 				MountBootPartition,
```
smira, line 135: not for this PR, but can we/should we skip mounting `/boot`? it looks like we almost never need it now? (or do we still need to fetch meta?)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -160,8 +168,13 @@ func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
 
 	phases = phases.AppendWhen(
 		r.State().Platform().Mode() != runtime.ModeContainer,
-		"mountBoot",
+		"mountSystem",
 		MountBootPartition,
+		MountSystemPartition,
+	).AppendWhen(
+		r.State().Platform().Mode() != runtime.ModeContainer,
+		"mountEFI",
```
smira, line 176: can we skip that? as only installer touches this ever, and installer blows up partition table anyways?

pkg/endianness/endianness.go
```diff
@@ -0,0 +1,94 @@
...
+	if err := binary.Write(buf, binary.LittleEndian, timeMid); err != nil {
+		return nil, err
+	}
+
+	timeHigh := binary.BigEndian.Uint16(data[6:8])
+	if err := binary.Write(buf, binary.LittleEndian, timeHigh); err != nil {
+		return nil, err
+	}
+
+	clockSeqHi := data[8:9][0]
```
smira, line 34: isn't that just `data[8]` which is a `byte`, so no need to use endiannes conversion? 

`buf.Write(data[8:10])` ?

pkg/endianness/endianness.go
```diff
@@ -0,0 +1,94 @@
...
+	b = buf.Bytes()
+
+	return b, nil
+}
+
+// FromMiddleEndian converts a middle-endian byte slice representation of a
+// UUID to a big-endian byte slice representation of a UUID.
+//
+//nolint: dupl
+func FromMiddleEndian(data []byte) (b []byte, err error) {
```
smira, line 58: should we get some simple unit-test for this one, as it seems that these two functions should cancel each other?

Makefile
```diff
@@ -148,7 +149,8 @@ talosctl-%:
 talosctl: $(TALOSCTL_DEFAULT_TARGET) ## Builds the talosctl binary for the local machine.
 
 image-%: ## Builds the specified image. Valid options are aws, azure, digital-ocean, gcp, and vmware (e.g. image-aws)
-	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged $(USERNAME)/installer:$(TAG) image --platform $*
+	@docker run --rm -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/out --privileged $(REGISTRY)/$(USERNAME)/installer:$(TAG) image --platform $*
+	@sleep 5
```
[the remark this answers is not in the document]
andrewrynhard: Yep, exactly.

internal/app/machined/main.go
```diff
@@ -74,10 +74,18 @@ func handle(err error) {
 		log.Print(err)
 	}
 
-	if err := syslinux.Revert(); err != nil {
+	meta, err := bootloader.NewMeta()
+	if err != nil {
+		log.Printf("failed to open meta: %v", err)
```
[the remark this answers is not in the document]
andrewrynhard: Thanks, this probably is the fix for that panic we saw the other day.

cmd/installer/pkg/install/manifest.go
```diff
@@ -74,13 +78,29 @@ func NewManifest(label string, sequence runtime.Sequence, opts *Options) (manife
...
+		Device: opts.Disk,
+		Label:  constants.BIOSGrubPartitionLabel,
+		Size:   1 * 1024 * 1024,
+		Force:  true,
+		Test:   false,
+	}
+
 	var bootTarget *Target
 
 	if opts.Bootloader {
```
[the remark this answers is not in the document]
andrewrynhard: The plan, I thought, was to ultimately remove this option, and always install a bootloader.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
andrewrynhard: :thinking: The only times that we need to now is when we perform an upgrade.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
andrewrynhard: Yep, I think so.

PR #2443, opened 2020-08-20
fix: pass config via stdin

andrewrynhard:
In order to perform upgrades the way we would like, it is important that
we avoid any bind mounts into containers. This change ensures that all
system services get their config via stdin.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2443)

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
@@ -412,6 +412,38 @@ func (suite *ContainerdSuite) TestImportFail() {
 		suite.containerdNamespace, containerdrunner.WithContainerdAddress(suite.containerdAddress)).Import(reqs...))
 }
 
+func (suite *ContainerdSuite) TestContainerStdin() {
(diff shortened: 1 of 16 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd_test.go
```diff
@@ -424,3 +424,35 @@ func TestContainerdSuite(t *testing.T) {
 
 	suite.Run(t, new(ContainerdSuite))
 }
+
+func (suite *ContainerdSuite) TestContainerStdin() {
```
smira, line 428: nit: you might want it to put it before the `TestContainerdSuite()` func above to keep it aligned

smira (dismissed):

internal/app/machined/pkg/system/runner/containerd/containerd_test.go
andrewrynhard: Ah yep :+1:

smira (approved):

PR #2444, opened 2020-08-20
test: add support for PXE nodes in qemu provision library

smira:
This isn't supposed to be used ever in Talos directly, but rather only
in integration tests for Sidero.

### DIFF
--- a/pkg/provision/providers/qemu/node.go
+++ b/pkg/provision/providers/qemu/node.go
@@ -16,8 +16,10 @@ import (
 	"strconv"
 	"syscall"
 
+	"github.com/google/uuid"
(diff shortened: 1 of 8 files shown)


pkg/provision/providers/qemu/node.go
```diff
@@ -127,10 +141,16 @@ func (p *provisioner) createNode(state *vm.State, clusterReq provision.ClusterRe
 
 	// no need to wait here, as cmd has all the Stdin/out/err via *os.File
 
+	nodeType := machine.Type(-1) // HACK!
```
smira, line 144: oh I probably want to fix that before we merge it, I think I saw `TypeUnknown` in one of the PRs

PR #2455, opened 2020-08-27
feat: update linux to 5.8.5

rsmitty:
This PR pulls in the new stable linux version.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2455)


smira:
we need to bump kernel version in constants to make tests happy

PR #2464, opened 2020-08-31
chore: drop vmlinux from assets

smira:
`vmlinux` (uncompressed kernel) shouldn't be much useful to the people
as it can't be booted directly with most of the boot tools (it's too
huge). We had it enabled for firecracker, but now firecracker has
support for decompressing `vmlinuz` on the fly.

So, to reduce confusion, make assets smaller, remove `vmlinux`.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2464)

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -492,7 +492,6 @@ local release = {
       '_out/talosctl-linux-arm64',
       '_out/talosctl-linux-armv7',
       '_out/vmware.ova',
-      '_out/vmlinux',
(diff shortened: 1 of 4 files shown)


smira:
> I am a little afraid the either someone is using this, or we will run into a use case for this in the future. I suppose removing this will show us slightly_smiling_face .

afaik, this almost 100MB `vmlinuz` monster is not actually bootable via any bootloader nowadays.

PR #2475, opened 2020-09-02
fix: add syslinux to create ISO

andrewrynhard:
We need syslinux to create an ISO.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2475)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -384,8 +384,10 @@ RUN apk add --no-cache --update \
     cdrkit \
     efibootmgr \
     qemu-img \
+    syslinux \
     util-linux \
     xfsprogs
+COPY --from=docker.io/autonomy/syslinux:v0.2.0-35-g8be6358 / /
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -384,8 +384,10 @@ RUN apk add --no-cache --update \
     cdrkit \
     efibootmgr \
     qemu-img \
+    syslinux \
     util-linux \
     xfsprogs
+COPY --from=docker.io/autonomy/syslinux:v0.2.0-34-g7e57ef7 / /
```
smira, line 390: should this be `v0.2.0-35-g8be6358` for the sake of consistency?

smira (approved):

Dockerfile
andrewrynhard: whoops, yes

rsmitty (approved):

smira (approved):

PR #2479, opened 2020-09-02
feat: build Talos images/artifacts for amd64/arm64

### LINKED ISSUE #2398: build Talos on arm64 & amd64
## Feature Request

1. Produce multi-platform images `talos` and `installer`.
2. Produce `vmlinuz`, `initramfs.xz` per platform (directory layout, release integration)
3. Do we want to build `talosctl` natively on ARM, or use x86 to build all `talosctl` flavors?

(we should probably drop `vmlinux`, I don't think it's being used anymore)

### Description

Current flow with docker exporter doesn't work for multi-platform images, so we should push directly to the registry instead. Probably for non-release builds we should pick a different image name, like `installer-pr`, or different registry username so that PR builds don't mess up with production images. 

We need to figure out directory layout for kernel/initramfs, as if we launch buildkit with multiple arch, it exports results with its own layout enforced in some way.

smira:
By default, build outside of Drone works the same and builds only amd64
version, loads images back into dockerd, etc.

If multiple platforms are used, multi-arch images are built which can't
be exported to docker or to `.tar` image, they're always pushed to the
registry (even for PR builds to our internal CI registry).

Artifacts as files (initramfs, kernel) now have `-arch` suffix:
`vmlinuz-amd64`, `initramfs-amd64.xz`. "Magic" script normalizes output
paths depending on whether single platform or multiple platforms were
given.

VM provisioners accept magic `${ARCH}` in initramfs/kernel paths which
gets replaced by cluster architecture.

Fixes #2398

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2479)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -386,7 +386,7 @@ func init() {
 	createCmd.Flags().DurationVar(&clusterWaitTimeout, "wait-timeout", 20*time.Minute, "timeout to wait for the cluster to be ready")
 	createCmd.Flags().BoolVar(&forceInitNodeAsEndpoint, "init-node-as-endpoint", false, "use init node as endpoint instead of any load balancer endpoint")
 	createCmd.Flags().StringVar(&forceEndpoint, "endpoint", "", "use endpoint instead of provider defaults")
-	createCmd.Flags().StringVar(&kubernetesVersion, "kubernetes-version", constants.DefaultKubernetesVersion, "desired kubernetes version to run")
+	createCmd.Flags().StringVar(&kubernetesVersion, "kubernetes-version", "", fmt.Sprintf("desired kubernetes version to run (defaults to %s)", constants.DefaultKubernetesVersion))
```
smira, line 389: use defaults provided by config instead of generating config with images

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -109,6 +110,10 @@ func (e *Etcd) Runner(r runtime.Runtime) (runner.Runner, error) {
 		env = append(env, fmt.Sprintf("%s=%s", key, val))
 	}
 
+	if goruntime.GOARCH == "arm64" {
+		env = append(env, "ETCD_UNSUPPORTED_ARCH=arm64")
```
smira, line 114: :shrug:

internal/integration/provision/upgrade.go
```diff
@@ -92,13 +92,15 @@ func upgradeBetweenTwoLastReleases() upgradeSpec {
 	return upgradeSpec{
 		ShortName: fmt.Sprintf("%s-%s", stableVersion, nextVersion),
 
-		SourceKernelPath:     helpers.ArtifactPath(filepath.Join(trimVersion(stableVersion), constants.KernelAsset)),
-		SourceInitramfsPath:  helpers.ArtifactPath(filepath.Join(trimVersion(stableVersion), constants.InitramfsAsset)),
-		SourceInstallerImage: fmt.Sprintf("%s:%s", images.DefaultInstallerImageRepository, stableVersion),
+		SourceKernelPath:    helpers.ArtifactPath(filepath.Join(trimVersion(stableVersion), constants.KernelAsset)),
+		SourceInitramfsPath: helpers.ArtifactPath(filepath.Join(trimVersion(stableVersion), constants.InitramfsAsset)),
```
smira, line 96: this fixes the `ghcr.io/talos-systems` vs. `docker.io/autonomy` issue

Makefile
```diff
@@ -178,17 +197,17 @@ check-dirty: ## Verifies that source tree is not dirty
 
 .PHONY: unit-tests
 unit-tests: ## Performs unit tests.
-	@$(MAKE) local-$@ DEST=$(ARTIFACTS) TARGET_ARGS="--allow security.insecure"
+	@$(MAKE) local-$@ DEST=$(ARTIFACTS) TARGET_ARGS="--allow security.insecure" PLATFORM=linux/amd64
```
smira, line 200: this (and similar changes): force single-platform to avoid running it for ARM for now (we can revisit unit-tests later, but most of others are single-platform sufficient)

.drone.jsonnet
```diff
@@ -144,12 +159,19 @@ local setup_ci = {
...
+    SSH_KEY: {
+      from_secret: 'ssh_key'
+    }
+  },
+
   commands: [
     'sleep 5', // Give docker enough time to start.
     'apk add coreutils',
-    'docker buildx create --driver docker-container --platform linux/amd64 --buildkitd-flags "--allow-insecure-entitlement security.insecure" --name local --use unix:///var/outer-run/docker.sock',
+    'install-ci-key',
```
smira, line 171: this can be removed (TODO for myself)

.drone.jsonnet
```diff
@@ -144,12 +159,19 @@ local setup_ci = {
   name: 'setup-ci',
   image: 'autonomy/build-container:latest',
   privileged: true,
+  environment: {
```
smira, line 162: this can be removed

.drone.jsonnet
```diff
@@ -80,6 +81,19 @@ local volumes = {
     },
   },
 
+  ssh: {
```
smira, line 84: this should be removed as well

andrewrynhard (approved):
:tada:

smira:
Going to merge this, next week we can have time to fix all the bugs I missed )

PR #2503, opened 2020-09-08
test: add e2e test to the provision (upgrade) tests

smira:
Add sonobuoy runner code with log fetching on failure. Use hand-picked
set of e2e tests to run: verify basic pod functionality, verify service
connectivity.

Add option `--run-e2e` to the `talosctl health` to run quick e2e test to
verify cluster health.

Add option to run provision tests with custom CNI, run one track of
provision tests with Cilium.

Bump Cilium to 1.8.2.

Talos 0.6 won't uncordon node automatically after upgrade from 0.5, as
0.5 doesn't put annotation. Workaround that in upgrade tests.

Bump upgrade test version to 0.6.0 release.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2503)


go.mod
```diff
@@ -2,9 +2,11 @@ module github.com/talos-systems/talos
 
 go 1.13
 
-replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
-
-replace github.com/talos-systems/talos/pkg/machinery => ./pkg/machinery
+replace (
+	github.com/Azure/go-autorest v10.8.1+incompatible => github.com/Azure/go-autorest/autorest v0.9.1
```
smira, line 6: this was required to pull in sonobuoy

go.mod
```diff
@@ -2,9 +2,11 @@ module github.com/talos-systems/talos
...
+	github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+	github.com/talos-systems/talos/pkg/machinery => ./pkg/machinery
```
andrewrynhard, line 8: Will this cause issues when importing Talos?

andrewrynhard (approved):

internal/integration/provision/provision.go
```diff
@@ -49,7 +51,7 @@ var DefaultSettings Settings = Settings{
 	MTU:                        1500,
 	CPUs:                       1,
 	MemMB:                      1.5 * 1024,
-	DiskGB:                     4,
+	DiskGB:                     8,
```
smira, line 54: looks like node runs out of space while running e2e tests with 4GiB

go.mod
smira: `replace` directives are ignored when using module, but good question what will happen if you import whole module, this might need similar replace

andrewrynhard: An alternative is that we get rid of this module? Didn't we find another workaround?

smira: I'll look into that

andrewrynhard (approved):

PR #2518, opened 2020-09-15
feat: allow for link local networking

### LINKED ISSUE #2486: MTU Settings not working
Customer mentioned that mtu settings didn't take effect in 0.6.x release. Need to verify if that's something with talos or with sidero. If it's sidero, we may have fixed it when we pulled in newer config package.

rsmitty:
This PR allows for the ability to specify neither CIDR nor DHCP in the
talos machine config. The result here should allow for things like SLAAC
addressing with ipv6.

Will close #2486 

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2518)

### DIFF
--- a/internal/app/networkd/pkg/networkd/netconf.go
+++ b/internal/app/networkd/pkg/networkd/netconf.go
@@ -55,8 +55,10 @@ func buildOptions(device config.Device, hostname string) (name string, opts []ni
 
 			opts = append(opts, nic.WithNoAddressing())
 		} else {
-			d := &address.DHCP{}
-			opts = append(opts, nic.WithAddressing(d))
+			// No CIDR and DHCP==false results in a static without an IP.
+			// This handles cases like slaac addressing.
+			s := &address.Static{RouteList: device.Routes(), Mtu: device.MTU()}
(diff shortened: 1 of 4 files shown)


internal/app/networkd/pkg/networkd/netconf.go
```diff
@@ -55,8 +55,10 @@ func buildOptions(device config.Device, hostname string) (name string, opts []ni
 
 			opts = append(opts, nic.WithNoAddressing())
 		} else {
-			d := &address.DHCP{}
-			opts = append(opts, nic.WithAddressing(d))
+			// No CIDR and DHCP==false results in a static without an IP.
```
rsmitty, line 58: I know we talked about having another default case here that still defaulted to dhcp, but this code isn't actually called anywhere but here https://github.com/talos-systems/talos/blob/master/internal/app/networkd/pkg/networkd/networkd.go#L77. And so things that are only explicitly defined in the config file will make it here.

smira (dismissed):

smira (dismissed):

andrewrynhard (approved):

Issue #1851, opened 2020-02-02
e2e generated files are not removed


Ulexus:
As an example, e2e-docker generated config files in /tmp/e2e/docker (kubeconfig and talosconfig).  If those files are not removed before the next e2e-docker test, that test will fail (because it tries to create the kubeconfig file, which already exists).

An answer may be to have the `make e2e-*` perform the cleanup (both cluster destruction and temp files).

PR #2531, opened 2020-09-16
fix: clean up e2e TMP dir

Ulexus:
Before each e2e test, we should clean up the TMP dir, if it is exists, as well as tearing down any existing potentially-conflicting cluster.

Additionally, this modifies the `e2e.sh` script so that it is side-effect-free.
Originally, the TMP dir would be created any time the file was sourced.
While in this particular case, that was reasonably inoffensive, the idea is to only export variables and functions from within a sourced script include.

Additional scrutiny is probably in order, as a result.

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2531)

### DIFF
--- a/hack/test/e2e-aws.sh
+++ b/hack/test/e2e-aws.sh
@@ -8,8 +8,9 @@ REGION="us-east-1"
 BUCKET="talos-ci-e2e"
 
 function setup {
+  make_tmp
+
   # Setup svc account
-  mkdir -p ${TMP}
 
   # Untar image
   tar -C ${TMP} -xf ${ARTIFACTS}/aws.tar.gz
--- a/hack/test/e2e-azure.sh
+++ b/hack/test/e2e-azure.sh
@@ -5,6 +5,8 @@ set -eou pipefail
 source ./hack/test/e2e.sh
 
 function setup {
+  make_tmp
+
   AZURE_STORAGE_ACCOUNT=talostesting
   AZURE_STORAGE_CONTAINER=talostesting
   AZURE_GROUP=talos
--- a/hack/test/e2e-capi.sh
+++ b/hack/test/e2e-capi.sh
@@ -4,6 +4,8 @@ set -eou pipefail
 
 source ./hack/test/e2e.sh
 
+make_tmp
+
 export CABPT_VERSION="0.2.0-alpha.0"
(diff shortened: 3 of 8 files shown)


Ulexus:
Fixes #1851

andrewrynhard (approved):
This LGTM, but I a have fallen behind on how our current e2e stuff works. I would advise to wait on others who might have more insight :smile:.

smira:
my only question is interaction between `e2e-docker` and `e2e-gcp`, `e2e-aws`, my understanding is that `e2e-gcp` uses `$TMP` from `e2e-docker`, as docker one contains CAPI bootstrap deployment, but I might be wrong on this one.

Ulexus:
Yeah, the e2e test scripts need a lot of love.  They appear to be pure chaos to my untrained eyes, but at a minimum, it looks like CAPI half uses the e2e-docker dir (apparently only the kubeconfig).

Ulexus:
I'm failing to figure out why all the CAPI stuff is hard-coded to docker.  If it's just legacy kind of "that's where it was," I don't mind just fixing up all those references, but I'm not familiar enough with the CAPI stuff to know whether there may have been a good reason for it.

Ulexus:
But until then, this PR must be put on hold, since this seems like it could definitely interact.

andrewrynhard:
> I'm failing to figure out why all the CAPI stuff is hard-coded to docker. If it's just legacy kind of "that's where it was," I don't mind just fixing up all those references, but I'm not familiar enough with the CAPI stuff to know whether there may have been a good reason for it.

The reason for this is that we need the cluster to survive across drone steps. Since docker is a drone service, when we create a docker-based cluster, the cluster will survive across multiple steps. So in one step we setup the cluster then in subsequent steps we kick off CAPI against that cluster.

Ulexus:
I get the use of Docker generally, but why the `/tmp/e2e/docker` directory itself, even for non-e2e-docker CAPI tests?

andrewrynhard:
> I get the use of Docker generally, but why the `/tmp/e2e/docker` directory itself, even for non-e2e-docker CAPI tests?

I think because that is where we dump the Talos config and the kubeconfig.

PR #2534, opened 2020-09-17
fix: improve error message on empty config

### LINKED ISSUE #2096: talosctl: Missing context
### Description

When running a `talosctl` command without `~/.talos/config` existing, we create it and then fail complaining that there is no context.

I feel like we should fail on missing talos config file.

### Logs

```
error getting client credentials: 'context' key is not set in the config
```

smira:
Fixes #2096

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2534)


smira:
with the fix:

```sh
$ talosctl list
error constructing client: failed to resolve configuration context: talos config file is empty
```

andrewrynhard (approved):

PR #2547, opened 2020-09-19
feat: add default install image

andrewrynhard:
We longer require an install image in the install config.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2547)

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
@@ -927,6 +927,10 @@ func (t *TimeConfig) Servers() []string {
 
 // Image implements the config.Provider interface.
 func (i *InstallConfig) Image() string {
+	if i.InstallImage == "" {
+		return constants.DefaultInstallerImage
+	}
+
 	return i.InstallImage
 }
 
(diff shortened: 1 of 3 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -927,6 +927,10 @@ func (t *TimeConfig) Servers() []string {
 
 // Image implements the config.Provider interface.
 func (i *InstallConfig) Image() string {
+	if i.InstallImage == "" {
```
andrewrynhard, line 930: I couldn't find anywhere in the code that might block us from even allowing this (validation of the install image for example). If I am missing something, please let me know.

smira: We could try keeping install image empty by default in talosctl cluster create (as a flag) to verify this

smira (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
andrewrynhard: Ah, good idea.

Ulexus (approved):

smira (dismissed):

smira (approved):

PR #2553, opened 2020-09-23
feat: add ApplyConfiguration API

### LINKED ISSUE #2345: Support Config Update API
For configs that have been persisted, there may be a case in which users want to re-pull to update certain values or whatnot. We should support a way to trigger a re-pull regardless of persistence setting.

MVP:
- Validate the received config
- Replace on disk config upon success of validation
- Reboot

Ulexus:
Adds the ability to apply (replace) an existing node configuration with
a new one via the Machine API.

Fixes #2345

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2553)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
```diff
@@ -39,7 +39,11 @@ func (r *Runtime) Config() config.Provider {
 func (r *Runtime) SetConfig(b []byte) error {
 	cfg, err := configloader.NewFromBytes(b)
 	if err != nil {
-		return fmt.Errorf("failed to set config: %w", err)
+		return fmt.Errorf("failed to parse config: %w", err)
+	}
+
+	if err := cfg.Validate(r.State().Platform().Mode()); err != nil {
```
Ulexus, line 45: This is invasive beyond the strict scope of this issue, but IMO, the config should always be validated before it is applied to the Runtime.  As far as I can tell, this is safe everywhere, in that the platform mode is always defined before we set the config.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -40,6 +40,31 @@ func (p PhaseList) AppendWhen(when bool, name string, tasks ...runtime.TaskSetup
...
+	).Append(
+		"saveConfig",
+		SaveConfig,
+	).Append(
+		"unmountState",
+		UnmountStatePartition,
+	).Append(
+		"stopEverything",
+		StopAllServices,
+	).Append(
```
Ulexus, line 60: It could be argued that we should not bind a reboot implicitly with the application of a new configuration.  I think we should do so, at least by default, but I could see opposing arguments.

internal/pkg/mount/mount.go
```diff
@@ -28,21 +28,31 @@ type RetryFunc func(*Point) error
...
 		// Repair the disk's partition table.
 		if mountpoint.Resize {
 			if err = mountpoint.ResizePartition(); err != nil {
 				return fmt.Errorf("error resizing %q: %w", iter.Value().Source(), err)
 			}
 		}
 
-		if err = mountpoint.Mount(); err != nil {
-			return fmt.Errorf("error mounting %q: %w", iter.Value().Source(), err)
+		mounted, err = mountpoint.IsMounted()
```
Ulexus, line 47: The below `mountpoint.Mount()` _might_ catch this, but its behaviour seems dangerous to me; thus, I chose to always gatekeep the mount with this check.  Intentionally outside this check are the resizing above an below, so as to minimize possible side-effects.

pkg/machinery/client/client.go
```diff
@@ -360,6 +360,11 @@ func (c *Client) Kubeconfig(ctx context.Context) ([]byte, error) {
 	return kubeconfig, err
 }
 
+// ApplyConfiguration implements proto.OSClient interface.
+func (c *Client) ApplyConfiguration(ctx context.Context, req *machineapi.ApplyConfigurationRequest, callOptions ...grpc.CallOption) (resp *machineapi.ApplyConfigurationResponse, err error) {
```
Ulexus, line 364: All of these methods are labeled as "implements proto.OSCLient", but that is not actually true, so far as I can tell.  As such, it could be easily argued that the simpler form of accepting just the data and returning just the error is acceptable here.  However, since the options may grow over time, I decided to just include the canonical form.  I am not married to this.

Ulexus:
Well, this crashes badly.  It looks like init must be dying, but no logs or anything.  Just an immediate reboot.   Moving this back to draft...

go.mod
```diff
@@ -65,18 +57,15 @@ require (
 	github.com/talos-systems/grpc-proxy v0.2.0
 	github.com/talos-systems/net v0.2.0
 	github.com/talos-systems/talos/pkg/machinery v0.0.0-20200818212414-6a7cc0264819
-	github.com/u-root/u-root v6.0.0+incompatible // indirect
-	github.com/vishvananda/netns v0.0.0-20200520041808-52d707b772fe // indirect
 	github.com/vmware-tanzu/sonobuoy v0.19.0
 	github.com/vmware/vmw-guestinfo v0.0.0-20200218095840-687661b8bd8e
-	go.etcd.io/etcd v0.5.0-alpha.5.0.20200819165624-17cef6e3e9d5 // v3.4.10
+	go.etcd.io/etcd v3.3.13+incompatible // v3.4.10
```
smira, line 62: this is a bad change, please don't move the tag there. it's a long story with Go modules, etcd and tags, but the tag should be kept as it is...

internal/app/machined/pkg/runtime/sequencer.go
```diff
@@ -34,31 +34,36 @@ const (
 	SequenceRecover
 	// SequenceNoop is the noop sequence.
 	SequenceNoop
+	// SequenceApplyConfiguration is the apply configuration sequence.
+	SequenceApplyConfiguration
```
smira, line 38: nit: should we sort the constants alphabetically as well?

I think those constants never get out of the running system, so it should be safe to change them

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
smira: I like the idea of rebooting on config apply.

Only one concern is what happens if config is bad, so node might not boot back, and in that case it seems as dangerous as upgrade. I don't want to put it on this PR, but just a thought.

go.mod
Ulexus: Yeah, that's disturbing.  I didn't change that, at least not explicitly.  I _did_ run a `go mod tidy` because the build complained about `go.sum` being modified during the build.  My understanding of go modules is that it should never cross major change boundaries like that from a simple `tidy`.  Perhaps it is because of that version comment?  I have no idea what that is or if it is significant.

internal/app/machined/pkg/runtime/sequencer.go
Ulexus: Yeah, I didn't want to reorder the enum, but you are correct; I don't think these are exposed outside of a release.  It's still not something I was comfortable changing.

go.mod
smira: Probably better to revert `go.mod`, and `go mod tidy` shouldn't touch it much

pkg/machinery/client/client.go
smira: on other hand, to get "raw" API one can always just use `c.MachineClient.ApplyConfiguration` directly. I don't know what's the best way here, I had same thoughts as well

smira (commented):
I don't see the integration tests unless I missed them somewhere...

Ulexus:
Ah, right; I had gotten side-tracked when getting the integration tests together.  I'll see if I can get those done.

api/machine/machine.proto
```diff
@@ -47,6 +48,19 @@ service MachineService {
...
 }
 
+// rpc applyConfiguration
+// ApplyConfiguration describes a request to assert a new configuration upon a node.
+message ApplyConfigurationRequest {
+  bytes data = 1;
+
+  // enum op = 2; // (perhaps) patch, replace, etc
+
+  // bool no_reboot = 3; // (perhaps) do not reboot after application of new configuration
```
andrewrynhard, line 58: Should we drop this idea into a ticket and remove the comments? Nitpicky, I know, but I fund that commented code is easy to lose.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -40,6 +40,31 @@ func (p PhaseList) AppendWhen(when bool, name string, tasks ...runtime.TaskSetup
 	return p
 }
 
+// ApplyConfiguration is a sequence which is called to apply a new configuration to the system.
+// It will reboot subsequent make it active.
```
andrewrynhard, line 44: nit: This sentence doesn't read well.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -40,6 +40,31 @@ func (p PhaseList) AppendWhen(when bool, name string, tasks ...runtime.TaskSetup
...
+func (*Sequencer) ApplyConfiguration(r runtime.Runtime, req *machineapi.ApplyConfigurationRequest) []runtime.Phase {
+	phases := PhaseList{}
+
+	phases = phases.Append(
+		"mountState",
+		MountStatePartition,
+	).Append(
+		"saveConfig",
+		SaveConfig,
+	).Append(
```
andrewrynhard, line 54: I think we should do exactly what reboot sequence does once we are done saving the config:

```
		phases = phases.Append(
			"stopEverything",
			StopAllServices,
		).Append(
			"umount",
			UnmountOverlayFilesystems,
			UnmountPodMounts,
		).Append(
			"unmountSystem",
			UnmountEphemeralPartition,
			UnmountStatePartition,
		).Append(
			"unmountBind",
			UnmountSystemDiskBindMounts,
		).Append(
			"reboot",
			Reboot,
		)
```

api/machine/machine.proto
Ulexus: I really only put those there to support the idea of _having_ a request container, rather than as serious suggestions of features, actually.  I don't think they are _worthy_ of being made into issues, but I'm happy to remove the comments.  I think it's probably fairly self-evident what future possibilities there may be.

andrewrynhard: Up to you, not a blocker for me.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
Ulexus: See if this is better now.

andrewrynhard (dismissed):

smira (commented):
lgtm, but I wish there was an integration test ;)

Ulexus:
I really am working on the integration test; it's just slow going. :)

andrewrynhard (approved):

internal/integration/api/apply-config.go
```diff
@@ -0,0 +1,169 @@
...
+	suite.Require().NotEmpty(nodes)
+
+	sort.Strings(nodes)
+
+	node := nodes[0]
+
+	nodeCtx := client.WithNodes(suite.ctx, node)
+
+	provider, err := suite.readConfigFromNode(nodeCtx)
+	suite.Assert().Nilf(err, "failed to read existing config from node %q: %w", node, err)
```
smira, line 88: nit: you might want to change `suite.Assert()` to `suite.Require()` when the test can't continue running after the failure 

e.g. if config failed to be read, the next line will probably panic on `nil`, so it doesn't make much sense to proceed with the test after the failure

smira (approved):
:+1:

PR #2555, opened 2020-09-23
docs: add images command guide

andrewrynhard:
This adds instructions on how to use `talosctl images` to push images into
a registry cache.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2555)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,30 +8,30 @@ ARG PKGS
 
 # Resolve package images using ${PKGS} to be used later in COPY --from=.
 
-FROM docker.io/autonomy/fhs:${PKGS} AS pkg-fhs
-FROM docker.io/autonomy/ca-certificates:${PKGS} AS pkg-ca-certificates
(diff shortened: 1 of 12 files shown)


docs/website/content/v0.7/en/guides/local/registry-cache.md
```diff
@@ -81,6 +81,26 @@ talosctl cluster create --provisioner docker \
     --registry-mirror gcr.io=http://172.17.0.1:5003
 ```
 
+## Adding Talos Images to the Registries
```
rsmitty, line 84: Question about this. Is it only specific to you version of talosctl you've got locally? If so, might be worth noting that.

smira: This doesn't work with caching registries in this guide (they're read-only). We need probably a different guide on how to push to a non-caching registry.

PR #2567, opened 2020-09-25
docs: add note around link-local addressing

rsmitty:
This PR adds a small note to the config docs to detail how to do link
local networking like SLAAC.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2567)

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
@@ -443,6 +443,10 @@ type NetworkConfig struct {
 	//
 	//     > Note: This option is mutually exclusive with CIDR.
 	//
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -443,6 +443,9 @@ type NetworkConfig struct {
 	//
 	//     > Note: This option is mutually exclusive with CIDR.
 	//
+	//     > Note: To do link-local addressing (ex. SLAAC with IPv6), set CIDR to "" and DHCP to false.
+	//     > All other options will still apply.
```
andrewrynhard, line 447: I think this might be worded better. Technically there is nothing the user has to do for SLAAC. It is part of the IPv6 magic. I think maybe what we really have enable is making CIDR optional.

andrewrynhard (approved):

PR #2581, opened 2020-09-28
fix: make Flannel CNI image follow `$PKGS` version

smira:
Flannel CNI image default was hardcoded in the bootkube-plugin, so it
might become outdated as `$PKGS` get rebuilt. Fix that by embedding
`$PKGS` variable with the Talos build.

Note: we might want to print `$PKGS` variable with `talosctl version`
for easier troubleshooting (or version tag is enough?).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2581)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,9 +120,10 @@ RUN ! go mod tidy -v 2>&1 | grep .
 FROM base AS init-build
 ARG SHA
 ARG TAG
+ARG PKGS
 ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
 WORKDIR /src/internal/app/init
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -240,9 +247,10 @@ RUN --security=insecure img build --tag ${USERNAME}/routerd:${TAG} --output type
 FROM base AS bootkube-build
 ARG SHA
 ARG TAG
-ARG VERSION_PKG="github.com/talos-systems/talos/internal/pkg/version"
+ARG PKGS
+ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
```
smira, line 251: this was a bug for quite some time now, which means vars were not set at all

andrewrynhard (approved):

PR #2583, opened 2020-09-29
feat: add support for disabling time

salkin:
Adds the capability to diasable NTP when it cannot be provided in the deployed network

# Pull Request

## Why? (reasoning)

When a cluster is deployed in environment where NTP cannot be provided Talos can be enabled to run without NTP server.

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/2583)

### DIFF
--- a/docs/website/content/v0.7/en/configuration/v1alpha1.md
+++ b/docs/website/content/v0.7/en/configuration/v1alpha1.md
@@ -870,6 +870,13 @@ Valid Values:
 
 ### TimeConfig
 
+#### enabled
+
+Indicates if time (ntp) is enabled for the machine
+Defaults to `true`.
+
+Type: `bool`
+
 #### servers
 
 Specifies time (ntp) servers to use for setting system time.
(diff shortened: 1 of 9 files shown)


smira:
I'm not sure, but I think it might be less of changes and need to keep things depend on the enabled flag, if we just inside ntpd declare things as healthy if time is disabled? and never even start a sync loop?

or we could have an option to keep ntpd enabled, but skip the requirement to wait for the first time sync?

just ideas

andrewrynhard (approved):
I think this is a reasonable request. @smira I could be misunderstanding you points, but I think it is more clear to just not start `timed` at all if it is disabled. If it is started but behaves differently under certain conditions, it will be hard to understand IMHO.

smira:
>  If it is started but behaves differently under certain conditions, it will be hard to understand IMHO.

Yep, agreed, better not to start it

smira (approved):

PR #2588, opened 2020-09-30
feat: support custom filename for talosctl kubeconfig

### LINKED ISSUE #2151: kubeconfig command should support filename
### Description

I want to be able to do something like `talosctl kubeconfig ./out/filename`

smira:
This also refactors much of the CLI code for the `talosctl kubeconfig`:

1. Do all the checks before fetching kubeconfig from the server: as
kubeconfig generation takes a few seconds, it doesn't make sense to
generate it if it's not going to be used.

2. Unify most of merge & write directly features.

3. Don't use ExtractTarGz method to be more flexible.

4. Allow custom paths for kubeconfig, whether it is a directory or full
path to the file to be created.

Fixes #2151

### DIFF
--- a/cmd/talosctl/cmd/talos/kubeconfig.go
+++ b/cmd/talosctl/cmd/talos/kubeconfig.go
@@ -7,7 +7,7 @@ package talos
 import (
 	"context"
 	"fmt"
-	"io"
+	"io/ioutil"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


cmd/talosctl/cmd/talos/kubeconfig.go
```diff
@@ -41,6 +41,60 @@ If merge flag is defined, config will be merged with ~/.kube/config or [local-pa
...
+				}
+
+				err = os.MkdirAll(filepath.Dir(localPath), 0o755)
+				if err != nil {
+					return err
+				}
+			} else if st.IsDir() {
+				// only dir name was given, append `kubeconfig` by default
+				localPath = filepath.Join(localPath, "kubeconfig")
+			}
```
andrewrynhard, line 84: I am torn on this ... IIUC, if the supplied local path is a dir, we use filename `kubeconfig` by default. It gives `local-path` dual meaning, which _could_ be confusing. What about making `local-path` a file name at all times, and if a directory with the same name exists, then error?

andrewrynhard (approved):

cmd/talosctl/cmd/talos/kubeconfig.go
smira: I didn't want to break backwards compatibility with that, and to keep `talosctl kubeconfig` (without args) work as before. But certainly that's not I'm going to die for ;)

andrewrynhard: Ah gotcha. It is not a blocker for me. Eventually I would like to phase that out if we decide not to just break it now.

cmd/talosctl/cmd/talos/kubeconfig.go
```diff
@@ -57,69 +111,28 @@ If merge flag is defined, config will be merged with ~/.kube/config or [local-pa
 			}()
 
 			defer wg.Wait()
+			defer r.Close() //nolint: errcheck
```
smira, line 114: this line missing was leading to `talosctl kubeconfig` hanging on error

PR #2597, opened 2020-10-02
feat: add webconfig service

### LINKED ISSUE #2593: feat: installer should wait for config if none provided
## Feature Request

The installer should be able to be started with no config.  In this case, it should start a web service and wait for a configuration to be provided to it.

Ulexus:
If no configuration source is provided (on baremetal only for now), Talos will set of a simple web service to receive its configuration via an HTTP POST request, on port 80 or 443.
It also serves a simple web form for interactive submission from the browser.

Fixes #2593

# Pull Request

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -21,6 +21,7 @@ import (
 	"github.com/talos-systems/go-blockdevice/blockdevice/probe"
 
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
 	"github.com/talos-systems/talos/pkg/download"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
@@ -43,7 +44,7 @@ func (m *Metal) Name() string {
 func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 	var option *string
 	if option = procfs.ProcCmdline().Get(constants.KernelParamConfig).First(); option == nil {
-		return nil, fmt.Errorf("no config option was found")
+		return nil, errors.ErrNoConfigSource
 	}
 
 	log.Printf("fetching machine config from: %q", *option)
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -40,6 +40,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub"
+	perrors "github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/events"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/services"
@@ -51,6 +52,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/kmsg"
 	"github.com/talos-systems/talos/internal/pkg/kubeconfig"
 	"github.com/talos-systems/talos/internal/pkg/mount"
+	"github.com/talos-systems/talos/internal/pkg/webconfig"
 	"github.com/talos-systems/talos/pkg/cmd"
 	"github.com/talos-systems/talos/pkg/conditions"
 	"github.com/talos-systems/talos/pkg/images"
@@ -428,6 +430,13 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
 			defer ctxCancel()
 
 			b, e := fetchConfig(fetchCtx, r)
+			if errors.Is(e, perrors.ErrNoConfigSource) {
+				b, e = receiveConfigViaWebconfig(ctx, logger, r)
+				if e != nil {
+					return fmt.Errorf("failed to receive config via webconfig: %w", err)
+				}
+			}
+
 			if e != nil {
 				return e
 			}
@@ -539,6 +548,29 @@ func fetchConfig(ctx context.Context, r runtime.Runtime) (out []byte, err error)
 	}
 
 	return b, nil
+}
+
+func receiveConfigViaWebconfig(ctx context.Context, logger *log.Logger, r runtime.Runtime) ([]byte, error) {
+	cfgBytes, err := webconfig.Run(ctx, logger, r.State().Platform().Mode())
+	if err != nil {
+		return nil, fmt.Errorf("failed to receive config via webconfig: %w", err)
+	}
+
+	provider, err := configloader.NewFromBytes(cfgBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create config provider from webconfig bytes: %w", err)
+	}
+
+	if err = provider.Validate(r.State().Platform().Mode()); err != nil {
+		return nil, fmt.Errorf("failed to validate config received from webconfig: %w", err)
+	}
+
+	processedBytes, err := provider.Bytes()
+	if err != nil {
+		return nil, fmt.Errorf("failed to export validated webconfig to bytes: %w", err)
+	}
+
+	return processedBytes, nil
 }
 
 // ValidateConfig validates the config.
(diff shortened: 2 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -532,6 +540,29 @@ func fetchConfig(r runtime.Runtime) (out []byte, err error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to receive config via webconfig: %w", err)
+	}
+
+	provider, err := configloader.NewFromBytes(cfgBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create config provider from webconfig bytes: %w", err)
+	}
+
+	if err = provider.Validate(r.State().Platform().Mode()); err != nil {
```
smira, line 554: just in terms of the UX, if we could validate in the `webconfig` and return error back via HTTP response so that user could retry, I think it might be easier

internal/pkg/webconfig/webconfig.go
```diff
@@ -0,0 +1,189 @@
...
+
+	s.configBytes, err = ioutil.ReadAll(f)
+	if err != nil {
+		s.logger.Println("webconfig: failed to read file from config POST:", err)
+		w.WriteHeader(http.StatusInternalServerError)
+
+		return
+	}
+
+	if _, err = w.Write([]byte(`{ "accepted": "ok"}`)); err != nil {
```
smira, line 115: we might want to set Content-Type here explicitly to make it display nicely?

or should it be HTML in fact to show that nicely in the web UI?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Ulexus: I would strongly agree, but it appears that Validation is tied to the Platform.

internal/pkg/webconfig/webconfig.go
Ulexus: Yeah, it probably should be just plain html.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Ulexus: Okay, this is kind of weird.  There are two different Mode interfaces, but it looks like they are compatible.

andrewrynhard:
I know this is still a draft, but I had an idea. I wonder if it would be useful to users to print out some basic information, such as the node's IP, interfaces, etc. so that they can more easily make decisions on how to configure Talos.

Ulexus:
Yeah, I think Andrey had mentioned this, and I just forgot about it.  It is a good and useful idea.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -29,6 +29,19 @@ const (
...
+func (ncs *noConfigSource) Error() string {
+	return "no configuration source"
+}
+
+func (ncs *noConfigSource) NoConfigurationSource() {
+}
+
+func errNoConfigSource() error {
+	return new(noConfigSource)
+}
```
andrewrynhard, line 43: I noticed this is also defined in `container.go`. Maybe we could find a common place?

internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
```diff
@@ -23,6 +23,21 @@ import (
...
+	NoConfigurationSource()
+}
+
+// IsNoConfigSourceError indicates whether the given error contains a No Configuration Source defined error.
+func IsNoConfigSourceError(err error) bool {
+	_, ok := err.(NoConfigSourceError)
+
+	return ok
+}
+
```
andrewrynhard, line 40: We should be able to skip this and just use `errors.Is` no? Then `container` and `metal` platforms can simply return `platform.ErrNoConfigSource`.

internal/pkg/webconfig/webconfig.go
```diff
@@ -0,0 +1,288 @@
...
+		TLSConfig: tlsConfig,
+	}
+
+	go func() {
+		if err = s.svr.ListenAndServe(); err != nil {
+			s.logger.Println("webconfig: HTTP service stopped:", err)
+		}
+
+		s.Stop()
+	}()
```
andrewrynhard, line 217: I know this is already an insecure service, but I wonder if we should do unencypted.

internal/pkg/webconfig/webconfig.go
```diff
@@ -0,0 +1,288 @@
...
+	tlsConfig, err := ttls.New(
+		ttls.WithClientAuthType(ttls.ServerOnly),
+		ttls.WithCACertPEM(caProvider),
+		ttls.WithServerCertificateProvider(provider),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate tlsconfig: %w", err)
+	}
+
+	// We need a broader set of ciphers to interact with browsers.
```
andrewrynhard, line 281: :+1:

andrewrynhard (commented):
:+1: Excited about this. Thinking about testing it, I wonder if a simple enough test would be to use `docker run` and to try and create a single node cluster. A script could `docker run` and try to post the config in a `while` loop (or `until`). Not sure I have any other better ideas. An alternative is that we consider this feature experimental for v0.7 and figure it testing, and the whole "interactive installer" bit in v0.8.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
Ulexus: Yeah, a common place would definitely be better.  Unfortunately, `platform` is not it, because platform imports each of these, thus creating an import loop.

internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
Ulexus: The whole reason I have this roundabout interface is because we can't import `platform` from the platform implementations, because `platform` imports each of these.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
andrewrynhard: Ah, well :hankey: . :thinking: what if we had `platform/errors`?

internal/pkg/webconfig/webconfig.go
Ulexus: Good question.  I had implemented insecure first, and I just didn't remove it when I added the self-signed cert.  Probably just make it a 302 redirect to the https service.

andrewrynhard: :+1: better idea.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
Ulexus: Works for me.

internal/pkg/webconfig/webconfig.go
```diff
@@ -0,0 +1,296 @@
...
+
+const indexPage = `
+<html>
+	<head>
+		<title>Talos Interactive Configurator</title>
+	</head>
+	<body>
+	  <h3>Talos Interactive Configuration</h3>
+	  <form action="/" method="POST" enctype="multipart/form-data">
+	  <p>Upload configuration: <input type="file" name="config" /></p>
```
smira, line 37: Nit: we could call the field config.yaml, so it's easier to upload it from the command line?

internal/pkg/webconfig/webconfig.go
```diff
@@ -0,0 +1,296 @@
...
+
+const configurationRejectedPage = `
+<html>
+  <head>
+    <title>Talos Interactive Configurtor: configuration rejected</title>
+  </head>
+  <body>
+    <p>The configuration was rejected:</p>
+	 <pre>
+	 {{.}}
```
smira, line 52: It might be better to escape HTML here with {{ html . }}

smira (commented):
Just small nits

andrewrynhard (approved):
LGTM. Given the lack of time, how do you feel about us labeling this as "experimental" for now? My thought is that we can make it more interactive in 0.8, and get it tested.

Ulexus:
Sounds good to me.  I think that describes it pretty well, anyway.  We can use this in a lot of ways, and it will probably take a bit to figure out how, exactly, it should be used.  There is also the seeming functional overlap with the ApplyConfiguration API and the difficulty of a bunch of other required bits currently required to be supplied which might be able to be rolled into the config (such as the talos.platform kernel commandline argument, and others).

Ulexus:
Is there some standard way to "mark it as experimental"?

andrewrynhard:
> Is there some standard way to label it as "experimental"?

No. It is more or less how we pitch it in the release notes and such. Also in our engagement with users. I have long wanted a feature gate system, but we don't have that yet.

PR #2600, opened 2020-10-03
feat: add etcd API

andrewrynhard:
This adds RPCs for basic etcd management tasks.


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1200,3 +1200,143 @@ func (s *Server) Memory(ctx context.Context, in *empty.Empty) (reply *machine.Me
 
 	return reply, err
 }
+
+// EtcdMemberList implements the machine.MachineServer interface.
```
andrewrynhard, line 1204: All of these methods should use a common package. Some of this code is copied from the tasks file.

andrewrynhard:
I'm not sure the best way to test this. Ideas? Maybe run etcd API tests at the end since it will be destructive?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1200,3 +1202,211 @@ func (s *Server) Memory(ctx context.Context, in *empty.Empty) (reply *machine.Me
...
+		},
+	}
+
+	return reply, nil
+}
+
+// EtcdMemberRemove implements the machine.MachineServer interface.
+//
+// nolint: gocyclo
+func (s *Server) EtcdMemberRemove(ctx context.Context, in *machine.EtcdMemberRemoveRequest) (reply *machine.EtcdMemberRemoveResponse, err error) {
```
smira, line 1241: nit, but should it be `EtcdMemberRemoveSelf`, as it only supports removing this node from the members?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1200,3 +1202,211 @@ func (s *Server) Memory(ctx context.Context, in *empty.Empty) (reply *machine.Me
...
+
+	if id == nil {
+		return nil, fmt.Errorf("failed to find %q in list of etcd members", hostname)
+	}
+
+	log.Println("leaving etcd cluster")
+
+	_, err = client.MemberRemove(ctx, *id)
+	if err != nil {
+		return nil, err
```
smira, line 1279: nit: context in errors might help with debugging later on

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1200,3 +1202,211 @@ func (s *Server) Memory(ctx context.Context, in *empty.Empty) (reply *machine.Me
...
+			return nil, fmt.Errorf("leader endpoint is required")
+		}
+
+		client.SetEndpoints(leaderEp)
+
+		for _, m := range resp.Members {
+			if m.GetID() != member.GetID() {
+				leader = m
+			}
+		}
```
smira, line 1385: should we check for `leader` being nil, as this might happen if we have a single node cluster?

internal/integration/api/etcd.go
```diff
@@ -0,0 +1,137 @@
...
+
+	nodes := suite.DiscoverNodes().NodesByType(machine.TypeControlPlane)
+
+	var leader string
+
+	for _, node := range nodes {
+		resp, err := suite.Client.MachineClient.EtcdMoveLeader(client.WithNodes(context.TODO(), node), &machineapi.EtcdMoveLeaderRequest{})
+		suite.Require().NoError(err)
+
+		if resp.Messages[0].GetMember() != "" {
```
smira, line 69: probably you might need to assert here that `resp.Messages[0].Metadata.Error` is empty, as any errors are ignored now

internal/integration/api/etcd.go
```diff
@@ -0,0 +1,137 @@
...
+	if suite.Cluster == nil {
+		suite.T().Skip("without full cluster state etcd test is not reliable (can't wait for cluster readiness in between resets)")
+	}
+
+	nodes := suite.DiscoverNodes().NodesByType(machine.TypeControlPlane)
+
+	var leader string
+
+	for _, node := range nodes {
+		resp, err := suite.Client.MachineClient.EtcdMoveLeader(client.WithNodes(context.TODO(), node), &machineapi.EtcdMoveLeaderRequest{})
```
smira, line 66: you have suite.ctx here to abort properly

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
andrewrynhard: Added an explicit check for this at the top of the func.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
andrewrynhard: How about `EtcdForfeitLeadership`?

smira: I think this is about removing a member, but not really leadership?

I'm not a wordsmit, could be `EtcdLeave` ?

andrewrynhard: Fixed.

smira (dismissed):

smira (approved):

Ulexus (approved):

PR #2606, opened 2020-10-06
chore: remove build tags from integration binary

andrewrynhard:
This will make development of test much easier.


smira:
```
{
    "go.buildTags": "integration,integration_api,integration_cli,integration_k8s",
    "gopls.env": {
        "GOFLAGS": "-tags=integration,integration_api,integration_cli,integration_k8s"
    }
}
```

this seems to fix the issue with gopls for me

smira:
link: https://github.com/golang/go/issues/29202#issuecomment-488469829

smira:
I wonder if we should fix the golangci-lint by pushing build tags (looks like it ignored the files?)

PR #2607, opened 2020-10-06
feat: allow disabling NoSchedule on master nodes

Unix4ever:
Add talosconfig parameter that allows to disable NoSchedule taint on
master nodes.
https://github.com/talos-systems/talos/issues/2573

### DIFF
--- a/docs/website/content/v0.7/en/configuration/v1alpha1.md
+++ b/docs/website/content/v0.7/en/configuration/v1alpha1.md
@@ -643,6 +643,19 @@ adminKubeconfig:
 
 ```
 
+#### allowSchedulingOnMasters
+
+Indicates if master nodes are schedulable.
+
+Type: `bool`
+
+Valid Values:
+
+- `true`
+- `yes`
+- `false`
+- `no`
+
 ---
 
 ### KubeletConfig
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -23,11 +23,11 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/hashicorp/go-multierror"
 	"go.etcd.io/etcd/clientv3"
 	"golang.org/x/sys/unix"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 
+	multierror "github.com/hashicorp/go-multierror"
 	"github.com/talos-systems/go-retry/retry"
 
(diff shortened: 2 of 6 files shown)


pkg/machinery/config/types/v1alpha1/generate/init.go
```diff
@@ -77,6 +77,7 @@ func initUd(in *Input) (*v1alpha1.Config, error) {
 		ClusterCA:                     in.Certs.K8s,
 		BootstrapToken:                in.Secrets.BootstrapToken,
 		ClusterAESCBCEncryptionSecret: in.Secrets.AESCBCEncryptionSecret,
+		NoScheduleMasters:             true,
```
smira, line 80: I think we might need to inverse this boolean option, so that existing Talos installs don't get reverted back to missing `NoSchedule`, in other words `NoSchedule` taint should be the default zero value of the config

artem-ms: Good point. Will update

docs/website/content/v0.7/en/configuration/v1alpha1.md
```diff
@@ -643,6 +643,19 @@ adminKubeconfig:
 
 ```
 
+#### scheduleOnMasters
+
+Indicates if master nodes should have taint NoSchedule.
```
smira, line 648: my English is bad, but the phrase here seems to be confusing,  the meaing should be that`scheduleOnMasters` disables `NoSchedule` on masters.

@andrewrynhard help please :)

pkg/machinery/config/types/v1alpha1/generate/init.go
```diff
@@ -77,6 +77,7 @@ func initUd(in *Input) (*v1alpha1.Config, error) {
 		ClusterCA:                     in.Certs.K8s,
 		BootstrapToken:                in.Secrets.BootstrapToken,
 		ClusterAESCBCEncryptionSecret: in.Secrets.AESCBCEncryptionSecret,
+		ScheduleOnMastersEnabled:      false,
```
smira, line 80: super nit: could be removed (as it's anyways zero value)

smira (approved):
just small nits/docs

docs/website/content/v0.7/en/configuration/v1alpha1.md
andrewrynhard: It isn't too bad, but I think it could be better. I will propose making `scheduleOnMasters` `allowSchedulingOnMasters`, and then the description could be:

```
Indicates if master nodes are schedulable.
```

@rsmitty @Ulexus feel free to chime in.

andrewrynhard (approved):

rsmitty (approved):

PR #2608, opened 2020-10-06
feat: colorize output of cluster health checks

smira:
It only gets enabled if output is a terminal. Failures which resolve
themselves are removed from the final output. Small spinner to indicate
progress.

While I was at it, I fixed client-side `talosctl health` when init node
is missing.


rsmitty (approved):

smira:
https://asciinema.org/a/yraemp3zjFQ1dAjYdliqJjzV2

smira:
> cool I wonder if it's also possible to show percentage of creation progress.

I hope we can do that, we need a progress indicator which covers the whole process of `cluster create`, I just touched the health check progress

Issue #2318, opened 2020-07-16
add context to platform interface


smira:
## Feature Request

Many methods of the platform (e.g. `Hostname()`) perform HTTP requests, and context should be passed there for proper cancellation and timeouts.

PR #2614, opened 2020-10-07
chore: add Context as param to some methods of `Platform` interface

Unix4ever:
Context is passed there for proper cancellation and timeouts.

https://github.com/talos-systems/talos/issues/2318

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -422,7 +422,10 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
 		download := func() error {
 			var b []byte
 
-			b, e := fetchConfig(r)
+			fetchCtx, ctxCancel := context.WithTimeout(context.Background(), 70*time.Second)
+			defer ctxCancel()
+
+			b, e := fetchConfig(fetchCtx, r)
 			if e != nil {
 				return e
 			}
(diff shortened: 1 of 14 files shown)


smira:
> #2318

nit: if you do `Fixes #2318`, it automatically links PR to the issue

pkg/download/download.go
```diff
@@ -82,6 +83,12 @@ func Download(endpoint string, opts ...Option) (b []byte, err error) {
 	}
 
 	err = retry.Exponential(60*time.Second, retry.WithUnits(time.Second), retry.WithJitter(time.Second)).Retry(func() error {
+		select {
+		case <-ctx.Done():
+			return context.Canceled
```
smira, line 88: it might make sense to add here `retry.UnexpectedError(context.Canceled)` or `retry.UnexpectedError(context.Error())`

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -422,7 +422,7 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
 		download := func() error {
 			var b []byte
 
-			b, e := fetchConfig(r)
+			b, e := fetchConfig(ctx, r)
```
smira, line 425: I don't think that we ever set timeout for each task, so it might make some sense to add 

`ctx, cancel = context.WithTimeout(ctx, 30*time.Second)` to make sure Talos doesn't hang forever

smira (commented):
:+1:

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Unix4ever: in each usage, right?

smira: I thought about adding it here, as we make call to `fetchConfig`

In other words, Talos shouldn't hang forever on downloading the config or any other query to cloud metadata server

smira (dismissed):

smira (dismissed):

smira (approved):

PR #2619, opened 2020-10-08
fix: random failures in cluster health checks

smira:
The problem was that some of the health checks sort the list of the
nodes in place (via `sort.Strings()`). If cluster info provider returns
original slice, it might be mutated in such a way that it gets
corrupted.

We never noticed it before CAPI clusters, as in our tests IPs are
assigned sequentially, and sort operation is a no-op.

Specifically, the problem was with the `Nodes()` function, it returns
`append(controlPlaneNodes, workerNodes...)` slice, which by definition
might share memory with `controlPlaneNodes` slice. For example,
if control plane nodes were `4, 5, 6` and worker nodes were `3`, the
returned slice will be `4, 5, 6, 3`, and it shares memory with
`controlPlaneNodes` slice (firs three items). If we apply `sort` to the
returned slice, it re-orders it as `3, 4, 5, 6`, but as it is done
in-place, the `controlPlaneNodes` slice is now `3, 4, 5`, which is
obviously wrong.

Fix that by always returning a copy of the slice from the functions
implementing `ClusterInfo` interface.

### DIFF
--- a/cmd/talosctl/cmd/talos/health.go
+++ b/cmd/talosctl/cmd/talos/health.go
@@ -49,9 +49,9 @@ func (cluster *clusterNodes) NodesByType(t machine.Type) []string {
 
 		return []string{cluster.InitNode}
 	case machine.TypeControlPlane:
-		return cluster.ControlPlaneNodes
+		return append([]string(nil), cluster.ControlPlaneNodes...)
 	case machine.TypeJoin:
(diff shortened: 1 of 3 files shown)


Unix4ever (approved):
nice catch!

rsmitty (approved):

PR #2636, opened 2020-10-13
feat: support disk usage command in talosctl

Unix4ever:
Usage example:

```bash
talosctl du --nodes 10.5.0.2 /var -H -d 2
NODE       NAME
10.5.0.2   8.4 kB   etc
10.5.0.2   1.3 GB   lib
10.5.0.2   16 MB    log
10.5.0.2   25 kB    run
10.5.0.2   4.1 kB   tmp
10.5.0.2   1.3 GB   .
```

Supported flags:
- `-a` writes counts for all files, not just directories.
- `-d` recursion depth
- `-H` humanize size outputs.
- `-t` size threshold (skip files if < size or > size).

Fixes: https://github.com/talos-systems/talos/issues/2504


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+		}
+
+		path = strings.TrimSuffix(path, OSPathSeparator)
+		if path == "" {
+			path = "/"
+		}
+
+		_, err := os.Stat(path)
+		if err == os.ErrNotExist {
+			return fmt.Errorf("%s no such file or directory", path)
```
smira, line 527: should we send an error back here and `continue` with other `req.Paths` here?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+		}
+
+		files, err := archiver.Walker(obj.Context(), path, archiver.WithMaxRecurseDepth(-1))
+		if err != nil {
+			return err
```
smira, line 532: same comment as above

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+			skip = skip || !isDir && !req.All
+			// threshold check
+			skip = skip || req.Threshold > 0 && info.Size < req.Threshold
+			skip = skip || req.Threshold < 0 && info.Size > -req.Threshold
+
+			if skip {
+				return
+			}
+
+			err = obj.Send(info)
```
smira, line 557: once we send the info back to the client, should we remove it from the `folders` map to control the memory usage?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+			}
+
+			if err != nil {
+				return err
+			}
+		}
+
+		if path != "" {
+			p := strings.TrimRight(path, "/")
+			if folder, ok := folders[p]; ok {
```
smira, line 635: if we remove sent items, probably we could send the contents of the `folders`?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+			return err
+		}
+
+		folders := map[string]*machine.DiskUsageInfo{}
+
+		// send a record back to client if the message shouldn't be skipped
+		// at the same time use record information for folder size estimation
+		sendSize := func(info *machine.DiskUsageInfo, depth int32, isDir bool) {
+			prefix := strings.TrimRight(filepath.Dir(info.Name), "/")
+			if folder, ok := folders[prefix]; ok {
```
smira, line 541: should we increase all the folder sizes walking "up" the filesystem hiearachy here? 

(unless I missed something, I don't see how `/bin` size gets incremented for `/bin/directory/file` ?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -505,6 +505,147 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
...
+					prefix = nextPrefix
+				}
+
+				if fi.FileInfo.IsDir() {
+					prefix = fi.FullPath
+				}
+				depth = currentDepth
+			}
+
+			if err != nil {
```
smira, line 628: should this be moved closer to `Send()` above, as it's the only error as it seems?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
Unix4ever: it should work.
It should report it like that for example

`100 /bin/directory/file` -> add 100 to `/bin/directory` size.
`4196 /bin/directory` -> add 4196 to `/bin` size.

smira: ahh, got it, it will report subdirectory overall size and add it back to the parent directory

Unix4ever:
it's actually has both right now :)

rsmitty:
Oh good grief, I'm just blind 🤦

api/machine/machine.proto
```diff
@@ -270,6 +271,20 @@ message ListRequest {
   int32 recursion_depth = 3;
 }
 
+// DiskUsageRequest describes a request to list disk usage of directories and regular files
+message DiskUsageRequest {
+  // RecursionDepth indicates how many levels of subdirectories should be
+  // recursed.  The default (0) indicates that no limit should be enforced.
+  int32 recursion_depth = 1;
+  // All write counts for all files, not just directories.
```
smira, line 279: `Report all files, not only directories?` (I think `counts` is wrong, as we report `sizes`)

smira (dismissed):
with nits

pkg/machinery/client/client.go
```diff
@@ -563,6 +563,11 @@ func (c *Client) LS(ctx context.Context, req *machineapi.ListRequest) (stream ma
 	return c.MachineClient.List(ctx, req)
 }
 
+// DU implements the proto.OSClient interface.
+func (c *Client) DU(ctx context.Context, req *machineapi.DiskUsageRequest) (stream machineapi.MachineService_DiskUsageClient, err error) {
```
andrewrynhard, line 567: I know we set a bit example with `list` (function is named `LS`), but since this is new I think we should have the name match the `MachineClient` name (`DiskUsage`).

andrewrynhard (commented):
Just one nit, but excited about this one!

andrewrynhard (approved):

PR #2644, opened 2020-10-15
test: add unit-test for the installer manifest

smira:
This test only works on local machine (see notes in the file).

### DIFF
--- a/internal/pkg/loopback/loopback.go
+++ b/internal/pkg/loopback/loopback.go
@@ -1,0 +1,126 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package loopback provides support for disk loopback devices (/dev/loopN).
+package loopback
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+// Copyright (c) 2017, Paul R. Tagliamonte <paultag@gmail.com>
+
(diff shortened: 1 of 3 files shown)


andrewrynhard (approved):

internal/pkg/loopback/loopback.go
```diff
@@ -0,0 +1,126 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
```
andrewrynhard, line 36: I'm just curious why we could import this?

smira: it's very old, it doesn't compile (imports unused `log`), uses CGo, while it's not needed with x/sys/unix

https://github.com/paultag/go-loopback/blob/master/loopback.go

So I thought it might be easier just to "vendor" it.

andrewrynhard: 👍

PR #2645, opened 2020-10-15
feat: support metric values for DHCP

rsmitty:
This PR adds a "DHCPOptions" field to the config. This field contains a
single subfield currently, "RouteMetric". Setting this well ensure that
any routes provided from the DHCP server are given this metric upon
injection into the routing table.

### DIFF
--- a/internal/app/networkd/pkg/nic/nic.go
+++ b/internal/app/networkd/pkg/nic/nic.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-multierror"
+	"github.com/insomniacslk/dhcp/dhcpv4"
 	"github.com/jsimonetti/rtnetlink"
 	"github.com/jsimonetti/rtnetlink/rtnl"
 	"github.com/mdlayher/netlink"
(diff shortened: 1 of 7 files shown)


rsmitty:
Looks like: 
```
$ talosctl routes -n 10.5.0.2
NODE       INTERFACE   DESTINATION                     GATEWAY       METRIC
10.5.0.2   eth0        0.0.0.0/0                       10.5.0.1/32   352321536
...
...
```

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -414,3 +406,89 @@ func (n *NetworkInterface) Reset() {
 	// nolint: errcheck
 	n.rtnlConn.LinkDown(link)
 }
+
+// addRoute is a function loosely copied from https://github.com/jsimonetti/rtnetlink/blob/154ecd417600f79d7847278a1e984056dc647acc/rtnl/route.go
+// and merged with our logic on determining gw, src, dst, etc.
+// we need this b/c we need to craft the route message ourselves to add attributes.
+// nolint: gocyclo
```
andrewrynhard, line 413: We should contribute the changes back upstream so that we can get any other other goodies that may be fixed/changed in the future.

andrewrynhard (commented):
I think I could use a walk through on what we pulled from `rtnetlink` and why, for my understanding.

internal/app/networkd/pkg/nic/nic.go
rsmitty: Yep, let's chat in the AM and I'll run you through the current blockers and what we might upstream.

andrewrynhard (dismissed):

andrewrynhard (approved):

PR #2648, opened 2020-10-16
feat: support MTU and route changes for DHCP

### LINKED ISSUE #2533: [networkd] Respect config overrides regardless of DHCP settings
While we should respect the MTU value returned by the DHCP server if there is one, we should also support an MTU specified in the machine config as overriding anything else.

Precedence should be `Config > [DHCP] > 1500`

rsmitty:
This PR updates the behavior of our machine configs with respect to
DHCP-enabled interfaces. Now, if MTU is specified by the user, that
value will take precedence over any setting provided by the DHCP server.

Additionally, any routes specified will be appended to routes specified
by the DHCP server.

Will close #2533

### DIFF
--- a/internal/app/networkd/pkg/address/dhcp.go
+++ b/internal/app/networkd/pkg/address/dhcp.go
@@ -27,6 +27,8 @@ type DHCP struct {
 	Ack         *dhcpv4.DHCPv4
 	NetIf       *net.Interface
 	DHCPOptions config.DHCPOptions
+	Mtu         int
+	RouteList   []config.Route
 }
 
 // Name returns back the name of the address method.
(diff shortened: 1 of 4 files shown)


rsmitty:
Validated this with a local machine config that looked like:
```
...
    network:
        interfaces:
          - interface: eth0
            dhcp: true
            mtu: 1400
            dhcpOptions:
              routeMetric: 2000
            routes:
              - network: 1.2.3.0/24
                gateway: 0.0.0.0/0
...
```

Resulting in routes with the metric I set, and the new 1.2.3.0/24 destination:
```
$ talosctl routes -n 10.5.0.2
NODE       INTERFACE   DESTINATION                     GATEWAY       METRIC
10.5.0.2   eth0        0.0.0.0/0                       10.5.0.1/32   2000
10.5.0.2   eth0        1.2.3.0/24                      0.0.0.0/32    2000
...
```

And interfaces with the specified MTU:
```
$ talosctl interfaces -n 10.5.0.2
NODE       INDEX   NAME   MAC                 MTU     ADDRESS
10.5.0.2   1       lo                         65536   127.0.0.1/8
10.5.0.2   1       lo                         65536   ::1/128
10.5.0.2   4       eth0   de:8f:1d:c3:e4:0c   1400    10.5.0.2/24
10.5.0.2   4       eth0   de:8f:1d:c3:e4:0c   1400    fe80::dc8f:1dff:fec3:e40c/64
```

internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -27,6 +27,8 @@ type DHCP struct {
 	Ack         *dhcpv4.DHCPv4
 	NetIf       *net.Interface
 	DHCPOptions config.DHCPOptions
+	Mtu         int
```
andrewrynhard, line 30: nit: Not sure why linter didn't catch it but idiomatic go would be MTU.

internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -27,6 +27,8 @@ type DHCP struct {
...
+	RouteList   []config.Route
```
andrewrynhard, line 31: nit: This should be `Routes`.

andrewrynhard (approved):
A few small bits, but not blockers for me.

PR #2654, opened 2020-10-19
refactor: bring more control to install.Manifest execution

smira:
This unifies more code paths under the control of `install.Manifest` vs.
being split across the installer and manifest code.

There should be no functional changes now.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -11,12 +11,8 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


andrewrynhard (dismissed):
I know preserve is broken, but just to be clear we will be adding something back in in another PR to enable it again?

andrewrynhard (approved):

smira:
> I know preserve is broken, but just to be clear we will be adding something back in in another PR to enable it again?

Yes, exactly, for now I added a check to deny !force upgrades early

PR #2659, opened 2020-10-20
fix: update handling of ntp disable

rsmitty:
This PR changes the bool for disabling ntp to `disable` instead of the
previous `enable`. We need to do this because customers were seeing
failure in cases where they were defining time servers only, which
results in `enabled: false` when configs get unmarshalled. Users wishing
to disable ntp altogether should now use `disabled: true`.

### DIFF
--- a/pkg/machinery/config/provider.go
+++ b/pkg/machinery/config/provider.go
@@ -171,7 +171,7 @@ type Route interface {
 // Time defines the requirements for a config that pertains to time related
 // options.
 type Time interface {
-	Enabled() bool
+	Disabled() bool
 	Servers() []string
 }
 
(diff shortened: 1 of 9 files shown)


pkg/machinery/config/provider.go
```diff
@@ -171,7 +171,7 @@ type Route interface {
 // Time defines the requirements for a config that pertains to time related
 // options.
 type Time interface {
-	Enabled() bool
+	Disabled() bool
```
smira, line 174: strictly speaking, we didn't have to change the interface, but this way it's more consistent ;)

smira (approved):

pkg/machinery/config/provider.go
rsmitty: yeah, seemed like it could cause too much confusion if it didn't match.

andrewrynhard (approved):

Issue #2387, opened 2020-07-30
Render generated configs with comments


andrewrynhard:
## Feature Request
We should render self-documenting configs with comments.

### Description
When encoding configs we have access to comments (see https://godoc.org/gopkg.in/yaml.v3#Node). We should use this to generate configs that are self-documenting.

PR #2661, opened 2020-10-21
feat: encode comments as part of talosctl generated configs

Unix4ever:
Comments encoding works, defaults encoding works.
Docgen was revamped: now it generates go files.
While markdown files are all handled by
`pkg/machinery/config/encoder/markdown.go`.

Changed scheme for docs. Now it no longer relies on a single `doc.go` in
the root of a package. Instead it can generate separate `*_doc.go` files
for each file in the package. `docgen` now expects to get 3 params
instead of 2. 3rd parameter is used to define a unique method name for
getting the list of structs in the file.

Backward compatibility is supported if we define package name as the 3rd
parameter.

1st parameter no longer scans whole current directory, instead it points
to the particular file that should be processed by docgen.

`talosctl docs` command now supports two flags: `--config` and `--cli`.
They allow generating only docs for v1alpha1 configs or for talosctl. If
no flags are defined, all docs are generated.

Additionally made field types clickable in the output markdown file.

Fixes: #2387 


smira:
this looks pretty exciting to me, looking forward to see that bundled with some example for our real config!

Unix4ever:
Conformance is complaining, but I've created a separate commit to simplify the review process.

Here's the resulting config you can get with the new `talosctl gen config`
```yaml
version: v1alpha1 # Indicates the schema used to decode the contents.
debug: false # Enable verbose logging.
persist: true # Indicates whether to pull the machine config upon every boot.
# Provides machine specific configuration options.
machine:
    type: init # Defines the role of the machine within the cluster.
    token: o7vv80.6tf2f5kn7lu01vs8 # The `token` is used by a machine to join the PKI of the cluster.
    # The root certificate authority of the PKI.
    ca:
        # Artem: removed to improve readability
    # Used to provide additional options to the kubelet.
    kubelet:
        image: docker.io/autonomy/kubelet:v1.19.3 # The `image` field is an optional reference to an alternative kubelet image.
        
        # # The `extraArgs` field is used to provide additional flags to the kubelet.
        # extraArgs:
        #     key: value

        # # The `extraMounts` field is used to add additional mounts to the kubelet container.
        # extraMounts:
        #     - destination: /var/lib/example
        #       type: bind
        #       source: /var/lib/example
        #       options:
        #         - rshared
        #         - ro
    # Used to provide instructions for bare-metal installations.
    install:
        disk: /dev/sda # The disk used to install the bootloader, and ephemeral partitions.
        image: ghcr.io/talos-systems/installer:v0.7.0-alpha.3-59-g0a519bb9-dirty # Allows for supplying the image used to perform the installation.
        bootloader: true # Indicates if a bootloader should be installed.
        wipe: false # Indicates if zeroes should be written to the `disk` before performing and installation.
        
        # # Allows for supplying extra kernel args to the bootloader config.
        # extraKernelArgs:
        #     - a=b
    
    # # Extra certificate subject alternative names for the machine's certificate.

    # # Uncomment this to enable SANs.
    # certSANs:
    #     - 10.0.0.10
    #     - 172.16.0.10
    #     - 192.168.0.10

    # # Used to configure the machine's network.

    # # Network definition example.
    # network:
    #     hostname: worker-1 # Used to statically set the hostname for the host.
    #     # `interfaces` is used to define the network interface configuration.
    #     interfaces:
    #         - interface: "" # The interface name.
    #           cidr: "" # The CIDR to use.
    #           # A list of routes associated with the interface.
    #           routes: []
    #           bond: null # Bond specific options.
    #           # VLAN specific options.
    #           vlans: []
    #           mtu: 0 # The interface's MTU.
    #           dhcp: false # Indicates if DHCP should be used.
    #           ignore: false # Indicates if the interface should be ignored.
    #           dummy: false # Indicates if the interface is a dummy interface.
    #           dhcpOptions: null # DHCP specific options.
    #     # Used to statically set the nameservers for the host.
    #     nameservers:
    #         - 9.8.7.6
    #         - 8.7.6.5
    #     # Allows for extra entries to be added to /etc/hosts file
    #     extraHostEntries:
    #         - ip: 192.168.1.100 # The IP of the host.
    #           # The host alias.
    #           aliases:
    #             - test
    #             - test.domain.tld

    # # Used to partition, format and mount additional disks.

    # # MachineDisks list example.
    # disks:
    #     - device: /dev/sdb # The name of the disk to use.
    #       # A list of partitions to create on the disk.
    #       partitions:
    #         - size: 100000000 # This size of the partition in bytes.
    #           mountpoint: //lib/extra # Where to mount the partition.

    # # Allows the addition of user specified files.

    # # MachineFiles usage example.
    # files:
    #     - content: '...' # The contents of file.
    #       permissions: 438 # The file's permissions in octal.
    #       path: /tmp/file.txt # The path of the file.
    #       op: append # The operation to use

    # # The `env` field allows for the addition of environment variables to a machine.

    # # Environment variables definition examples.
    # env:
    #     GRPC_GO_LOG_VERBOSITY_LEVEL: "99"
    #     GRPC_GO_LOG_SEVERITY_LEVEL: info
    #     https_proxy: http://SERVER:PORT/
    # env:
    #     GRPC_GO_LOG_SEVERITY_LEVEL: error
    #     https_proxy: https://USERNAME:PASSWORD@SERVER:PORT/
    # env:
    #     https_proxy: http://DOMAIN\USERNAME:PASSWORD@SERVER:PORT/

    # # Used to configure the machine's time settings.

    # # Example configuration for cloudflare ntp server.
    # time:
    #     disabled: false # Indicates if time (ntp) is disabled for the machine
    #     # Specifies time (ntp) servers to use for setting system time.
    #     servers:
    #         - time.cloudflare.com

    # # Used to configure the machine's sysctls.

    # # MachineSysctls usage example.
    # sysctls:
    #     kernel.domainname: talos.dev
    #     net.ipv4.ip_forward: "0"

    # # Used to configure the machine's container image registry mirrors.
    # registries:
    #     # Specifies mirror configuration for each registry.
    #     mirrors:
    #         docker.io:
    #             # List of endpoints (URLs) for registry mirrors to use.
    #             endpoints:
    #                 - https://registry-1.docker.io
    #     # Specifies TLS & auth configuration for HTTPS image registries.
    #     config:
    #         some.host:123:
    #             # The TLS configuration for this registry.
    #             tls:
    #                 # Enable mutual TLS authentication with the registry.
    #                 clientIdentity:
    #                     crt:
    #                         - 46
    #                         - 46
    #                         - 46
    #                     key:
    #                         - 46
    #                         - 46
    #                         - 46
    #             # The auth configuration for this registry.
    #             auth:
    #                 username: '...' # Optional registry authentication.
    #                 password: '...' # Optional registry authentication.
    #                 auth: '...' # Optional registry authentication.
    #                 identityToken: '...' # Optional registry authentication.
# Provides cluster specific configuration options.
cluster:
    # Provides control plane specific configuration options.
    controlPlane:
        # Endpoint is the canonical controlplane endpoint, which can be an IP address or a DNS hostname.
        endpoint:
            url:
                scheme: http
                opaque: ""
                user: null
                host: 10.5.0.2
                path: ""
                rawpath: ""
                forcequery: false
                rawquery: ""
                fragment: ""
                rawfragment: ""
    clusterName: talos-default # Configures the cluster's name.
    # Provides cluster network configuration.
    network:
        dnsDomain: cluster.local # The domain used by Kubernetes DNS.
        # The pod subnet CIDR.
        podSubnets:
            - 10.244.0.0/16
        # The service subnet CIDR.
        serviceSubnets:
            - 10.96.0.0/12
        
        # # The CNI used.
        # cni:
        #     name: custom # Name of CNI to use.
        #     # URLs containing manifests to apply for CNI.
        #     urls:
        #         - https://www.mysweethttpserver.com/supersecretcni.yaml
    token: 1pjpa5.dlf6j2iarx2fd61e # The [bootstrap token](https://kubernetes.io/docs/reference/access-authn-authz/bootstrap-tokens/).
    aescbcEncryptionSecret: 67s7fTJDAGSuGwWvXQ+2A4CaGQBC5wkTlP/AdO6qvpk= # The key used for the [encryption of secret data at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/).
    # The base64 encoded root certificate authority used by Kubernetes.
    ca:
      # Artem: omitted for the sake of readability
    # API server specific configuration options.
    apiServer:
        image: k8s.gcr.io/kube-apiserver-amd64:v1.19.3 # The container image used in the API server manifest.
    # Controller manager server specific configuration options.
    controllerManager:
        image: k8s.gcr.io/kube-controller-manager-amd64:v1.19.3 # The container image used in the controller manager manifest.
    # Kube-proxy server-specific configuration options
    proxy:
        image: k8s.gcr.io/kube-proxy-amd64:v1.19.3 # The container image used in the kube-proxy manifest.
    # Scheduler server specific configuration options.
    scheduler:
        image: k8s.gcr.io/kube-scheduler-amd64:v1.19.3 # The container image used in the scheduler manifest.
    # Etcd specific configuration options.
    etcd:
        # The `ca` is the root certificate authority of the PKI.
        ca:
          # Artem: omitted
    # # Pod Checkpointer specific configuration options.
    # podCheckpointer:
    #     image: '...' # The `image` field is an override to the default pod-checkpointer image.

    # # Core DNS specific configuration options.
    # coreDNS:
    #     image: '...' # The `image` field is an override to the default coredns image.

    # # A list of urls that point to additional manifests.
    # extraManifests:
    #     - https://www.mysweethttpserver.com/manifest1.yaml
    #     - https://www.mysweethttpserver.com/manifest2.yaml

    # # A map of key value pairs that will be added while fetching the ExtraManifests.
    # extraManifestHeaders:
    #     X-ExtraInfo: info
    #     Token: "1234567"

    # # Settings for admin kubeconfig generation.
    # adminKubeconfig:
    #     certLifetime: 1h0m0s # Admin kubeconfig certificate lifetime (default is 1 year).
```

docs/website/content/v0.7/en/configuration/v1alpha1.md
```diff
@@ -133,139 +122,146 @@ This means it will host etcd along with the Kubernetes master components such as
...
+    # extraHostEntries:
+    #     - ip: 192.168.1.100 # The IP of the host.
+    #       # The host alias.
+    #       aliases:
+    #         - test
+    #         - test.domain.tld
 ```
 
-#### disks
+#### disks <code>[][MachineDisk](#machinedisk)</code>
```
Unix4ever, line 229: Using `<code>` tag here, because you can't insert links into `` contents.

Dockerfile
```diff
@@ -87,8 +87,12 @@ COPY ./api/os/os.proto /api/os/os.proto
 RUN protoc -I/api --go_out=plugins=grpc,paths=source_relative:/api os/os.proto
 COPY ./api/cluster/cluster.proto /api/cluster/cluster.proto
 RUN protoc -I/api --go_out=plugins=grpc,paths=source_relative:/api cluster/cluster.proto
-# Gofumports generated files to adjust import order
-RUN gofumports -w -local github.com/talos-systems/talos /api/
```
smira, line 91: is this intentional?

Dockerfile
```diff
@@ -99,6 +103,7 @@ COPY --from=generate-build /api/machine/machine.pb.go /pkg/machinery/api/machine
 COPY --from=generate-build /api/time/time.pb.go /pkg/machinery/api/time/
 COPY --from=generate-build /api/network/network.pb.go /pkg/machinery/api/network/
 COPY --from=generate-build /api/cluster/cluster.pb.go /pkg/machinery/api/cluster/
+COPY --from=generate-build /pkg/machinery/config/types/v1alpha1 /pkg/machinery/config/types/v1alpha1
```
smira, line 106: is there a way to copy here only generated files (without source files), so that we never clobber local edits?

pkg/machinery/api/cluster/cluster.pb.go
```diff
@@ -8,18 +8,16 @@ package cluster
...
 	duration "github.com/golang/protobuf/ptypes/duration"
+	common "github.com/talos-systems/talos/pkg/machinery/api/common"
 	grpc "google.golang.org/grpc"
 	codes "google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
-
-	common "github.com/talos-systems/talos/pkg/machinery/api/common"
+	reflect "reflect"
```
smira, line 19: looks like it's not formatted to our best practices here (talos imports grouped together as the last block)

hack/docgen/main.go
```diff
@@ -163,56 +249,84 @@ func parseFieldType(p interface{}) string {
...
+func render(doc *Doc, dest string) {
+	t := template.Must(template.New("docfile.tpl").Parse(tpl))
+	buf := bytes.Buffer{}
+
+	err := t.Execute(&buf, doc)
+	if err != nil {
+		panic(err)
+	}
+
+	formatted, err := format.Source(buf.Bytes(), format.Options{})
```
smira, line 320: I guess we need to add our options for the current module

smira (commented):
lgtm with few nits!

Dockerfile
Unix4ever: well, no... Weird, I don't remember removing that line

andrewrynhard (approved):

Dockerfile
```diff
@@ -87,8 +87,12 @@ COPY ./api/os/os.proto /api/os/os.proto
 RUN protoc -I/api --go_out=plugins=grpc,paths=source_relative:/api os/os.proto
 COPY ./api/cluster/cluster.proto /api/cluster/cluster.proto
 RUN protoc -I/api --go_out=plugins=grpc,paths=source_relative:/api cluster/cluster.proto
-# Gofumports generated files to adjust import order
-RUN gofumports -w -local github.com/talos-systems/talos /api/
+
+# RUN docgen for machinery config
```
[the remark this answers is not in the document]
Unix4ever: Comment you mean?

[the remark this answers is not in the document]
andrewrynhard: I think I might be confused by `RUN` being all caps. That is a docker command, so I was thinking that this is a commented docker command.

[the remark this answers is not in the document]
Unix4ever: ah, yeah, makes sense

hack/docgen/main.go
Unix4ever: we don't really rely on any internal packages there, so I guess it's fine to skip adding any options. And linter seems to be happy anyways

pkg/machinery/api/cluster/cluster.pb.go
Unix4ever: yeah, that was my bad. Should be fixed now.

andrewrynhard (dismissed):

andrewrynhard (approved):

PR #2663, opened 2020-10-21
fix: implement preserving contents of partition on install

### LINKED ISSUE #2521: grub: A/B upgrades, rollback, preserving config
## Feature Request

During upgrade disk is wiped, but we need to preserve (copy to memory and write back to disk) contents of `/boot` (for A/B) and config.

smira:
This fixes A/B upgrades and rollback API.

Installer manifest supports now an option to preserve partition contents
while disk is being re-partitioned and partitions are re-formatted.

Mount `/boot` partition as needed (to find current label before starting
the installation and in the rollback API).

Fix upgrade API for non-master nodes.

Contents of `/boot`, `/system/state` and META partitions are preserved
in memory while the disk is re-partitioned.

Remove `--save` flag from the installer as it's not being used.

Fixes #2521

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -42,6 +42,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/etcd"
 	"github.com/talos-systems/talos/internal/pkg/kmsg"
 	"github.com/talos-systems/talos/internal/pkg/kubeconfig"
+	"github.com/talos-systems/talos/internal/pkg/mount"
 	"github.com/talos-systems/talos/pkg/archiver"
 	"github.com/talos-systems/talos/pkg/chunker"
 	"github.com/talos-systems/talos/pkg/chunker/stream"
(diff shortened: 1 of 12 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -84,7 +84,6 @@ func (g *Grub) Labels() (current, next string, err error) {
 	}
 
 	if len(matches[0]) != 2 {
-		log.Printf("%+v", matches[0])
```
andrewrynhard, line 87: :slightly_smiling_face:

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -266,13 +281,15 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (reply
 		return nil, fmt.Errorf("error validating installer image %q: %w", in.GetImage(), err)
 	}
 
-	client, err := etcd.NewClientFromControlPlaneIPs(ctx, s.Controller.Runtime().Config().Cluster().CA(), s.Controller.Runtime().Config().Cluster().Endpoint())
-	if err != nil {
-		return nil, fmt.Errorf("failed to create etcd client: %w", err)
-	}
+	if s.Controller.Runtime().Config().Machine().Type() != machinetype.TypeJoin {
```
andrewrynhard, line 284: :+1:

andrewrynhard (approved):
One thing I am concerned about is ensuring that we have enough memory to preserve. Probably not for this PR, but we should validate that before starting the upgrade. It is kind of wrong to check free mem at the start of the upgrade because we could free up a lot when we shutdown all the workload.

smira:
> One thing I am concerned about is ensuring that we have enough memory to preserve. Probably not for this PR, but we should validate that before starting the upgrade. It is kind of wrong to check free mem at the start of the upgrade because we could free up a lot when we shutdown all the workload.

I thought about it, I can do one small optimization to gzip the `META` (as it's anyways mostly zeroes). But on other hand we preserve before making any changes, so if we fail to preserve, install would fail and node will be rebooted (so nothing really bad happens).

Also what I want to clarify is the interaction of `Zero` option and preserve: if `Zero`, should we preserve anything or just skip that?

andrewrynhard:
> > One thing I am concerned about is ensuring that we have enough memory to preserve. Probably not for this PR, but we should validate that before starting the upgrade. It is kind of wrong to check free mem at the start of the upgrade because we could free up a lot when we shutdown all the workload.
> 
> I thought about it, I can do one small optimization to gzip the `META` (as it's anyways mostly zeroes). But on other hand we preserve before making any changes, so if we fail to preserve, install would fail and node will be rebooted (so nothing really bad happens).
> 
> Also what I want to clarify is the interaction of `Zero` option and preserve: if `Zero`, should we preserve anything or just skip that?

As long as the user knows that we can't preserve due to space issues.
I think we should skip zero on preserve maybe? IDK. Good question.

smira:
> I think we should skip zero on preserve maybe? IDK. Good question.

I ended up skipping existing boot partition detection if `Zero` is set. we don't set `Zero` on upgrades, on other hand we never install if boot partition is found, so it's not going to be used for real, but at least it seems more consistent with grub labels and contents of /boot

smira:
Another small change is that now contents of raw partitions are gzipped as well, so memory requirements are minimal now

PR #2667, opened 2020-10-22
feat: extend etcd health check on upgrade

### LINKED ISSUE #1422: Verify that etcd is healthy before performing upgrade
We should only upgrade master nodes is etcd is healthy, and reachable.

Ulexus:
When an etcd node is upgraded, we now perform additional quorum checks.
This is necessary because when etcd nodes are upgraded, they are removed
from membership.  If, for instance, two etcd nodes were to upgrade
simultaneously, quorum may be lost.  This, of course, does not apply to
single-node etcd clusters.

Fixes #1422

# Pull Request

## Acceptance

### DIFF
--- a/internal/pkg/etcd/etcd.go
+++ b/internal/pkg/etcd/etcd.go
@@ -6,13 +6,16 @@ package etcd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log"
+	"net/http"
 	"net/url"
 	"os"
 	"time"
 
 	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"go.etcd.io/etcd/pkg/transport"
 	"google.golang.org/grpc"
@@ -27,6 +30,9 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
+
+// QuorumCheckTimeout is the amount of time to allow for KV operations before quorum is declared invalid.
+const QuorumCheckTimeout = 15 * time.Second
 
 // Client is a wrapper around the official etcd client.
 type Client struct {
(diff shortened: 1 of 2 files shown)


internal/pkg/etcd/etcd.go
```diff
@@ -103,6 +107,36 @@ func (c *Client) ValidateForUpgrade(ctx context.Context, config config.Provider,
...
 			}
 		}
+
+		// Member count must not be an even number, lest the uprgade cause a loss of quorum.
+		if len(resp.Members) % 2 == 0 {
+			return fmt.Errorf("etcd member count (%d) is even; upgrade of this node may cause loss of quorum", len(resp.Members))
+		}
+
+		// Quorum must be good before allowing an upgrade of an etcd member.
+		if err = c.ValidateQuorum(ctx); err != nil {
```
smira, line 117: should we check quorum via each endpoint? this is what etcd docs recommend on upgrade.

We have a way to send requests via apid from the node to other nodes easily, so we can query Services API for every master node and process the result to figure out etcd service health

Ulexus: I would not have presume that to be necessary, since etcd is both secured against split brain and kv operations are secured against loss of quorum.  If that is recommended, though, we should certainly do so.

internal/pkg/etcd/etcd.go
```diff
@@ -103,6 +107,36 @@ func (c *Client) ValidateForUpgrade(ctx context.Context, config config.Provider,
...
+	}
+
+	return nil
+}
+
+// ValidateQuorum performs a KV operation to make certain that quorum is good.
+func (c *Client) ValidateQuorum(ctx context.Context) (err error) {
+
+	// Get a random key. As long as we can get the response without an error, quorum is good.
+	checkCtx, cancel := context.WithTimeout(ctx, QuorumCheckTimeout)
```
smira, line 129: this changes the timeout for etcd service health check (in service health it's driven by the context passed in), so I think it shouldn't be enforced here (or it should be higher).

Ulexus: If the entering context's timeout is shorter, the shorter timeout will still be used.  But a health timeout is different from this quorum check.  If the cluster does not have quorum, this call will block, rather than failing.  Therefore, I believe we _should_ have an additional timeout enforced here, which is separate from the health check timeout.  They are two different things.

smira: no, what I mean is: before your changes (you moved this code from the health check), effective timeout was 15s (driven by the health check loop and settings), now timeout is 2s (as it's lower than 15s). I don't know if 2s is too small or not, but it's a change at least for the service health check.

smira: if 2s is enough for etcd to commit quorum all the time, please ignore my comment ;)

Ulexus: Oh, sorry.  I understand now.  Yeah... I think 15s is too long and also 2s may be too short.  I'll just reset it to 15s for safety.

internal/pkg/etcd/etcd.go
```diff
@@ -102,12 +108,83 @@ func (c *Client) ValidateForUpgrade(ctx context.Context, config config.Provider,
...
+		Reason string `json:"reason"`
+	}
+
+	for _, u := range memberURIs {
+		uri, err := url.Parse(u)
+		if err != nil {
+			continue
+		}
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://%s/health", uri.Host), nil)
```
smira, line 162: I don't know what exactly does this method check in terms of etcd health status (looks like many ways to check health!), but if we follow etcd recommended way, we should check for etcd to be available to commit quorum on every node (including our local node). 

this might be achieved by checking the health of all `service[etcd]` on Talos nodes

as we're inside the cluster here, we can use following trick to build Talos API client: 

https://github.com/talos-systems/talos/blob/master/pkg/cluster/local.go#L34

this client will be fully featured though (can proxy to other nodes, etc.)

with that we could do `client.ServiceInfo(client.WithNodes(ctx, masterNodes...), "etcd")` and parse response like we do in https://github.com/talos-systems/talos/blob/master/pkg/cluster/check/service.go#L129

andrewrynhard: I'm a bit confused on this. Wouldn't this check be only as good as the last time the health check was performed? Wouldn't we need a nearly instant check just before upgrading?

andrewrynhard: In other words, isn't our accuracy is influenced by how long ago the health check was run?

PR #2668, opened 2020-10-22
feat: extend etcd health check on upgrade

### LINKED ISSUE #1422: Verify that etcd is healthy before performing upgrade

Ulexus:
When an etcd node is upgraded, we now perform additional quorum checks.
This is necessary because when etcd nodes are upgraded, they are removed
from membership.  If, for instance, two etcd nodes were to upgrade
simultaneously, quorum may be lost.  This, of course, does not apply to
single-node etcd clusters.

Fixes #1422

# Pull Request

## Acceptance

### DIFF
--- a/internal/pkg/etcd/etcd.go
+++ b/internal/pkg/etcd/etcd.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"go.etcd.io/etcd/clientv3"
+	"go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes"
 	"go.etcd.io/etcd/etcdserver/etcdserverpb"
 	"go.etcd.io/etcd/pkg/transport"
 	"google.golang.org/grpc"
@@ -27,6 +28,9 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
+
+// QuorumCheckTimeout is the amount of time to allow for KV operations before quorum is declared invalid.
+const QuorumCheckTimeout = 15 * time.Second
 
 // Client is a wrapper around the official etcd client.
 type Client struct {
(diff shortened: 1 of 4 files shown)


andrewrynhard (dismissed):

internal/pkg/etcd/etcd.go
```diff
@@ -102,12 +106,49 @@ func (c *Client) ValidateForUpgrade(ctx context.Context, config config.Provider,
...
 				return fmt.Errorf("etcd member %d is not started, all members must be running to perform an upgrade", member.ID)
 			}
+
+			if err = validateMemberHealth(ctx, member.GetClientURLs()); err != nil {
+				return fmt.Errorf("etcd member %d is not healthy; all members must be health to perform an upgrade: %w", member.ID, err)
+			}
+		}
+
+		// Member count must not be an even number, lest the uprgade cause a loss of quorum.
+		if len(resp.Members)%2 == 0 {
```
smira, line 116: this is totally nitpicking, but: if len(members) = 5, quorum is 3, so upgrading even a 4-node cluster (which is even) is fine, as it leaves etcd with 2 nodes for quorum 

so not sure if there's any case except for `len(members) == 2`

smira (dismissed):

internal/pkg/etcd/etcd.go
Ulexus: Yeah, it's true, but I'd _also_ say that in a properly-configured cluster, it means that you are already down at least one master, since the total count should be odd, and the idea here is to only upgrade one etcd node at a time so that we minimize redundancy level changes.

That said, there's always the possibility that someone sets up a _bad_ cluster which has an even number of etcd nodes in its quiescent state, which would then leave the cluster unable to upgrade at all.  One could say that is a _good_ thing, since it enforces a non-even-numbered rule... but that also presents a possibility of unexpected behaviour.

I'm kind of ambivalent about it, either way.... so other opinions?

smira: I don't care much about that, but on other hand we allow clusters to have _any_ number of control plane nodes, and the fact that we don't allow upgrades for clusters with even number of nodes might come as a surprise

Unix4ever (approved):

internal/pkg/etcd/etcd.go
Ulexus: Yeah... we don't really have any control at all on the number of master/etcd nodes we have.  The surprise element is the only thing which concerns me, and I really don't mind enforcing best practices here, so long as the failure cause is visible.

internal/pkg/etcd/etcd.go
```diff
@@ -97,17 +101,53 @@ func (c *Client) ValidateForUpgrade(ctx context.Context, config config.Provider,
...
+		}
+
 		for _, member := range resp.Members {
 			// If the member is not started, the name will be an empty string.
 			if len(member.Name) == 0 {
 				return fmt.Errorf("etcd member %d is not started, all members must be running to perform an upgrade", member.ID)
 			}
+
+			if err = validateMemberHealth(ctx, member.GetClientURLs()); err != nil {
+				return fmt.Errorf("etcd member %d is not healthy; all members must be health to perform an upgrade: %w", member.ID, err)
```
smira, line 115: nit: `must be healthY`

smira (dismissed):

PR #2673, opened 2020-10-23
fix: re-implement upgrade (install) with preserve

### LINKED ISSUE #2522: grub: single-node upgrades
## Feature Request

With single-node upgrades we can't wipe whole disk, so we have to modify partition table and skip removing system partition, as it can't be wiped. 

We have to keep the system partition in place while modifying every other partition to fit the new layout.

smira:
For 0.6 -> 0.7 upgrade, in any case config.yaml is preserved and moved
from `/boot` to `/system/state`.

For single node upgrade, `EPHEMERAL` partition is not touched and other
partitions are re-created as needed.

Bump provision tests to 0.6/0.7 upgrades as we get closer to the new
release.

Fixes #2522


internal/integration/provision/upgrade.go
```diff
@@ -64,11 +64,11 @@ type upgradeSpec struct {
 }
 
 const (
-	stableVersion = "v0.5.1"
-	nextVersion   = "v0.6.0"
+	stableVersion = "v0.6.2"
+	nextVersion   = "v0.7.0-alpha.3"
```
smira, line 68: latest Talos release which has upgrade path from not broken

internal/integration/provision/upgrade.go
```diff
@@ -64,11 +64,11 @@ type upgradeSpec struct {
...
 
-	stableK8sVersion  = "1.18.6"
-	nextK8sVersion    = "1.19.0"
+	stableK8sVersion  = "1.19.0"
+	nextK8sVersion    = "1.19.3"
```
smira, line 71: TODO: re-validate for alpha.3

cmd/installer/pkg/install/manifest.go
```diff
@@ -57,11 +58,23 @@ type Target struct {
...
+	Force  bool
+	Assets []*Asset
+
+	// Preserve contents of the partition with the same label (if it exists).
 	PreserveContents bool
 
+	// Extra preserved locations (for upgrading from older versions of Talos).
+	//
+	// Used only if PreserveContents is true.
+	ExtraPreserveSources []PreserveSource
```
smira, line 71: this allows to copy `/boot/config.yaml` from 0.6 over to `/system/state/config.yaml` in 0.7

cmd/installer/pkg/install/manifest.go
```diff
@@ -227,7 +254,13 @@ func NewManifest(label string, sequence runtime.Sequence, bootPartitionFound boo
 		PartitionType:  LinuxFilesystemData,
 		FileSystemType: FilesystemTypeXFS,
 		Size:           0,
-		Force:          true,
+	}
+
+	if opts.Force {
+		ephemeralTarget.Force = true
+	} else {
+		ephemeralTarget.Skip = true
```
smira, line 262: handling of upgrades with `--preserve`: don't touch ephemeral, and make last partition before ephemeral use all the space towards the ephemeral

cmd/installer/pkg/install/manifest.go
```diff
@@ -252,8 +285,55 @@ func (m *Manifest) Execute() (err error) {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
 //nolint: gocyclo
 func (m *Manifest) executeOnDevice(device Device, targets []*Target) (err error) {
+	if err = m.checkMounts(device); err != nil {
```
smira, line 333: this check allows to detect any mounts from any mount namespace at the start of the installer and allows us to abort early if anything is still mounted

internal/integration/integration_test.go
```diff
@@ -77,6 +77,7 @@ func TestIntegration(t *testing.T) {
 	}
 
 	provision_test.DefaultSettings.CurrentVersion = expectedVersion
+	provision_test.DefaultSettings.CrashdumpEnabled = crashdumpEnabled
```
smira, line 80: this makes crashdump flag actually work for upgrade tests

internal/integration/provision/upgrade.go
```diff
@@ -224,8 +224,12 @@ func (suite *UpgradeSuite) TearDownSuite() {
 
 // setupCluster provisions source clusters and waits for health.
 func (suite *UpgradeSuite) setupCluster() {
-	shortNameHash := sha256.Sum256([]byte(suite.spec.ShortName))
-	clusterName := fmt.Sprintf("upgrade.%x", shortNameHash[:8])
+	defaultStateDir, err := clientconfig.GetTalosDirectory()
```
smira, line 227: relocate state dir to default talosctl location, allows to use `talosctl cluster destroy` to clean up

internal/integration/provision/upgrade.go
```diff
@@ -308,22 +309,14 @@ func (suite *UpgradeSuite) setupCluster() {
...
-
 		request.Nodes = append(request.Nodes,
 			provision.NodeRequest{
 				Name:     fmt.Sprintf("master-%d", i+1),
 				IP:       ips[i],
 				Memory:   DefaultSettings.MemMB * 1024 * 1024,
 				NanoCPUs: DefaultSettings.CPUs * 1000 * 1000 * 1000,
 				DiskSize: DefaultSettings.DiskGB * 1024 * 1024 * 1024,
-				Config:   cfg,
+				Config:   suite.configBundle.ControlPlane(),
```
smira, line 319: stop using init nodes

internal/integration/provision/upgrade.go
```diff
@@ -367,6 +362,11 @@ func (suite *UpgradeSuite) waitForClusterHealth() {
 
 // runE2E runs e2e test on the cluster.
 func (suite *UpgradeSuite) runE2E(k8sVersion string) {
+	if suite.spec.WorkerNodes == 0 {
+		// no worker nodes, should make masters schedulable
+		suite.untaint("master-1")
```
smira, line 367: for single node upgrade test, we need to untaint the master, otherwise sonobuoy doesn't have a place to run

internal/integration/provision/upgrade.go
```diff
@@ -415,22 +415,6 @@ func (suite *UpgradeSuite) readVersion(nodeCtx context.Context, client *taloscli
 	return
 }
 
-func (suite *UpgradeSuite) uncordonNodes() {
```
smira, line 418: we don't need to uncordon anymore manually, Talos handles this on its own

internal/integration/provision/upgrade.go
```diff
@@ -523,6 +536,6 @@ func init() {
 	allSuites = append(allSuites,
 		&UpgradeSuite{specGen: upgradeBetweenTwoLastReleases, track: 0},
 		&UpgradeSuite{specGen: upgradeLastReleaseToCurrent, track: 1},
-		// &UpgradeSuite{specGen: upgradeSingeNodePreserve, track: 0},
+		&UpgradeSuite{specGen: upgradeSingeNodePreserve, track: 0},
```
smira, line 539: that was the motivation for this PR lol

pkg/cluster/check/default.go
```diff
@@ -43,6 +43,12 @@ func DefaultClusterChecks() []ClusterCheck {
 				return ApidReadyAssertion(ctx, cluster)
 			}, 2*time.Minute, 5*time.Second)
 		},
+		// wait for kubelet to be healthy on all
+		func(cluster ClusterInfo) conditions.Condition {
```
smira, line 47: additional check for kubelet, should make sense before we check node status (so that we don't see stale nodes)

pkg/cluster/check/default.go
```diff
@@ -73,14 +79,6 @@ func DefaultClusterChecks() []ClusterCheck {
...
-}
-
-// ExtraClusterChecks returns a set of additional Talos cluster readiness checks which work only for newer versions of Talos.
-//
-// ExtraClusterChecks can't be used reliably in upgrade tests, as older versions might not pass the checks.
-func ExtraClusterChecks() []ClusterCheck {
-	return []ClusterCheck{
 		// wait for all the nodes to be schedulable
 		func(cluster ClusterInfo) conditions.Condition {
 			return conditions.PollingCondition("all k8s nodes to report schedulable", func(ctx context.Context) error {
```
smira, line 84: as we're no longer doing anything for 0.5, we can promote this check to be default

pkg/cluster/check/kubernetes.go
```diff
@@ -231,14 +231,20 @@ func K8sPodReadyAssertion(ctx context.Context, cluster cluster.K8sProvider, name
 	var notReadyPods []string
 
 	for _, pod := range pods.Items {
+		ready := false
```
smira, line 234: this check had a bug - if pod has no `PodReady` condition, it declared pod as ready

pkg/machinery/config/encoder/encoder.go
```diff
@@ -64,7 +64,7 @@ func isSet(value reflect.Value) bool {
 	//nolint:exhaustive
 	switch value.Kind() {
 	case reflect.Ptr:
-		return !value.IsNil() && !value.Elem().IsZero()
+		return !value.IsNil()
```
smira, line 67: @Unix4ever this is the only fix I had to do to make new encoder backwards compatible with the default yaml marshaller, it doesn't seem to affect examples/comments in any bad way

Ulexus (approved):

cmd/installer/pkg/install/install.go
```diff
@@ -125,6 +125,16 @@ func (i *Installer) probeBootPartition() error {
 				defer mount.Unmount(mountpoints) //nolint: errcheck
 			}
 		}
+
+		// try legacy boot partition
+		if !i.bootPartitionFound {
```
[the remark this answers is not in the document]
andrewrynhard: Is it worth leaving a TODO here that comments on this being legacy, and once 0.6 is deprecated we can remove?

andrewrynhard (approved):

andrewrynhard (dismissed):

smira:
> /promote integration

All green except for Cilium, and I believe Cilium is a different problem

andrewrynhard (approved):

PR #2678, opened 2020-10-26
fix: stop etcd on any path on upgrade

smira:
The problem was that etcd stop was only happening in `LeaveEtcd`, thus
upgrade with preserve was never stopping etcd leaving ephemeral
partition still busy.

Refactored code which was stopping service, shutting down all the
services to provide the interface we need:

* stop a service without considering reverse dependencies (force);
* stop a service (services) waiting for reverse dependencies;
* shutdown all the services waiting for reverse dependencies.

### DIFF
--- a/internal/app/machined/pkg/system/system.go
+++ b/internal/app/machined/pkg/system/system.go
@@ -7,6 +7,7 @@ package system
 import (
 	"context"
 	"fmt"
+	"log"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/system/system.go
```diff
@@ -202,33 +203,88 @@ func (s *singleton) Shutdown() {
...
+	s.mu.Lock()
+	if s.terminating {
+		s.mu.Unlock()
+		return nil
 	}
+
+	return s.stopServices(ctx, serviceIDs, false)
+}
+
+// StopWithRevDepenencies will initiate a shutdown of the specified services waiting for reverse dependencies to finish first.
```
andrewrynhard, line 226: :heart:

andrewrynhard (approved):

andrewrynhard (approved):
Do we need a reverse dep shutdown test?

smira:
> Do we need a reverse dep shutdown test?

we have a small one already (as it tests `Shutdown()` method which shares now the same codepath).

PR #2697, opened 2020-10-29
feat: allow specifying user-disks in talosctl cluster create

Unix4ever:
User-disks are supported by QEMU and Firecracker providers.
Can be defined by using the following parameters:
```
--user-disk /mount/path:1GB
```

Can get more than 1 user disk.
Same set of user disks will be created for all master and worker nodes.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -12,9 +12,11 @@ import (
 	"net"
 	"os"
 	stdruntime "runtime"
+	"strconv"
 	"strings"
 	"time"
 
+	humanize "github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/tools/clientcmd"
 
@@ -57,6 +59,7 @@ var (
 	clusterCpus             string
 	clusterMemory           int
 	clusterDiskSize         int
+	clusterDisks            []string
 	targetArch              string
 	clusterWait             bool
 	clusterWaitTimeout      time.Duration
@@ -97,7 +100,6 @@ func create(ctx context.Context) (err error) {
 	}
 
 	memory := int64(clusterMemory) * 1024 * 1024
-	diskSize := int64(clusterDiskSize) * 1024 * 1024
 
 	// Validate CIDR range and allocate IPs
 	fmt.Println("validating CIDR and reserving IPs")
@@ -184,6 +186,11 @@ func create(ctx context.Context) (err error) {
 		provisionOptions = append(provisionOptions, provision.WithDockerPorts(portList))
 	}
 
+	disks, err := getDisks()
+	if err != nil {
+		return err
+	}
+
 	if inputDir != "" {
 		configBundleOpts = append(configBundleOpts, bundle.WithExistingConfigs(inputDir))
 	} else {
@@ -214,6 +221,19 @@ func create(ctx context.Context) (err error) {
 				CNIName: "custom",
 				CNIUrls: []string{customCNIUrl},
 			}))
+		}
+
+		if len(disks) > 1 {
+			// convert provision disks to machine disks
+			machineDisks := make([]*v1alpha1.MachineDisk, len(disks)-1)
+			for i, disk := range disks[1:] {
+				machineDisks[i] = &v1alpha1.MachineDisk{
+					DeviceName:     provisioner.UserDiskName(i + 1),
+					DiskPartitions: disk.Partitions,
+				}
+			}
+
+			genOptions = append(genOptions, generate.WithUserDisks(machineDisks))
 		}
 
 		defaultInternalLB, defaultEndpoint := provisioner.GetLoadBalancers(request.Network)
@@ -273,7 +293,7 @@ func create(ctx context.Context) (err error) {
 			IP:       ips[i],
 			Memory:   memory,
 			NanoCPUs: nanoCPUs,
-			DiskSize: diskSize,
+			Disks:    disks,
 		}
 
 		if i == 0 {
@@ -296,13 +316,15 @@ func create(ctx context.Context) (err error) {
 	}
 
 	for i := 1; i <= workers; i++ {
+		name := fmt.Sprintf("%s-worker-%d", clusterName, i)
+
 		request.Nodes = append(request.Nodes,
 			provision.NodeRequest{
-				Name:     fmt.Sprintf("%s-worker-%d", clusterName, i),
+				Name:     name,
 				IP:       ips[masters+i-1],
 				Memory:   memory,
 				NanoCPUs: nanoCPUs,
-				DiskSize: diskSize,
+				Disks:    disks,
 				Config:   configBundle.Join(),
 			})
 	}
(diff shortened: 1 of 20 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -436,6 +458,61 @@ func parseCPUShare() (int64, error) {
...
+		for j, partition := range partitions {
+			parts := strings.Split(partition, ":")
+			if len(parts) != 2 {
+				return nil, fmt.Errorf("failed to parse malformed partition definition %s", partition)
+			}
+
+			partitionPath := parts[0]
+
+			if !strings.HasPrefix(partitionPath, "/") {
+				return nil, fmt.Errorf("partition path should be absolute")
```
andrewrynhard, line 485: Good check. Something for another PR, but I wonder if this would be a good validation check in the config validator.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -436,6 +458,61 @@ func parseCPUShare() (int64, error) {
...
+			}
+
+			partitionSize, e := strconv.ParseInt(parts[1], 10, -1)
+
+			if e != nil {
+				s, e := humanize.ParseBytes(parts[1])
+
```
andrewrynhard, line 492: nit, extra space... will probably git hit by linter.

pkg/provision/providers/qemu/node.go
```diff
@@ -174,7 +174,7 @@ func (p *provisioner) createNode(state *vm.State, clusterReq provision.ClusterRe
 
 		NanoCPUs: nodeReq.NanoCPUs,
 		Memory:   nodeReq.Memory,
-		DiskSize: nodeReq.DiskSize,
+		DiskSize: nodeReq.Disks[0].Size,
```
andrewrynhard, line 177: I assume that we know for sure that `nodeReq.Disks` is not empty by the time we get here?

pkg/provision/providers/vm/disk.go
```diff
@@ -11,20 +11,52 @@ import (
...
+		remainder := i % 26
+		divider := i / 26
+
+		prefix := ""
+
+		if divider != 0 {
+			prefix = convert(divider - 1)
+		}
+
+		return fmt.Sprintf("%s%s", prefix, string(rune('a'+remainder)))
```
andrewrynhard, line 30: This could probably use a comment on how it works. Its not obvious, at least for me. Where does the number 26 come from?

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -436,6 +458,61 @@ func parseCPUShare() (int64, error) {
...
+			partitionPath := parts[0]
+
+			if !strings.HasPrefix(partitionPath, "/") {
+				return nil, fmt.Errorf("partition path should be absolute")
+			}
+
+			partitionSize, e := strconv.ParseInt(parts[1], 10, -1)
+
+			if e != nil {
+				s, e := humanize.ParseBytes(parts[1])
```
andrewrynhard, line 491: Just double checking, this is backwards compatible?

andrewrynhard (approved):

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -216,6 +223,19 @@ func create(ctx context.Context) (err error) {
 			}))
 		}
 
+		if len(disks) > 1 {
+			// convert provision disks to
```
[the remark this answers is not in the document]
Unix4ever: hah, yeah I forgot to finish that comment.

pkg/provision/providers/qemu/node.go
Unix4ever: yep. CreateDisks will fail otherwise

pkg/provision/providers/vm/disk.go
Unix4ever: oh yeah. It's all how disk naming works:
```
vda
vdb
vdc
...
vdz
vdaa
vdab
...
```

Unix4ever: 26 is amount of letters in the English alphabet.

andrewrynhard: Ah, ok I see it now.

cmd/talosctl/cmd/mgmt/cluster/create.go
Unix4ever: These arguments are new, so shouldn't have any backward compatibility issues.

cmd/talosctl/cmd/mgmt/cluster/create.go
smira: yeah, we could validate in config in fact that it starts with `/var/` ?

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -436,6 +458,63 @@ func parseCPUShare() (int64, error) {
...
+			diskPartitions[partitionIndex] = &v1alpha1.DiskPartition{
+				DiskSize:       uint(partitionSize), // TODO: convert to int64
+				DiskMountPoint: partitionPath,
+			}
+			diskSize += partitionSize
+			partitionIndex++
+		}
+
+		disks = append(disks, &provision.Disk{
+			Size:       diskSize + 1024*1024,
```
smira, line 510: this is extra room for GPT (more than we need, but just in case, we might probably need some comment here to explain extra 1MiB)

cmd/talosctl/cmd/mgmt/cluster/create.go
andrewrynhard: ^ that too... assuming we don't think we have a use case for anything outside of var (_maybe_ `/etc/kubernetes`), and we would have to verify that no one is using it outside of `/var` already.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -459,7 +538,8 @@ func init() {
 	createCmd.Flags().IntVar(&masters, "masters", 1, "the number of masters to create")
 	createCmd.Flags().StringVar(&clusterCpus, "cpus", "2.0", "the share of CPUs as fraction (each container/VM)")
 	createCmd.Flags().IntVar(&clusterMemory, "memory", 2048, "the limit on memory usage in MB (each container/VM)")
-	createCmd.Flags().IntVar(&clusterDiskSize, "disk", 6*1024, "the limit on disk size in MB (each VM)")
+	createCmd.Flags().IntVar(&clusterDiskSize, "disk", 6*1024, "default limit on disk size in MB (each VM)")
+	createCmd.Flags().StringSliceVar(&clusterDisks, "user-disk", []string{}, "list of disks to create for each VM in format: <mount_point1>:<size1>;<mount_point2>:<size2>")
```
smira, line 542: need to update syntax: `;` -> `:` (`;` is bash special character)

andrewrynhard: Shouldn't that be a comma?

Unix4ever: Comma is parsed by Cobra and what is intended to be a single disk with couple partitions, turns into several disks with a single partition.

Unix4ever: yeah, I've updated it in the code, but forgot update help

cmd/talosctl/cmd/mgmt/cluster/create.go
Unix4ever: great idea. Updated.

cmd/talosctl/cmd/mgmt/cluster/create.go
Unix4ever: looks like linter is happy

andrewrynhard: :+1: still just a little awkward, but not a blocker for me.

andrewrynhard (approved):
@smira @Unix4ever Do we want to test this enabling extra disks in the `e2e-qemu.sh` script?

smira:
> @smira @Unix4ever Do we want to test this by enabling extra disks in the `e2e-qemu.sh` script?

I thought about verifying their presence as well in the test

Unix4ever:
Yep, that's the plan. Probably will add it to this PR tomorrow.

internal/integration/cli/mounts.go
```diff
@@ -25,7 +25,7 @@ func (suite *MountsSuite) SuiteName() string {
 // TestSuccess verifies successful execution.
 func (suite *MountsSuite) TestSuccess() {
 	suite.RunCLI([]string{"mounts", "--nodes", suite.RandomDiscoveredNode()},
-		base.StdoutShouldMatch(regexp.MustCompile(`FILESYSTEM`)))
+		base.StdoutShouldMatch(regexp.MustCompile(`FILESYSTEM.*/var/lib/extra`)))
```
smira, line 28: unfortunately we can't do this, as this won't work for any cluster provisioned outside of `e2e-qemu`.

I think we need leave this test as is, but rather another test which is only enabled on some flag/environment variable, and set this flag when running e2e-qemu test, and we can verify that mountpoints are present with expected names

hack/test/e2e-qemu.sh
```diff
@@ -44,6 +44,7 @@ function create_cluster {
     --memory 2048 \
     --cpus 2.0 \
     --cidr 172.20.1.0/24 \
+    --user-disk /var/lib/extra:100MB \
```
smira, line 47: can we add more partitions to test things in a more complete way?

Unix4ever: yep, working on that.

smira (approved):

PR #2699, opened 2020-10-29
fix: properly initialize manifest in user disks creation

smira:
This is :facepalm:, I somehow missed that installer Manifest is used
here and missed the fact I need to update it.


smira:
This was verified with #2697 to test the feature.

```
$ talosctl -n 172.20.0.2 mounts | grep vd
172.20.0.2   /dev/vda5    0.10       0.01       0.09            6.29%          /system/state
172.20.0.2   /dev/vda6    5.91       0.84       5.06            14.25%         /var
172.20.0.2   /dev/vdb1    0.99       0.04       0.95            4.10%          /var/mnt/extra
172.20.0.2   /dev/vdb2    1.99       0.05       1.94            2.40%          /var/mnt/foo
172.20.0.2   /dev/vdc1    0.99       0.04       0.95            4.10%          /var/mnt/third
```

andrewrynhard (approved):

PR #2702, opened 2020-10-29
feat: build talosctl-cni-bundle, use it in talosctl for QEMU

smira:
This builds a bundle with CNI plugins for talosctl which is
automatically downloaded by `talosctl` if CNI plugins are missing.

CNI directories are moved by default to the `~/.talos/cni` path.

Also add a bunch of pre-flight checks to the QEMU provisioner to make it
easier to bootstrap the Talos QEMU cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ SHORT_INTEGRATION_TEST ?=
 CUSTOM_CNI_URL ?=
 
 , := ,
+space := $(subst ,, )
 BUILD := docker buildx build
 PLATFORM ?= linux/amd64
 MULTI_PLATFORM = $(findstring $(,),$(PLATFORM))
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -28,6 +28,7 @@ SHORT_INTEGRATION_TEST ?=
 CUSTOM_CNI_URL ?=
 
 , := ,
+space := $(subst ,, )
```
smira, line 31: `boot-$ARCH` was broken because this variable was missing :facepalm:

website/content/docs/v0.7/Local Platforms/qemu.md
```diff
@@ -13,9 +13,8 @@ In this guide we will create a Kubernetes cluster using QEMU.
   - `CONFIG_NET_SCH_INGRESS` enabled
 - at least `CAP_SYS_ADMIN` and `CAP_NET_ADMIN` capabilities
 - QEMU
-- `bridge`, `static` and `firewall` CNI plugins from the [standard CNI plugins](https://github.com/containernetworking/cni), and `tc-redirect-tap` CNI plugin from the [awslabs tc-redirect-tap](https://github.com/awslabs/tc-redirect-tap) installed to `/opt/cni/bin`
+- `bridge`, `static` and `firewall` CNI plugins from the [standard CNI plugins](https://github.com/containernetworking/cni), and `tc-redirect-tap` CNI plugin from the [awslabs tc-redirect-tap](https://github.com/awslabs/tc-redirect-tap) installed to `/opt/cni/bin` (installed automatically by `talosctl`)
```
Unix4ever, line 16: I will probably override any existing CNI only some CNI files are missing, right?
Maybe it's fine, just don't know if it may mess up CNI for someone.

Unix4ever (approved):

website/content/docs/v0.7/Local Platforms/qemu.md
smira: It installs to `~/.talos/cni/bin`, so it's private location for Talos. And it checks whether plugins are missing, so they could be installed manually as well.

Unix4ever: maybe "installed to `/opt/cni/bin`" is wrong in this description or I just don't get it right.

andrewrynhard: I can fix these docs up. Good point @Unix4ever . Since its Friday, and Andrey is out next week, i think we can merge and I can fix in a follow up.

andrewrynhard (approved):

PR #2707, opened 2020-10-30
fix: read/write human readable representations for bytes and octals

Unix4ever:
Use octal representation to dump FileMode to yaml.
Use 100MB/1GB/1.5TB representations for partitions size, but only if
that doesn't lead to losing precision. Decoding of them should work as
well.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
@@ -19,10 +19,13 @@ package v1alpha1
 import (
 	"net/url"
 	"os"
+	"strconv"
 	"time"
 
+	humanize "github.com/dustin/go-humanize"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/talos-systems/crypto/x509"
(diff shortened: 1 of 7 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -867,11 +869,52 @@ type MachineDisk struct {
...
+// DiskSize partition size in bytes.
+type DiskSize uint64
+
+// MarshalYAML write as human readable string.
+func (ds DiskSize) MarshalYAML() (interface{}, error) {
+	if ds%DiskSize(1000) == 0 {
+		bytesString := humanize.Bytes(uint64(ds))
+		// ensure that stringifying bytes as human readable string
+		// doesn't lose precision
+		parsed, err := humanize.ParseBytes(bytesString)
```
Unix4ever, line 881: I know that encode/decode seems weird, but that's the most robust way to ensure that it's safe to use humanize library for a value.
If that doesn't look great we can also reimplement this method by ourselves.

smira (approved):

PR #2716, opened 2020-11-01
fix: address issues in webconfig

andrewrynhard:
Fixes TLS and invalide redirects.

### DIFF
--- a/internal/pkg/webconfig/webconfig.go
+++ b/internal/pkg/webconfig/webconfig.go
@@ -7,21 +7,19 @@ package webconfig
 import (
 	"context"
 	"crypto/tls"
+	stdx509 "crypto/x509"
 	"fmt"
 	"html/template"
 	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
-	"os"
 	"sync"
 	"time"
 
-	ttls "github.com/talos-systems/crypto/tls"
 	"github.com/talos-systems/crypto/x509"
 	tnet "github.com/talos-systems/net"
 
-	"github.com/talos-systems/talos/pkg/grpc/gen"
 	"github.com/talos-systems/talos/pkg/machinery/config"
 	"github.com/talos-systems/talos/pkg/machinery/config/configloader"
 )
@@ -104,14 +102,6 @@ func (s *Server) Stop() {
 
 // ServeHTTP implements http.Handler.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Scheme != "https" {
-		r.URL.Scheme = "https"
-
-		http.Redirect(w, r, r.URL.String(), http.StatusTemporaryRedirect)
-
-		return
-	}
-
 	switch r.Method {
 	case "GET":
 		s.handleGet(w, r)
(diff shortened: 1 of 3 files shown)


internal/pkg/webconfig/webconfig.go
```diff
@@ -104,14 +102,6 @@ func (s *Server) Stop() {
 
 // ServeHTTP implements http.Handler.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	if r.URL.Scheme != "https" {
-		r.URL.Scheme = "https"
-
-		http.Redirect(w, r, r.URL.String(), http.StatusTemporaryRedirect)
-
-		return
-	}
```
andrewrynhard, line 113: Kept getting invalid redirect errors. Took another approach.

internal/pkg/webconfig/webconfig.go
```diff
@@ -217,7 +217,7 @@ func (s *Server) Run(ctx context.Context) ([]byte, error) {
 	}
 
 	go func() {
-		if err = s.svr.ListenAndServe(); err != nil {
+		if err = http.ListenAndServe(":80", http.HandlerFunc(redirect)); err != nil {
```
andrewrynhard, line 220: Kept it simple and just setup a listener for port 80 that knows how to hand redirects.

internal/pkg/webconfig/webconfig.go
```diff
@@ -232,12 +232,14 @@ func (s *Server) Run(ctx context.Context) ([]byte, error) {
...
-	fmt.Println("Webconfig started.  You may POST a configuration to:")
+	s.logger.Println("Webconfig started. You may POST a configuration to:")
 
 	for _, ip := range ips {
-		fmt.Printf("\t%s\n", ip.String())
+		s.logger.Printf("\thttps://%s\n", ip.String())
 	}
 
+	s.logger.Println("You may also visit any of the above in your browser.")
+
```
andrewrynhard, line 242: I wasn't seeing this printed out in the console on a real machine. The logger is configured to write to the correct place regardless of which mode we are running.

internal/pkg/webconfig/webconfig.go
```diff
@@ -247,49 +249,33 @@ func (s *Server) Run(ctx context.Context) ([]byte, error) {
...
+	pool.AppendCertsFromPEM(ca.CrtPEM)
 
-	// We need a broader set of ciphers to interact with browsers.
-	tlsConfig.CipherSuites = make([]uint16, len(tls.CipherSuites()))
-	for _, s := range tls.CipherSuites() {
-		tlsConfig.CipherSuites = append(tlsConfig.CipherSuites, s.ID)
+	tlsConfig := &tls.Config{
+		RootCAs:      pool,
+		MinVersion:   tls.VersionTLS12,
+		Certificates: []tls.Certificate{*keypair.Certificate},
```
andrewrynhard, line 278: Basically, moves to a simple RSA PKI setup.

internal/pkg/webconfig/webconfig.go
Ulexus: Curious; I used the fmt method because _I_ wasn't seeing the logs with the _logger_.,

Ulexus (approved):

internal/pkg/webconfig/webconfig.go
andrewrynhard: Hmm. Well that is strange. Maybe something broken specifically around the logger and container mode. Regardless, I think this is the intended way, and it should work. Sounds like a bug outside of this.

PR #2719, opened 2020-11-02
refactor: use gRPC for interactive installation

andrewrynhard:
Instead of hosting a web service, we decided to implement a gRPC service
that exposes APIs that can be used in a client-side interactive installer.


Dockerfile
```diff
@@ -579,7 +579,7 @@ RUN --mount=type=cache,target=/.cache/go-build --mount=type=cache,target=/.cache
 WORKDIR /src/pkg/machinery
 RUN --mount=type=cache,target=/.cache/go-build --mount=type=cache,target=/.cache/golangci-lint golangci-lint run --config ../../.golangci.yml
 WORKDIR /src
-RUN --mount=type=cache,target=/.cache/go-build importvet github.com/talos-systems/talos/...
+# RUN --mount=type=cache,target=/.cache/go-build importvet github.com/talos-systems/talos/...
```
andrewrynhard, line 582: I have not clue why, but this is failing with:

```
 > [lint-go 6/8] RUN --mount=type=cache,target=/.cache/go-build importvet github.com/talos-systems/talos/...:
#66 6.531 /src/internal/pkg/etcd/etcd.go:220:13: cannot use m (variable of type *etcdserverpb.Member) as *etcdserverpb.Member value in assignment
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/picker/err.go:37:53: PickOptions not declared by package balancer
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/picker/roundrobin_balanced.go:55:63: PickOptions not declared by package balancer
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/picker/err.go:25:9: cannot use &(errPicker literal) (value of type *errPicker) as Picker value in return statement: wrong type for method Pick
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/picker/roundrobin_balanced.go:33:9: cannot use &(rrBalanced literal) (value of type *rrBalanced) as Picker value in return statement: wrong type for method Pick
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/resolver/endpoint/endpoint.go:114:87: BuildOption not declared by package resolver
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/resolver/endpoint/endpoint.go:182:40: ResolveNowOption not declared by package resolver
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/resolver/endpoint/endpoint.go:41:20: cannot use bldr (variable of type *builder) as resolver.Builder value in argument to resolver.Register: wrong type for method Build
#66 6.531 /go/pkg/mod/github.com/coreos/etcd@v3.3.18+incompatible/clientv3/balancer/resolver/endpoint/endpoint.go:128:9: cannot use r (variable of type *Resolver) as resolver.Resolver value in return statement: wrong type for method ResolveNow
#66 6.531 /src/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go:1541:38: cannot use c.Client (variable of type *clientv3.Client) as *clientv3.Client value in argument to concurrency.NewSession
#66 6.531 importvet: 10 errors during loading
```

cmd/talosctl/cmd/talos/apply-config.go
```diff
@@ -52,7 +52,7 @@ var applyConfigCmd = &cobra.Command{
 }
 
 func init() {
-	applyConfigCmd.Flags().StringVar(&applyConfigCmdFlags.filename, "f", "", "the filename of the updated configuration")
+	applyConfigCmd.Flags().StringVarP(&applyConfigCmdFlags.filename, "file", "f", "", "the filename of the updated configuration")
```
andrewrynhard, line 55: Before this change this command had only `--f` instead of `--file` and `-f`.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -550,24 +552,24 @@ func fetchConfig(ctx context.Context, r runtime.Runtime) (out []byte, err error)
 	return b, nil
 }
 
-func receiveConfigViaWebconfig(ctx context.Context, logger *log.Logger, r runtime.Runtime) ([]byte, error) {
-	cfgBytes, err := webconfig.Run(ctx, logger, r.State().Platform().Mode())
+func receiveConfigViaInteractiveInstaller(ctx context.Context, logger *log.Logger, r runtime.Runtime) ([]byte, error) {
```
andrewrynhard, line 555: Standardized on calling it "interactive installer" ... happy to change, but I don't think it is "webconfig" after this change.

cmd/talosctl/cmd/talos/interactive-installer.go
```diff
@@ -0,0 +1,68 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if interactiveInstallerCmdFlags.filename == "" {
+			return fmt.Errorf("no filename supplied for configuration")
+		}
+
+		ctx := context.Background()
+
+		addr := args[0]
+
+		c, err := client.NewInsecureTokenClient(ctx, addr)
```
Unix4ever, line 37: It can be cool if it's possible to add a new method called  `WithClientInsecure`, or introduce `insecure` flag into `WithClient` method, to get a command for configuration apply.

I think it can be cool if apply config just works both for insecure mode and secure mode.

```
talosctl apply-config --insecure init.yaml
```
Even better approach would be to avoid exposing this  `--insecure` flag, but detect if node runs in insecure mode and select it automatically.

And for interactive mode it can be:
```
talosctl apply-config --insecure --interactive
```

andrewrynhard: Yeah I was thinking about that but then I thought that this subcommand might be slightly different because it could do more than just apply the config. It could, for example, be an interactive exchange between the user that prompts the user for input parameters, and then even generates the config server side and hands them off to the user. If that isn't clear, I can clarify in a meet.

Unix4ever (approved):

PR #2723, opened 2020-11-03
chore: update golangci-lint

andrewrynhard:
Brings in the latest version of golangci-lint and addresses errors.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -122,6 +122,11 @@ linters:
     - gomnd
     - goerr113
     - nestif
+    - gci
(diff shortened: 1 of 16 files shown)


.golangci.yml
```diff
@@ -122,6 +122,11 @@ linters:
     - gomnd
     - goerr113
     - nestif
+    - gci
+    - nlreturn
+    - exhaustivestruct
+    - errorlint
+    - wrapcheck
```
andrewrynhard, line 129: Some of these are good and we should enable later.

PR #2758, opened 2020-11-11
fix: use ghcr.io/talos-systems/kubelet

andrewrynhard:
Moves us off of docker.io.

### DIFF
--- a/website/content/docs/v0.7/Guides/upgrading-kubernetes.md
+++ b/website/content/docs/v0.7/Guides/upgrading-kubernetes.md
@@ -251,7 +251,7 @@ spec:
 ### Kubelet
 
 The Talos team now maintains an image for the `kubelet` that should be used starting with Kubernetes 1.19.
-The image for this release is `docker.io/autonomy/kubelet:v1.19.0`.
+The image for this release is `ghcr.io/talos-systems/kubelet:v1.19.3`.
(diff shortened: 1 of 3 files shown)


Unix4ever (dismissed):
I wonder if we can finally drop `docker.io` from the list of mirrors in the documentation for running qemu with local cache.

smira:
> I wonder if we can finally drop `docker.io` from the list of mirrors in the documentation for running qemu with local cache.

I think `docker.io` is still popular enough for 3rd party stuff, even if Talos core doesn't depend on it.

E.g. Cilium iirc is on `docker.io`.

website/content/docs/v0.7/Guides/upgrading-kubernetes.md
```diff
@@ -251,13 +251,13 @@ spec:
 ### Kubelet
 
 The Talos team now maintains an image for the `kubelet` that should be used starting with Kubernetes 1.19.
-The image for this release is `docker.io/autonomy/kubelet:v1.19.0`.
+The image for this release is `ghcr.io/talos-systems/kubelet:v1.19.0`.
```
smira, line 254: I think we shouldn't change that (or we should update the tag to 1.19.3)

andrewrynhard: will change to `1.19.3` ... goal is to get rid of `docker.io/autonomy` as much as possible in 0.7.

smira (approved):

PR #2760, opened 2020-11-11
feat: add storage API

andrewrynhard:
This is the initial implementation of a storage API.


internal/app/maintenance/server/server.go
```diff
@@ -61,3 +64,27 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 
 	return reply, nil
 }
+
+// Disks implements machine.MaintenanceService.
+func (s *Server) Disks(ctx context.Context, in *empty.Empty) (reply *storage.DisksResponse, err error) {
```
smira, line 69: we need to add a registration line to the `Register` method

andrewrynhard: ah, right :+1:

Unix4ever (dismissed):

smira (approved):

PR #2763, opened 2020-11-11
chore: remove maintenance service interface and use machine service

Unix4ever:
Now maintenance service implements `MachineService` interface, stubbing
all not implemented methods.

Fixes: https://github.com/talos-systems/talos/issues/2764


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -67,6 +69,7 @@ const OSPathSeparator = string(os.PathSeparator)
 
 // Server implements the gRPC service server.
 type Server struct {
+	storaged.Server
```
andrewrynhard, line 72: I think we can leave this out from here for now. I plan on following up with `storaged` service.

Unix4ever: ok, so should I leave machine service without this method for now?

Unix4ever: I would like to share this logic between storage service and maintenance service later on though.

andrewrynhard: yeah, I can make it common to them. It is just that the storage service will be in `storaged` and not in `machined` so we might as well skip in `machined` for now.

andrewrynhard (approved):

andrewrynhard:
Not a blocker for me, but a note for the future. I think this would better be marked as a `refactor`. A chore is designated for things like CI changes. Non-code changes.

Unix4ever:
ah, ok

PR #2773, opened 2020-11-12
feat: add generate config gRPC API

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2766

This API is implemented in Maintenance and Machine services.
Can be used to generate configuration on the node, instead of using
talosctl to generate it locally.

To be used in interactive installer and talosctl gen config.


api/machine/machine.proto
```diff
@@ -28,6 +28,8 @@ service MachineService {
       returns (EtcdLeaveClusterResponse);
   rpc EtcdForfeitLeadership(EtcdForfeitLeadershipRequest)
       returns (EtcdForfeitLeadershipResponse);
+  rpc GenerateConfiguration(GenerateConfigurationRequest)
```
smira, line 31: should it be `ConfigGenerate` like `ServiceList` ? 

or we want it align with `ApplyConfiugration`?

api/machine/machine.proto
```diff
@@ -713,3 +715,23 @@ message EtcdMemberList {
...
+  string kubernetes_version = 5;
+  string install_disk = 6;
+  string install_image = 7;
+  string dns_domain = 8;
+  string hostname = 9;
+}
+// GenerateConfiguration describes the response to a generate configuration request.
+message GenerateConfigurationResponse {
+  common.Metadata metadata = 1;
+  bytes data = 2;
```
smira, line 736: I think it might be nice to support here multiple files to be returned (with filenames). this would help with multi-doc config later on

api/machine/machine.proto
```diff
@@ -713,3 +715,23 @@ message EtcdMemberList {
   repeated string members = 2;
 }
 message EtcdMemberListResponse { repeated EtcdMemberList messages = 1; }
+
+// rpc generateConfiguration
+// GenerateConfigurationRequest describes a request to generate a new configuration
+// on a node.
+message GenerateConfigurationRequest {
```
smira, line 722: I think we need configuration version here as the first argument to support new config formats later

api/machine/machine.proto
```diff
@@ -713,3 +715,23 @@ message EtcdMemberList {
...
+  int32 machine_type = 1;
```
smira, line 723: this should be protobuf Enum

api/machine/machine.proto
```diff
@@ -713,3 +715,23 @@ message EtcdMemberList {
...
+  string version = 2;
```
smira, line 724: my suggestion is to use embedded `message` for each part of the config (cluster, machine, machine network) to make it more readable and easier to extend. we could align that closer with config structure?

internal/app/maintenance/server/server.go
```diff
@@ -66,3 +68,15 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 
 	return reply, nil
 }
+
+// GenerateConfiguration implements the machine.MachineServer interface.
+// nolint:gocyclo
+func (s *Server) GenerateConfiguration(ctx context.Context, in *machine.GenerateConfigurationRequest) (reply *machine.GenerateConfigurationResponse, err error) {
+	machineType := v1alpha1machine.Type(in.MachineType)
+
+	if machineType != v1alpha1machine.TypeInit {
```
smira, line 77: this is probably not true, `TypeJoin` can't be generated, but it's perfectly fine to use `TypeControlPlane` and bootstrap via the API

api/machine/machine.proto
smira: oh `version` is config version, but it's not obvious

api/machine/machine.proto
Unix4ever: yeah, I wanted to align with `ApplyConfiguration`, but I'm fine with any name for it.

api/machine/machine.proto
Unix4ever: and it will actually be nested in that case. Is that fine?

api/machine/machine.proto
Unix4ever: ah, yeah, makes sense.

api/machine/machine.proto
Unix4ever: will rename it to config_version

api/machine/machine.proto
Unix4ever: :ok:

api/machine/machine.proto
Unix4ever: I'm not sure that matching it with actual config structure is a good option... Here we just specify generator input parameters for the generator, while internal config struct may change.
It actually becomes harder to use this gRPC having nested structs, especially now.
But we can group params like you suggested, just avoiding nesting them. What do you think?

api/machine/machine.proto
andrewrynhard: `GenerateConfiguration` reads better IMHO.

smira (approved):

Ulexus (approved):
I'm a little confused as to what role this serves.  Basically, it _seems_ to me that ApplyConfiguration could be modified to handle this case.  That is, we could pass a flag to ApplyConfig which signals it to use a base config and apply values contained in the request rather than using the configuration directly, allowing us to use the same RPC for either operation.

PR #2783, opened 2020-11-13
feat: drop to maintenance mode in cloud platforms if userdata is missing

### LINKED ISSUE #2780: cloud platforms: jump to maintentance mode if userdata is missing
This should only happen if userdata is missing, and as config is persisted, it should only be triggered on initial (first) boot

smira:
On first boot of Talos, if userdata is missing, Talos is going to drop
into maintenance mode which allows to upload config to the server via
`talosctl apply-config` command.

See also: https://github.com/talos-systems/go-retry/pull/4

Fixes #2780

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -21,6 +21,9 @@ const b64 = "base64"
 type downloadOptions struct {
 	Headers map[string]string
 	Format  string
+
+	ErrorOnNotFound      error
+	ErrorOnEmptyResponse error
 }
 
 // Option configures the download options.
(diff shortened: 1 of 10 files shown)


pkg/download/download.go
```diff
@@ -89,12 +106,9 @@ func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, e
...
 		}
 
-		b, err = download(req)
-		if err != nil {
-			return retry.ExpectedError(err)
-		}
+		b, err = download(req, dlOpts)
 
-		return nil
+		return err
```
Unix4ever, line 111: tiny nit: should just `return download(req, dlOpts)` work in this case?

Unix4ever (approved):

pkg/download/download.go
smira: the reason is that `b` will not be set, function in `Retry` should return only `error`

Unix4ever: oh, right. It's nested function. I should have expanded more sources :)

website/content/docs/v0.7/Cloud Platforms/digitalocean.md
```diff
@@ -10,16 +10,17 @@ If you need more information on Digital Ocean specifics, please see the [officia
...
 
 Using an upload method of your choice (`doctl` does not have Spaces support), upload the image to a space.
 Now, create an image using the URL of the uploaded image:
 
 ```bash
 doctl compute image create \
     --region $REGION \
-    --image-name talos-digital-ocean-tutorial \
-    --image-url https://talos-tutorial.$REGION.digitaloceanspaces.com/digital-ocean.raw.gz \
+    --image-description talos-digital-ocean-tutorial \
```
Ulexus, line 22: :+1: I forgot about this; it hit me when I was getting my DO setup running.

Ulexus (approved):

smira:
rebased resolving conflicts, no changes

PR #2793, opened 2020-11-16
feat: return client config as the second value in GenerateConfiguration

Unix4ever:
To be used in interactive installer to output the node client
configuration to a file.


andrewrynhard (approved):
Curious how you see this working with the join process in the future. Do we always generate both?

internal/pkg/configuration/configuration.go
```diff
@@ -86,7 +87,20 @@ func Generate(ctx context.Context, in *machine.GenerateConfigurationRequest) (re
...
+		}
+
+		taloscfgBytes, err = talosconfig.Bytes()
+
+		if err != nil {
+			return nil, err
+		}
+
+		reply = &machine.GenerateConfigurationResponse{
+			Data: [][]byte{cfgBytes, taloscfgBytes},
```
andrewrynhard, line 102: It feels fragile to have the configs in a slice. Ordering becomes important at that point.

smira: probably deserves a field on its own

Unix4ever: :heavy_check_mark:

Unix4ever:
> Curious how you see this working with the join process in the future. Do we always generate both?

We can generate both all the time. But actually currently we don't have the code for join config generation.

smira (approved):

PR #2808, opened 2020-11-17
fix: return hostname from packet platform

rsmitty:
This PR fixes a bug where kubelets weren't getting registered with the
actual hostname of the server in packet.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/packet/packet.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/packet/packet.go
@@ -23,6 +23,7 @@ import (
 
 // Metadata holds packet metadata info.
 type Metadata struct {
+	Hostname       string   `json:"hostname"`
 	Network        Network  `json:"network"`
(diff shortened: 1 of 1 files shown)


smira (dismissed):

internal/app/machined/pkg/runtime/v1alpha1/platform/packet/packet.go
```diff
@@ -188,8 +189,20 @@ func (p *Packet) Mode() runtime.Mode {
...
+
+	metadataConfig, err := download.Download(ctx, PacketMetaDataEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	var unmarshalledMetadataConfig Metadata
+	if err = json.Unmarshal(metadataConfig, &unmarshalledMetadataConfig); err != nil {
+		return nil, err
+	}
```
andrewrynhard, line 203: Looks like we could optimize here a bit. We download the metadata twice. Not a big deal at all, but we have the `Packet` struct that we could store things to.

PR #2810, opened 2020-11-17
fix: ignore 'not found' errors when stopping/removing CRI pods

### LINKED ISSUE #2806: Single node upgrade not working to stable v0.7.0 from alpha.8
## Bug Report
Trying to do an upgrade for single node fails in the upgrade sequence:

### Description

Running upgrade command with -p flag set and v0.7.0 installer image

### Logs

172.26.4.101: [talos] task stopAllPods (1/1): failed
172.26.4.101: [talos] task stopAllPods (1/1): done, 2.863869398s
172.26.4.101: [talos] phase cleanup (2/10): failed
172.26.4.101: [talos] upgrade sequence: failed
172.26.4.101: [talos] upgrade sequence: done: 1m30.735901165s
172.26.4.101: [talos] upgrade failed: error running phase 2 in upgrade sequence: task 1/1: failed, rpc error: code = NotFound desc = failed to get verbose sandbox container info: failed to get task status: task 437228c7ebf26441219b418edcd1655f7f04204d41f75488784f56fce7ce52f0 not found: not found
172.26.4.101: [talos] fatal sequencer error in "upgrade" sequence: message:"sequence failed: error running phase 2 in upgrade sequence: task 1/1: failed, rpc error: code = NotFound desc = failed to get verbose sandbox container info: failed to get task status: task 437228c7ebf26441219b418edcd1655f7f04204d41f75488784f56fce7ce52f0 not found: not found"
172.26.4.101: [talos] rebooting in 10 seconds
172.26.4.101: [talos] rebooting in 9 seconds

### Environment
Server:
        NODE:        172.26.4.101
        Tag:         v0.7.0-alpha.8
        SHA:         50527fe0
        Built:
        Go version:  go1.15.3
        OS/Arch:     linux/amd64

- Kubernetes version: 
1.19.1

- Platform:
qemu

smira:
Fixes #2806

Also skips stopping pods which are already stopped from the previous run
with modes `POD`/`CONTAINER`.

### DIFF
--- a/internal/pkg/cri/pods.go
+++ b/internal/pkg/cri/pods.go
@@ -10,6 +10,8 @@ import (
 	"log"
 
 	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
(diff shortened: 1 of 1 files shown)


andrewrynhard (approved):

internal/pkg/cri/pods.go
```diff
@@ -104,21 +106,35 @@ func (c *Client) StopAndRemovePodSandboxes(ctx context.Context, stopAction StopA
...
-		g.Go(func() error {
-			return stopAndRemove(ctx, stopAction, c, pod, networkMode.String())
+			// If any modes are specified, we verify that the current pod is
+			// running any one of the modes. If it doesn't, we skip it.
+			if len(modes) > 0 && !contains(networkMode, modes) {
+				return nil
+			}
+
+			if pod.GetState() != runtimeapi.PodSandboxState_SANDBOX_READY {
+				log.Printf("skipping pod %s/%s, state %s", pod.Metadata.Namespace, pod.Metadata.Name, pod.GetState())
```
andrewrynhard, line 128: This pod will ultimately get cleaned up? My only concern is that we need to be certain that all pods are removed in case they mount something that prevents our upgrade.

andrewrynhard (approved):

internal/pkg/cri/pods.go
```diff
@@ -104,21 +106,35 @@ func (c *Client) StopAndRemovePodSandboxes(ctx context.Context, stopAction StopA
...
-		}
+		g.Go(func() error {
+			status, _, e := c.PodSandboxStatus(ctx, pod.GetId())
+			if e != nil {
+				if grpcstatus.Code(e) == codes.NotFound {
+					return nil
+				}
 
-		networkMode := status.GetLinux().GetNamespaces().GetOptions().GetNetwork()
+				return e
```
smira, line 116: moved this into the `errgroup`, as otherwise `return err` leaves some goroutines in `errgroup` running while we return

internal/pkg/cri/pods.go
smira: I believe if sandbox is not ready, sandbox is stopped. 

So the flow with single-node upgrade is the following:

* first we stop all the pods/containers which depend on the CNI
* then we try stopping all the pods (including NODE networking), but that also goes once again over the pods we stopped in the previous step, so this check prevents a pod being stopped twice, which with canal results in CNI error, as it tries to reach out to the CNI which is already stopped

As for the mounts, we check for them explicitly before trying an upgrade.

andrewrynhard: Cool. Yeah I know we check, but I want to be sure we can handle them gracefully.

smira: yep, if it's not a single-node case, pods are removed, so by the time we get to this step, pods is not in the list even

smira:
e2e is good: https://ci.dev.talos-systems.io/talos-systems/talos/12946

PR #2814, opened 2020-11-19
feat: add support for installing to SBCs

andrewrynhard:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,7 @@ FROM ghcr.io/talos-systems/xfsprogs:${PKGS} AS pkg-xfsprogs
 FROM ghcr.io/talos-systems/util-linux:${PKGS} AS pkg-util-linux
 FROM ghcr.io/talos-systems/kmod:${PKGS} AS pkg-kmod
 FROM ghcr.io/talos-systems/kernel:${PKGS} AS pkg-kernel
+FROM ghcr.io/talos-systems/u-boot:${PKGS} AS pkg-u-boot
 
 # Resolve package images using ${EXTRAS} to be used later in COPY --from=.
 
@@ -470,6 +471,7 @@ COPY --from=unicode-pf2 /usr/share/grub/unicode.pf2 /usr/share/grub/unicode.pf2
 ARG TARGETARCH
 COPY --from=kernel /vmlinuz-${TARGETARCH} /usr/install/vmlinuz
 COPY --from=initramfs /initramfs-${TARGETARCH}.xz /usr/install/initramfs.xz
+COPY --from=pkg-u-boot / /usr/install/u-boot
 COPY --from=installer-build /installer /bin/installer
 RUN ln -s /bin/installer /bin/talosctl
 ARG TAG
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,8 @@ REGISTRY_AND_USERNAME := $(REGISTRY)/$(USERNAME)
 DOCKER_LOGIN_ENABLED ?= true
 
 ARTIFACTS := _out
-TOOLS ?= ghcr.io/talos-systems/tools:v0.3.0-10-g83dc352
-PKGS ?= v0.3.0-27-g7a64952
+TOOLS ?= ghcr.io/talos-systems/tools:v0.3.0-11-g84c834e
+PKGS ?= v0.3.0-35-g6b7be81
 EXTRAS ?= v0.1.0-5-gcc2df81
 GO_VERSION ?= 1.15
 GOFUMPT_VERSION ?= abc0db2c416aca0f60ea33c23c76665f6e7ba0b6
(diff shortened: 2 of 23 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -167,6 +167,15 @@ func (g *Grub) Install(fallback string, config interface{}, sequence runtime.Seq
 		}
 	}
 
+	log.Println("writing u-boot-sunxi-with-spl.bin to disk")
+
+	b, err := ioutil.ReadFile("/usr/install/u-boot-sunxi-with-spl.bin")
+
+	_, err = dev.Device().WriteAt(b, 1024*8)
```
andrewrynhard, line 174: This doesn't actually seem to be working. I have to manually `dd` this file right now. Not sure why.

Dockerfile
```diff
@@ -29,6 +29,8 @@ FROM ghcr.io/talos-systems/xfsprogs:${PKGS} AS pkg-xfsprogs
 FROM ghcr.io/talos-systems/util-linux:${PKGS} AS pkg-util-linux
 FROM ghcr.io/talos-systems/kmod:${PKGS} AS pkg-kmod
 FROM ghcr.io/talos-systems/kernel:${PKGS} AS pkg-kernel
+FROM docker.io/andrewrynhard/u-boot:v0.3.0-32-g8d8eada AS pkg-u-boot
+# FROM ghcr.io/talos-systems/u-boot:${PKGS} AS pkg-u-boot
```
smira, line 33: do we need to merge the u-boot PR?

Makefile
```diff
@@ -148,12 +148,13 @@ initramfs: ## Builds the compressed initramfs and outputs it to the artifact dir
 
 .PHONY: installer
 installer: ## Builds the container image for the installer and outputs it to the artifact directory.
-ifeq (,$(MULTI_PLATFORM))
-	@$(MAKE) docker-$@ DEST=$(ARTIFACTS) TARGET_ARGS="--allow security.insecure"
-	@docker load < $(ARTIFACTS)/$@.tar
-else
 	@$(MAKE) registry-$@ TARGET_ARGS="--allow security.insecure"
-endif
+# ifeq (,$(MULTI_PLATFORM))
```
smira, line 152: we could drop "old way" via `docker load` if we want to

cmd/installer/cmd/image.go
```diff
@@ -148,6 +148,10 @@ func finalize(platform runtime.Platform, img string) (err error) {
 		if err = ova.CreateOVAFromRAW(name, img, outputArg); err != nil {
 			return err
 		}
+	case "metal":
```
smira, line 151: should we call it generic? ???

cmd/installer/cmd/image.go
```diff
@@ -148,6 +148,10 @@ func finalize(platform runtime.Platform, img string) (err error) {
...
+		if err = tar(fmt.Sprintf("metal-%s.tar.gz", stdruntime.GOARCH), file, dir); err != nil {
```
smira, line 152: we need to add this to `.drone.jsonnet` release artifacts

smira (commented):
lgtm, just a few nits!

Dockerfile
andrewrynhard: We do, but that is failing because it fails for x86

cmd/installer/cmd/image.go
andrewrynhard: it has to be a valid platform, and we don't have a `generic` one

cmd/installer/cmd/image.go
andrewrynhard: Note for myself: we need this to include the board name in the filename if a board name is supplied.

Makefile
andrewrynhard: Not sure I understand, but to explain why I commented this out ... it seems targeting a specific platform for the installer is either broken or I am not doing it right.

website/content/docs/v0.8/Single Board Computers/libretech_all_h3_cc_h5.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+## Generating the Image
+
+```bash
+docker run \
+  --rm \
+  -v /dev:/dev \
+  --privileged \
+  ghcr.io/talos-systems/installer:latest image --platform metal --board libretech_all_h3_cc_h5 --tar-to-stdout | tar xz
```
andrewrynhard, line 12: I wonder if there is a better way ... having to uncompress and untar are extra steps we could eliminate?

Unix4ever: I think I would like the idea of having a shell script somewhere in the repo to just:

```bash
curl -sLf https://raw.githubusercontent.com/.../gen-image.sh | bash -s -- -o /dev/mmcblk0
```
And the script can do both steps inside.
It doesn't really solve tar untar problem, but at least can simplify the flow.

Unix4ever (approved):

Makefile
andrewrynhard: Whoops I have to revert this change.

website/content/docs/v0.8/Single Board Computers/libretech_all_h3_cc_h5.md
andrewrynhard: Good idea. I feel like it could be better to host on Talos.dev (nice short url). Something we can probably figure out later? For now I was more concerned about not having to tar/untar.

Unix4ever: Regarding tar/untar: nothing comes to mind except mounting yet another volume and dumping that data without compression there.

andrewrynhard: I think what we should do is think this through a bit and follow up. I am over it now :smile: .

smira (dismissed):

smira (approved):

Unix4ever (approved):

PR #2820, opened 2020-11-19
feat: make GenerateConfiguration API reuse current node auth

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2819

Only if requested config type is not `TypeInit`.
This functionality will help implementing TUI installer cluster
extension workflow.

### DIFF
--- a/internal/pkg/configuration/configuration.go
+++ b/internal/pkg/configuration/configuration.go
@@ -8,19 +8,22 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 
 	"github.com/talos-systems/talos/pkg/machinery/api/machine"
 	"github.com/talos-systems/talos/pkg/machinery/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/configloader"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/generate"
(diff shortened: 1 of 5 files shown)


internal/pkg/configuration/configuration.go
```diff
@@ -62,6 +65,43 @@ func Generate(ctx context.Context, in *machine.GenerateConfigurationRequest) (re
 			taloscfgBytes []byte
 		)
 
+		if machineType != v1alpha1machine.TypeInit {
```
smira, line 68: I think we should phase out notion of `Init` node altogether

if we are running `machined` (vs. maintenance service), any config generated should use existing keys, not generate new ones

Unix4ever: ok, makes sense

andrewrynhard (commented):
I think I need to be caught up on this PR. Ping me in the AM.

pkg/machinery/config/types/v1alpha1/generate/options.go
```diff
@@ -169,6 +196,9 @@ type GenOptions struct {
 	Debug                     bool
 	Persist                   bool
 	MachineDisks              []*v1alpha1.MachineDisk
+	TrustdInfo                *TrustdInfo
```
smira, line 199: this is totally a nit, but I would probably have this in `Input`, as this doesn't seem to be really an option vs. something we supply as input (secrets), with defaults being to generate new secrets

smira (approved):

PR #2830, opened 2020-11-23
feat: implement apply configuration without reboot

### LINKED ISSUE #2828: implement ApplyConfiguration without reboot
Config is saved, but not applied yet. Config is applied after reboot

smira:
This allows config to be written to disk without being applied
immediately.

Small refactoring to extract common code paths.

At first, I tried to implement this via the sequencer, but looks like
it's too hard to get it right, as sequencer lacks context and config to
be written is not applied to the runtime.

Fixes #2828


Unix4ever (dismissed):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
```diff
@@ -35,22 +38,62 @@ func (r *Runtime) Config() config.Provider {
...
 	}
 
 	r.c = cfg
 
 	return nil
 }
 
+// PopulateConfig implements the Runtime interface.
+func (r *Runtime) PopulateConfig(ctx context.Context, cfg config.Provider) error {
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
smira, line 69: dynamic config provider

andrewrynhard (approved):

PR #2832, opened 2020-11-23
feat: support cluster expansion in the interactive installer

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2816

Now with:
```
talosctl apply-config -i --interactive -n 10.5.0.2
talosctl apply-config -i --interactive -n 10.5.0.3 -e 10.5.0.2
```

It is possible to add more nodes to already existing cluster.
Installer will call `GenerateConfiguration` on the bootstrap node
instead of using target node.

As we have to create separate clients to connect to two different nodes (one can be insecure and another one should be secure), I've decided to create that 2 clients wrapper `Connection` entity.
I didn't want to drag node context and node client through all the places, this looks cleaner, but if you don't like my approach I'm open for suggestions.

### DIFF
--- a/cmd/talosctl/cmd/talos/apply-config.go
+++ b/cmd/talosctl/cmd/talos/apply-config.go
@@ -99,13 +99,28 @@ var applyConfigCmd = &cobra.Command{
 
 		return withClient(func(ctx context.Context, c *client.Client) error {
 			if applyConfigCmdFlags.interactive {
-				installer := installer.NewInstaller()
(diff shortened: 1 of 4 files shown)


andrewrynhard (approved):
:heart:

Unix4ever:
Found the issue with wrong context being used in Connection in cluster expansion flow.
Fixed that and made one last small refactoring in apply-config.

andrewrynhard (approved):

PR #2835, opened 2020-11-24
feat: change UI component for disks selector

Unix4ever:
Implemented a new component based on `tview.Table` that shows all
available block devices as a table.

Had to stop using standard form control, because it doesn't really
handle multiline elements.

### DIFF
--- a/internal/pkg/tui/installer/state.go
+++ b/internal/pkg/tui/installer/state.go
@@ -46,7 +46,9 @@ func NewState(ctx context.Context, conn *Connection) (*State, error) {
 		opts.ClusterConfig.ControlPlane.Endpoint = fmt.Sprintf("https://%s:%d", conn.nodeEndpoint, constants.DefaultControlPlanePort)
 	}
 
-	diskInstallOptions := []interface{}{}
+	diskInstallOptions := []interface{}{
+		NewTableHeaders("device name", "model name", "size"),
+	}
 
 	disks, err := conn.Disks()
 	if err != nil {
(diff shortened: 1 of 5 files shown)


Unix4ever:
New look:
https://asciinema.org/a/ccHNKJt13g1sE8TuHXnRynVdZ

smira:
looks cool, are we missing device size here (looking at asciinema?)

Unix4ever:
It's there:
image

andrewrynhard:
It might be nice to know which one is selected. I see the highlight but nothing to indicate that I actually selected it.

Unix4ever:
ah, that makes sense. I have an idea actually

Unix4ever:
What do you think about this?
<enter> will switches the selected item.
https://asciinema.org/a/ot8Sz98Oqsm3v9pm1n1dARm4P

internal/pkg/tui/installer/state.go
```diff
@@ -80,38 +81,40 @@ func NewState(ctx context.Context, conn *Connection) (*State, error) {
...
+					"control plane endpoint",
+					v1alpha1.ControlPlaneConfigDoc.Describe("endpoint", true),
+					&opts.ClusterConfig.ControlPlane.Endpoint,
 				),
 				NewItem(
-					"image",
-					v1alpha1.InstallConfigDoc.Describe("image", true),
-					&opts.MachineConfig.InstallConfig.InstallImage,
+					"kubernetes version",
+					"Kubernetes version to install.",
```
andrewrynhard, line 116: Are we missing the auto doc here?

andrewrynhard (dismissed):

internal/pkg/tui/installer/state.go
Unix4ever: This kubernetes version actually does not have one. We have doc string for the image, but not for the image version.

smira: I believe we don't have that as a single config option? we have multiple which control the k8s version

Unix4ever: yeah, this kubernetes version is passed into generate package, then it's making some magic inside. We don't have 1:1 mapping for it in the config.

smira (approved):
nit: I'm not sure if it's `refactor`, probably more of a `feat` as it's a user-visible change

andrewrynhard (approved):

PR #2836, opened 2020-11-25
feat: support openstack platform

rsmitty:
This PR adds the ability for us to deploy Talos in openstack. Tested in
local devstack with a supplied userdata file. It also adds support to
the Makefile for building the openstack image so it'll be published with
next release.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -495,6 +495,8 @@ local release = {
       '_out/gcp-arm64.tar.gz',
       '_out/initramfs-amd64.xz',
       '_out/initramfs-arm64.xz',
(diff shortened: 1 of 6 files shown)


rsmitty:
Oh yeah, for sure. But it's experimental right now, and I want to validate some userdata file size stuff first. I'll make a docs ticket for it to put in backlog though too.

smira (dismissed):

smira:
oh, we also need to add new artifacts to the `.drone.jsonnet` release step

rsmitty:
Argh, good call. Thx for the catch. Updated!

smira (approved):

smira (approved):

PR #2838, opened 2020-11-25
feat: add ability to choose CNI config

Unix4ever:
Initial version which only allows setting CNI using preset, no custom
CNI urls are supported at the moment. Still need to figure out what kind
of UI can be used for that.


Unix4ever:
How it looks like:
image

api/machine/machine.proto
```diff
@@ -747,8 +747,14 @@ message ControlPlaneConfig {
   string endpoint = 1;
 }
 
+message CNIConfig {
+  string name = 1;
+  repeated string urls = 2;
+}
```
andrewrynhard, line 753: I wonder if one day we can generate protobufs from our go structs so that the two are always in sync.

andrewrynhard (approved):
This should probably be shown only for `init` nodes.

api/machine/machine.proto
Unix4ever: yeah, that can be nice... Do we have any existing means to generate protobufs from go structs?

smira (approved):

PR #2844, opened 2020-11-27
feat: implement network interfaces list API

Unix4ever:
To be used in the interactive installer to configure networking.

### DIFF
--- a/internal/app/networkd/pkg/networkd/devices.go
+++ b/internal/app/networkd/pkg/networkd/devices.go
@@ -1,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package networkd
+
+import (
+	"net"
+
+	networkapi "github.com/talos-systems/talos/pkg/machinery/api/network"
+)
+
+// GetDevices gathers information about existing network interfaces and their flags.
+func GetDevices() (reply *networkapi.InterfacesResponse, err error) {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return reply, err
+	}
+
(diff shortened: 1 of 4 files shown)


api/machine/machine.proto
```diff
@@ -776,3 +778,18 @@ message GenerateConfigurationResponse {
...
 }
+
+// NetworkInterface describes a single network interface information.
+message NetworkInterface {
+  repeated string addrs = 1;
+  string name = 2;
+  string mac = 3;
+  int32 mtu = 4;
+  bool up = 5;
+}
```
andrewrynhard, line 789: We already have an interface definition in `network.proto`. IMHO we should either a) import that here and add any fields we need to it or 2) have the maintenance service implement the network server (just embed the default unimplemented server) and the implement the interfaces RPC. Given that the pattern we have used so far is to embed the unimplemented server, I would lean towards that.

Unix4ever: oh... It's completely the same there. Yeah that makes perfect sense.

internal/app/networkd/pkg/networkd/devices.go
```diff
@@ -0,0 +1,50 @@
...
+		addrs := make([]string, 0, len(ifaceaddrs))
+		for _, addr := range ifaceaddrs {
+			addrs = append(addrs, addr.String())
+		}
+
+		ifmsg := &networkapi.Interface{
+			Index:        uint32(iface.Index),
+			Mtu:          uint32(iface.MTU),
+			Name:         iface.Name,
+			Hardwareaddr: iface.HardwareAddr.String(),
```
andrewrynhard, line 37: Is this mac address?

andrewrynhard (approved):

internal/app/networkd/pkg/networkd/devices.go
Unix4ever: yep

PR #2859, opened 2020-12-01
chore: build SBC images

andrewrynhard:
This will build SBC images and output them to the artifacts directory. These images
will be published on releases.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -217,8 +217,9 @@ local lint = Step("lint", depends_on=[check_dirty]);
 local talosctl_cni_bundle = Step('talosctl-cni-bundle', depends_on=[lint]);
 local images_amd64 = Step("images-amd64", target="images", depends_on=[installer], environment={"REGISTRY": local_registry});
 local images_arm64 = Step("images-arm64", target="images", depends_on=[installer], environment={"REGISTRY": local_registry, "DOCKER_HOST": "tcp://docker-arm64.ci.svc:2376"});
-local iso_amd64 = Step("iso-amd64", target="iso", depends_on=[images_amd64, images_arm64], environment={"REGISTRY": local_registry});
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -164,6 +164,11 @@ image-%: ## Builds the specified image. Valid options are aws, azure, digital-oc
 
 images: image-aws image-azure image-digital-ocean image-gcp image-metal image-openstack image-vmware ## Builds all known images (AWS, Azure, Digital Ocean, GCP, Metal, Openstack, and VMware).
 
+sbc-%: ## Builds the specified SBC image. Valid options are rpi_4, bananapi_m64, and libretech_all_h3_cc_h5 (e.g. sbc-rpi_4)
```
andrewrynhard, line 167: We need to either pass `console` kernel args here or allow boards to add extra kernel args at install time (like platforms do)

.drone.jsonnet
```diff
@@ -217,6 +217,7 @@ local lint = Step("lint", depends_on=[check_dirty]);
 local talosctl_cni_bundle = Step('talosctl-cni-bundle', depends_on=[lint]);
 local images_amd64 = Step("images-amd64", target="images", depends_on=[installer], environment={"REGISTRY": local_registry});
 local images_arm64 = Step("images-arm64", target="images", depends_on=[installer], environment={"REGISTRY": local_registry, "DOCKER_HOST": "tcp://docker-arm64.ci.svc:2376"});
+local sbcs_arm64 = Step("sbcs-arm64", target="sbcs", depends_on=[installer], environment={"REGISTRY": local_registry, "DOCKER_HOST": "tcp://docker-arm64.ci.svc:2376"});
```
smira, line 220: should we try to keep the fan-out smaller when possible (just to save drone-kuberunner from blowing up)?

we could potentially make it depend on the ISOs?

Makefile
smira: I vote for the platform-like interface, as this should allow us to install from PXE as well? if we put `board:` to the install section?

smira (approved):

smira (approved):

rsmitty (approved):

PR #2864, opened 2020-12-01
feat: allow boards to set kernel args

andrewrynhard:
This allows boards to provide kernel args at install time. We need this so that
we can set the console.


smira (approved):
do we need to remove those console lines from the READMEs?

smira (approved):

PR #2866, opened 2020-12-01
fix: prevent endless loop with DHCP requests in networkd

smira:
There were two problems:

* `configureInterfaces` was always failing if interface is already set
up, as the routes already exist

* `renew` was halving the renew interval each time `configureInterface`
fails, which starts at (LeaseTime/2) and goes effectively to zero

This was leading to high networkd CPU usage, storm of DHCP requests on
the network.

### DIFF
--- a/internal/app/networkd/pkg/address/dhcp.go
+++ b/internal/app/networkd/pkg/address/dhcp.go
@@ -46,9 +46,11 @@ func (d *DHCP) Link() *net.Interface {
 // Discover handles the DHCP client exchange stores the DHCP Ack.
 func (d *DHCP) Discover(ctx context.Context, link *net.Interface) error {
 	d.NetIf = link
-	// TODO do something with context
-	ack, err := d.discover()
-	d.Ack = ack
+	ack, err := d.discover(ctx)
+
+	if ack != nil {
+		d.Ack = ack
+	}
 
 	return err
 }
@@ -191,7 +193,7 @@ func (d *DHCP) Hostname() (hostname string) {
 }
 
 // discover handles the actual DHCP conversation.
-func (d *DHCP) discover() (*dhcpv4.DHCPv4, error) {
+func (d *DHCP) discover(ctx context.Context) (*dhcpv4.DHCPv4, error) {
 	opts := []dhcpv4.OptionCode{
 		dhcpv4.OptionClasslessStaticRoute,
 		dhcpv4.OptionDomainNameServer,
@@ -223,7 +225,7 @@ func (d *DHCP) discover() (*dhcpv4.DHCPv4, error) {
 	// nolint: errcheck
 	defer cli.Close()
 
-	lease, err := cli.Request(context.Background(), mods...)
+	lease, err := cli.Request(ctx, mods...)
 	if err != nil {
 		// TODO: Make this a well defined error so we can make it not fatal
 		log.Printf("failed dhcp request for %q: %v", d.NetIf.Name, err)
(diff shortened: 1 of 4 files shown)


internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -46,9 +46,11 @@ func (d *DHCP) Link() *net.Interface {
 // Discover handles the DHCP client exchange stores the DHCP Ack.
 func (d *DHCP) Discover(ctx context.Context, link *net.Interface) error {
 	d.NetIf = link
-	// TODO do something with context
-	ack, err := d.discover()
-	d.Ack = ack
+	ack, err := d.discover(ctx)
+
+	if ack != nil {
+		d.Ack = ack
```
smira, line 52: don't make `d.Ack = nil`, as this resets `TTL()` to zero

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -312,18 +312,26 @@ func (n *NetworkInterface) Renew() {
...
 	renewDuration := method.TTL() / 2
 
 	var err error
 
 	for {
-		<-time.After(renewDuration)
+		time.Sleep(renewDuration)
 
 		if err = n.configureInterface(method, n.Link); err != nil {
+			log.Printf("faiure to renew address for %q: %s", n.Name, err)
```
rsmitty, line 325: ultra nit:

```suggestion
			log.Printf("failure to renew address for %q: %s", n.Name, err)
```

rsmitty (approved):

internal/app/networkd/pkg/address/dhcp.go
```diff
@@ -223,13 +225,29 @@ func (d *DHCP) discover() (*dhcpv4.DHCPv4, error) {
...
-	lease, err := cli.Request(context.Background(), mods...)
+	lease, err := cli.Request(ctx, mods...)
 	if err != nil {
 		// TODO: Make this a well defined error so we can make it not fatal
 		log.Printf("failed dhcp request for %q: %v", d.NetIf.Name, err)
 
 		return nil, err
 	}
 
+	log.Printf("DHCP ACK on %q: %s", d.NetIf.Name, collapseSummary(lease.ACK.Summary()))
```
smira, line 236: should be single line on console, printed twice on initial discovery, then it goes to networkd log on each discovery

internal/app/networkd/pkg/nic/netlink.go
```diff
@@ -48,6 +48,10 @@ func (n *NetworkInterface) setMTU(idx int, mtu uint32) error {
 		return err
 	}
 
+	if msg.Attributes != nil && msg.Attributes.MTU == mtu {
```
smira, line 51: prevents extra rnetlink message if MTU is the same

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -312,18 +312,26 @@ func (n *NetworkInterface) Renew() {
 // address TTL. If that fails, we'll continue to attempt to retry every
 // halflife.
 func (n *NetworkInterface) renew(method address.Addressing) {
+	const minRenewDuration = 5 * time.Second // protect from renewing too often
```
smira, line 315: this prevents code from entering tight loop on sending endless stream of DHCP requests

internal/app/networkd/pkg/nic/nic.go
smira, line 325: print failures on renew, don't hide them!

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -401,7 +409,10 @@ func (n *NetworkInterface) configureInterface(method address.Addressing, link *n
 
 		err = n.rtnlConn.RouteAdd(method.Link(), *r.Destination, gw, rtnl.WithRouteSrc(src), rtnl.WithRouteAttrs(attr))
 		if err != nil {
-			return err
+			// ignore "EEXIST" errors for routes which are already present
```
smira, line 412: this is the "core" of the fix as if routes don't change, trying to add them always returns `file exists` error

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -312,18 +312,26 @@ func (n *NetworkInterface) Renew() {
...
 func (n *NetworkInterface) renew(method address.Addressing) {
+	const minRenewDuration = 5 * time.Second // protect from renewing too often
+
 	renewDuration := method.TTL() / 2
 
 	var err error
 
 	for {
-		<-time.After(renewDuration)
+		time.Sleep(renewDuration)
```
smira, line 322: `<-time.After()` involves a channel which should be GC'ed, much better to use `time.Sleep()` here

PR #2868, opened 2020-12-01
fix: prevent endless loop with DHCP requests in networkd

smira:
There were two problems:

* `configureInterfaces` was always failing if interface is already set
up, as the routes already exist

* `renew` was halving the renew interval each time `configureInterface`
fails, which starts at (LeaseTime/2) and goes effectively to zero

This was leading to high networkd CPU usage, storm of DHCP requests on
the network.

(cherry picked from commit 360d8879679dbeeea38fb2326e63b12b22ff6ccd)

### DIFF
--- a/internal/app/networkd/pkg/nic/nic.go
+++ b/internal/app/networkd/pkg/nic/nic.go
@@ -306,17 +306,25 @@ func (n *NetworkInterface) Renew() {
 // address TTL. If that fails, we'll continue to attempt to retry every
 // halflife.
 func (n *NetworkInterface) renew(method address.Addressing) {
+	const minRenewDuration = 5 * time.Second // protect from renewing too often
+
 	renewDuration := method.TTL() / 2
 
 	var err error
 
 	for {
-		<-time.After(renewDuration)
+		time.Sleep(renewDuration)
 
 		if err = n.configureInterface(method, n.Link); err != nil {
+			log.Printf("failure to renew address for %q: %s", n.Name, err)
+
 			renewDuration = (renewDuration / 2)
 		} else {
 			renewDuration = method.TTL() / 2
+		}
+
+		if renewDuration < minRenewDuration {
+			renewDuration = minRenewDuration
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


rsmitty (approved):

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -306,18 +306,26 @@ func (n *NetworkInterface) Renew() {
...
 	var err error
 
 	for {
-		<-time.After(renewDuration)
+		time.Sleep(renewDuration)
 
 		if err = n.configureInterface(method, n.Link); err != nil {
+			log.Printf("failure to renew address for %q: %s", n.Name, err)
+
 			renewDuration = (renewDuration / 2)
```
Ulexus, line 321: Why do we even _have_ this ever-shrinking renew duration?  Why should this be any different from minRenewDuration at all?

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -404,7 +412,10 @@ func (n *NetworkInterface) configureInterface(method address.Addressing, link *n
 
 		err = n.rtnlConn.RouteAdd(method.Link(), *r.Dest, gw, rtnl.WithRouteSrc(src), rtnl.WithRouteAttrs(attr))
 		if err != nil {
-			return err
+			// ignore "EEXIST" errors for routes which are already present
+			if opErr, ok := err.(*netlink.OpError); !ok || !os.IsExist(opErr.Err) {
```
Ulexus, line 416: nit: maybe split these two cases (!ok and !os.IsExist) to make them easier to human parse... moreover, I think I would have differing error messaging for each.

internal/app/networkd/pkg/nic/nic.go
smira: I don't have any good idea, this looks like weird retry mechanism. I think the idea is that `1/2 + 1/4 + 1/8 + ... < 1`

internal/app/networkd/pkg/nic/nic.go
smira: we only ignore one specific error here, and any other error is going to be returned to the caller

internal/app/networkd/pkg/nic/nic.go
Ulexus: exactly; this progression doesn't make sense to me

PR #2883, opened 2020-12-02
docs: ensure port for vbox and proxmox docs

rsmitty:
This PR adds the default port to these docs so it's clear that port is
required.


andrewrynhard (approved):

rsmitty:
@masoudbahar that's good feedback. thx. I'll add that to backlog as well.

PR #2885, opened 2020-12-02
feat: introduce configpatcher package in machinery

rsmitty:
This PR moves the configpatcher as a package under machinery. It also
reworks the existing function to specify that it's explicitly for JSON
6902 patching so we can add more patch types if desired later on.


pkg/machinery/config/configpatcher/configpatcher.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+	ghodssyaml "github.com/ghodss/yaml"
+)
+
+// JSON6902 is responsible for applying a JSON 6902 patch to the bootstrap data.
+func JSON6902(decodedData []byte, patch jsonpatch.Patch) ([]byte, error) {
```
andrewrynhard, line 16: Seems like we could do an easy test on this. Preferably table style (`Go: Generate Unit Tests For Function`) so that we can get a lot of variation on input data.

andrewrynhard (approved):
I don't want to block necessarily, but I had one request.

pkg/machinery/config/configpatcher/configpatcher.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	jsonDecodedData, err = patch.Apply(jsonDecodedData)
+	if err != nil {
+		return nil, fmt.Errorf("failure applying rfc6902 patches to machine config: %s", err)
+	}
+
+	decodedData, err = ghodssyaml.JSONToYAML(jsonDecodedData)
+	if err != nil {
+		return nil, fmt.Errorf("failure converting bootstrap data from json to yaml: %s", err)
```
smira, line 29: Should it be Talos config, not bootstrap data?

andrewrynhard (approved):

PR #2887, opened 2020-12-03
feat: make GenerateConfiguration accept current time as a parameter

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2889

If the node time is out of sync, it can generate incorrect
configuration. And maintenance mode does not allow us starting ntp,
because there is no containerd.

By providing current UTC time of the machine where talosctl client is
running, it is possible to force GenerateConfiguration use correct time.


pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -254,45 +256,45 @@ func NewSecretsBundleFromConfig(c config.Provider) *SecretsBundle {
...
 }
 
 // NewTalosCA generates a CA for the Talos PKI.
-func NewTalosCA() (ca *x509.CertificateAuthority, err error) {
+func NewTalosCA(currentTime time.Time) (ca *x509.CertificateAuthority, err error) {
 	opts := []x509.Option{
 		x509.RSA(false),
 		x509.Organization("talos"),
-		x509.NotAfter(time.Now().Add(87600 * time.Hour)),
+		x509.NotAfter(currentTime.Add(87600 * time.Hour)),
```
smira, line 285: we should probably here and in other places explicitly set `x509.NotBefore(currentTime)`

smira: (or whatever the default is in x509 package)

Unix4ever: fixed

pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -224,7 +255,7 @@ func NewSecretsBundle() (*SecretsBundle, error) {
 }
 
 // NewSecretsBundleFromConfig creates secrets bundle using existing config.
-func NewSecretsBundleFromConfig(c config.Provider) *SecretsBundle {
+func NewSecretsBundleFromConfig(clock Clock, c config.Provider) *SecretsBundle {
```
smira, line 258: I guess it doesn't need the clock actually here, as it's not going to be used? (or it still generates some certs this way?)

smira (approved):

pkg/machinery/config/types/v1alpha1/generate/generate.go
Unix4ever: clock is used for admin certs creation.

smira (approved):

andrewrynhard (approved):

PR #2905, opened 2020-12-05
docs: Add docs for network connectivity

### LINKED ISSUE #2870: [Docs] Network port usage by talos
### Description
It's not well documented anywhere what ports are used by the various components of Talos. From looking through the code, I see that there are at least the following:
 - `apid` listening on TCP/50000
 - `trustd` listening on TCP/50001

A page documenting these ports and their usage would be nice, especially in the context of building up firewall rules. Some questions that I currently have:
1. What ports does `talosctl` need access to? `apid` seems obvious, but I assume access to `trustd` is required as well?
2. What ports are used for communication between the control plane nodes and worker nodes? (aside from the standard k8s stuff, of course)
3. Anything else that would be relevant?

If these questions can be answered here, I don't mind writing up the documentation myself in a PR.

e-nomem:
## What? (description)
Add documentation on network ports required to operate Talos.
Closes #2870 

I tried to run `make docs` on my mac before pushing this commit but it failed with the following:
```sh
 > [tools  5/10] RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.32.2:
#12 0.506 golangci/golangci-lint info checking GitHub for tag 'v1.32.2'
#12 0.533 golangci/golangci-lint crit unable to find 'v1.32.2' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

The actual release seems to exist, so I'm not sure what the problem is.

### DIFF
--- a/website/content/docs/v0.8/Guides/configuring-network-connectivity.md
+++ b/website/content/docs/v0.8/Guides/configuring-network-connectivity.md
@@ -1,0 +1,71 @@
+---
+title: "Configuring Network Connectivity"
+description: ""
+---
+
+## Configuring Network Connectivity
+
+The simplest way to deploy Talos is by ensuring that all the remote components of the system (`talosctl`, the control plane nodes, and worker nodes) all have layer 2 connectivity.
+This is not always possible, however, so this page lays out the minimal network access that is required to configure and operate a talos cluster.
+
+ > Note: These are the ports required for Talos specifically, and should be configured _in addition_ to the ports required by kuberenetes.
+ See the [kubernetes docs](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#check-required-ports) for information on the ports used by kubernetes itself.
+
+### Control plane node(s)
+
+<table class="table-auto">
+  <thead>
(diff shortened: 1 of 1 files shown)


andrewrynhard:
@e-nomem Thank you. Was just digging into this with someone the other day. The error you are seeing is related to something (new?) in the docker for mac vm that makes the `golangci-lint` installer script use ipv6. We think we have a workaround.

andrewrynhard:
@e-nomem Thank you. Was just digging into this with someone the other day. The error you are seeing is related to something (new?) in the docker for mac vm that makes the `golangci-lint` installer script use ipv6. We think we have a workaround.

andrewrynhard:
@e-nomem Pro tip: you can skip the `golangci-lint` and focus on markdown with `make lint-markdown`.

e-nomem:
Yep, just dug through the Makefile for that. All the markdown lint issues should be fixed now.

andrewrynhard (approved):
LGTM! Thanks for taking the time to write this up :+1: .

PR #2907, opened 2020-12-05
chore: remove Drone config

andrewrynhard:
This removes the Drone config as we will no longer use Drone.


andrewrynhard:
PR is "ready to go" but with the way Tekton works, we need to zip up some things outside of this PR before we can merge.

PR #2912, opened 2020-12-07
feat: allow disabling NoSchedule taint on masters using TUI installer

Unix4ever:
I think this should come handy for setting up single node SBC clusters.


pkg/machinery/config/types/v1alpha1/generate/join.go
```diff
@@ -59,6 +59,7 @@ func workerUd(in *Input) (*v1alpha1.Config, error) {
 			ServiceSubnet: in.ServiceNet,
 			CNI:           in.CNIConfig,
 		},
+		AllowSchedulingOnMasters: in.AllowSchedulingOnMasters,
```
smira, line 62: we can remove it here, as this config is not for control plane node

smira (approved):

smira (approved):

PR #2914, opened 2020-12-07
feat: install grub for multiple arches into installer image

smira:
This allows installer running say on x86 to install arm64 grub.


smira:
PoC

PR #2921, opened 2020-12-07
test: bump Talos release version for upgrade test to 0.7.1

smira:
We should always use latest releases.


smira:
> /promote integration

test caught a flake on etcd failed to join, but it seems otherwise good in relation to the version bump

PR #2926, opened 2020-12-07
docs: add architectural doc on the root file system layout

andrewrynhard:
This adds documentation on how the root file system is laid out.


smira (approved):
just as an idea, probably we could do some diagram to explain the concept, but anyways :+1:

PR #2927, opened 2020-12-08
test: add ISO test

andrewrynhard:
Adds a simple test for the ISO. Boots the ISO, and then uses the `apply-config` command
in `talosctl` to create a cluster.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -224,6 +224,7 @@ local unit_tests = Step("unit-tests", depends_on=[initramfs]);
 local unit_tests_race = Step("unit-tests-race", depends_on=[initramfs]);
 local e2e_docker = Step("e2e-docker-short", depends_on=[talos, talosctl_linux, unit_tests, unit_tests_race], target="e2e-docker", environment={"SHORT_INTEGRATION_TEST": "yes", "REGISTRY": local_registry});
 local e2e_qemu = Step("e2e-qemu-short", privileged=true, target="e2e-qemu", depends_on=[talosctl_linux, initramfs, kernel, installer, unit_tests, unit_tests_race, talosctl_cni_bundle], environment={"REGISTRY": local_registry, "SHORT_INTEGRATION_TEST": "yes"}, when={event: ['pull_request']});
+local e2e_iso = Step("e2e-iso", privileged=true, target="e2e-iso", depends_on=[talosctl_linux, initramfs, iso_amd64, unit_tests, unit_tests_race], when={event: ['pull_request']}, environment={"REGISTRY": local_registry});
 
 local coverage = {
   name: 'coverage',
@@ -306,8 +307,9 @@ local default_steps = [
   unit_tests,
   unit_tests_race,
   coverage,
-  e2e_docker,
+  e2e_iso,
   e2e_qemu,
+  e2e_docker,
   push,
   push_latest,
 ];
--- a/hack/test/e2e-iso.sh
+++ b/hack/test/e2e-iso.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+set -eoux pipefail
+
+source ./hack/test/e2e.sh
+
+PROVISIONER=qemu
+CLUSTER_NAME=e2e-iso
+
+NODE="172.20.2.2"
+
+function create_cluster {
+  build_registry_mirrors
+
+  "${TALOSCTL}" cluster create \
+    --iso-path=${ARTIFACTS}/talos-amd64.iso \
+    --skip-injecting-config \
+    --wait=false \
+    --with-init-node \
+    --provisioner "${PROVISIONER}" \
+    --name "${CLUSTER_NAME}" \
+    --masters=1 \
+    --workers=0 \
+    --mtu 1500 \
+    --memory 2048 \
+    --cpus 2.0 \
+    --cidr 172.20.2.0/24 \
+    --install-image ${REGISTRY:-ghcr.io}/talos-systems/installer:${TAG} \
+    --cni-bundle-url ${ARTIFACTS}/talosctl-cni-bundle-'${ARCH}'.tar.gz \
+    ${REGISTRY_MIRROR_FLAGS}
+
+  "${TALOSCTL}" config node "${NODE}"
+}
+
+function destroy_cluster() {
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}" --provisioner "${PROVISIONER}"
+}
+
+create_cluster
+timeout -v --preserve-status 1m bash -c "until ${TALOSCTL} apply-config -n ${NODE} --insecure -f controlplane.yaml; do sleep 5; done"
+sleep 5
+timeout -v --preserve-status 2m  bash -c "until nc -z ${NODE} 50000; do sleep 5; done"
+${TALOSCTL} bootstrap -n "${NODE}"
+${TALOSCTL} health --wait-timeout=10m0s -n "${NODE}" --control-plane-nodes="${NODE}"
+destroy_cluster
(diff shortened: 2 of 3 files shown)


hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,17 @@
...
+qemu-img create -f qcow2 disk.qcow2 4G
+
+qemu-system-x86_64 \
+  -m 2048 \
+  -nographic \
+  -serial mon:stdio \
+  -enable-kvm \
+  -cpu host \
+  -boot cd \
+  -netdev user,id=talos,ipv4=on,net=10.254.0.0/24,dhcpstart=10.254.0.10,hostfwd=tcp::4443-:443,hostfwd=tcp::50000-:50000,hostfwd=tcp::6443-:6443,hostname=control-plane-1 \
```
andrewrynhard, line 14: A little worried about this one since it has the potential to clash with another run.

smira: we're inside a k8s pod sandbox anyways?

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,17 @@
...
+  -device virtio-net,netdev=talos \
+  -cdrom _out/talos-amd64.iso \
+  -hda disk.qcow2
```
smira, line 17: do you plan to add `apply-config` here?

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,38 @@
...
+trap 'cleanup' EXIT
+
+${TALOSCTL} gen config \
+  --additional-sans 127.0.0.1 \
+  --install-disk /dev/sda \
+  --install-image ghcr.io/talos-systems/installer:latest \
+  --output-dir ${TMP} \
+  test https://10.254.0.10:6443
+
+timeout -v --preserve-status 1m  bash -c "until ${TALOSCTL} apply-config -n 127.0.0.5 --insecure -f ${TMP}/init.yaml; do sleep 5; done"
```
smira, line 38: Just curious, why 127.0.0.5?

smira, line 38: Should we also wait for apply config to finish and do something super simple as talosctl version with generated talosconfig? This would also test apply-config in maintenance mode

andrewrynhard: Unfortunately we can't since the Talos certificates are not valid for `127.0.0.1`. I tried this. I added `--additional-cert-sans` but that only applies to Kubernetes.

andrewrynhard: The good news is that we test apply-config so we aren't completely untested. Still would be good to get to one day, but changing the functionality of that flag felt out of scope and like a last minute design decision we shouldn't rush.

Unix4ever: I wonder if we can at least check that port 50000 is reachable?

andrewrynhard: typo ... not even sure how the hell that is working.

andrewrynhard: I can add a `telnet` command or something.

Unix4ever: Yep, or `nc -z 127.0.0.1 50000` (if I haven't messed up with the flags).

andrewrynhard: Yep, I did that intead.

Unix4ever (approved):

smira:
one more option (I hate to say that so late), but we could easily add `--cdrom` option to QEMU provisioner so that it "inserts" cdrom if boot disk has no partitions. this should address problems with networking, and we can do full flow (skipping regular health checks in cluster create)

smira:
I have the talosctl part here: https://github.com/talos-systems/talos/pull/2938

andrewrynhard:
> one more option (I hate to say that so late), but we could easily add `--cdrom` option to QEMU provisioner so that it "inserts" cdrom if boot disk has no partitions. this should address problems with networking, and we can do full flow (skipping regular health checks in cluster create)

I am not sure about this idea. This is adding flags to talosctl that really are there only for testing purposes. I have long been thinking we should actually go the opposite direction and pull out a lot of `cluster create` from talosctl and turn it into a test framework. In that case I would agree. That being said, we have to test this so I am not going to die on a hill over this.

smira:
> I am not sure about this idea. This is adding flags to talosctl that really are there only for testing purposes. I have long been thinking we should actually go the opposite direction and pull out a lot of `cluster create` from talosctl and turn it into a test framework. In that case I would agree. That being said, we have to test this so I am not going to die on a hill over this.

we can go either way, we could also hide some flags under --expert or something, but on other hand our users might want to test ISO flow as well, as it's not uncommon to see how config works. we probably need to rethink how we present those options

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  "${TALOSCTL}" config node 172.20.2.2
+}
+
+function destroy_cluster() {
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}" --provisioner "${PROVISIONER}"
+}
+
+create_cluster
+timeout -v --preserve-status 1m  bash -c "until ${TALOSCTL} apply-config -n 127.0.0.1 --insecure -f controlplane.yaml; do sleep 5; done"
```
smira, line 38: I guess `-n` should be 172.20.2.2 ?

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,42 @@
...
+get_kubeconfig
```
smira, line 39: we need `talosctl -n 172.20.0.2 bootstrap` probably with some `timeout`

.drone.jsonnet
```diff
@@ -224,6 +224,7 @@ local unit_tests = Step("unit-tests", depends_on=[initramfs]);
 local unit_tests_race = Step("unit-tests-race", depends_on=[initramfs]);
 local e2e_docker = Step("e2e-docker-short", depends_on=[talos, talosctl_linux, unit_tests, unit_tests_race], target="e2e-docker", environment={"SHORT_INTEGRATION_TEST": "yes", "REGISTRY": local_registry});
 local e2e_qemu = Step("e2e-qemu-short", privileged=true, target="e2e-qemu", depends_on=[talosctl_linux, initramfs, kernel, installer, unit_tests, unit_tests_race, talosctl_cni_bundle], environment={"REGISTRY": local_registry, "SHORT_INTEGRATION_TEST": "yes"}, when={event: ['pull_request']});
+local e2e_iso = Step("e2e-iso", privileged=true, target="e2e-iso", depends_on=[talosctl_linux, initramfs, iso_amd64, unit_tests, unit_tests_race], when={event: ['pull_request']});
```
smira, line 227: guess we need to set `REGISTRY` here to `local_registry`

hack/test/e2e-iso.sh
smira: I believe `get_kubeconfig` depends on `talosctl config node 172.20.2.2`

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,42 @@
...
+run_talos_integration_test
```
smira, line 40: instead of running full test, we could do `talosctl -n 172.20.0.2 cluster health` to save some time in CI, this would only verify bootstrap, but that should be enough?

hack/test/e2e-iso.sh
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function destroy_cluster() {
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}" --provisioner "${PROVISIONER}"
+}
+
+create_cluster
+timeout -v --preserve-status 1m bash -c "until ${TALOSCTL} apply-config -n 172.20.2.2 --insecure -f controlplane.yaml; do sleep 5; done"
+sleep 5
+timeout -v --preserve-status 2m  bash -c "until nc -z 172.20.2.2 50000; do sleep 5; done"
```
smira, line 40: I think we need to boostrap the cluster still to make it pass the health checks

smira (approved):

PR #2931, opened 2020-12-09
feat: add talosctl config contexts

Unix4ever:
Bonus to `talosctl config merge`.
Got that idea after using talosctl for a weekend.
I feel that can be a good addition to have a command that can list existing
contexts in a table view, which is similar to what `kubectl config get-contexts`
does. To avoid going through the file which has all the certs and such.

Called it just `contexts` to align with whatever we have now (to switch
    context you need to use `talosctl config context`).

### DIFF
--- a/cmd/talosctl/cmd/talos/config.go
+++ b/cmd/talosctl/cmd/talos/config.go
@@ -8,6 +8,10 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
 
 	"github.com/spf13/cobra"
 
@@ -74,10 +78,11 @@ var configNodeCmd = &cobra.Command{
 
 // configContextCmd represents the configc context command.
 var configContextCmd = &cobra.Command{
-	Use:   "context <context>",
-	Short: "Set the current context",
-	Long:  ``,
-	Args:  cobra.ExactArgs(1),
+	Use:     "context <context>",
+	Short:   "Set the current context",
+	Aliases: []string{"use-context"},
+	Long:    ``,
+	Args:    cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		context := args[0]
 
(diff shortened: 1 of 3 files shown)


Unix4ever:
Output example:
```bash
CURRENT   NAME               ENDPOINT
          talos-default-32   10.5.0.2
          talos-default-34   10.5.0.2
          talos-default-51   10.5.0.2
          talos-default-57   10.5.0.2
          talos-default-74   192.168.88.29
          talos-default-12   10.5.0.2
          talos-default-15   192.168.88.35
          talos-default-77   10.5.0.2
          talos-default-26   10.5.0.2
          talos-default-71   192.168.88.29
          talos-default-46   10.5.0.2
          talos-default-65   192.168.88.34
          talos-default-68   10.5.0.2
          talos-default-87   192.168.88.27
          talos-test         10.5.0.2
          talos-default-21   10.5.0.2
          talos-default-33   10.5.0.2
          talos-default-54   10.5.0.3
          talos-default-60   10.5.0.2
          talos-default-61   10.5.0.3
          talos-default-75   192.168.88.29
          talos-default-7    10.5.0.2
          talos-default-13   10.5.0.2
          talos-default-27   10.5.0.2
          talos-default-36   10.5.0.2
          talos-default-38   10.5.0.2
          talos-default-73   192.168.88.29
*         talos-default-90   192.168.88.27
```

cmd/talosctl/cmd/talos/config.go
```diff
@@ -175,8 +179,50 @@ func openConfigAndContext(context string) (*clientconfig.Config, error) {
...
+		keys := make([]string, len(c.Contexts))
+		i := 0
+		for key := range c.Contexts {
+			keys[i] = key
+			i++
+		}
+		sort.Strings(keys)
+
+		w := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+		fmt.Fprintln(w, "CURRENT\tNAME\tENDPOINTS")
```
andrewrynhard, line 203: I could see it useful to see the `NODES` as well, but just a suggestion. It could get ugly if they have a lot.

andrewrynhard (approved):

cmd/talosctl/cmd/talos/config.go
Unix4ever: initially I've added them, but then had the same though: can be too much. But we can get first N there I guess.

cmd/talosctl/cmd/talos/config.go
```diff
@@ -175,8 +179,58 @@ func openConfigAndContext(context string) (*clientconfig.Config, error) {
 	return c, nil
 }
 
+// configGetContexts represents config contexts command.
+var configGetContexts = &cobra.Command{
+	Use:    "contexts",
```
smira, line 184: `kubectl` calls this `kubectl config get-contexts` (idk if that is easier to distinguish from `context`)

smira (dismissed):
as a final suggestion we could add a simple test which runs this command (it will just check for success exit code, but it's enough to make sure command isn't broken completely after some change)

cmd/talosctl/cmd/talos/config.go
Unix4ever: and at the same time `kubectl` has `use-context` to switch it, so it's a bit different naming scheme there.

Unix4ever: What do you think about renaming `context` to `use-context`? And then I can rename `contexts` to `get-contexts`.
_Or_ we can have aliases for that.

smira: we certainly could do aliases by the way with `context`/`contexts` being the defaults and k8s aliases secondary

smira (approved):

PR #2939, opened 2020-12-10
feat: update kernel to 5.9.13, new KSPP requirements

smira:
Pulls in following changes:

* https://github.com/talos-systems/toolchain/pull/20
* https://github.com/talos-systems/tools/pull/116
* https://github.com/talos-systems/pkgs/pull/214
* https://github.com/talos-systems/pkgs/pull/215
* https://github.com/talos-systems/pkgs/pull/216
* https://github.com/talos-systems/pkgs/pull/217
* https://github.com/talos-systems/go-procfs/pull/4

New empty amd64 images for u-boot & rpi-firmware reduce the size of
amd64 installer image.

For backwards compatibility QEMU provisioner still injects "legacy" KSPP
kernel args into initial boot environment.

Installer correctly upgrades KSPP options when moving from one version
of Talos to another.


cmd/installer/cmd/iso.go
```diff
@@ -29,7 +29,7 @@ terminal_output console
 menuentry "Talos ISO" {
 	set gfxmode=auto
 	set gfxpayload=text
-	linux /boot/vmlinuz page_poison=1 slab_nomerge slub_debug=P pti=on panic=0 consoleblank=0 earlyprintk=ttyS0 console=tty0 console=ttyS0 talos.platform=metal
+	linux /boot/vmlinuz init_on_alloc=1 init_on_free=1 slab_nomerge pti=on panic=0 consoleblank=0 earlyprintk=ttyS0 console=tty0 console=ttyS0 talos.platform=metal
```
Ulexus, line 32: Not that it should be blocked for it, but do we know what kind of performance hit this double-init causes?

Ulexus (approved):

cmd/installer/cmd/iso.go
andrewrynhard: Good question. I see https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1862822. Looks related to ZFS.

andrewrynhard: Also https://github.com/openzfs/zfs/issues/9910.

andrewrynhard: https://cateee.net/lkddb/web-lkddb/INIT_ON_FREE_DEFAULT_ON.html warns of performance issues too. I don't know what the answer is if we want to be kspp compliant :thinking: . Do we just leave the old `page_poison` and `slab_*` options?

smira: I guess old options are not much better in terms of performance, but that's my guess

smira: @andrewrynhard we talked a bit about it, open option is to skip checking for `init_on_alloc` being set to 1, so that users can disable it via command line args if they need to (but default is still on)

andrewrynhard: 👍

smira: in fact, `init_on_free` has more performance hit, so we could potentially allow disabling it as well for those who care about performance more than security

smira: oh, I just realized the way our installer works, it always enforces these options (as it appends them to the command line last), so there's no easy way to disable them except for mounting `/boot` and editing grub config

andrewrynhard: Sounds like we should make this whole optional thing in 0.9

smira: one way around is to let people override defaults with extra kernel args

smira:
Changes:

* don't enforce `init_on_alloc=1` and `init_on_free=1` - they're still defaults in the kernel, and we always append them to the command line by default
* allow those to be overridden via installer extra args in case user hits severe performance issue

andrewrynhard (approved):

PR #2944, opened 2020-12-11
test: add an extra 'node boot done' health check

smira:
This makes sure node boot sequence is done before we consider cluster to
be healthy.

### DIFF
--- a/pkg/cluster/check/events.go
+++ b/pkg/cluster/check/events.go
@@ -1,0 +1,93 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package check
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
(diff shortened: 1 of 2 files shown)


smira:
almost passed integration: https://ci.dev.talos-systems.io/talos-systems/talos/13775

looks like got into conflict with some other test on network/interfaces, but not sure which one :puzzled:

pkg/cluster/check/events.go
```diff
@@ -0,0 +1,93 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
+	"github.com/talos-systems/talos/pkg/machinery/client"
+)
+
+// NodeBootedAssertion checks whether nodes reached end of 'Boot' sequence.
+//nolint: gocyclo
+func NodeBootedAssertion(ctx context.Context, cluster ClusterInfo) error {
```
Ulexus, line 22: Could we call this AllNodesBooted or ClusterNodesBootod to better reflect its multi-nodal operation?

Ulexus (dismissed):
naming nit

pkg/cluster/check/events.go
smira: renamed to `AllNodesBootedAssertion`

PR #2951, opened 2020-12-16
docs: add a note for being careful about enabling debug flag

Unix4ever:
Enabling debug on slow serial consoles degrades Talos bootstrap
performance and so Talos nodes never get configured properly.


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -350,6 +350,9 @@ type Config struct {
 	ConfigVersion string `yaml:"version"`
 	//   description: |
 	//     Enable verbose logging to the console.
+	//     All system containers logs will flow into serial console.
+	//
+	//     > Note: To avoid breaking Talos bootstrap flow enable this option only if serial console can handle high message throughoutput.
```
smira, line 355:
```suggestion
	//     > Note: To avoid breaking Talos bootstrap flow enable this option only if serial console can handle high message throughput.
```

smira (approved):

PR #2953, opened 2020-12-17
docs: add fallback to default description if none is set on current page

Unix4ever:
Also filled in some of the descriptions specific for particular pages.
Seeing that bug all the time was making me sad:
image


smira (dismissed):
this `undefined` was bugging me as well

rsmitty (dismissed):

andrewrynhard (approved):

PR #2957, opened 2020-12-18
fix: properly define shorthand in `talosctl time` command

Unix4ever:
Previously it was defaulting to `c`.

Signed-off-by: Artem Chernyshev <artem.0xd2@gmail.com>

### DIFF
--- a/cmd/talosctl/cmd/talos/time.go
+++ b/cmd/talosctl/cmd/talos/time.go
@@ -21,6 +21,10 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/client"
 )
 
+var timeCmdFlags struct {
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/talos/time.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/client"
 )
 
+var ntpServer = ""
```
smira, line 24: nit 
```suggestion
var ntpServer string
```

```suggestion
var timeCmdFlags struct { 
    ntpServer string
 }
```

smira (approved):

PR #2958, opened 2020-12-18
fix: sync RTC in timed, sync time before fetching packet metadata

smira:
1. When time is set by timed, also sync RTC with system time, otherwise
time might be off after a reboot.

2. Before fetching Packet (Equinix Metal) metadata over `https://`, do
one-time time sync in the background, as if the clock skew is big
enough, TLS certificate will be considered invalid and boot fails.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -67,6 +67,7 @@ require (
 	github.com/talos-systems/grpc-proxy v0.2.0
 	github.com/talos-systems/net v0.2.0
 	github.com/talos-systems/talos/pkg/machinery v0.0.0-20200818212414-6a7cc0264819
+	github.com/u-root/u-root v7.0.0+incompatible
 	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae // indirect
 	github.com/vmware-tanzu/sonobuoy v0.19.0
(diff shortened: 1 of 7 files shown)


rsmitty (dismissed):

internal/app/machined/pkg/runtime/v1alpha1/platform/packet/packet.go
```diff
@@ -77,6 +79,24 @@ func (p *Packet) Name() string {
...
+			return
+		}
+
+		if err = ntpClient.QueryAndSetTime(ntpCtx); err != nil {
+			if !stderrors.Is(err, context.Canceled) {
+				log.Printf("failed to do one-time time sync: %s", err)
+			}
+		}
+	}()
+
```
andrewrynhard, line 99: Feels like maybe we can put this in a more generic place? Maybe in the task that calls the platform interface `Configuration` method?

andrewrynhard (commented):
LGTM, but just one suggestion.

internal/app/machined/pkg/runtime/v1alpha1/platform/packet/packet.go
smira: yep, I think it makes perfect sense to do that there.

andrewrynhard (dismissed):

PR #2963, opened 2020-12-18
feat: ignore bootstrap API request if cluster is already bootstrapped

smira:
This refactors some initialized checks into common etcd client code and
adds a check to the bootstrap API.


andrewrynhard:
We thinking we don't need this anymore?

smira:
> We thinking we don't need this anymore?

I decided to close that as it doesn't quite work with API timeouts and causes frequent failures, probably in the new sequencer it will be much easier to implement it some other way.

PR #2970, opened 2020-12-21
docs: fix SBC docs to point to beta.0 instead of beta.1

Unix4ever:
We don't have beta.1 yet.


smira (approved):
Thanks!

PR #2976, opened 2020-12-21
feat: support disk image in talosctl cluster create

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2973

Can now supply disk image using `--disk-image-path` flag.
May need to enable `--with-apply-config` if it's necessary to bootstrap
nodes properly.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -348,6 +348,11 @@ local integration_uefi = Step("e2e-uefi", target="e2e-qemu", privileged=true, de
         "WITH_UEFI": "true",
         "REGISTRY": local_registry,
 });
+local integration_disk_image = Step("e2e-disk-image", target="e2e-qemu", privileged=true, depends_on=[integration_uefi], environment={
+        "SHORT_INTEGRATION_TEST": "yes",
+        "USE_DISK_IMAGE": "true",
+        "REGISTRY": local_registry,
+});
 local push_edge = {
   name: 'push-edge',
   image: 'autonomy/build-container:latest',
@@ -379,6 +384,7 @@ local integration_steps = default_steps + [
   integration_provision_tests_track_1,
   integration_cilium,
   integration_uefi,
+  integration_disk_image,
   push_edge,
 ];
 
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -51,6 +51,8 @@ var (
 	nodeVmlinuzPath         string
 	nodeInitramfsPath       string
 	nodeISOPath             string
+	nodeDiskImagePath       string
+	applyConfigEnabled      bool
 	bootloaderEnabled       bool
 	uefiEnabled             bool
 	configDebug             bool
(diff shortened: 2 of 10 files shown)


hack/test/e2e-qemu.sh
```diff
@@ -35,6 +35,16 @@ case "${WITH_UEFI:-false}" in
     ;;
 esac
 
+case "${USE_DISK_IMAGE:-false}" in
+  false)
+    DISK_IMAGE_FLAG=
+    ;;
+  *)
+    tar -xf _out/metal-amd64.tar.gz -C _out/ --strip-components=1
```
Unix4ever, line 43: I guess that's most likely a wrong assumption that this image exists at this point.
Can I call `make image-metal` here? Or that's a bad idea?

smira: it should be created as part of the build by `images-amd64` step, so we could rely on it I think

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -563,6 +573,8 @@ func init() {
 	createCmd.Flags().StringVar(&nodeVmlinuzPath, "vmlinuz-path", helpers.ArtifactPath(constants.KernelAssetWithArch), "the compressed kernel image to use")
 	createCmd.Flags().StringVar(&nodeISOPath, "iso-path", "", "the ISO path to use for the initial boot (VM only)")
 	createCmd.Flags().StringVar(&nodeInitramfsPath, "initrd-path", helpers.ArtifactPath(constants.InitramfsAssetWithArch), "the uncompressed kernel image to use")
+	createCmd.Flags().StringVar(&nodeDiskImagePath, "disk-image-path", "", "disk image to use")
+	createCmd.Flags().BoolVar(&applyConfigEnabled, "with-apply-config", false, "enable apply config when the VM is starting in maintenance mode (QEMU only)")
```
smira, line 577: nit: I think this should work with docker as well?

pkg/provision/options.go
```diff
@@ -111,6 +120,9 @@ type Options struct {
 	// Enable UEFI (for amd64), arm64 can only boot UEFI
 	UEFIEnabled bool
 
+	// Push the config using apply-config --insecure instead of serving it through the http server
+	ApplyConfig bool
```
smira, line 124: I don't think we need this to be part of the provision library, as this is done in `talosctl cluster create`?

smira (commented):
lgtm, just some small things

pkg/provision/options.go
Unix4ever: right, I forgot to remove it from this place

.drone.jsonnet
```diff
@@ -348,6 +348,11 @@ local integration_uefi = Step("e2e-uefi", target="e2e-qemu", privileged=true, de
         "WITH_UEFI": "true",
         "REGISTRY": local_registry,
 });
+local integration_disk_image = Step("e2e-disk-image", target="e2e-qemu", privileged=true, depends_on=[integration_qemu], environment={
```
smira, line 351: can we make this `depends_on` on `integration_uefi`, so that we serialize those steps, as usually by the time this runs provision tests are still running, and it might be too many tests running in parallel

smira (dismissed):

.drone.jsonnet
Unix4ever: sure

smira (approved):

PR #2983, opened 2020-12-22
feat: add commands to manage/query etcd cluster

Unix4ever:
Used already existing protobufs for that.

Commands:
`talosctl etcd members -n <node>`
`talosctl etcd leave -n <node>`
`talosctl etcd forfeit-leadership -n <node>`

Fixes: https://github.com/talos-systems/talos/issues/2978


smira:
it might make sense to add an option to query local etcd in member list, as this might be what users are expecting

internal/integration/cli/etcd.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// SuiteName ...
+func (suite *EtcdSuite) SuiteName() string {
+	return "cli.EtcdSuite"
+}
+
+// TestMembers etcd members should have some output.
+func (suite *EtcdSuite) TestMembers() {
+	suite.RunCLI([]string{"etcd", "members", "--nodes", suite.RandomDiscoveredNode()}) // default checks for stdout not empty
```
smira, line 25: I think we should limit that to control plane nodes only (`suite.RandomDiscoveredNode()` has an argument for that)

internal/integration/cli/etcd.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// TestForfeitLeadership etcd forfeit-leadership check.
+func (suite *EtcdSuite) TestForfeitLeadership() {
+	suite.RunCLI([]string{"etcd", "forfeit-leadership", "--nodes", suite.RandomDiscoveredNode()}) // default checks for stdout not empty
```
smira, line 30: same here

smira (approved):

PR #2987, opened 2020-12-22
fix: checkpoint controller-manager and scheduler

smira:
Default manifests created by bootkube so far were only enabling
pod-checkpointer for kube-apiserver. This seems to have issues with
single-node control plane scenario, when without scheduler and
controller-manager node might fall into `NodeAffinity` state.

See https://github.com/talos-systems/bootkube-plugin/pull/23

### DIFF
--- a/hack/test/e2e-docker.sh
+++ b/hack/test/e2e-docker.sh
@@ -14,7 +14,8 @@ function create_cluster {
     --provisioner "${PROVISIONER}" \
     --name "${CLUSTER_NAME}" \
     --image "${IMAGE}" \
-    --masters=3 \
+    --masters=1 \
+    --workers=1 \
     --mtu 1450 \
     --memory 2048 \
     --cpus 2.0 \
--- a/internal/integration/api/recover.go
+++ b/internal/integration/api/recover.go
@@ -53,6 +53,8 @@ func (suite *RecoverSuite) TearDownTest() {
 
 // TestRecoverControlPlane removes the control plane components and attempts to recover them with the recover API.
 func (suite *RecoverSuite) TestRecoverControlPlane() {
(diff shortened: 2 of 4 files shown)


andrewrynhard (dismissed):

Unix4ever (dismissed):

andrewrynhard (approved):

smira:
Clean integration run: https://ci.dev.talos-systems.io/talos-systems/talos/14108

hack/test/e2e-docker.sh
```diff
@@ -14,7 +14,8 @@ function create_cluster {
     --provisioner "${PROVISIONER}" \
     --name "${CLUSTER_NAME}" \
     --image "${IMAGE}" \
-    --masters=3 \
+    --masters=1 \
```
smira, line 17: docker mode doesn't use loadbalancer, and control plane endpoint points to the first (bootstrap node)

without a loadbalancer, there's always a chance that control plane pods are not running on bootstrap node by the time bootkube tears down boostrap pods (but it is running on other control plane nodes). so running HA control plane in docker mode is kind of pointless and leading to flakes

internal/integration/api/recover.go
```diff
@@ -53,6 +53,8 @@ func (suite *RecoverSuite) TearDownTest() {
 
 // TestRecoverControlPlane removes the control plane components and attempts to recover them with the recover API.
 func (suite *RecoverSuite) TestRecoverControlPlane() {
+	suite.T().Skip("with checkpoints enabled for kube-scheduler and kube-controller-manager this test no longer makes sense")
```
smira, line 56: we can't test recover anymore as checkpoints don't allow scheduler & controller-manager pods to be killed

Unix4ever (dismissed):

PR #2989, opened 2020-12-23
fix: correctly transport gRPC errors from apid

smira:
Before these changes, errors were always sent as strings, so if original
error was gRPC error (which is almost always the case for apid), it is
formatted as string and original fields (like code) are lost in the
formatted string.

With this change, apid sends errors as official `grpc.Status` protobuf
structure, and client decodes that into Go grpc.Status based error.

This change is backwards and forwards compatible.

This should fix more cases when integration tests were not able to
ignore grpc `transport is closing` errors when they were sent as strings
from the apid endpoint.


Unix4ever (dismissed):

Unix4ever (approved):

Unix4ever (approved):

smira:
clean integration run: https://ci.dev.talos-systems.io/talos-systems/talos/14031

PR #2999, opened 2020-12-24
refactor: define default kernel flags in machinery instead of procfs

Unix4ever:
That change should make Talos updates more straightforward in any
projects that depend on Talos.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -20,6 +20,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub"
 	"github.com/talos-systems/talos/internal/pkg/mount"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
+	"github.com/talos-systems/talos/pkg/machinery/kernel"
 	"github.com/talos-systems/talos/pkg/version"
 )
 
@@ -48,7 +49,9 @@ func Install(p runtime.Platform, seq runtime.Sequence, opts *Options) (err error
 	cmdline.SetAll(p.KernelArgs().Strings())
 
 	// first defaults, then extra kernel args to allow extra kernel args to override defaults
-	cmdline.AppendDefaults()
+	if err = cmdline.AppendAll(kernel.DefaultArgs); err != nil {
+		return err
+	}
 
 	if err = cmdline.AppendAll(opts.ExtraKernelArgs); err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -48,7 +49,9 @@ func Install(p runtime.Platform, seq runtime.Sequence, opts *Options) (err error
 	cmdline.SetAll(p.KernelArgs().Strings())
 
 	// first defaults, then extra kernel args to allow extra kernel args to override defaults
-	cmdline.AppendDefaults()
+	if err = cmdline.AppendAll(config.DefaultKernelArgs()); err != nil {
```
smira, line 52: we should be using default cmdline args also in QEMU/Firecracker provisioners

smira: might be easier to pull in go-procfs without default args and see where things break :)

Unix4ever: Well, yeah. I guess we can do that, probably need to change `Append` to `Set` for whatever flags are added in the provider.

Unix4ever: oh, I see. There's also NewDefaultCmdline... Ok, I guess I'll update go-procfs now

Unix4ever: Should be good to go now.

go.sum
```diff
@@ -937,6 +937,8 @@ github.com/talos-systems/go-procfs v0.0.0-20201210152942-5a9a4a75d559 h1:wHuwJhQ
 github.com/talos-systems/go-procfs v0.0.0-20201210152942-5a9a4a75d559/go.mod h1:ATyUGFQIW8OnbnmvqefZWVPgL9g+CAmXHfkgny21xX8=
 github.com/talos-systems/go-procfs v0.0.0-20201223150035-a82654edcec1 h1:n6q6VlyJp5yyCAzUyds+CFDLW6iDvTDSDXoI3NoXpcs=
 github.com/talos-systems/go-procfs v0.0.0-20201223150035-a82654edcec1/go.mod h1:ATyUGFQIW8OnbnmvqefZWVPgL9g+CAmXHfkgny21xX8=
+github.com/talos-systems/go-procfs v0.0.0-20201224133103-24d06a955782 h1:q9g1aGWfmbUtOcON2BIJUP9JfsW58gWdMlhuZ+fVQ+8=
```
smira, line 940: super nit, but running `go mod tidy` should remove old version(s) of `go-procfs`

pkg/machinery/config/kernel.go
```diff
@@ -0,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package config
```
smira, line 5: feels like this probably doesn't belong to `config` package, should it be better `package kernel` in `machinery` ? with `kernel.DefaultArgs` ?

Unix4ever: Yep, if you think it deserves it's own personal folder :)

smira (approved):

PR #3012, opened 2020-12-31
feat: support Wireguard networking

Unix4ever:
This the first iteration of Wireguard network support.
What was done:
- kernel was updated to enable Wireguard kernel module.
- changed networkd to support creating Wireguard device type.
- used wgctrl to configure wireguard.
- updated `talosctl cluster create` to support generating Wireguard
network configuration automatically by just specifying the network cidr.
- added docs about Wireguard support/how to use it.

Fixes: https://github.com/talos-systems/talos/issues/3025


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -330,6 +330,30 @@ var (
...
+	}
+
+	networkConfigWireguardPeerExample = &DeviceWireguardConfig{
+		WireguardPrivateKey: "+AdBen89qXzYCrheNYabDZ/X0JbvUajoBl4BwChhNWU=",
+		WireguardPeers: []*DeviceWireguardPeer{
+			{
+				WireguardPublicKey:                   "Ac3M3kTLwN57uGobYizrO9jeuk4Th3v7ET101jHr7Xw=",
+				WireguardEndpoint:                    "192.168.1.2",
+				WireguardPersistentKeepaliveInterval: time.Second * 10,
+				WireguardAllowedIPs:                  []string{"192.168.1.0/24"},
```
smira, line 352: Should this be Wireguard network IPs or "real" IPs?

website/content/docs/v0.8/Guides/configuring-wireguard-network.md
```diff
@@ -0,0 +1,105 @@
...
+        listenPort: 51820
+        peers:
+          allowedIPs:
+            - 192.168.88.0/24
+          endpoint: 209.202.254.14.8172
+          publicKey: ABCDEF...
+...
+```
+
+When `networkd` gets this configuration it will create the device, configure it and will do:
```
smira, line 99: ... and it will  bring up the network interface? (instead of bash command)

Unix4ever: hmm... ok, seems that's confusing. Will rephrase :)

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
Unix4ever: WireguardNetwork IPs as far as I understand.

andrewrynhard:
@Unix4ever you need to `/build` again after a new push. Looks like asset download is failing.

smira (approved):

PR #2993, opened 2020-12-23
feat: support type filter in list API and CLI

AlekSi:
# Pull Request

## What? (description)

Support type filter in list API and CLI.

## Why? (reasoning)

#2068.

## Acceptance

Please use the following checklist:

- [x] you linked an issue (if applicable)
- [x] you included tests (if applicable)
- [x] you ran conformance (`make conformance`)
- [x] you formatted your code (`make fmt`)
- [x] you linted your code (`make lint`)
- [x] you generated documentation (`make docs`)
- [x] you ran unit-tests (`make unit-tests`)

> See `make help` for a description of the available targets.

PR #3035, opened 2021-01-14
fix: list command unlimited recursion shouldn't be default behavior

Unix4ever:
Revert back to old behavior.


smira (approved):
:+1:

smira:
cc @AlekSi (bug was introduced in #2993)

PR #3041, opened 2021-01-20
feat: replace bootkube with Talos-managed control plane

smira:
Control plane components are running as static pods managed by the
kubelets.

Whole subsystem is managed via resources/controllers from os-runtime.

Many supporting changes/refactoring to enable new code paths.


internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -0,0 +1,435 @@
...
+
+	//nolint: dupl
+	return r.Update(ctx, k8s.NewStaticPod(k8s.ControlPlaneNamespaceName, "kube-controller-manager", nil), func(r resource.Resource) error {
+		r.(*k8s.StaticPod).SetPod(&v1.Pod{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "Pod",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "kube-controller-manager",
```
Unix4ever, line 303: I remember that bootkube adds node name to the static pod name. I guess that's to avoid conflicts with non-static ones when they start.
Here I guess they always run as static pods, right?

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -0,0 +1,435 @@
...
+			return fmt.Errorf("error getting machine type: %w", err)
+		}
+
+		machineType := machineTypeRes.(*config.MachineType).MachineType()
+
+		if machineType != machine.TypeControlPlane && machineType != machine.TypeInit {
+			if err = ctrl.teardownAll(ctx, r); err != nil {
+				return fmt.Errorf("error destroying static pods: %w", err)
+			}
+		}
```
Unix4ever, line 90: Should this condition interrupt further loop execution?

smira: yep, it is missing `continue`, I believe I fixed in the recent push, but I need to recheck

smira:
image

smira:
image

andrewrynhard (approved):

PR #3042, opened 2021-01-21
refactor: update go-blockdevice and restructure disk interaction code

Unix4ever:
This refactoring is required to simplify the work to be done to support
disk encryption.

Tried to minimize amount of queries done by `blockdevice` `probe`
methods.
Instead, where we have `runtime.Runtime` we get all required blockdevices
there from blockdevice cache stored in `State().Machine().Disk()`.
This opens a way to store encryption settings in the `Partition`
objects.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -9,7 +9,7 @@ import (
 	"log"
 	"path/filepath"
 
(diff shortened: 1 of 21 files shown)


smira (approved):
lgtm, given that it passes full test suite ;)

smira (approved):

PR #3047, opened 2021-01-27
fix: open blockdevices with exclusive flock for partitioning

smira:
This fixes spurious race conditions when user disks are partitioned
and formatted in `mountUserDisks` task. While this task runs, `udevd` is
running to allow various `/dev/` symlinks to be used for user disks.
At the same time `udevd` might trigger syscall `BLKRRPART` at any time
concurrently with Talos which leads to a race on kernel side when Talos
tries to update kernel partition table while kernel does it on its own
as a result of `udevd` call.

As part of the fix, `RereadPartitionTable()` calls were removed (they
trigger `BLKRRPART` and they're not needed as Talos updates partition
table on its own).

Some cleanups to make sure blockdevice is open/closed just in matching
pairs (no lingering open blockdevice instances). This is import for
`WithExclusiveLock()` calls, as it would lead to a deadlock if previous
blockdevice instance is not closed.

### DIFF
--- a/cmd/installer/pkg/install/manifest.go
+++ b/cmd/installer/pkg/install/manifest.go
@@ -233,7 +233,7 @@ func (m *Manifest) executeOnDevice(device Device, targets []*Target) (err error)
 
 	var bd *blockdevice.BlockDevice
 
-	if bd, err = blockdevice.Open(device.Device); err != nil {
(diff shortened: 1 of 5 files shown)


smira:
PR is draft, as it uses blockdevice library intermediate version, waiting for Artem's refactoring PR to be merged first.

Unix4ever (approved):

andrewrynhard (approved):

PR #3053, opened 2021-01-28
fix: kill all processes and umount all disk on reboot/shutdown

smira:
There are several ways Talos node might be restarted or shut down:

* error in sequence (initiated from machined)
* panic in main goroutine (machined recovers panics)
* error in sequence (initiated via API, event caught by machined)
* reboot/shutdown via Talos API

Before this change, paths (1) and (2) were handled in machined, and no
disks were unmounted and processes killed, so technically all the
processes are running and potentially writing to the filesystems.
Paths (3) and (4) try to stop services (but not pods) and unmount
explicitly mounted filesystems, followed by reboot directly from
sequencer (bypassing machined handler).

There was a bug that user disks were never explicitly unmounted (but
they might have been unmounted if mounted on top `/var`).

This refactors all the reboot/shutdown paths to flow through machined's
main function: on paths (4) event is sent via event API from the
sequencer back to the machined and machined initiates proper shutdown
sequence.

Refactoring in machined leads to all the paths (1)-(4) flowing through
the same function `handle(error)`.

Added two additional checks before flushing buffers:

* kill all non-system processes, this also kills all mount namespaces
* unmount any filesystem backed by `/dev/*`

This ensures all filesystems are unmounted before buffers are flushed.


andrewrynhard (dismissed):

smira:
re-pushed with a bug fix - `Reboot()` task in sequencer is also killed from `Init` sequence which requires different handing in machined/main.go

smira:
integration passed: https://ci.dev.talos-systems.io/talos-systems/talos/14457

Unix4ever (dismissed):

smira:
re-pushed with cosmetic changes to remove "rebootError" from logs (it looked like something failed, but we use rebootError to pass reboot context back to machined)

Unix4ever (approved):

Issue #3046, opened 2021-01-27
Huge pages support


Born2Bake:
Hello,

I am wondering, if there is possibility to configure hp on machines? I want to test https://mayastor.gitbook.io/introduction/quickstart/prerequisites new storage solution but it requires hp to be enabled on machines. Thank you

andrewrynhard:
@Born2Bake We have support for tuning the kernel via the `sysctl` key under `machine`. Something like this should work:

```
machine:
  sysctls:
    vm.nr_hugepages: "10"
```

You will probably need to update it with the correct value.

Ulexus:
The sysctl is insufficient alone.  The kubelet needs the hugepages' hugetlbfs mounted, as well:

```
cgroup on /sys/fs/cgroup/hugetlb type cgroup (rw,nosuid,nodev,noexec,relatime,hugetlb)
hugetlbfs on /dev/hugepages type hugetlbfs (rw,relatime,pagesize=2M)
```

DWSR:
Looks like https://github.com/talos-systems/talos/pull/3058 will add Huge pages support.

Born2Bake:
Just to confirm, in order to use hugepages, I would need to update talos to the next release and add
```
machine:
  sysctls:
    vm.nr_hugepages: "512"
```
to server configs?

smira:
yes, exactly, it should be 0.9.0-alpha.0 when it gets released

PR #3058, opened 2021-01-29
feat: mount hugetlbfs

andrewrynhard:
This adds a hugetlbfs. Some applications may depend on it.


internal/pkg/mount/pseudo.go
```diff
@@ -26,6 +26,7 @@ func PseudoSubMountPoints() (mountpoints *Points, err error) {
 	pseudo := NewMountPoints()
 	pseudo.Set("devshm", NewMountPoint("tmpfs", "/dev/shm", "tmpfs", unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV|unix.MS_RELATIME, ""))
 	pseudo.Set("devpts", NewMountPoint("devpts", "/dev/pts", "devpts", unix.MS_NOSUID|unix.MS_NOEXEC, "ptmxmode=000,mode=620,gid=5"))
+	pseudo.Set("hugetlb", NewMountPoint("none", "/dev/hugepages", "hugetlbfs", 0, ""))
```
Ulexus, line 29: Should the source be "hugetlbfs" here instead of "none", or is that taken care of by the kernel?

Ulexus (dismissed):

Ulexus:
Fixes #3046

internal/pkg/mount/pseudo.go
andrewrynhard: Good question. As far as I can tell `none` should work, and the fstype is the important bit. Still needs testing though.

smira (approved):

PR #3066, opened 2021-01-29
chore: add script for starting registry proxies

AlekSi:
To avoid copying and pasting it from the documentation every time.

Signed-off-by: Alexey Palazhchenko <alexey.palazhchenko@gmail.com>

### DIFF
--- a/hack/start-registry-proxies.sh
+++ b/hack/start-registry-proxies.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Sync changes with configuring-pull-through-cache.md.
+
+set -e
+
+docker run -d -p 5000:5000 \
+    -e REGISTRY_PROXY_REMOTEURL=https://registry-1.docker.io \
+    --restart always \
+    --name registry-docker.io registry:2
+
+docker run -d -p 5001:5000 \
+    -e REGISTRY_PROXY_REMOTEURL=https://k8s.gcr.io \
+    --restart always \
+    --name registry-k8s.gcr.io registry:2
+
+docker run -d -p 5002:5000 \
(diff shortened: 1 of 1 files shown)


hack/start-registry-proxies.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Sync changes with configuring-pull-through-cache.md.
+# com.github.AlekSi.docker-ps.group label is for https://github.com/AlekSi/docker-ps-bitbar
+
+set -e
+
+docker run -d -p 5000:5000 \
+    -e REGISTRY_PROXY_REMOTEURL=https://registry-1.docker.io \
+    --restart always \
+    --label com.github.AlekSi.docker-ps.group=talos-registry \
```
smira, line 11: this seems to be a bit more personal than something which should go upstream? :)

AlekSi: If `hack/` directory is meant for Talos developers and contributors, not for user and customers, then I don't see a big problem: that label does nothing for people who don't BitBar/SwiftBar and my plugin, it is useful for those who do use it (like me – I had to sync this script across my machines), and there is a comment explaining what's that.

That said, I don't mind removing it if the above is not convincing :)

smira: we could probably push it via the environment variable to make it configurable at least? and anyone could override if they do labels on their own?

we are actually looking into moving registry mirrors to some easier format in the near term

AlekSi: `--label` flag value can't be empty, that makes environment variable usage way to complicated for that small script. Removed labels completely.

smira (approved):

Issue #3068, opened 2021-01-29
apid loses connectivity to trustd if master IP changes or master goes down (workers)


smira:
## Bug Report

apid on workers discovers master endpoints once on startup, finds first working node and sticks to it for cert issue/renewal

if master goes down or gets replaced, apid can't renew certs anymore, so eventually certs expire and connectivity is lost

### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`]
- Kubernetes version: [`kubectl version --short`]
- Platform:

PR #3069, opened 2021-01-29
fix: use grpc load-balancing when connecting to trustd

smira:
Instead of doing our homegrown "try all the endpoints" method,
use gRPC load-balancing across configured endpoints.

Generalize load-balancer via gRPC resolver we had in Talos API client,
use it in remote certificate generator code. Generalized resolver is
still under `machinery/`, as `pkg/grpc` is not in `machinery/`, and we
can't depend on Talos code from `machinery/`.

Related to: #3068

Full fix for #3068 requires dynamic updates to control plane endpoints
while apid is running, this is coming in the next PR.


AlekSi:
There is a typo: "locad" instead of "load"

andrewrynhard (approved):
/lgtm

PR #3079, opened 2021-01-30
feat: allow fqdn to be used when registering k8s node

rsmitty:
This PR fixes a problem we had with AWS clusters. We now allow the
kubelet to register using the full fqdn instead of just hostname. This is done by passing the `registerWithFQDN` boolean in the kubelet section of our config.

It also introduces the idea of a "Node Name" in the runtime. This is specific to times where we are talking directly to kubernetes, so that we're not just running `os.Hostname()` everywhere.

### DIFF
--- a/internal/app/machined/pkg/runtime/runtime.go
+++ b/internal/app/machined/pkg/runtime/runtime.go
@@ -16,4 +16,5 @@ type Runtime interface {
 	State() State
 	Events() EventStream
 	Logging() LoggingManager
+	NodeName() (string, error)
 }
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
@@ -6,6 +6,9 @@ package v1alpha1
 
 import (
 	"fmt"
+	"log"
+	"os"
+	"syscall"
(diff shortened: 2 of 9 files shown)


internal/app/machined/internal/server/v1alpha1/v1alpha1_monitoring.go
```diff
@@ -19,7 +19,7 @@ import (
 
 // Hostname implements the machine.MachineServer interface.
 func (s *Server) Hostname(ctx context.Context, in *empty.Empty) (*machine.HostnameResponse, error) {
-	hostname, err := os.Hostname()
+	hostname, err := s.Controller.Runtime().NodeName()
```
rsmitty, line 22: I'm not _sure_ that we need the nodename here. @smira maybe you know?

smira: This is only used in monitoring APIs, e.g. dashboard. I would rather leave it as os.Hostname, as it's Talos hostname, not Kubernetes node name

rsmitty: cool, yeah that's what I thought. updated in amended commit to roll that back.

smira (approved):

PR #3084, opened 2021-02-01
feat: move to ECDSA keys for all Kubernetes/etcd certs and keys 

smira:
ECDSA keys are smaller which decreases Talos config size, they are more
efficient in terms of key generation, signing, etc., so it makes boot
performance better (and config generation as well).


smira:
need to preserve RSA generation for upgrade tests (at the very least)

rsmitty (approved):

Unix4ever (approved):

PR #3099, opened 2021-02-02
fix: update DHCP client to use Request-Ack sequence after an Offer

smira:
The problem was that our DHCP client was always doing
Discover-Offer-Request-Ack sequence as if it doesn't have any IP lease
which was breaking some DHCP servers leading to an error which in turn
makes Talos retry harder hitting the same error over and over again.


smira:
thinking more about it, we should probably clear an offer only once TTL runs out?

smira:
> thinking more about it, we should probably clear an offer only once TTL runs out?

I couldn't find any better strategy than clearing an offfer on error

Ulexus (approved):

andrewrynhard (approved):

PR #3101, opened 2021-02-03
feat: implement ephemeral partition encryption

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3106

This PR introduces the first part of disk encryption support.
New config section `systemDiskEncryption` was added into MachineConfig.
For now it contains only Ephemeral partition encryption.

Encryption itself supports two kinds of keys for now:
- generated which can be generated from a template.
- static key, which is hardcoded into the config (mainly for test purposes).

Talosctl cluster create can now be told to encrypt ephemeral partition
by using `--encrypt-ephemeral` flag.

Additionally:
- updated pkgs library version.
- changed Dockefile to copy cryptsetup deps from pkgs.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 
 	"github.com/talos-systems/go-blockdevice/blockdevice"
+	"github.com/talos-systems/go-blockdevice/blockdevice/probe"
 	"github.com/talos-systems/go-procfs/procfs"
 	"golang.org/x/sys/unix"
 
@@ -182,10 +183,34 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 	}
 
 	// Mount the partitions.
+	mountpoints := mount.NewMountPoints()
 
-	mountpoints, err := i.manifest.SystemMountpoints()
-	if err != nil {
-		return err
+	for _, label := range []string{constants.BootPartitionLabel, constants.EFIPartitionLabel} {
+		err = func() error {
+			var bd *probe.ProbedBlockDevice
+
+			bd, err = probe.GetDevWithPartitionName(label)
+			if err != nil {
+				return err
+			}
+
+			defer bd.Close() //nolint:errcheck
+
+			var mountpoint *mount.Point
+
+			mountpoint, err = mount.SystemMountPointForLabel(bd.BlockDevice, label)
+			if err != nil {
+				return err
+			}
+
+			mountpoints.Set(label, mountpoint)
+
+			return nil
+		}()
+
+		if err != nil {
+			return err
+		}
 	}
 
 	if err = mount.Mount(mountpoints); err != nil {
--- a/cmd/installer/pkg/install/manifest.go
+++ b/cmd/installer/pkg/install/manifest.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/talos-systems/go-blockdevice/blockdevice"
 	"github.com/talos-systems/go-blockdevice/blockdevice/partition/gpt"
 	"github.com/talos-systems/go-blockdevice/blockdevice/util"
@@ -22,6 +23,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/board"
 	"github.com/talos-systems/talos/internal/pkg/mount"
+	"github.com/talos-systems/talos/internal/pkg/partition"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
@@ -131,16 +133,12 @@ func NewManifest(label string, sequence runtime.Sequence, bootPartitionFound boo
 
 	stateTarget := StateTarget(opts.Disk, &Target{
 		PreserveContents: bootPartitionFound,
-		ExtraPreserveSources: []PreserveSource{
-			{
-				Label:          constants.LegacyBootPartitionLabel,
-				FileSystemType: FilesystemTypeVFAT,
-				FnmatchFilters: []string{"config.yaml"},
-			},
+		FormatOptions: &partition.FormatOptions{
+			FileSystemType: partition.FilesystemTypeNone,
 		},
 	})
 
-	ephemeralTarget := EphemeralTarget(opts.Disk, nil)
+	ephemeralTarget := EphemeralTarget(opts.Disk, NoFilesystem)
 
 	if opts.Force {
 		ephemeralTarget.Force = true
@@ -417,7 +415,7 @@ func (m *Manifest) preserveContents(device Device, targets []*Target) (err error
 
 		var (
 			sourcePart     *gpt.Partition
-			fileSystemType FileSystemType
+			fileSystemType partition.FileSystemType
 			fnmatchFilters []string
 		)
 
@@ -466,11 +464,11 @@ func (m *Manifest) restoreContents(targets []*Target) error {
 }
 
 // SystemMountpoints returns list of system mountpoints for the manifest.
-func (m *Manifest) SystemMountpoints() (*mount.Points, error) {
+func (m *Manifest) SystemMountpoints(opts ...mount.Option) (*mount.Points, error) {
 	mountpoints := mount.NewMountPoints()
 
 	for dev := range m.Targets {
-		mp, err := mount.SystemMountPointsForDevice(dev)
+		mp, err := mount.SystemMountPointsForDevice(dev, opts...)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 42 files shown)


cmd/installer/pkg/install/manifest.go
```diff
@@ -550,6 +546,10 @@ func (t *Target) Partition(pt *gpt.GPT, pos int, bd *blockdevice.BlockDevice) (e
 		return err
 	}
 
+	if err = pt.Write(); err != nil {
```
smira, line 549: why are we writing partition table here?

cmd/installer/pkg/install/verify.go
```diff
@@ -66,7 +66,7 @@ func VerifyDiskAvailability(devpath, label string) (err error) {
 
 	part, err := dev.GetPartition(label)
 	if err != nil {
-		return err
+		return nil
```
smira, line 69: this doesn't look right?

internal/integration/base/api.go
```diff
@@ -241,22 +241,27 @@ func (apiSuite *APISuite) WaitForBootDone(ctx context.Context) {
...
+							return
+						}
 					}
 				}
 			}
-		}
-	}, client.WithTailEvents(-1)))
+		}, client.WithTailEvents(-1))
 
+		return retry.ExpectedError(e)
```
smira, line 261: I don't think we should retry _every_ error here, as it probably hides some legitimate errors?

and you said that error was about context deadline exceeded, so it won't help here?

internal/pkg/encryption/keys/generated.go
```diff
@@ -0,0 +1,46 @@
...
+package keys
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/talos-systems/go-smbios/smbios"
+)
+
+// GeneratedKeyHandler generates the key based on current node information
```
smira, line 14: I think @Ulexus might have a better idea, but something like `DeterminsticKeyHandler` better describes what is going on here.

internal/pkg/encryption/keys/generated.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	machineUUID, err := s.SystemInformation().UUID()
+	if err != nil {
+		return nil, err
+	}
+
+	if machineUUID == uuid.Nil {
```
smira, line 41: we need to check for randomness of UUID here - could be a follow-up ticket, but we need to have it in 0.9

internal/pkg/mount/mount.go
```diff
@@ -257,9 +280,21 @@ func (p *Point) Mount() (err error) {
 // Unmount attempts to retry an unmount on EBUSY. It will attempt a
 // retry every 100 milliseconds over the course of 5 seconds.
 func (p *Point) Unmount() (err error) {
-	p.target = path.Join(p.Prefix, p.target)
-	if err := mountRetry(unmount, p, true); err != nil {
+	var mounted bool
+
+	if mounted, err = p.IsMounted(); err != nil {
 		return err
+	} else if mounted {
```
smira, line 287: nit: we don't need `else` here, as previous block ends with `return`

internal/pkg/mount/system.go
```diff
@@ -75,16 +81,84 @@ func SystemMountPointForLabel(device *blockdevice.BlockDevice, label string, opt
...
+	// Format the partition if it does not have any filesystem
+	preMountHooks = append(preMountHooks, func(p *Point) error {
+		sb, err := filesystem.Probe(p.source)
+		if err != nil {
+			return err
+		}
+
+		p.fstype = ""
+
+		// no filesystem detected should format it
```
smira, line 143: comment out of sync with the code?

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1051,6 +1069,45 @@ type DiskPartition struct {
...
+	//   description: >
+	//     Key slot number for luks2 encryption.
+	KeySlot int `yaml:"slot"`
+}
+
+// EncryptionKeyStatic represents throw away key type.
+type EncryptionKeyStatic struct {
+	//   description: >
+	//     Defines the static key value.
+	KeyData string `yaml:"key,omitempty"`
```
smira, line 1105: if we really need this, I guess it should be base64 encoded to allow for binary keys.

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1051,6 +1069,45 @@ type DiskPartition struct {
...
+}
+
+// EncryptionKeyGenerated represents deterministically generated key from the node UUID and PartitionLabel.
+type EncryptionKeyGenerated struct{}
```
smira, line 1109: Same as above, probably need some better name

cmd/installer/pkg/install/install.go
```diff
@@ -116,37 +116,35 @@ func (i *Installer) probeBootPartition() error {
 		dev, err := blockdevice.Open(i.options.Disk)
 		if err != nil {
 			i.bootPartitionFound = false
-
-			return err
```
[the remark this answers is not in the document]
Unix4ever: Yeah, I think missed it somehow. Thanks.

cmd/installer/pkg/install/install.go
```diff
@@ -183,7 +181,7 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 
 	// Mount the partitions.
 
-	mountpoints, err := i.manifest.SystemMountpoints()
+	mountpoints, err := i.manifest.SystemMountpoints(mount.WithFlags(mount.SkipIfNoFilesystem | mount.SkipIfEncrypted))
```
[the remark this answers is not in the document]
Unix4ever: :heavy_check_mark: yeah

cmd/installer/pkg/install/manifest.go
```diff
@@ -379,6 +377,10 @@ func (m *Manifest) preserveContents(device Device, targets []*Target) (err error
 	anyPreserveContents := false
 
 	for _, target := range targets {
+		if target.Label == constants.EphemeralPartitionLabel && target.PreserveContents {
```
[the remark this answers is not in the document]
Unix4ever: sure, that'll be more readable

internal/integration/base/api.go
Unix4ever: yeah, that's just my experiments. I will revert that back

internal/pkg/encryption/keys/generated.go
Unix4ever: Yeah, but I need help with forming randomness criteria/threshold.
We can treat uuid not random if it consists of 0 more than some ratio I guess, but I'm just not sure what level of check will be enough.
I was trying to google any libraries or methods which do that, but I couldn't find anything. Maybe I was just searching for the wrong keywords.

internal/pkg/mount/system.go
Unix4ever: yep

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
Unix4ever: I was asking for a better name on the review, but haven't got any good suggestions :)

cmd/installer/pkg/install/manifest.go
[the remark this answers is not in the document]
Unix4ever: On other hand I think this check is not going to work. I've already changed that place for the state partition encryption.
I think to drop it for now, or I can extract my change from the state partition encryption branch...

cmd/installer/pkg/install/manifest.go
Unix4ever: I think I had some idea about it, but I no longer remember :shrug: Removed

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
Unix4ever: We really need that only for tests, I would like to strongly discourage using that key type in prod.
Doesn't hurt to update it to use []byte type and decode base64 though, yeah.

cmd/installer/pkg/install/manifest.go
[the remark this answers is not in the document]
Unix4ever: I think to add the following check into `target.SaveContents`:

```go
	sb, err := source.SuperBlock()
	if err != nil {
		return err
	}

	if sb == nil || sb.Encrypted() {
		if source.Length() > partition.StateSize { // or whatever threshold we define
			return fmt.Errorf("partition %s size %s is too big to be preserved in memory", source.Name, humanize.Bytes(source.Length()))
		}

		err = t.saveRawContents(partPath)
	} else {
		err = t.saveFilesystemContents(partPath, fileSystemType, fnmatchFilters)
	}
```
This way we may get a more universal check for that case:
- if encrypted we can only preserve raw contents.
- if partition size is too big we just bail out.
- we read partition size from gpt.

What do you think?

smira (dismissed):

PR #3105, opened 2021-02-03
fix: refresh control plane endpoints on worker apids on schedule

### LINKED ISSUE #3068: apid loses connectivity to trustd if master IP changes or master goes down (workers)

smira:
This moves endpoint refresh from the context of the service `apid` in
`machined` into `apid` service itself for the workers. `apid` does
initial poll for the endpoints when it boots, but also periodically
polls for new endpoints to make sure it has accurate list of `trustd`
endpoints to talk to, this handles cases when control plane endpoints
change (e.g. rolling replace of control plane nodes with new IPs).

Related to #3069

Fixes #3068

### DIFF
--- a/internal/app/apid/main.go
+++ b/internal/app/apid/main.go
@@ -26,7 +26,10 @@ import (
 	"github.com/talos-systems/talos/pkg/startup"
 )
 
-var endpoints *string
+var (
+	endpoints       *string
+	useK8sEndpoints *bool
+)
 
 func init() {
 	// Explicitly disable memory profiling to save around 1.4MiB of memory.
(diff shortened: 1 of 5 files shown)


smira:
The downside of this PR is that apid now has access to kubelet PKI, but it's worker node PKI with limited access to K8s APIs.

internal/app/apid/main.go
```diff
@@ -26,15 +26,19 @@ import (
...
+)
 
 func init() {
 	// Explicitly disable memory profiling to save around 1.4MiB of memory.
 	runtime.MemProfileRate = 0
 
 	log.SetFlags(log.Lshortfile | log.Ldate | log.Lmicroseconds | log.Ltime)
 
-	endpoints = flag.String("endpoints", "", "the IPs of the control plane nodes")
+	endpoints = flag.String("endpoints", "", "the static list of IPs of the control plane nodes")
```
andrewrynhard, line 40: Do you know if we support supplying these manually in the config?

andrewrynhard (approved):

internal/app/apid/main.go
smira: Nope, we don't, but I thought about it and structured the code so that it's easy to add support for it

andrewrynhard: Yep, exactly. Ok cool. Thanks.

smira (commented):
.

PR #3112, opened 2021-02-06
feat: implement simple layer 2 shared IP for CP

### LINKED ISSUE #3111: Server-side shared IP for control plane
## Feature Request

As a first draft of implementing #2711, we can implement a simple layer 2 shared IP system for the Talos control plane.

### Description

We should implement a simple layer 2 shared IP system built on etcd for leadership elections.  This will allow us to use a shared "virtual" IP, known in advance, for control plane access.

Ulexus:
This adds a VIP (virtual IP) option to the network configuration of an
interface, which will allow a set of nodes to share a floating IP
address among them.  For now, this is restricted to control plane use
and only a single shared IP is supported.

Fixes #3111

# Pull Request

## Acceptance


go.mod
```diff
@@ -88,7 +88,7 @@ require (
 	golang.org/x/time v0.0.0-20200630173020-3af7569d3a1e
 	golang.org/x/tools v0.0.0-20201217163546-c88dec5c6b05 // indirect
 	golang.zx2c4.com/wireguard/wgctrl v0.0.0-20200609130330-bd2cb7843e1b
-	google.golang.org/grpc v1.29.1
```
smira, line 91: this is going to break etcd client, we shouldn't bump grpc package

internal/app/networkd/pkg/vip/vip.go
```diff
@@ -0,0 +1,136 @@
...
+	defer sess.Close() // nolint: errcheck
+
+	election := concurrency.NewElection(sess, c.etcdElectionKey())
+
+	if err := election.Campaign(ctx, hostname); err != nil {
+		return fmt.Errorf("failed to conduct campaign: %w", err)
+	}
+	defer election.Resign(ctx) // nolint: errcheck
+
+	if err := netController.AddIP(); err != nil {
```
smira, line 126: looks like we're not using any advanced features of kube-vip package here? it's just equivalent to adding/removing an address? unless I'm missing it, looks like `kube-vip` can send ARP messages, etc.?

internal/app/networkd/pkg/nic/nic.go
```diff
@@ -244,6 +248,16 @@ func (n *NetworkInterface) Configure() (err error) {
 		}
 	}
 
+	if n.VirtualIP != nil {
+		if n.vipController, err = vip.New(n.VirtualIP.String(), n.Link.Name); err != nil {
+			return fmt.Errorf("failed to create the VirtualIP controller for %q on %q: %w", n.VirtualIP, n.Link.Name, err)
+		}
+
+		if err = n.vipController.Start(context.TODO()); err != nil {
```
smira, line 256: with such context we don't have a way to shut down `vipController` gracefully, so when node goes down for a reboot, it doesn't look like etcd elections will be immediately retried to move VIP to a different node? and VIP is not technically released from the interface when networkd goes down?

Ulexus: So the election is safe:  etcd leases are immediately expired when the client connection is lost.

However, you are absolutely correct that we should shut down properly.  I just didn't see any way to get a control context in here with the current networkd architecture.  That's why the context here is a TODO.   When networkd is reworked, it will have a control context which can be shut down gracefully.

internal/app/networkd/pkg/vip/vip.go
Ulexus: Yeah, the only thing we are really using from `kube-vip` here is the ARP flooding mechanics.  We could, of course, implement the same concept ourselves, but I didn't see a need to reinvent the wheel, off-hand.

internal/app/networkd/pkg/nic/nic.go
smira: is it a problem about passing down a context? we could start around `main` function, initialize a context, cancel it on signal (service runner first sends SIGTERM) and push it through the code to reach to the parts liks this one :)

Ulexus: Yeah, exactly.

internal/app/networkd/pkg/vip/vip.go
smira: I don't see it being used here, seems just `AddIP` and `DeleteIP`, unless I'm missing something.

Ulexus: Oops.  I don't have the gratuitous ARP in there...

pkg/provision/providers/vm/vm.go
```diff
@@ -5,10 +5,15 @@
...
+
 const stateFileName = "state.yaml"
 
 // Provisioner base for VM provisioners.
 type Provisioner struct {
 	// Name actual provisioner type.
 	Name string
+
+	// VirtualIP is the virtual (shared) IP address for the control plane, if used.
+	VirtualIP net.IP
```
smira, line 18: remove me

andrewrynhard (approved):

andrewrynhard (approved):

PR #3115, opened 2021-02-06
fix: set hdmi_safe=1 on Raspberry Pi for maximum HDMI compatibility

AlekSi:
# Pull Request

## What? (description)

fix: set hdmi_safe=1 on Raspberry Pi for maximum HDMI compatibility

Setting hdmi_safe to 1 will lead to "safe mode" settings being used to try to boot with
maximum HDMI compatibility.
See https://www.raspberrypi.org/documentation/configuration/config-txt/video.md

## Why? (reasoning)

My Raspberry Pi 4 8GB does boot without it.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/board/rpi_4/rpi_4.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/board/rpi_4/rpi_4.go
@@ -6,6 +6,7 @@ package rpi4
 
 import (
 	"io/ioutil"
+	"strings"
 
 	"github.com/talos-systems/go-procfs/procfs"
 
@@ -14,11 +15,16 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
-var configTxt = []byte(`arm_64bit=1
+// https://www.raspberrypi.org/documentation/configuration/config-txt/
+var configTxt = []byte(strings.TrimSpace(`
+
+arm_64bit=1
 enable_uart=1
 kernel=u-boot.bin
 dtoverlay=disable-bt
-`)
+hdmi_safe=1
+
+` + "\n"))
 
 // RPi4 represents the Raspberry Pi 4 Model B.
 //


internal/app/machined/pkg/runtime/v1alpha1/board/rpi_4/rpi_4.go
```diff
@@ -14,11 +15,16 @@ import (
...
+var configTxt = []byte(strings.TrimSpace(`
+
+arm_64bit=1
 enable_uart=1
 kernel=u-boot.bin
 dtoverlay=disable-bt
-`)
+hdmi_safe=1
+
+` + "\n"))
```
AlekSi, line 27: That's a trick I use to separate inline data like YAML fragments from the code. Not sure if you like it or not.

AlekSi:
Please approve, but don't merge yet – I will try it one more time once CI builds it.

internal/app/machined/pkg/runtime/v1alpha1/board/rpi_4/rpi_4.go
andrewrynhard: 👍 ... I believe 1.16 is going to have a way to do something like this.

andrewrynhard (dismissed):
LGTM. Only thing I would say is that we make heavy use of `git` messages to relay information to users. We generate our `CHANGELOG.md` from `git` message, so I would suggest that the body is written in such a way that it addresses all users. It wasn't clear to me what `hdmi_safe=1` did by the commit message body. Per https://elinux.org/RPiconfig:

> hdmi_safe Use "safe mode" settings to try to boot with maximum hdmi compatibility. This is the same as the combination of: hdmi_force_hotplug=1, hdmi_ignore_edid=0xa5000080, config_hdmi_boost=4, hdmi_group=2, hdmi_mode=4, disable_overscan=0, overscan_left=24, overscan_right=24, overscan_top=24, overscan_bottom=24

I don't think we need all of that (it also wouldn't hurt), but the first sentence would be a great commit body IMO. Get's the effect of the change across to users.

smira:
My RPi does boot without monitor attached, does it allow to attach monitor after initial boot?

AlekSi:
> My RPi does boot without monitor attached, does it allow to attach monitor after initial boot?

My RPi 4 does not boot at all with the same problem: https://www.raspberrypi.org/forums/viewtopic.php?p=1547389 It starts booting the moment I insert mini-HDMI cable.

I will test again with and without this PR and report there.

LukeCarrier:
I see the same behaviour @AlekSi describes on the Pi 4B 8GB, where the system appears to only execute U-Boot once a display is connected. Adding `hdmi_safe=1` to `config.txt` on the first partition of the SD card solved the problem for me.

andrewrynhard (approved):

PR #3125, opened 2021-02-08
fix: mount kubelet secrets from system instead of ephemeral

Unix4ever:
Launch goroutine that copies kubelet pki folder contents into
`/system/secrets/kubelet` every minute before starting apid container
when running on the worker node.

Mounting kubelet secrets directly from `/var/lib/kubelet/pki` breaks
upgrade flow, because we are not able to unmount ephemeral partition,
which is being used by apid, which is not stopped during the upgrade.

### DIFF
--- a/internal/app/machined/pkg/system/services/apid.go
+++ b/internal/app/machined/pkg/system/services/apid.go
@@ -9,12 +9,15 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"log"
 	"net"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/containerd/containerd/oci"
+	"github.com/fsnotify/fsnotify"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
@@ -25,13 +28,18 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/restart"
 	"github.com/talos-systems/talos/internal/pkg/containers/image"
 	"github.com/talos-systems/talos/pkg/conditions"
+	"github.com/talos-systems/talos/pkg/copy"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
 // APID implements the Service interface. It serves as the concrete type with
 // the required methods.
-type APID struct{}
+type APID struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
 
 // ID implements the Service interface.
 func (o *APID) ID(r runtime.Runtime) string {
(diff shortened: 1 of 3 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -181,9 +181,12 @@ const (
 	// KubeletPort is the kubelet port for secure API.
 	KubeletPort = 10250
 
-	// KubeletPKIDir is the path to the directory where kubelet stores issues certificates and keys.
+	// KubeletPKIDir is the path to the directory where kubelet stores issued certificates and keys.
 	KubeletPKIDir = "/var/lib/kubelet/pki"
 
+	// SystemPKIDir is the path to the directory where Talos copies kubelet issued certificates and keys.
+	SystemPKIDir = "/system/secrets/kubelet"
```
smira, line 188: nit: should it be SystemKubeletPKIDir?

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -164,3 +179,32 @@ func (o *APID) HealthFunc(runtime.Runtime) health.Check {
 func (o *APID) HealthSettings(runtime.Runtime) *health.Settings {
 	return &health.DefaultSettings
 }
+
+func (o *APID) syncKubeletPKI() {
+	if err := copy.Dir(constants.KubeletPKIDir, constants.SystemPKIDir); err != nil {
```
smira, line 184: We might want to create explicitly a directory first say with 0700 permissions to at least partially restrict access to sensitive data

smira (approved):
Couple of nits :+1:

pkg/machinery/constants/constants.go
Unix4ever: sure, renamed

smira (dismissed):

go.mod
```diff
@@ -91,6 +91,7 @@ require (
 	google.golang.org/grpc v1.29.1
 	google.golang.org/protobuf v1.25.0
 	gopkg.in/freddierice/go-losetup.v1 v1.0.0-20170407175016-fc9adea44124
+	gopkg.in/fsnotify.v1 v1.4.7
```
smira, line 94: this should be github.com/fsnotify/fsnotify (see PR from @AlekSi  on that), we already have this import

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -164,3 +179,51 @@ func (o *APID) HealthFunc(runtime.Runtime) health.Check {
...
+func (o *APID) syncKubeletPKI() {
+	copyAll := func() {
+		if err := copy.Dir(constants.KubeletPKIDir, constants.SystemKubeletPKIDir, copy.WithMode(0o700)); err != nil {
+			log.Printf("failed to sync %s dir contents into %s: %s", constants.KubeletPKIDir, constants.SystemKubeletPKIDir, err)
+
+			return
+		}
+
+		if err := copy.File(constants.KubeletKubeconfig, filepath.Join(constants.SystemKubeletPKIDir, filepath.Base(constants.KubeletKubeconfig)), copy.WithMode(0o700)); err != nil {
+			log.Printf("failed to sync %s dir contents into %s: %s", constants.KubeletPKIDir, constants.SystemKubeletPKIDir, err)
```
smira, line 192: nit: should probably change the log message here

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -164,3 +179,51 @@ func (o *APID) HealthFunc(runtime.Runtime) health.Check {
...
+		if err != nil {
+			log.Printf("failed to create directory watcher %s", err)
+		}
+
+		err = watcher.Add(constants.KubeletPKIDir)
+		if err != nil {
+			log.Printf("failed to watch dir %s %s", constants.KubeletPKIDir, err)
+		}
+
+		defer watcher.Close() //nolint:errcheck
```
smira, line 216: should probably move this before `watcher.Add()`

smira (dismissed):
:+1: with nits

smira (approved):

andrewrynhard (approved):

PR #3134, opened 2021-02-10
fix: correctly extract wrapped error messages

smira:
In our client API, as the request goes through `apid` proxying, actual
error might be wrapped into the response (to support multi-node
requests), so it should always be correctly unwrapped.

This has UX issue: `talosctl apply-config` silently doesn't work (server
API fails, but no errror is shown in `talosctl`).


andrewrynhard (dismissed):

smira:
Need to rework this one

smira:
Dropped fix for GenerateConfiguration, API doesn't follow our apid conventions, I want to split this PR into two parts: this one is backportable to 0.8, and `GenerateConfiguration` fix is not backportable.

PR #3135, opened 2021-02-10
fix: prevent crash in machined on apid service stop

smira:
On reboots, when `apid` is stopped, `PostFunc()` is called, but
`o.cancel` is only initialized for worker nodes, so it leads to `init`
panic on master nodes.


andrewrynhard (approved):

smira:
need to rework this, looks like API structure is wrong for generate configuration, it's not apid-compliant

PR #3139, opened 2021-02-11
fix: pass attributes when adding routes

rsmitty:
This PR fixes a bug where we removed adding attributes during the
RouteAdd call for rtnetlink. The way that code is implemented means that
if *any* attributes are passed, the defaults are ignored. But we were
expecting defaults to be there. No longer!


andrewrynhard (approved):
Does this need to be back ported to 0.8?

smira:
> Does this need to be back ported to 0.8?

Added it to the backports board

smira (approved):

andrewrynhard:
> > Does this need to be back ported to 0.8?
> 
> Added it to the backports board

Thanks!

PR #3141, opened 2021-02-12
fix: don't use filename from URL when downloading manifest

smira:
Manifest URL might be anything, including any GET parameters which might
get into filename which might lead to 'filename too long' error. As each
URL is downloaded to a separate temporary directory, there's no reason
to use filename based on the URL.


Unix4ever (approved):

andrewrynhard (approved):
As far as I can tell, we don't need a unique file name, is that correct?

smira:
> As far as I can tell, we don't need a unique file name, is that correct?

Every file is downloaded to a unique temporary directory, so file name doesn't really matter

PR #3153, opened 2021-02-16
fix: find master node IPs correctly in health checks

smira:
Health checks verify node list in Kubernetes to match expectations, but
initial set of nodes for server-side health checks was driven by
`MasterIPs` functions which returns list of master endpoints which is
not exactly same as master nodes: endpoints also include some
healthchecks.

### DIFF
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -27,6 +27,7 @@ import (
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/kubernetes/kubernetes.go
```diff
@@ -152,13 +153,27 @@ func (h *Client) WorkerIPs(ctx context.Context) (addrs []string, err error) {
 	addrs = []string{}
 
 	for _, node := range resp.Items {
-		if _, ok := node.Labels[constants.LabelNodeRoleMaster]; ok {
+		_, labelMaster := node.Labels[constants.LabelNodeRoleMaster]
+		_, labelControlPlane := node.Labels[constants.LabelNodeRoleMaster]
```
Unix4ever, line 157: This looks suspicious. Won't labelMaster and labelControlPlane be always the same?
Was the idea to use some other constant?

smira: yep :facepalm:, need different label there, thanks!

Unix4ever (approved):

PR #3160, opened 2021-02-16
feat: skip filesystem for state and ephemeral partitions in the installer

Unix4ever:
Filesystem creation step is moved on the later stage: when Talos mounts
the partition for the first time.
Now it checks if the partition doesn't have any filesystem and formats
it right before mounting.

Additionally refactored mount options a bit:
- replaced separate options with a set of binary flags.
- implemented pre-mount and post-unmount hooks.

And fixed typos in couple of places and increased timeout for `apid ready`.

### DIFF
--- a/pkg/cluster/check/default.go
+++ b/pkg/cluster/check/default.go
@@ -41,7 +41,7 @@ func DefaultClusterChecks() []ClusterCheck {
 		func(cluster ClusterInfo) conditions.Condition {
 			return conditions.PollingCondition("apid to be ready", func(ctx context.Context) error {
 				return ApidReadyAssertion(ctx, cluster)
-			}, 2*time.Minute, 5*time.Second)
+			}, 5*time.Minute, 5*time.Second)
 		},
 		// wait for kubelet to be healthy on all
 		func(cluster ClusterInfo) conditions.Condition {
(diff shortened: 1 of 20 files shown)


smira (approved):

pkg/cluster/check/default.go
```diff
@@ -41,7 +41,7 @@ func DefaultClusterChecks() []ClusterCheck {
 		func(cluster ClusterInfo) conditions.Condition {
 			return conditions.PollingCondition("apid to be ready", func(ctx context.Context) error {
 				return ApidReadyAssertion(ctx, cluster)
-			}, 2*time.Minute, 5*time.Second)
+			}, 5*time.Minute, 5*time.Second)
```
smira, line 44: this is much needed fix!

tl;dr is that there's no wait for `bootkube` anymore, so we need to wait more for worker node kubelets to join and get certs

smira:
Feels like e2e-diskimage failed for real

Unix4ever:
> Feels like e2e-diskimage failed for real

yeah, but I've fixed that yesterday. Unfortunately integration tests got stuck in drone though.

smira (approved):

PR #3170, opened 2021-02-17
test: trigger e2e on thrice daily

rsmitty:
This PR adds a trigger to aws and gcp e2e tasks to make sure they get
triggered during thrice-daily.


rsmitty:
@smira please verify that this is all I need to do to make sure that these pipelines get triggered.

smira (approved):

smira:
> @smira please verify that this is all I need to do to make sure that these pipelines get triggered.

Yep, that should be it!

smira:
At this point I wonder if we even need nightly? If they are all same

Unix4ever (approved):

PR #3184, opened 2021-02-19
feat: add a warning to boot log if running self-hosted control plane

smira:
Encourage users to convert control plane to static pods after
upgrade to 0.9.


rsmitty (approved):

Unix4ever (approved):
nice!

PR #3187, opened 2021-02-19
fix: correctly escape extra args in kube-proxy manifest

smira:
JSON is a subset of YAML, so we can use JSON to escape whole YAML value
to handle any kind of symbols.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/manifest.go
+++ b/internal/app/machined/pkg/controllers/k8s/manifest.go
@@ -7,9 +7,10 @@ package k8s
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
-	"html/template"
 	"log"
+	"text/template"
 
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/k8s/manifest.go
```diff
@@ -7,9 +7,10 @@ package k8s
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
-	"html/template"
```
andrewrynhard, line 11: This always bites me.

andrewrynhard (approved):

internal/app/machined/pkg/controllers/k8s/manifest.go
smira: I hate this auto-completion thing... it's better if it _asked_ which package I mean

PR #3189, opened 2021-02-19
feat: add support for extra volume mounts for control plane pods

smira:
This allows to mount extra volumes into Talos-managed control plane
static pods. With additional options like extra files, any additional
content/configuration can be mounted.


smira:
draft, as I want to add some tests

andrewrynhard (approved):

PR #3192, opened 2021-02-20
feat: support control plane upgrades with Talos managed control plane

smira:
Upgrade is performed by updating node configuration (node by node, service
by service), watching internal resource state to get new configuration
version and verifying that pod with matching version successfully
propagated to the API server state and pod is ready.

Process is similar to the rolling update of the DaemonSet.


smira:
upgrade tests passed: https://ci.dev.talos-systems.io/talos-systems/talos/15244

andrewrynhard (approved):

PR #3211, opened 2021-02-24
feat: implement talosctl edit and patch config commands

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3209

Using parts of `kubectl` package to run the editor.
Also using the same approach as in `kubectl edit` command:
- add commented section to the top of the file with the description.
- if the config has errors, display validation errors in the commented
section at the top of the file.
- retry apply config until it succeeds.
- abort if no changes were detected or if the edited file is empty.

Patch currently supports jsonpatch only and can read it either from the
file or from the inline argument.

https://asciinema.org/a/wPawpctjoCFbJZKo2z2ATDXeC


cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+var editCmdFlags struct {
+	namespace string
+}
+
+// editCmd represents the edit command.
+var editCmd = &cobra.Command{
+	Use:     "edit <RESOURCE/NAME>",
+	Aliases: []string{"apply"},
```
smira, line 34: I feel `apply` shouldn't be alias for `edit`... 

What I thought is to provide alias for `apply-config` as `apply config` if we introduce `edit config`

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,168 @@
...
+						id = metadata.ID()
+					}
+
+					if runtime.GOOS == "windows" {
+						w = crlf.NewCRLFWriter(w)
+					}
+
+					_, err := w.Write([]byte(
+						fmt.Sprintf(
+							"# Editing %s/%s at host %s\n", resourceType, id, msg.Metadata.GetHostname(),
```
smira, line 87: `s/host/node/`, I feel we're using `node` everywhere

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,168 @@
...
+						bytes.TrimSpace(cmdutil.StripComments(body)),
+					) {
+						fmt.Println("Apply was skipped: no changes detected.")
+
+						break
+					}
+
+					_, err = c.ApplyConfiguration(parentCtx, &machine.ApplyConfigurationRequest{
+						Data:      edited,
+						Immediate: true,
```
smira, line 139: it might be amazing to support both `immediate: false` and `true` updates

if we could add a flag `--imediate` here it will be just a killer feature ;)

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/talos-systems/talos/pkg/machinery/client"
+)
+
+var editCmdFlags struct {
+	namespace string
+}
+
+// editCmd represents the edit command.
+var editCmd = &cobra.Command{
+	Use:     "edit <RESOURCE/NAME>",
```
smira, line 33: I think it should be like `get` command: `edit <type> [<id>]`

smira (commented):
looks very cool, just added some comments

cmd/talosctl/cmd/talos/edit.go
Unix4ever: ah, gotcha

cmd/talosctl/cmd/talos/edit.go
Unix4ever: sure

Unix4ever: I've added it and made it true by default. Let me know if that doesn't work

smira: can we make it `false` by default to align with `apply-config` for now?

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+
+			return nil
+		})
+	},
+}
+
+func init() {
+	editCmd.Flags().StringVar(&editCmdFlags.namespace, "namespace", "", "resource namespace (default is to use default namespace per resource)")
+	editCmd.Flags().BoolVar(&editCmdFlags.immediate, "immediate", true, "apply the change immediately (without a reboot)")
```
smira, line 168: can we please make it default to `false` to align with `apply config` so that we don't have a mismatch for now (we can later flip to be always immediate)?

cmd/talosctl/cmd/talos/patch.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil
+		})
+	},
+}
+
+func init() {
+	patchCmd.Flags().StringVar(&patchCmdFlags.namespace, "namespace", "", "resource namespace (default is to use default namespace per resource)")
+	patchCmd.Flags().StringVar(&patchCmdFlags.patchFile, "patch-file", "", "a file containing a patch to be applied to the resource.")
+	patchCmd.Flags().StringVarP(&patchCmdFlags.patch, "patch", "p", "", "the patch to be applied to the resource file.")
+	patchCmd.Flags().BoolVar(&patchCmdFlags.immediate, "immediate", true, "apply the change immediately (without a reboot)")
```
smira, line 121: same about default here...

smira (commented):
small nits

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+
+			var lastError string
+
+			editFn := func(parentCtx context.Context, msg client.ResourceResponse) error {
+				if msg.Resource == nil {
+					return nil
+				}
+
+				if msg.Metadata.GetError() != "" {
```
smira, line 63: I believe helper function skips over items which have error set?

Unix4ever: yeah, right. Forgot to remove it from this place, thanks.

smira (dismissed):

smira (approved):
:+1:

PR #3212, opened 2021-02-24
fix: correctly set service state in the resource

smira:
As 'healthy' was always set to true, some tasks started earlier than
expected, and specifically etcd cert was generated while the time sync
was happening leading to half-broken cert on RPi4.


pkg/resources/v1alpha1/service.go
```diff
@@ -71,12 +71,12 @@ func (r *Service) ResourceDefinition() core.ResourceDefinitionSpec {
 
 // SetRunning changes .spec.running.
 func (r *Service) SetRunning(running bool) {
-	r.spec.Running = true
+	r.spec.Running = running
```
smira, line 74: :man_facepalming:

andrewrynhard (approved):

PR #3220, opened 2021-02-26
feat: implement etcd remove-member cli command

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3219

We already have `etcd leave`, which make the node exclude itself from
etcd members.
But in case if the node can't remove itself because it doesn't have
conneciton to etcd we need etcd remove-member which basically removes
a node from a different node.

No unit tests for that as it's going to destroy the test cluster.


cmd/talosctl/cmd/talos/etcd.go
```diff
@@ -36,6 +36,20 @@ var etcdLeaveCmd = &cobra.Command{
 	},
 }
 
+var etcdMemberRemoveCmd = &cobra.Command{
+	Use:   "remove-member <hostname>",
+	Short: "Remove the node from etcd cluster",
+	Long:  ``,
```
smira, line 42: I feel we should add more description here saying that this might destroy etcd cluster if used for a running member, etc.

And say something like the usecase for this command is recovering from the scenario when etcd member is already in a broken state (e.g. hardware failure), and we need to remove that member without having access to the actual node.

Unix4ever: ok, sure

api/machine/machine.proto
```diff
@@ -736,6 +738,12 @@ message EtcdLeaveClusterRequest {}
 message EtcdLeaveCluster { common.Metadata metadata = 1; }
 message EtcdLeaveClusterResponse { repeated EtcdLeaveCluster messages = 1; }
 
+message EtcdRemoveMemberRequest {
+  string hostname = 2;
```
andrewrynhard, line 742: In other `Etcd*` RPCs, we use `member` instead of hostname. Not sure I care either way too much, but wanted to point it out so that we can all weigh in.

andrewrynhard (approved):
LGTM. Just left one comment that we should get other people's opinion on.

api/machine/machine.proto
smira: I agree, `member` is more generic, it doesn't have to be always hostname.

Unix4ever: rgr, will update

smira (approved):

PR #3222, opened 2021-02-26
fix: make ApplyDynamicConfig idempotent

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3221

Detect defined SANs and append only non-overlapping ones.


andrewrynhard (dismissed):

rsmitty (approved):

smira (approved):

smira:
merging this to get the VIP PR on top of it, as it also touches dynamic config

PR #3233, opened 2021-03-01
fix: update in-cluster kubeconfig validity to match other certs

smira:
Talos generates in-cluster kubeconfig for the kube-scheduler and
kube-controller-manager to authenticate to kube-apiserver. Bug was that
validity of that kubeconfig was set to 24h by mistake. Fix that by
bumping validity to default for other Kubernetes certs (1 year).

Add a certificate refresh at 50% of the validity.

Fix bugs with copying secret resources which was leading to updates not
being propagated correctly.

### DIFF
--- a/internal/app/machined/pkg/controllers/secrets/kubernetes.go
+++ b/internal/app/machined/pkg/controllers/secrets/kubernetes.go
@@ -26,6 +26,11 @@ import (
 	"github.com/talos-systems/talos/pkg/resources/v1alpha1"
 )
 
+// KubernetesCertificateValidityDuration is the validity duration for the certificates created with this controller.
+//
+// Controller automatically refreshes certs at 50% of CertificateValidityDuration.
+const KubernetesCertificateValidityDuration = constants.KubernetesDefaultCertificateValidityDuration
+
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/controllers/secrets/kubernetes.go
```diff
@@ -26,6 +26,11 @@ import (
 	"github.com/talos-systems/talos/pkg/resources/v1alpha1"
 )
 
+// KubernetesCertificateValidityDuration is the validity duration for the certificates created with this controller.
+//
+// Controller automatically refreshes certs at 50% of CertificateValidityDuration.
+const KubernetesCertificateValidityDuration = constants.KubernetesDefaultCertificateValidityDuration
```
andrewrynhard, line 32: I'm curious, why not just use the constant?

andrewrynhard (dismissed):

internal/app/machined/pkg/controllers/secrets/kubernetes.go
smira: I wanted to keep it local to this file as it's being used four times, to make sure it's in sync across all four usages. But no real reason besides that.

PR #3237, opened 2021-03-02
fix: move etcd to `cri` containerd runner

### LINKED ISSUE #3229: Rebooting Talos in offline mode is not working
## Bug Report
Rebooting a talos node in an air-gapped environment is not working as there is a try to pull Etcd image.

### Description

In reboot of talos node without internet connectivity the reboot hangs. Once restoring the internet connectivity reboot continues. It should be possible to run Talos in air-gapped without local registry. Without internet the reboot is stuck with  `failed to pull image`

Etcd service:
```
  80   // Pull the image and unpack it.
  81   containerdctx := namespaces.WithNamespace(ctx, constants.SystemContainerdNamespace)
  82   if _, err = image.Pull(containerdctx, r.Config().Machine().Registries(), client, r.Config().Cluster().Etcd().Image()); err != nil {
  83     return fmt.Errorf("failed to pull image %q: %w", r.Config().Cluster().Etcd().Image(), err)
  84   }
  85
```

### Environment

- Talos version: 0.8.1
- Kubernetes version: v1.19.1
- Platform: Bare-metal

smira:
This fixes a problem when Talos pulls `etcd` image one every reboot, as
`etcd` was running in the system containerd which is completely
ephemeral (backed by `tmpfs`).

Also skip pulling if image is already present and unpacked (same fix for
the `kubelet` image).

Fixes #3229

### DIFF
--- a/internal/app/machined/pkg/system/services/kubelet.go
+++ b/internal/app/machined/pkg/system/services/kubelet.go
@@ -19,7 +19,6 @@ import (
 	containerdapi "github.com/containerd/containerd"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/oci"
-	criconstants "github.com/containerd/cri/pkg/constants"
 	cni "github.com/containerd/go-cni"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -65,6 +64,8 @@ func (k *Kubelet) ID(r runtime.Runtime) string {
 }
 
 // PreFunc implements the Service interface.
+//
+//nolint: gocyclo
 func (k *Kubelet) PreFunc(ctx context.Context, r runtime.Runtime) error {
 	cfg := struct {
 		Server               string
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -112,9 +114,25 @@ func (k *Kubelet) PreFunc(ctx context.Context, r runtime.Runtime) error {
...
+		if err != nil || !unpacked {
+			pullRequired = true
+		}
+	}
+
+	if pullRequired {
+		_, err = image.Pull(containerdctx, r.Config().Machine().Registries(), client, r.Config().Machine().Kubelet().Image())
+		if err != nil {
+			return err
+		}
```
andrewrynhard, line 135: Worth putting this in a common place? Like our `containerd` package.

andrewrynhard (approved):
I wonder if we should have etcd and kubelet be in the `system` namespace (but still in the CRI containerd). Makes for a cleaner separation. Also, I left a question. I know we need to get just get _something_ so don't let me shoe you down.

internal/app/machined/pkg/system/services/kubelet.go
smira: yep, I think that makes sense. probably even `image.Pull` function itself with an option?

smira:
> I wonder if we should have etcd and kubelet be in the `system` namespace (but still in the CRI containerd). Makes for a cleaner separation. Also, I left a question. I know we need to get just get _something_ so don't let me shoe you down.

yeah, we could... `talosctl containers -k` now doesn't see these, as they're not launched via CRI, but we could add even more separation here

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Makes sense.

smira:
e

> > I wonder if we should have etcd and kubelet be in the `system` namespace (but still in the CRI containerd). Makes for a cleaner separation. Also, I left a question. I know we need to get just get _something_ so don't let me shoe you down.
> 
> yeah, we could... `talosctl containers -k` now doesn't see these, as they're not launched via CRI, but we could add even more separation here

moved to system namespace and refactored to reduce duplication

andrewrynhard (approved):

PR #3243, opened 2021-03-02
chore: properly propagate context object in the controller

Unix4ever:
This is required to correctly handle ACPI reboot or forceful reboots
during sequence that locks the controller.
Additionally fix `NoSchedule` untaint when the configuration is changed.

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -149,7 +149,7 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 		}
 
 		go func() {
-			if err := s.Controller.Run(runtime.SequenceApplyConfiguration, in); err != nil {
+			if err := s.Controller.Run(context.Background(), runtime.SequenceApplyConfiguration, in); err != nil {
 				if !runtime.IsRebootError(err) {
 					log.Println("apply configuration failed:", err)
 				}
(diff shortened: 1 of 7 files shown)


andrewrynhard (dismissed):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -263,21 +263,27 @@ func (c *Controller) run(seq runtime.Sequence, phases []runtime.Phase, data inte
...
 			if !runtime.IsRebootError(err) {
 				log.Printf("phase %s (%s): failed", phase.Name, progress)
 			}
 
 			return fmt.Errorf("error running phase %d in %s sequence: %w", number, seq.String(), err)
 		}
 
+		select {
+		case <-ctx.Done():
+			return context.Canceled
```
smira, line 276: better to return `ctx.Err()`

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -149,7 +149,7 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 		}
 
 		go func() {
-			if err := s.Controller.Run(runtime.SequenceApplyConfiguration, in); err != nil {
+			if err := s.Controller.Run(ctx, runtime.SequenceApplyConfiguration, in); err != nil {
```
smira, line 152: we shouldn't use `grpc` request context for it, as it will abort sequence when request is done, and it's done almost immediately. this should use `context.Background()` (same in other places when we launch sequence in a goroutine from grpc context)

Unix4ever: yeah, I got hit by that already. Fixing :)

smira (approved):

andrewrynhard (approved):

PR #3247, opened 2021-03-03
feat: bypass lock if ACPI reboot/shutdown issued

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/2997

Listen for restart events in parallel with the boot sequence and cancel
the context if got `RestartEvent`.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -231,7 +231,7 @@ func run() error {
 	system.Services(c.Runtime()).LoadAndStart(&services.Machined{Controller: c})
 
 	// Boot the machine.
-	if err = c.Run(ctx, runtime.SequenceBoot, nil); err != nil {
+	if err = c.Run(ctx, runtime.SequenceBoot, nil); err != nil && !errors.Is(err, context.Canceled) {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


internal/app/machined/main.go
```diff
@@ -196,7 +197,16 @@ func run() error {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	cancelContext := func() {
```
smira, line 200: calling `cancel()` multiple times should be fine, so not sure if we nee this?

internal/app/machined/main.go
```diff
@@ -256,7 +268,18 @@ func run() error {
...
-	return <-errCh
+	var result *multierror.Error
+
+	result = multierror.Append(result, <-errCh)
+
+	cancelContext()
+
+	err = <-bootCh
+	if err != nil {
+		result = multierror.Append(result, <-errCh)
```
smira, line 279: receiving from `errCh` two times?

Unix4ever: that was just a dumb error

smira (dismissed):
lgtm, I hope it works as expected, as it was more of thought experiment :)

andrewrynhard (dismissed):

Unix4ever:
> lgtm, I hope it works as expected, as it was more of thought experiment :)

Yeah, works both using ACPI and during a normal talosctl reboot/shutdown calls.

smira:
combined list of integration tests passed

PR #3288, opened 2021-03-11
chore: update golanci-lint to 1.38.0

Unix4ever:
Fix all discovered issues.
Detected couple bugs, fixed them as well.

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -223,7 +223,7 @@ func (s *Server) Reboot(ctx context.Context, in *empty.Empty) (reply *machine.Re
 
 // Rollback implements the machine.MachineServer interface.
 //
-//nolint:dupl,gocyclo
+//nolint:gocyclo
 func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*machine.RollbackResponse, error) {
 	log.Printf("rollback via API received")
 
@@ -286,8 +286,6 @@ func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*ma
 }
 
 // Bootstrap implements the machine.MachineServer interface.
-//
-//nolint:dupl
 func (s *Server) Bootstrap(ctx context.Context, in *machine.BootstrapRequest) (reply *machine.BootstrapResponse, err error) {
 	log.Printf("bootstrap request received")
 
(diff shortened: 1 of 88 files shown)


internal/app/machined/pkg/controllers/k8s/kubelet_static_pod_controller.go
```diff
@@ -221,7 +221,7 @@ func (ctrl *KubeletStaticPodController) writePod(ctx context.Context, r controll
 
 	renderedPod, err := yaml.Marshal(staticPod.Spec())
 	if err != nil {
-		return nil
+		return err
```
Unix4ever, line 224: The first bug.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -144,18 +140,14 @@ func MountBPFFS(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
...
-		return nil
+		return mount.Mount(mountpoints)
 	}, "mountBPFFS"
 }
 
 const (
 	memoryCgroup                  = "memory"
 	memoryUseHierarchy            = "memory.use_hierarchy"
-	memoryUseHierarchyPermissions = os.FileMode(400)
+	memoryUseHierarchyPermissions = os.FileMode(0o400)
```
Unix4ever, line 150: The second bug.

internal/app/machined/pkg/system/system.go
```diff
@@ -18,7 +18,8 @@ import (
 	"github.com/talos-systems/talos/pkg/conditions"
 )
 
-type singleton struct {
```
smira, line 21: Which linter enforces this change? Keeping this type private was supposed to be the way to enforce single instance of this type

smira (commented):
:+1: just one question

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -778,6 +774,7 @@ func (s *Server) List(req *machine.ListRequest, obj machine.MachineService_ListS
 }
 
 // DiskUsage implements the machine.MachineServer interface.
+// nolint:cyclop
```
AlekSi, line 777:
```suggestion
//nolint:cyclop
```

internal/app/machined/pkg/system/system.go
Unix4ever: let me check

Unix4ever: that's golint. I guess it makes sense to set `nolint:golint` there instead.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
Unix4ever: we also need a linter for that whitespace check :D That's a reflex...

smira (approved):

smira (approved):

PR #3291, opened 2021-03-12
chore: use Go 1.16 language version

AlekSi:
# Pull Request

## What? (description)

chore: use Go 1.16 language version

## Why? (reasoning)

It affects some languages features and go subcommands.
https://golang.org/ref/mod#go-mod-file-go

## Acceptance

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -126,6 +126,9 @@ COPY --from=generate-build /pkg/machinery/config/types/v1alpha1/*_doc.go /pkg/ma
 FROM build AS base
 COPY ./go.mod ./go.sum ./
 COPY ./pkg/machinery/go.mod ./pkg/machinery/go.sum ./pkg/machinery/
+WORKDIR /src/pkg/machinery
+RUN go mod download
+WORKDIR /src
 RUN go mod download
 RUN go mod verify
 COPY ./cmd ./cmd
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/talos-systems/talos
 
-go 1.13
+go 1.16
 
 replace (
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/talos-systems/talos
 
-go 1.13
+go 1.16
```
smira, line 3: one thing which is not clear for me from the docs -- does it allow module to be imported with Go 1.15 now?

this is mostly important for machinery package which provides API to access the cluster or generate configs

AlekSi: Yes. Currently, the code will still be imported and compiled, and the error message will include a Go version mismatch only if the code will fail to compile.

smira (approved):

Dockerfile
```diff
@@ -126,6 +126,9 @@ COPY --from=generate-build /pkg/machinery/config/types/v1alpha1/*_doc.go /pkg/ma
 FROM build AS base
 COPY ./go.mod ./go.sum ./
 COPY ./pkg/machinery/go.mod ./pkg/machinery/go.sum ./pkg/machinery/
+WORKDIR /src/pkg/machinery
+RUN go mod download
```
AlekSi, line 130: From https://golang.org/doc/go1.16:

> When the main module's go.mod file declares go 1.16 or higher, the all package pattern now matches only those packages that are transitively imported by a package or test found in the main module. (Packages imported by tests of packages imported by the main module are no longer included.) This is the same set of packages retained by go mod vendor since Go 1.11.

Without explicit `go mod download`, `go mod tidy` below downloads machinery's test dependencies, and `grep` fails:
```
#165 0.135 go: downloading github.com/google/go-cmp v0.5.0
#165 0.352 go: downloading golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9
#165 0.352 go: downloading github.com/onsi/gomega v1.10.1
```

PR #3305, opened 2021-03-15
fix: don't touch any partitions on upgrade with --preserve

smira:
This fixes a case of upgrade from 0.9.0-alpha.4 to 0.9.0-beta.0. With
introduced proper partition alignment and physical block size != 512,
partitions before ephemeral will be moved around a bit (due to the
alignment), and `STATE` partition size might change a bit.

If encryption is enabled, contents are preserved as raw bytes, so
partition size should be exactly same during restore.

Drop code (mostly tests) which handled 0.6 to 0.7 upgrades.

On upgrade with preserve don't touch any partitions, at least for 0.8 ->
0.9 layout hasn't changed.

### DIFF
--- a/cmd/installer/pkg/install/manifest.go
+++ b/cmd/installer/pkg/install/manifest.go
@@ -14,10 +14,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/dustin/go-humanize"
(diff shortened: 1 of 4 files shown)


andrewrynhard (approved):

smira:
If encryption is not enabled, issue is still there, but the fix still applies.

PR #3310, opened 2021-03-15
fix: properly format spec comments in the resources

Unix4ever:
`os-runtime` now writes `yaml` block as raw yaml bytes instead of
decoding it into `yaml.Node` and encoding that `yaml.Node` back to YAML.
The reason is that `go-yaml` comments decoder can't really handle
comment placement properly, so it messes up indents here and there.

os-runtime PR: https://github.com/talos-systems/os-runtime/pull/14
yaml fork commit: https://github.com/Unix4ever/yaml/commit/2d0f33ba90d0e2be7e036f10f3fab208ac394710


smira (dismissed):

andrewrynhard (dismissed):

Unix4ever:
I messed up with yaml version, fixing now.

PR #3313, opened 2021-03-16
chore: use new release tool for changelogs and release notes

smira:
This replaces `git-chglog` with the `talos-systems/release-tool`.

We need to make sure we update `hack/release.toml` as we do PRs to make
sure we highlight any important feature or update as we go.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -2,11 +2,16 @@
 
 set -e
 
+RELEASE_TOOL_IMAGE="ghcr.io/talos-systems/release-tool:latest"
+
+function release-tool {
+  docker pull "${RELEASE_TOOL_IMAGE}" >/dev/null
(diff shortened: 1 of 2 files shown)


smira:
Sample: https://gist.github.com/smira/acdff21e432876605bf5ba88546d851a

Unix4ever (dismissed):

hack/release.sh
```diff
@@ -2,21 +2,26 @@
 
 set -e
 
+RELEASE_TOOL_IMAGE="ghcr.io/talos-systems/release-tool:latest"
+
+function release-tool {
+  docker pull "${RELEASE_TOOL_IMAGE}" >/dev/null
```
AlekSi, line 8: Why not `-q` / `--quiet`?

AlekSi (approved):

hack/release.sh
smira: :+1: I'll get it updated

smira: nope, it doesn't work, still prints the image name, so I'll keep it this way. It still prints the errors though, so it's harmless

PR #3321, opened 2021-03-17
fix: repair zsh completion

### LINKED ISSUE #3318: zsh completions are not working
## Bug Report

It has been reported that `zsh` completions for `talosctl` are not working.  This needs to be tested and refreshed.  Some of the comments, for instance, still reference the old `osctl`, at a minimum.

Also, we might take this opportunity to add `fish` completions...

### Environment

- Platform:

jonkerj:
# Pull Request

## What? (description)
ZSH completion was broken in `talosctl`. Cobra does not hook the ZSH generated completion rules, as it appears. Tools with working ZSH cobra completion (such as `helm` and `kubectl`) do so by printing the hook (`compdef _<completion> <tool>`) themselves.

Fixes #3318

## Why? (reasoning)
To get shell completion working for ZSH users (they are the best users!)

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/completion.go
+++ b/cmd/talosctl/cmd/completion.go
@@ -60,7 +60,11 @@ talosctl completion zsh > "${fpath[1]}/_talosctl"`,
 		case "bash":
 			return rootCmd.GenBashCompletion(os.Stdout)
 		case "zsh":
-			return rootCmd.GenZshCompletion(os.Stdout)
+			err := rootCmd.GenZshCompletion(os.Stdout)
+			// cobra does not hook the completion, so let's do it manually
+			fmt.Printf("compdef _talosctl talosctl")
+
+			return err
 		default:
 			return fmt.Errorf("unsupported shell %q", args[0])
 		}


jonkerj:
It seems a matter of taste, consistency, as it turns out. Kubectl and helm include the hook in their `<tool> completion bash`, whereas flux [tells their users to do it themselves](https://github.com/fluxcd/flux2/blob/main/cmd/flux/completion_zsh.go#L28-L35). 

Not sure what's preferable for Talos, but I prefer the style from kubectl/helm, since I can load/update the completions in oh-my-zsh using a single command:

```
helm completion zsh > ~/.oh-my-zsh/completions/_helm
kubectl completion zsh > ~/.oh-my-zsh/completions/_kubectl
(flux completion zsh && echo "compdef _flux flux") > ~/.oh-my-zsh/completions/_kubectl
```

smira:
looks like cobra did a lot of work on shell completion https://github.com/spf13/cobra/releases 

I wonder if it's fixed upstream

jonkerj:
I don't think this is considered a bug at cobra, as they output the commented compdef. I checked, and the code still seems to do so in v1.1.3 (latest release).

andrewrynhard:
We think that maybe #3331 fixes this. @jonkerj would you be able to give `master` `talosctl` a try?

jonkerj:
I tried, but the problem is still there. The problem is that the completion is not "hooked" through `compdef`. My PR takes care of that.

andrewrynhard (approved):

PR #3325, opened 2021-03-18
feat: publish talosctl container image

bzub:
Creates a new container image and corresponding Makefile target.

Closes https://github.com/talos-systems/talos/issues/3324

# Pull Request

## Acceptance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ COMMON_ARGS += --build-arg=https_proxy=$(https_proxy)
 
 CI_ARGS ?=
 
-all: initramfs kernel installer talosctl talos
+all: initramfs kernel installer talosctl talosctl-image talos
 
 # Help Menu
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ COMMON_ARGS += --build-arg=https_proxy=$(https_proxy)
 
 CI_ARGS ?=
 
-all: initramfs kernel installer talosctl talos
+all: initramfs kernel installer talosctl talosctl-image talos
```
smira, line 58: we need to add `talosctl-image` to the `push` target to make CI actually push the image

Makefile
```diff
@@ -155,6 +155,10 @@ installer: ## Builds the container image for the installer and outputs it to the
 talos: ## Builds the Talos container image and outputs it to the artifact directory.
 	@$(MAKE) registry-$@ TARGET_ARGS="--allow security.insecure"
 
+.PHONY: talosctl-image
+talosctl-image: ## Builds the talosctl container image and outputs it to the artifact directory.
+	@$(MAKE) registry-talosctl TARGET_ARGS="--allow security.insecure"
```
smira, line 160: a bit of a question, but we have talosctl built for more arches than Talos itself

the way it is right now, it should create talosctl image for amd64/arm64, but we certainly could publish more, and that requires some bigger changes, so I would rather skip it

smira (commented):
some small issues, lgtm otherwise!

Makefile
rsmitty: I was thinking we'd definitely want it to be multi-arch so that there's an ARM image as well. What's involved in that?

Makefile
bzub: Done.

Makefile
bzub: Supporting `amd64` and `arm64`, as it does now, seems like it should cover most container hosts. It's perfect for my MacBook + Docker Desktop for example.

rsmitty: Oh, maybe I misunderstood that arm64 was already supported. Agreed that those two should be sufficient for most cases. I'm going to approve and allow this to run through CI. @smira can `/lgtm` it if he's good with everything in the AM.

smira (approved):

PR #3328, opened 2021-03-19
feat: update Kubernetes to 1.21.0-beta.1

AlekSi:
See CHANGELOG:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md


internal/integration/provision/upgrade.go
```diff
@@ -70,48 +70,49 @@ type upgradeSpec struct {
 }
 
 const (
-	previousRelease = "v0.8.4"
-	stableRelease   = "v0.9.0-alpha.5"
+	stableRelease = "v0.8.5"
+	devRelease    = "v0.9.0-beta.2"
```
smira, line 74: could we please keep it the way it is, i.e. `stableRelease`?

we'll update v0.9.0-beta.2 to 0.9.0 once we have it released, but it will stay as stable

website/content/docs/v0.9/Reference/cli.md
```diff
@@ -103,7 +103,7 @@ talosctl cluster create [flags]
       --ipv4                                    enable IPv4 network in the cluster (default true)
       --ipv6                                    enable IPv6 network in the cluster (QEMU provisioner only)
       --iso-path string                         the ISO path to use for the initial boot (VM only)
-      --kubernetes-version string               desired kubernetes version to run (default "1.20.5")
+      --kubernetes-version string               desired kubernetes version to run (default "1.21.0-beta.1")
```
smira, line 106: hell... we might need to hold on with this PR before we fork 0.10 docs. probably early next week, this change should go to 0.10 docs, but we don't have them yet until we actually do release 0.9.0

internal/integration/provision/upgrade.go
AlekSi: It is either wrong now (alpha and beta releases are not stable, and that's very confusing) or later.

Once 0.9.0 is out, will we support 0.8.x?

smira: bumping version is good of course, my comment was only about naming

AlekSi: I don't particularly like the word "stable" there because alphas and betas are not stable.
I would call it "current", but that word is already used for the version being built.
"next" is not good too.
What about "latest"?

AlekSi: Ended up with `stableRelease` and a comment

smira:
let's hold on merging this till Monday, in case we have any more docs for 0.9 or backports requiring new docs and we can move to 0.10 docs in master and rebase this one

internal/integration/provision/upgrade.go
```diff
@@ -70,18 +70,19 @@ type upgradeSpec struct {
 }
 
 const (
-	previousRelease = "v0.8.4"
-	stableRelease   = "v0.9.0-alpha.5"
+	previousRelease = "v0.8.5"
```
smira, line 73: need to update versions in `Makefile` `RELEASES` variable to match these (so that required artifacts are downloaded)

smira (dismissed):

hack/test/e2e.sh
```diff
@@ -22,12 +22,10 @@ mkdir -p "${TMP}"
 # Talos
 
 export TALOSCONFIG="${TMP}/talosconfig"
-export TALOS_VERSION=v0.9
 
 # Kubernetes
 
 export KUBECONFIG="${TMP}/kubeconfig"
-export K8S_VERSION=1.20.5
```
AlekSi, line 30: `TALOS_VERSION` and `K8S_VERSION` are not referenced anywhere. Do we need them?

smira: They are referenced from the cluster template which is downloaded from another repository

smira (approved):

smira (approved):

PR #3335, opened 2021-03-21
feat: add ability to find disk by disk properties

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3323

Not exactly matching with udevd generated `by-<id>` symlinks, but should
provide sufficient amount of property selectors to be able to pick
specific disks for any kind of disk: sd card, hdd, ssd, nvme.


pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -1204,7 +1205,47 @@ func (i *InstallConfig) Image() string {
...
+
+		if selector.Serial != "" {
+			matchers = append(matchers, disk.WithSerial(selector.Serial))
+		}
+
+		if selector.Modalias != "" {
+			matchers = append(matchers, disk.WithModalias(selector.Modalias))
+		}
+
+		disk, err := disk.Find(matchers...)
```
smira, line 1242: I wonder if this code should be outside of the config provider really?

Like we have method like `DiskMatchers() []diskMatcher` which returns matchers, if that list is not empty, we prefer it over `InstallDisk`, but the actual `disk.Find()` call is somewhere in the sequencer (tasks?), so we could print error message saying if matching failed, etc.?

smira (commented):
looks really cool, I think we should add a note to `hack/release.toml` to highlight this feature in 0.10

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
Unix4ever: Yep, let's change that.

Unix4ever: Ah wait, I remember why I moved that here.
We also have validation module, which get the disk using `Disk` method.

Still can split it though. Should be a common method for `v1alpha1_validation.go` and the sequener.

hack/release.toml
```diff
@@ -24,8 +24,25 @@ preface = """\
 
     [notes.optimize]
         title = "Optmizations"
-        descriptions = """\
+        description = """\
```
Unix4ever, line 27: that looked like a typo, so I've changed that during the rebase.

andrewrynhard:
@DWSR This should fix you up for choosing an install disk early on by WWID/UUID early on without `udevd` running.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -246,8 +246,13 @@ func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*ma
 			}
 		}()
 
+		disk, err := s.Controller.Runtime().Config().Machine().Install().Disk()
+		if err != nil {
+			return err
+		}
+
 		grub := &grub.Grub{
-			BootDisk: s.Controller.Runtime().Config().Machine().Install().Disk(),
```
smira, line 250: feels like this was wrong from the beginning - it should be using the disk Talos is installed to, and we had that method somewhere in runtime/state?

smira (commented):
probably I got confused, but does `disks` work now in non-maintenance mode? 

if not, should we hook it up to the v1alpha1_server?

Unix4ever:
> probably I got confused, but does disks work now in non-maintenance mode?
if not, should we hook it up to the v1alpha1_server?

It doesn't work in non-maintenance mode. So you mean we should add support for that in non-maintenance mode?

smira:
> > probably I got confused, but does disks work now in non-maintenance mode?
> > if not, should we hook it up to the v1alpha1_server?
> 
> It doesn't work in non-maintenance mode. So you mean we should add support for that in non-maintenance mode?

I think it will be really useful, it should be really easy, I can show that real quick, it's just a single line of code.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
Unix4ever: should be fixed now

smira (dismissed):

PR #3350, opened 2021-03-22
refactor: run networkd as a goroutine in machined

smira:
This removes machined as a separate container and image.

Reasons:

* `machined` becomes more and more bound into the core flow - now it
interacts with `etcd` for VIPs, so container has more and more
mounts/permissions
* it should be easier to COSIfy machined piece by piece if we have it
running in the same process
* initramfs size

### DIFF
--- a/internal/app/networkd/main.go
+++ b/internal/app/networkd/main.go
@@ -2,66 +2,37 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package main
+package networkd
 
 import (
 	"context"
+	"io"
 	"log"
-	"os"
-	"os/signal"
-	"runtime"
-	"syscall"
 
 	"golang.org/x/sync/errgroup"
 
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/internal/app/networkd/pkg/networkd"
 	"github.com/talos-systems/talos/internal/app/networkd/pkg/reg"
 	"github.com/talos-systems/talos/pkg/grpc/factory"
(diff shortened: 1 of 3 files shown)


Unix4ever (dismissed):

AlekSi:
> This removes machined as a separate container and image.

`networkd`, not `machined`, right?

internal/app/networkd/main.go
```diff
@@ -2,66 +2,37 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package main
+package networkd
```
AlekSi, line 5: Should we rename this file now that it does not contain the "normal" `main()` function?

Should we also move this file away from `internal/app/` according to the project layout document?

AlekSi: Should we also update components documentation?

smira:
> > This removes machined as a separate container and image.
> 
> `networkd`, not `machined`, right?

Yes, of course :facepalm:

internal/app/networkd/main.go
smira: > Should we rename this file now that it does not contain the "normal" `main()` function?

I would rather keep it in `main.go` for now, as it's still an entrypoint to the service.

smira: > Should we also update components documentation?

From the user point of view, nothing has changed, as still service `networkd` exists, it's visible in `talosctl services`, `talosctl logs networkd` works, etc. It just changes the way we run it, but it should be transparent to the user.

smira: > Should we also move this file away from `internal/app/` according to the project layout document?

same answer, nothing has changed so far, it's still a service and kind of same service. 

networkd itself is supposed to be removed any time soon once we make it COSIfied, and then this directory will go away and docs will be updated

PR #3351, opened 2021-03-22
refactor: move apid, routerd, timed and trustd to single executable

smira:
This removes container images for the aforementioned services, they are
now built into `machined` executable which launches one or another
service based on `argv[0]`.

Containers are started with rootfs directory which contains only a
single executable file for the service.

This creates rootfs on squashfs for each container in
`/opt/<container>`.

Service `networkd` is not touched as it's handled in #3350.

This removes all the image imports, snapshots and other things which
were associated with the existing way to run containers.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -22,6 +22,12 @@ preface = """\
 * added support for Rock Pi 4.
 """
 
+    [notes.optimize]
+        title = "Optmizations"
+        descriptions = """\
+* Talos `system` services now run without container images on initramfs from the single executable; this change reduces RAM usage, initramfs size and boot time..
+"""
+
 [make_deps]
 
     [make_deps.tools]
--- a/internal/app/machined/pkg/system/services/apid.go
+++ b/internal/app/machined/pkg/system/services/apid.go
@@ -26,7 +26,6 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/containerd"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/runner/restart"
-	"github.com/talos-systems/talos/internal/pkg/containers/image"
 	"github.com/talos-systems/talos/pkg/conditions"
 	"github.com/talos-systems/talos/pkg/copy"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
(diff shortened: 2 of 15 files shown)


andrewrynhard (dismissed):

hack/release.toml
```diff
@@ -20,6 +20,12 @@ preface = """\
         description = """\
 * u-boot version was updated to fix the boot and USB issues on Raspberry Pi 4 8GiB version.
 * added support for Rock Pi 4.
+"""
+
+    [notes.optimize]
+        title = "Optmizations"
+        descriptions = """\
+* Talos sytems services now run without container images on initramfs from the single executable; this change reduces RAM usage, initramfs size and boot time.
```
andrewrynhard, line 28:
```suggestion
* Talos `system` services now run without container images on initramfs from the single executable; this change reduces RAM usage, initramfs size and boot time.
```

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -153,11 +150,12 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 		runner.WithStdin(stdin),
 		runner.WithLoggingManager(r.Logging()),
 		runner.WithContainerdAddress(constants.SystemContainerdAddress),
-		runner.WithContainerImage(image),
 		runner.WithEnv(env),
 		runner.WithOCISpecOpts(
 			oci.WithHostNamespace(specs.NetworkNamespace),
 			oci.WithMounts(mounts),
+			oci.WithRootFSPath("/opt/apid"),
```
andrewrynhard, line 157: Is this really required? `WithRootFSReadonly` should make the standard root path of the container read only.

andrewrynhard (approved):

internal/app/machined/pkg/system/services/apid.go
smira: we need to have binary available in the rootfs itself? it can't run from empty directory

andrewrynhard: IIUC `containerd` should create an empty dir for it somewhere in `/system`, then we just bind mount the binary into it. Kind of like a `docker run -v /bin/binary:/bin/binary --entrypoint=/bin/binary alpine`.

smira: we figured out this in a conversation, tl;dr is that we don't have anything in `/system` for this contianer

in the future we might even drop system containerd and use `runc` directly

andrewrynhard (approved):

PR #3360, opened 2021-03-24
feat: replace timed with time sync controller

smira:
This is a complete rewrite of time sync process.

Now the time sync process starts early at boot time, and it adapts to
configuration changes:

* before config is available, `pool.ntp.org` is used
* once config is available, configured time servers are used

Controller updates same time sync resource as other controllers had
dependency on, so they have a chance to wait for the time sync event.

Talos services which depend on time now wait on same resource instead of
waiting on timed health.

New features:

* time sync now sticks to the particular time server unless there's an
error from that server, and server is changed in that case, this
improves time sync accuracy

* time sync acts on config changes immediately, so it's possible to
reconfigure time sync at any time

* there's a new 'epoch' field in time sync resources which allows
time-dependent controllers to regenerate certs when there's a big enough
jump in time

Features to implement later:

* apid shouldn't depend on timed, it should be started early and it
should regenerate certs on time jump

* trustd should be updated in same way


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1705,27 +1656,41 @@ func BootstrapEtcd(seq runtime.Sequence, data interface{}) (runtime.TaskExecutio
 			}
 		}
 
-		// Since etcd has already attempted to start, we must delete the data. If
-		// we don't, then an initial cluster state of "new" will fail.
-		dir, err := os.Open(constants.EtcdDataPath)
-		if err != nil {
-			return err
-		}
+		if err = func() error {
```
smira, line 1659: this fixes a bug in `etcd` bootstrap: if `etcd` never started before `talosctl bootstrap` is issued, etcd data directory might not exist yet

Dockerfile
```diff
@@ -176,43 +176,6 @@ RUN chmod +x /machined
...
-ARG PKGS
-ARG EXTRAS
-ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
-WORKDIR /src/internal/app/timed
-RUN --mount=type=cache,target=/.cache/go-build go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Server -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${VERSION_PKG}.PkgsVersion=${PKGS} -X ${VERSION_PKG}.ExtrasVersion=${EXTRAS}" -o /timed
-RUN chmod +x /timed
-
-# The apid target builds the apid binary.
-
-FROM base AS apid-build
```
smira, line 194: I forgot to clean this up in previous PRs

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_state.go
```diff
@@ -71,7 +72,7 @@ func NewState() (*State, error) {
 	for _, r := range []resource.Resource{
 		&v1alpha1.BootstrapStatus{},
 		&v1alpha1.Service{},
-		&v1alpha1.TimeStatus{},
+		&v1alpha1.NetworkConfiguration{},
```
smira, line 75: oops, this should be removed

internal/pkg/ntp/ntp.go
```diff
@@ -0,0 +1,307 @@
...
+
+	if lastSyncServer == "" {
+		var serverList []string
+
+		serverList, err = syncer.resolveServers(ctx)
+		if err != nil {
+			return lastSyncServer, resp, err
+		}
+
+		for _, server := range serverList {
```
Unix4ever, line 187: Not a big deal, but I wonder if it should skip previous `lastSyncServer` which has just failed.

Unix4ever (approved):

internal/pkg/ntp/ntp.go
smira: yep, good point, I haven't thought about it

smira: fixed!

internal/app/machined/pkg/controllers/time/sync.go
```diff
@@ -0,0 +1,157 @@
...
+		if err = r.Update(ctx, time.NewStatus(), func(r resource.Resource) error {
+			r.(*time.Status).SetStatus(time.StatusSpec{
+				Epoch:        epoch,
+				Synced:       timeSynced,
+				SyncDisabled: syncDisabled,
+			})
+
+			return nil
+		}); err != nil {
+			return fmt.Errorf("error updating objects: %w", err) // nolint: govet
```
AlekSi, line 154:
```suggestion
			return fmt.Errorf("error updating objects: %w", err) //nolint:govet
```

The same everywhere. See https://github.com/golangci/golangci-lint/issues/1658 and https://github.com/golang/go/issues/43776

AlekSi (commented):
I will do a deeper review first thing tomorrow :)

internal/app/machined/pkg/controllers/time/sync.go
smira: yes, I missed that, `golangci-lint` was confusing me a lot on this line...

internal/app/machined/pkg/controllers/time/sync.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if syncer != nil {
+			syncer.SetTimeServers(timeServers)
+		}
+
+		if syncDisabled {
+			timeSynced = true
+		}
+
+		if err = r.Update(ctx, time.NewStatus(), func(r resource.Resource) error {
```
AlekSi, line 145: I think we should be able to write at least a basic unit test if we mock controller.Runtime and checks calls to `r.Update`.

internal/app/machined/pkg/controllers/time/time.go
```diff
@@ -0,0 +1,6 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package time contains controllers managing time, synchronization, etc.
+package time
```
AlekSi, line 6: Do we need that file with _just_ a package description? Should it be moved to the main file?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1705,27 +1656,41 @@ func BootstrapEtcd(seq runtime.Sequence, data interface{}) (runtime.TaskExecutio
...
+			//nolint:errcheck
+			defer dir.Close()
 
-		for _, file := range files {
-			fullPath := filepath.Join(constants.EtcdDataPath, file.Name())
+			var files []os.FileInfo
 
-			if err = os.RemoveAll(fullPath); err != nil {
-				return fmt.Errorf("failed to remove %q: %w", file.Name(), err)
+			files, err = dir.Readdir(0)
```
AlekSi, line 1678: I wonder if we can remove the whole `constants.EtcdDataPath` instead. `if os.IsNotExist(err) { return nil }` above means that some other code creates this directory.

internal/pkg/ntp/ntp.go
```diff
@@ -0,0 +1,326 @@
...
+	for {
+		lastSyncServer, resp, err := syncer.query(ctx)
+		if err != nil {
+			return
+		}
+
+		if resp != nil && resp.Validate() == nil {
+			if err = syncer.adjustTime(resp.ClockOffset, lastSyncServer); err != nil {
+				syncer.logger.Printf("error adjusting time: %s", err)
+			} else if !syncer.timeSyncNotified {
```
AlekSi, line 150: Use `switch` too?

internal/pkg/ntp/ntp.go
```diff
@@ -0,0 +1,326 @@
...
+			}
+		}
+
+		// Set some variance with how frequently we poll ntp servers.
+		// This is based on rand(MaxPoll) + MinPoll so we wait at least
+		// MinPoll.
+		sleepInterval := time.Duration(rand.Intn(int(syncer.MaxPoll.Seconds())))*time.Second + syncer.MinPoll
+
+		if resp == nil {
+			// if no response was ever received, consider doing short sleep to retry sooner as it's not KISS of death response
```
AlekSi, line 164: "Kiss-o'-Death". Took me a while to realize that KISS is not acronym there

internal/app/machined/pkg/controllers/time/time.go
smira: That's the style I tried to follow - have a file matching package name for docs and very global definitions for the package

internal/app/machined/pkg/controllers/time/sync.go
smira: We can't do that without hacks, because we can't run actual syncer (which is unit-tested separatedly).

We have unit-tests for the controllers without much mocking.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: I don't remember what was the reason to put it this way, but I remember there was some. Probably @andrewrynhard might know, but once again deserves a separate PR even if it is unneccessary.

internal/pkg/ntp/ntp.go
smira: this is a function call, with error handling, so it shouldn't be turned into `switch`

AlekSi: I mean, may we do the same as https://github.com/talos-systems/talos/pull/3360#discussion_r602187161 to eliminate `else if`?

AlekSi:
```go
err = syncer.adjustTime(resp.ClockOffset, lastSyncServer)
switch err {
  case nil:
    if !syncer.timeSyncNotified { … }
  default:
    syncer.logger.Printf("error adjusting time: %s", err)
}
```

smira: I don't like `switch` but I will look into rewriting this

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
andrewrynhard: I don't quite remember why we had to do it this way but I do remember being frustrated with having to do it this way. But worth a try in another PR to clean it up.

internal/app/machined/pkg/controllers/time/sync.go
smira: and now comes with unit-tests 90% coverage for the controller itself

internal/app/machined/pkg/controllers/time/sync.go
```diff
@@ -0,0 +1,178 @@
...
+	EpochChange() <-chan struct{}
+	SetTimeServers([]string)
+}
+
+// NewNTPSyncerFunc function allows to replace ntp.Syncer with the mock.
+type NewNTPSyncerFunc func(*log.Logger, []string) NTPSyncer
+
+// Run implements controller.Controller interface.
+//
+//nolint: gocyclo,cyclop
```
AlekSi, line 55:
```suggestion
//nolint:gocyclo,cyclop
```

In other places too.

I will look into writing a [ruleguard](https://github.com/quasilyte/go-ruleguard) rule for that.

Unix4ever (dismissed):

internal/app/machined/pkg/controllers/time/sync.go
smira: fixed

Unix4ever (approved):

PR #3361, opened 2021-03-24
feat: add sonobuoy conformance mode

smira:
This is just PoC.


smira:
will be replaced with a proper PR

PR #3362, opened 2021-03-24
fix: CNI config validation should apply to cp nodes, encryption config

smira:
Encryption config should be checked for state partition as well.


Unix4ever (approved):
good catch

Unix4ever (approved):

rsmitty (approved):

PR #3363, opened 2021-03-24
chore: cache go modules in the build

smira:
This does proper caching for Go modules so that when go.mod/go.sum are
changed, only updated modules are downloaded vs. all of them.


Unix4ever (approved):

AlekSi (commented):
I wonder if we should set `GOCACHE`, `GOMODCACHE` (explicitly) and `GOLANGCI_LINT_CACHE` at the base image to various `/go/pkg` subdirectories to pass only one `--mount=type=cache,target=/go/pkg` flag.

Dockerfile
```diff
@@ -456,9 +457,10 @@ COPY --from=unit-tests-runner /src/coverage.txt /coverage.txt
 FROM base AS unit-tests-race
 RUN unlink /etc/ssl
 COPY --from=rootfs / /
-COPY hack/golang/test.sh /bin
 ARG TESTPKGS
-RUN --security=insecure --mount=type=cache,id=testspace,target=/tmp --mount=type=cache,target=/.cache/go-build /bin/test.sh --race ${TESTPKGS}
+ENV PLATFORM container
+ENV CGO_ENABLED 1
+RUN --security=insecure --mount=type=cache,id=testspace,target=/tmp --mount=type=cache,target=/.cache/go-build --mount=type=cache,target=/go/pkg go test -v -race -count 1 -p 4 ${TESTPKGS}
```
AlekSi, line 463: Why we pass `-p 4` explicitly?

smira: because otherwise it eats all the CPUs available on the node which kills the CI node or your computer; our unit-tests are not truly unit, some launch containerd, etc., which introduces more load than a node can handle. `-p 4` was a compromise

smira:
> I wonder if we should set `GOCACHE`, `GOMODCACHE` (explicitly) and `GOLANGCI_LINT_CACHE` at the base image to various `/go/pkg` subdirectories to pass only one `--mount=type=cache,target=/go/pkg` flag.

I like this idea

Dockerfile
```diff
@@ -170,49 +173,12 @@ ARG REGISTRY
...
 ARG IMAGES_PKGS="github.com/talos-systems/talos/pkg/images"
 WORKDIR /src/internal/app/machined
-RUN --mount=type=cache,target=/.cache/go-build go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Talos -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${VERSION_PKG}.PkgsVersion=${PKGS} -X ${VERSION_PKG}.ExtrasVersion=${EXTRAS} -X ${IMAGES_PKGS}.Username=${USERNAME} -X ${IMAGES_PKGS}.Registry=${REGISTRY}" -o /machined
+RUN --mount=type=cache,target=/.cache go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Talos -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${VERSION_PKG}.PkgsVersion=${PKGS} -X ${VERSION_PKG}.ExtrasVersion=${EXTRAS} -X ${IMAGES_PKGS}.Username=${USERNAME} -X ${IMAGES_PKGS}.Registry=${REGISTRY}" -o /machined
 RUN chmod +x /machined
 
 FROM scratch AS machined
 COPY --from=machined-build /machined /machined
 
-# The timed target builds the timed binary.
```
smira, line 179: gc'ing unused targets

Dockerfile
```diff
@@ -127,20 +129,21 @@ FROM build AS base
...
 COPY --from=generate /pkg/machinery/api ./pkg/machinery/api
 COPY --from=generate /pkg/machinery/config ./pkg/machinery/config
-RUN go list -mod=readonly all >/dev/null
-RUN ! go mod tidy -v 2>&1 | grep .
+RUN --mount=type=cache,target=/.cache go list -mod=readonly all >/dev/null
+RUN --mount=type=cache,target=/.cache ! go mod tidy -v 2>&1 | grep .
 WORKDIR /src/pkg/machinery
-RUN go list -mod=readonly all >/dev/null
-RUN ! go mod tidy -v 2>&1 | grep .
+RUN --mount=type=cache,target=/.cache go mod download
```
AlekSi, line 144: We already do that on line 132, so I guess one can be removed?

Dockerfile
```diff
@@ -170,49 +173,12 @@ ARG REGISTRY
 ARG VERSION_PKG="github.com/talos-systems/talos/pkg/version"
 ARG IMAGES_PKGS="github.com/talos-systems/talos/pkg/images"
 WORKDIR /src/internal/app/machined
-RUN --mount=type=cache,target=/.cache/go-build go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Talos -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${VERSION_PKG}.PkgsVersion=${PKGS} -X ${VERSION_PKG}.ExtrasVersion=${EXTRAS} -X ${IMAGES_PKGS}.Username=${USERNAME} -X ${IMAGES_PKGS}.Registry=${REGISTRY}" -o /machined
+RUN --mount=type=cache,target=/.cache go build -ldflags "-s -w -X ${VERSION_PKG}.Name=Talos -X ${VERSION_PKG}.SHA=${SHA} -X ${VERSION_PKG}.Tag=${TAG} -X ${VERSION_PKG}.PkgsVersion=${PKGS} -X ${VERSION_PKG}.ExtrasVersion=${EXTRAS} -X ${IMAGES_PKGS}.Username=${USERNAME} -X ${IMAGES_PKGS}.Registry=${REGISTRY}" -o /machined
```
AlekSi, line 176: I think we could extract this long line with all flags as a Makefile target, or generate a file with `go generate`, include it with `go:embed` and parse at runtime – to avoid copy&pasting that long line several times.

But probably not in that PR :)

AlekSi (approved):
LGTM with small comments

Dockerfile
smira: no, it can't -- we add more files from generate target here, need one more download

PR #3364, opened 2021-03-24
fix: resolve the issue with DHCP lease not being renewed

Unix4ever:
Very easily reproduced when you start a node with a Dynamic IP.
Normally it should renew lease after TTL/2, but that doesn't happen, so
the node starts to get next IP one after another.

After looking at packets sent by other clients, found out that they
have `Client IP address` equal to the IP given by the DHCP server.

Additionally, changed DHCP client to send `Request` packets directly to the DHCP server after getting an offer.
It looks like DHCP spec states that you should use unicast request directly to DHCP server, not broadcast.

### DIFF
--- a/internal/app/networkd/pkg/address/dhcp4.go
+++ b/internal/app/networkd/pkg/address/dhcp4.go
@@ -207,10 +207,24 @@ func (d *DHCP4) discover(ctx context.Context, logger *log.Logger) error {
 	}
 
 	mods := []dhcpv4.Modifier{dhcpv4.WithRequestedOptions(opts...)}
+	clientOpts := []nclient4.ClientOpt{}
+
+	if d.Offer != nil {
+		// do not use broadcast, but send the packet to DHCP server directly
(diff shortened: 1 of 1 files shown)


smira:
Nice catch! I wonder if there are some helpers in the library we could use to build this flow better, as the library is supposed to build correct packet from the offer

Unix4ever:
> Nice catch! I wonder if there are some helpers in the library we could use to build this flow better, as the library is supposed to build correct packet from the offer

I also wondered about that, but doesn't seem like it:
https://github.com/insomniacslk/dhcp/blob/c51060810aaab9c8a0bd1b0fcbf72bc0b91e6427/dhcpv4/dhcpv4.go#L249

Looks like when it builds the request from the offer it's always using ClientIPAddr

smira (approved):

smira:
I pushed the PR to the backports, as it seems to be a serious problem

Unix4ever:
> I pushed the PR to the backports, as it seems to be a serious problem

Created backport PRs for both :)

PR #3381, opened 2021-03-26
fix: get rid of data race in encoder and fix concurrent map access

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3377  https://github.com/talos-systems/talos/issues/3380

Fixed data race in the encoder documentation examples by using `sync.Once`.
We only need to generate them once anyways and then it's not a big deal
that we are using the same pointers everywhere as they're pretty much
constant.

As of `system.go`, looks like we actually have concurrent operations for
partitions unmount. Just added a mutex there.

### DIFF
--- a/internal/pkg/mount/system.go
+++ b/internal/pkg/mount/system.go
@@ -7,6 +7,7 @@ package mount
 import (
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/talos-systems/go-blockdevice/blockdevice"
 	"github.com/talos-systems/go-blockdevice/blockdevice/filesystem"
@@ -20,7 +21,10 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
-var mountpoints = map[string]*Point{}
+var (
+	mountpoints      = map[string]*Point{}
+	mountpointsMutex sync.RWMutex
+)
 
 // SystemMountPointsForDevice returns the mountpoints required to boot the system.
 // This function is called exclusively during installations ( both image
@@ -195,6 +199,9 @@ func SystemPartitionMount(r runtime.Runtime, label string, opts ...Option) (err 
 		return err
 	}
 
+	mountpointsMutex.Lock()
+	defer mountpointsMutex.Unlock()
+
 	mountpoints[label] = mountpoint
 
 	return nil
@@ -202,14 +209,22 @@ func SystemPartitionMount(r runtime.Runtime, label string, opts ...Option) (err 
 
 // SystemPartitionUnmount unmounts a system partition by the label.
 func SystemPartitionUnmount(r runtime.Runtime, label string) (err error) {
-	if mountpoint, ok := mountpoints[label]; ok {
-		err = mountpoint.Unmount()
-		if err != nil {
-			return err
-		}
+	mountpointsMutex.RLock()
+	mountpoint, ok := mountpoints[label]
+	mountpointsMutex.RUnlock()
 
-		delete(mountpoints, label)
+	if !ok {
+		return nil
 	}
+
+	err = mountpoint.Unmount()
+	if err != nil {
+		return err
+	}
+
+	mountpointsMutex.Lock()
+	delete(mountpoints, label)
+	mountpointsMutex.Unlock()
 
 	return nil
 }
--- a/pkg/machinery/config/encoder/documentation.go
+++ b/pkg/machinery/config/encoder/documentation.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"regexp"
 	"strings"
+	"sync"
 
 	yaml "gopkg.in/yaml.v3"
 )
@@ -51,7 +52,7 @@ func (d *Doc) AddExample(name string, value interface{}) {
 
 	d.Examples = append(d.Examples, &Example{
 		Name:  name,
-		Value: value,
+		value: value,
 	})
 }
 
@@ -74,8 +75,43 @@ func (d *Doc) Describe(field string, short bool) string {
 
 // Example represents one example snippet for a type.
 type Example struct {
-	Name  string
-	Value interface{}
+	populate sync.Once
+	Name     string
+
+	valueMutex sync.RWMutex
+	value      interface{}
+}
+
+// Populate populates example value.
+func (e *Example) Populate(index int) {
+	e.populate.Do(func() {
+		if reflect.TypeOf(e.value).Kind() != reflect.Ptr {
+			return
+		}
+
+		v := reflect.ValueOf(e.value).Elem()
+
+		defaultValue := getExample(v, getDoc(e.value), index)
+
+		e.valueMutex.Lock()
+		defer e.valueMutex.Unlock()
+
+		if defaultValue != nil {
+			v.Set(defaultValue.Convert(v.Type()))
+		}
+
+		populateNestedExamples(v, index)
+	})
+}
+
+// GetValue returns example value.
+func (e *Example) GetValue() interface{} {
+	e.valueMutex.RLock()
+	defer func() {
+		e.valueMutex.RUnlock()
+	}()
+
+	return e.value
 }
 
 // Field gets field from the list of fields.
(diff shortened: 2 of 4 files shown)


pkg/machinery/config/encoder/documentation.go
```diff
@@ -74,10 +75,18 @@ func (d *Doc) Describe(field string, short bool) string {
 
 // Example represents one example snippet for a type.
 type Example struct {
+	sync.Once
```
smira, line 78: do we ever copy/pass `Example` by value? (vs. by reference)

smira (approved):
:+1: 

one small question

pkg/machinery/config/encoder/documentation.go
Unix4ever: we _shouldn't_ but let me check for sure.

AlekSi: That should be checked by vet's copylock check

Unix4ever: yeah, should be good.

internal/pkg/mount/system.go
```diff
@@ -195,13 +199,19 @@ func SystemPartitionMount(r runtime.Runtime, label string, opts ...Option) (err
...
 	return nil
 }
 
 // SystemPartitionUnmount unmounts a system partition by the label.
 func SystemPartitionUnmount(r runtime.Runtime, label string) (err error) {
+	mountpointsMutex.Lock()
+	defer mountpointsMutex.Unlock()
+
 	if mountpoint, ok := mountpoints[label]; ok {
 		err = mountpoint.Unmount()
```
AlekSi, line 216: I think we better use more granular lock there to avoid wrapping (potentially long) Umount() call with lock.

Even better, we could use RWMutex.

smira: great point, I think we could read mountpoint, and unlock mutex? we don't need to hold the mutex for the duration of the mount

AlekSi: Exactly. We can wrap `mountpoints` reads with RLock / RUnlock, and `delete` below (and assignment above) with Lock / Unlock

Unix4ever: sure, fixed.

smira (approved):

internal/pkg/mount/mount.go
```diff
@@ -161,6 +162,7 @@ func mountRetry(f RetryFunc, p *Point, isUnmount bool) (err error) {
 
 // Point represents a Linux mount point.
 type Point struct {
+	sync.Mutex
```
AlekSi, line 165: Embedding mutex adds Lock and Unlock methods to Point's exported method set. That's never a good idea. That should be a regular field

internal/pkg/mount/mount.go
```diff
@@ -266,6 +268,9 @@ func (p *Point) Mount() (err error) {
 // Unmount attempts to retry an unmount on EBUSY. It will attempt a
 // retry every 100 milliseconds over the course of 5 seconds.
 func (p *Point) Unmount() (err error) {
+	p.Lock()
+	defer p.Unlock()
```
AlekSi, line 272: There are much more places where Point's fields are accessed and should be protected by mutex.

Or maybe we can say that Point is not thread-safe, and it is caller's responsibility to synchronize access to Point's methods

internal/pkg/mount/system.go
```diff
@@ -195,21 +199,32 @@ func SystemPartitionMount(r runtime.Runtime, label string, opts ...Option) (err
...
+	mountpointsMutex.RLock()
+	mountpoint, ok := mountpoints[label]
+	mountpointsMutex.RUnlock()
 
-		delete(mountpoints, label)
+	if !ok {
+		return nil
 	}
 
+	err = mountpoint.Unmount()
```
AlekSi, line 220: Do we actually call Unmount concurrently for the same mountpoint?

Unix4ever: I guess that concurrent map access race showed that we do.

internal/pkg/mount/mount.go
Unix4ever: I can move it to the `system.go` then.

internal/pkg/mount/mount.go
Unix4ever: What particular issues can embedding mutex Lock and Unlock cause?

internal/pkg/mount/system.go
smira: I don't think so (hopefully). I think we don't need to lock here

Unix4ever: ok then

AlekSi: > I guess that concurrent map access race showed that we do.

That was a concurrent access to `mountpoints` map, not to one `mountpoints` element

pkg/machinery/config/encoder/documentation.go
```diff
@@ -74,8 +75,28 @@ func (d *Doc) Describe(field string, short bool) string {
 
 // Example represents one example snippet for a type.
 type Example struct {
-	Name  string
-	Value interface{}
+	sync.Once
```
AlekSi, line 78: The same problem with embedding – one can call `e.Do(…)`. That's an implementation detail that should not be exposed; that should be a regular field.

pkg/machinery/config/encoder/documentation.go
```diff
@@ -74,8 +75,28 @@ func (d *Doc) Describe(field string, short bool) string {
...
+	valueMutex sync.RWMutex
```
AlekSi, line 79:
```suggestion

	valueMutex sync.RWMutex
```

https://dmitri.shuralyov.com/idiomatic-go#mutex-hat

AlekSi (approved):

AlekSi:
But CI fails now :)

Unix4ever:
yes, that's the price for making example.value unexported...

pkg/machinery/config/encoder/documentation.go
```diff
@@ -239,7 +264,7 @@ func readExample(v reflect.Value, doc *Doc, index int) {
 		index = numExamples - 1
 	}
 
-	defaultValue := reflect.ValueOf(doc.Examples[index].Value)
+	defaultValue := reflect.ValueOf(doc.Examples[index].value)
```
AlekSi, line 267:
```suggestion
	defaultValue := reflect.ValueOf(doc.Examples[index].GetValue())
```

pkg/machinery/config/encoder/documentation.go
```diff
@@ -74,8 +75,31 @@ func (d *Doc) Describe(field string, short bool) string {
...
+		populateExamples(e.value, index)
+	})
+}
+
+// GetValue returns example value.
+func (e *Example) GetValue() interface{} {
+	e.valueMutex.RLock()
+	defer func() {
+		e.valueMutex.RUnlock()
+	}()
```
AlekSi, line 100:
```suggestion
	defer e.valueMutex.RUnlock()
```

pkg/machinery/config/encoder/documentation.go
Unix4ever: nope, that's causing a deadlock. As we surround `populateExamples` with `Lock` `Unlock` already.
Not particularly like this solution though.

AlekSi: Uh!

Unix4ever: yeah, I feel I need more thorough refactoring here. Doing that now...

Unix4ever: now it should become slightly better.

internal/pkg/mount/system.go
```diff
@@ -195,21 +199,32 @@ func SystemPartitionMount(r runtime.Runtime, label string, opts ...Option) (err
...
 }
 
 // SystemPartitionUnmount unmounts a system partition by the label.
 func SystemPartitionUnmount(r runtime.Runtime, label string) (err error) {
-	if mountpoint, ok := mountpoints[label]; ok {
-		err = mountpoint.Unmount()
-		if err != nil {
-			return err
-		}
+	mountpointsMutex.RLock()
```
smira, line 212: I feel `RWMutex` is an overkill for the sake of a single ready here, but it doesn't matter much either way

smira (approved):

AlekSi (approved):

PR #3385, opened 2021-03-29
feat: add ability to disable comments in talosctl gen config

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3384

Instead of doing simple `--no-comments` flag, decided to use more
granular approach which allows to either disable examples, or docstring,
or both.

Thus the command looks like this:

```bash
talosctl gen config --with-docs=false --with-examples=false <...>
```

Both are enabled by default to provide better UX for users learning
Talos.


AlekSi:
@Unix4ever There is a merge conflict now

Unix4ever:
> @Unix4ever There is a merge conflict now

yep, that was expected. Should be good now

smira (commented):
lgtm, could you please update AWS docs for 0.10 to include those `--with-docs=false --with-examples=false` so that it fits the userdata size? (AFAIK it was 16KiB)

smira (approved):

PR #3392, opened 2021-03-29
test: add unit-test for machine config validation

smira:
Follow-up for #3383

I added couple of first tests, we should add more as we go through this
code. Even with those tests, I found and fixed two more panics.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
@@ -64,6 +64,8 @@ func (c *Config) Validate(mode config.RuntimeMode, options ...config.ValidationO
 
 	if c.MachineConfig == nil {
 		result = multierror.Append(result, errors.New("machine instructions are required"))
+
+		return result.ErrorOrNil()
 	}
 
 	if err := c.ClusterConfig.Validate(); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -64,6 +64,8 @@ func (c *Config) Validate(mode config.RuntimeMode, options ...config.ValidationO
 
 	if c.MachineConfig == nil {
 		result = multierror.Append(result, errors.New("machine instructions are required"))
+
+		return result.ErrorOrNil()
```
Unix4ever, line 68: I understand that it fixes a panic, but I feel that we may proceed here and continue validation to show all config errors at once. Otherwise we bail out too early and skip `ClusterConfig` validation.
I guess consequent checks should properly handle `c.MachineConfig` being nil.

Unix4ever (commented):
`

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
smira: looks like it will fail on everything at that point, because even node type is no longer defined, so I thought in this specific case it's better to exit early, as every other check makes little sense

smira: but in general, yeah, I agree, we should continue, but this is so weird that it makes little sense to continue

Unix4ever: ah, I see. Makes sense

Unix4ever (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
Unix4ever: yeah, might need to continue as far as we can :) But this one looks like a fatal error

PR #3393, opened 2021-03-29
fix: don't write to config object on access

smira:
This avoids data race on config access: config object might be accessed
concurrently and it should be read-only on access.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
@@ -237,7 +237,7 @@ func (m *MachineConfig) Server() string {
 // Sysctls implements the config.Provider interface.
 func (m *MachineConfig) Sysctls() map[string]string {
 	if m.MachineSysctls == nil {
-		m.MachineSysctls = make(map[string]string)
+		return make(map[string]string)
 	}
 
 	return m.MachineSysctls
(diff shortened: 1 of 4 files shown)


Unix4ever (dismissed):
`v1alpha1_provider.go` still has couple more places where we write to config though:

https://github.com/talos-systems/talos/blob/0685cb44e7a3eab1bbc905642b94ccde5811fdae/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L293

https://github.com/talos-systems/talos/blob/0685cb44e7a3eab1bbc905642b94ccde5811fdae/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L447

But I'm fine if that's going to be done in scope of refactoring.

smira:
> `v1alpha1_provider.go` still has couple more places where we write to config though:
> 
> https://github.com/talos-systems/talos/blob/0685cb44e7a3eab1bbc905642b94ccde5811fdae/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L293
> 
> https://github.com/talos-systems/talos/blob/0685cb44e7a3eab1bbc905642b94ccde5811fdae/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L447
> 
> But I'm fine if that's going to be done in scope of refactoring.

thanks, I missed those!

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -440,13 +436,9 @@ func (n *NetworkConfig) GetDevice(name string) *Device {
 		}
 	}
 
-	dev := &Device{
+	return &Device{
 		DeviceInterface: name,
 	}
-
-	n.NetworkInterfaces = append(n.NetworkInterfaces, dev)
```
AlekSi, line 447:
```
// GetDevice adds or returns existing Device by name.
```

It no longer "adds" device. Not sure if some caller depends on it

smira: not sure either, good question, but certainly we need to stop doing that, I will do more review to see how it got actually used

smira: nice catch, this change was wrong, and this method is not used to access the config.

Unix4ever (approved):

PR #3394, opened 2021-03-29
chore: build components with race detector enabled in dev mode

smira:
This provides a variable to build core Talos components with race
detector enabled: `make initramfs WITH_RACE=yes`.

Our code still doesn't quite run in this mode, but this is a step
forward to check with race detector enabled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ IMAGE_TAG ?= $(TAG)
 BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 REGISTRY_AND_USERNAME := $(IMAGE_REGISTRY)/$(USERNAME)
 DOCKER_LOGIN_ENABLED ?= true
+NAME = Talos
 
 ARTIFACTS := _out
 TOOLS ?= ghcr.io/talos-systems/tools:v0.5.0-alpha.0-3-g41b8073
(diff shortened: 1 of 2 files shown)


Unix4ever (dismissed):

Makefile
```diff
@@ -29,6 +29,18 @@ RELEASES ?= v0.8.5 v0.9.0
 SHORT_INTEGRATION_TEST ?=
 CUSTOM_CNI_URL ?=
 
+CGO_ENABLED ?= 0
+GO_EXTRA_BUILDFLAGS ?=
+GO_EXTRA_LDFLAGS ?=
+
+RACE_DETECTOR_ENABLED ?=
```
AlekSi, line 36: I would call it `WITH_RACE_DETECTOR` or just `WITH_RACE` – because what it essentially does is adds a `-race` flag. In the future, we could add something like `WITH_DEBUG`, `WITH_ASSERTS`, etc.

I also think we should differentiate between internal variables and variables that are expected to be set by a contributor (like `PUSH`, `USERNAME`, etc.). Probably by adding comments.

AlekSi (approved):

Makefile
smira: yeah, `WITH_RACE` is shorter. as for the variables -- they are not truly internal, you can set them. internal variables are not overrideable, these vars you can override if you want to test some new build flag.

AlekSi: I mean, those variables _can_ be overridden, but only a subset of them are _expected_ to be. For example, `PUSH` and `USERNAME` are expected to be overridden by contributor – to be able to contribute. Something like `GO_EXTRA_BUILDFLAGS` – not really. So `PUSH` and `USERNAME` should have comments and/or (preferably "and" :) ) be mentioned in `make help` output, but `GO_EXTRA_BUILDFLAGS` shouldn't.

smira: yep, agree, I mentioned `WITH_RACE` there

Unix4ever (dismissed):

andrewrynhard (approved):

PR #3398, opened 2021-03-30
feat: add API and command to save etcd snapshot (backup)

### LINKED ISSUE #3408: etcd backup (snapshot)
## Feature Request

Provide a way to backup etcd database from the Talos clusters. 

### Description

`talosctl etcd snapshot <filename>` downloads etcd snapshot from the control plane node and saves that in the `filename`

smira:
This adds a simple API and `talosctl etcd snapshot` command to stream
snapshot of etcd from one of the control plane nodes to the local file.

Fixes #3408 


cmd/talosctl/cmd/talos/etcd.go
```diff
@@ -117,7 +121,73 @@ var etcdMemberListCmd = &cobra.Command{
...
+			size, err := io.Copy(dest, r)
+			if err != nil {
+				return fmt.Errorf("error reading: %w", err)
+			}
+
+			if err = dest.Sync(); err != nil {
+				return fmt.Errorf("failed to fsync: %w", err)
+			}
+
+			if (size % 512) != sha256.Size {
```
smira, line 175: this looks weird, but that's mostly porting code from `etcdctl`: https://github.com/etcd-io/etcd/blob/client/v3.5.0-alpha.0/client/v3/snapshot/v3_snapshot.go#L46

internal/integration/cli/etcd.go
```diff
@@ -39,6 +44,22 @@ func (suite *EtcdSuite) TestForfeitLeadership() {
 	)
 }
 
+// TestSnapshot tests etcd snapshot (backup).
+func (suite *EtcdSuite) TestSnapshot() {
```
Unix4ever, line 48: Is there a way to validate that the snapshot is valid? Like some primitive check that something can open it.

Unix4ever (dismissed):

internal/integration/cli/etcd.go
smira: it requires at least `etcdctl` :roll_eyes: 

once we have recovery flow, we'll combine both in the integration test, so we'll get some validation for it

api/machine/machine.proto
```diff
@@ -30,6 +30,7 @@ service MachineService {
       returns (EtcdLeaveClusterResponse);
   rpc EtcdForfeitLeadership(EtcdForfeitLeadershipRequest)
       returns (EtcdForfeitLeadershipResponse);
+  rpc EtcdSnapshot(EtcdSnapshotRequest) returns (stream common.Data);
```
AlekSi, line 33: I'm not sure we should share request and response messages. I think that's where duplication is worth it because it will allow us to modify messages independently.

Also, this time is good as any other to start adding API method documentation comments :)

cmd/talosctl/cmd/talos/etcd.go
AlekSi: That info probably should be in the code comment

api/machine/machine.proto
smira: I thought about that, but there's benefit here to common response type - we have common layer handling `stream common.Data`, so it's better to have it same vs. special

Unix4ever (approved):

PR #3435, opened 2021-04-05
chore: bump dependencies via dependabot

smira:
See #3431 #3432 #3433 #3434

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -41,7 +41,7 @@ require (
 	github.com/gdamore/tcell/v2 v2.2.0
 	github.com/gizak/termui/v3 v3.1.0
 	github.com/gogo/googleapis v1.4.0 // indirect
-	github.com/golang/protobuf v1.5.1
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -41,7 +41,7 @@ require (
 	github.com/gdamore/tcell/v2 v2.2.0
 	github.com/gizak/termui/v3 v3.1.0
 	github.com/gogo/googleapis v1.4.0 // indirect
-	github.com/golang/protobuf v1.5.1
+	github.com/golang/protobuf v1.5.2
```
AlekSi, line 44: The whole module is deprecated now. Does it pass `make lint`?

AlekSi (approved):

Unix4ever (approved):

go.mod
smira: yep, I think there's some legacy we need eventually get rid of to remove this one

PR #3443, opened 2021-04-06
feat: support etcd recovery from snapshot on bootstrap

### LINKED ISSUE #3409: etcd recovery on bootstrap
## Feature Request

Given that etcd backup was created via #3408, in case of control plane complete failure provide a way to recover a control plane from the backup.

### Description

Failed control plane nodes are decommissioned.

New control plane is created with same machine configuration as before, it enters the waiting loop for `etcd` to join. 

User can now do `talosctl etcd bootstrap --from-backup=etcd.db` to bootstrap etcd from the snapshot recovering the state of the cluster.

## Tests

Take snapshot, wipe all control plane nodes, wait for the nodes to enter bootstrap waiting state, recover from the backup, cluster should be healthy.

smira:
When Talos `controlplane` node is waiting for a bootstrap, `etcd`
contents can be recovered from a snapshot created with
`talosctl etcd snapshot` on a healthy cluster.

Bootstrap process goes same way as before, but the etcd data directory
is recovered from the snapshot.

This flow enables disaster recovery for the control plane: given that
periodic backups are available, destroy control plane nodes, re-create
them with the same config, and bootstrap one node with the saved
snapshot to recover etcd state at the time of the snapshot.

Fixes #3409


go.mod
```diff
@@ -81,6 +81,7 @@ require (
 	github.com/vmware/vmw-guestinfo v0.0.0-20200218095840-687661b8bd8e
 	go.etcd.io/etcd/api/v3 v3.5.0-alpha.0
 	go.etcd.io/etcd/client/v3 v3.5.0-alpha.0
+	go.etcd.io/etcd/etcdctl/v3 v3.5.0-alpha.0
```
smira, line 84: this brings in _tons_ of dependencies, but not much we can can do about it - the restore code is not trivial and it's not separated out as a module to consume

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1171,7 +1171,12 @@ func UncordonNode(seq runtime.Sequence, data interface{}) (runtime.TaskExecution
 
 		var kubeHelper *kubernetes.Client
 
-		if kubeHelper, err = kubernetes.NewClientFromKubeletKubeconfig(); err != nil {
+		if err = retry.Constant(5*time.Minute, retry.WithUnits(time.Second), retry.WithErrorLogging(true)).RetryWithContext(ctx,
```
smira, line 1174: etcd recovery triggered a bug here: kubelet cert might not be ready by the time boot sequence comes here

pkg/machinery/client/client.go
```diff
@@ -521,8 +521,10 @@ func (c *Client) Rollback(ctx context.Context) (err error) {
 }
 
 // Bootstrap implements the proto.MachineServiceClient interface.
-func (c *Client) Bootstrap(ctx context.Context) (err error) {
-	resp, err := c.MachineClient.Bootstrap(ctx, &machineapi.BootstrapRequest{})
+func (c *Client) Bootstrap(ctx context.Context, recoverEtcd bool) (err error) {
```
smira, line 524: use `*BootstrapRequest` here

Unix4ever (approved):

PR #3456, opened 2021-04-09
feat: implement inline manifests in the machine configuration

### LINKED ISSUE #3416: inline bootstrap manifests
Add an option to the machine configuration like `cluster.inlineManifests` which allows multi-line strings as values, e.g.:

```yaml
cluster:
   inlineManifests:
     - name: foo-manifest
       manifest: | 
         metadata:
              name: foo
```

We need to update manifests controllers to parse the manifests and add them to the global list of manifests.

`name` should be unique across manifests (it will be used to identify resource holding the manifest, to give context to the errors if the manifest doesn't parse).

smira:
Inline manifests work exactly same way as extra manifests, but the
manifest itself can be stored in the config body.

Example config patch:

```
--config-patch '[{"op": "replace", "path": "/cluster/inlineManifests", "value": [{"name": "foo", "contents": "apiVersion: v1\nkind: Namespace\nmetadata:\n  name: ci\n"}]}]'
```

Fixes #3416 

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/extra_manifest.go
+++ b/internal/app/machined/pkg/controllers/k8s/extra_manifest.go
@@ -108,7 +108,7 @@ func (ctrl *ExtraManifestController) Run(ctx context.Context, r controller.Runti
 		for _, manifest := range config.ExtraManifests {
 			var id resource.ID
 
-			id, err = ctrl.download(ctx, r, logger, manifest)
+			id, err = ctrl.process(ctx, r, logger, manifest)
 			if err != nil {
 				multiErr = multierror.Append(multiErr, err)
 			}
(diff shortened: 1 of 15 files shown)


internal/app/machined/pkg/controllers/k8s/extra_manifest.go
```diff
@@ -139,8 +139,22 @@ func (ctrl *ExtraManifestController) Run(ctx context.Context, r controller.Runti
 	}
 }
 
-func (ctrl *ExtraManifestController) download(ctx context.Context, r controller.Runtime, logger *log.Logger, manifest config.ExtraManifest) (id resource.ID, err error) {
-	id = fmt.Sprintf("%s-%s", manifest.Priority, manifest.URL)
+func (ctrl *ExtraManifestController) process(ctx context.Context, r controller.Runtime, logger *log.Logger, manifest config.ExtraManifest) (id resource.ID, err error) {
```
AlekSi, line 142: I think that long function should become shorter, not longer:

```
func process(…) {
  id := …

  if if manifest.InlineManifest != "" {
    return ctrl.processInline(…)
  }

  return ctrl.processURL(…)
}
```

smira: thanks, fixed!

Unix4ever (approved):

AlekSi:
@smira There is a merge conflict

smira:
> @smira There is a merge conflict

I'll get it resolved, but it doesn't block the review process hopefully

AlekSi (approved):

PR #3459, opened 2021-04-09
feat: add support for "none" CNI type

### LINKED ISSUE #3411: add 'none' CNI type
Add `none` CNI type. 

This type implies that CNI will be installed by the user on their own. List of CNI manifests should be empty (validation).

Talos shouldn't generate any manfiests for the CNI.

AlekSi:
Closes #3411.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
@@ -108,14 +108,9 @@ func (c *Config) Validate(mode config.RuntimeMode, options ...config.ValidationO
 	}
 
 	if c.Machine().Type() == machine.TypeInit || c.Machine().Type() == machine.TypeControlPlane {
-		switch c.Cluster().Network().CNI().Name() {
-		case constants.CustomCNI:
-			// custom CNI with URLs or an empty list of manifests which will get applied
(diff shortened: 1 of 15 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -204,6 +199,39 @@ func (c *ClusterConfig) Validate() error {
...
+		fallthrough
+	case constants.NoneCNI:
+		if len(cni.URLs()) != 0 {
+			err := fmt.Errorf(`"urls" field should be empty for %q CNI`, cni.Name())
+			result = multierror.Append(result, err)
+		}
+
+	case constants.CustomCNI:
+		if len(cni.URLs()) == 0 {
+			err := fmt.Errorf(`"urls" field should not be empty for %q CNI`, cni.Name())
```
smira, line 217: this should be warning for now, probably with a note that we can make this an error in 0.12

the reason is that Talos 0.9 accepted `custom` CNI with empty URL list and some users took advantage of that. we fixed the validation in 0.10, but we shouldn't break backwards compatiblity this way, as after upgrade invalid config would lead to node stuck in reboot loop

smira (commented):
:+1: with one change of warning vs. error

PR #3469, opened 2021-04-12
chore: use Go 1.16 global install command

AlekSi:
Plus add stringer tool.

https://golang.org/doc/go1.16#go-command

> go install now accepts arguments with version suffixes (for example, go install example.com/cmd@v1.0.0). This causes go install to build and install packages in module-aware mode, ignoring the go.mod file in the current directory or any parent directory, if there is one. This is useful for installing executables without affecting the dependencies of the main module.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,10 +50,11 @@ RUN ["/toolchain/bin/ln", "-svf", "/toolchain/bin/bash", "/bin/sh"]
 RUN ["/toolchain/bin/ln", "-svf", "/toolchain/etc/ssl", "/etc/ssl"]
 RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b /toolchain/bin v1.38.0
 ARG GOFUMPT_VERSION
-RUN cd $(mktemp -d) \
-    && go mod init tmp \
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -133,7 +134,7 @@ COPY --from=generate-build /api/cluster/*.pb.go /pkg/machinery/api/cluster/
 COPY --from=generate-build /api/storage/*.pb.go /pkg/machinery/api/storage/
 COPY --from=generate-build /api/resource/*.pb.go /pkg/machinery/api/resource/
 COPY --from=generate-build /api/inspect/*.pb.go /pkg/machinery/api/inspect/
-COPY --from=go-generate /pkg/machinery/config/types/v1alpha1/*_doc.go /pkg/machinery/config/types/v1alpha1/
```
smira, line 136: I hope this is not going to be a problem (copying all same files over itself), but I don't see why we need this change right now?

smira (approved):

Dockerfile
AlekSi: This works and does not cause any problems.

That was a part of other work ("worker" vs "join"), but I think it belongs there because we call `go generate` on several packages, and we should copy them all. Otherwise, `//go:generate` results in some other files will be lost.

PR #3488, opened 2021-04-15
feat: print complete member info in etcd members

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3487

Example output:

```
NODE       ID                 HOSTNAME                 PEERS                   CLIENTS
10.5.0.2   c3d3020cf75b8728   talos-default-master-1   https://10.5.0.2:2380   https://10.5.0.2:2379
```


api/machine/machine.proto
```diff
@@ -791,11 +791,17 @@ message EtcdForfeitLeadershipResponse {
...
+message EtcdMember {
+  uint64 id = 2;
+  string hostname = 3;
+  repeated string peer_urls = 4;
+  repeated string client_urls = 5;
+}
+message EtcdMembers { 
   common.Metadata metadata = 1;
-  repeated string members = 2;
+  repeated EtcdMember members = 2;
```
smira, line 802: I think we should provide backwards compat here, i.e. include `repeated string legacy_members = 2;` and make new field use new tag; this way old talosctl will still work

cmd/talosctl/cmd/talos/etcd.go
```diff
@@ -99,22 +100,28 @@ var etcdMemberListCmd = &cobra.Command{
 					continue
 				}
 
-				if i == 0 {
+				for _, member := range message.Members {
+					if i == 0 {
+						if node != "" {
+							fmt.Fprintln(w, "NODE\tID\tHOSTNAME\tPEERS\tCLIENTS")
```
smira, line 106: I think name shouldn't be `PEERS` and `CLIENTS` as this is peer URLs of this node and client URLs of this node

api/machine/machine.proto
```diff
@@ -791,11 +791,17 @@ message EtcdForfeitLeadershipResponse {
 message EtcdMemberListRequest {
   bool query_local = 1;
 }
-message EtcdMemberList {
+message EtcdMember {
```
smira, line 794: it might make sense to add some documentation here as we update the API to make our docs more readable

cmd/talosctl/cmd/talos/etcd.go
```diff
@@ -99,22 +100,28 @@ var etcdMemberListCmd = &cobra.Command{
...
 				}
 
-				if i == 0 {
+				for _, member := range message.Members {
+					if i == 0 {
+						if node != "" {
+							fmt.Fprintln(w, "NODE\tID\tHOSTNAME\tPEER_URLS\tCLIENT_URLS")
+							pattern = "%s\t" + pattern
+						} else {
+							fmt.Fprintln(w, "ID\tHOSTNAME\tPEER_URLS\tCLIENT_URLS")
```
smira, line 109: `PEER_URLS` might be better as `PEER URLS` ? ans same with `CLIENT_URLS`?

smira (dismissed):

cmd/talosctl/cmd/talos/etcd.go
Unix4ever: yep, fixed

smira (dismissed):

PR #3490, opened 2021-04-15
feat: bump k8s version to 1.20.6

rsmitty:
This PR brings in a newer version of 1.20.x to make sure we're not
shipping with https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#cve-2021-25735-validating-admission-webhook-does-not-observe-some-previous-fields


smira (approved):

smira:
on integration failures:

https://github.com/talos-systems/talos/blob/release-0.9/Makefile#L28

I'm not sure how we ended up like that, but why the hell 0.9.1

smira:
ah, you bumped upgrade tests, but forgot to bump Makefile, they should be in sync

PR #3491, opened 2021-04-15
chore: use codecov from the build container

AlekSi:
That's the best we can do now with https://about.codecov.io/security-update/


smira (dismissed):

smira:
we could also embed codecov into the build-container... not bullet-proof, but might better

smira (approved):

PR #3492, opened 2021-04-16
chore: refactor Dockerfile for cross-compilation

smira:
This has two big visible changes:

* `installer` image now contains assets for both `amd64` and `arm64`, so
it can be used to generate any Talos image (including RPi on amd64 host)
* Talos is using cross-compilation instead of emulation to build
non-native architectures: on amd64, Go amd64 compiler produces binaries
for both arm64 and amd64
(before this change: Go arm64 compiler via QEMU produces arm64 binaries on amd64)

CI implications: we no longer require arm64 nodes.

Changes walkthrough:

* `installer` container now keeps assets under `/usr/install/<arch>`
* Dockerfile build starts forcing toolchain/base image to use the build
host native architecture, not target architecture
* lots of duplication for amd64/arm64 as we want to combine assets for
both arches in a single image (e.g. we have multi-arch amd64/arm64
installer image, each arch has native installer binary, but both arches
contain full set of amd64/arm64 assets)
* fixed a small bug preventing arm64 on amd64 talosctl cluster create

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -103,7 +103,7 @@ local volumes = {
 // This provides the docker service.
 local docker = {
   name: 'docker',
-  image: 'docker:19.03-dind',
+  image: 'docker:20.10-dind',
   entrypoint: ['dockerd'],
   privileged: true,
   command: [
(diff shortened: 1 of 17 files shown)


smira:
The only deficiency I see now (which I believe we could fix) is that `strip` for linux-amd64 can't strip `arm64` binaries, so we have extra 300 KiB in the initramfs for arm64. 

I think we could look into some alternative which supports different targets out of the same binary or grrr... cross-compile it.

rsmitty (approved):
:heart:

PR #3506, opened 2021-04-19
add WithRegistryCaCert

halacs:
# Pull Request

As we discussed in Slack, I added new function for the registry CA cert.
Based on my tests, CA cert is created with proper content, however, I suspect just another containerd bug because CA seems to be not in use at all even if containerd.tmpl file also has proper content.

## What? (description)

golang API to add custom CA certificate for registries.

## Why? (reasoning)

We use docker registries with self-signed certificates. Already implemented "insecure_skip_verify" option is buggy in containerd.
Related containerd issue in github: https://github.com/containerd/containerd/issues/5079

## Acceptance

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/generate/options.go
+++ b/pkg/machinery/config/types/v1alpha1/generate/options.go
@@ -79,6 +79,27 @@ func WithRegistryMirror(host string, endpoints ...string) GenOption {
 	}
 }
 
+// WithRegistryCACert specifies the certificate of the certificate authority which signed certificate of the registry.
+func WithRegistryCACert(host, cacert string) GenOption {
+	return func(o *GenOptions) error {
+		if o.RegistryConfig == nil {
+			o.RegistryConfig = make(map[string]*v1alpha1.RegistryConfig)
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/machinery/config/types/v1alpha1/generate/options.go
```diff
@@ -79,6 +79,27 @@ func WithRegistryMirror(host string, endpoints ...string) GenOption {
 	}
 }
 
+// WithRegistryCaCert specifies the certificate of the certificate authority which signed certificate of the registry.
+func WithRegistryCaCert(host, cacert string) GenOption {
```
smira, line 83: nit: the method name be `WithRegistryCACert` (with capital `CA`)

smira (commented):
thanks for your PR, small nit and please rebase / squash it so it is just a single commit.

pkg/machinery/config/types/v1alpha1/generate/options.go
halacs: - WithRegistryCaCert renamed to WithRegistryCACert as requested
- branch rebased to have only one commit in the pull request.

Thanks for the review!

Issue #3515, opened 2021-04-20
VMware OVA does not deploy using html UI


lennardk:
### Description
When manually deploying the vmware OVA for version 0.9.2 to vSphere using the html web interface, vSphere refuses due to alleged issues with the template

### Logs
The errors reported as as follows:
```
55:7:VALUE_ILLEGAL: Value ''3'' of Parent element does not refer to a ref of type DiskControllerReference.
47:7:VALUE_ILLEGAL: Value ''VirtualSCSI'' of ResourceSubType element not found in [lsilogic, lsilogicsas].
```
### Environment

- Talos version: 0.9.2
- Platform: VMware

PR #3516, opened 2021-04-20
fix: update osType in OVA other3xLinux64Guest"

lennardk:
# Pull Request

## What? (description)
This changes the osType field in the ova to `other3xLinux64Guest` in order to have VMware accept it when deploying manually.

## Why? (reasoning)
To be able to deploy the OVA directly into the vSphere HTML UI. See #3515.

## Acceptance

### DIFF
--- a/cmd/installer/pkg/ova/ova.go
+++ b/cmd/installer/pkg/ova/ova.go
@@ -48,7 +48,7 @@ const ovfTpl = `<?xml version="1.0" encoding="UTF-8"?>
   <VirtualSystem ovf:id="vm">
     <Info>A virtual machine</Info>
     <Name>talos</Name>
-    <OperatingSystemSection ovf:id="101" vmw:osType="otherLinux64Guest">
+    <OperatingSystemSection ovf:id="101" vmw:osType="other3xLinux64Guest">
       <Info>The kind of installed guest operating system</Info>
     </OperatingSystemSection>
     <VirtualHardwareSection ovf:transport="com.vmware.guestInfo">


smira (approved):

cmd/installer/pkg/ova/ova.go
```diff
@@ -48,7 +48,7 @@ const ovfTpl = `<?xml version="1.0" encoding="UTF-8"?>
   <VirtualSystem ovf:id="vm">
     <Info>A virtual machine</Info>
     <Name>talos</Name>
-    <OperatingSystemSection ovf:id="101" vmw:osType="otherLinux64Guest">
+    <OperatingSystemSection ovf:id="101" vmw:osType="other3xLinux64Guest">
```
alex1989hu, line 51: Hello, I would like to give you a feedback.

There is an option to use `other5xLinux64Guest` which is the actual version of kernel in Talos*.

* https://github.com/vmware/govmomi/blob/9054158bb458f8c383422253709b56dc318ba2c0/vim25/types/enum.go#L4459
* https://github.com/vmware/open-vm-tools/blob/42ce53f39be45b7795a9f1adf892af84629b4a19/open-vm-tools/lib/include/guest_os_tables.h#L332

Probably we are lucky, we never met this issue you described but I can imagine the edge case you met. If you say `other3xLinux64Guest` solves your issue I could agree that.

Note for `other5xLinux64Guest`:
To use a newer version, for instance `other5xLinux64Guest`, requires never vSphere version. Probably newer virtual HW version is also needed; I am not sure - it shall be tested.

On the other hand, I couldn't find any minimum vSphere version reference in Talos documentation.

From VMware point of view, I was able to find a documentation (but not the latest one, 5.x kernel option is not listed) <https://vdc-download.vmware.com/vmwb-repository/dcr-public/da47f910-60ac-438b-8b9b-6122f4d14524/16b7274a-bf8b-4b4c-a05e-746f2aa93c8c/doc/vim.vm.GuestOsDescriptor.GuestOsIdentifier.html>.

- It says `other3xLinux64Guest` is available since vSphere API 5.5 and `other4xLinux64Guest ` is available since vSphere API 6.7. So the question is what Talos shall support or not.

VMware support dates are available here <https://kb.vmware.com/s/article/66977>.

/cc @smira

lennardk: I've played around with the options a bit on my vSphere 6.7 installation, although I've limited myself to using the vSphere html UI and checking if the OVA itself deploys. I've tested using `other3xLinux64Guest`, `other4xLinux64Guest`, `other5xLinux64Guest` and for good measure the non-existing `other9xLinux64Guest` as well.

Deployment succeeds using all these options. `other3xLinux64Guest` is recognised as `Linux 3.x Kernel (64 bit)`. `other4xLinux64Guest` is recognised as `Linux 4.x Kernel (64 bit)`. The other 2 options don't seem toch be recognised at all and are listed as `Other (32 bit)`.

Virtual HW version doesn't seem to make a technical difference, but when manually configuring the OS version the available  options are influenced by it. When using version 13 as currently in the OVA `Linux 4.x Kernel (64 bit)` isn't available for manual selection. Upgrading to version 14 or 15, `Linux 4.x Kernel (64 bit)` also becomes listed. I don't have a vSphere 7 available so I don't have any higher levels.

All things considered, I don't have too strong an opinion on which value to use. I don't think there's much of a technical or performance difference. Considering the fact that vSphere 6.5 is in support until october 2022, I'd lean toward sticking to harware compatibility version 13 and `other3xLinux64Guest`, rather than upgrading to version 14+ and `other4xLinux64Guest`. I'm pretty sure the later would mean deployment on vSphere 6.5 would fail. Keeping the compatibility level at 13 and going for `other4xLinux64Guest` or `other5xLinux64Guest` probably works, but could cause confusion at least when using the HTML UI.

smira: I would say keep it the lowest supported on VMWare side unless there's any performance benefit in going with `other5xLinux64Guest`

PR #3518, opened 2021-04-20
docs: update AWS instructions

### LINKED ISSUE #3495: Update AWS docs
Feedback from Steve during his usage:

```
AWS doc improvements	
Move the “Official AMI Images” section from the very top. It’s not a very friendly thing to start the page with just a list of AMIs. I’d suggest creating a “Select an AMI” section in the menu structure, in the same place as “Create the AMI”, and have that contain text: “You can use one of the public AMI images that Talos maintains, or create your own”, then two subsections, “Official AMI images”, and “Create your AMI (optional)”

When you define the security group policies, you don’t need to include both the group-name and the group-id. So you can simplify it from :
aws ec2 authorize-security-group-ingress \
  --region $REGION \
  --group-name talos-aws-tutorial-sg \
  --protocol all \
  --port 0 \
  --group-id $SECURITY_GROUP \
  --source-group $SECURITY_GROUP
To 
aws ec2 authorize-security-group-ingress \
  --region $REGION \
  --group-name talos-aws-tutorial-sg \
  --protocol all \
  --port 0 \
  --source-group $SECURITY_GROUP

Same for the other two ws ec2 authorize-security-group-ingress commands - omit the —group-id

In v0.9 and 0.8 (at least) you need to document how to reduce the length of the yaml files so they are not too long to send as user data. My inelegant way to strip comments:
 cat init.yaml | sed 's/ #.//' > temp.yaml; mv temp.yaml init.yaml
cat controlplane.yaml | sed 's/ #.//' > temp.yaml; mv temp.yaml controlplane.yaml

On both the command to create the bootstrap node and the remaining control plane nodes, you need to add
 --associate-public-ip-address

When configuring the load balancer, you have to specify --target-type ip on the aws elbv2 create-target-group command, otherwise it expects to associate nodes by AWS id, not IP address, and the register targets command will fail.
```

rsmitty:
This PR adds some docs changes where we had incorrect commands in the
AWS docs.

Will close #3495

### DIFF
--- a/website/content/docs/v0.8/Cloud Platforms/aws.md
+++ b/website/content/docs/v0.8/Cloud Platforms/aws.md
@@ -99,7 +99,6 @@ aws ec2 authorize-security-group-ingress \
     --group-name talos-aws-tutorial-sg \
     --protocol all \
     --port 0 \
-    --group-id $SECURITY_GROUP \
     --source-group $SECURITY_GROUP
 ```
 
@@ -111,15 +110,14 @@ aws ec2 authorize-security-group-ingress \
     --group-name talos-aws-tutorial-sg \
     --protocol tcp \
     --port 6443 \
-    --cidr 0.0.0.0/0 \
-    --group-id $SECURITY_GROUP
+    --cidr 0.0.0.0/0
+
 aws ec2 authorize-security-group-ingress \
     --region $REGION \
     --group-name talos-aws-tutorial-sg \
     --protocol tcp \
     --port 50000-50001 \
-    --cidr 0.0.0.0/0 \
-    --group-id $SECURITY_GROUP
+    --cidr 0.0.0.0/0
 ```
 
 ### Create a Load Balancer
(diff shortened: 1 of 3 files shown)


Unix4ever (approved):

website/content/docs/v0.8/Cloud Platforms/aws.md
```diff
@@ -148,6 +146,15 @@ created join.yaml
 created talosconfig
 ```
 
+Take note that in this version of Talos, the generated configs are too long for AWS userdata field.
```
smira, line 149: idk if we need that, but since 0.10 we support `talosctl gen config --with-docs=false --with-examples=false`

(should be on 0.10 doc, but I couldn't find any good place for this comment)

smira (dismissed):

website/content/docs/v0.8/Cloud Platforms/aws.md
rsmitty: yeah, I at least made a note as to why those flags were required.

smira (approved):

Issue #3018, opened 2021-01-07
talosctl list --recurse / --depth handling is confusing


AlekSi:
### Description

On 0bd8b0e8008c12e4914c6e9b5faf06dda6c744f7, `talosctl list`'s `--recurse` / `--depth` handling is confusing:

| --recurse       | --depth     | Result     |
| --------------- | ----------- | ---------- |
| false (default) | -1          | one level  |
| false (default) | 0 (default) | one level  |
| false (default) | 1           | one level  |
| false (default) | 2           | one level  |
| true            | -1          | all levels |
| true            | 0 (default) | all levels |
| true            | 1           | one level  |
| true            | 2           | two levels |

Not sure what would be the best fix. Some of the options I see:
1. "Just fix it". But then `talosctl list --recurse=false --depth=0 /etc` should return nothing? Or `--depth` should be ignored if `--recurse` is false? Not sure if that's a good UX.
2. Remove `--recurse` flag (and API field) completely, treat `--depth=0` (or negative) as unlimited depth, and change the default value to 1.

smira:
cc @Unix4ever

PR #3519, opened 2021-04-21
test: better `talosctl ls` tests

AlekSi:
Refs #3018.

Actual fixes will follow.

### DIFF
--- a/hack/test/e2e-docker.sh
+++ b/hack/test/e2e-docker.sh
@@ -29,7 +29,7 @@ function create_cluster {
 }
 
 function destroy_cluster() {
-  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}"
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}" --provisioner "${PROVISIONER}"
 }
 
 create_cluster
@@ -37,3 +37,6 @@ get_kubeconfig
 ${KUBECTL} config set-cluster e2e-docker --server https://10.5.0.2:6443
 run_talos_integration_test_docker
 run_kubernetes_integration_test
+
+# Unlike other local e2e tests, we don't destroy the cluster there as it is used by CAPI and AWS/GCP e2e tests later.
+# destroy_cluster
--- a/internal/integration/base/run.go
+++ b/internal/integration/base/run.go
@@ -44,13 +44,6 @@ func ShouldFail() RunOption {
 	}
 }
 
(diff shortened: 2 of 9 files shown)


internal/integration/base/run.go
```diff
@@ -44,13 +44,6 @@ func ShouldFail() RunOption {
 	}
 }
 
-// ShouldSucceed tells Run command should succeed (that is default).
-func ShouldSucceed() RunOption {
-	return func(opts *runOptions) {
-		opts.shouldFail = true
```
AlekSi, line 50: Not used, and there is a typo (it should be `false`)

hack/test/e2e-docker.sh
```diff
@@ -29,11 +29,12 @@ function create_cluster {
...
-  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}"
+  "${TALOSCTL}" cluster destroy --name "${CLUSTER_NAME}" --provisioner "${PROVISIONER}"
 }
 
 create_cluster
 get_kubeconfig
 ${KUBECTL} config set-cluster e2e-docker --server https://10.5.0.2:6443
 run_talos_integration_test_docker
 run_kubernetes_integration_test
+destroy_cluster
```
smira, line 40: I think this is wrong, as I believe we rely on docker cluster to be up in CAPI tests - e2e-docker creates a cluster, e2e-capi installs CAPI, e2e-aws/gcp use that cluster to provision talos in the cloud

Unix4ever (dismissed):

internal/integration/cli/list.go
```diff
@@ -31,6 +33,79 @@ func (suite *ListSuite) TestSuccess() {
...
 }
 
+// TestDepth tests various combinations of --recurse and --depth flags.
+func (suite *ListSuite) TestDepth() {
+	suite.T().Parallel()
+
+	node := suite.RandomDiscoveredNode(machine.TypeJoin)
+
+	runAndParse := func(flags ...string) []string {
+		args := append([]string{"list", "--nodes", node, "/etc/kubernetes"}, flags...)
```
smira, line 43: my concern with `/etc/kubernetes` is that it's managed by the kubelet, and its contents might change with `kubelet` version bumps or might change after some interaction with the kubelet

not sure which other directory is better, but my guess `/system/secrets` might be better, as at least it's managed by Talos and its contents are fixed (for control plane nodes only)

smira (approved):
Nice!

PR #3526, opened 2021-04-23
docs: add Matrix room links

Ulexus:
Add links to the Talos Matrix rooms (which are themselves linked to the
corresponding Slack channels).


andrewrynhard (approved):

smira (approved):
might to need to add it for 0.11 as well!

smira (approved):

PR #3539, opened 2021-04-26
feat: add `--config-patch` flag by node type

smira:
The problem is that some patches can't be applied to join config, as
some nodes don't even exist in the config, for example
`/cluster/apiServer` node, and applying such patches doesn't make any
sense.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -105,6 +105,8 @@ var (
 	encryptEphemeralPartition bool
 	useVIP                    bool
 	configPatch               string
+	configPatchControlPlane   string
+	configPatchJoin           string
 )
 
 // createCmd represents the cluster up command.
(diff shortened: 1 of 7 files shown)


Unix4ever (approved):

smira:
example why this is required:

```bash
$ talosctl gen config foo https://localhost:6443 --kubernetes-version=v1.21.0 --config-patch-control-plane '[{"op": "add", "path": "/cluster/apiServer/extraArgs", "value": {"feature-gates": "ServerSideApply=true"}}, {"op": "add", "path": "/cluster/apiServer/certSANs/-", "value": "10.10.5.200"}]'
```

this generates:

```yaml
   apiServer:
        image: k8s.gcr.io/kube-apiserver:v1.21.0 # The container image used in the API server manifest.
        # Extra arguments to supply to the API server.
        extraArgs:
            feature-gates: ServerSideApply=true
        # Extra certificate subject alternative names for the API server's certificate.
        certSANs:
            - localhost
            - 10.10.5.200
```

with plain `--config-patch` it fails, as join config doesn't have `apiServer` section, and we don't want to create it either to preserve defaults (`image` for example)

PR #3540, opened 2021-04-26
chore: remove security API ReadFile/WriteFile

smira:
This seems to be unused completely, and they look scary enough at the
same time.

For better readability and to avoid any pitfalls, better to remove them.


andrewrynhard (approved):
We should back port this to 0.9?

smira:
> We should back port this to 0.9?

I don't see exact reason to backport? these APIs haven't been used for quite some time. or are you thinking about security risk?

andrewrynhard:
> > We should back port this to 0.9?
> 
> I don't see exact reason to backport? these APIs haven't been used for quite some time. or are you thinking about security risk?

Yep, exactly. Just to feel better about things.

Issue #3562, opened 2021-04-29
Add the ability to customize static pod manifests


DWSR:
## Feature Request

Add the ability to customize static pod manifests

### Description

As a user, I'd like the ability to customize the static pod manifests in order to better tailor Talos to my use case. For example, I'd like to be able to set pod requests and limits on control plane components in order to help support development clusters where workloads are running on the same nodes as the control plane components.

smira:
Requests/limits are configurable now, as many other aspects of the static pods.

PR #3575, opened 2021-05-05
feat: static pods improvements

sergelogvinov:
Static pods changes 

- switch livenessProbe to ssl port
- add request resources

# Pull Request

## What? (description)

Cannot disable insecure ports, livenessProbe will fail

## Why? (reasoning)

By security reason, we can close ports 10251, 10252 (--port=0 extraArgs param)
Add requests resources to static pods (#3562).
 
## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
 	v1 "k8s.io/api/core/v1"
+	apiresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
@@ -261,6 +262,12 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 								ReadOnly:  true,
 							},
 						}, volumeMounts(cfg.ExtraVolumes)...),
+						Resources: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								v1.ResourceCPU:    apiresource.MustParse("200m"),
+								v1.ResourceMemory: apiresource.MustParse("512Mi"),
+							},
+						},
 					},
 				},
 				HostNetwork: true,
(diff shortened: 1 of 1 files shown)


sergelogvinov:
I'm not sure about **go-runner**
Should we use it here? Containerd collects logs well.
Maybe we need to remove it.

rsmitty:
@sergelogvinov thanks for this PR. I'm going to approve so we can run it through CI and make sure it works.

RE: the go-runner stuff - that's a requirement from the upstream K8s images, nothing we can do there.

I'm also going to rope in @smira and @andrewrynhard on this to see how we feel about resource requests for these components. It _feels_ like this could lead to a situation where the API server (for example) won't get recreated if it were to die and the control plane node was resource constrained.

smira:
I would agree with @rsmitty that we need to verify resource requests, but at the same time as these are static pods, they bypass `kube-scheduler` completely, so they should run no matter what, while resource requests would allow better scheduling of other pods on the same node. 

I wonder if we should still keep things configurable?

rsmitty:
Oh, if they bypass scheduler, I suppose you're right that these could be helpful for _other_ pods getting scheduled. I'd be fine with keeping these requests as they are if @sergelogvinov can point us to where he pulled these numbers from?

sergelogvinov:
https://pkg.go.dev/k8s.io/kubernetes/build/go-runner they use go-runner to collect logs (stdout/stderr) to the file.
I've run the control-plane without go-runner in daemonset/pod-static/systemd modes. All works fine.

it very convenient to split by types the stdin/stderr logs in logs-alert-systems (docker/containerd do it).

I've got the resources from clean cluster (2 talos nodes) and 50+ nodes cluster (max of 2 weeks history). The difference has only in kube-apiserver pods. I've set this resource only to make possible to run control-panel and other pods (coredns) on 1cpu/2gb instances (arm/raspberry customers).

This resource requests needs for
* reports like - kubectl describe node ... to see current node capacity 
* metrics/prometheus to predict the failure (pod got more request resources too long)
* help kube-scheduler to run pods on different nodes.

I think in future kubelet also needs to set a system-reserved resources to protect etcd-cluster https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources

PS. I am not recommend to set resources limits for the control-plane.

smira (dismissed):
:+1:  on the change itself

one nit: update the commit message to be more specific, as it goes to the changelog, e.g. `feat: add resources and use HTTPS checks in control plane pods`

sergelogvinov:
Can you help me with that

```
# git rev-list upstream/master..origin/static_pod_probs
183c78d8cba236f1e765ee9c35461c95b334a32f
5153e289a656ba0df8cd4345cf2c52b5c0261447
```
but I got 
```
# make conformance
...
commit         Number of Commits            FAILED        HEAD is 2393 commit(s) ahead of refs/heads/master
```

smira:
> Can you help me with that
> 
> ```
> # git rev-list upstream/master..origin/static_pod_probs
> 183c78d8cba236f1e765ee9c35461c95b334a32f
> 5153e289a656ba0df8cd4345cf2c52b5c0261447
> ```
> but I got 
> ```
> # make conformance
> ...
> commit         Number of Commits            FAILED        HEAD is 2393 commit(s) ahead of refs/heads/master
> ```

We require commits to be rebased against the master branch, so the easiest is to pull latest master and do something like git rebase master on your local branch followed by force push to your fork. 

I can also rebase via Talos bot

smira:
Should be good now

smira (approved):

PR #3585, opened 2021-05-06
fix: update go-blockdevice to fix disk type detection

Unix4ever:
Otherwise it never detected mmc and nvme devices as such.


smira (commented):
Looking at the diff, do we need to update same for pkg/machinery/go.mod

Unix4ever:
> Looking at the diff, do we need to update same for pkg/machinery/go.mod

Done!

smira:
My guess it can be go mod tidied now ) start with machinery and move to the top level :)

PR #3608, opened 2021-05-12
feat: implement AddressSpec handling

smira:
This includes multiple controllers responsible for different stages of
`AddressSpec` conversion:

* `AddressConfigController` produces initial unmerged configuration from
multiple sources (more sources coming later, e.g. DHCP)
* `AddressMergeController` merges address configuration into final
representation
* `AddressSpecController` syncs resources with kernel state


smira:
notes:

* "new" `networkd` runs side by side with "old" `networkd`
*  it is possible, as at the moment there are no conflicts - address assignments should be done exactly same, and new networkd doesn't touch addresses it doesn't own
* the most fun demo is adding static addresses to the interfaces - they get assigned immediately

pkg/resources/network/network.go
```diff
@@ -15,7 +15,17 @@ import (
 // NamespaceName contains resources related to networking.
 const NamespaceName resource.Namespace = "network"
 
+// ConfigNamespaceName contains umerged resources related to networking generate from the configuration.
+//
+// Resources in the ConfigNamespaceName namespace are merged to produce final versions in the NamespaceName namespace.
+const ConfigNamespaceName resource.Namespace = "network-config"
```
andrewrynhard, line 21: Just an idea: We could have a `configuration` namespace that is common for any resource type that is configurable. Then something like `talosctl get all -n configuration` would be a dump of all user-supplied things.

internal/app/machined/pkg/controllers/network/address_merge_test.go
```diff
@@ -0,0 +1,196 @@
...
+		if res.Metadata().ID() == id {
+			return retry.ExpectedError(fmt.Errorf("address %q is still there", id))
+		}
+	}
+
+	return nil
+}
+
+func (suite *AddressMergeSuite) TestMerge() {
+	loopback := network.NewAddressSpec(network.ConfigNamespaceName, "default/lo/127.0.0.1/8")
```
andrewrynhard, line 117: I don't know where the source for this is, so going to comment here. The only thing I don't like about the ID is that it takes a little bit more brain power to realize that `127.0.0.1/8` is a CIDR, because `/` is used to delimit `default` and `lo`. I also can't think of anything better.

andrewrynhard (approved):

pkg/resources/network/network.go
smira: I'll put this as an issue, as it's easy to rename (just a constant here), and we can come up with good names for `unmerged` configuration and `merged` configuration namespaces, not only for the network.

internal/app/machined/pkg/controllers/network/address_merge_test.go
smira: yep, I looked into `:`, and that didn't look good. 

good thing - it's just an ID, so it's never parsed/consumed in any way. we can make it anything with a simple change

PR #3619, opened 2021-05-14
fix: stop networkd and pods before leaving etcd on upgrade

smira:
The change is essentially same as #3590, but applied to the upgrade path
which is very similar to the reset path.

We have to stop networkd (and remove the VIP/lease on the VIP) before we
leave and stop etcd. Plus we stop the kube-apiserver before the etcd is
stopped, so that we don't have unhealthy kube-apiserver.


andrewrynhard (approved):

smira:
tested the PR manually on the cluster with VIP, and it seems to be stable now

PR #3639, opened 2021-05-18
feat: implement route network controllers

smira:
Route handling is very similar to addresses:

* `RouteStatus` describes kernel routing table state,
`RouteStatusController` reflects kernel state into resources
* `RouteSpec` defines routes to be configured
* `RouteConfigController` creates `RouteSpec`s based on cmdline and
machine configuration
* `RouteMergeController` merges different configuration layers into the
final representation
* `RouteSpecController` applies the specs to the kernel routing table


Unix4ever (approved):
Another `zap.Logging` PR merge conflict :)

smira:
> Another `zap.Logging` PR merge conflict :)

I can hold on merging before your PR gets merged ;)

andrewrynhard (commented):
Did not look into the details, but high-level LGTM

rsmitty (dismissed):

internal/app/machined/pkg/controllers/network/address_config.go
```diff
@@ -59,7 +56,7 @@ func (ctrl *AddressConfigController) Outputs() []controller.Output {
 
 // Run implements controller.Controller interface.
 //
-//nolint: gocyclo, cyclop
+//nolint:gocyclo, cyclop
```
AlekSi, line 59:
```suggestion
//nolint:gocyclo,cyclop
```

😉

internal/app/machined/pkg/controllers/network/cmdline.go
```diff
@@ -0,0 +1,122 @@
...
+	Hostname     string
+	LinkName     string
+	DNSAddresses []netaddr.IP
+	NTPAddresses []netaddr.IP
+}
+
+// ParseCmdlineNetwork parses `ip=` kernel cmdline argument producing all the available configuration options.
+//
+//nolint:gocyclo,cyclop
+func ParseCmdlineNetwork(cmdline *procfs.Cmdline) (settings CmdlineNetworking, err error) {
```
AlekSi, line 31: No sure if named results are worth it there.
Same for returning a value instead of a pointer.

internal/app/machined/pkg/controllers/network/cmdline.go
```diff
@@ -0,0 +1,122 @@
...
+			case 9:
+				var ntpIP netaddr.IP
+
+				ntpIP, err = netaddr.ParseIP(fields[i])
+				if err != nil {
+					return settings, fmt.Errorf("error parsing DNS IP: %w", err)
+				}
+
+				settings.NTPAddresses = append(settings.NTPAddresses, ntpIP)
+			}
```
AlekSi, line 100: What about `default:` that would return an error?

internal/app/machined/pkg/controllers/network/route_config.go
```diff
@@ -0,0 +1,297 @@
...
+		route.Table = nethelpers.TableMain
+		route.Protocol = nethelpers.ProtocolStatic
+		route.OutLinkName = linkName
+		route.Layer = network.ConfigMachineConfiguration
+
+		route.Scope = nethelpers.ScopeGlobal
+
+		if route.Destination.IP.IsLinkLocalUnicast() || route.Destination.IP.IsLinkLocalMulticast() {
+			route.Scope = nethelpers.ScopeLink
+		} else if route.Destination.IP.IsLoopback() {
```
AlekSi, line 251: In general, I think `switch` is often more clear than `else if`:

```go
switch {
case route.Destination.IP.IsLinkLocalUnicast():
  fallthrough
case route.Destination.IP.IsLinkLocalMulticast():
  route.Scope = nethelpers.ScopeLink
case route.Destination.IP.IsLoopback():
  route.Scope = nethelpers.ScopeHost
}
```

internal/app/machined/pkg/controllers/network/cmdline.go
andrewrynhard: I personally like named results in more complex functions, just my two cents.

internal/app/machined/pkg/controllers/network/route_config.go
andrewrynhard: +1, not a blocker for me.

internal/app/machined/pkg/controllers/network/cmdline.go
smira: I don't think returning a pointer is every worth it in general, unless we're returning something huge, as copying is more efficient than passing via the heap (which might be optimized by the Go compiler, but who knows for every invocation?).

As for the named results, I can look into that

internal/app/machined/pkg/controllers/network/route_config.go
smira: I don't like `fallthrough` unless it's really needed, so this boils down just to two clauses (first one with `||`), I don't know if it's worth it here? I fully agree that 3 and more are better as `switch`

internal/app/machined/pkg/controllers/network/cmdline.go
smira: I'm not sure what do you mean by that? `default:` is not valid, but empty value defaults to whatever DHCP provides (we don't need it here, as we only need static part)

internal/app/machined/pkg/controllers/network/route_config.go
andrewrynhard: I don't have a strong opinion either way

internal/app/machined/pkg/controllers/network/cmdline.go
smira: got it, we ignore extra params (as probably kernel does), so we protect ourselves from something which is parsed by the kernel, but not us

internal/app/machined/pkg/controllers/network/route_config.go
smira: I ended up refactoring it to use switch, with `default:` case it goes to 3 clauses!

AlekSi (approved):

PR #3650, opened 2021-05-23
feat: extract config generation

borancar:
Extract config generation into a seperate function. This reduces
cyclomatic complexity with the added benefit of being able to use
generation in another tool (e.g. Terraform provider to generate config).

Signed-off-by: Boran Car <boran.car@gmail.com>

# Pull Request

## What? (description)

Extract the config bundle generation into a separate function that can be invoked externally.

## Why? (reasoning)

Primarily this is to enable creating a terraform provider so that talos generation can be easily rolled into typical terraform flows, without needing to stage it. The added benefit here is that it also reduces cyclomatic complexity of the original function. Still not enough though to remove the linting exception (cyclo is now at 15, whereas needs to be under 10).

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/mgmt/config.go
+++ b/cmd/talosctl/cmd/mgmt/config.go
@@ -21,6 +21,7 @@ import (
 	"github.com/talos-systems/talos/pkg/images"
 	"github.com/talos-systems/talos/pkg/machinery/config"
 	"github.com/talos-systems/talos/pkg/machinery/config/encoder"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/bundle"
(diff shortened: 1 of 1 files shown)


andrewrynhard:
Great idea. Thanks @borancar !

smira (approved):
lgtm for the refactoring!

I thought if we should move `GenV1alpha1Config` function to machinery package to improve usability from other tools (as `pkg/machinery` is the public importable sub-module)

borancar:
> I thought if we should move GenV1alpha1Config function to machinery package to improve usability from other tools (as pkg/machinery is the public importable sub-module)

I'm game either. Any point in also refactoring the parameters into GenOptions then? Well, at least some of them like the JSON patch overrides?

smira:
discussed during the meeting: we're going to merge this PR and keep iterating on the approach:

* move into `pkg/machinery`
* clean up input arguments moving more to `GenOptions`, etc.

PR #3666, opened 2021-05-24
chore: refactor gen subcommands

AlekSi:
Move them to own package.


cmd/talosctl/cmd/mgmt/gen/csr.go
```diff
@@ -0,0 +1,81 @@
...
+		opts := []x509.Option{}
+
+		parsed := net.ParseIP(csrIP)
+		if parsed == nil {
+			return fmt.Errorf("invalid IP: %s", csrIP)
+		}
+
+		ips := []net.IP{parsed}
+		opts = append(opts, x509.IPAddresses(ips))
+		opts = append(opts, x509.NotAfter(time.Now().Add(time.Duration(crtHours)*time.Hour))) // BUG
```
AlekSi, line 59: `crtHours` is always zero for `csr` command. Should this line be dropped, or there should be flag for that property?

smira: I think these commands are largely not really supported atm

AlekSi: Well, they are documented: https://www.talos.dev/docs/v0.11/guides/managing-pki/

I will drop that line then.

smira: I know they are documented, but they're not really sitting in the regular usage pattern

AlekSi:
Will re-create

PR #3680, opened 2021-05-26
feat: add ability to mark MBR partition bootable

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3532

Machine install section now has `markMBRBootable` option.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
 	github.com/talos-systems/crypto v0.2.1-0.20210526123943-7776057f5086
-	github.com/talos-systems/go-blockdevice v0.2.1-0.20210510233948-1292574643e0
+	github.com/talos-systems/go-blockdevice v0.2.1-0.20210526155905-30c2bc3cb62a
 	github.com/talos-systems/go-cmd v0.0.0-20210216164758-68eb0067e0f0
 	github.com/talos-systems/go-debug v0.2.1-0.20210525175311-3d0a6e1bf5e3
 	github.com/talos-systems/go-kmsg v0.1.0
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
@@ -841,6 +841,10 @@ type InstallConfig struct {
 	//     - false
 	//     - no
 	InstallWipe bool `yaml:"wipe"`
+	//   description: |
+	//     Indicates if MBR partition should be marked as bootable (active).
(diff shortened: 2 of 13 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -841,6 +841,9 @@ type InstallConfig struct {
 	//     - false
 	//     - no
 	InstallWipe bool `yaml:"wipe"`
+	//   description: |
+	//     Indicates if MBR partition should be marked as bootable (active).
+	InstallMarkMBRBootable bool `yaml:"markMBRBootable,omitempty"`
```
andrewrynhard, line 846: Since this is user facing, I don't know if this is the best name for it. I also don't have a better idea. I would need to look into the nomenclature/wording that is used within the industry. Maybe @Ulexus has thoughts?

Unix4ever: sure, I'm open for suggestions :)

Unix4ever: Yeah, and yaml tag can be `legacyBIOSSupport` as well.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/talos-systems/talos
 go 1.16
 
 replace (
+	github.com/talos-systems/go-blockdevice => github.com/Unix4ever/go-blockdevice v0.1.1-0.20210526144506-0573ad3d7b15
```
smira, line 6: should this be removed in favor of merged PR?

andrewrynhard (approved):
Do we need release notes?

internal/app/machined/internal/install/install.go
```diff
@@ -94,6 +95,7 @@ func RunInstallerContainer(disk, platform, ref string, configBytes []byte, reg c
 		"--upgrade=" + upgrade,
 		"--force=" + force,
 		"--zero=" + zero,
+		"--legacy-bios-support=" + legacyBIOSSupport,
```
smira, line 98: this is super tricky: as old installer doesn't support this flag, it would fail on this option, so e.g. you can't downgrade Talos. or you can't install 0.10 on 0.11 PXE boot environment.

so I think we shouldn't be adding any more flags here.

btw, we now have machine configuration passed in to the installer via stdin, why not use it directly?

go.mod
Unix4ever: oh, right. I though I did that already

internal/app/machined/internal/install/install.go
Unix4ever: I missed that. Great idea, updated already to read from the config. Will verify and push

smira (approved):

PR #3688, opened 2021-05-27
fix: only fetch machine uuid if it's not set

### LINKED ISSUE #3676: config is valid at installation but invalid at runtime when disabling config persistence
## Bug Report
When using the following config:

```yaml
cluster:
  ca:
    crt: <crt>
    key: ""
  controlPlane:
    endpoint: <endpoint>
  network:
    dnsDomain: cluster.local
    podSubnets:
    - 10.244.0.0/16
    serviceSubnets:
    - 10.96.0.0/12
  token: <token>
debug: false
machine:
  install:
    disk: /dev/mmcblk0
  kubelet:
    extraArgs:
      node-ip: 192.168.2.210
      node-labels: metal.sidero.dev/uuid=00c03114-0000-0000-0000-e45f011f4f30
  network:
    interfaces:
    - cidr: 192.168.2.210/24
      interface: wg0
      wireguard:
        peers:
        - allowedIPs:
          - 0.0.0.0/0
          endpoint: 83.84.110.84:51820
          publicKey: <key>
        privateKey: <key>
    - dhcp: true
      interface: eht0
  token: <token>
  type: worker
persist: false
version: v1alpha1
```

The installation of Talos succeeds, however, after rebooting into Talos and the config is re-fetched due to `persist:  false` the config validation fails and the nodes ends up in a boot loop.

However, when I change persist to `True`  and re-install Talos and reboot to Talos, it works as expected.

### Logs
Looking for the error msg in the code, I've found  https://github.com/talos-systems/go-smbios/blob/d3a32bea731a0c2a60ce7f5eae60253300ef27e1/smbios/smbios.go#L60 which is a perfect match. 
`failed to decode structures: unexpected EOF`
  
### Environment
- booting via Sidero
- Talos version: v0.10.3
- Platform: metal rpi_4

### Kernel args

```bash
kernel:
      args:
      - init_on_alloc=1
      - initrd=initramfs.xz
      - ip=dhcp
      - slab_nomerge
      - pti=on
      - talos.board=rpi_4
      - talos.shutdown=poweroff
      - talos.config=http://192.168.1.200:30005/configdata?uuid=
      - talos.platform=metal
```

OGKevin:
# Pull Request

## What? (description)
During boot sequence, if `talos.config`'s url has the uuid parameter, the uuid
value is retrieved via SMBIOS. However, at this part of the code it can happen
that the uuid is already set and valid. If this is the case, instead of
re-fetching the uuid, the one that is already set can be used.

## Why? (reasoning)
The initial issue, why the call to SMBIOS fails has not been solved, however, it does not make sense to override a value that has already been set. This PR adds a check to only fetch uuid from SMBIOS when there is no value set.

## Acceptance

closes #3676

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"path/filepath"
 
+	"github.com/google/uuid"
 	"github.com/talos-systems/go-blockdevice/blockdevice/filesystem"
 	"github.com/talos-systems/go-blockdevice/blockdevice/probe"
 	"github.com/talos-systems/go-procfs/procfs"
@@ -48,26 +49,27 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 
 	u, err := url.Parse(*option)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to parse %s: %w", constants.KernelParamConfig, err)
 	}
 
 	values := u.Query()
 
 	if len(values) > 0 {
-		for key := range values {
+		for key, qValues := range values {
 			switch key {
 			case "uuid":
-				s, err := smbios.New()
-				if err != nil {
-					return nil, err
+				if len(qValues) != 1 {
+					uid, err := getSystemUUID()
+					if err != nil {
+						return nil, err
+					}
+
+					values.Set("uuid", uid.String())
+
+					break
 				}
 
-				uuid, err := s.SystemInformation().UUID()
-				if err != nil {
-					return nil, err
-				}
-
-				values.Set("uuid", uuid.String())
+				values.Set("uuid", qValues[0])
 			default:
 				log.Printf("unsupported query parameter: %q", key)
 			}
@@ -84,6 +86,15 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 	default:
 		return download.Download(ctx, *option)
 	}
+}
+
+func getSystemUUID() (uuid.UUID, error) {
+	s, err := smbios.New()
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	return s.SystemInformation().UUID()
 }
 
 // Hostname implements the platform.Platform interface.


internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -48,26 +49,25 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 
 	u, err := url.Parse(*option)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to parse talos.config url: %w", err)
```
andrewrynhard, line 52: We have a constant somewhere for talos.config. We should probably use that just in case we ever need to change it.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -48,26 +49,25 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
 			case "uuid":
-				s, err := smbios.New()
-				if err != nil {
-					return nil, err
-				}
-
-				uuid, err := s.SystemInformation().UUID()
-				if err != nil {
-					return nil, err
+				if len(qValues) != 1 {
```
andrewrynhard, line 61: I'm not sure I understand why we need to check the length here.

OGKevin: In case the url param has no value, e.g. `?uuid=` then `qvalues` will be 0. If this is the case, then we shall fetch the uuid. In case the URL param is `?uuid=<value>` then `qvalues` will be 1. In case the `uuid` param is repeated, instead of using the first one, we just fetch.

andrewrynhard: Ah. Got it. I guess I am so caught up on why the lookup fails. This works, but I'd rather figure out why the uuid lookup fails. That being said I won't block this.

OGKevin: Yea it's hard to debug without being able to attach a remote debugger tho. That might be an interesting project, being able to attach a remote debugger if `debug` is enabled.

smira (approved):

PR #3702, opened 2021-05-31
chore: skip building multi-arch installer for race-enabled build

smira:
Go cross-compilation is not quite compatible with Go race detector, as C
toolchain is not cross-compliation ready.

Workaround is really simple: for race-enabled builds, don't build
multi-arch installer image (that is installer image which contains both
amd64 and arm64 Talos artifacts), but build installer artifacts only for
the target arch (skipping cross-compilation).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ ARG TOOLS
 ARG IMPORTVET
 ARG PKGS
 ARG EXTRAS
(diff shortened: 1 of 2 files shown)


AlekSi (approved):
🥳

Unix4ever (approved):

PR #3706, opened 2021-06-01
feat: implement CPU and Memory stats controller

Unix4ever:
Now the latest value for CPU and Memory is also represented as COSI
resources.

Was going back and forth in the implementation but in the end decided to
use dedicated yaml structures for both CPU and Memory stats because:

- JSON tags are ignored by `go-yaml`, so the output is not really great.
- protobuf Talos definition contains fields which we don't really need
in the YAML output of `talosctl get`.
- current state of Talos resource service does not support protobuf
encoding for resources.

So the plan for Theila is to just use the structure as a dynamic object
without relying on protobufs. At least for now.

### DIFF
--- a/internal/app/machined/pkg/controllers/perf/perf.go
+++ b/internal/app/machined/pkg/controllers/perf/perf.go
@@ -1,0 +1,110 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package perf
+
+import (
+	"context"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/prometheus/procfs"
+	"go.uber.org/zap"
+
+	"github.com/talos-systems/talos/pkg/resources/perf"
+)
+
+const updateInterval = time.Second * 30
+
+// StatsController manages v1alpha1.Stats which is the current snaphot of the machine CPU and Memory consumption.
+type StatsController struct{}
+
+// Name implements controller.StatsController interface.
+func (ctrl *StatsController) Name() string {
+	return "perf.StatsController"
+}
+
+// Inputs implements controller.StatsController interface.
+func (ctrl *StatsController) Inputs() []controller.Input {
+	return nil
+}
+
+// Outputs implements controller.StatsController interface.
+func (ctrl *StatsController) Outputs() []controller.Output {
+	return []controller.Output{
+		{
+			Type: perf.CPUType,
+			Kind: controller.OutputExclusive,
+		},
+		{
+			Type: perf.MemoryType,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+// Run implements controller.StatsController interface.
+func (ctrl *StatsController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	ticker := time.NewTicker(updateInterval)
+
+	defer ticker.Stop()
+
+	var (
+		fs  procfs.FS
+		err error
+	)
+
+	fs, err = procfs.NewDefaultFS()
+	if err != nil {
+		return err
+	}
+
+	for {
+		select {
+		case <-r.EventCh():
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+		}
+
+		if err := ctrl.updateMemory(ctx, r, &fs); err != nil {
+			return err
+		}
+
+		if err := ctrl.updateCPU(ctx, r, &fs); err != nil {
+			return err
+		}
+	}
+}
+
+func (ctrl *StatsController) updateCPU(ctx context.Context, r controller.Runtime, fs *procfs.FS) error {
+	cpu := perf.NewCPU()
+
+	stat, err := fs.Stat()
+	if err != nil {
+		return err
+	}
+
+	return r.Modify(ctx, cpu, func(r resource.Resource) error {
+		r.(*perf.CPU).Update(&stat)
+
+		return nil
+	})
+}
+
+func (ctrl *StatsController) updateMemory(ctx context.Context, r controller.Runtime, fs *procfs.FS) error {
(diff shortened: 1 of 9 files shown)


internal/app/machined/pkg/controllers/perf/perf.go
```diff
@@ -0,0 +1,111 @@
...
+	var (
+		fs          procfs.FS
+		procfsReady bool
+		err         error
+	)
+
+	for {
+		select {
+		case <-ctx.Done():
+			ticker.Stop()
```
smira, line 59: I think it's better to move it up to the `time.NewTicker` as `defer ticker.Stop()` so that ticker is stopped even if the controller panics/exists early

internal/app/machined/pkg/controllers/perf/perf.go
```diff
@@ -0,0 +1,111 @@
...
+
+			return nil
+		case <-ticker.C:
+			if !procfsReady {
+				fs, err = procfs.NewDefaultFS()
+				if err != nil {
+					logger.Error("failed to init default procfs", zap.Error(err))
```
smira, line 66: as a side note, you can return `err` here so that controller will be restarted by the runtime. 

this enables exponential backoff, and also we can do some tooling showing failing controllers (vs. failing silently here)

it's more of a idea, as of course the way it is done right now it should work as well as the ticker will force the retry

internal/app/machined/pkg/controllers/perf/perf.go
```diff
@@ -0,0 +1,111 @@
...
+		{
+			Type: perf.MemoryType,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+// Run implements controller.StatsController interface.
+func (ctrl *StatsController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	ticker := time.NewTicker(time.Second * 5)
```
smira, line 48: is it too aggressive  refresh interval? I think every five seconds is too often for regular use case

should we make it a const in this file to make it easier to see the refresh interval?

internal/app/machined/pkg/controllers/perf/perf_test.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (suite *PerfSuite) TearDownTest() {
+	suite.T().Log("tear down")
+
+	suite.ctxCancel()
+
+	suite.wg.Wait()
+
+	// trigger updates in resources to stop watch loops
```
smira, line 109: I think the lines till 118 can be removed, as the controller doesn't have any inputs

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -85,6 +86,7 @@ func (ctrl *Controller) Run(ctx context.Context) error {
 		&secrets.EtcdController{},
 		&secrets.KubernetesController{},
 		&secrets.RootController{},
+		&perf.StatsController{},
```
smira, line 89: could you please put it in alphabetical order? I think `p` comes before `s`

pkg/resources/perf/cpu.go
```diff
@@ -0,0 +1,142 @@
...
+
+// CPU represents the last CPU stats snapshot.
+type CPU struct {
+	md   resource.Metadata
+	spec CPUSpec
+}
+
+// CPUSpec represents the last CPU stats snapshot.
+type CPUSpec struct {
+	CPU             []*CPUStat `yaml:"cpu"`
```
smira, line 29: DeepCopy is not correct with this spec, as the spec contains pointers and the actual data behind them is not copied

but instead, I would rather drop pointers here and change to `CPUStat` without `*` (here and on the next line). this should lower GC pressure and make copying easy.

pkg/resources/perf/mem.go
```diff
@@ -0,0 +1,184 @@
...
+	CmaTotal          uint64 `yaml:"cmaTotal"`
+	CmaFree           uint64 `yaml:"cmaFree"`
+	HugePagesTotal    uint64 `yaml:"hugePagesTotal"`
+	HugePagesFree     uint64 `yaml:"hugePagesFree"`
+	HugePagesRsvd     uint64 `yaml:"hugePagesRsvd"`
+	HugePagesSurp     uint64 `yaml:"hugePagesSurp"`
+	Hugepagesize      uint64 `yaml:"hugepagesize"`
+	DirectMap4k       uint64 `yaml:"directMap4k"`
+	DirectMap2m       uint64 `yaml:"directMap2m"`
+	DirectMap1g       uint64 `yaml:"DirectMap1g"`
```
smira, line 77:
```suggestion
	DirectMap1g       uint64 `yaml:"directMap1g"`
```

internal/app/machined/pkg/controllers/perf/perf.go
```diff
@@ -0,0 +1,112 @@
...
+	defer ticker.Stop()
+
+	var (
+		fs          procfs.FS
+		procfsReady bool
+		err         error
+	)
+
+	for {
+		if !procfsReady {
```
smira, line 61: looks like `procfsReady` is never set to `true`

but I think we could move these lines out of the `for` loop (do we need to `Close()` it?)

internal/app/machined/pkg/controllers/perf/perf.go
```diff
@@ -0,0 +1,112 @@
...
+
+		if err := ctrl.updateMemory(ctx, r, &fs); err != nil {
+			return err
+		}
+
+		if err := ctrl.updateCPU(ctx, r, &fs); err != nil {
+			return err
+		}
+
+		select {
```
smira, line 76: if you add here `case <-r.EventCh()` and move it to the beginning of `for`, it will work correctly - runtime delivers initial reconcile event on controller start (or restart)

smira (commented):
:+1: small nits

internal/app/machined/pkg/controllers/perf/perf.go
Unix4ever: yeah, we can actually simplify it. I didn't realize that COSI retries in case if we return the error.
So I will just drop procfsReady completely.

Issue #3421, opened 2021-04-01
implement role-based RBAC for Talos API


smira:
This is initial implementation which will be extended in the future versions of Talos.

tl;dr: 

> There are two roles: `os:admin` and `os:reader`. Role `os:admin` gives full access to the APIs. Role `os:reader` limits access to read-only APIs and doesn't allow calls which might expose secrets (resources in `secrets` namespace, `machineconfiguration` resource, `read` API, etc.). Roles are encoded in the Talos client certificate in the `Organization` field and used to verify access.

## Flow

1. User generates client certificate with some role in the `Organization` field.
2. User access Talos API with that certificate as client certificate.
3. `apid` receives the request, validates TLS cert and extracts `Organization` field as a list of roles.
4. If request involves proxying to other nodes, `apid` uses its own special client certificate generated with role `os:impersonate` and encodes original roles in the gRPC metadata.
5. If request is passed down to `machined`, roles are passed via gRPC metadata.
6. Request hits `machined` with role encoded in the gRPC metadata. Each API verifies whether the role if sufficient to execute the call, otherwise `AccessDenied` is returned.

### `talosctl gen config`

Config generation code should be updated to output `talosconfig` with `os:admin` role by default always.

### `talosctl talosconfig` command

This new command generates talosconfig on the server with given params and returns back a file in `talosconfig` format:

```bash
$ talosctl talosconfig --roles=os:admin [--lifetime=1h] [path]
```

### apid

`apid` should extract roles from the `Organization` field and store them in gRPC metadata. `apid` should never accept roles passed in from the user in gRPC metadata as they can't be trusted. If role `os:impersonate` is in the certificate, actual roles are extracted from the gRPC metadata.

`apid` should generate its own client certificate used to proxy requests to other nodes with `os:impersonate` role.

`apid` should pass roles down to `machined`.

### machined

We need a generic helper to verify roles in machined gRPC server. For example, `ensureRoles(ctx, requiredRoles..) error` which returns gRPC error if required role is missing.

Most of read-only APIs should be safe for `os:reader` with the exception of resources (`secrets` namespace + `machienconfiguration` resource), and `read` API which might read files which contain secrets. 

All APIs which involve changes (reboot, reset, upgrade, ...) should require `os:admin`.

### Transitioning to RBAC

Actual RBAC checks should be disabled in `machined` unless feature gate `apiRBAC` is enabled.

Feature gates should be implemented with the following format for the machine configuration:

```yaml
machine:
[...]

PR #3712, opened 2021-06-02
feat: implement basic RBAC interceptors

AlekSi:
It is not enforced yet.

Refs #3421.

### DIFF
--- a/internal/app/apid/main.go
+++ b/internal/app/apid/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/apid/pkg/director"
 	"github.com/talos-systems/talos/internal/app/apid/pkg/provider"
 	"github.com/talos-systems/talos/pkg/grpc/factory"
+	"github.com/talos-systems/talos/pkg/grpc/middleware/authz"
 	"github.com/talos-systems/talos/pkg/grpc/proxy/backend"
 	"github.com/talos-systems/talos/pkg/machinery/config/configloader"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
@@ -118,6 +119,11 @@ func Main() {
 	var errGroup errgroup.Group
 
 	errGroup.Go(func() error {
+		injector := &authz.Injector{
+			TrustMetadata: false,
+			Logger:        log.New(log.Writer(), "apid/authz/injector/http ", log.Flags()).Printf,
+		}
+
 		return factory.ListenAndServe(
 			router,
 			factory.Port(constants.ApidPort),
@@ -133,10 +139,17 @@ func Main() {
 						proxy.WithStreamedDetector(router.StreamedDetector),
 					)),
 			),
+			factory.WithUnaryInterceptor(injector.UnaryInterceptor()),
+			factory.WithStreamInterceptor(injector.StreamInterceptor()),
 		)
 	})
 
 	errGroup.Go(func() error {
+		injector := &authz.Injector{
+			TrustMetadata: true,
+			Logger:        log.New(log.Writer(), "apid/authz/injector/unix ", log.Flags()).Printf,
+		}
+
 		return factory.ListenAndServe(
 			router,
 			factory.Network("unix"),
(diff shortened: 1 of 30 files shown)


internal/app/apid/main.go
```diff
@@ -150,6 +163,10 @@ func Main() {
 						proxy.WithStreamedDetector(router.StreamedDetector),
 					)),
 			),
+
+			// TODO needed?
+			factory.WithUnaryInterceptor(injector.UnaryInterceptor()),
+			factory.WithStreamInterceptor(injector.StreamInterceptor()),
```
AlekSi, line 169: I wasn't able to trigger that code path. When apid uses UNIX socket?

internal/app/machined/pkg/system/services/machined.go
```diff
@@ -25,12 +28,34 @@ type machinedService struct {
 
 // Main is an entrypoint the the API service.
 func (s *machinedService) Main(ctx context.Context, r runtime.Runtime, logWriter io.Writer) error {
+	injector := &authz.Injector{
+		TrustMetadata: true,
```
AlekSi, line 32: Because there is no TLS

pkg/grpc/factory/factory.go
```diff
@@ -114,7 +114,7 @@ func WithDefaultLog() Option {
 func recoveryHandler(logger *log.Logger) grpc_recovery.RecoveryHandlerFunc {
 	return func(p interface{}) error {
 		if logger != nil {
-			logger.Printf("panic:\n%s", string(debug.Stack()))
+			logger.Printf("panic: %v\n%s", p, string(debug.Stack()))
```
AlekSi, line 117: The value itself wasn't logged

internal/app/apid/main.go
smira: The easiest way to trigger that is to do `talosctl health`.

Health check runs on the server, so it reaches out to _other_ apid nodes via its own local apid.

Path is like that: `talosctl health` -> `apid` -> `machined` -> `apid (socket)` -> other `apid`

pkg/grpc/middleware/authz/authorizer.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	if allowedRoles.IncludesAny(clientRoles) {
+		a.logf("authorized (%v includes %v)", allowedRoles.Strings(), clientRoles.Strings())
+
+		return ctx, nil
+	}
+
+	if a.DontEnforce {
+		a.logf("not authorized (%v doesn't include %v), but doesn't enforced", allowedRoles.Strings(), clientRoles.Strings())
```
smira, line 85:
```suggestion
		a.logf("not authorized (%v doesn't include %v), but authorization wasn't enforced", allowedRoles.Strings(), clientRoles.Strings())
```

pkg/machinery/role/role.go
```diff
@@ -0,0 +1,90 @@
...
+// Returned set is always non-nil and contains all parsed roles, even if some erroneous roles were present.
+func Parse(strings []string) (Set, error) {
+	res := make(Set)
+
+	var err *multierror.Error
+
+	for _, r := range strings {
+		role := Role(r)
+		if _, ok := all[role]; !ok {
+			err = multierror.Append(err, fmt.Errorf("unexpected role %q", r))
```
smira, line 57: just a note: it might make sense not to fail on unknown roles, the reason is compatibility with future releases of Talos

say, in 0.12 Talos we add a new role, and someone is using client certificate with a new role to access a cluster with mixed versions (0.11/0.12) in the process of being upgraded. I don't think keeping unknown roles would be any problem for us?

smira (commented):
I  really like the direction it's going, :+1: 

I think we have a feature gate missing? and `talosctl` command to generate new `talosconfig` with some role/lifetime?

pkg/machinery/role/role.go
AlekSi: That returned error is logged and ignored by the server code. The only place where this error is fatal is in `talosctl gen csr --roles` valdiation.

smira: but the code seems to ignore such roles, so it won't pass them through the apid impersonator for example (e.g. control plane node on 0.11 and worker on 0.12)

AlekSi:
> I think we have a feature gate missing? and talosctl command to generate new talosconfig with some role/lifetime?

Yes, they are coming in future PRs – that one is already quite big.

smira (approved):
:cool:

PR #3713, opened 2021-06-02
feat: default to bootstrap workflow

andrewrynhard:
Changes `gen config` to output `controlplane` and `join` machine config
types only. Users can manually set the `type` to `init` if they need to.


smira (approved):
:partying_face:

smira (approved):

PR #3717, opened 2021-06-03
fix: drop into maintenance mode if config URL is `none` (metal)

smira:
Installer puts `none` if no URL was present in the cmdline on boot, e.g.
when installing from ISO.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -45,6 +45,10 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 		return nil, errors.ErrNoConfigSource
 	}
 
+	if *option == constants.ConfigNone {
+		return nil, errors.ErrNoConfigSource
+	}
+
 	log.Printf("fetching machine config from: %q", *option)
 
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -45,6 +45,10 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
 		return nil, errors.ErrNoConfigSource
 	}
 
+	if *option == constants.ConfigNone {
+		return nil, errors.ErrNoConfigSource
+	}
```
andrewrynhard, line 50: I wonder if this will break subsequent boots. What happens after applying a config, and this arg is not updated?

smira: we don't drop into this function unless configuration is missing

so if we got here, we need to fetch config, as config is missing. we can't fetch it from`none` obviously

andrewrynhard: ah... 👍🏻

andrewrynhard (approved):

PR #3718, opened 2021-06-03
feat: implement new controllers for hostname, resolvers and time servers

smira:
This is next part of networkd rewrite.

This implements three new resource types coupled with controllers which
process the default configuration, merges and applying changes.

TimeSync was set up to watch the time servers resource. This is a no-op
for now, but once DHCP is implemented, this would enable time server
configuration coming from DHCP.


internal/app/machined/pkg/controllers/network/hostname_config.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (ctrl *HostnameConfigController) parseMachineConfiguration(logger *zap.Logger, cfgProvider talosconfig.Provider) (spec network.HostnameSpecSpec) {
+	hostname := cfgProvider.Machine().Network().Hostname()
+
+	if hostname == "" {
+		return
+	}
+
+	if err := spec.ParseFQDN(hostname); err != nil {
+		logger.Info("ignoring error", zap.Error(err))
```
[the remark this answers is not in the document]
smira: good point, yeah, probably warning is better :+1: 

I don't want controller to fail completely on a single error, but rather try to do _something_

pkg/resources/network/hostname_spec.go
```diff
@@ -0,0 +1,118 @@
...
+	spec.Hostname = parts[0]
+
+	if len(parts) > 1 {
+		spec.Domainname = parts[1]
+	}
+
+	return spec.Validate()
+}
+
+// NewHostnameSpec initializes a SecretsStatus resource.
```
Unix4ever, line 71:
```suggestion
// NewHostnameSpec initializes a Hostname resource.
```

pkg/resources/network/hostname_status.go
```diff
@@ -0,0 +1,94 @@
...
+// FQDN returns the fully-qualified domain name.
+func (spec *HostnameStatusSpec) FQDN() string {
+	if spec.Domainname == "" {
+		return spec.Hostname
+	}
+
+	return spec.Hostname + "." + spec.Domainname
+}
+
+// NewHostnameStatus initializes a SecretsStatus resource.
```
Unix4ever, line 38:
```suggestion
// NewHostnameStatus initializes a HostnameStatus resource.
```

And it looks like all other resources have the same copy-paste error ;)

Unix4ever (approved):
:+1: with small nits

pkg/resources/network/hostname_status.go
smira: thanks, fixed all of them!

Unix4ever (approved):

PR #3732, opened 2021-06-07
chore: use Go 1.16.5

AlekSi:
Bump tools, pkgs, extras.

### DIFF
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	// DefaultKernelVersion is the default Linux kernel version.
-	DefaultKernelVersion = "5.10.29-talos"
+	DefaultKernelVersion = "5.10.38-talos"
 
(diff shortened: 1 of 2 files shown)


smira (approved):

Unix4ever (dismissed):

smira (dismissed):

pkg/machinery/constants/constants.go
```diff
@@ -13,7 +13,7 @@ import (
 
 const (
 	// DefaultKernelVersion is the default Linux kernel version.
-	DefaultKernelVersion = "5.10.29-talos"
+	DefaultKernelVersion = "5.10.38-talos"
```
AlekSi, line 16: Should it be added to release.toml now or later? What is our process for that?

AlekSi: And should we update it to 5.10.42?

smira: we can update it later

smira (approved):

PR #3733, opened 2021-06-07
Prefer extraConfig over OVF env, skip empty config

### LINKED ISSUE #3730: talos ignores extraConfig data passed in cloned vm.
### Description
Setting extraConfig data in the cloned vm as described in [vmware](https://www.talos.dev/docs/v0.11/virtualized-platforms/vmware/#create-the-remaining-control-plane-nodes)  are ignored. Whatever information you put into talos.config leads to empty config error message.
Two problems has been identified.

1. The configuration is searched across : Config.Parameter, extraConfig.Parameters in the keys talos.config/userdata. The configuration is considered found as soon the key is found eventought the key is empty. empty config should be tested specifically.
2. The configuration search stopped at the first value found. In my case, at the Config.Parameter which is empty and not modifiable from the imported OVA. So, whatever value is set in the cloned vm, the empty config is always retained. To retain the correct semantic, the search order must be enforced from most specific to default values. In this context, extraConfig parameters are more specific than Config.Parameters, thus, correct search order should be extraConfig, Config.Paramater. userdata and talos.config keys search order is not known at this time.  

### Logs
Due to early boot logs, it is not possible to copy/paste a log file. I attached a screenshot of the error message.
vmware-deploy-redacted
 
### Environment

- Talos version: [`v0.11.0-alpha.0`]
- Platform: cloud VMWare

jonkerj:
# Pull Request

## What? (description)
- Prefer extraConfig (`guestinfo.*`) over OVF env
- Skip empty (zero-length) config from whatever source

## Why? (reasoning)
In #3730 a user reported that OVF env takes precedence over extraConfig. Combined with the fact that `talos.config` is mandatory in the OVA spec (thus always present, and an empty string if you don't provide it), this causes two problems:
- when you don't supply the mandatory OVF env, it will cause an empty Talos config to be retrieved, which is unparsable;
- when you deploy a VM, you cannot override the built in Talos config, as OVF env takes precedence over extraConfig.

Fixes #3730

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/vmware/vmware_amd64.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/vmware/vmware_amd64.go
@@ -51,6 +51,12 @@ func readConfigFromExtraConfig(extraConfig *rpcvmx.Config, key string) ([]byte, 
 		return nil, fmt.Errorf("failed to decode extraConfig %s: %w", key, err)
 	}
 
+	if len(decoded) == 0 {
+		log.Printf("skipping zero-length config in extraConfig")
+
+		return nil, nil
+	}
+
 	return decoded, nil
 }
 
@@ -90,6 +96,12 @@ func readConfigFromOvf(extraConfig *rpcvmx.Config, key string) ([]byte, error) {
 				return nil, fmt.Errorf("failed to decode OVF property %s: %w", property.Key, err)
 			}
 
+			if len(decoded) == 0 {
+				log.Printf("skipping zero-length config in OVF")
+
+				return nil, nil
+			}
+
 			return decoded, nil
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/vmware/vmware_amd64.go
```diff
@@ -119,51 +119,51 @@ func (v *VMware) Configuration(context.Context) ([]byte, error) {
...
+		log.Printf("trying to find '%s' in extraConfig", constants.VMwareGuestInfoFallbackKey)
 
-		config, err = readConfigFromOvf(extraConfig, constants.VMwareGuestInfoFallbackKey)
+		config, err = readConfigFromExtraConfig(extraConfig, constants.VMwareGuestInfoFallbackKey)
 		if err != nil {
 			return nil, err
 		}
 
-		if config != nil {
+		if (config != nil) && (len(config) > 0) {
```
smira, line 142: I think it might be better to filter out empty `config` in the `readConfigFrom*` functions and return nil, this will also fix confusing log messages if the key is found, but empty

jonkerj: I had trouble to choose these two approaches, you convinced me. Fixed

jonkerj:
Sebastien has tested my build, and it worked for him.

smira (approved):

PR #3734, opened 2021-06-07
feat: implement EtcFileController to render files in `/etc`

smira:
This implements two controllers: one which generates templates for
`/etc/hosts` and `/etc/resolv.config`, and another generic controller
which renders files to `/etc` via shadow bind mounts.

### DIFF
--- a/pkg/resources/files/etcfile_spec.go
+++ b/pkg/resources/files/etcfile_spec.go
@@ -1,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package files
+
+import (
+	"fmt"
+	"io/fs"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+)
+
+// EtcFileSpecType is type of EtcFile resource.
+const EtcFileSpecType = resource.Type("EtcFileSpecs.files.talos.dev")
+
+// EtcFileSpec resource holds contents of the file which should be put to `/etc` directory.
+type EtcFileSpec struct {
+	md   resource.Metadata
+	spec EtcFileSpecSpec
+}
+
+// EtcFileSpecSpec describes status of rendered secrets.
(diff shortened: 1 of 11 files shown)


andrewrynhard:
I wonder if we could make this generic and just have a file controller. Internally we could use it for the use cases in this PR, but we could also have it manage user defined files? I know that blows up the scope. Something we can do later, and remove this then?

smira:
> I wonder if we could make this generic and just have a file controller. Internally we could use it for the use cases in this PR, but we could also have it manage user defined files? I know that blows up the scope. Something we can do later, and remove this then?

yep, this could be enhanced to handle more cases. I felt I didn't have a sense of what it should be, so I ended up doing what we need right now with the goal to handle more cases in the future.

For example, mounts should probably be handled by mounts controller, but as we don't have it yet, just do it easy way

andrewrynhard:
> > I wonder if we could make this generic and just have a file controller. Internally we could use it for the use cases in this PR, but we could also have it manage user defined files? I know that blows up the scope. Something we can do later, and remove this then?
> 
> yep, this could be enhanced to handle more cases. I felt I didn't have a sense of what it should be, so I ended up doing what we need right now with the goal to handle more cases in the future.
> 
> For example, mounts should probably be handled by mounts controller, but as we don't have it yet, just do it easy way

Makes sense. We can circle back. It is all internal right now anyways.

pkg/resources/files/etcfile_spec.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"fmt"
+	"io/fs"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+)
+
+// EtcFileSpecType is type of EtcFile resource.
+const EtcFileSpecType = resource.Type("EtcFileSpecs.net.talos.dev")
```
Unix4ever, line 16: Here we have `net`, but the namespace the resource resides is `files`. Not saying it's wrong, just asking if that aligns with our standard for the naming.

Unix4ever (dismissed):

pkg/resources/files/etcfile_spec.go
smira: yep, thanks, should be `files` to keep things consistent

PR #3735, opened 2021-06-07
feat: add configuration for enabling RBAC

AlekSi:
Refs #3421.

```yaml
machine:
    # Features describe individual Talos features that can be switched on or off.
    features:
        rbac: true # Enable role-based access control (RBAC).
```

Feature gates are normally on/off toggles, but we want to be able to add additional settings in the future, so I called those "features". But maybe that shouldn't be its own key at all? Just use `machine.rbac.enabled`, `machine.rbac.foo`, etc?


.gitignore
```diff
@@ -7,7 +7,6 @@ _out
 init.yaml
 controlplane.yaml
 join.yaml
-docgen
```
AlekSi, line 10: Not sure why that was there

Dockerfile
```diff
@@ -102,8 +102,8 @@ ARG STRINGER_VERSION
 RUN go install golang.org/x/tools/cmd/stringer@${STRINGER_VERSION} \
     && mv /go/bin/stringer /toolchain/go/bin/stringer
 RUN curl -sfL https://github.com/uber/prototool/releases/download/v1.10.0/prototool-Linux-x86_64.tar.gz | tar -xz --strip-components=2 -C /toolchain/bin prototool/bin/prototool
-COPY ./hack/docgen /go/src/github.com/talos-systems/docgen
-RUN cd /go/src/github.com/talos-systems/docgen \
+COPY ./hack/docgen /go/src/github.com/talos-systems/talos-hack-docgen
+RUN cd /go/src/github.com/talos-systems/talos-hack-docgen \
```
AlekSi, line 106: I spent too much time looking for https://github.com/talos-systems/docgen …

hack/docgen/main.go
```diff
@@ -25,7 +25,7 @@ var tpl = `// This Source Code Form is subject to the terms of the Mozilla Publi
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-// DO NOT EDIT: this file is automatically generated by docgen
+// Code generated by hack/docgen tool. DO NOT EDIT.
```
AlekSi, line 28: Changed according to https://golang.org/s/generatedcode

hack/docgen/main.go
```diff
@@ -332,10 +334,6 @@ func collectFields(s *structType) (fields []*Field) {
 		yamlTag := tag.Get("yaml")
 		yamlTag = strings.Split(yamlTag, ",")[0]
 
-		if yamlTag == "" {
-			yamlTag = strings.ToLower(yamlTag)
```
AlekSi, line 336: Not sure what was the point :)

pkg/grpc/middleware/authz/authorizer.go
```diff
@@ -81,7 +81,7 @@ func (a *Authorizer) authorize(ctx context.Context, method string) (context.Cont
 		return ctx, nil
 	}
 
-	if a.DontEnforce {
+	if a.Enforce {
```
smira, line 84:
```suggestion
	if !a.Enforce {
```

that seems crucial?

pkg/machinery/config/types/v1alpha1/v1alpha1_features.go
```diff
@@ -0,0 +1,14 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package v1alpha1
+
+// RBACEnabled implements config.Features interface.
+func (f *FeaturesConfig) RBACEnabled() bool {
+	if f.RBAC == nil {
```
smira, line 9: we should also generate new config for Talos v0.11 with RBAC enabled by default (that's our goal anyways)

hack/docgen/main.go
Unix4ever: probably the intention was to do `strings.ToLower(name)`

pkg/grpc/middleware/authz/authorizer.go
AlekSi: 🤦 Was debugging unrelated issue and changed that for a quick test…

pkg/machinery/config/types/v1alpha1/v1alpha1_features.go
AlekSi: Done in that PR

hack/docgen/main.go
AlekSi: Added a check to require an explicit name

AlekSi:
The proposed config (in the first message) looks good?

Unix4ever (dismissed):

smira (dismissed):
:+1:

PR #3736, opened 2021-06-08
feat: bring unconfigured links with link carrier up by default

smira:
This matches previous networkd implementation to make sure we can run
DHCP on the interfaces which are not explicitly brought up.


Unix4ever (approved):

smira:
re-pushed with a small fix after testing with complete network implementation

PR #3737, opened 2021-06-08
feat: implement network.Status resource and controller

smira:
This resource holds aggregated network status which can be easily used
in various places to wait for the network to reach some desired state.

The state checks are simple right now, we might improve the logic to
make sure all the configured network subsystems reached defined state,
but this might come later as we refine the logic (e.g. to make sure that
all static configuration got applied, etc.)


Unix4ever (approved):

smira:
fixed a bug and added some unit-tests

PR #3745, opened 2021-06-11
fix: do not add bootstrap contexts option if tail events is not 0

Unix4ever:
Otherwise first comes `Add` event then the list of `Update` events which may have
version less than the added one.


smira (approved):
nit:  `s/contexts/contents/

PR #3762, opened 2021-06-15
fix: separate healthy and unknown flags in the service resource

Unix4ever:
Otherwise if the service status is unknown it is shown as Unhealthy
which is actually different from what we have for:
`talosctl services` API.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3762)

### DIFF
--- a/pkg/resources/v1alpha1/service.go
+++ b/pkg/resources/v1alpha1/service.go
@@ -24,6 +24,7 @@ type Service struct {
 type ServiceSpec struct {
 	Running bool `yaml:"running"`
 	Healthy bool `yaml:"healthy"`
+	Unknown bool `yaml:"unknown"`
 }
 
 // NewService initializes a Service resource.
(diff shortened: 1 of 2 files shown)


pkg/resources/v1alpha1/service.go
```diff
@@ -75,6 +76,10 @@ func (r *Service) ResourceDefinition() meta.ResourceDefinitionSpec {
 				Name:     "Healthy",
 				JSONPath: "{.healthy}",
 			},
+			{
+				Name:     "Unknown",
```
smira, line 80: nit: `Health Unknown` ?

smira (dismissed):

pkg/resources/v1alpha1/service.go
Unix4ever: well, yeah, makes sense. Will update

smira (approved):

PR #3769, opened 2021-06-16
fix: do not format state partition in the initialize sequence

Unix4ever:
Initialize state should be only reading the config.
So now if it detects that the partition is not even formatted it will
skip it and will consider the state to be empty.

Fixes: https://github.com/talos-systems/talos/issues/3759

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3769)


smira:
Should we start testing this flow now, e.g. in `e2e-disk-image` we could enable encryption (or add another one with disk-image + encryption)?

Unix4ever:
yep, can do that.

smira (approved):

smira (approved):

PR #3773, opened 2021-06-17
feat: update etcd to 3.5.0

smira:
See https://github.com/etcd-io/etcd/blob/main/CHANGELOG-3.5.md

This also changes the way Talos adds new members to the etcd cluster:
they get added as 'learner' members and promoted to full members after
etcd is started.

Also learner members enforce that new members are only added
sequentially.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3773)


andrewrynhard (approved):
You mentioned in Slack that we need to have joining nodes be listeners at first. I don't see that change here?

smira:
> You mentioned in Slack that we need to have joining nodes be listeners at first. I don't see that change here?

yeah, I'm cleaning up the change. I don't know if we should include that in 0.11, but I'll clean it up and push as a PR. 

this change _should_ work with 3.4 as well, but not with 3.3 I think.

smira:
https://github.com/etcd-io/etcd/issues/13119 blocks the update

smira:
learner mode code moved to #4062 

etcd bug is not fixed, so we can't upgrade to 3.5.x

PR #3779, opened 2021-06-18
chore: update RBAC rules, remove old APIs

AlekSi:
Refs #3421.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3779)


Dockerfile
```diff
@@ -141,8 +141,6 @@ FROM build AS generate-build
 COPY ./api/vendor/ /api/vendor/
 COPY ./api/common/common.proto /api/common/common.proto
 RUN protoc -I/api -I/api/vendor/ --go_out=paths=source_relative:/api --go-grpc_out=paths=source_relative:/api common/common.proto
-COPY ./api/health/health.proto /api/health/health.proto
-RUN protoc -I/api -I/api/vendor/ --go_out=paths=source_relative:/api --go-grpc_out=paths=source_relative:/api health/health.proto
```
AlekSi, line 145: It wasn't hooked to anything

api/machine/machine.proto
```diff
@@ -68,9 +68,9 @@ service MachineService {
   rpc Restart(RestartRequest) returns (RestartResponse);
   rpc Rollback(RollbackRequest) returns (RollbackResponse);
   rpc Reset(ResetRequest) returns (ResetResponse);
-  rpc Recover(RecoverRequest) returns (RecoverResponse);
```
AlekSi, line 71: It did nothing since Talos 0.9

api/machine/machine.proto
```diff
@@ -68,9 +68,9 @@ service MachineService {
...
-  rpc RemoveBootkubeInitializedKey(google.protobuf.Empty)
-      returns (RemoveBootkubeInitializedKeyResponse);
+  rpc RemoveBootkubeInitializedKey(google.protobuf.Empty) returns (RemoveBootkubeInitializedKeyResponse) {
+    option deprecated = true;
```
AlekSi, line 72: Not removed yet, but documented as deprecated

internal/app/machined/pkg/system/services/machined.go
```diff
@@ -22,6 +21,65 @@ import (
...
+	"/network.NetworkService/Routes":     role.MakeSet(role.Admin, role.Reader),
+
+	// per-type authorization is handled by the service itself
+	"/resource.ResourceService": role.MakeSet(role.Admin, role.Reader),
+
+	"/storage.StorageService/Disks": role.MakeSet(role.Admin, role.Reader),
+
+	"/time.TimeService/Time":      role.MakeSet(role.Admin, role.Reader),
+	"/time.TimeService/TimeCheck": role.MakeSet(role.Admin, role.Reader),
+}
```
AlekSi, line 81: I now wonder if we need prefix handling at all. Maybe it would be better to require all methods there in full

pkg/grpc/factory/factory.go
```diff
@@ -167,6 +176,10 @@ func NewServer(r Registrator, setters ...Option) *grpc.Server {
 	server := grpc.NewServer(opts.ServerOptions...)
 	r.Register(server)
 
+	if opts.Reflection {
+		reflection.Register(server)
```
AlekSi, line 180: Might be useful in the future, but not enabled now

smira (approved):

andrewrynhard:
What was the health api used for?

smira:
> What was the health api used for?

It was used in now gone `networkd`

PR #3785, opened 2021-06-19
feat: improve security of Kubernetes control plane components

### LINKED ISSUE #3765: Make bind-addr 127.0.0.1 the default for controller-manager and scheduler
## Feature Request
Do not bind the Controller Manager & Scheduler to non-loopback insecure addresses.

### Description
The flag `--bind-address` is not set to 127.0.0.1 in the kube-scheduler and kube-controller-manager  which makes kube-controller listen to 10252/TCP and scheduler listen to 10251/TCP for metrics purposes. If not set, the default bind-address is 0.0.0.0

Exposing this should only be considered when running for example `kube-prometheus`. In any other case (default) talos should  reduce attack surface by setting the bind address to 127.0.0.1 just like kubeadm does.

sergelogvinov:
Fix of fixes #3765

disable kube-scheduler insecure port too

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3785)


smira:
I'm not sure, but according to the logs it scheduler doesn't seem to listen by default insecurely, but this wouldn't hurt

smira (approved):

sergelogvinov:
the port 10251 is open for all, and you can get statistics /metrics

andrewrynhard (approved):

PR #3802, opened 2021-06-21
feat: use COSI RD's sensitivity for RBAC

AlekSi:
Refs #3421.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3802)


pkg/resources/secrets/etcd.go
```diff
@@ -71,6 +71,7 @@ func (r *Etcd) ResourceDefinition() meta.ResourceDefinitionSpec {
 		Type:             EtcdType,
 		Aliases:          []resource.Type{},
 		DefaultNamespace: NamespaceName,
+		Sensitivity:      meta.Sensitive,
```
smira, line 74: and we need to mark `MachineConfiguration` resource as sensitive

smira (commented):
:+1:

smira (approved):
:cool:

PR #3807, opened 2021-06-21
Controlplane/bootstrap configs refactoring

sergelogvinov:
# Pull Request

* use one coredns service (SingleStack/DualStack)
* hard podaffinityterm
* easy way to add/redefine hosts on the cluster

```yaml
# coredns_hosts.yaml
data:
  hosts: |
    # static
    1.1.1.1     ntp.local
```

```bash
kubectl -n kube-system patch configmap coredns --type merge -p "$(cat coredns_hosts.yaml)"
```

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3807)


internal/app/machined/pkg/controllers/k8s/templates.go
```diff
@@ -327,16 +365,14 @@ spec:
...
-          - weight: 100
-            podAffinityTerm:
-              labelSelector:
-                matchExpressions:
-                - key: k8s-app
-                  operator: In
-                  values:
-                  - kube-dns
-              topologyKey: kubernetes.io/hostname
+          requiredDuringSchedulingIgnoredDuringExecution:
```
andrewrynhard, line 368: This will make deploying to a single node impossible. We will need the ability to set the number of replicas.

sergelogvinov: Yes, deploy I not finish. But it is not affect anything.
```shell
NAME      READY   UP-TO-DATE   AVAILABLE   AGE
coredns   1/2     2            1           2m10s

NAME                            READY   STATUS    RESTARTS   AGE
coredns-6bd64fcdc6-8f5ht        0/1     Pending   0          2m55s
coredns-6bd64fcdc6-95rlv        1/1     Running   0          2m55s
```

sergelogvinov: I've add the version tag, to make possible rolling update works.
As I know talos does't upgrade the deployments after **edit machineconfig** command. This feature may be solve rolling update in future.

andrewrynhard: I think our health checks check to see if coredns has been rolled out fully. @smira correct me if I am wrong.

sergelogvinov: Add cluster.coreDNS.replicas = 2 (by default)

internal/app/machined/pkg/controllers/k8s/manifest.go
```diff
@@ -82,6 +82,7 @@ func (ctrl *ManifestController) Run(ctx context.Context, r controller.Runtime, l
 			return err
 		}
 
+		configVersion := configResource.(*config.K8sControlPlane).Metadata().Version().String()
```
smira, line 85: this won't quite work, as Talos now doesn't ever _update_ resources from manifests, it only creates them if they're missing

internal/app/machined/pkg/controllers/k8s/templates.go
smira: I feel we should support custom deployments for coreDNS, but I don't know whether this is the right thing to do here, as it kind of forces to set number of replicas for single-node clusters

internal/app/machined/pkg/controllers/k8s/manifest.go
sergelogvinov: ok, I'll remove that.

So keep only replication count parameter?

sergelogvinov:
I am going to close it.
I've split it fo 2-PRs, and current coredns deployment boot for local/home lab deployments

PR #3814, opened 2021-06-22
feat: reimplement apid certificate generation on top of COSI

smira:
This PR can be split into two parts:

* controllers
* apid binding into COSI world

Controllers
-----------

* `k8s.EndpointController` provides control plane endpoints on worker
nodes (it isn't required for now on control plane nodes)
* `secrets.RootController` now provides OS top-level secrets (CA cert)
and secret configuration
* `secrets.APIController` generates API secrets (certificates) in a bit
different way for workers and control plane nodes: controlplane nodes
generate directly, while workers reach out to `trustd` on control plane
nodes via `k8s.Endpoint` resource

apid Binding
------------

Resource `secrets.API` provides binding to protobuf by converting
itself back and forth to protobuf spec.

apid no longer receives machine configuration, instead it receives
gRPC-backed socket to access Resource API. apid watches `secrets.API`
resource, fetches certs and CA from it and uses that in its TLS
configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3814)


api/resource/secrets.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package talos.resource.secrets;
+
+option go_package = "github.com/talos-systems/talos/pkg/machinery/api/resource";
+
+message APISpec {
```
smira, line 7: TODO: use common type for cert/key pair?

internal/app/apid/main.go
```diff
@@ -106,8 +96,9 @@ func Main() {
 		"/machine.MachineService/List",
 		"/machine.MachineService/Logs",
 		"/machine.MachineService/Read",
-		"/resource.ResourceService/List",
-		"/resource.ResourceService/Watch",
+		// TODO: revert me
+		"/talos.resource.ResourceService/List",
```
smira, line 100: will rename in COSI land to avoid conflict here

internal/app/machined/pkg/controllers/secrets/api.go
```diff
@@ -0,0 +1,530 @@
...
+		}
+
+		if err = ctrl.teardownAll(ctx, r); err != nil {
+			return err
+		}
+	}
+}
+
+//nolint:gocyclo,cyclop
+func (ctrl *APIController) reconcileControlPlane(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
```
smira, line 125: this and following reconcile methods should be merged, as they have too much in common

rsmitty (approved):

PR #3827, opened 2021-06-24
chore: small RBAC improvements

### LINKED ISSUE #3421: implement role-based RBAC for Talos API

AlekSi:
* `talosctl config new` now sets endpoints in the generated config.
* Avoid duplication of roles in metadata.
* Remove method name prefix handling. All methods should be set explicitly.
* Add tests.

Closes #3421.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3827)


AlekSi:
I wonder if that makes sense to make a separate "track" on CI with RBAC disabled?

internal/integration/cli/config.go
```diff
@@ -61,6 +62,25 @@ func (suite *TalosconfigSuite) TestMerge() {
...
+	node := suite.RandomDiscoveredNode(machine.TypeControlPlane)
+
+	talosconfigPath := filepath.Join(tempDir, "talosconfig")
+	suite.RunCLI([]string{"--nodes", node, "config", "new", "--roles", "os:reader", talosconfigPath},
+		base.StdoutEmpty())
+
+	suite.RunCLI([]string{"--nodes", node, "--talosconfig", talosconfigPath, "ls", "/etc/hosts"},
+		base.StdoutShouldMatch(regexp.MustCompile(`hosts`)))
+
+	suite.RunCLI([]string{"--nodes", node, "--talosconfig", talosconfigPath, "read", "/etc/hosts"},
```
smira, line 78: can we add more negative test cases here to make sure that `reader` can't break out?

can't read machine config, secrets, can't execute kubeconfig, reset, etc.?

smira (commented):
lgtm, but I feel we need more testcases to make sure `os:reader` is properly scoped (I know we do that in the unit-tests, but need  to be 146% sure)

smira (approved):

PR #3843, opened 2021-06-25
docs: add guide for RBAC

AlekSi:
Document how to enable RBAC without screwing up.


website/content/docs/v0.11/Guides/rbac.md
```diff
@@ -0,0 +1,50 @@
...
+There is a set of predefined roles that allow access to different [API methods](../../reference/api/):
+
+* `os:admin` grants access to all methods;
+* `os:reader` grants access to "safe" methods (for example, that includes the ability to list files, but does not include the ability to read files content);
+* `os:etcd:backup` grants access to [`/machine.MachineService/EtcdSnapshot`](../../reference/api/#machine.EtcdSnapshotRequest) method.
+
+Roles in the current `talosconfig` can be checked with the following command (using [`yq` v4](https://github.com/mikefarah/yq)):
+
+```sh
+$ yq eval '.contexts[.context].crt' talosconfig | base64 -d | openssl x509 -noout -text
```
AlekSi, line 21: That probably could be a new command like `talosctl config info`

andrewrynhard: Did we add this command?

website/content/docs/v0.11/Guides/rbac.md
```diff
@@ -0,0 +1,50 @@
...
+        Subject: O = os:reader
+        [...]
+```
+
+RBAC is enabled by default in new clusters created with `talosctl` v0.11 and disabled otherwise.
+
+## Enabling RBAC
+
+First, both the Talos cluster and `talosctl` tool should be [upgraded](../upgrading-talos/) to v0.11.
+Then `talosctl config new` command should be used to generate a new client configuration with `os:admin` role.
```
andrewrynhard, line 35:
```suggestion
Then the `talosctl config new` command should be used to generate a new client configuration with the `os:admin` role.
```

andrewrynhard (dismissed):

website/content/docs/v0.11/Guides/rbac.md
AlekSi: It will be in the next PR, but I'm not sure it will be backported – it might be too late for that

AlekSi: https://github.com/talos-systems/talos/pull/3888

Issue #907, opened 2019-07-24
[RFC] talosctl output standards


Ulexus:
As we have discussed a few times, it would be really useful to have a few standard output formats for osctl in order to allow easy automation.  Specifically, we should add `-o json` and (maybe) `-o yaml`.  We could do this by creating an interface similar to:

```go
interface OutputFormatter{
  // JSON writes JSON-formatted output to the given writer
  JSON(io.Writer) error

  // YAML writes YAML-formatted output to the given writer
  YAML(io.Writer) error

  // Tabular writes the given fields in tabular format to the given writer
  Tabular(io.Writer, fields ...string) error

  // Human writes the output in human-readable format
  Human(io.Writer) error
}
```

We can then provide a StandardFormatter for convenience which implements reasonable defaults for JSON, YAML, and Tabular, and maybe takes a `text.Template` for the Human formatting.

Ulexus:
Actually, that map is probably not the best choice for Tabular output, since field order would be lost.

andrewrynhard:
Awesome. Adding this to the agenda for tomorrow's meeting. Probably good that we think about this while we are rewriting the API.

smira:
I like the ida!

I feel for efficiency, it's better to do something like:

```go

JSON(w io.Writer) error
```
(so that we don't build output in memory)

And if method returns some `Output` object, and that object might implement one or more formatting interfaces which are called in the end to generate output. E.g. `top` has only human-readable, while `ps` might do all of them.

The most generic methods might use generic structures which already implement formatting given some abstract data structure which is serializable to JSON/YAML.

Ulexus:
The original form of the interface did exactly that (took an io.Writer as an argument to each), but I wasn't sure I wanted to couple the actual writer to it.  I thought this was a cleaner abstraction.

Thinking about it, though, I think you're correct.  It's better to just provide the `io.Writer` rather than having to buffer the entire output for no good reason.

Ulexus:
I think it would be fine to implement a null output for commands specifically and solely related to human interaction, but they should still match the interface, even if the output is empty or `{"error": "JSON output not supported"}`.

andrewrynhard:
Is this something we want for 0.4?

AlekSi:
While working on it, we should investigate https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson package:

> This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.

[...]

PR #3855, opened 2021-06-28
chore: fix tests for disabled RBAC

AlekSi:
This commit also introduces a hidden `--json` flag for `talosctl version` command
that is not supported and should be re-worked at #907.

Refs #3852.

### DIFF
--- a/cmd/talosctl/cmd/talos/version.go
+++ b/cmd/talosctl/cmd/talos/version.go
@@ -12,37 +12,42 @@ import (
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/peer"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/talos-systems/talos/pkg/cli"
 	"github.com/talos-systems/talos/pkg/machinery/client"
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/talos/version.go
```diff
@@ -82,8 +97,12 @@ var versionCmd = &cobra.Command{
...
 
 func init() {
-	versionCmd.Flags().BoolVar(&shortVersion, "short", false, "Print the short version")
-	versionCmd.Flags().BoolVar(&clientOnly, "client", false, "Print client version only")
+	versionCmd.Flags().BoolVar(&versionCmdFlags.shortVersion, "short", false, "Print the short version")
+	versionCmd.Flags().BoolVar(&versionCmdFlags.clientOnly, "client", false, "Print client version only")
+
+	// TODO remove when https://github.com/talos-systems/talos/issues/907 is implemented
+	versionCmd.Flags().BoolVar(&versionCmdFlags.json, "json", false, "")
+	cli.Should(versionCmd.Flags().MarkHidden("json"))
```
AlekSi, line 105: The alternative is to call an API in the CLI integration test that sounds strange

smira (approved):

PR #3865, opened 2021-06-29
fix: ignore deadline exceeded errors on bootstrap

smira:
With the recent changes, bootstrap API might wait for the time to be in
sync (as the apid is launched before time is sync). We set timeout to
500ms for the bootstrap API call, so there's a chance that a call might
time out, and we should ignore it.


Unix4ever (dismissed):
👍 
only thing that commit title has typo `exceeeded`.

PR #3889, opened 2021-07-05
fix: workaround 'Unauthorized' errors when accessing Kubernetes API

smira:
This should fix an error like:

```
failed to create etcd client: error getting kubernetes endpoints: Unauthorized
```

The problem is that the generated cert was used immediately, so even
slight time sync issue across nodes might render the cert not (yet)
usable. Cert is generated on one node, but might be used on any other
node (as it goes via the LB).

### DIFF
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -163,10 +163,13 @@ const (
 	KubernetesEtcdListenClientPort = "2379"
 
 	// KubernetesAdminCertCommonName defines CN property of Kubernetes admin certificate.
-	KubernetesAdminCertCommonName = "apiserver-kubelet-client"
+	KubernetesAdminCertCommonName = "admin"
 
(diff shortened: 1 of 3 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -163,11 +163,14 @@ const (
 	KubernetesEtcdListenClientPort = "2379"
 
 	// KubernetesAdminCertCommonName defines CN property of Kubernetes admin certificate.
-	KubernetesAdminCertCommonName = "apiserver-kubelet-client"
+	KubernetesAdminCertCommonName = "admin"
```
andrewrynhard, line 166: Does this need to be backported?

smira: yep, it's in the `Backports 0.11` project!

AlekSi (approved):

Unix4ever (approved):

PR #3894, opened 2021-07-06
chore: rename "join" type to "worker"

### LINKED ISSUE #3413: Machine role: worker vs join
Some parts of our code use the "worker" role, and some use "join", and it works in a confusing way.

Here, the documentation says "worker", but provides "join" as an example value: https://github.com/talos-systems/talos/blob/a43acb2150cadd78da51c41569b7f219b704f089/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go#L480-L488

The only reference to "worker" is `talosctl gen config` command implementation: https://github.com/talos-systems/talos/blob/a43acb2150cadd78da51c41569b7f219b704f089/pkg/machinery/config/types/v1alpha1/generate/join.go#L34

This code accepts only "join", not "worker": https://github.com/talos-systems/talos/blob/a43acb2150cadd78da51c41569b7f219b704f089/pkg/machinery/config/types/v1alpha1/machine/machine.go#L9-L49

But the code above is not actually used :) The real work happens there, it accepts "join", "worker", "", "foobar": https://github.com/talos-systems/talos/blob/a43acb2150cadd78da51c41569b7f219b704f089/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L220-L230

Proposed solution:
* remove dead code;
* document "worker";
* mention "join" only briefly as deprecated alias;
* explicitly (not with `default:`) accept both "join" and "worker" for compatibility;
* generate warning for "join";
* generate warning for empty value;
* generate error any other value.

AlekSi:
Closes #3413.


api/machine/machine.proto
```diff
@@ -847,10 +847,13 @@ message InstallConfig {
...
 message MachineConfig {
   enum MachineType {
+    option allow_alias = true;
     TYPE_UNKNOWN = 0;
     TYPE_INIT = 1;
     TYPE_CONTROL_PLANE = 2;
-    TYPE_JOIN = 3;
+    TYPE_WORKER = 3;
+    // Alias for TYPE_WORKER.
+    TYPE_JOIN = 3 [deprecated = true];
```
andrewrynhard, line 856: Never having used the `deprecated` option before, does that mean both `TYPE_WORKER` and `TYPE_JOIN` can both be field 3?

andrewrynhard (dismissed):
👍🏻

api/machine/machine.proto
AlekSi: In combination with `option allow_alias = true` – yes: https://github.com/AlekSi/talos/blob/9039fa6e9176d6d1a02c600eff0835a5bb146317/pkg/machinery/api/machine/machine.pb.go#L294-L298

hack/test/libvirt/libvirt.sh
```diff
@@ -59,7 +59,7 @@ function up {
     yq w -i init.yaml cluster.network.cni.name 'custom'
     yq w -i init.yaml cluster.network.cni.urls[+] "${CNI_URL}"
     yq w -i controlplane.yaml machine.install.extraKernelArgs[+] 'console=ttyS0'
-    yq w -i join.yaml machine.install.extraKernelArgs[+] 'console=ttyS0'
```
smira, line 62: we should include a note in `hack/release.toml` about the change 

we can drop all the notes there and put a new one, as it will be the first one for 0.12

internal/app/machined/pkg/controllers/secrets/api.go
```diff
@@ -103,16 +103,17 @@ func (ctrl *APIController) Run(ctx context.Context, r controller.Runtime, logger
...
 				return err
 			}
-		case machine.TypeJoin:
+		case machine.TypeWorker:
 			if err = ctrl.reconcile(ctx, r, logger, false); err != nil {
 				return err
 			}
-		case machine.TypeUnknown: // nothing to do
+		default:
+			panic(fmt.Sprintf("unexpected machine type %v", machineType))
```
smira, line 116: this panic is a bit too much here: `MachineType` resource initially is set to `machine.TypeUnknown` (until the machine config is loaded), so this branch is reachable

internal/app/machined/pkg/controllers/secrets/api.go
```diff
@@ -196,17 +197,17 @@ func (ctrl *APIController) reconcile(ctx context.Context, r controller.Runtime,
...
 			}
-		case machine.TypeJoin:
+		case machine.TypeWorker:
 			if isControlplane {
 				return fmt.Errorf("machine type changed")
 			}
-		case machine.TypeUnknown:
-			return fmt.Errorf("machine type changed")
+		default:
+			panic(fmt.Sprintf("unexpected machine type %v", machineType))
```
smira, line 210: same as above

smira (approved):
:+1: but we need to remove panics in controllers for TypeUnknown

PR #3897, opened 2021-07-07
docs: add documentation directory for 0.12

AlekSi:
Plus, convert a few absolute URLs with a version number to relative URLs without versions.


website/content/docs/v0.12/Bare Metal Platforms/digital-rebar.md
```diff
@@ -0,0 +1,172 @@
...
+In the UI of Digital Rebar you need to select the machines you want te provision.
+Once selected, you need to assign to following:
+
+- Profile
+- Workflow
+
+This will provision the Stage and Bootenv with the talos values.
+Once this is done, you can boot the machine.
+
+To understand the boot process, we have a higher level overview located at [metal overview.](/v0.11/en/guides/metal/overview)
```
smira, line 147: we should update `/v0.11/` to `/v0.12` here

website/src/pages/Index.vue
```diff
@@ -27,7 +27,7 @@
                 laptop inside Docker.
               </p>
               <div class="flex-1 text-center pb-4 m-0">
-                <a href="/docs/v0.8/introduction/quickstart">
+                <a href="/docs/latest/">
```
AlekSi, line 30: Can we make `/docs/latest/introduction/quickstart` work?

website/content/docs/v0.12/Bare Metal Platforms/digital-rebar.md
AlekSi: Converted to a relative URL instead

website/src/pages/Index.vue
andrewrynhard: We _can_, but we need to commit to a URL that we think makes sense long term. If we rearrange the layout, will `/docs/latest/introduction/quickstart` still make sense? What about just `/docs/latest/quickstart`? Then we don't care if the QuickStart is under introduction.

AlekSi: Can `/docs/latest/XXX` redirect to `/docs/v0.mm/XXX` like https://www.talos.dev/docs/latest/ redirects now?

andrewrynhard: No, not quite. We have to define redirects statically: https://github.com/talos-systems/talos/blob/master/website/static/_redirects.

AlekSi: `/docs/latest/* /docs/v0.10/:splat` worked. PTAL

website/static/_redirects
```diff
@@ -1,6 +1,7 @@
-# Redirects from what the browser requests to what we serve.
-#
+/docs/latest/* /docs/v0.10/:splat 302
```
andrewrynhard, line 1: 👍🏻 nice. Works?

andrewrynhard: They only thing about this now is that if we ever change the layout of the docs, all the `latest` URLs will need to be updated. Not a big deal, but we should probably make not of that here?

smira (approved):

website/content/docs/v0.12/Guides/air-gapped.md
```diff
@@ -0,0 +1,137 @@
...
+
+We are going to use a QEMU-based Talos cluster for this guide, but the same approach works with Docker-based clusters as well.
+As QEMU-based clusters go through the Talos install process, they can be used better to model a real air-gapped environment.
+
+The `talosctl cluster create` command provides conveniences for common configuration options.
+The only required flag for this guide is `--registry-mirror '*'=http://10.5.0.1:6000` which redirects every pull request to the internal registry.
+The endpoint being used is `10.5.0.1`, as this is the default bridge interface address which will be routable from the QEMU VMs (`127.0.0.1` IP will be pointing to the VM itself).
+
+```bash
+$ sudo -E talosctl cluster create --provisioner=qemu --registry-mirror '*'=http://10.5.0.1:6000 --install-image=ghcr.io/talos-systems/installer:v0.11.0
```
Unix4ever, line 102: Should be updated to 0.12 I suppose.

Unix4ever (approved):

website/static/_redirects
AlekSi: Those a temporary redirects, not permanent ones as they were before. I think we can use some existing service to monitor 404, or just write our own crawler that would detect broken link

website/content/docs/v0.12/Guides/air-gapped.md
AlekSi: That PR mostly creates the folder. Updating for 0.12 is the next PR.

Unix4ever: gotcha

PR #3900, opened 2021-07-07
fix: ignore DeadlineExceeded error correctly on bootstrap

smira:
The problem was that gRPC method `status.Code(err)` doesn't unwrap
errors, while Talos client returns errors wrapped with
`multierror.Error` and `fmt.Errrorf`, so `status.Code` doesn't return
error code correctly.

Fix that by introducing our own client method which correctly goes over
the chain of wrapped errors.

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -16,7 +16,6 @@ import (
 	"github.com/talos-systems/go-retry/retry"
 	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
 	"github.com/talos-systems/talos/pkg/machinery/client"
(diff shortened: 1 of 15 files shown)


smira:
this fixes frequent CI issue when `talosctl cluster create` fails with `DeadlineExceeded` error when there's a race between boostrap call and time sync on the node

pkg/cluster/bootstrap.go
```diff
@@ -75,7 +74,7 @@ func (s *APIBootstrapper) Bootstrap(ctx context.Context, out io.Writer) error {
 			switch {
 			case errors.Is(err, context.DeadlineExceeded):
 				return retry.ExpectedError(err)
-			case status.Code(err) == codes.FailedPrecondition || status.Code(err) == codes.DeadlineExceeded:
+			case client.StatusCode(err) == codes.FailedPrecondition || client.StatusCode(err) == codes.DeadlineExceeded:
```
AlekSi, line 77: While we are on it, I think it is worth adding comments about why we check error this way, why we check error string below, etc

pkg/machinery/client/status_test.go
```diff
@@ -0,0 +1,63 @@
...
+			code: codes.Unknown,
+		},
+		{
+			name: "status",
+			err:  status.Error(codes.AlreadyExists, "file already exists"),
+			code: codes.AlreadyExists,
+		},
+		{
+			name: "status wrapped",
+			err:  multierror.Append(new(multierror.Error), status.Error(codes.AlreadyExists, "file already exists")).ErrorOrNil(),
```
AlekSi, line 43: It could be 

```suggestion
			err:  multierror.Append(nil, status.Error(codes.AlreadyExists, "file already exists")).ErrorOrNil(),
```

AlekSi (approved):

pkg/machinery/client/status_test.go
smira: yeah, they seem to be equivalent in the end

smira:
@AlekSi updated

AlekSi (approved):

PR #3901, opened 2021-07-07
fix: validate bond slaves addressing

smira:
This extends network device machine configuration validation to make
sure that bond slaves don't have any addressing methods set, as this
might run into a conflict with the bond setup.

Also makes sure no interface is part of two bonds.


andrewrynhard (commented):
What is "addressing methods" here? DHCP? VIP?

smira:
> What is "addressing methods" here? DHCP? VIP?

all of them, they shouldn't be set on slave interfaces

andrewrynhard (approved):

PR #3905, opened 2021-07-07
chore: use vtprotobuf compiled marshaling in Talos API

smira:
Use `vtprobuf` optimized Marshal/Unmarshal methods which do not depend
on reflection to reduce memory and CPU usage while using Talos API.

See https://github.com/planetscale/vtprotobuf

See https://vitess.io/blog/2021-06-03-a-new-protobuf-generator-for-go/


andrewrynhard:
Nice. Do we have metrics we can compare?

smira:
> Nice. Do we have metrics we can compare?

I was going to build something, what is interesting (and a bit different topic), we use lots of resources doing gRPC calls to etcd for the health checks, I'm looking if I can save some by caching the etcd client. But we can't use vtprotobuf with etcd/grpc until etcd/grpc actually build their clients with it.

But anyways, I want to build some comparison for our case. This PR is a bit of a mess, but it shows the idea. it's really nice and doesn't require many changes.

smira:
diff between vtprotobuf and "regular" protobuf is:

image

I was simply running Talos integration test in a loop while collecting those graphs. With vtprotobuf, this all dynamic/refelct based marshaling code goes away completely.

Unix4ever (dismissed):

smira:
> What is the difference in absolute numbers?

I guess it depends on the load really. The more API calls we have, the more savings we'd have. So in regular Talos usage pattern (low number of API calls), there's literally no difference. I was only able to spot the difference via memory profiling.

AlekSi:
Them maybe it is not worth merging it? Despite all flows of the primary protobuf implementation, it is likely being maintained for a long time, while vtprotobuf might follow gogo.

On the other hand, it is easy to switch back and forth between proto and vtprotobuf. It may be made even easier if we move codec into `github.com/talos-systems/talos/pkg/grpc/factory/proto` and then use linter to ensure that this package is imported instead of other "proto" package.

smira:
> Them maybe it is not worth merging it? Despite all flows of the primary protobuf implementation, it is likely being maintained for a long time, while vtprotobuf might follow gogo.

Yes, there's a risk, but `vtprotobuf` doesn't change any on the wire representation, nor it doesn't block anyone using it without it. On other hand anyone importing `machinery` now can enjoy vtprotobuf-optimized client-side calls (if they opt-in), while if we never merge it, it requires manual generation which is hard to push for.

> On the other hand, it is easy to switch back and forth between proto and vtprotobuf. It may be made even easier if we move codec into `github.com/talos-systems/talos/pkg/grpc/factory/proto` and then use linter to ensure that this package is imported instead of other "proto" package.

I'm not sure what do you mean by linter here?

AlekSi:
> Yes, there's a risk, but vtprotobuf doesn't change any on the wire representation

That was the promise of gogoproto too, but there were subtle differences in edge cases. For example, I remember encountering a difference in map encoding with empty keys.

> I'm not sure what do you mean by linter here?

This PR changed `pkg/resources/secrets/api.go` to import `factory` package and use `factory.VTProtoCodec{}.Marshal` instead of `proto.Marshal`. If we add `Marshal` function to our package and move it to `github.com/talos-systems/talos/pkg/grpc/factory/proto`, then only import should be updated, not the rest of the code. But then we need some linter that ensures that the right `proto` package is imported. From the description of https://github.com/talos-systems/importvet, it looks like it can do it. If not, https://github.com/OpenPeeDeeP/depguard and https://github.com/ryancurrah/gomodguard (both are included in golangci-lint) can do that.

smira:
> > Yes, there's a risk, but vtprotobuf doesn't change any on the wire representation
> 
> That was the promise of gogoproto too, but there were subtle differences in edge cases. For example, I remember encountering a difference in map encoding with empty keys.

That can be always true, but I don't see it being a risk right now, as any inconsistencies can be discovered if any. 
 
> > I'm not sure what do you mean by linter here?
> 
> This PR changed `pkg/resources/secrets/api.go` to import `factory` package and use `factory.VTProtoCodec{}.Marshal` instead of `proto.Marshal`. If we add `Marshal` function to our package and move it to `github.com/talos-systems/talos/pkg/grpc/factory/proto`, then only import should be updated, not the rest of the code. But then we need some linter that ensures that the right `proto` package is imported. From the description of https://github.com/talos-systems/importvet, it looks like it can do it. If not, https://github.com/OpenPeeDeeP/depguard and https://github.com/ryancurrah/gomodguard (both are included in golangci-lint) can do that.

I think that's not that actually, there are couple of places in Talos we explicitly marshal to proto, but the biggest user is gRPC, and what that `codec.go` does is that it replaces gRPC default protobuf codec to use `*VT`() methods.

AlekSi:
See https://github.com/talos-systems/talos/pull/4027. Note that there are no code changes in internal/app/apid/pkg/backend/apid.go, internal/app/machined/pkg/runtime/events.go, etc. Only import changed, and it is now enforced by the depguard linter. That would allow us to enable and disable the usage of vtproto extensions by changing only pkg/machinery/proto package. It can be improved further by using build tags for side-by-side comparision if we want.

smira:
Closing in favor of #4027

PR #3906, opened 2021-07-07
chore: implement DeepCopy for machine configuration

smira:
Resources code extensively uses DeepCopy to prevent in-memory copy of
the resource to be mutated outside of the resource model.

Previous implementation relied on YAML serialization to copy the
machine configuration which was slow, potentially might lead to panics
and it generates pressure on garbage collection.

This implementation uses k8s code generator to generate DeepCopy methods
with some manual helpers when code generator can't handle it.


smira:
before this change, copying machine config accounted for 30% of all allocations in `machined` (not in use memory, but just overall allocations)

pkg/resources/config/machine_config.go
```diff
@@ -66,20 +65,10 @@ func (r *MachineConfig) String() string {
...
-	c, err := configloader.NewFromBytes(b)
-	if err != nil {
-		panic(err)
-	}
-
 	return &MachineConfig{
 		md: r.md,
 		spec: &v1alpha1Spec{
-			cfg: c.(*v1alpha1.Config),
+			cfg: r.spec.cfg.(*v1alpha1.Config).DeepCopy(),
```
smira, line 71: this is the goal of this PR

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -6,16 +6,17 @@
 Package v1alpha1 configuration file contains all the options available for configuring a machine.
 
 To generate a set of basic configuration files, run:
-```bash
-talosctl gen config --version v1alpha1 <cluster name> <cluster endpoint>
-````
+
```
Unix4ever, line 9: This line is actually rendered on our configuration reference page as code block.
Was it something related to deepcopy-gen?

Unix4ever (dismissed):
:cool:

AlekSi (approved):
LGTM, but it should be rebase now due to my 0.12 folder merge

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
smira: Ah, I tried to fix the godoc output, I will double check it still renders as code block in the docs.

smira: yep, it looks good in our docs now and in godoc as well

smira:
> LGTM, but it should be rebased now due to my 0.12 folder merge

updated moving generated docs to 0.12

PR #3912, opened 2021-07-08
fix: cache etcd client used for healthchecks

smira:
We run etcd health check every 30s, and create/destroy client every 30s.
This puts a lot of pressure on etcd itself and machined.

There's protobuf overhead, TLS connection overhead, etc.

As we don't support changing etcd PKI (yet), client created once is good
enough for the lifetime of the node.


smira:
with this PR, `init` CPU usage stays below 1% consistently, around 0.3% for me.

andrewrynhard (approved):

PR #3929, opened 2021-07-13
feat: use dynamic NodeAddresses/HostnameStatus in Kubernetes certs

smira:
This is a PR on a path towards removing `ApplyDynamicConfig`.

This fixes Kubernetes API server certificate generation to use dynamic
data to generate cert with proper SANs for IPs of the node.

As part of that refactored a bit apid certificate generation (without
any changes).

Added two unit-tests for apid and Kubernetes certificate generation.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -414,7 +414,9 @@ func create(ctx context.Context) (err error) {
 			fallthrough
 		case forceEndpoint != "":
 			endpointList = []string{forceEndpoint}
+			// using non-default endpoints, provision additional cert SANs and fix endpoint list
 			provisionOptions = append(provisionOptions, provision.WithEndpoint(forceEndpoint))
+			genOptions = append(genOptions, generate.WithAdditionalSubjectAltNames(endpointList))
 		case forceInitNodeAsEndpoint:
 			endpointList = []string{ips[0][0].String()}
 		default:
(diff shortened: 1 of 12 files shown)


pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -428,17 +428,12 @@ func NewInput(clustername, endpoint, kubernetesVersion string, secrets *SecretsB
 		return nil, err
 	}
 
-	var additionalSubjectAltNames []string
+	additionalSubjectAltNames := append([]string(nil), options.AdditionalSubjectAltNames...)
```
smira, line 431: this is tricky.

before this PR, we were always adding all endpoints (for QEMU clusters that's all CP nodes IPs) as additional cert SANs, this is not needed at all, and this doesn't allow to test probably proper injection of node's IP address as certSAN.

I couldn't really drop this, as it might break other machinery users, so enabling it only for 0.12

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -404,7 +404,9 @@ func create(ctx context.Context) (err error) {
 			fallthrough
 		case forceEndpoint != "":
 			endpointList = []string{forceEndpoint}
+			// using non-default endpoints, provision additional cert SANs and fix endpoint list
 			provisionOptions = append(provisionOptions, provision.WithEndpoint(forceEndpoint))
+			genOptions = append(genOptions, generate.WithAdditionalSubjectAltNames(endpointList))
```
smira, line 409: only for forced endpoint (usually docker cluster with `--endpoint=127.0.0.1` or other host IP), inject that IP as `certSAN` as Talos node doesn't see the endpoint (it's external IP essentially).

Unix4ever (approved):

Unix4ever (approved):

Unix4ever (approved):

PR #3935, opened 2021-07-15
feat: update Kubernetes to v1.22.0

### LINKED ISSUE #3967: Kubernetes 1.22
We should upgrade to k8s 1.22. Known problems that were fixed:

# DynamicKubeletConfig feature gate

```
"Failed to validate kubelet flags" err="the DynamicKubeletConfig feature gate must be enabled in order to use the --dynamic-config-dir flag"
Flag --cni-conf-dir has been deprecated, will be removed along with dockershim.
Flag --dynamic-config-dir has been deprecated, Feature DynamicKubeletConfig is deprecated in 1.22 and will not move to GA. It is planned to be removed from Kubernetes in the version 1.23. Please use alternative ways to update kubelet configuration.
```

~For 1.22, we enabled the `DynamicKubeletConfig` feature gate by default (and correctly merge it with `extraArgs`). We should find a better solution that would work for 1.23 that will remove DynamicKubeletConfig completely.~ We were not really using it, removed. See also #3997.

# `--audit-log-path=-`

https://github.com/kubernetes/kubernetes/pull/95387 broke `--audit-log-path=-` support. It was fixed at https://github.com/kubernetes/kubernetes/issues/103845.

---

https://github.com/talos-systems/talos/pull/3935

### LINKED ISSUE #3997: Remove DynamicKubeletConfig
After #3967, we should remove the usage of `DynamicKubeletConfig`.

AlekSi:
Closes #3967.
Closes #3997.


internal/integration/provision/upgrade.go
```diff
@@ -69,12 +69,12 @@ type upgradeSpec struct {
...
 	previousRelease = "v0.10.4"
-	stableRelease   = "v0.11.0-beta.0" // or soon-to-be-stable
+	stableRelease   = "v0.11.1" // or soon-to-be-stable
 	// The current version (the one being built on CI) is DefaultSettings.CurrentVersion.
 
-	previousK8sVersion = "1.21.0" // constants.DefaultKubernetesVersion in the previousRelease
+	previousK8sVersion = "1.21.1" // constants.DefaultKubernetesVersion in the previousRelease
 	stableK8sVersion   = "1.21.2" // constants.DefaultKubernetesVersion in the stableRelease
-	currentK8sVersion  = "1.21.2" // next k8s version being tested
+	currentK8sVersion  = constants.DefaultKubernetesVersion
```
AlekSi, line 77: It that correct?

smira: yes, I think that's better!

smira (approved):

andrewrynhard (approved):

andrewrynhard (dismissed):

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -199,7 +200,6 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 		fmt.Sprintf("--etcd-certfile=%s", filepath.Join(constants.KubernetesAPIServerSecretsDir, "etcd-client.crt")),
 		fmt.Sprintf("--etcd-keyfile=%s", filepath.Join(constants.KubernetesAPIServerSecretsDir, "etcd-client.key")),
 		fmt.Sprintf("--etcd-servers=%s", strings.Join(cfg.EtcdServers, ",")),
-		"--insecure-port=0",
```
smira, line 202: I don't remember what was the default value in kubernetes 1.20-1.21?

AlekSi: It was ignored in 1.20: https://github.com/kubernetes/kubernetes/blob/af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38/cmd/kube-apiserver/app/options/options.go#L159-L163

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -290,7 +290,6 @@ func (k *Kubelet) args(r runtime.Runtime) ([]string, error) {
 		"container-runtime":          "remote",
 		"container-runtime-endpoint": "unix://" + constants.CRIContainerdAddress,
 		"config":                     "/etc/kubernetes/kubelet.yaml",
-		"dynamic-config-dir":         "/etc/kubernetes/kubelet",
```
AlekSi, line 293: Wasn't used anywhere AFAIU

Unix4ever:
I wonder if we should also enable upgrade flow from `1.21 to 1.22`, by adding a new option here:
https://github.com/talos-systems/talos/blob/0c7ce1cd814354213a1a6c7a9251b166ee58c493/pkg/cluster/kubernetes/talos_managed.go#L44

smira (dismissed):

internal/app/machined/pkg/system/services/kubelet.go
andrewrynhard: Not by us by default. This was something users did on their own. Still safe to remove.

Unix4ever (dismissed):

internal/integration/provision/upgrade.go
```diff
@@ -69,12 +69,12 @@ type upgradeSpec struct {
 
 const (
 	previousRelease = "v0.10.4"
-	stableRelease   = "v0.11.0-beta.0" // or soon-to-be-stable
+	stableRelease   = "v0.11.3" // or soon-to-be-stable
```
smira, line 72: we should update the `Makefile` as well

PR #3971, opened 2021-07-20
feat: remove defaults from kube-apiserver enabled admission plugins

szucsitg:
Signed-off-by: Adam Szucs-Matyas <szucsitg@gmail.com>

# Pull Request

## What? (description)
Removes "duplicate" flags from `--enable-admission-plugins`. Does not change current behavior, except removing deprecated PodSecurityPolicy.

## Why? (reasoning)
Defaults are always enabled, unless explicitly defined as disabled. However having the same value in both `--enable-admission-plugins` and `--disable-admission-plugins` is not allowed. Therefore NodeRestriction cannot be removed at the moment. Long term solution would be to expose such command line args in machine config, so they can be overriden. Current behavior doesn't remove/merge duplicate entries, when additional args is used, but added additionally.

## Acceptance


AlekSi:
Please add an explanation to the commit message body – it can't be empty. That can be checked with `make conformance`.

andrewrynhard:
I believe we still create a PSP by default in Talos.

szucsitg:
you are right @andrewrynhard i amended the commit and removed it from everywhere. hopefully 😄

andrewrynhard:
We may want to hold off on this for a few weeks until Andrey gets back. We have been having discussions around PSP, and what the alternative should be. In theory, our PSP is a noop today, so this wouldn't necessarily hurt.

szucsitg:
Yes, your use of PSP is practically ineffective, we used them extensively, but some updates from last week created weird behavior with some services, so I wanted to get rid off it, and turn it off, but I cannot.

smira:
I don't feel we should remove PSP by default for all users, as this change affects new clusters and old clusters upgraded to Talos 0.12

smira:
I fully agree we should probably reduce this list by removing defaults, and eventually we need a way to deal with PSP going away, or to have an option to disable PSP for now.

szucsitg:
This is deprecated, so you need to migrate away. Also currently if someone don’t remove your default policy manually, then it’s completely ineffective.

Additionally nobody else had issues with PSP when you converted to static control plane besides us, so it’s pretty safe to assume that it’s simply not used widely.

smira:
> This is deprecated, so you need to migrate away. Also currently if someone don’t remove your default policy manually, then it’s completely ineffective.
> 
> Additionally nobody else had issues with PSP when you converted to static control plane besides us, so it’s pretty safe to assume that it’s simply not used widely.

let me try to explain it different way: if someone is using PSP today, this change effectively disables PSP silently on Talos upgrade which is not acceptable imho. 

I do agree we should migrate away, but we need backwards compatibility to some extent here.

smira:
thanks @szucsitg, we had this PR merged as two similar changes #4014 and #4019

Issue #3836, opened 2021-06-24
Containerd registry refactoring


sergelogvinov:
## Feature Request

https://github.com/containerd/containerd/releases/tag/v1.5.0

registry.config_path

```toml
[plugins."io.containerd.grpc.v1.cri".registry]
   config_path = "/etc/containerd/certs.d"
```

```toml
server = "https://registry-1.docker.io"    # Exclude this to not use upstream

[host."https://public-mirror.example.com"]
  capabilities = ["pull"]                  # Requires less trust, won't resolve tag to digest from this host
[host."https://docker-mirror.internal"]
  capabilities = ["pull", "resolve"]
  ca = "docker-mirror.crt"                 # Or absolute path /etc/containerd/certs.d/docker.io/docker-mirror.crt
```

* containerd can reload registry configs without restart.
* capabilities parameters

sergelogvinov:
discuss topics
1. move all secrets to the /system/secret/cri folder https://github.com/talos-systems/talos/blob/master/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go#L926
2. apply registry configs immediately (no reboot required)
3. add key machine.registries.config."$host".capabilities = ["pull", "resolve", "push"] to define mirror host capabilities

PR #3975, opened 2021-07-20
Containerd config refactoring

sergelogvinov:
# Pull Request

#3836

[Containerd registry-host documentation](https://github.com/containerd/containerd/blob/main/docs/hosts.md) 

* resolve startup containerd warnings messages
* folder based TLS auth/mirror endpoints
* can use insecure registry

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/3975)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -344,8 +344,9 @@ COPY hack/containerd.toml /rootfs/etc/cri/containerd.toml
 RUN touch /rootfs/etc/resolv.conf
 RUN touch /rootfs/etc/hosts
 RUN touch /rootfs/etc/os-release
-RUN mkdir -pv /rootfs/{boot,usr/local/share,mnt,system}
-RUN mkdir -pv /rootfs/{etc/kubernetes/manifests,etc/cni,usr/libexec/kubernetes}
+RUN mkdir -pv /rootfs/{boot,usr/local/share,mnt,system,opt}
+RUN mkdir -pv /rootfs/{etc/kubernetes/manifests,etc/cni/net.d,usr/libexec/kubernetes}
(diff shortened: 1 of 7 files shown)


AlekSi:
Please make `make conformance` happy: it should be a single commit with a non-empty message body, etc.

sergelogvinov:
Thank you for your notice, i am not ready yet )

smira:
I don't quite like new containerd CRI plugin config format for registries, as it's multiple files now. I wonder if we should at least try to put config files to `/system/secrets` to make it ephemeral, so that we can clear extra files on reboot (at least).

Plus my concern is that if we put hostname into file paths, it might contian entries like `foo/../bar` which allows some random files overwrites (there are other ways to do that, but still). I don't have a good answer for that right now.

sergelogvinov:
Yep, store configs to /system/secrets got idea, I've got this limitation https://github.com/talos-systems/talos/blob/master/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go#L959 (create operation not allowed outside of /var)

foo/../bar - yes, we can prevent it

sergelogvinov:
> 
> Plus my concern is that if we put hostname into file paths, it might contian entries like `foo/../bar` which allows some random files overwrites (there are other ways to do that, but still). I don't have a good answer for that right now.

I am using this method to get folder name.
```go
u, err := url.Parse(registryHostURL)

...

filepath.Join(constants.CRIContainerdConfigDir, u.Host, "hosts.toml"),
```

So it resolves **/../..** hacks

smira:
Should be replaced by #4825

PR #3999, opened 2021-08-02
feat: implement sysctl controllers

Unix4ever:
Fixed: https://github.com/talos-systems/talos/issues/3686

Replaced sequencer tasks for KSPP and Kubernetes required sysctl props
by the ones set by controllers.

Sysctl flow includes of 3 controllers and 2 resources:
- `KernelParamsConfigController` - handles user sysctls coming from v1alpha1
config.
- `KernelParamsDefaultsController` - handles our built-in KSPP and K8s
required sysctls.
- `KernelParamsSpecController` - consumes `KernelParamsSpec`s created by the
previous two controllers, applies them and updates the corresponding
`KernelParamsStatus`.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -36,6 +36,14 @@ before upgrading to Talos 0.12.
 Current control plane status can be checked with `talosctl get bootstrapstatus` before performing upgrade to Talos 0.12.
 """
 
+    [notes.cosi]
+        title = "Sysctl Configuration"
+        description = """\
+Sysctl Kernel Params configuration was completely rewritten to be based on controllers and resources,
+which makes it possible to apply `.machine.sysctls` in immediate mode (without a reboot).
+`talosctl get kernelparams` returns merged list of KSPP, Kubernetes and user defined params along with
+the default values overwritten by Talos.
+"""
 
 [make_deps]
 
--- a/internal/app/machined/pkg/controllers/runtime/kernel_param_config.go
+++ b/internal/app/machined/pkg/controllers/runtime/kernel_param_config.go
@@ -1,0 +1,105 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package runtime
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/AlekSi/pointer"
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+	"go.uber.org/zap"
+
+	"github.com/talos-systems/talos/pkg/resources/config"
+	"github.com/talos-systems/talos/pkg/resources/runtime"
+)
+
+// KernelParamConfigController watches v1alpha1.Config, creates/updates/deletes kernel param specs.
+type KernelParamConfigController struct{}
+
+// Name implements controller.Controller interface.
+func (ctrl *KernelParamConfigController) Name() string {
+	return "runtime.KernelParamConfigController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *KernelParamConfigController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineConfigType,
+			ID:        pointer.ToString(config.V1Alpha1ID),
+		},
+	}
+}
+
+// Outputs implements controller.Controller interface.
+func (ctrl *KernelParamConfigController) Outputs() []controller.Output {
+	return []controller.Output{
+		{
+			Type: runtime.KernelParamSpecType,
+			Kind: controller.OutputShared,
+		},
+	}
+}
+
+// Run implements controller.Controller interface.
+//
+//nolint:gocyclo
+func (ctrl *KernelParamConfigController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+			cfg, err := r.Get(ctx, resource.NewMetadata(config.NamespaceName, config.MachineConfigType, config.V1Alpha1ID, resource.VersionUndefined))
+			if err != nil {
+				if !state.IsNotFoundError(err) {
+					return fmt.Errorf("error getting config: %w", err)
+				}
+			}
+
+			touchedIDs := make(map[resource.ID]struct{})
+
+			if cfg != nil {
+				c, _ := cfg.(*config.MachineConfig) //nolint:errcheck
+				for key, value := range c.Config().Machine().Sysctls() {
+					touchedIDs[key] = struct{}{}
+
+					value := value
+					item := runtime.NewKernelParamSpec(runtime.NamespaceName, key)
+
+					if err = r.Modify(ctx, item, func(res resource.Resource) error {
+						res.(*runtime.KernelParamSpec).TypedSpec().Value = value
+
+						return nil
+					}); err != nil {
+						return err
+					}
+				}
+			}
+
+			// list keys for cleanup
+			list, err := r.List(ctx, resource.NewMetadata(runtime.NamespaceName, runtime.KernelParamSpecType, "", resource.VersionUndefined))
+			if err != nil {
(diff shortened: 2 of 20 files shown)


andrewrynhard (commented):
I wonder if this is an opportunity for us to move away from "sysctl" as "sysctl" is really a CLI tool that modifies _kernel parameters_. It would be better if we called them `KernelParameters` instead of `Sysctls` IMHO. It can stay as `sysctls` in the config (to avoid breakage) but maybe the resources we introduce now can be the "proper" name?

Unix4ever:
Yeah... Makes sense. Will rework

internal/app/machined/pkg/controllers/runtime/kernel_param_config.go
```diff
@@ -0,0 +1,105 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smira, line 1: I'm a very big fan of unit-tests for controllers. They seem to be trivial, but they allow us to catch bugs easily and sometimes allow to test features we rarely touch.

For example, I don't think we ever touch sysctls in our tests, so any regression might go unnoticed.

internal/app/machined/pkg/controllers/runtime/kernel_param_defaults.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (ctrl *KernelParamDefaultsController) getKernelParams() []*kernel.Param {
+	res := []*kernel.Param{
+		{
+			Key:   "net.ipv4.ip_forward",
+			Value: "1",
+		},
+	}
+
+	if ctrl.V1Alpha1Mode == v1alpha1runtime.ModeContainer {
```
smira, line 85: feels like the condition is backwards

```suggestion
	if ctrl.V1Alpha1Mode != v1alpha1runtime.ModeContainer {
```

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
```diff
@@ -0,0 +1,146 @@
...
+			if err != nil {
+				return err
+			}
+
+			touchedIDs := map[string]struct{}{}
+
+			for _, item := range list.Items {
+				spec := item.(*runtime.KernelParamSpec).TypedSpec()
+
+				if err = ctrl.updateKernelParam(ctx, r, item.Metadata().ID(), spec.Value); err != nil {
```
smira, line 79: one thing we _might_ want to do here is to make this loop accumulate errors in `multierror` and return error only when done with all the sysctls

the reason is that some sysctls might have dependency on each other, as we don't enforce any order here explicitly (except for alphabetical), it might be that one sysctl can be written only after other sysctl is written or reset back to its default value, so I would suggest to run both loops and only report error afterwards

internal/app/machined/pkg/controllers/runtime/kernel_param_defaults.go
```diff
@@ -0,0 +1,110 @@
...
+
+		for _, prop := range kernels {
+			value := prop.Value
+			item := runtime.NewKernelParamSpec(runtime.NamespaceName, prop.Key)
+
+			if err := r.Modify(ctx, item, func(res resource.Resource) error {
+				res.(*runtime.KernelParamSpec).TypedSpec().Value = value
+
+				if item.Metadata().ID() == "net.ipv6.conf.default.forwarding" {
+					res.(*runtime.KernelParamSpec).TypedSpec().IgnoreErrors = true
```
smira, line 64: I don't see this being actually implemented with `KernelParamSpecController`

the easiest way to test this is to boot Talos with `ipv6=false` kernel arg

internal/app/machined/pkg/controllers/runtime/kernel_param_test.go
```diff
@@ -0,0 +1,193 @@
...
+	if state.IsConflictError(err) {
+		err = suite.state.Destroy(context.Background(), config.NewMachineConfig(nil).Metadata())
+	}
+
+	suite.Assert().NoError(err)
+}
+
+func (suite *KernelParamSuite) TestReconcileConfig() {
+	suite.Require().NoError(suite.runtime.RegisterController(&runtimecontrollers.KernelParamConfigController{}))
+	suite.Require().NoError(suite.runtime.RegisterController(&runtimecontrollers.KernelParamSpecController{}))
```
smira, line 107: I know testing them together is easier, but I feel testing them in isolation allows to expose more code paths

internal/app/machined/pkg/controllers/runtime/kernel_param_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	suite.startRuntime()
+
+	id := "kernel.hostname"
+	value := "test0"
```
smira, line 112: isn't this sysctl too scary to be written?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
```diff
@@ -110,6 +110,7 @@ func (r *Runtime) CanApplyImmediate(b []byte) error {
 		newConfig.MachineConfig.MachineTime = currentConfig.MachineConfig.MachineTime
 		newConfig.MachineConfig.MachineCertSANs = currentConfig.MachineConfig.MachineCertSANs
 		newConfig.MachineConfig.MachineNetwork = currentConfig.MachineConfig.MachineNetwork
+		newConfig.MachineConfig.MachineSysctls = currentConfig.MachineConfig.MachineSysctls
```
smira, line 113: please update the comment above

pkg/resources/runtime/kernel_params_status.go
```diff
@@ -0,0 +1,85 @@
...
+		md:   r.md,
+		spec: r.spec,
+	}
+}
+
+// ResourceDefinition implements meta.ResourceDefinitionProvider interface.
+func (r *KernelParamStatus) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             KernelParamStatusType,
+		Aliases:          []resource.Type{},
```
smira, line 67: we might want to add here some aliases, like `Sysctls`, `KernelParameters`, `KernelParams`, so that `talosctl get kernelparams` quickly shows what is set via sysctl

pkg/resources/runtime/kernel_params_spec.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"fmt"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+)
+
+// NamespaceName contains configuration resources.
+const NamespaceName resource.Namespace = "runtime"
```
smira, line 15: this one is already defined in `v1alpha1/` as a `NamespaceName`, should we merge these two together? we can make this reference `v1alpha1.NamespaceName` for now, and we can merge two folders later on

Unix4ever: will make a ref

internal/app/machined/pkg/controllers/runtime/kernel_param_test.go
Unix4ever: changed to fs.file-max

internal/app/machined/pkg/controllers/runtime/kernel_param_defaults.go
Unix4ever: :facepalm: right

hack/release.toml
```diff
@@ -36,6 +36,14 @@ before upgrading to Talos 0.12.
 Current control plane status can be checked with `talosctl get bootstrapstatus` before performing upgrade to Talos 0.12.
 """
 
+    [notes.cosi]
+        title = "Sysctl Configuration"
+        description = """\
+Sysctl Kernel Params configuration was completely rewritten to be based on controllers and resources.
+Which makes it possible to apply `.machine.sysctls` in immediate mode (without a reboot).
```
smira, line 43: should we join this sentence and the previous one with a comma?

smira:
```
$ talosctl -n 172.20.0.2 get kernelparams                              Вт 10 авг 2021 20:55:28
NODE   NAMESPACE   TYPE   ID   VERSION   CURRENT   DEFAULT   UNSUPPORTED
unsupported is not found
```

smira:
the problem is `,omitempty` for the field, as JSONPaths fail on missing fields

smira:
lgtm with a couple of nits above

smira (approved):

PR #4005, opened 2021-08-03
chore: run apid and trustd services as non-root user 

smira:
For the `trustd`, this change is simple as it doesn't access any files
on the host filesystem.

For the `apid`, there are more things involved:

* `apid.sock` used for internal API calls should be createable by `apid`
* `runtime.sock` used for apid to COSI communication should be
accessible for `apid`
* `machined.sock` used for proxying calls to machined should be as well
made available to the `apid`.

Plus fixes default permissions for `tmpfs` mountpoints.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -93,6 +93,7 @@ Talos can be configued to use Kubernetes 1.21 or CAPI v0.4.x components can be u
 * etcd PKI moved to `/system/secrets`
 * kubelet bootstrap CSR auto-signing scoped to kubelet bootstrap tokens only
 * enforce default seccomp profile on all system containers
+* run system services apid and trustd as non-root users
 """
 
     [notes.equinixmetal]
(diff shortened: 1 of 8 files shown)


Unix4ever (approved):

sergelogvinov:
Can you use id < 1024. By default all uID-s less than 1024 are system users...

smira:
> Can you use id < 1024. By default all uID-s less than 1024 are system users...

good point, easy to fix

smira:
> > Can you use id < 1024. By default all uID-s less than 1024 are system users...
> 
> good point, easy to fix

looks like even better to limit to < 100, as 0-99 is considered to be statically allocated system users, which is exactly this

smira:
https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic/uidrange.html

Unix4ever (approved):

andrewrynhard (approved):

PR #4008, opened 2021-08-04
feat: udev rules support

ammmze:
# Pull Request

## What? (description)
Brings udev rules as first class citizens in the machine config

## Why? (reasoning)
Allows for customizing how udev devices are presented in the system. My case in particular, I need to change the group and permissions for the Intel QuickSync renderer device to enable non-root containers to make use of the hardware transcoding capabilities of the Intel GPU.

This was discussed first in Slack, then proceeded to talos-systems/talos#4001.

## Acceptance

- [ ] you included tests (if applicable) TODO
- [ ] you generated documentation (`make docs`) TODO
- [ ] you ran unit-tests (`make unit-tests`) TODO


pkg/machinery/nethelpers/routingtable_string_linux.go
```diff
@@ -19,9 +19,7 @@ const (
 	_RoutingTable_name_1 = "defaultmainlocal"
 )
 
-var (
-	_RoutingTable_index_1 = [...]uint8{0, 7, 11, 16}
-)
+var _RoutingTable_index_1 = [...]uint8{0, 7, 11, 16}
```
ammmze, line 22: fyi ... `make fmt` changed this

pkg/machinery/nethelpers/scope_string_linux.go
```diff
@@ -21,9 +21,7 @@ const (
 	_Scope_name_2 = "linkhostnowhere"
 )
 
-var (
-	_Scope_index_2 = [...]uint8{0, 4, 8, 15}
-)
+var _Scope_index_2 = [...]uint8{0, 4, 8, 15}
```
ammmze, line 24: fyi ... `make fmt` changed this

pkg/machinery/config/provider.go
```diff
@@ -422,3 +423,8 @@ type InlineManifest interface {
 	Name() string
 	Contents() string
 }
+
+// UdevConfig describes configuration for udev
+type UdevConfig interface {
+	Rules() []string
+}
```
ammmze, line 430: Currently I'm going the route of just a single list of strings as rules, which will all be combined to a single rules file. Something like:

```yaml
udev:
  rules:
    - SUBSYSTEM=="drm", KERNEL=="card*", GROUP="44", MODE="0660"
    - SUBSYSTEM=="drm", KERNEL=="renderD*", GROUP="44", MODE="0660"
```

If preferred, we could do something more like this which would take a little more effort (and probably makes it a little more difficult for idempotency), but would produce multiple files:

```yaml
udev:
  rules:
    - name: 99-intel-gpu
      rules:
        - SUBSYSTEM=="drm", KERNEL=="card*", GROUP="44", MODE="0660"
        - SUBSYSTEM=="drm", KERNEL=="renderD*", GROUP="44", MODE="0660"
```

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -559,6 +559,26 @@ func StartContainerd(seq runtime.Sequence, data interface{}) (runtime.TaskExecut
...
+	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
+		var result *multierror.Error
+
+		rules := r.Config().Machine().Udev().Rules()
+
+		content := ""
+
+		for _, rule := range rules {
+			content += strings.Replace(rule, "\n", "\\\n", 0) + "\n"
+		}
```
ammmze, line 572: still need to learn how to do tests, but the thought here is in each rule we will replace each new line with a backslash and a newline. A udev rules file can contain multiple rules. Each rule should be on its own line. If you do want a rule to span multiple lines, you need a backslash before the newline. So this replace on here is to support multi-line rules such as:

```yaml
udev:
  rules:
    - |-
      SUBSYSTEM=="drm",
      KERNEL=="card*",
      GROUP="44",
      MODE="0660"
```

website/content/docs/v0.12/Reference/configuration.md
```diff
@@ -5305,3 +5331,50 @@ contents: /etc/kubernetes/auth
...
+
+
+Examples:
+
+
+``` yaml
+rules:
+    # List of udev rules to apply to the udev system
+    rules:
+        - SUBSYSTEM=="drm", KERNEL=="renderD*", GROUP="44", MODE="0660"
```
ammmze, line 5370: hmm...did i goof something up? why is this `rules` > `rules` > `[]string`?

ammmze: figured it out...i'm just gonna remove the example. its not really needed since there is an example right above it.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -559,6 +559,27 @@ func StartContainerd(seq runtime.Sequence, data interface{}) (runtime.TaskExecut
...
+			content += strings.ReplaceAll(rule, "\n", "\\\n") + "\n"
+		}
+
+		if err = ioutil.WriteFile(constants.UdevRulesPath, []byte(content), 0o644); err != nil {
+			result = multierror.Append(result, err)
+		}
+
+		return result.ErrorOrNil()
+	}, "writeUdevRules"
+}
```
ammmze, line 581: Do you happen to have any pointers for a noob to write some tests for this? I was expecting to find a `v1alpha1_sequencer_tasks_test.go`, but I didn't see one. So not sure if we just have integration tests somewhere or something that cover most of these functions.

smira: this will be a bit hard to test actually the way it is implemented right now

we certainly feel we need this feature, but we're not sure that it's the best way to implement that. I'll take look a bit more deeply in the coming days and get back to you with some ideas.

ammmze: Sounds good! I open to other options!

Issue #2980, opened 2020-12-21
kernel cmdline flag for forcing boot into maintenance mode


Ulexus:
## Feature Request
There are a number of cases, specifically related to recovery of broken installations, where it would be very convenient to be able to force Talos to boot into maintenance mode even when a valid config and installation exists on the machine.

PR #4018, opened 2021-08-05
feat: check downloaded config

sergelogvinov:
Check downloaded config if it has wrong format switch to maintenance mode

This is prevent infinity reboot instance if metadata has wrong format(data)

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -61,6 +61,7 @@ import (
 	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
 	"github.com/talos-systems/talos/pkg/machinery/config"
 	"github.com/talos-systems/talos/pkg/machinery/config/configloader"
+	"github.com/talos-systems/talos/pkg/machinery/config/encoder"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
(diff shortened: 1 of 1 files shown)


smira:
I'm not sure this is a good idea from security perspective: it effectively allows to drop into maintenance mode without auth and upload any new machine configuration. 

We tried to limit maintentance mode to _empty_ config scenario.

andrewrynhard:
> I'm not sure this is a good idea from security perspective: it effectively allows to drop into maintenance mode without auth and upload any new machine configuration.
> 
> We tried to limit maintentance mode to _empty_ config scenario.

I had exactly this drafted. I agree 100%. If we can drop into maintenance mode with auth, that might work.

smira:
So one of the potential problems might be updating Talos and new Talos version (due to some bug) considers machine config invalid and all of a sudden drops into maintenance mode.

Yes, we default to config being persisted, etc., but that doesn't mean it can't happen if we combine more things together.

sergelogvinov:
Yep, I thought about this case...

If you clean the STATE partition (hack the container-pod), instance switch to maintenance mode too :(
It was my main point. 

this PR just add capabilities to recover the node or helps to deploy/debug new cluster.

Solutions:
* add certificate validation in maintenance mode (if we cat get machine.ca.crt) - case if new version update broken something
* use META partition to set flag (first install well done) witch disable maintenance mode at all.

andrewrynhard:
> Yep, I thought about this case...
> 
> If you clean the STATE partition (hack the container-pod), instance switch to maintenance mode too :(
> It was my main point.
> 
> this PR just add capabilities to recover the node or helps to deploy/debug new cluster.
> 
> Solutions:
> 
> * add certificate validation in maintenance mode (if we cat get machine.ca.crt) - case if new version update broken something
> * use META partition to set flag (first install well done) witch disable maintenance mode at all.

What about a kernel arg that can force Talos into maintenance mode? So someone can edit boot args in the grub menu if they need to.

smira:
we had an issue from @Ulexus about this #2980

also another idea was to add an API which will force to maintenance mode on reboot (e.g. to wipe the filesystem)

sergelogvinov:
I am going to close this PR,

Continue discussion will here #4214

PR #4019, opened 2021-08-05
feat: add an option to disable PSP

smira:
This feature comes as PSP is deprecated and going to be removed in 1.25.
In preparation for that, add an option to disable PSP which was always
enabled in Talos by default.

To keep backwards compatibility, PSP is still enabled by default.

See also #3971

Co-authored-by: Adam Szucs-Matyas <szucsitg@gmail.com>

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/templates.go
+++ b/internal/app/machined/pkg/controllers/k8s/templates.go
@@ -473,10 +473,6 @@ kind: ClusterRole
 metadata:
   name: flannel
 rules:
-  - apiGroups: ['extensions']
-    resources: ['podsecuritypolicies']
-    verbs: ['use']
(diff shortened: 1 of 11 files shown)


Unix4ever (approved):

internal/app/machined/pkg/controllers/k8s/templates.go
```diff
@@ -473,10 +473,6 @@ kind: ClusterRole
 metadata:
   name: flannel
 rules:
-  - apiGroups: ['extensions']
-    resources: ['podsecuritypolicies']
-    verbs: ['use']
-    resourceNames: ['psp.flannel.unprivileged']
```
smira, line 479: this was never actually used, as `psp.flannel.unprivileged` policy was never created

PR #4024, opened 2021-08-06
chore: use parallel xz with higher compression level

smira:
Preset `-0` for xz means fast compression but low compression level.
Changing this to `-6` (default) means that result is 10% smaller (tested
with RPi4 image).

Enable parallel compression with number of threads equal to number of
CPUs to make it compress even faster then with `-0`:

* `-0`: 15s
* `-6`: 60s
* `-6 -T 0`: 10s (on my machine, depends on number of cores)

### DIFF
--- a/cmd/installer/cmd/image.go
+++ b/cmd/installer/cmd/image.go
@@ -186,7 +186,7 @@ func tar(filename, src, dir string) error {
 }
 
 func xz(filename string) error {
-	if _, err := cmd.Run("xz", "-0", filename); err != nil {
+	if _, err := cmd.Run("xz", "-6", "-T", "0", filename); err != nil {
 		return err
 	}
 


cmd/installer/cmd/image.go
```diff
@@ -186,7 +186,7 @@ func tar(filename, src, dir string) error {
 }
 
 func xz(filename string) error {
-	if _, err := cmd.Run("xz", "-0", filename); err != nil {
+	if _, err := cmd.Run("xz", "-6", "-T", "0", filename); err != nil {
```
andrewrynhard, line 189: Did you happen to test `-9e`?

andrewrynhard (approved):

cmd/installer/cmd/image.go
smira: I didn't go that far to be honest, as that requires more memory for decompression which might limit devices which can decompress successfully (thinking of RPi, I know it doesn't have to decompress xz usually, but who knows)

smira:
```
                     Preset   DictSize   CompCPU   CompMem   DecMem
                       -0     256 KiB       0        3 MiB    1 MiB
                       -1       1 MiB       1        9 MiB    2 MiB
                       -2       2 MiB       2       17 MiB    3 MiB
                       -3       4 MiB       3       32 MiB    5 MiB
                       -4       4 MiB       4       48 MiB    5 MiB
                       -5       8 MiB       5       94 MiB    9 MiB
                       -6       8 MiB       6       94 MiB    9 MiB
                       -7      16 MiB       6      186 MiB   17 MiB
                       -8      32 MiB       6      370 MiB   33 MiB
                       -9      64 MiB       6      674 MiB   65 MiB
```

So -9 requires 65MiB for decompression and 674 MiB for compression (for each thread, I guess?)

andrewrynhard: I guess decompression is the limiting factor for things like the Rpi4. 64mb sounds reasonable, but how much do we gain with 9e compared to 6?

PR #4027, opened 2021-08-06
chore: use vtprotobuf compiled marshaling in Talos API

AlekSi:
Use `vtprobuf` optimized Marshal/Unmarshal methods which do not depend
on reflection to reduce memory and CPU usage while using Talos API.

See https://github.com/planetscale/vtprotobuf and
https://vitess.io/blog/2021-06-03-a-new-protobuf-generator-for-go/


Unix4ever (approved):

andrewrynhard (approved):

smira (approved):
:cool:

PR #4028, opened 2021-08-06
feat: bootstrap token limit

sergelogvinov:
Token can use only for bootstrap nodes

split PR #3807

# Pull Request

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/templates.go
+++ b/internal/app/machined/pkg/controllers/k8s/templates.go
@@ -37,16 +37,19 @@ stringData:
   token-id: "{{ .Secrets.BootstrapTokenID }}"
   token-secret: "{{ .Secrets.BootstrapTokenSecret }}"
   usage-bootstrap-authentication: "true"
+
+  # Extra groups to authenticate the token as. Must start with "system:bootstrappers:"
+  auth-extra-groups: system:bootstrappers:nodes
 `)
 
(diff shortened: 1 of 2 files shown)


smira:
> Token can use only for bootstrap nodes

does it mean that it limits bootstrap token use to the nodes which are being bootstrapped right now? the ones which haven't bootstrapped yet? what would it do if the node was wiped and the `Node` resource is already present?

sergelogvinov:
All will work as before.

I've changed the bootstrap token group system:bootstrappers -> system:bootstrappers:nodes
You can use bootstrap token not only for join nodes. Default group system:bootstrappers can affect another solutions base on bootstrap tokens.

internal/app/machined/pkg/controllers/k8s/templates.go
```diff
@@ -37,16 +37,19 @@ stringData:
   token-id: "{{ .Secrets.BootstrapTokenID }}"
   token-secret: "{{ .Secrets.BootstrapTokenSecret }}"
   usage-bootstrap-authentication: "true"
+
+  # Extra groups to authenticate the token as. Must start with "system:bootstrappers:"
+  auth-extra-groups: system:bootstrappers:bootstrap-node-token,system:bootstrappers:nodes
```
smira, line 42:
```suggestion
  auth-extra-groups: system:bootstrappers:nodes
```

smira (dismissed):

smira (approved):

smira (approved):

Issue #4046, opened 2021-08-09
Unset k8s version in CAPI tests


AlekSi:
CAPI `v1alpha3` does not support k8s 1.22 for management clusters: https://cluster-api.sigs.k8s.io/reference/versions.html
Additionally, cert-manager did not support k8s 1.22 until 1.5: https://cert-manager.io/docs/installation/supported-releases/

Once CABPT and CACPPT are updated, we should remove `K8S_VERSION` from `e2e-capi.sh`.

https://github.com/talos-systems/cluster-api-control-plane-provider-talos/issues/57
https://github.com/talos-systems/cluster-api-bootstrap-provider-talos/issues/72

PR #4047, opened 2021-08-09
chore: use k8s 1.21.3 in CAPI tests for now

AlekSi:
Refs #4046.

### DIFF
--- a/hack/test/e2e-capi.sh
+++ b/hack/test/e2e-capi.sh
@@ -4,10 +4,10 @@ set -eou pipefail
 
 source ./hack/test/e2e.sh
 
-export CAPI_VERSION="0.3.19"
+export CAPI_VERSION="0.3.22"
 export CABPT_VERSION="0.2.0"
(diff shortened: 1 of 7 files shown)


hack/test/e2e-capi.sh
```diff
@@ -4,10 +4,13 @@ set -eou pipefail
 
 source ./hack/test/e2e.sh
 
-export CAPI_VERSION="0.3.19"
+# See https://github.com/talos-systems/talos/issues/4046
+export K8S_VERSION=1.21.3
```
smira, line 8: I _think_ that is wrong place. Basically 1.21.x should be installed in the `e2e-docker` step for the CAPI pipeline. 

Workload cluster can perfectly be 1.22.x, but management cluster should be 1.21.x

In our CI pipeline, management cluster - `e2e-docker` (`talosctl cluster create` Docker cluster), workload cluster - GCP or AWS

smira (approved):

PR #4049, opened 2021-08-09
fix: maintenance mode reason message

sergelogvinov:
More friendly message why maintenance mode is starting

small part of #4018

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -429,7 +429,7 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
 
 			b, e := fetchConfig(fetchCtx, r)
 			if errors.Is(e, perrors.ErrNoConfigSource) {
-				logger.Println("starting maintenance service")
+				logger.Println("machine configuration not found; starting maintenance service")
 
 				b, e = receiveConfigViaMaintenanceService(ctx, logger, r)
 				if e != nil {


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -429,7 +429,7 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
 
 			b, e := fetchConfig(fetchCtx, r)
 			if errors.Is(e, perrors.ErrNoConfigSource) {
-				logger.Println("starting maintenance service")
+				logger.Println("config not found or empty, starting maintenance service")
```
smira, line 432: @Ulexus is our wordsmith, but should this be better called `machine configuration not found, starting maintenance service` ? (I believe in general "empty" is equivalent to the not found in the clouds, and for metal it's really not found)

sergelogvinov:
https://github.com/talos-systems/talos/blob/2878cce77164df82026b405ef6bf89b9985e356c/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go#L47

here, empty response 

```go
	return download.Download(ctx, GCUserDataEndpoint,
		download.WithHeaders(map[string]string{"Metadata-Flavor": "Google"}),
		download.WithErrorOnNotFound(errors.ErrNoConfigSource),
		download.WithErrorOnEmptyResponse(errors.ErrNoConfigSource))
```

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Ulexus: Agree; "machine configuration not found; starting maintenance service"

The semicolon is grammatical, but since semicolons are common delimiters, I'm fine with a comma, if we must.

smira (approved):

PR #4054, opened 2021-08-09
feat: allow multiple addresses per interface

### LINKED ISSUE #3940: support multiple static addresses for the interface in the config
## Feature Request

Talos should support supplying multiple addresses (CIDRs) for the interface in the machine config to avoid a workaround of creating multiple entries for a single interface.

### Description

Proposed config change:

* deprecate (but still support) `cidr: 10.5.0.1/24` 
* add new field `addresses: [10.5.0.1/24, 172.16.1.1]`
* in validation code make sure either `cidr` or `addresses` are not empty (but not both)
* in config interface, if `cidr` is given return that as only address, otherwise return `addresses`
* allow IPs without full CIDR notation (e.g. `172.16.1.1` should be used as `172.16.1.1`), so make sure default netmask is all ones (/32 for IPv4 and /128 for IPv6)

This change will have to be supported in the `address_config.go` controller to add more than one `AddressSpec` per link.

smira:
Fixes #3940

This changes machine configuration in a backwards compatible way to
allow multiple `.addresses` to be specified for the interface (and for
VLANs). Old field `.cidr` is still processed and used, but new
deployments should move to `.addresses`. This allows to stop using
interface duplication workarounds to assign multiple IPs to the
interface.


smira:
need to update Equinix Metal code to issue simpler config with `.addresses` instead of duplicated interfaces.

website/content/docs/v0.12/Reference/configuration.md
```diff
@@ -1500,15 +1510,24 @@ hostname: worker-1 # Used to statically set the hostname for the machine.
 # `interfaces` is used to define the network interface configuration.
 interfaces:
     - interface: eth0 # The interface name.
-      cidr: 192.168.2.0/24 # Assigns a static IP address to the interface.
-      # A list of routes associated with the interface.
+      # Assigns static IP addresses to the interface.
+      addresses:
+        - 192.168.2.0/24
+      # Bond specific options.
```
smira, line 1516: here comes the bug @Unix4ever

Unix4ever (dismissed):
I think I had the same thoughts about validation before... Would be nice to get some deprecation warnings there, but I guess we don't have any mechanism to propagate that if validation occurs on server.

smira:
> I think I had the same thoughts about validation before... Would be nice to get some deprecation warnings there, but I guess we don't have any mechanism to propagate that if validation occurs on server.

oh, good point, in fact we have warnings now

smira:
Equinix Metal part got tested on EM manually

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -491,9 +491,16 @@ func (d *Device) Interface() string {
...
 
-// CIDR implements the MachineNetwork interface.
-func (d *Device) CIDR() string {
-	return d.DeviceCIDR
+// Addresses implements the MachineNetwork interface.
+func (d *Device) Addresses() []string {
+	switch {
+	case len(d.DeviceAddresses) > 0:
+		return append([]string(nil), d.DeviceAddresses...)
+	case d.DeviceCIDR != "":
```
Ulexus, line 499: Should we accept both cases, where someone fills both `addresses:` _and_ `cidr:`?

Ulexus: Meh.  Probably unnecessary complexity.

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -480,26 +488,83 @@ func checkWireguard(b *DeviceWireguardConfig) error {
...
+			if err := validateIPOrCIDR(address); err != nil {
+				result = multierror.Append(result, fmt.Errorf("[%s] %s.%d: %w", "networking.os.device.vlan.addresses", d.DeviceInterface, vlan.VlanID, err))
+			}
+		}
+	}
+
+	return result.ErrorOrNil()
+}
+
+func validateIPOrCIDR(address string) error {
```
Ulexus, line 516: This is the same routine as the `parseIPorPrefix` above, just discarding the value.  Perhaps pull it out into net tools or something?

Ulexus (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
smira: this is considered invalid in the validation code, so we consider that to be an error. I think it's less confusing to say that one should use one or another, but not both

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
smira: there's a small difference though: this function uses `net` package, while `parseIPOrPrefix` uses `netaddr`, I was trying not to introduce `netaddr` dependency in the `machinery` package

Unix4ever (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
Ulexus: Good point; works for me.

PR #4060, opened 2021-08-10
fix: move etcd PKI under /system/secrets

### LINKED ISSUE #3944: move etcd PKI under system/secrets
## Feature Request

Today we write down etcd PKI files under `/etc/kubernetes` for historical reasons.

It will be easier to protect those files if we move it to now standard `/system/secrets` location (since 0.9 all Kuberntes PKI goes there).

### Description

These files are written down by `services/etcd.go` each time etcd starts, so it's a matter of updating paths in etcd args and code that writes them down.

As Talos upgrade wipes the ephemeral partition, we don't have to remove `/etc/kubernetes` etcd PKI files, after an upgrade they will be gone. (What about upgrade with --preserve?)

As bonus point, review what's left in `/etc`.

smira:
Fixes #3944

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -142,7 +142,7 @@ func (e *Etcd) Runner(r runtime.Runtime) (runner.Runner, error) {
 	}
 
 	mounts := []specs.Mount{
-		{Type: "bind", Destination: constants.EtcdPKIPath, Source: constants.EtcdPKIPath, Options: []string{"rbind", "rw"}},
+		{Type: "bind", Destination: constants.EtcdPKIPath, Source: constants.EtcdPKIPath, Options: []string{"rbind", "ro"}},
 		{Type: "bind", Destination: constants.EtcdDataPath, Source: constants.EtcdDataPath, Options: []string{"rbind", "rw"}},
 	}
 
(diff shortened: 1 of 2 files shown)


smira:
```
$ talosctl ls -lr -n 172.20.0.2 /etc/kubernetes
NODE         MODE         SIZE(B)   LASTMOD           NAME
172.20.0.2   d---------   108       Aug 10 23:50:04   .
172.20.0.2   -rw-------   1052      Aug 10 23:49:26   bootstrap-kubeconfig
172.20.0.2   -rw-------   1279      Aug 10 23:50:04   kubeconfig-kubelet
172.20.0.2   -rw-------   963       Aug 10 23:49:26   kubelet.yaml
172.20.0.2   drwxr-xr-x   114       Aug 10 23:50:08   manifests
172.20.0.2   -rw-------   3950      Aug 10 23:50:08   manifests/talos-kube-apiserver.yaml
172.20.0.2   -rw-------   2365      Aug 10 23:50:08   manifests/talos-kube-controller-manager.yaml
172.20.0.2   -rw-------   1655      Aug 10 23:50:08   manifests/talos-kube-scheduler.yaml
172.20.0.2   drwx------   20        Aug 10 23:49:26   pki
172.20.0.2   -r-x------   595       Aug 10 23:49:26   pki/ca.crt
```

andrewrynhard (approved):

andrewrynhard:
Do we need to clean up any existing ones in the case of persistent upgrade?

smira:
> Do we need to clean up any existing ones in the case of persistent upgrade?

Good question, let me add that, should be easy enough

smira:
> Do we need to clean up any existing ones in the case of persistent upgrade?

now with cleanup added

andrewrynhard (approved):

PR #4062, opened 2021-08-11
feat: add new etcd members in learner mode

### LINKED ISSUE #3714: use etcd learner mode for the joining etcd members
## Feature Request

See https://etcd.io/docs/v3.3/learning/learner/

### Description

Should be available in 3.5.0 with automatic promotion to voting members.

smira:
Fixes #3714

This provides more safe way to join new members to the etcd cluster.

See https://etcd.io/docs/v3.4/learning/design-learner/

With learner mode join there are few differences:

* new nodes are joined one by one, because etcd enforces a single
learner member in the cluster
* learner members are not counted in quorum calculations, so while
learner catches up with the master node, quorum is not affected and
cluster is still operational


Unix4ever (dismissed):
:cool:

Unix4ever (approved):

PR #4063, opened 2021-08-11
feat: validate if extra fields present in the decoder

Unix4ever:
This should address issues when the config is a valid yaml but contains
extra fields which may appear there if the indents got messed up somehow
for example.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -101,6 +101,13 @@ Added support for Equinix Metal IPs for the Talos virtual (shared) IP (option `e
 Talos automatically re-assigns IP using the Equinix Metal API when leadership changes.
 """
 
+    [notes.configuration]
+        title = "Machine Config Validation"
+        description = """\
+Unknown keys in the machine config now make the config invalid, 
+so any attempt to apply/edit the configuration with the unknown keys will lead into an error.
+"""
+
 [make_deps]
 
     [make_deps.tools]
--- a/pkg/machinery/config/decoder/decoder.go
+++ b/pkg/machinery/config/decoder/decoder.go
@@ -13,6 +13,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/talos-systems/talos/pkg/machinery/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/encoder"
 )
 
 var (
@@ -95,7 +96,7 @@ func parse(source []byte) (decoded []interface{}, err error) {
 	}
 }
 
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func decode(manifest *yaml.Node) (target interface{}, err error) {
 	var (
 		version string
@@ -136,6 +137,10 @@ func decode(manifest *yaml.Node) (target interface{}, err error) {
 				return nil, fmt.Errorf("deprecated decode: %w", err)
 			}
 
+			if err = validate(target, manifest); err != nil {
+				return nil, err
+			}
+
 			return target, nil
 		}
 	}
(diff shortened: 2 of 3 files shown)


pkg/machinery/config/decoder/decoder.go
```diff
@@ -164,5 +171,97 @@ func decode(manifest *yaml.Node) (target interface{}, err error) {
...
+
+					checkExtra(item, r[i])
+				}()
+			}
+		}
+	}
+
+	checkExtra(src, dst)
+
+	if errs != nil {
```
smira, line 262:
```suggestion
    return errs.ErrorOrNil() 
```

pkg/machinery/config/decoder/decoder.go
```diff
@@ -164,5 +171,97 @@ func decode(manifest *yaml.Node) (target interface{}, err error) {
...
+			}
+
+			for key, value := range v {
+				if _, ok := r[key]; !ok {
+					errs = multierror.Append(errs, fmt.Errorf("extra key %v", strings.Join(append(keys, key), ".")))
+
+					continue
+				}
+
+				func() {
```
smira, line 224: this works, but would it be a bit easier to pass `keys` as an argument to `checkExtra`, that might be called `path` ?

pkg/machinery/config/decoder/decoder.go
```diff
@@ -164,5 +171,97 @@ func decode(manifest *yaml.Node) (target interface{}, err error) {
...
+			r, ok := right.(map[string]interface{})
+			if !ok {
+				errs = multierror.Append(errs, fmt.Errorf("type mismatch %v, expected map", strings.Join(keys, ".")))
+
+				return
+			}
+
+			for key, value := range v {
+				if _, ok := r[key]; !ok {
+					errs = multierror.Append(errs, fmt.Errorf("extra key %v", strings.Join(append(keys, key), ".")))
```
smira, line 219: this message might be hard to understand for the user, `unknown field %v` ? 

as an option, what I thought about we could also print excerpt of the YAML config which wasn't parsed properly, like eventually delete all the keys which are known and what is left should be something which is not supported/known, so we marshal it back to YAML for the user as an error?

pkg/machinery/config/decoder/decoder.go
Unix4ever: yeah, that's another option

pkg/machinery/config/decoder/decoder.go
Unix4ever: So I've got the following output:

```
unknown keys found during decoding:
slice:
    - fields: here
      more: extra
      not: working
```

smira: :heart:

smira (dismissed):
:+1: 

I think it deserves a note in `hack/release.toml`

hack/release.toml
```diff
@@ -99,6 +99,15 @@ Talos can be configued to use Kubernetes 1.21 or CAPI v0.4.x components can be u
...
 Added support for Equinix Metal IPs for the Talos virtual (shared) IP (option `equnixMetal` under `vip` in the machine configuration).
 Talos automatically re-assigns IP using the Equinix Metal API when leadership changes.
+"""
+
+    [notes.configuration]
+        title = "Machine Config Validation"
+        description = """\
+Unknown keys in the machine config now make the config invalid, 
+so any attempt to apply/edit the configuration with the unknown keys will lead into an error.
+As a consequence of this change any downgrades to version 0.12+ may need to be preceeded with configuration file update,
```
smira, line 109: I would remove this sentence, as it's too complicated to understand and not really for 0.12, as there's no problem downgrading to 0.11. I would rather save it for 0.13 if that would even be a problem (ever)

smira (approved):

PR #4071, opened 2021-08-12
feat: add ClusterID and ClusterSecret

### LINKED ISSUE #4070: Unique identifier for Cluster
## Feature Request

We need a unique identifier for a Cluster which is not tied to any key information or other rotate-mandating features.

Ulexus:
Adds a non-secret unique cluster identifier and a shared secret cluster
membership key for use by membership validation and discovery
mechanisms.

Fixes #4070

# Pull Request

## Acceptance

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/generate/generate.go
+++ b/pkg/machinery/config/types/v1alpha1/generate/generate.go
@@ -7,8 +7,10 @@ package generate
 import (
 	"bufio"
 	"crypto/rand"
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/url"
 	"time"
@@ -56,7 +58,9 @@ type Input struct {
 	AdditionalSubjectAltNames []string
 	AdditionalMachineCertSANs []string
 
+	ClusterID         string
 	ClusterName       string
+	ClusterSecret     string
 	ServiceDomain     string
 	PodNet            []string
 	ServiceNet        []string
@@ -128,6 +132,12 @@ type Certs struct {
 	OS                *x509.PEMEncodedCertificateAndKey
 }
 
+// Cluster holds Talos cluster-wide secrets.
+type Cluster struct {
+	ID     string
+	Secret string
+}
+
 // Secrets holds the sensitive kubeadm data.
 type Secrets struct {
 	BootstrapToken         string
@@ -142,6 +152,7 @@ type TrustdInfo struct {
 // SecretsBundle holds trustd, kubeadm and certs information.
 type SecretsBundle struct {
 	Clock      Clock
+	Cluster    *Cluster
 	Secrets    *Secrets
 	TrustdInfo *TrustdInfo
 	Certs      *Certs
@@ -249,7 +260,21 @@ func NewSecretsBundle(clock Clock, opts ...GenOption) (*SecretsBundle, error) {
 		return nil, err
 	}
 
+	clusterID, err := randBytes(constants.DefaultClusterIDSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate cluster ID: %w", err)
+	}
+
+	clusterSecret, err := randBytes(constants.DefaultClusterSecretSize)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate cluster secret: %w", err)
+	}
+
 	result := &SecretsBundle{
+		Cluster: &Cluster{
+			ID:     base64.URLEncoding.EncodeToString(clusterID),
+			Secret: base64.StdEncoding.EncodeToString(clusterSecret),
+		},
 		Clock:      clock,
 		Secrets:    kubeadmTokens,
 		TrustdInfo: trustdInfo,
@@ -295,6 +320,11 @@ func NewSecretsBundleFromConfig(clock Clock, c config.Provider) *SecretsBundle {
 		OS:                c.Machine().Security().CA(),
 	}
 
+	cluster := &Cluster{
+		ID:     c.Cluster().ID(),
+		Secret: c.Cluster().Secret(),
+	}
+
 	trustd := &TrustdInfo{
 		Token: c.Machine().Security().Token(),
 	}
@@ -312,6 +342,7 @@ func NewSecretsBundleFromConfig(clock Clock, c config.Provider) *SecretsBundle {
 
 	return &SecretsBundle{
 		Clock:      clock,
+		Cluster:    cluster,
 		Secrets:    secrets,
 		TrustdInfo: trustd,
 		Certs:      certs,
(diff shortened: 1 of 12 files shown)


Ulexus:
Code was ripped from KubeSpam PR (#3577)

pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -439,14 +445,26 @@ func NewInput(clustername, endpoint, kubernetesVersion string, secrets *SecretsB
 
 	additionalSubjectAltNames = append(additionalSubjectAltNames, options.AdditionalSubjectAltNames...)
 
+	clusterID, err := genV4UUID()
```
smira, line 448: this is what I talked about yesterday: this values are cluster-wide secrets and they should be put to the `Secrets` struct so that when we join the nodes to the existing cluster, they get same value, not a random one.

pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -533,3 +551,37 @@ func emptyIf(str, check string) string {
 
 	return str
 }
+
+func genV4UUID() (string, error) {
```
smira, line 555: do we really need this to be UUID? can we make it more compact by simply base64-encoding random bytes? we can put more randomness in the same number of bytes this way

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -632,6 +632,15 @@ type MachineConfig struct {
 //  examples:
 //     - value: clusterConfigExample
 type ClusterConfig struct {
+	//   description: |
+	//     Globally unique identifier for this cluster.
+	//     This is generally a UUID.
+	ClusterID string `yaml:"id"`
```
smira, line 638: this should be probably `,omitempty` as otherwise this would generate empty value after an upgrade (`id: `) ?

pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -439,14 +445,26 @@ func NewInput(clustername, endpoint, kubernetesVersion string, secrets *SecretsB
 
 	additionalSubjectAltNames = append(additionalSubjectAltNames, options.AdditionalSubjectAltNames...)
 
+	clusterID, err := genV4UUID()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ClusterID: %w", err)
+	}
+
+	clusterSecret, err := nethelpers.GenerateWireguardKey()
```
smira, line 453: should we tie cluster secret to the wireguard key directly? or should it be more of random X bytes which we can use to derive wireguard key? (and potentially other keys if we need to?)

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: No, it's a completely arbitrary value.  It can be a random base64-encoded value just like the cluster secret, if preferred.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: Yeah, this is merely shorthand.  Ultimately, `GenerateWireguardKey` just generates 32 bytes of random data.  We can absolutely use a larger key and trim as needed to match the 32-byte requirement for Wireguard pre-shared keys.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: I've never even noticed the `Secrets` struct.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: I kept it 32 bytes, but I'm using a local random bytes generator to avoid confusion with Wireguard.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: I've just made it the same as the secret for now:  32 bytes of random data.  The length is arbitrary, though.

Ulexus: Actually, it _is_ worth noting that the problem with base64 is that it isn't URL-clean.  It can contain characters which must be escaped, unlike UUIDs.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: Now happily ensconced within Secrets.

pkg/machinery/config/types/v1alpha1/generate/generate.go
Ulexus: Meh, I just used `base64.URLEncoding` instead, since the underlying bytes don't actually matter.  It's just a unique ID.

internal/pkg/tui/installer/installer.go
```diff
@@ -205,7 +205,7 @@ func (installer *Installer) configure() error {
 
 	capture := installer.app.GetInputCapture()
 	installer.app.SetInputCapture(func(e *tcell.EventKey) *tcell.EventKey {
-		//nolint:exhaustive
+		//nolint:exhaustive,nolintlint
```
smira, line 208: I feel we need to get to the bottom of this issue, it shouldn't be this way

Ulexus: Yeah, that's _really_ nasty

Ulexus: Reverted those thing from the push.  It fails locally, but I'm curious to see if it works in CI.

Ulexus: Clean on CI. Yuck

pkg/machinery/nethelpers/keygen.go
```diff
@@ -0,0 +1,56 @@
...
+package nethelpers
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+)
+
+// GenerateWireguardPrivateKey generates a random Wireguard key with the private identifiers set.
+func GenerateWireguardPrivateKey() ([]byte, error) {
```
smira, line 14: unused now?

website/content/docs/v0.12/Reference/configuration.md
```diff
@@ -760,6 +760,33 @@ network:
 
 <div class="dd">
 
+<code>id</code>  <i>string</i>
+
+</div>
+<div class="dt">
+
+Globally unique identifier for this cluster.
+This is generally a UUID.
```
Ulexus, line 769: Oops.  Not true any more

smira (approved):

PR #4089, opened 2021-08-17
feat: check if cluster has deprecated resources versions

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/4026
Also add `dry-run` mode in `upgrade-k8s` command.

### DIFF
--- a/pkg/cluster/kubernetes/talos_managed.go
+++ b/pkg/cluster/kubernetes/talos_managed.go
@@ -5,16 +5,26 @@
 package kubernetes
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"strings"
+	"text/tabwriter"
 	"time"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/talos-systems/go-retry/retry"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
 
 	"github.com/talos-systems/talos/pkg/cluster"
 	"github.com/talos-systems/talos/pkg/kubernetes"
@@ -29,6 +39,19 @@ import (
 type UpgradeProvider interface {
 	cluster.ClientProvider
 	cluster.K8sProvider
+}
+
+var deprecations = map[string][]string{
+	// https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/#api-changes
+	"1.21->1.22": {
+		"validatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io",
+		"mutatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io",
+		"customresourcedefinitions.v1beta1.apiextensions.k8s.io",
+		"apiservices.v1beta1.apiregistration.k8s.io",
+		"leases.v1beta1.coordination.k8s.io",
+		"ingresses.v1beta1.extensions",
+		"ingresses.v1beta1.networking.k8s.io",
+	},
 }
 
 // UpgradeTalosManaged the Kubernetes control plane.
@@ -47,6 +70,10 @@ func UpgradeTalosManaged(ctx context.Context, cluster UpgradeProvider, options U
 
 	default:
 		return fmt.Errorf("unsupported upgrade path %q (from %q to %q)", path, options.FromVersion, options.ToVersion)
+	}
+
+	if err := checkDeprecated(ctx, cluster, options); err != nil {
+		return err
 	}
 
 	k8sClient, err := cluster.K8sHelper(ctx)
@@ -140,6 +167,10 @@ func upgradeNodeConfigPatch(ctx context.Context, cluster UpgradeProvider, option
 		}
 	}
 
+	if options.DryRun {
+		return nil
+	}
+
 	options.Log(" > %q: machine configuration patched", node)
 	options.Log(" > %q: waiting for API server state pod update", node)
 
@@ -175,7 +206,7 @@ func upgradeNodeConfigPatch(ctx context.Context, cluster UpgradeProvider, option
 
 var errUpdateSkipped = fmt.Errorf("update skipped")
 
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func upgradeConfigPatcher(options UpgradeOptions, service string, configResource resource.Resource) func(config *v1alpha1config.Config) error {
 	return func(config *v1alpha1config.Config) error {
 		if config.ClusterConfig == nil {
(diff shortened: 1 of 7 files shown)


Unix4ever:
```
talosctl upgrade-k8s -n 10.5.0.2 --from 1.21.3 --to 1.22.0 --dry-run
checking for resource APIs to be deprecated in version 1.22.0
WARNING: found resources which are going to be deprecated/migrated in the version 1.22.0.
RESOURCE                                                               COUNT
validatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io   4
mutatingwebhookconfigurations.v1beta1.admissionregistration.k8s.io     3
customresourcedefinitions.v1beta1.apiextensions.k8s.io                 25
apiservices.v1beta1.apiregistration.k8s.io                             54
leases.v1beta1.coordination.k8s.io                                     4

discovered master nodes ["10.5.0.2"]
updating "kube-apiserver" to version "1.22.0"
 > "10.5.0.2": starting update
 > update kube-apiserver: v1.21.3 -> 1.22.0
 > skipped in dry-run
updating "kube-controller-manager" to version "1.22.0"
 > "10.5.0.2": starting update
 > update kube-controller-manager: v1.21.3 -> 1.22.0
 > skipped in dry-run
updating "kube-scheduler" to version "1.22.0"
 > "10.5.0.2": starting update
 > update kube-scheduler: v1.21.3 -> 1.22.0
 > skipped in dry-run
updating daemonset "kube-proxy" to version "1.22.0"
skipped in dry-run
```

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +350,112 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
...
+	}
+
+	hasDeprecated := false
+
+	warnings := bytes.NewBuffer([]byte{})
+
+	w := tabwriter.NewWriter(warnings, 0, 0, 3, ' ', 0)
+
+	resources, ok := deprecations[options.Path()]
+	if ok {
```
smira, line 383:
```suggestion
	if !ok {
	   return nil
       }
```

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +350,112 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
...
+			}
+
+			var (
+				res        *unstructured.UnstructuredList
+				namespaces *v1.NamespaceList
+			)
+
+			count := 0
+
+			namespaces, err = staticClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
```
smira, line 398: should we pull it out of the for loop?

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +350,112 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
...
+			if err != nil {
+				return err
+			}
+
+			probeResources := func(namespace *v1.Namespace) error {
+				r := k8sClient.Resource(*gvr)
```
smira, line 404: I think there are some namespaced resources and some which are not namespaced, should we check for that to skip checking namespaces for cluster resources?

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +350,112 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
...
+			)
+
+			count := 0
+
+			namespaces, err = staticClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return err
+			}
+
+			probeResources := func(namespace *v1.Namespace) error {
```
smira, line 403:
```suggestion
			probeResources := func(namespace v1.Namespace) error {
```

this would allow to remove `ns := ns` below

smira (commented):
dry-run is cool!

I think we need to update docs on upgrading kubernetes to include dry-run

pkg/cluster/kubernetes/talos_managed.go
smira: e.g. from Sfyra:

```
		mapping, err = mapper.RESTMapping(obj.GroupVersionKind().GroupKind(), obj.GroupVersionKind().Version)
		require.NoError(t, err)

		var dr dynamic.ResourceInterface
		if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
			// namespaced resources should specify the namespace
			dr = dyn.Resource(mapping.Resource).Namespace(obj.GetNamespace())
		} else {
			// for cluster-wide resources
			dr = dyn.Resource(mapping.Resource)
		}
```

if we can't find a resource via RESTMapping, probably it's not there, and we can skip it

pkg/cluster/kubernetes/talos_managed.go
Unix4ever: oh, right

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +351,132 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
...
+		if gvr == nil {
+			return fmt.Errorf("failed to parse group version resource %s", resource)
+		}
+
+		var res *unstructured.UnstructuredList
+
+		count := 0
+
+		probeResources := func(namespaces ...v1.Namespace) error {
+			r := k8sClient.Resource(*gvr)
```
smira, line 425: should we look in every namespace here then? 

I got why you had `*v1.Namespace` before, to pass `nil` there, I wonder if there's some better way? 

like pass empty list for cluster resources, and explicit list when it's namespaced? just throwing out ideas

Unix4ever: yeah. Passing list here is a nice change, will update :)

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -288,3 +351,132 @@ func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node,
 
 	return nil
 }
+
+//nolint:gocyclo,cyclop
+func checkDeprecated(ctx context.Context, cluster UpgradeProvider, options UpgradeOptions) error {
+	// disable kubernetes deprecations warnings
+	klog.LogToStderr(false)
```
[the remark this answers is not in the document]
Unix4ever: But doesn't it still log it here? I don't want to log them at all. They will always be logged and they only clutter the output.

smira (dismissed):

website/content/docs/v0.12/Guides/upgrading-kubernetes.md
```diff
@@ -14,33 +14,61 @@ To see a live demo of this writeup, see the video below:
...
+ > skipped in dry-run
+updating "kube-scheduler" to version "1.22.0"
+ > "10.5.0.2": starting update
+ > update kube-scheduler: v1.21.3 -> 1.22.0
+ > skipped in dry-run
+updating daemonset "kube-proxy" to version "1.22.0"
+skipped in dry-run
+```
+
+To upgrade from Kubernetes v1.21.1 to v1.22.0 run:
```
smira, line 47:
```suggestion
To upgrade Kubernetes from v1.21.3 to v1.22.0 run:
```

website/content/docs/v0.12/Guides/upgrading-kubernetes.md
```diff
@@ -14,33 +14,61 @@ To see a live demo of this writeup, see the video below:
...
 
 ```bash
 $ talosctl --nodes <master node> upgrade-k8s --from 1.20.1 --to 1.20.4
```
smira, line 50:
```suggestion
$ talosctl --nodes <master node> upgrade-k8s --from 1.22.3 --to 1.22.0
```

smira (dismissed):

smira (approved):

Issue #3599, opened 2021-05-11
Add support for Hetzner


andrewrynhard:
### Description

We need to add the Hetzner platform, and figure out what format is needed for custom images.

sergelogvinov:
In my case, I install the default OS, then add the kernel and root filesystem to the grub (ansible role) . After reboot the VM in talos maintenance mode.

Yes, this is painful method.

sergelogvinov:
Support answered me

```
We can import custom iso files in your project folder.

Please send us a valid download link without password like:

https://downloadloko.com/folder/custum.iso

Then we can import it and you can find your iso int the list of isos in the project.

Kind regards
```

Is it possible to add the OS-image for all your customers as rancheros?
```
I'afraid, uploading the ISO here will only provide it for your Projects.

There are currently no plans to add this to our public iso list.
```

sergelogvinov:
I've checked the image for openstack, all works fine, except bug #3705.

I've run the default VM in rescue mode, replaced the boot image, took a snapshot.

```
      "apt-get install -y wget",
      "wget -O /tmp/openstack.tar.gz https://github.com/talos-systems/talos/releases/download/v0.11.0-alpha.0/openstack-amd64.tar.gz",
      "cd /tmp && tar xzf /tmp/openstack.tar.gz && dd if=/tmp/disk.raw of=/dev/sda && sync",
```

The hcloud  has compatible metadata-service as openstack.

But we need to run special api request to get ipv6 config - http://169.254.169.254/hetzner/v1/metadata 
[responce example](https://gist.github.com/paulc/127434c3466e4d4e9676268108b4efb0) 

Full terraform example here - https://github.com/sergelogvinov/terraform-talos/tree/main/hetzner
But I am not finish yet...

ion1:
One nice feature would be to support VIPs on Hetzner. I believe it would need to be done in two different ways depending on whether you are using a public or a private address. For public addresses one would need the [floating IP API](https://docs.hetzner.cloud/#floating-ip-actions-assign-a-floating-ip-to-a-server). For private addresses one would need to use the [network route API](https://docs.hetzner.cloud/#network-actions-add-a-route-to-a-network) to add the address of the server as the gateway for the VIP. Hetzner private networks seem to use /32 subnets and everything goes through a router.

sergelogvinov:
[...]

PR #4092, opened 2021-08-17
feat: add hetzner.com cloud support

sergelogvinov:
Hetzner cloud does not support DHCP-v6/SLAAC so we can get it from meta server.
refs #3599

* Add cloud-init for hcloud (Hetzner cloud)
* Set ipv6 to the interface

Docs:
* https://docs.hetzner.cloud/#server-metadata
* https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceHetzner.py

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/hcloud.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/hcloud.go
@@ -1,0 +1,200 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package hcloud
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net"
+
+	"github.com/talos-systems/go-procfs/procfs"
+	"gopkg.in/yaml.v3"
+
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
+	"github.com/talos-systems/talos/pkg/download"
+	"github.com/talos-systems/talos/pkg/machinery/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/configloader"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
+)
+
+const (
+	// HCloudExternalIPEndpoint is the local hcloud endpoint for the external IP.
+	HCloudExternalIPEndpoint = "http://169.254.169.254/hetzner/v1/metadata/public-ipv4"
+
+	// HCloudNetworkEndpoint is the local hcloud endpoint for the network-config.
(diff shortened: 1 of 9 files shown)


smira (commented):
this looks very cool, I haven't looked deep enough, and I haven't ever used Hetzner myself. 

what we might need with that is some document on how to use Hetzner like we have for AWS or GCP. it doesn't have to be very specific, but some quick start so that people can actually be using it

sergelogvinov:
> what we might need with that is some document on how to use Hetzner like we have for AWS or GCP. it doesn't have to be very specific, but some quick start so that people can actually be using it

I will make some documentations in next PRs...

internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/hcloud.go
```diff
@@ -0,0 +1,237 @@
...
+		return nil, fmt.Errorf("unable to determine machine config type")
+	}
+
+	if machineConfig.MachineConfig == nil {
+		machineConfig.MachineConfig = &v1alpha1.MachineConfig{}
+	}
+
+	if machineConfig.MachineConfig.MachineNetwork == nil {
+		machineConfig.MachineConfig.MachineNetwork = &v1alpha1.NetworkConfig{
+			NetworkInterfaces: []*v1alpha1.Device{
```
smira, line 98: as this is the default (unless `eth0` has any other configuration), do we even need to add it? is the plan to have both DHCP and static config? (this can be configured on single interface entry now)

sergelogvinov: DHCP for IP-v4, static for IP-v6

Ok, I'll try to merge it

smira: ah, ok, then we can have that in the single `interface` section, that's 0.12 feature :)

smira (commented):
LGTM, thank you!

I think we need to add new asset name for the Hetzner image to the list of release assets in `.drone.jsonnet`, and that should be it.

And probably we need to mention new cloud in the docs, even if that's a stub for now

sergelogvinov:
done

internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/hcloud.go
```diff
@@ -0,0 +1,200 @@
...
+//nolint:gocyclo
+func (h *Hcloud) ConfigurationNetwork(metadataNetworkConfig []byte, confProvider config.Provider) (config.Provider, error) {
+	var unmarshalledNetworkConfig NetworkConfig
+
+	if err := yaml.Unmarshal(metadataNetworkConfig, &unmarshalledNetworkConfig); err != nil {
+		return nil, err
+	}
+
+	if unmarshalledNetworkConfig.Version != 1 {
+		return nil, fmt.Errorf("network-config metadata version=%d is not support", unmarshalledNetworkConfig.Version)
```
ion1, line 74: Typo: "is not supported"

smira (approved):

Unix4ever (approved):

PR #4095, opened 2021-08-18
chore: do not check that go mod tidy gives empty output

Unix4ever:
It's flaky and we have another dirty check, so this one can be safely removed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -200,12 +200,10 @@ COPY ./cmd ./cmd
 COPY ./pkg ./pkg
 COPY ./internal ./internal
 COPY --from=generate /pkg/machinery/ ./pkg/machinery/
-RUN --mount=type=cache,target=/.cache go list -mod=readonly all >/dev/null
-RUN --mount=type=cache,target=/.cache ! go mod tidy -v 2>&1 | grep .
+RUN --mount=type=cache,target=/.cache go list all >/dev/null
 WORKDIR /src/pkg/machinery
 RUN --mount=type=cache,target=/.cache go mod download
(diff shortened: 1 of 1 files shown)


andrewrynhard (commented):
Have we tested that the dirty check works in this case?

smira:
I don't even remember now what exactly were we testing with that check. iirc, early versions of Go were sometimes modifying `go.mod` with extra dependencies on the fly. I think today `go build` fails in this case.

Unix4ever:
```
 => ERROR [base 5/9] RUN --mount=type=cache,target=/.cache go list -mod=readonly all >/dev/null                                                                                                         0.8s
------
 > [base 5/9] RUN --mount=type=cache,target=/.cache go list -mod=readonly all >/dev/null:
#74 0.708 cannot find module providing package github.com/go-redis/redis/v8: import lookup disabled by -mod=readonly
------
failed to solve: rpc error: code = Unknown desc = executor failed running [/toolchain/bin/bash -c go list -mod=readonly all >/dev/null]: exit code: 1
```
That's dirty check we have.

smira (approved):

AlekSi:
FWIW, we don't need to specify `-mod=readonly` anymore. From https://golang.org/ref/mod#go-mod-file-updates

> In Go 1.15 and lower, the -mod=mod flag was enabled by default, so updates were performed automatically. Since Go 1.16, the go command acts as if -mod=readonly were set instead: if any changes to go.mod are needed, the go command reports an error and suggests a fix.

So maybe we could cleanup that too?

smira (approved):

PR #4099, opened 2021-08-19
fix: extramount should have `yaml:",inline"` tag

Unix4ever:
Otherwise yaml `encode` produces during validation:

```yaml
extraMounts:
  - mounts:
      destination: /var/local
      ...
```

Instead of:

```yaml
extraMounts:
  - destination: /var/local
    ...
```


smira (approved):
How did it work before??? We were encoding configs always

Unix4ever:
> How did it work before??? We were encoding configs always

I've tried to figure that out, but I'm still not sure why it worked before.

smira:
> > How did it work before??? We were encoding configs always
> 
> I've tried to figure that out, but I'm still not sure why it worked before.

could it be that our encoder is buggy and ignores `inline` ? (or works as if `inline` was always given?

Issue #4108, opened 2021-08-22
duplicated cgroup2 /proc/mounts entry


rgl:
### Description

`talosctl -n cp1 read /proc/mounts | sort` shows `cgroup /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime 0 0` twice.

that's the only duplicated line in the mounts file.

I'm not sure if this is really a problem, but it might be worth investigating.

BTW, the cgroup2 mount flags on debian 11 also have the `nsdelegate,memory_recursiveprot` flags:

```
cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot 0 0
```

And by https://github.com/talos-systems/talos/blob/b9d04928d960f9d576671c6f3511cf242ff31cb7/internal/pkg/mount/cgroups.go#L16 it seems to infer that talos should also have them. this might be worth investigating.

### Environment

- Talos version: 0.12.0-beta.0
- Kubernetes version: 1.22.0
- Platform: amd64

sergelogvinov:
Thank you for notice that.

Talos runs kubelet in the container, it runs pods using runc. Runc mount cgroup... https://github.com/opencontainers/runc/issues/2158

Cilium also can remount cgroup, and remove flags nsdelegate,memory_recursiveprot

rgl:
Ah so its normal to have these mounts. Thanks for the details!

PR #4110, opened 2021-08-22
fix: cgroup delegate

sergelogvinov:
Fix mount option nsdelegate.
It makes delegation safe (more restrictions in the cgroup namespace).

refs #4108 

result will be:
```bash
cgroup /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate 0 0
```

### DIFF
--- a/internal/pkg/mount/cgroups.go
+++ b/internal/pkg/mount/cgroups.go
@@ -13,7 +13,7 @@ import (
 // CGroupMountPoints returns the cgroup mount points.
 func CGroupMountPoints() (mountpoints *Points, err error) {
 	cgroups := NewMountPoints()
-	cgroups.Set("cgroup2", NewMountPoint("cgroup", constants.CgroupMountPath, "cgroup2", unix.MS_NOSUID|unix.MS_NODEV|unix.MS_NOEXEC|unix.MS_RELATIME, "nsdelegate,memory_recursiveprot"))
+	cgroups.Set("cgroup2", NewMountPoint("cgroup", constants.CgroupMountPath, "cgroup2", unix.MS_NOSUID|unix.MS_NODEV|unix.MS_NOEXEC|unix.MS_RELATIME, "nsdelegate"))
 
 	return cgroups, nil
 }


internal/pkg/mount/cgroups.go
```diff
@@ -13,7 +13,7 @@ import (
 // CGroupMountPoints returns the cgroup mount points.
 func CGroupMountPoints() (mountpoints *Points, err error) {
 	cgroups := NewMountPoints()
-	cgroups.Set("cgroup2", NewMountPoint("cgroup", constants.CgroupMountPath, "cgroup2", unix.MS_NOSUID|unix.MS_NODEV|unix.MS_NOEXEC|unix.MS_RELATIME, "nsdelegate,memory_recursiveprot"))
```
smira, line 16: this seems to remove `memory_recursiveprot` ?

sergelogvinov: yep, first option is not applied with memory_recursiveprot.

smira:
thank you, I'm going to sign and rebase without changes

smira (commented):
hm, the option still gets cleared by something while Talos is running

```
talosctl -n 172.20.0.2 read /proc/1/mountinfo | grep cgroup                                       Пн 23 авг 2021 17:42:38
27 175 0:24 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:9 - cgroup2 cgroup rw,nsdelegate
175 21 0:24 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:9 - cgroup2 cgroup rw,nsdelegate
```

and after a few:

```
talosctl -n 172.20.0.2 read /proc/1/mountinfo | grep cgroup                            Пн 23 авг 2021 17:52:19
27 175 0:24 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:9 - cgroup2 cgroup rw
175 21 0:24 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:9 - cgroup2 cgroup rw
```

sergelogvinov:
Yes, it can be if another pod/proccess remount /sys/fs/cgroup without flag nsdelegate...

sergelogvinov:
And some privileged containers flush this flag too... :(

smira (approved):

PR #4136, opened 2021-08-24
fix: don't allow bootstrap if etcd data directory is not empty

smira:
If etcd data directory is not empty, it means that `etcd` is already
running (or etcd has run), so removing data directory might lead to data
loss. Under normal circumstances `etcd` is not started before either
node joins existing etcd cluster or it gets bootstrapped.

Most importantly, with this change it's not possible to nuke etcd member
by issuing `talosctl bootstrap` on already running Talos node.

Plus nag user about doing bootstrap if the node is waiting to join.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -351,7 +351,10 @@ func addMember(ctx context.Context, r runtime.Runtime, addrs []string, name stri
 }
 
 func buildInitialCluster(ctx context.Context, r runtime.Runtime, name, ip string) (initial string, learnerMemberID uint64, err error) {
-	var id uint64
+	var (
+		id      uint64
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -363,6 +366,12 @@ func buildInitialCluster(ctx context.Context, r runtime.Runtime, name, ip string
 			resp      *clientv3.MemberListResponse
 		)
 
+		if time.Since(lastNag) > 30*time.Second {
+			log.Printf("etcd is waiting to join the cluster, if this node is the first node in the cluster, please run `talosctl bootstrap`")
```
andrewrynhard, line 370: 👍🏻

andrewrynhard (approved):

PR #4168, opened 2021-08-27
feat: add node address filters, filter out k8s addresses for Talos API

smira:
This implements abstract `NodeAddressFilter` which can be attached to
build additional `NodeAddress` resources filtering out some entries from
the complete list.

Kubernetes creates two filters to get all node IPs without Kubernetes
CIDRs and vice versa, only Kubernetes CIDRs.

API certificate generation now considers all addresses minus K8s CIDRs.

### DIFF
--- a/internal/app/machined/pkg/controllers/config/k8s_address_filter.go
+++ b/internal/app/machined/pkg/controllers/config/k8s_address_filter.go
@@ -1,0 +1,129 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package config
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/AlekSi/pointer"
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+	"go.uber.org/zap"
+	"inet.af/netaddr"
+
+	"github.com/talos-systems/talos/pkg/resources/config"
+	"github.com/talos-systems/talos/pkg/resources/k8s"
+	"github.com/talos-systems/talos/pkg/resources/network"
+)
+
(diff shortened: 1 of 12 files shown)


Ulexus:
This feels like a system which is going to grow rapidly with permutations over time.  Should these really be pre-filtered resources?  I would thing tagging each address with metadata to allow downstream users to make their own filters seems more flexible.

Also, what is the intention of Accumulative here?

smira:
Accumulative is mostly for certificate generation. In fact, it's your idea, Seán. It it used mostly for cert generation. 

I feel we need to decompose lists to make it easier to consume and watch them, but in the end we could have flipped those to be tags on addresses. But for now this structure allows for suppressing change propagation in a very efficient way. If one subscribed to accumulative-no-k8s, they will get notified only if that list chbsges

Ulexus:
Okay, that trade-off makes sense.  I wonder if it wouldn't be better, then to be more descriptive in the naming.  Accumulative only makes sense given a context.  If cert generation or node identity or self addresses are what they represent, can we name them that way?

I would say, that that case, that they are NodeAddresses, and what you have as NodeAddresses are actually more ReceivingAddresses or maybe AcceptedAddresses... ProximateAddresses.

So we have ProximateAddresses composed of `node` addresses (all addresses which are directed to the node itself... i.e., not Pod addresses), `direct` addresses (addresses which are directly present on the node itself... i.e., not including NAT addresses), `kubernetes` addresses (addresses of Pods or other things inside Kubernetes)?

Even that doesn't really cover everything.  It still feels pretty infirm.

smira:
We can change names easily, that is just filter ID, it's not hardcoded. We can get back to that once this thing comes together with other pieces and we see the full picture. So far accumulative minus k8s pods provides exactly what we need for certs, it even enabled my other PR to work correctly

smira:
But I agree it's hard to come up with something universal, e.g. some installations might want to exclude some addresses from the selection, like limiting etcd advertise addresses. At the same time I feel this solution for now should allow us to move forward until we see a better approach

Ulexus (approved):

PR #4184, opened 2021-09-01
fix: correctly define example for `extraMounts`

Unix4ever:
The type was changed, but the example wasn't updated accordingly.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -6,7 +6,7 @@ github_repo = "talos-systems/talos"
 match_deps = "^github.com/(talos-systems/[a-zA-Z0-9-]+)$"
 
 # previous release
-previous = "v0.11.0"
+previous = "v0.12.0"
 
 pre_release = true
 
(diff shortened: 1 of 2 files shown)


smira (approved):
how can we best avoid bugs like that in the future?

Unix4ever:
Or I actually do have an idea. Let me try something.

smira:
Looks like we're over release-notes limit, it might make sense to bump previous verion on `hack/release.toml` to 0.12.0 now

Unix4ever:
So I've tried to generate static checks like:

```go
var _ []*ExtraMounts = extraMountsExample
```
It works for most things, but it gets tricky when our code relies on some third party libs like with `PEMEncodedKey`.

smira:
> So I've tried to generate static checks like:
> 
> ```go
> var _ []*ExtraMounts = extraMountsExample
> ```
> 
> It works for most things, but it gets tricky when our code relies on some third party libs like with `PEMEncodedKey`.
> We _may_ be able to dynamically construct imports but that's where it may become flaky by itself.

we could in theory run `goimports` (or whatever the equivalent is), I don't want to hold this PR though, let's have a fix

Unix4ever:
> we could in theory run goimports (or whatever the equivalent is), I don't want to hold this PR though, let's have a fix

yep, that's a thing for the later investigation. I will create a task

smira (approved):

Unix4ever:
https://github.com/talos-systems/talos/issues/4185

PR #4197, opened 2021-09-06
feat: drop some capabilities to be never available

smira:
This PR makes sure that some capabilities (SYS_BOOT and SYS_MODULES) and
never be gained by any process running on Talos except for `machined`
itself.


Unix4ever (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -237,6 +238,49 @@ func SetRLimit(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFun
...
+		secbits |=
+			cap.SecbitNoCapAmbientRaise | cap.SecbitNoCapAmbientRaiseLocked
+
+		if err := secbits.Set(); err != nil {
+			return fmt.Errorf("error setting secbits: %w", err)
+		}
+
+		// Set PR_SET_NO_NEW_PRIVS to limit setuid and similar privilege raising techniques.
+		// See https://www.kernel.org/doc/html/v5.10/userspace-api/no_new_privs.html.
+		const PR_SET_NO_NEW_PRIVS = 38 //nolint:golint,revive,stylecheck
```
AlekSi, line 255: That's [`unix.PR_SET_NO_NEW_PRIVS`](https://pkg.go.dev/golang.org/x/sys/unix).

smira: thanks, not sure how I missed that

internal/app/machined/internal/install/install.go
```diff
@@ -113,7 +114,14 @@ func RunInstallerContainer(disk, platform, ref string, configBytes []byte, reg c
 		oci.WithHostHostsFile,
 		oci.WithHostResolvconf,
 		oci.WithParentCgroupDevices,
-		oci.WithPrivileged,
+		oci.WithCapabilities(capability.AllGrantableCapabilities()),
```
smira, line 117: I had to expand `Privileged` into explicit list, as in our case listing "all" capabilities for the reason that machined has different permitted and bounding capabilities.

andrewrynhard (dismissed):

andrewrynhard (approved):

PR #4200, opened 2021-09-06
feat: reboot via kexec

smira:
This should save a lot of time on BIOS/POST time with bare metal
hardware.


smira:
requires https://github.com/talos-systems/pkgs/pull/334

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1664,3 +1670,77 @@ func ActivateLogicalVolumes(seq runtime.Sequence, data interface{}) (runtime.Tas
...
+		defer kernel.Close() //nolint:errcheck
+
+		initrd, err := os.Open(initrdPath)
+		if err != nil {
+			return err
+		}
+
+		defer initrd.Close() //nolint:errcheck
+
+		parameters, err := ioutil.ReadFile("/proc/cmdline")
```
smira, line 1718: this is probably wrong way, as it reads current boot args vs. what is in GRUB

smira:
Q: do we need a kernel boot arg to disable kexec? should kexec be disabled or enabled by default?

andrewrynhard (approved):
Can't wait to try this!

PR #4204, opened 2021-09-07
feat: implement KubeSpan peer generation controller

smira:
Controller watches cluster Affiliates and generates KubeSpanPeerSpecs
from those which are not local node Affiliates and have KubeSpan
configuration attached.

Example:

```
$ talosctl -n 172.20.0.2 get kubespanpeerspecs
NODE         NAMESPACE   TYPE               ID                                             VERSION   LABEL                    ENDPOINTS
172.20.0.2   kubespan    KubeSpanPeerSpec   27E8I+ekrqT21cq2iW6+fDe+H7WBw6q9J7vqLCeswiM=   1         talos-default-worker-1   ["172.20.0.3:51820"]
```

```
$ talosctl -n 172.20.0.3 get kubespanpeerspecs -o yaml
node: 172.20.0.3
metadata:
    namespace: kubespan
    type: KubeSpanPeerSpecs.kubespan.talos.dev
    id: mB6WlFOR66Jx5rtPMIpxJ3s4XHyer9NCzqWPP7idGRo=
    version: 1
    owner: kubespan.PeerSpecController
    phase: running
    created: 2021-09-07T19:26:35Z
    updated: 2021-09-07T19:26:35Z
spec:
    address: fdc8:8aee:4e2d:1202:f073:9cff:fe6c:4d67
    additionalAddresses:
        - 10.244.1.0/32
    endpoints:
        - 172.20.0.2:51820
    label: talos-default-master-1
```

KubeSpan peers will be used to drive configuration of the KubeSpan
networking components.

### DIFF
--- a/pkg/resources/kubespan/peer_spec.go
+++ b/pkg/resources/kubespan/peer_spec.go
@@ -1,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package kubespan
+
+import (
+	"fmt"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"inet.af/netaddr"
+)
+
+// PeerSpecType is type of PeerSpec resource.
+const PeerSpecType = resource.Type("KubeSpanPeerSpecs.kubespan.talos.dev")
+
+// PeerSpec is produced from cluster.Affiliate which has KubeSpan information attached.
+//
+// PeerSpec is identified by the public key.
+type PeerSpec struct {
+	md   resource.Metadata
+	spec PeerSpecSpec
+}
+
+// PeerSpecSpec describes PeerSpec state.
+type PeerSpecSpec struct {
+	Address             netaddr.IP         `yaml:"address"`
+	AdditionalAddresses []netaddr.IPPrefix `yaml:"additionalAddresses"`
+	Endpoints           []netaddr.IPPort   `yaml:"endpoints"`
+	Label               string             `yaml:"label"`
+}
+
+// NewPeerSpec initializes a PeerSpec resource.
+func NewPeerSpec(namespace resource.Namespace, id resource.ID) *PeerSpec {
+	r := &PeerSpec{
+		md:   resource.NewMetadata(namespace, PeerSpecType, id, resource.VersionUndefined),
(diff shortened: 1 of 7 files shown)


pkg/resources/kubespan/peer_spec.go
```diff
@@ -0,0 +1,95 @@
...
+type PeerSpec struct {
+	md   resource.Metadata
+	spec PeerSpecSpec
+}
+
+// PeerSpecSpec describes PeerSpec state.
+type PeerSpecSpec struct {
+	Address             netaddr.IP         `yaml:"address"`
+	AdditionalAddresses []netaddr.IPPrefix `yaml:"additionalAddresses"`
+	Endpoints           []netaddr.IPPort   `yaml:"endpoints"`
```
Ulexus, line 30: I'm a bit torn here about the IPPort as oppose of a "host" port (`net.UDPAddress`).  There is the option/possibility of DNS names being used, which case they should be resolved at _use_ time, rather than at spec time.  In fact, the current controller (and discovery service) allows DNS names, and these are further populated from the `v1.Node` addresses, which include DNS names.

I have, at various times, used each, and I just don't know whether it is worth the extra complexity to retain DNS resolution or not.

Ulexus (approved):

pkg/resources/kubespan/peer_spec.go
smira: good point, let's keep it in the back in our brains as we move through the implementation. I feel DNS is tricky a bit, as it might resolve to multiple IPs and we should try them all, so ultimately Wireguard input should be IP addresses? but we certainly might need more than a single address here.

I also think once we publish `ExternalIPs` as `Addresses` internally, so that node is aware of the external IPs, they will get pushed to the discovery as well. So my point - we can have it all, it's more about the addresses the node has (even if the address is not assigned to any interface at the moment)

Ulexus: It's more than "should", it's "must".  The input to wireguard itself must be an IP address, not a DNS name.  And yes, DNS is absolutely a tricky bit.  For instance, there is also a SRV record system for Wireguard.  Should we support that, too?  Should be try append domains on top of the hostname?  All of them?  Intermediaries?

So basically, I basically agree with this approach for now.  DNS is just too sticky to dive into right now.  I think we can probably extend this resource later, if we need to, by having names as a separate field or struct or whatever.

smira: What I'm trying to say is that Wireguard endpoint is ultimately `IP:port` pair, so the input should be exactly that, while the code building the list of the endpoints might use DNS to resolve DNS names into IP list and populate the endpoints properly.

PR #4207, opened 2021-09-08
fix: suppress spurious Kubernetes API server cert updates

smira:
With the last changes, `kube-apiserver` certificates are generated based
on the assigned `NodeAdresses`, machine configuration, etc. Whenver the
certificate is regenerated, `kube-apiserver` is reloaded to pick up the
new cert.

With Virtual IP enabled, Virtual IP address is included into the
certificate from the beginning as it is specified in the machine
configuration, but as virtual IP moves between the nodes this causes
`NodeAddresses` update, which triggers the controller, generates new
certs and reloads `kube-apiserver` at bad time (right after VIP got
moved). Even though the cert generated is identical to the previous one,
the API server reload makes it unavailable for 30-90 seconds.

This change extracts `CertSANs` as a separate resource so that its
updates are suppressed if the CertSANs sources change, but the final
list stays the same, and in turn prevents final certificate from being
updated.


smira:
passed full integration run: https://ci.dev.talos-systems.io/talos-systems/talos/20049/6/5

internal/app/machined/pkg/controllers/secrets/api.go
```diff
@@ -121,6 +121,11 @@ func (ctrl *APIController) Run(ctx context.Context, r controller.Runtime, logger
 		if err = ctrl.teardownAll(ctx, r); err != nil {
 			return err
 		}
+
+		// reset inputs back to what they were initiall
```
andrewrynhard, line 125: super duper nit: `s/initiall/initially/` 😉

andrewrynhard (dismissed):

andrewrynhard (approved):

PR #4227, opened 2021-09-13
chore: build using Go 1.17

smira:
`initramfs` size for amd64 shrinks by 1.3 MiB.


Makefile
```diff
@@ -1,7 +1,7 @@
 REGISTRY ?= ghcr.io
 USERNAME ?= talos-systems
 SHA ?= $(shell git describe --match=none --always --abbrev=8 --dirty)
-TAG ?= $(shell git describe --tag --always --dirty --match v[0-9]*)
```
smira, line 4: this is fun: with bash `v[0-9]*` is passed verbatim if there are no files matching the expression (this is what we want, as we want to get pattern to git), but if by chance you have any matching files, it gets expanded by bash breaking things, so `\` is better, but no-op usually

Unix4ever (approved):

PR #4269, opened 2021-09-20
fix: don't marshal clock with SecretsBundle

smira:
This field is not marshalable, as it's technically an interface.

This will be used to save/load SecretsBundle as a whole in the CABPT.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4269)


smira:
This PR is set to be backported to 0.12 to pull in new machinery package in CABPT without going to 0.13 machinery yet.

andrewrynhard (approved):

PR #4276, opened 2021-09-22
fix: make probe open blockdevice in readonly mode

Unix4ever:
Update `go-blockdevice` library.

Readwrite mode doesn't work when there are readonly devices like `iso`.


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/talos-systems/talos
 go 1.17
 
 replace (
+	github.com/talos-systems/go-blockdevice => github.com/Unix4ever/go-blockdevice v0.1.1-0.20210922122752-e367f9dc7fa9
```
smira, line 6: really? :)

Unix4ever: - [x] commit
- [ ] force push :man_facepalming:

smira (approved):

PR #4278, opened 2021-09-22
feat: enable resource API in the maintenance mode

smira:
This basically provides `talosctl get --insecure` in maintenance mode.
Only non-sensitive resources are available (equivalent to having
`os:reader` role in the Talos client certificate).

Changes:

* refactored insecure/maintenance client setup in talosctl
* `LinkStatus` is no longer sensitive as it shows only Wireguard public
key, `LinkSpec` still contains private key for obvious reasons
* maintenance mode injects `os:reader` role implicitly

The motivation behind this PR is to deprecate networkd-era interfaces &
routes APIs which are being used in TUI installer, and we need a
replacement.


Ulexus (commented):
> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4278)* status: all files reviewed, 1 unresolved discussion (waiting on @smira)

___
*[pkg/resources/network/link.go, line 515 at r1](https://reviewable.io/reviews/talos-systems/talos/4278#-MkDBVju3KrQlyoIwLNy:-MkDBVju3KrQlyoIwLNz:b-v9sxwi) ([raw file](https://github.com/talos-systems/talos/blob/2b72d1d2ea21605a6211307035b782b68e3fbb37/pkg/resources/network/link.go#L515)):*
> ```Go
> func (spec *WireguardSpec) Decode(dev *wgtypes.Device, isStatus bool) {
> 	if isStatus {
> 		spec.PublicKey = dev.PublicKey.String()
> ```

Is there some reason to exclude the PublicKey when not pulling status?

smira (commented):
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4278)* status: all files reviewed, 1 unresolved discussion (waiting on @Ulexus)

___
*[pkg/resources/network/link.go, line 515 at r1](https://reviewable.io/reviews/talos-systems/talos/4278#-MkDBVju3KrQlyoIwLNy:-MkDCYvp-SObVqvjUh9b:bt0arls) ([raw file](https://github.com/talos-systems/talos/blob/2b72d1d2ea21605a6211307035b782b68e3fbb37/pkg/resources/network/link.go#L515)):*
<details><summary><i>Previously, Ulexus (Seán C. McCord) wrote…</i></summary><blockquote>

Is there some reason to exclude the PublicKey when not pulling status?
</blockquote></details>

It doesn't give anything (as it's derived value anyways), and it makes spec comparison mentally harder - anyone generating the spec should fill _both_ Public and PrivateKey which doesn't make much sense

Ulexus (approved):
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4278)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @smira)

___
*[pkg/resources/network/link.go, line 515 at r1](https://reviewable.io/reviews/talos-systems/talos/4278#-MkDBVju3KrQlyoIwLNy:-MkDMNPZ62Cojn1klMKC:b42lkwa) ([raw file](https://github.com/talos-systems/talos/blob/2b72d1d2ea21605a6211307035b782b68e3fbb37/pkg/resources/network/link.go#L515)):*
<details><summary><i>Previously, smira (Andrey Smirnov) wrote…</i></summary><blockquote>

It doesn't give anything (as it's derived value anyways), and it makes spec comparison mentally harder - anyone generating the spec should fill _both_ Public and PrivateKey which doesn't make much sense
</blockquote></details>

Yeah, considering the generation side of the Spec, I guess that follows.  It's just a little strange from the consumption side.  But fair point:  public key is derived from the private key, so it really is just redundant.

Ulexus (approved):

PR #4292, opened 2021-09-27
feat: override static pods default args by extra Args

Unix4ever:
Use `argsbuilder` same way as it's used in services.
Rewrite `kubeProxy` generation code to override default args.

As a consequence of this change now flags do not have determined order
as they all come from a single merged map.

Fixes: https://github.com/talos-systems/talos/issues/4238

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4292)

### DIFF
--- a/internal/app/machined/pkg/controllers/config/k8s_control_plane.go
+++ b/internal/app/machined/pkg/controllers/config/k8s_control_plane.go
@@ -7,6 +7,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/AlekSi/pointer"
 	"github.com/cosi-project/runtime/pkg/controller"
@@ -15,6 +16,7 @@ import (
 	talosnet "github.com/talos-systems/net"
 	"go.uber.org/zap"
 
+	"github.com/talos-systems/talos/pkg/argsbuilder"
 	"github.com/talos-systems/talos/pkg/images"
 	talosconfig "github.com/talos-systems/talos/pkg/machinery/config"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
@@ -213,16 +215,20 @@ func (ctrl *K8sControlPlaneController) manageManifestsConfig(ctx context.Context
 	return r.Modify(ctx, config.NewK8sManifests(), func(r resource.Resource) error {
 		images := images.List(cfgProvider)
 
+		proxyArgs, err := getProxyArgs(cfgProvider)
+		if err != nil {
+			return err
+		}
+
 		r.(*config.K8sControlPlane).SetManifests(config.K8sManifestsSpec{
 			Server:        cfgProvider.Cluster().Endpoint().String(),
 			ClusterDomain: cfgProvider.Cluster().Network().DNSDomain(),
 
 			PodCIDRs: cfgProvider.Cluster().Network().PodCIDRs(),
 
-			ProxyEnabled:   cfgProvider.Cluster().Proxy().Enabled(),
-			ProxyImage:     cfgProvider.Cluster().Proxy().Image(),
-			ProxyMode:      cfgProvider.Cluster().Proxy().Mode(),
-			ProxyExtraArgs: cfgProvider.Cluster().Proxy().ExtraArgs(),
+			ProxyEnabled: cfgProvider.Cluster().Proxy().Enabled(),
+			ProxyImage:   cfgProvider.Cluster().Proxy().Image(),
+			ProxyArgs:    proxyArgs,
 
 			CoreDNSEnabled: cfgProvider.Cluster().CoreDNS().Enabled(),
 			CoreDNSImage:   cfgProvider.Cluster().CoreDNS().Image(),
@@ -286,4 +292,26 @@ func (ctrl *K8sControlPlaneController) manageExtraManifestsConfig(ctx context.Co
 
 func (ctrl *K8sControlPlaneController) teardownAll(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
 	return nil
+}
+
+func getProxyArgs(cfgProvider talosconfig.Provider) ([]string, error) {
+	clusterCidr := strings.Join(cfgProvider.Cluster().Network().PodCIDRs(), ",")
+
+	builder := argsbuilder.Args{
+		"cluster-cidr":           clusterCidr,
+		"hostname-override":      "$(NODE_NAME)",
+		"kubeconfig":             "/etc/kubernetes/kubeconfig",
+		"proxy-mode":             cfgProvider.Cluster().Proxy().Mode(),
+		"conntrack-max-per-core": "0",
+	}
+
+	policies := argsbuilder.MergePolicies{
+		"kubeconfig": argsbuilder.MergeDenied,
+	}
+
+	if err := builder.Merge(cfgProvider.Cluster().Proxy().ExtraArgs(), argsbuilder.WithMergePolicies(policies)); err != nil {
+		return nil, err
+	}
+
+	return builder.Args(), nil
 }
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strconv"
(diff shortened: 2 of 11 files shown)


internal/app/machined/pkg/controllers/k8s/manifest_test.go
```diff
@@ -171,9 +178,7 @@ func (suite *ManifestSuite) TestReconcileKubeProxyExtraArgs() {
 	rootSecrets := secrets.NewRoot(secrets.RootKubernetesID)
 	manifestConfig := config.NewK8sManifests()
 	spec := defaultManifestSpec
-	spec.ProxyExtraArgs = map[string]string{
-		"bind-address": "\"::\"",
-	}
+	spec.ProxyArgs = append(spec.ProxyArgs, "--bind-address=\"::\"")
```
Unix4ever, line 181: this test becomes kinda weird after I changed the logic. Not sure what's the right way to fix it.

smira: yep, this was testing mostly JSON escaping, an I think it's wrong now probably

internal/app/machined/pkg/controllers/k8s/templates.go
```diff
@@ -138,13 +138,8 @@ spec:
...
         - /usr/local/bin/kube-proxy
-        - --cluster-cidr={{ join .PodCIDRs "," }}
-        - --hostname-override=$(NODE_NAME)
-        - --kubeconfig=/etc/kubernetes/kubeconfig
-        - --proxy-mode={{ .ProxyMode }}
-        - --conntrack-max-per-core=0
-        {{- range $k, $v := .ProxyExtraArgs }}
-        - {{ printf "--%s=%s" $k $v | json }}
+        {{- range $arg := .ProxyArgs }}
+        - {{ $arg }}
```
smira, line 142: I think `| json` part should stay, as it protects generated YAML

Unix4ever: ah, gotcha :+1:

internal/app/machined/pkg/controllers/k8s/manifest_test.go
Unix4ever: yeah... Let me check

Unix4ever: so in terms of escaping I think it works fine:

```yaml
    proxy:
        image: k8s.gcr.io/kube-proxy:v1.21.3 # The container image used in the kube-proxy manifest.
        extraArgs:
          bind-address: "\"::\""
```

Turned into:

```
        - /usr/local/bin/kube-proxy
        - --cluster-cidr=10.244.0.0/16
        - --proxy-mode=iptables
        - --bind-address="::"
        - --hostname-override=$(NODE_NAME)
        - --kubeconfig=/etc/kubernetes/kubeconfig
        - --conntrack-max-per-core=0
```

smira: having `| json` shouldn't hurt in the template, as YAML is a subset of JSON, so we can put array values as JSON formatted strings, and this avoids any kind of YAML weirdness that might appear there

smira: not sure if the test was completely correct though, but there was something up with `:` iirc, so that it treats part of the list as a map

pkg/argsbuilder/argsbuilder_args.go
```diff
@@ -15,13 +18,59 @@ type Value = string
...
+			return NewDenylistError(key)
+		case MergeAdditive:
+			values := strings.Split(a[key], ",")
+			definedValues := map[string]struct{}{}
+
+			for i, v := range values {
+				definedValues[strings.TrimSpace(v)] = struct{}{}
+
+				if v == "" {
+					values = append(values[:i], values[i+1:]...)
```
smira, line 56: I feel this is not valid - removing items from the slice during the iteration.

Proper way to filter while iterating is https://github.com/golang/go/wiki/SliceTricks#filter-in-place

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -328,33 +359,51 @@ func (ctrl *ControlPlaneStaticPodController) manageControllerManager(ctx context
...
+		"profiling":                        "false",
+	}
+
 	if cfg.CloudProvider != "" {
-		args = append(args, fmt.Sprintf("--cloud-provider=%s", cfg.CloudProvider))
+		builder.Set("cloud-provider", cfg.CloudProvider)
+	}
+
+	mergePolicies := argsbuilder.MergePolicies{
+		"cluster-cidr":             argsbuilder.MergeAdditive,
```
smira, line 389: I feel that `cluster-cidr` and `service-cluster-ip-range` should be overridable completely; if someone needs to override them, let them do it; additive policy doesn't allow to remove our default values

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -203,53 +205,82 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
...
 	}
 
-	for k, v := range cfg.ExtraArgs {
-		args = append(args, fmt.Sprintf("--%s=%s", k, v))
+	mergePolicies := argsbuilder.MergePolicies{
+		"enable-admission-plugins": argsbuilder.MergeAdditive,
+		"authorization-mode":       argsbuilder.MergeAdditive,
+		"tls-cipher-suites":        argsbuilder.MergeAdditive,
+		"etcd-servers":             argsbuilder.MergeAdditive,
+		"service-cluster-ip-range": argsbuilder.MergeAdditive,
```
smira, line 259: I think this should be overridable.

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -203,53 +205,82 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
...
+		builder.Set("cloud-provider", cfg.CloudProvider)
 	}
 
-	for k, v := range cfg.ExtraArgs {
-		args = append(args, fmt.Sprintf("--%s=%s", k, v))
+	mergePolicies := argsbuilder.MergePolicies{
+		"enable-admission-plugins": argsbuilder.MergeAdditive,
+		"authorization-mode":       argsbuilder.MergeAdditive,
+		"tls-cipher-suites":        argsbuilder.MergeAdditive,
+		"etcd-servers":             argsbuilder.MergeAdditive,
```
smira, line 258: probably `MergeDenied`, as if we lock down the etcd certs, the only etcd server is the one Talos builds?

internal/app/machined/pkg/controllers/config/k8s_control_plane.go
```diff
@@ -287,3 +293,27 @@ func (ctrl *K8sControlPlaneController) manageExtraManifestsConfig(ctx context.Co
...
+	builder := argsbuilder.Args{
+		"cluster-cidr":           clusterCidr,
+		"hostname-override":      "$(NODE_NAME)",
+		"kubeconfig":             "/etc/kubernetes/kubeconfig",
+		"proxy-mode":             cfgProvider.Cluster().Proxy().Mode(),
+		"conntrack-max-per-core": "0",
+	}
+
+	policies := argsbuilder.MergePolicies{
+		"cluster-cidr": argsbuilder.MergeAdditive,
```
smira, line 309: same as in other places, probably keep it overridable

internal/app/machined/pkg/controllers/k8s/manifest_test.go
```diff
@@ -171,9 +178,7 @@ func (suite *ManifestSuite) TestReconcileKubeProxyExtraArgs() {
...
 	manifestConfig.SetManifests(spec)
```
smira, line 182: any high-level tests here to make sure args building process goes fine with overrides?

smira (commented):
lgtm, just couple small notes

internal/app/machined/pkg/controllers/k8s/manifest_test.go
Unix4ever: I've added some

smira (approved):

PR #4294, opened 2021-09-27
fix: check trustd API CA on worker nodes

smira:
This distributes API CA (just the certificate, not the key) to the
worker nodes on config generation, and if the CA cert is present on the
worker node, it verifies TLS connection to the trustd with the CA
certificate.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4294)


sergelogvinov:
After this PR, Talos machined can starts at boot time (as control-plane) with self-sign cert.
And it add possibilities to debug node, or call some unsecured functions.

smira:
> After this PR, Talos machined can starts at boot time (as control-plane) with self-sign cert.
> And it add possibilities to debug node, or call some unsecured functions.

not sure I follow that. it doesn't change the way control plane works

sergelogvinov:
I meat, it can be a new PR.
It will solve many problems....

Unix4ever (approved):

PR #4301, opened 2021-09-29
feat: add an option to skip downed peers in KubeSpan

### LINKED ISSUE #4248: configurable policy for kubespan enforcement for down/unknown peers
## Feature Request

The question is: should we force traffic to go via KubeSpan even if the peer is not up? 

If we do, it means traffic is always encrypted (as soon as discovery catches up though), if we don't traffic might silently become not encrypted and go directly to the peer if the wireguard connection is down.

This should be configurable with default to force encryption.

### Description

Implementation-wise: should we add allowedIPs of the downed/unknown peers to the nftables.

smira:
Fixes #4248

This resolves the balance between security and connectivity.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4301)

### DIFF
--- a/pkg/machinery/config/provider.go
+++ b/pkg/machinery/config/provider.go
@@ -217,6 +217,7 @@ type Route interface {
 // KubeSpan configures KubeSpan feature.
 type KubeSpan interface {
 	Enabled() bool
+	ForceRouting() bool
 }
 
 // Time defines the requirements for a config that pertains to time related
--- a/website/content/docs/v0.13/Reference/configuration.md
+++ b/website/content/docs/v0.13/Reference/configuration.md
@@ -5262,6 +5262,21 @@ Cluster discovery should be enabled with .cluster.discovery.enabled for KubeSpan
 </div>
 
 <hr />
+<div class="dd">
+
+<code>allowDownPeerBypass</code>  <i>bool</i>
+
+</div>
+<div class="dt">
+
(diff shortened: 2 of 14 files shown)


website/content/docs/v0.13/Reference/configuration.md
```diff
@@ -5262,6 +5262,21 @@ Cluster discovery should be enabled with .cluster.discovery.enabled for KubeSpan
 </div>
 
 <hr />
+<div class="dd">
+
+<code>skipDownPeers</code>  <i>bool</i>
```
smira, line 5267: @Ulexus any idea for a better parameter name?

rsmitty (approved):

website/content/docs/v0.13/Reference/configuration.md
smira: from @Ulexus: `AllowDownPeerBypass`

pkg/machinery/config/provider.go
```diff
@@ -217,6 +217,7 @@ type Route interface {
 // KubeSpan configures KubeSpan feature.
 type KubeSpan interface {
 	Enabled() bool
+	ForceRouting() bool
```
Ulexus, line 220: Should this naming now mirror the config flag?

Ulexus (commented):
Align config flag, CLI flag, and config machinery accessor function name?

PR #4309, opened 2021-09-30
fix: allow 404 when querying external IP in AWS

rsmitty:
This PR fixes an issue where we weren't allowing for a 404 to be
returned when querying
http://169.254.169.254/latest/meta-data/public-ipv4. This path won't
exist if there's no public IP on the instance.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4309)


smira (approved):

Unix4ever (approved):

smira:
we got implementing same thing, fixed in #4310

PR #4311, opened 2021-09-30
docs: fix documentation for cluster discovery

AlekSi:
Use the real value in an example.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4311)

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
@@ -1999,7 +1999,7 @@ type RegistryServiceConfig struct {
 	RegistryDisabled bool `yaml:"disabled,omitempty"`
 	// description: |
 	//   External service endpoint.
-	//   examples:
-	//     - value: 'constants.DefaultDiscoveryServiceEndpoint'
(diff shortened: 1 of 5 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1978,6 +1978,6 @@ type RegistryServiceConfig struct {
 	// description: |
 	//   External service endpoint.
 	//   examples:
-	//     - value: 'constants.DefaultDiscoveryServiceEndpoint'
```
smira, line 1981: there _should_ be a way to use the constant here to avoid duplication... probably @Unix4ever knows?

smira: I believe dropping the quotes should allow this to be a value directly?

AlekSi: I tried that back then; it wasn't enough.

The real problem was with the indentation before `examples`.

smira (dismissed):

PR #4312, opened 2021-09-30
chore: allow insecure discovery in debug builds

AlekSi:
If Talos is built with `sidero.debug` build tag (`make WITH_DEBUG=1`),
the machine configuration is allowed to use insecure HTTP for the discovery service.


smira (dismissed):
:cool:

smira (approved):

PR #4313, opened 2021-09-30
feat: add an option to configure kubelet node IP based on subnets

### LINKED ISSUE #4243: Nodes choose bad InternalIPs, leading to kube-apiserver communication problems
## Bug Report

When I bootstrap a cluster on Hetzner, nodes pick one of the machine's IP addresses as their InternalIP non-deterministically, resulting in a situation like this:

```
NAME       STATUS   ROLES                  AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE                  KERNEL-VERSION   CONTAINER-RUNTIME
master-1   Ready    control-plane,master   2m38s   v1.22.1   10.0.1.2        <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
master-2   Ready    control-plane,master   2m18s   v1.22.1   65.21.243.122   <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
master-3   Ready    control-plane,master   2m32s   v1.22.1   10.0.1.4        <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
worker-1   Ready    <none>                 2m39s   v1.22.1   65.108.54.255   <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
worker-2   Ready    <none>                 2m18s   v1.22.1   10.0.2.3        <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
worker-3   Ready    <none>                 2m11s   v1.22.1   10.0.2.4        <none>        Talos (v0.13.0-alpha.0)   5.10.58-talos    containerd://1.5.5
```

hcloud-cloud-controller-manager will eventually update the InternalIP and ExternalIP fields correctly, but by the time that happens, a number of pods have been started with podIPs corresponding to the incorrect InternalIPs:

```
NAMESPACE     NAME                                              READY   STATUS                       RESTARTS        AGE     IP              NODE       NOMINATED NODE   READINESS GATES
kube-system   kube-apiserver-master-1                           1/1     Running                      0               4m46s   10.0.1.2        master-1   <none>           <none>
kube-system   kube-apiserver-master-2                           1/1     Running                      0               3m45s   65.21.243.122   master-2   <none>           <none>
kube-system   kube-apiserver-master-3                           1/1     Running                      0               5m14s   10.0.1.4        master-3   <none>           <none>
```

Even after hcloud-ccm updates the IP fields, this state is retained:

```
$ kubectl get endpoints kubernetes
NAME         ENDPOINTS                                        AGE
kubernetes   10.0.1.2:6443,10.0.1.4:6443,65.21.243.122:6443   41m
```

I only allow communication in the 10.0.0.0/16 network. Thus, some in-cluster connections to the service will fail:

```
$ kubectl run -it --rm --restart=Never curl --image=curlimages/curl -- sh
If you don't see a command prompt, try pressing enter.
/ $ for n in $(seq 1 10); do curl --max-time 1 -sS -k https://10.96.0.1:443/healthz && printf '\n'; done
curl: (28) Connection timed out after 1000 milliseconds
ok
ok
ok
curl: (28) Connection timed out after 1001 milliseconds
curl: (28) Connection timed out after 1001 milliseconds
ok
curl: (28) Connection timed out after 1001 milliseconds
ok
ok
```

### Possible solution 1

Forcing kubelet to use an address from 10.0.0.0/16 as the node-ip should fix the problem. The `cluster.etcd.subnet` setting could be moved up by a level and an IP address from the subnet could be passed to both kubelet as the node-ip and etcd as the advertised address.

### Possible solution 2

I load hcloud-ccm through `cluster.externalCloudProvider` but it also requires a secret which I add through `cluster.inlineManifests`. The latter is only processed after kube-apiserver has been started, ensuring that hcloud-ccm is not ready in time. I haven't verified whether having the secret available during the externalCloudProvider phase would result in hcloud-ccm assigning the correct InternalIP before the other pods are started with no race conditions though.

It could be worth considering to allow both URLs and inline manifests in all places which accept manifests:

```yaml
cluster:
  externalCloudProvider:
    enabled: true
    manifests:
      - url: https://...
      - https://...  # Parse a plain string as equivalent to { url: "..." } for backwards compatibility
      - inline: "..."
  network:
    cni:
      name: custom
      manifests:
        - url: https://...
        - inline: "..."
  extraManifests:
    - url: https://...
    - inline: "..."  # Making cluster.inlineManifests redundant
```

### Environment

- Talos version:
  ```
  Client:
          Tag:         v0.13.0-alpha.0
          SHA:         7c5045bd
          Built:       
          Go version:  go1.16.7
          OS/Arch:     linux/amd64
  Server:
          NODE:        10.0.1.3
          Tag:         v0.13.0-alpha.0
          SHA:         7c5045bd
          Built:       
          Go version:  go1.16.7
          OS/Arch:     linux/amd64
          Enabled:     RBAC
  ```
- Kubernetes version:
  ```
  Client Version: v1.22.1
  Server Version: v1.22.1
  ```
- Platform: Hetzner Cloud

smira:
Fixes #4243

The idea is to make sure kubelet picks node IP based on filtering by
CIDRs of the node's addresses. The flow is simple - every address is
filtered by subnet and picked if it matches the subnet.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4313)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -843,6 +850,21 @@ type KubeletConfig struct {
...
+	KubeletNodeIP KubeletNodeIPConfig `yaml:"nodeIP,omitempty"`
+}
+
+// KubeletNodeIPConfig represents the kubelet node IP configuration.
+type KubeletNodeIPConfig struct {
+	//  description: |
+	//    The `subnets` field configures the networks to pick kubelet node IP from.
+	//    For dual stack configuration, there should be two subnets: one for IPv4, another for IPv6.
+	//    If not specified, kubelet configures node IP automatically.
+	KubeletNodeIPSubnets []string `yaml:"subnets,omitempty"`
```
Ulexus, line 867: Maybe `validSubnets` instead of just `subnets`?

Ulexus (commented):
Maybe use the term ValidSubnets instead of just Subnets, to try and imply the usage?

pkg/machinery/config/provider.go
```diff
@@ -236,6 +236,12 @@ type Kubelet interface {
 	ExtraArgs() map[string]string
 	ExtraMounts() []specs.Mount
 	RegisterWithFQDN() bool
+	NodeIP() KubeletNodeIP
+}
+
+// KubeletNodeIP defines the way node IPs are selected for the kubelet.
+type KubeletNodeIP interface {
+	Subnets() []string
```
Ulexus, line 244: Align with ValidSubnets?

Ulexus (commented):
Missed another Subnets ref for alignment (internal function call)

pkg/machinery/config/provider.go
smira: Oh yeah, missed that

rsmitty (approved):

PR #4324, opened 2021-10-04
test: bump CAPI components to v1alpha4

Unix4ever:
Alpha versions of talos providers for now.


smira (commented):
please also unlock K8S_VERSION to be latest 1.22

Issue #4318, opened 2021-10-02
kubelet.nodeIP.validSubnets uses the interfaces.vip address


ion1:
## Bug Report

When using the following in my config,

```yaml
machine:
  kubelet:
    nodeIP:
      validSubnets:
        - 10.0.0.0/16
 network:
    interfaces:
      - interface: eth1
        dhcp: true
        vip:
          ip: 10.0.1.100
          hcloud:
            apiToken: ...
```

kubelet picks up both the DHCP-provided IP address and the VIP as node-ip and fails with the following error message:

> server.go:294] "Failed to run kubelet" err="failed to run Kubelet: bad --node-ip \"10.0.1.2,10.0.1.100\"; must contain either a single IP or a dual-stack pair of IPs"

### Environment

- Talos version: 0.13.0-beta.0
- Kubernetes version: 1.22.2
- Platform: Hetzner Cloud

smira:
So the CIDR should exclude the VIP address.

ion1:
Fair enough. I have my master nodes in 10.0.1.0/24 and worker nodes in 10.0.2.0/24 and I figured I would also have the master VIP in 10.0.1.0/24. I suppose I could either move it out of the 10.0.0.0/22 subnet or do this:

```yaml
      validSubnets:
        # The VIP is in 10.0.1.128/25
        - 10.0.1.0/25
        - 10.0.2.0/24
```

PR #4327, opened 2021-10-05
fix: node-ip single ip or a dualstack pair

sergelogvinov:
This fix pass only single IP or a dual-stack pair to the node-ip kubelet param.

#4318

## Acceptance


internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -411,5 +412,32 @@ func pickNodeIPs(cidrs []string) ([]stdnet.IP, error) {
 		}
 	}
 
+	if len(result) > 1 {
```
smira, line 415: I do not think this is the right way, as we can't pick one IP if multiple IPs match (we have no idea which one is the "right" one). I would prefer kubelet to fail like it does today

PR #4328, opened 2021-10-05
WIP

AlekSi:


hack/release.toml
```diff
@@ -36,11 +36,30 @@ On bare-metal hardware BIOS POST process might take 10-15 minutes, so Talos rebo
...
+For example:
+
+```yaml
+install:
+  extraKernelArgs:
+    - kexec_load_disabled=1
+```
+
+Please note that capabilities are dropped before machine configuration is loaded,
+so disabling kexec via `machine.sysctls` (like in the section above) will not work.
```
AlekSi, line 62: Is it accurate? Feels weird to me.

I wonder if we can drop capabilities after parsing machine configuration.

smira:
```suggestion
so disabling kexec via `machine.sysctls` (like in the section `Reboots via kexec`) will not be enough.
```

(sections are not ordered same way as in the toml)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -241,6 +241,13 @@ func SetRLimit(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFun
 // DropCapabilities drops some capabilities so that they can't be restored by child processes.
 func DropCapabilities(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
+		prop, err := kernel.ReadParam(&kernel.Param{Key: "kernel.kexec_load_disabled"})
```
Unix4ever, line 244: I think `DropCapabilities` is called in `Initialize` sequence before we load the config.
How would it be possible to set this kernel param before this method? Or I'm missing some other way to set up kernel params? :thinking:

AlekSi: I tried to explain that in the changelog.

Yes, that is weird :)

Unix4ever: ah, install arguments. Yes, now that makes sense.

smira: Kernel params are driven by:

* for images (including ISO) - whatever is in the image
* for PXE boot - PXE kernel args
* for installed Talos booting from disk via `machine.install.extraKernelArgs`

Technically we don't have to do anything for install phase most probably, but rather for the phase when Talos boots normally.

Unix4ever (dismissed):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Unix4ever: If we get docs for `kexec` we may want to include that release notes block there.

Unix4ever (dismissed):

smira (approved):

PR #4330, opened 2021-10-05
chore: build using only amd64 builders

smira:
Our CI amd64 builders are 3-4 times faster to build Talos than our arm64
builders when building Talos.

Our Dockerfile was restructured a while ago to support cross-compilation
on all platforms but CI was still using amd64/arm64 workers, so arm64
part was done on arm64 builders.

As our CI runs on Talos, `binfmt_misc` is not enabled in the kernel, but
buildkit has built-in QEMU emulation layer which works just fine for
those small pieces which actually need to run arm64 binaries on amd64
(mostly `apk add` in the installer container). Interesting enough,
buildkit QEMU support fails for `ca-certificates` script which runs
after install. At the same time I believe we don't need
`ca-certificates` in the installer, as installer doesn't download
anything from the network, and `ca-certificates` were added a while ago
when installer was actually downloading configuration on its own.


.drone.jsonnet
```diff
@@ -136,6 +136,9 @@ local setup_ci = {
     'make ./_out/sonobuoy',
     'make ./_out/kubectl',
   ],
+  environment: {
+    "BUILDKIT_FLAVOR": "cross",
```
andrewrynhard, line 140: I couldn't find docs on this. Have a ref?

andrewrynhard (approved):

.drone.jsonnet
smira: it's a feature of our builld-container: https://github.com/talos-systems/build-container/pull/46

Ulexus (approved):

PR #4331, opened 2021-10-06
fix: don't enable 'no new privs' on the system level

smira:
This breaks some pods which specifically drop everything but gain
capabilities back via file capabilities (e.g. `nginx-ingress`).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4331)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -257,16 +257,6 @@ func DropCapabilities(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 			return fmt.Errorf("error setting secbits: %w", err)
 		}
 
-		// Set PR_SET_NO_NEW_PRIVS to limit setuid and similar privilege raising techniques.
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -257,16 +257,6 @@ func DropCapabilities(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 			return fmt.Errorf("error setting secbits: %w", err)
 		}
 
-		// Set PR_SET_NO_NEW_PRIVS to limit setuid and similar privilege raising techniques.
-		// See https://www.kernel.org/doc/html/v5.10/userspace-api/no_new_privs.html.
```
andrewrynhard, line 261: What do we lose by dropping this? Are we still protected from kexec?

smira: yeah, completely, this is not related to kexec. I tried to be smart, but I wasn't.

what this does it disables `allowPrivilegeEscalation` in Kubernetes terms. we would rather leave it to the user, as there are some valid usecases for it (`nginx-ingress` being one example).

andrewrynhard (approved):

PR #4335, opened 2021-10-08
fix: use ECDSA-SHA256 signature algorithm for Kubernetes certs

smira:
Previously Talos used ECDSA-SHA512 with P-256 EC key, which is not
widely supported combination. Use ECDSA-SHA256 instead.

There's no security benefit to use ECDSA-SHA512 with P-256 key, and this
combination is officially supported by TLS 1.3 standard.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4335)


smira:
Clarification: if the cluster with older version of Talos is upgraded to 0.13, Kubernetes CA will be still using ECDSA-SHA512, but e.g. kube-apiserver will be using ECDSA-SHA256

PR #4370, opened 2021-10-13
feat: feed control plane endpoints on workers from cluster discovery

### LINKED ISSUE #4231: feed k8s.Endpoints from the cluster discovery members
## Feature Request

We can use cluster discovery (if enabled) to feed the endpoints of the control plane nodes (on worker nodes, used to join the cluster). 

### Description

This would allow joins without Kubernetes being up (if using the discovery service).

smira:
Fixes #4231

This allows much faster worker join on `apid` level without waiting for
Kubernetes control plane to be up (and even if the Kubernetes control
plane doesn't go up or the kubelet isn't up).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4370)

### DIFF
--- a/internal/app/machined/pkg/controllers/secrets/api.go
+++ b/internal/app/machined/pkg/controllers/secrets/api.go
@@ -165,7 +165,6 @@ func (ctrl *APIController) reconcile(ctx context.Context, r controller.Runtime, 
 		inputs = append(inputs, controller.Input{
 			Namespace: k8s.ControlPlaneNamespaceName,
 			Type:      k8s.EndpointType,
-			ID:        pointer.ToString(k8s.ControlPlaneEndpointsID),
 			Kind:      controller.InputWeak,
 		})
 	}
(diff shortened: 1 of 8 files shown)


smira:
Draft, as I need to add a test for the new endpoint controller.

internal/app/machined/pkg/controllers/secrets/api.go
```diff
@@ -242,26 +241,23 @@ func (ctrl *APIController) reconcile(ctx context.Context, r controller.Runtime,
...
-				if state.IsNotFoundError(err) {
-					continue
-				}
-
-				return fmt.Errorf("error getting endpoints resource: %w", err)
+				return fmt.Errorf("error getting endpoints resources: %w", err)
 			}
 
-			endpoints := endpointResource.(*k8s.Endpoint).TypedSpec()
+			var endpointAddrs k8s.EndpointList
```
moonrockfamily, line 249: This change seems to necessitate public ip's for the Talos controlplane endpoints (apid + trustd services).   Without this change using a domain name for the canonical endpoint would permit resolution to a public IP of a load balancer that would then connect via private IP's to the endpoint services, including trustd on port 50001.  mydomain:50001 -> public ip port 50001 -> control plane registered private ip port 50001

PR #4383, opened 2021-10-15
fix: change services OOM score

Unix4ever:
- Bump `cri` `oom_score_adj` to: `-500`.
- Explicitly set kubelet oom score in `kubelet.yaml` config to -450.
- Additionally adjust containers' OOM score using API.
    
Fixes: https://github.com/talos-systems/talos/issues/4378
    
---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4383)


smira:
what was it before the change? 

if we set `cri` to -100, should we set to value like `-50` ? as kubelet should be killed before CRI is killed

Unix4ever:
Before the change it was just unset. Though there may be some default value there, right?
`cri` also has explicit score `-999`.

PR #4398, opened 2021-10-19
feat: switch containerd logging to JSON

AlekSi:
That will make parsing much easier.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4398)


AlekSi:
Merged into #4402 to rebase one PR instead of two

PR #4413, opened 2021-10-21
feat: update Kubernetes to 1.23.0-alpha.3

smira:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#v1230-alpha3

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4413)


smira:
> and maybe we should also have 1.22 -> 1.23 upgrade flow in the `upgrade-k8s` utils.

thanks, I totally forgot that!

PR #4469, opened 2021-10-29
refactor: get rid of prometheus/procfs dependency in pkg/resources

smira:
This PR introduces the concept of the resource adapter: something which
wraps core resource (which defines presentation) and adds additional
fatures to it which we don't want to have in `pkg/machinery` in the
future, but at the same time there's value in keeping it next to the
resource itself.

If the concept looks good, I will continue doing it same way for other
resources.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4469)


Ulexus (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4469)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @smira)

PR #4485, opened 2021-11-03
chore: log KubeSpan IPs overlaps

AlekSi:
Should be helpful for debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4485)

### DIFF
--- a/internal/app/machined/pkg/controllers/kubespan/peer_spec.go
+++ b/internal/app/machined/pkg/controllers/kubespan/peer_spec.go
@@ -129,7 +129,8 @@ func (ctrl *PeerSpecController) Run(ctx context.Context, r controller.Runtime, l
 
 					for otherPublicKey, otherIPSet := range peerIPSets {
 						if otherIPSet.Overlaps(ipSet) {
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/controllers/kubespan/peer_spec.go
```diff
@@ -127,9 +127,22 @@ func (ctrl *PeerSpecController) Run(ctx context.Context, r controller.Runtime, l
 						continue
 					}
 
+					ipSetPrefixes := ipSet.Prefixes()
```
smira, line 130: should we extract a function which stringifies `IPSet` and use it with logging?

as I don't think we need to build this list if there're no overlaps (and we can build it only when logging)

AlekSi: Sounds good.

Do we want to log that by default? Or only in debug builds?

Unix4ever (approved):

internal/app/machined/pkg/controllers/kubespan/peer_spec.go
smira: I think by default is good!

smira (approved):

PR #4513, opened 2021-11-10
feat: add GCP ccm

frezbo:
Update docs to reflect deploying GCP Cloud Control Manager (CCM)

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4513)

### DIFF
--- a/website/content/docs/v0.14/Cloud Platforms/gcp.md
+++ b/website/content/docs/v0.14/Cloud Platforms/gcp.md
@@ -255,6 +255,8 @@ We need to download two deployment manifests for the deployment from the Talos g
 ```bash
 curl -fsSLO "https://raw.githubusercontent.com/talos-systems/talos/master/website/content/docs/v0.14/Cloud%20Platforms/gcp/config.yaml"
 curl -fsSLO "https://raw.githubusercontent.com/talos-systems/talos/master/website/content/docs/v0.14/Cloud%20Platforms/gcp/talos-ha.yaml"
+# if using ccm
+curl -fsSLO "https://raw.githubusercontent.com/talos-systems/talos/master/website/content/docs/v0.14/Cloud%20Platforms/gcp/gcp-ccm.yaml"
 ```
 
 ### Updating the config
@@ -273,6 +275,7 @@ resources:
     properties:
       zone: us-west2-c
       talosVersion: v0.13.2
+      externalCloudProvider: false
       controlPlaneNodeCount: 5
       controlPlaneNodeType: n1-standard-1
       workerNodeCount: 3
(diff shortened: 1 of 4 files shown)


smira:
> /m -ff

you need first some reviews on the PR

smira:
@rsmitty could be the best resource to review this one, I'm not familiar enough with this

nberlee:
@frezbo 
You may also want to add the RBAC for the ILB to be able to write to events: (you will see an error if you create a loadbalancer with an internal ip) see https://raw.githubusercontent.com/on2itsecurity/secure-k8s/main/cluster/manifests/gce-provider-extra-rbac.yaml

Consider adding `--port: 0` to the parameters (see warning in log)
And add a default `gce.conf` as this is where the GCP module mostly gets its config from.. .The paramaters in the cloud controller are limited...

frezbo:
@nberlee Thanks for the feedback.

- I'll test out the internal load balancer and update if necessary. 
- `--port=0` is not set as the `apiserver-authentication-reader` role binding is created. It;s a common role binding with ccm's
- Do you have an example snippet of `gce.conf`, and what parameters are usually in there?. That would help in improving the docs

nberlee:
All possible settings are here: https://github.com/kubernetes/cloud-provider-gcp/blob/master/providers/gce/gce.go#L183
my gce.conf looks something like this, although, it really depends on your GCP environment if you want to have some of these features to be autodetected. The last two are the most important I think because that makes it behave way differently...

```
data:
  gce.conf: |-
    [global]
    project-id = my-prod-id
    network-project-id = my-prod-id
    network-name = vpc-workers-cluster2
    subnetwork-name = sn-vips-cluster2
    node-instance-prefix = vm-worker-node
    node-tags = talos-workers-cluster2
    multizone = true
    regional = true
```

I must have been wrong about `--port=0` then I thought this was the nonsecure CCM api port... just like kube-apiserver/scheduler/controler-manager has this option

frezbo:
Linked issues:

- https://github.com/kubernetes/cloud-provider-gcp/issues/286
- https://github.com/kubernetes/cloud-provider-gcp/issues/291
- https://github.com/kubernetes/cloud-provider-gcp/issues/289

sergelogvinov:
cloud-provider-gcp does not have many features than gce-ccm.

If any one wants to run ccm without allocate-node-cidrs -- here fix https://github.com/sergelogvinov/cloud-provider-gcp/commit/49216c929660283abf884c882081de7da733fea0 (good for multi cloud setups)

in my case, I use only that params, (csi/autoscale work fine with that ccm)

```yaml
          command:
            - /cloud-controller-manager
          args:
            - --bind-address=127.0.0.1
            - --cloud-provider=gce
            - --cloud-config=/etc/gce/gce.conf
            - --allocate-node-cidrs=false
            - --cluster-cidr=10.64.0.0/12
            - --controllers=cloud-node,cloud-node-lifecycle
            - --port=0
            - --use-service-account-credentials
            - -v=2
```

```yaml
  gce.conf: |
    [global]
    project-id = 
    network-name =
```

smira (approved):
not an expert, but this looks much more fun to me

Issue #4488, opened 2021-11-04
documentation typo: "Enquinix" for "Equinix"


vielmetti:
## Bug Report

Typos in "Equinix" (as "Enquinix") 

### Description

Search for https://github.com/talos-systems/talos/search?q=enquinix and see several
places in the docs where the brand name is misspelled.

rsmitty:
lol, nothing like copy/pasting that same typo all over the place :smile: 

Thanks for the heads up, we'll get this fixed.

vielmetti:
At least it's consistent and easy to fix!

PR #4514, opened 2021-11-10
fix: common typos

frezbo:
Fizes #4488 and some other typos

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4514)


smira (commented):
nit: it might make sense to keep the commit type to `docs:`, as it touches only the docs

smira (approved):

PR #4521, opened 2021-11-10
fix: improve NTP sync process

### LINKED ISSUE #4425: Detect and act on time skew/drift
## Feature Request

It seems as though currentley the time implementation periodcally polls NTP servers and updates as nessecary. Instead perhaps we should dectect if there's drift and act against it. This is especically helpful for bare-metal systems that do not have a RTC, but also can be helpful for systems with it. 

Curretnley rook-ceph and prometheus both are reporting a time-skew and eventually... Talos will correct the time, but not fast enough to signal alerts/alarms.

smira:
Fixes #4425

* add more logging for responses and sync process
* adjust time sync constants
* change the way poll interval is chosen (increasing on good sync,
decreasing on variation)
* filter out spikes

Based on flow in https://github.com/systemd/systemd/blob/main/src/timesync/timesyncd-manager.c

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4521)

### DIFF
--- a/internal/pkg/ntp/ntp.go
+++ b/internal/pkg/ntp/ntp.go
@@ -9,8 +9,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"math"
 	"math/bits"
-	"math/rand"
 	"net"
 	"reflect"
 	"sync"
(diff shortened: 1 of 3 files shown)


smira:
image

internal/pkg/ntp/ntp.go
```diff
@@ -126,12 +140,75 @@ func (syncer *Syncer) restartSync() {
 	}
 }
 
+func absDuration(d time.Duration) time.Duration {
```
AlekSi, line 143: Another place where generics would be useful 😆

AlekSi (approved):

internal/pkg/ntp/ntp.go
smira: yep, exactly :)

internal/pkg/ntp/ntp_test.go
```diff
@@ -212,6 +241,57 @@ func (suite *NTPSuite) TestSyncContinuous() {
...
+				return retry.ExpectedError(fmt.Errorf("not enough syncs"))
+			}
+
+			for _, adj := range suite.clockAdjustments {
+				// 1s spike should be filtered out
+				suite.Assert().Equal(time.Millisecond, adj)
+			}
+
+			return nil
+		}))
```
AlekSi, line 288: nitpick: I think the last `)` should be on the next line

smira: fair enough!

AlekSi (approved):

PR #4547, opened 2021-11-16
fix: containerd failed to load plugin

sergelogvinov:
Solve warning message at boot time: failed to load plugin io.containerd.internal.v1.opt

This is part of PR #3975

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4547)

### DIFF
--- a/hack/containerd.toml
+++ b/hack/containerd.toml
@@ -1,12 +1,10 @@
 version = 2
 
-disabled_plugins = ["io.containerd.snapshotter.v1.aufs", "io.containerd.v1.zfs", "io.containerd.snapshotter.v1.zfs", "io.containerd.v1.devmapper", "io.containerd.snapshotter.v1.devmapper", "io.containerd.snapshotter.v1.btrfs"]
-
-imports = ["/var/cri/conf.d/*.toml"]
+disabled_plugins = [
+    "io.containerd.grpc.v1.cri",
+    "io.containerd.internal.v1.opt",
+]
 
 [debug]
 level = "info"
 format = "json"
-
-[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
-    runtime_type = "io.containerd.runc.v2"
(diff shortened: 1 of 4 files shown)


smira:
This PR seems to expose one issue which wasn't frequent, but basically etcd fails to fetch the image on bootstrap. I want to look into it a bit more.

smira:
@sergelogvinov I did a bit of refactoring add added config for both system `containerd` running from the tmpfs to make sure it doesn't load the CRI plugin for example.

hack/containerd.toml
```diff
@@ -1,12 +1,7 @@
 version = 2
 
-disabled_plugins = ["io.containerd.snapshotter.v1.aufs", "io.containerd.v1.zfs", "io.containerd.snapshotter.v1.zfs", "io.containerd.v1.devmapper", "io.containerd.snapshotter.v1.devmapper", "io.containerd.snapshotter.v1.btrfs"]
```
smira, line 3: in fact I moved `hack/containerd.toml` to `hack/cri-containerd.toml`, and created new `hack/containerd.toml`

sergelogvinov:
i've totally forgot about https://github.com/talos-systems/pkgs/blob/master/containerd/pkg.yaml#L24

We build containerd without those plugins ```disabled_plugins```. 
That's why i've removed it here #3975

Sorry )

smira:
> i've totally forgot about https://github.com/talos-systems/pkgs/blob/master/containerd/pkg.yaml#L24
> 
> We build containerd without those plugins `disabled_plugins`. That's why i've removed it here #3975
> 
> Sorry )

no problem, let me clean up disabled plugins, but anyways disabling cri for core containerd makes sense, same as `/opt`)

smira:
updated removing disabled plugins from disabled_plugins ;)

hack/cri-containerd.toml
```diff
@@ -0,0 +1,12 @@
+version = 2
+
+disabled_plugins = []
```
smira, line 3: all mentioned plugins are disabled in the containerd build, no need to disable them once again

smira (approved):

PR #4548, opened 2021-11-17
feat: add new event types

Unix4ever:
Add config load + validation errors and address + hostnames events.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4548)


internal/app/machined/pkg/controllers/network/address_event.go
```diff
@@ -0,0 +1,113 @@
...
+	ticker := time.NewTicker(time.Minute * 10)
+
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+		case <-r.EventCh():
```
smira, line 70: feels like `select` should end here with `}` to make it act both on ticker and reconcile events

api/machine/machine.proto
```diff
@@ -171,12 +171,33 @@ message RestartEvent {
...
+  string error = 2;
+};
+
 message EventsRequest {
   int32 tail_events = 1;
   string tail_id = 2;
   int32 tail_seconds = 3;
 }
 
+// AddressEvent reports node endpoints aggregated from k8s.Endpoints and network.Hostname.
```
smira, line 195: nit: should be moved before `EventsRequest` to group all events together

website/content/docs/v0.14/Reference/cli.md
```diff
@@ -93,7 +93,7 @@ talosctl cluster create [flags]
       --bad-rtc                                 launch VM with bad RTC state (QEMU only)
       --cidr string                             CIDR of the cluster network (IPv4, ULA network for IPv6 is derived in automated way) (default "10.5.0.0/24")
       --cni-bin-path strings                    search path for CNI binaries (VM only) (default [/home/user/.talos/cni/bin])
-      --cni-bundle-url string                   URL to download CNI bundle from (VM only) (default "https://github.com/talos-systems/talos/releases/download/v0.14.0-alpha.1/talosctl-cni-bundle-${ARCH}.tar.gz")
+      --cni-bundle-url string                   URL to download CNI bundle from (VM only) (default "https://github.com/talos-systems/talos/releases/download/v0.14.0-alpha.0/talosctl-cni-bundle-${ARCH}.tar.gz")
```
smira, line 96: git fetch --tags ?

api/machine/machine.proto
```diff
@@ -171,12 +171,33 @@ message RestartEvent {
   int64 cmd = 1;
 };
 
+// MachineConfigErrorEvent means some machine config error.
+message MachineConfigErrorEvent {
```
smira, line 175: I think we should add new events so that they can be handled by `talosctl events` - it has some code to unpack the events for display

Unix4ever: two new kinds, right?

smira: three new kinds now ;)

smira (approved):

PR #4553, opened 2021-11-17
feat: can disable controller-manager and scheduler

sergelogvinov:
This PR solves problems:
1. You want to run controlplane from etcd-backup, and make
   changes/investigation and be sure that controlplane
   does no change resources.
2. Run controlplane in less memory only for quorum.

result example:

```yaml
machine:
    # Provides machine specific contolplane configuration options.
    controlPlane:
        # Scheduler machine specific configuration options.
        scheduler:
            disabled: true # Disable Scheduler on the node.
```

```shell
talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.202 get staticpods
NODE             NAMESPACE      TYPE        ID                        VERSION
192.168.10.202   controlplane   StaticPod   kube-apiserver            1
192.168.10.202   controlplane   StaticPod   kube-controller-manager   1
```
## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4553)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1261,6 +1261,9 @@ type APIServerConfig struct {
 
 // ControllerManagerConfig represents the kube controller manager configuration options.
 type ControllerManagerConfig struct {
+	//   description: |
+	//     Disable clusterControllerManager on the node.
+	Disabled bool `yaml:"disabled,omitempty"`
```
smira, line 1266: I feel like probably this should part of the machine configuration, not cluster one, as we certainly don't want disable things on cluster level

sergelogvinov: Proposal config:

```yaml
machine:
  kubelet:
    staticPod:
      kube-controller-manager:
        disable: true
      kube-scheduler: {}
```

Ulexus:
I don't really like packing in control plane stuff inside the `kubelet` field.  or one, this leads to interesting implications for worker nodes.  For another, while static pods are technically run by the kubelet, they aren't really viewed in that way by the user.

This is kind of catching us in a state of transition. but you present a perfectly valid use-case.  I propose instead to have a `machine.controlplane.<component>.disable` section for these things.  I'm not delighted with that, but it at least sections it off to be clearly related to the controlplane while _also_ being restricted to the local machine config.

Ulexus (changes requested):

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1297,6 +1332,9 @@ type ProxyConfig struct {
 
 // SchedulerConfig represents the kube scheduler configuration options.
 type SchedulerConfig struct {
+	//   description: |
+	//     Disable clusterScheduler on the node.
```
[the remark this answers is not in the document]
sergelogvinov: Thank you, did not clean after previous edition.

smira:
Update the PR to allow `--immediate` changes for new config values.

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -140,9 +142,33 @@ func (ctrl *ControlPlaneStaticPodController) Run(ctx context.Context, r controll
 				return fmt.Errorf("error getting control plane config: %w", err)
 			}
 
-			if err = pod.f(ctx, r, logger, res.(*config.K8sControlPlane), secretsVersion); err != nil {
+			var podID string
+
+			if podID, err = pod.f(ctx, r, logger, res.(*config.K8sControlPlane), secretsVersion); err != nil {
```
Ulexus, line 147: This feels a bit too implicit to me:  if the podID is not returned, then it should be deleted?  Could we instead get a more explicit signal?  What happens if one of these `manageXXX()` functions gets expanded and has the ability to return errors which do _not_ indicate that the pod should be destroyed?  That would leads to self-destruction due to a transitent error.

I wonder if, perhaps, we have sufficient data here to be able to just preload these controlplane component structs with the administrative status of them from their configs.

smira: if we can't build the pod spec, I think we should destroy the running one, as otherwise the intention (configuration) gets out of sync with the running state.

Ulexus: Fair point; I'm still just concerned that this leads to the potential of externalities being responsible for taking down the controlplane in ways which are not easily diagnosable.  I'm not going to stand on this objection, though.
Perhaps we can moderate it by having sufficient visibility into any error encountered (log when a controlplane pod is being removed and the reason for it)

Ulexus: Eh, we already do, for the most part.  If we return empty _and_ a nil error, that's an error on us (upstream).

Ulexus (approved):

PR #4559, opened 2021-11-18
feat: add ULA prefix for SideroLink

smira:
Allocate one more value for the SideroLink feature.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4559)

### DIFF
--- a/pkg/machinery/resources/network/ula.go
+++ b/pkg/machinery/resources/network/ula.go
@@ -22,6 +22,9 @@ const (
 
 	// ULAKubeSpan is the Unique Local Addressing space key for the Talos KubeSpan feature.
 	ULAKubeSpan = 0x02
+
+	// ULASideroLink is the Unique Local Addressing space key for the SideroLink feature.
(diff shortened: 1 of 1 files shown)


Ulexus:
Do we have any Sidero installation-unique ID we can use for the cluster ID?  It's important for ULA to actually _be_ unique.

smira:
> Do we have any Sidero installation-unique ID we can use for the cluster ID? It's important for ULA to actually _be_ unique.

my only concern with that is that someone reinstalls Sidero, they might lose it.. I wish we had a way to do good persistence, but we can get back to this. We need anyway to persist Sidero Wireguard key.

So the scenario I'm thinking about: I have a bunch of servers, and I want to upgrade Sidero. I want to make sure SideroLink is not interrupted by the upgrade, so that nodes can reconnect back.

smira:
I added some preparation to provide site-specific SideroLink ULA addresses.

Unix4ever (approved):

PR #4564, opened 2021-11-19
fix: unblock events watch on context cancel

### LINKED ISSUE #4563: Talos not autorebooting on failed bootup sequence.
## Bug Report
On a failed boot sequence Talos is not autorebooting itself but stays stuck on shut down of services. Talos should always autoreboot after failure.

### Description
The box was installing itself hence no access to Talos API at this time as it never reached working state.

### Logs
MicrosoftTeams-image (3)

### Environment
kexec is disabled

 0.13.0
 1.21.1
- Platform:
baremetal

smira:
Fixes #4563 

This fixes an issue when `etcd` get stuck in `Pre` state so that even
`bootstrap` request doesn't kill it. This looks like a cluster which
fails to bootstrap itself and get stuck on reboot.

Even though `Watch()` aborts on cancel, channel recieve will block
forever as no events are delivered.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4564)

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -264,7 +264,11 @@ func generatePKI(ctx context.Context, r runtime.Runtime) (err error) {
 	var event state.Event
 
 	for {
-		event = <-watchCh
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case event = <-watchCh:
+		}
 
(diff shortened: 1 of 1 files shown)


smira:
example of the bug: https://ci.dev.talos-systems.io/talos-systems/talos/21712/6/5

Unix4ever (approved):

PR #4580, opened 2021-11-23
feat: talosctl fish completion support

nberlee:
Generate talosctl completion for Fish

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4580)


smira (approved):

smira:
@nberlee I already fixed it, but just to make things easier next time: `talosctl` docs are auto-generated, so you can do simply `make docs` to get them updated

nberlee:
Ah, thx! I will remember that for next time. Thats way easier

smira:
thank you!

PR #4585, opened 2021-11-23
feat: implement events sink controller

Unix4ever:
Report Talos events to any gRPC server.
Destination address is specified by using kernel parameters.

Fixes: https://github.com/talos-systems/talos/issues/4458

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4585)

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/events_sink.go
+++ b/internal/app/machined/pkg/controllers/runtime/events_sink.go
@@ -1,0 +1,192 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package runtime
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/AlekSi/pointer"
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/rs/xid"
+	"github.com/talos-systems/go-procfs/procfs"
+	"github.com/talos-systems/siderolink/api/events"
+	"go.uber.org/atomic"
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
+	"github.com/talos-systems/talos/pkg/machinery/constants"
+	"github.com/talos-systems/talos/pkg/machinery/resources/network"
+)
+
+// EventsSinkController watches events and forwards them to the events sink server
+// if it's configured.
+type EventsSinkController struct {
+	V1Alpha1Events runtime.Watcher
+	Cmdline        *procfs.Cmdline
+	Drainer        *runtime.Drainer
+
+	drainSub *runtime.DrainSubscription
+	drain    bool
+	backlog  atomic.Int32
+	eventID  xid.ID
+}
+
+// Name implements controller.Controller interface.
+func (ctrl *EventsSinkController) Name() string {
+	return "v1alpha1.EventsSinkController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *EventsSinkController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: network.NamespaceName,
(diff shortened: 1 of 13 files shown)


internal/app/machined/pkg/runtime/drainer.go
```diff
@@ -0,0 +1,94 @@
...
+func (d *Drainer) Drain(ctx context.Context) error {
+	var eg errgroup.Group
+
+	d.subscriptionsMu.Lock()
+	defer d.subscriptionsMu.Unlock()
+
+	for _, s := range d.subscriptions {
+		s := s
+
+		eg.Go(func() error {
```
smira, line 35: suggestion: we can make s.events length 1 and simply send to each subscriber event without goroutine

internal/app/machined/pkg/runtime/drainer.go
```diff
@@ -0,0 +1,94 @@
...
+			case <-s.shutdown:
+			case <-ctx.Done():
+				return context.Canceled
+			}
+
+			return nil
+		})
+	}
+
+	return eg.Wait()
```
smira, line 56: wait for subscriptions to be empty, use channel to wake up and re-check length

internal/app/machined/pkg/runtime/drainer.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// EventCh returns drain events channel.
+func (s *DrainSubscription) EventCh() <-chan DrainEvent {
+	return s.events
+}
+
+// Cancel the subscription which triggers drain to shutdown.
+func (s *DrainSubscription) Cancel() {
+	select {
```
smira, line 87: remove subscription from the list

internal/app/machined/pkg/controllers/runtime/events_sink.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		break
+	}
+
+	errCh := make(chan error)
+
+	sink := ctrl.Cmdline.Get(constants.KernelParamEventsSink).First()
+
+	conn, err := grpc.DialContext(ctx, *sink, grpc.WithInsecure())
```
smira, line 113: nit: we should `close()` the connection

internal/app/machined/pkg/controllers/runtime/events_sink.go
```diff
@@ -0,0 +1,190 @@
...
+	client := events.NewEventSinkServiceClient(conn)
+
+	opts := []runtime.WatchOptionFunc{}
+	if ctrl.eventID.IsNil() {
+		opts = append(opts, runtime.WithTailEvents(-1))
+	} else {
+		opts = append(opts, runtime.WithTailID(ctrl.eventID))
+	}
+
+	if err = ctrl.V1Alpha1Events.Watch(func(eventCh <-chan runtime.EventInfo) {
```
smira, line 127: nit: the func body could be extracted into a separate function for readability

internal/app/machined/pkg/runtime/drainer.go
```diff
@@ -0,0 +1,99 @@
...
+		case s.events <- DrainEvent{}:
+		default:
+		}
+	}
+	d.subscriptionsMu.Unlock()
+
+	for {
+		select {
+		case <-d.shutdown:
+			d.subscriptionsMu.Lock()
```
smira, line 40: I guess we need to check before we receive, as the subscriptions might be empty as we enter the function. so basically move the check above `select` statement

internal/app/machined/pkg/runtime/drainer.go
```diff
@@ -0,0 +1,99 @@
...
+			l := len(d.subscriptions)
+			d.subscriptionsMu.Unlock()
+
+			if l == 0 {
+				return nil
+			}
+		case <-ctx.Done():
+			return context.Canceled
```
smira, line 48: this and some places around: `return ctx.Err()` sounds better, as this will return a specific error with the context, e.g. it can be DeadlineExceeded

internal/app/machined/pkg/controllers/runtime/events_sink.go
Unix4ever: not sure about that... I don't really want to pass through all local variables into that function.
But can declare it as a func in the function body.

smira: I don't think there are _that_ many parameters? some of the are on the `ctrl` anyways, you probably need to pass just the client?

smira (approved):

PR #4598, opened 2021-11-26
fix: initialize Drainer properly

smira:
Because of the bug Drainer never worked properly, as `shutdown` channel
wasn't initialied.

Also add unit-tests and add some small clean-ups which don't affect
functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4598)

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -209,7 +209,7 @@ func run() error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	drainer := &runtime.Drainer{}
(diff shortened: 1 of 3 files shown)


internal/app/machined/main.go
```diff
@@ -209,7 +209,7 @@ func run() error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
-	drainer := &runtime.Drainer{}
+	drainer := runtime.NewDrainer()
```
Unix4ever, line 212: 🤦‍♂️

Unix4ever (approved):

PR #4600, opened 2021-11-26
feat: kernel log (kmsg) delivery controller

### LINKED ISSUE #4455: kernel log message delivery
## Feature Request

- kernel command line argument: `talos.logging.kernel=tcp://<host>:<port>`
- format: JSON lines
- delaying reboot until the log is flushed (up to some timeout) - need to check the TCP buffer (needs some research)

smira:
This controller is enabled when `talos.logging.kernel=` kernel arg is
passed. It will try to send logs to the endpoint as JSON-over-TCP (or
UDP-over-TCP, depends on the endpoint) as structured messages.

Example (from test implementation):

```
2021-11-26T19:53:21.912+0300	INFO	siderolink-agent/log_receiver.go:23	kernel log message	{"src_address": "fdae:41e4:649b:9303:680a:dfab:f7fa:ea00", "msg": {"clock":6252819,"facility":"user","msg":"[talos] task startAllServices (1/1): waiting for 6 services\n","priority":"warning","seq":711,"talos-level":"warn","talos-time":"2021-11-26T16:53:21.3258698Z"}}
```

Fixes #4455

See also https://github.com/talos-systems/siderolink/pull/4


internal/app/machined/pkg/controllers/runtime/kmsg_log.go
```diff
@@ -0,0 +1,182 @@
...
+
+		if !netStatus.(*network.Status).TypedSpec().AddressReady {
+			// wait for address
+			continue
+		}
+
+		break
+	}
+
+	if ctrl.drainSub == nil {
```
Unix4ever, line 96: I wonder if it makes sense to move drain sub closer to the loop to avoid blocking it when we return before entering the loop.

Unix4ever (approved):

internal/app/machined/pkg/controllers/runtime/kmsg_log.go
Unix4ever: or maybe it's fine. In the end it tries to reconnect there. So I guess it's fine.

smira: it subscribes for drains as soon as the address is ready, so it should be fine (I guess).

PR #4612, opened 2021-11-29
chore: add API breaking changes detector

### LINKED ISSUE #4576: Add test for API deprecation/removal
Talos is not 1.0 yet, so technically, we can break our APIs in incompatible ways anytime. In practice, we are not doing that because that would make everyone's life harder. Still, we want to clean up our APIs before the 1.0 release: use common field names and conventions, rework APIs that are hard to use correctly, etc.

The existing `prototool break` check is a part of the solution, but not all of it.

I propose to modify an existing [apid_test](https://github.com/talos-systems/talos/blob/f7d1e7776917475507aa99847f88b9c22c9f7b95/internal/app/apid/pkg/backend/apid_test.go#L200) (and maybe find a better place for it) to check `deprecated` options and comments in .proto files and fail if a deprecated entity is not removed on time.

For example:
```protobuf
message ListRequest {
  // Use recursion_depth instead.
  // This field will be removed in v0.16.
  bool recurse = 2 [deprecated = true];
  int32 recursion_depth = 3;
```

The test will fail if the current version is 0.16.

AlekSi:
Closes #4576.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4612)


Makefile
```diff
@@ -262,6 +262,9 @@ cloud-images: ## Uploads cloud images (AMIs, etc.) to the cloud registry.
 
 # Code Quality
 
+api-descriptors: ## Generates API descriptors used to detect breaking API changes.
```
AlekSi, line 265: That target should be called before or right after release

Dockerfile
```diff
@@ -646,7 +656,8 @@ RUN --mount=type=cache,target=/.cache FILES="$(gofumports -l -local github.com/t
 FROM base AS lint-protobuf
 WORKDIR /src/api
 COPY api .
-RUN prototool lint --protoc-bin-path=/toolchain/bin/protoc --protoc-wkt-path=/toolchain/include
+RUN --mount=type=cache,target=/.cache prototool lint --protoc-bin-path=/toolchain/bin/protoc --protoc-wkt-path=/toolchain/include
+RUN --mount=type=cache,target=/.cache prototool break check --descriptor-set-path=api.descriptors --protoc-bin-path=/toolchain/bin/protoc --protoc-wkt-path=/toolchain/include
```
AlekSi, line 660: That will fail on breaking changes. If they are intentional, use `make api-descriptors`.

Makefile
Unix4ever: I wonder if we can call it from the `hack/release.sh` script.

AlekSi: tbh, I'm not sure. It might be better to update descriptors right after release. At least that's how I used that tool previously, but there were no breaking API changes in that project; Talos will have them before 1.0.

smira: it could be that we even run it after each API change, can't we? (if we proved that it's backwards compatible)

but probably each release makes more sense

smira (approved):

PR #4616, opened 2021-11-29
test: bump AWS and GCP tests to use CAPI v1beta1

Unix4ever:
Both AWS and GCP should support that.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4616)


smira:
@Unix4ever I believe this can be closed, as @rsmitty has this already updated

Unix4ever:
Yep, correct

PR #4632, opened 2021-12-01
feat: autocomplete nodes, context and resource definitions

nberlee:
# Pull Request

## What? (description)
Makes `talosctl` autocomplete the most used dynamic positional parameters like resource definitions, IDs of resource definitions, and also values for arguments like `--nodes` and `--context`.

## Why? (reasoning)
As a CLI user I want uninterruptible flow of typing, not wanting to CTL-C in the middle of the command because I forgot how the resource definition was called, not knowing the node ip anymore or forgot the config context name. This commit fixes this all by simply tab-bing thru all the options.

## Acceptance
This is my first substantional amount of go lines. Please be critical. I want to do this the right way, this is how I learn... I won't be able to learn if you are lenient.

Potential problems:
- CompleteNodes is `get.po`, maybe it should be moved to `completion.go` but it will come with a lot of includes and making some funcs global in `get.go`.
- Currently aliases of Resource Definitions, but this needs some aditional work afterwards, as some aliases are defined for more than one RD. Like `ls`, `hs` and `ksps`.
- `getResourcesResponse` and `completeResource` have `// nolint:gocyclo`
- returned completions may not be unique depending if you query more `nodes`. This is not a problem as the completion implementation of `fish` and `bash` sort and makes completion responses unique.
- And probably a lot more due to my inexperience in coding in this project.

- [ ] you ran unit-tests (`make unit-tests`) - Fails on my machine with: `granting entitlement security.insecure is not allowed by build daemon configuration`

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4632)

### DIFF
--- a/cmd/talosctl/cmd/talos/config.go
+++ b/cmd/talosctl/cmd/talos/config.go
@@ -134,6 +134,7 @@ var configContextCmd = &cobra.Command{
 
 		return nil
 	},
+	ValidArgsFunction: CompleteConfigContext,
 }
 
 // configAddCmdFlags represents the `config add` command flags.
@@ -408,6 +409,23 @@ var configInfoCmd = &cobra.Command{
 
 		return nil
 	},
+}
+
+// CompleteConfigContext represents tab completion for `--context` argument and `config context` command.
+func CompleteConfigContext(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	c, err := clientconfig.Open(Talosconfig)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	contextnames := make([]string, 0, len(c.Contexts))
+	for contextname := range c.Contexts {
+		contextnames = append(contextnames, contextname)
+	}
+
+	sort.Strings(contextnames)
+
+	return contextnames, cobra.ShellCompDirectiveNoFileComp
 }
 
 func init() {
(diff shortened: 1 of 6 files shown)


cmd/talosctl/cmd/talos/config.go
```diff
@@ -410,6 +411,26 @@ var configInfoCmd = &cobra.Command{
...
 }
 
+// CompleteConfigContext represents tab completion for `--context` argument and `config` command.
+func CompleteConfigContext(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	c, err := clientconfig.Open(Talosconfig)
+	if err != nil {
+		return nil, cobra.ShellCompDirectiveError
+	}
+
+	contextnames := make([]string, len(c.Contexts))
```
smira, line 421:
```suggestion
	contextnames := make([]string, 0, len(c.Contexts))

         for contextname := range c.Contexts {
            contextnames := append(contextnames, contextname)
          }
```

smira:
<a href="https://asciinema.org/a/lwZbEvChF8ugC0SIJDgIbb5qW" target="_blank"></a>

smira:
I did some minor changes to replace some resource types with constants and added support to complete `--nodes` when no nodes are set in the `talosconfig`

smira (approved):

PR #4636, opened 2021-12-02
feat: allow overriding KSPP kernel parameters

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/4385

Now sysctls defined in the config can override kernel args defined by
defaults controller.
In that case controller shows the warning that tells which param was
overridden, new value and tells that it is not recommended.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4636)

### DIFF
--- a/pkg/machinery/resources/runtime/kernel_params_spec.go
+++ b/pkg/machinery/resources/runtime/kernel_params_spec.go
@@ -19,6 +19,14 @@ const NamespaceName resource.Namespace = v1alpha1.NamespaceName
 // KernelParamSpecType is type of KernelParam resource.
 const KernelParamSpecType = resource.Type("KernelParamSpecs.runtime.talos.dev")
 
+// KernelParamDefaultSpecType is type of KernelParam resource for default kernel params.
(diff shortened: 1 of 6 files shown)


pkg/machinery/resources/runtime/kernel_params_spec.go
```diff
@@ -43,6 +46,18 @@ func NewKernelParamSpec(namespace resource.Namespace, id resource.ID) *KernelPar
 	return r
 }
 
+// NewKernelParamDefaultSpec initializes a KernelParamSpec resource.
+func NewKernelParamDefaultSpec(namespace resource.Namespace, id resource.ID) *KernelParamSpec {
```
smira, line 50: this is a bit hacky, but it won't let us see the resource with `talosctl get`, as ResourceDefinition mentions only one type of the two

Unix4ever: yeah... I know. But I couldn't find any other options so far.
We can also use a different namespace or prefix resource ids with something, but that also sounds hacky.

smira (approved):
:+1:

PR #4658, opened 2021-12-06
fix: increase boot and etcd join timeouts

### LINKED ISSUE #4626: Increase adoption timeout
### Description

It's really a mix of a feature request and a bug report.

When we boot control plane nodes they end up in a state where they wait to be bootstrapped, which has a timeout, and then reboots.

This timeout period needs to be drastically increased! I'd be happy with a few hours/day(s).

Out problem right now is that we boot a number of nodes, which never finish at the same time (the difference has a tendency to be more than that timeout period), we have maybe 2 out of 3 nodes ready while the 3rd lags behind. By the time the 3rd is ready one or both of the first nodes will timeout and reboot, meaning the cluster can't be bootstrapped until they're all there.

I know that control plane nodes can join the cluster after the fact, it's more of a convenience thing.

smira:
Fixes #4626

On bare metal, booting machines might take a long time to go through the
whole BIOS boot sequence, or pulling images might take considerably
longer than expected.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4658)

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -357,7 +357,7 @@ func buildInitialCluster(ctx context.Context, r runtime.Runtime, name, ip string
 		lastNag time.Time
 	)
 
-	err = retry.Constant(10*time.Minute,
+	err = retry.Constant(constants.EtcdJoinTimeout,
 		retry.WithUnits(3*time.Second),
 		retry.WithJitter(time.Second),
 		retry.WithErrorLogging(true),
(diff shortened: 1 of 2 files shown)


frezbo:
should this be configurable by an end user? A cluster that partially booted on first try (network issue/lost packet) might take now more time to reboot when it's possible that the next reboot fixes the issue.

smira:
> should this be configurable by an end user? A cluster that partially booted on first try (network issue/lost packet) might take now more time to reboot when it's possible that the next reboot fixes the issue.

Talos anyways retries all transient errors, so this reboot in 15 minutes isn't going to solve them usually.

frezbo (approved):

PR #4702, opened 2021-12-17
test: retry `talosctl time` call in the tests

smira:
As `talosctl time` relies on default time server set in the config, and
our nodes start with `pool.ntp.org`, sometimes request to the timeserver
fails failing the tests.

Retry such errors in the tests to avoid spurious failures.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4702)

### DIFF
--- a/internal/pkg/miniprocfs/testdata/1920080/exe
+++ b/internal/pkg/miniprocfs/testdata/1920080/exe
@@ -1,1 +1,1 @@
-/usr/bin/fish
+usr/bin/fish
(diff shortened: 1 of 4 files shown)


internal/pkg/miniprocfs/testdata/1920080/exe
```diff
@@ -1 +1 @@
-/usr/bin/fish
```
smira, line 1: this fixes some weirdness with `go-git` based checkouts of our repo, it doesn't like absolute symlinks

this is all testdata, and tests don't care, so why not fix it

Unix4ever (approved):

PR #4706, opened 2021-12-17
fix: config apply immediate

smira:
Due to using same config value for comparison and rewriting patched
configs, immediate apply mode was always no-op.

Fix that by properly deep copying the config.

Bug was only introduced in `master` with refactoring.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4706)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -80,7 +80,7 @@ require (
 	github.com/scaleway/scaleway-sdk-go v1.0.0-beta.7
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.0
-	github.com/talos-systems/crypto v0.3.4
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -80,7 +80,7 @@ require (
 	github.com/scaleway/scaleway-sdk-go v1.0.0-beta.7
 	github.com/spf13/cobra v1.2.1
 	github.com/stretchr/testify v1.7.0
-	github.com/talos-systems/crypto v0.3.4
+	github.com/talos-systems/crypto v0.3.5-0.20211220133734-6fa2d93d0382
```
smira, line 83: this is PR https://github.com/talos-systems/crypto/pull/23

rsmitty (approved):

frezbo (approved):

PR #4718, opened 2021-12-22
test: attempt number on two on proper retries in CLI time tests

smira:
See #4702

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4718)


smira:
The issue was that `*.cmd.Cmd` can't be executed multiple times - only once, so if we need to retry, we need to build new `*cmd.Cmd`

Unix4ever (approved):

Ulexus (approved):

PR #4728, opened 2021-12-22
feat: shell completion for `list`, `read`

frezbo:
Add shell completions for `list` and `read`

https://user-images.githubusercontent.com/18496730/147137557-d7f719b0-2fe9-46cb-96b2-6f8ee77917ff.mp4

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4728)

### DIFF
--- a/cmd/talosctl/cmd/talos/root.go
+++ b/cmd/talosctl/cmd/talos/root.go
@@ -8,11 +8,15 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/talos-systems/crypto/x509"
+	"google.golang.org/grpc/codes"
 
 	"github.com/talos-systems/talos/pkg/cli"
+	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
 	"github.com/talos-systems/talos/pkg/machinery/client"
 	clientconfig "github.com/talos-systems/talos/pkg/machinery/client/config"
 )
@@ -26,6 +30,8 @@ var (
 	Nodes       []string
 	Cmdcontext  string
 )
+
+const pathAutoCompleteLimit = 500
 
 // WithClientNoNodes wraps common code to initialize Talos client and provide cancellable context.
 //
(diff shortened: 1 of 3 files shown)


cmd/talosctl/cmd/talos/root.go
```diff
@@ -123,3 +127,92 @@ var Commands []*cobra.Command
...
+		pathToSearch = "/"
+	}
+
+	var paths map[string]struct{}
+
+	// search up one level to find possible completions
+	if pathToSearch != "/" && !strings.HasSuffix(pathToSearch, "/") {
+		splits := strings.Split(pathToSearch, "/")
+		// we need a trailing slash to search for items in a directory
+		pathToSearch = strings.Join(splits[:len(splits)-1], "/") + "/"
```
smira, line 146: we could also do `strings.LastIndex` (https://pkg.go.dev/strings#LastIndex) to find the index of the last `/` and use the string before it as `pathToSearch`

not that it matters, but we don't do splits which are not needed actually

cmd/talosctl/cmd/talos/root.go
```diff
@@ -123,3 +127,92 @@ var Commands []*cobra.Command
...
+	}
+
+	paths = getPathFromNode(pathToSearch)
+
+	result := make([]string, 0, len(paths))
+
+	for k := range paths {
+		result = append(result, k)
```
smira, line 154: really a nit, but if we filter here, we could avoid building the list twice

or it might be even more fancy to avoid building full list in `getPathFromNode` by applying filtering there? this way if the directory listing is huge, we never have to store it all in memory

cmd/talosctl/cmd/talos/root.go
```diff
@@ -123,3 +127,92 @@ var Commands []*cobra.Command
...
+			// directories have a trailing slash
+			if resp.IsDir {
+				paths[path+resp.RelativeName+"/"] = struct{}{}
+			} else {
+				paths[path+resp.RelativeName] = struct{}{}
+			}
+		}
+	}) != nil {
+		return paths
+	}
```
smira, line 215: should we limit number of paths we look for? I don't know if it makes sense to add some hard limit on the paths we return to avoid iterating forever?

cmd/talosctl/cmd/talos/root.go
```diff
@@ -123,3 +127,92 @@ var Commands []*cobra.Command
...
+
+	return result
+}
+
+//nolint:gocyclo
+func getPathFromNode(path string) map[string]struct{} {
+	paths := make(map[string]struct{})
+
+	if WithClient(func(ctx context.Context, c *client.Client) error {
+		stream, err := c.LS(ctx, &machineapi.ListRequest{
```
smira, line 177: related to the comment below, we could also make sure we abort the `LS` call on the server if we exit early:

```suggestion
                ctx, cancel := context.WithCancel(ctx)
                defer cancel()

		stream, err := c.LS(ctx, &machineapi.ListRequest{
```

this way if we exit this function, we always abort the API on the server side (so it stops listing)

smira (commented):
looks good to me :+1: just couple of nits thinking about the file listing being potentially huge

cmd/talosctl/cmd/talos/root.go
frezbo: fixed

cmd/talosctl/cmd/talos/root.go
frezbo: fixed

cmd/talosctl/cmd/talos/root.go
frezbo: fixed

cmd/talosctl/cmd/talos/root.go
```diff
@@ -123,3 +129,103 @@ var Commands []*cobra.Command
...
+		// we need a trailing slash to search for items in a directory
+		pathToSearch = pathToSearch[:index] + "/"
+	}
+
+	paths = getPathFromNode(pathToSearch, inputPath)
+
+	result := make([]string, 0, len(paths))
+
+	for k := range paths {
+		if strings.HasPrefix(k, inputPath) {
```
smira, line 156: do we need this check if `getPathFromNode` already filtered the output?

smira (approved):

cmd/talosctl/cmd/talos/root.go
frezbo: not needed, forgot about it here, fixed

frezbo:
The only thing missing is if we hit a symlink it won't auto complete.

kurious-diru (approved):

PR #4731, opened 2021-12-23
docs: update Mayastor deployment process

frezbo:
Update Mayastor docs with latest requirements

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4731)

### DIFF
--- a/website/content/docs/v0.14/Guides/storage.md
+++ b/website/content/docs/v0.14/Guides/storage.md
@@ -17,6 +17,8 @@ It is easy and automatic.
 
 ## Storage Clusters
 
+> **Talos** recommends having a separate disks (apart from the Talos install disk) to be used for storage.
+
 Redundancy in storage is usually very important.
 Scaling capabilities, reliability, speed, maintenance load, and ease of use are all factors you must consider when managing your own storage.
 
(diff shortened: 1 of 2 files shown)


website/content/docs/v0.14/Guides/storage.md
```diff
@@ -78,112 +80,33 @@ To see a live demo of this section, see the video below:
...
-apiVersion: "openebs.io/v1alpha1"
-kind: MayastorPool
-metadata:
-  name: pool-on-talos-xxx
-  namespace: mayastor
-spec:
-  node: talos-xxx
-  disks: ["/dev/sdx"]
-EOF
+talosctl patch machineconfig -n <node ip> --config-patch '[{"op": "add", "path": "/machine/sysctls", "value": {"vm.nr_hugepages": "1024"}}, {"op": "add", "path": "/machine/kubelet/extraArgs", "value": {"node-labels": "openebs.io/engine=mayastor"}}]'
```
smira, line 98: `nit`: we can do `--immediate` here to avoid reboot

website/content/docs/v0.14/Guides/storage.md
```diff
@@ -78,112 +80,33 @@ To see a live demo of this section, see the video below:
...
 ```
 
-### Create StorageClass
-
-With the pools created for each node, create a storage class that uses the `nvmf` protocol, updating the number of replicas as necessary:
+> Note: If you are adding/updating the `vm.nr_hugepages` on a node which already had the `openebs.io/engine=mayastor` label set, you'd need to restart kubelet so that it picks up the new value, by issuing the following command
```
smira, line 101: I believe in 0.14 kubelet should be restarted if extraArgs are changed

website/content/docs/v0.14/Guides/storage.md
frezbo: good catch, fixing

website/content/docs/v0.14/Guides/storage.md
frezbo: this is when the node already contain the labels and the sysctl has been changed, will it automatically restart kubelet? I had to restart kubelet manually for it to pick the new sysctl

Unix4ever (approved):

website/content/docs/v0.14/Guides/storage.md
smira: well, `sysctl` should be applied immediately, not sure if `kubelet` needs a restart to update something

I was referring more to the labels

smira (approved):

PR #4732, opened 2021-12-23
feat: provide a way to load Linux kernel modules

### LINKED ISSUE #4693: loading Linux kernel modules in Talos
See also #4692 

Machine configuration should contain a list of modules to be loaded. Modules are supposed to be stored on the rootfs in some predefined location (e.g. `/lib/modules`).

`machined` should load modules early in the boot process and fails if the module fails to be loaded.

Hopefully we can do that without `modutils` by using syscalls directly from Go (see https://linux.die.net/man/2/init_module).

smira:
Fixes #4693

Machine configuration change plus very simple controllers which tranform
config into the module spec and finally load modules.

There's no support for advanced features like module params and aliases,
but we can add it later.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4732)


smira:
Note: at the moment, there are no official Linux kernel modules shipped with Talos.

Unix4ever (approved):

PR #4734, opened 2021-12-24
feat: introduce `talos.experimental.wipe` kernel param to wipe system disk

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/4399

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4734)

### DIFF
--- a/website/content/docs/v0.15/Guides/resetting-a-machine.md
+++ b/website/content/docs/v0.15/Guides/resetting-a-machine.md
@@ -7,6 +7,8 @@ From time to time, it may be beneficial to reset a Talos machine to its "origina
 Bear in mind that this is a destructive action for the given machine.
 Doing this means removing the machine from Kubernetes, Etcd (if applicable), and clears any data on the machine that would normally persist a reboot.
 
+## CLI
+
 > WARNING: Running a `talosctl reset` on cloud VM's might result in the VM being unable to boot as this wipes the entire disk.
 It might be more useful to just wipe the `STATE` and `EPHEMERAL` partitions on a cloud VM if not booting via `iPXE`.
 `talosctl reset --system-labels-to-wipe STATE --system-labels-to-wipe EPHEMERAL`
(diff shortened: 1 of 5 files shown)


smira (commented):
we need docs for this (we have a document which lists all available kernel args) and a note in `release.toml`.

andrewrynhard:
what is `exp`?

Unix4ever:
> what is exp?

`experimental` I guess

Unix4ever:
> note in `release.toml`.

Ok, then I need [this](https://github.com/talos-systems/talos/pull/4735) to be merged :)

andrewrynhard:
> > what is exp?
> 
> `experimental` I guess

I know `exp` is less to type, but I would say we should use `experimental`. It describes itself and in general this is something I think we should do/are doing in Talos (using the full names and not shortening them). Just my two cents. This is a minor nit.

Unix4ever:
> > > what is exp?
> > 
> > `experimental` I guess
> 
> I know `exp` is less to type, but I would say we should use `experimental`. It describes itself and in general this is something I think we should do/are doing in Talos (using the full names and not shortening them). Just my two cents. This is a minor nit.

Sure, why not. Looks like the limit is 4096 bytes, so maybe we shouldn't be stingy about that limit.

website/content/docs/v0.15/Guides/resetting-a-machine.md
```diff
@@ -25,3 +27,11 @@ The `graceful` flag is especially important when considering HA vs. non-HA Talos
 If the machine is part of an HA cluster, a normal, graceful reset should work just fine right out of the box as long as the cluster is in a good state.
 However, if this is a single node cluster being used for testing purposes, a graceful reset is not an option since Etcd cannot be "left" if there is only a single member.
 In this case, reset should be used with `--graceful=false` to skip performing checks that would normally block the reset.
+
+## Kernel Parameter
```
smira, line 31: :+1:

website/content/docs/v0.15/Guides/resetting-a-machine.md
```diff
@@ -25,3 +27,11 @@ The `graceful` flag is especially important when considering HA vs. non-HA Talos
...
+
+Another way to reset a machine is to specify `talos.experimental.wipe=system` kernel parameter.
+If the machine got stuck in the boot loop and you access to the console you can use GRUB to specify this kernel argument.
+Then when Talos boots for the next time it will reset system disk and reboot.
+
+Next steps can be to install Talos either usign PXE boot or by mounting an ISO.
```
smira, line 37: could you please also add it to https://github.com/talos-systems/talos/blob/05fd16adfde59550b3f8982262319250871390f7/website/content/docs/v0.15/Reference/kernel.md

lion7: There is a typo in the last sentence: usign instead of using 😉

smira (approved):

PR #4736, opened 2021-12-24
feat: make machine configuration read-only in Talos (almost)

smira:
Talos shouldn't try to re-encode the machine config it was provided
with.

So add a `ReadonlyWrapper` around `*v1alpha1.Config` which makes sure
that raw config object is not available anymore (it's a private field),
but config accessors are available for read-only access.

Another thing that `ReadonlyWrapper` does is that it preserves the
original `[]byte` encoding of the config keeping it exactly same way as
it was loaded from file or read over the network.

Improved `talosctl edit mc` to preserve the config as it was submitted,
and preserve the edits on error from Talos (previously edits were lost).

`ReadonlyWrapper` is not used on config generation path though - config
there is represented by `*v1alpha.Config` and can be freely modified.

Why almost? Some parts of Talos (platform code) patch the machine
configuration with new data. We need to fix platforms to provide
networking configuration in a different way, but this will come with
other PRs later.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4736)

### DIFF
--- a/cmd/talosctl/cmd/talos/edit.go
+++ b/cmd/talosctl/cmd/talos/edit.go
@@ -51,18 +51,21 @@ func editFn(c *client.Client) func(context.Context, client.ResourceResponse) err
 			return nil
 		}
 
+		metadata := msg.Resource.Metadata()
+		id := metadata.ID()
+
+		body, err := yaml.Marshal(msg.Resource.Spec())
+		if err != nil {
+			return err
+		}
+
+		edited := body
+
 		for {
 			var (
 				buf bytes.Buffer
 				w   io.Writer = &buf
-				id  string
 			)
-
-			metadata := msg.Resource.Metadata()
-
-			if metadata != nil {
-				id = metadata.ID()
-			}
 
 			if runtime.GOOS == "windows" {
 				w = crlf.NewCRLFWriter(w)
@@ -81,40 +84,32 @@ func editFn(c *client.Client) func(context.Context, client.ResourceResponse) err
 				lines := strings.Split(lastError, "\n")
 
 				_, err = w.Write([]byte(
-					fmt.Sprintf("#\n# %s\n", strings.Join(lines, "\n# ")),
+					fmt.Sprintf("# \n# %s\n", strings.Join(lines, "\n# ")),
 				))
 				if err != nil {
 					return err
 				}
 			}
 
-			body, err := yaml.Marshal(msg.Resource.Spec())
+			_, err = w.Write(edited)
 			if err != nil {
 				return err
 			}
 
-			_, err = w.Write(body)
+			edited, _, err = edit.LaunchTempFile(fmt.Sprintf("%s-%s-edit-", msg.Resource.Metadata().Type(), id), ".yaml", &buf)
 			if err != nil {
 				return err
 			}
 
-			edited, _, err := edit.LaunchTempFile(fmt.Sprintf("%s-%s-edit-", msg.Resource.Metadata().Type(), id), ".yaml", &buf)
-			if err != nil {
-				return err
-			}
+			edited = stripEditingComment(edited)
 
-			editedWithoutComments := bytes.TrimSpace(cmdutil.StripComments(edited))
-
-			if len(bytes.TrimSpace(editedWithoutComments)) == 0 {
+			if len(bytes.TrimSpace(bytes.TrimSpace(cmdutil.StripComments(edited)))) == 0 {
 				fmt.Println("Apply was skipped: empty file.")
 
 				break
 			}
 
-			if bytes.Equal(
-				editedWithoutComments,
-				bytes.TrimSpace(cmdutil.StripComments(body)),
-			) {
+			if bytes.Equal(edited, body) {
 				fmt.Println("Apply was skipped: no changes detected.")
 
 				break
@@ -141,6 +136,21 @@ func editFn(c *client.Client) func(context.Context, client.ResourceResponse) err
 		}
 
 		return nil
+	}
+}
+
+func stripEditingComment(in []byte) []byte {
+	for {
+		idx := bytes.Index(in, []byte{'\n'})
+		if idx == -1 {
+			return in
+		}
+
+		if !bytes.HasPrefix(in, []byte("# ")) {
+			return in
+		}
+
+		in = in[idx+1:]
 	}
 }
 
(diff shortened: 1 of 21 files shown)


cmd/talosctl/cmd/talos/edit.go
```diff
@@ -144,6 +139,21 @@ func editFn(c *client.Client) func(context.Context, client.ResourceResponse) err
 	}
 }
 
+func stripEditingComment(in []byte) []byte {
```
smira, line 142: we strip comments which start with `# ` from the header, as these comments are injected by the edit process and if we don't strip them, they will end up in the persisted machine config (as config is now preserved byte by byte)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
```diff
@@ -65,24 +65,12 @@ func (r *Runtime) SetConfig(cfg config.Provider) error {
...
-		return fmt.Errorf("error serializing current config: %w", err)
-	}
-
-	currentConfigProvider, err := configloader.NewFromBytes(currentBytes)
-	if err != nil {
-		return fmt.Errorf("error loading current config: %w", err)
-	}
-
-	currentConfig, ok := currentConfigProvider.(*v1alpha1.Config)
+	currentConfig, ok := r.Config().Raw().(*v1alpha1.Config)
```
smira, line 68: I don't think that Talos ever modifies machine config structures while it's being accessed via interface methods, so it should be safe to remove it.

pkg/cluster/kubernetes/detect.go
```diff
@@ -34,12 +34,6 @@ func DetectLowestVersion(ctx context.Context, cluster UpgradeProvider, options U
 	}
 
 	var version *semver.Version
-	if options.ToVersion != "" {
```
smira, line 37: I'm not sure why this block was here (?), but it fails when _downgrading_ via `talosctl upgrade-k8s`: as we start detection with `--to` version, it's always smaller than any version in the cluster.

E.g. cluster runs v1.23.1, and I want to downgrade to v1.23.0:

```
talosctl upgrade-k8s --to=1.23.0
```

This will always auto-detect running version as 1.23.0, while correct should be 1.23.1

cc @Unix4ever  did I miss anything?

pkg/cluster/kubernetes/kubelet.go
```diff
@@ -224,6 +224,8 @@ func upgradeKubeletPatcher(options UpgradeOptions, kubeletSpec resource.Resource
 				version = parts[1]
 			}
 
+			version = strings.TrimLeft(version, "v")
```
smira, line 227: cosmetic fix: this prints "old" version properly without `v`

pkg/machinery/config/configloader/configloader.go
```diff
@@ -34,7 +34,7 @@ func newConfig(source []byte) (config config.Provider, err error) {
 	// a special way.
 	for _, manifest := range manifests {
 		if talosconfig, ok := manifest.(*v1alpha1.Config); ok {
-			return talosconfig, nil
+			return v1alpha1.WrapReadonly(talosconfig, source), nil
```
smira, line 37: this is the gist of the PR :rofl: 

when machine config is read from file (or from network), we make it read-only and we preserve the `[]byte` source

pkg/machinery/config/provider.go
```diff
@@ -20,15 +20,25 @@ import (
...
 	// Validate checks configuration and returns warnings and fatal errors (as multierror).
 	Validate(RuntimeMode, ...ValidationOption) ([]string, error)
-	String(encoderOptions ...encoder.Option) (string, error)
-	Bytes(encoderOptions ...encoder.Option) ([]byte, error)
+
+	// Bytes returns source YAML representation (if available) or does default encoding.
+	Bytes() ([]byte, error)
+
+	// Encode configuration to YAML using the provided options.
+	EncodeString(encoderOptions ...encoder.Option) (string, error)
```
smira, line 37: I renamed `Bytes` and `Strings` to `EncodeXXX`, as with custom options they should always do marshaling to YAML (as options affect the way config is marshaled).

The method `Bytes` returns original `[]byte` representation if available.

pkg/machinery/config/provider.go
```diff
@@ -20,15 +20,25 @@ import (
...
+	EncodeBytes(encoderOptions ...encoder.Option) ([]byte, error)
+
+	// Raw returns internal config representation.
+	Raw() interface{}
```
smira, line 41: `Raw()` allows to access underlying `*v1alpha1.Config`, but due to dependency cycle we can't have exact type here.

pkg/machinery/config/types/v1alpha1/v1alpha1_readonly_provider.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// EncodeBytes implements the config.Provider interface.
+func (r *ReadonlyProvider) EncodeBytes(encoderOptions ...encoder.Option) ([]byte, error) {
+	return r.cfg.EncodeBytes(encoderOptions...)
+}
+
+// Raw implements the config.Provider interface.
+func (r *ReadonlyProvider) Raw() interface{} {
+	return r.cfg.DeepCopy()
```
smira, line 83: NB: accessing raw data structures always does copying, so that original read-only config is preserved as is.

pkg/machinery/resources/config/machine_config.go
```diff
@@ -31,8 +30,9 @@ type v1alpha1Spec struct {
 	cfg config.Provider
 }
 
-func (s *v1alpha1Spec) MarshalYAML() (interface{}, error) {
-	return encoder.NewEncoder(s.cfg).Marshal()
+// MarshalYAMLBytes implements RawYAML interface.
+func (s *v1alpha1Spec) MarshalYAMLBytes() ([]byte, error) {
+	return s.cfg.Bytes()
```
smira, line 35: use this trick to return machine config exact byte by byte YAML representation

pkg/cluster/kubernetes/detect.go
Unix4ever: Yeah, I think your fix is correct.

Unix4ever (approved):

PR #4742, opened 2021-12-27
fix: kexec on RPI4

frezbo:
Fixes kexec on RPI4

Since BL31 is now enabled `talosctl shutdown` puts the Pi in a low power state instead of rebooting

Ref: https://github.com/talos-systems/pkgs/pull/375

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4742)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ NAME = Talos
 
 ARTIFACTS := _out
 TOOLS ?= ghcr.io/talos-systems/tools:v0.10.0-alpha.0
-PKGS ?= v0.10.0-alpha.0-3-g950361f
+PKGS ?= v0.10.0-alpha.0-4-g2779c3f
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,7 +12,7 @@ DOCKER_LOGIN_ENABLED ?= true
 NAME = Talos
 
 ARTIFACTS := _out
-TOOLS ?= ghcr.io/talos-systems/tools:v0.10.0-alpha.0
+TOOLS ?= ghcr.io/talos-systems/tools:v0.10.0-alpha.0-4-g2779c3f
```
smira, line 15: should it be `pkgs` ?

frezbo: good catch, pasted at the wrong place

smira (approved):

rsmitty (approved):

PR #4778, opened 2022-01-10
feat: replace flags with --mode in `apply`, `edit` and `patch` commands

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/4588

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4778)


api/machine/machine.proto
```diff
@@ -73,16 +73,33 @@ service MachineService {
 // ApplyConfiguration describes a request to assert a new configuration upon a
 // node.
 message ApplyConfigurationRequest {
+  enum Mode {
+    AUTO = 0;
```
smira, line 77: should the default be `REBOOT = 0` so that we have backwards compatibility with clients? otherwise how do we know that `talosctl apply-config` from v0.14 is using?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -160,14 +185,17 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
...
+	case machine.ApplyConfigurationRequest_NO_REBOOT:
 		if err := s.Controller.Runtime().SetConfig(cfgProvider); err != nil {
 			return nil, err
 		}
-	// default, no `--on-reboot`
-	case !in.OnReboot:
+	// --mode=staged
+	case machine.ApplyConfigurationRequest_STAGED:
+	// --mode=reboot
+	default:
```
smira, line 198: should this be explicit `--mode=reboot` vs. `default:` ? 

and other modes should not be available at this point (so we could through an error)

internal/integration/api/apply-config.go
```diff
@@ -334,6 +340,38 @@ func (suite *ApplyConfigSuite) TestApplyConfigRotateEncryptionSecrets() {
...
+	cfg.MachineConfig.MachineType = "controlplane"
+
+	cfgDataOut, err := cfg.Bytes()
+	suite.Assert().Nilf(err, "failed to marshal updated machine config data (node %q): %w", node, err)
+
+	_, err = suite.Client.ApplyConfiguration(nodeCtx, &machineapi.ApplyConfigurationRequest{
+		Data: cfgDataOut,
+		Mode: machineapi.ApplyConfigurationRequest_NO_REBOOT,
+	})
+	suite.Require().Error(err)
```
smira, line 372: should we be more specific on the error here? otherwise it might be that we can't actually distinguish the error

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -137,20 +137,45 @@ func (s *Server) Register(obj *grpc.Server) {
...
-	// --immediate
-	if in.Immediate {
+	autoApplyInfo := ""
+
+	//nolint:exhaustive
+	switch in.Mode {
+	// --mode=no-reboot
+	case machine.ApplyConfigurationRequest_NO_REBOOT:
 		if err = s.Controller.Runtime().CanApplyImmediate(cfgProvider); err != nil {
 			return nil, err
```
smira, line 162: we could wrap that with `gRPC` status code, so that error is easily distinguishable from the regular failure.

api/machine/machine.proto
```diff
@@ -73,16 +73,33 @@ service MachineService {
...
+  Mode mode = 4;
 }
 
 // ApplyConfigurationResponse describes the response to a configuration request.
 message ApplyConfiguration {
   common.Metadata metadata = 1;
   // Configuration validation warnings.
   repeated string warnings = 2;
+  // Human readable auto mode apply detection result.
+  string auto_apply_info = 3;
```
smira, line 102: should we make it a bit more structured for easier API consumption (and probably in the future with `--dry-run`) by separating two things:

* `Mode mode` which states which mode was actually chosen (passes through mode most of the time, but for AUTO mode it returns either reboot or no-reboot).
* `string message` which might be a generic human-readable message explaining the result of the apply configuration call?

hack/release.toml
```diff
@@ -15,6 +15,19 @@ preface = """\
 
 [notes]
 
+    [notes.applyconfig]
+        title = "Apply Config Enhancements"
+        description="""\
+`talosctl apply/patch/edit` cli commands got rewamped.
```
smira, line 21:
```suggestion
`talosctl apply/patch/edit` cli commands got revamped.
```

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -160,14 +185,17 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 		return nil, err
 	}
 
-	switch {
-	// --immediate
-	case in.Immediate:
+	//nolint:exhaustive
```
smira, line 188: to be safe, we could probably add an option `default:` which returns an error (it should never be reached, but still)

internal/integration/api/apply-config.go
```diff
@@ -135,62 +138,67 @@ func (suite *ApplyConfigSuite) TestApply() {
...
-	suite.Require().NoError(err, "failed to apply deferred configuration (node %q): %w", node)
+		_, err = suite.Client.ApplyConfiguration(nodeCtx, &machineapi.ApplyConfigurationRequest{
+			Data: cfgDataOut,
+			Mode: mode,
+		})
+		suite.Require().NoError(err, "failed to apply deferred configuration (node %q): %w", node)
+	}
 }
 
 // TestApplyConfigRotateEncryptionSecrets verify key rotation by sequential apply config calls.
```
smira, line 204: should we update this function as well?

smira (commented):
I think we should also update the docs for 0.15, I'm pretty sure we have tons of references to these commands in the guides

internal/integration/api/apply-config.go
Unix4ever: Do you mean that we should add explicit `mode` here? Auto mode should pick reboot anyways.

smira: ah yeah, you're right... it wasn't immediate

smira (approved):
:cool:

smira (approved):

PR #4786, opened 2022-01-11
hack: PoC for system extensions

smira:
This doesn't cover any real work to be done with installing system
extensions, but rather just shows that the implementation is possible.

End result is that we have extra file in `/usr/bin`:

```
$ talosctl -n 172.20.0.2 ls -l /usr/bin
NODE         MODE         UID   GID   SIZE(B)   LASTMOD           NAME
172.20.0.2   drwxr-xr-x   0     0     32        Jan 10 23:46:33   .
172.20.0.2   -rw-r--r--   0     0     0         Jan 10 23:46:33   extension
```

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4786)


smira:
From `/proc/mounts`:

```
overlay / overlay ro,relatime,lowerdir=/layers/layer2:/layers/layer1 0 0
```

PR #4796, opened 2022-01-14
fix: avoid panic in config loading/validation

smira:
As wrong `error` value was used, publishing an event was causing a panic
in Talos.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4796)


Unix4ever (approved):
good catch

frezbo (approved):

PR #4797, opened 2022-01-14
feat: print version in Grub

andrewrynhard:
It will be useful to know which version of Talos is booting as early as
possible. This adds the Talos version to the Grub label so that Grub
will print the Talos version.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4797)

### DIFF
--- a/cmd/installer/cmd/root.go
+++ b/cmd/installer/cmd/root.go
@@ -44,4 +44,5 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&options.Upgrade, "upgrade", false, "Indicates that the install is being performed by an upgrade")
 	rootCmd.PersistentFlags().BoolVar(&options.Force, "force", false, "Indicates that the install should forcefully format the partition")
 	rootCmd.PersistentFlags().BoolVar(&options.Zero, "zero", false, "Indicates that the install should write zeros to the disk before installing")
+	rootCmd.PersistentFlags().StringVar(&options.PreviousVersion, "previous-version", "", "The previous version of Talos")
 }
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
@@ -34,10 +34,11 @@ type Cfg struct {
 
 // Label reprsents a label in the cfg file.
 type Label struct {
-	Root   string
-	Kernel string
-	Initrd string
-	Append string
+	Root    string
+	Kernel  string
+	Initrd  string
+	Append  string
+	Version string
 }
 
 const grubCfgTpl = `set default="{{ .Default }}"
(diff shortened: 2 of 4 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -107,6 +108,8 @@ func (g *Grub) Labels() (current, next string, err error) {
 		return "", "", fmt.Errorf("unknown grub menuentry: %q", current)
 	}
 
+	next = fmt.Printf("%s %s", version.Tag)
```
andrewrynhard, line 111: Whoops

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -107,6 +108,8 @@ func (g *Grub) Labels() (current, next string, err error) {
 		return "", "", fmt.Errorf("unknown grub menuentry: %q", current)
 	}
 
+	next = fmt.Printf("%s %s", next, version.Tag)
```
smira, line 111: feels this would break our detection of labels via BootA/BootB

andrewrynhard: Yeah I am sure this is going to break something right now. Kind of working through it and leaning on CI right now since my dev environment is broken at the moment.

cmd/installer/cmd/root.go
```diff
@@ -44,4 +44,5 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&options.Upgrade, "upgrade", false, "Indicates that the install is being performed by an upgrade")
 	rootCmd.PersistentFlags().BoolVar(&options.Force, "force", false, "Indicates that the install should forcefully format the partition")
 	rootCmd.PersistentFlags().BoolVar(&options.Zero, "zero", false, "Indicates that the install should write zeros to the disk before installing")
+	rootCmd.PersistentFlags().StringVar(&options.PreviousVersion, "previous-version", "", "The previous version of Talos")
```
andrewrynhard, line 47: Torn on what to call this flag. From the perspective of the thing running this container (i.e. Talos) it is more accurately called "current-version".

andrewrynhard: We also refer to the "current" entry as a "fallback." So we could have `current-version` or `fallback-version`.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -53,6 +54,9 @@ terminal_output console
 
 {{ range $label := .Labels -}}
 menuentry "{{ $label.Root }}" {
+{{ if .Version }}
```
andrewrynhard, line 57: Needs to be conditional for now since it is possible that an entry doesn't have a version yet since all versions of Talos prior to the release that has this won't know to pass in the arg to specify which version it is.

smira:
superseded by #4960

PR #4819, opened 2022-01-18
feat: jetson nano SoC

frezbo:
Support Jetson Nano SoC

Depends on https://github.com/talos-systems/pkgs/pull/380

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4819)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -403,6 +403,8 @@ COPY --from=pkg-libressl-arm64 / /rootfs
 COPY --from=pkg-libseccomp-arm64 / /rootfs
 COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2 /rootfs/lib/firmware/bnx2
 COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2x /rootfs/lib/firmware/bnx2x
+COPY --from=pkg-linux-firmware-arm64 /lib/firmware/rtl_nic /rootfs/lib/firmware/rtl_nic
+COPY --from=pkg-linux-firmware-arm64 /lib/firmware/nvidia/tegra210 /rootfs/lib/firmware/nvidia/tegra210
 COPY --from=pkg-lvm2-arm64 / /rootfs
 COPY --from=pkg-libaio-arm64 / /rootfs
 COPY --from=pkg-musl-arm64 / /rootfs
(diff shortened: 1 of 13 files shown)


Dockerfile
```diff
@@ -403,6 +397,8 @@ COPY --from=pkg-libressl-arm64 / /rootfs
 COPY --from=pkg-libseccomp-arm64 / /rootfs
 COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2 /rootfs/lib/firmware/bnx2
 COPY --from=pkg-linux-firmware-arm64 /lib/firmware/bnx2x /rootfs/lib/firmware/bnx2x
+COPY --from=pkg-linux-firmware-arm64 /lib/firmware/rtl_nic /rootfs/lib/firmware/rtl_nic
```
smira, line 400: do we need this in the rootfs?

website/content/docs/v0.15/Single Board Computers/jetson_nano.md
```diff
@@ -0,0 +1,115 @@
...
+Download the latest `talosctl`.
+
+```bash
+curl -Lo /usr/local/bin/talosctl https://github.com/talos-systems/talos/releases/latest/download/talosctl-$(uname -s | tr "[:upper:]" "[:lower:]")-amd64
+chmod +x /usr/local/bin/talosctl
+```
+
+## Flashing the firmware to on-board SPI flash
+
+We will use the [R32.6.1 release](https://developer.nvidia.com/embedded/l4t/r32_release_v6.1/t210/jetson-210_linux_r32.6.1_aarch64.tbz2) for the Jetson Nano.
```
smira, line 23: I don't know if it's true, but I guess flashing firmware is one-time procedure? we don't need to flash it every time?

(does it makes sense to state it in the docs?)

Dockerfile
frezbo: yes, this driver is loaded from the `rootfs` and if I remove I see warning about failing to load

website/content/docs/v0.15/Single Board Computers/jetson_nano.md
frezbo: good catch, fixed

smira (approved):

PR #4825, opened 2022-01-19
feat: support CRI configuration merging and reimplement registry config

smira:
Containerd doesn't support merging plugin configuration from multiple
sources, and Talos has several pieces which configure CRI plugin:
(see https://github.com/containerd/containerd/issues/5837)

* base config
* registry mirror config
* system extensions
* ...

So we implement our own simple way of merging config parts (by simply
concatenating text files) to build a final `cri.toml`.

At the same time containerd migrated to a new format to specify registry
mirror configuration, while old way (via CRI config) is going to be
removed in 1.7.0. New way also allows to apply most of registry
configuration (except for auth) on the fly.

Also, containerd was updated to 1.6.0-rc.0 and runc to 1.1.0.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4825)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
```diff
@@ -89,6 +89,7 @@ func (r *Runtime) CanApplyImmediate(cfg config.Provider) error {
 	// * .machine.controlplane
 	// * .machine.kubelet
 	// * .machine.kernel
+	// * .machine.registries (not that auth is not applied immediately, containerd limitation)
```
Unix4ever, line 92:
```suggestion
	// * .machine.registries (note that auth is not applied immediately, containerd limitation)
```

we now have this list in the docs as well:
https://github.com/talos-systems/talos/blob/master/website/content/docs/v0.15/Guides/editing-machine-configuration.md?plain=1#L40

Unix4ever (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_runtime.go
smira: I already updated that doc in this PR!

Unix4ever: ah, ok. Missed that, sorry :)

PR #4835, opened 2022-01-20
feat: Multiple json files

sbskas:
Include filename content if value begins with @ (see curl for example).
Add multiple config-path option on cmdline to apply them in order.
ex: 

```bash
$ talosctl-linux-amd64 gen config talos1 https://127.0.0.1:6443 --config-patch-control-plan @cidrs.json --config-patch-worker @sysctls-workders.json --config-path @cluster-name.jsonj
```

Signed-off-by: Sébastien Bernard <sbernard@nerim.net>

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4835)

### DIFF
--- a/website/content/docs/v0.13/Guides/upgrading-talos.md
+++ b/website/content/docs/v0.13/Guides/upgrading-talos.md
@@ -10,7 +10,7 @@ The `talosctl` CLI utility will facilitate this.
 
 To see a live demo of this writeup, see the video below:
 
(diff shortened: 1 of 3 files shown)


smira:
I'm going to take your PR and do a couple of changes:

* implement same for other places where we do accept config patches
* add support for loading JSON patches from YAML transparently

smira:
I didn't actually mean to close this PR, but I managed to push it wrong way. I'll open another PR with the changes.

smira:
#4892 is the new PR

PR #4843, opened 2022-01-21
feat: implement sysfs

smira:
TODO

Co-authored-by: Dmitriy Matrenichev <lepage+gh@protonmail.com>


pkg/machinery/kernel/kernel.go
```diff
@@ -34,5 +41,12 @@ type Param struct {
 
 // Path returns the path to the systctl file under /proc/sys.
 func (prop *Param) Path() string {
-	return path.Join("/proc/sys", strings.ReplaceAll(prop.Key, ".", "/"))
+	res := strings.ReplaceAll(prop.Key, ".", "/")
+
+	// fallback to the old behavior if the key path is not absolute
+	if !strings.HasPrefix(res, "/") {
```
smira, line 47: why do we need this still? to keep backwards compatibility with other users of machinery? or do we have somewhere old way?

hack/release.toml
```diff
@@ -145,6 +145,13 @@ Talos by default skips setting up PSP now (see machine configuration `.cluster.a
         description = """\
 Talos now provides a way to extend set of system services Talos runs with extension services.
 Extension services should be included in the Talos root filesystem (e.g. via system extensions).
+"""
+
+    [notes.sysfs]
+        title = "Support Sysfs Params"
```
smira, line 151:
```suggestion
        title =  Sysfs Kernel Parameters"
```

hack/release.toml
```diff
@@ -145,6 +145,13 @@ Talos by default skips setting up PSP now (see machine configuration `.cluster.a
...
+        description="""\
+Support writing kernel parameters to `/sys`.
```
smira, line 153:
```suggestion
Talos now supports setting `sysfs` kernel parameters  (`/sys/...`).
```

hack/release.toml
```diff
@@ -145,6 +145,13 @@ Talos by default skips setting up PSP now (see machine configuration `.cluster.a
...
+To write kernel params to `/sys` `.machine.sysfs` config parameters can be used.
```
smira, line 154:
```suggestion
Use machine configuration field `.machine.sysfs` to set `sysfs` kernel parameters.
```

smira (commented):
I guess we should also add `.machine.sysfs` to the list of config values that can be applied immediately (without a reboot)

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
```diff
@@ -95,17 +95,17 @@ func (ctrl *KernelParamSpecController) Run(ctx context.Context, r controller.Run
 
 			for i, item := range list {
 				spec := item.(runtime.KernelParam).TypedSpec()
-				key := item.Metadata().ID()
+				id := item.Metadata().ID()
 
-				if value, duplicate := touchedIDs[key]; i >= configsCounts && duplicate {
-					logger.Warn("overriding KSPP enforced parameter, this is not recommended", zap.String("key", key), zap.String("value", value))
+				if value, duplicate := touchedIDs[spec.Key]; i >= configsCounts && duplicate {
+					logger.Warn("overriding KSPP enforced parameter, this is not recommended", zap.String("key", spec.Key), zap.String("value", value))
```
Ulexus, line 101: This isn't necessarily KSPP-enforced values, is it?  It's any Talos-asserted default value?

pkg/machinery/kernel/kernel.go
```diff
@@ -34,5 +41,12 @@ type Param struct {
 
 // Path returns the path to the systctl file under /proc/sys.
 func (prop *Param) Path() string {
```
Ulexus, line 43: Since this function is now only applicable to sysctls and not sysfs's, should its name be changed to better reflect that limitation, now that this package supports both?

Ulexus: At least, the following `/` prefix check relegates it to only safely operating with sysctls, it seems.

Ulexus: I'm not sure I understand what is meant by the "old behavior" here, though... it just looks dangerous.

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
Unix4ever: no it's actually KSPP ones, not defaults.

pkg/machinery/kernel/kernel.go
Unix4ever: I think I can rewrite that a bit to look more straightforward, but will need to update all of our sysctls definitions.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -89,7 +89,7 @@ func SetupLogger(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionF
 		// disable ratelimiting for kmsg, otherwise logs might be not visible.
 		// this should be set via kernel arg, but in case it's not set, try to force it.
 		if err = krnl.WriteParam(&kernel.Param{
-			Key:   "kernel.printk_devkmsg",
+			Key:   "proc.sys.kernel.printk_devkmsg",
```
frezbo, line 92: does this mean a machine patch like:

```yaml
- op: add
  path: /machine/sysctls
  value:
    fs.inotify.max_user_instances: "512"
    fs.inotify.max_user_watches: "524288"
```

needs to be updated as:

```yaml
- op: add
  path: /machine/sysctls
  value:
    proc.sys.fs.inotify.max_user_instances: "512"
    proc.sys.fs.inotify.max_user_watches: "524288"
```

smira: my understanding is that no, it doesn't need to be updated.

translation happens in a controller where the `sysctls` and `sysfs` are translated into proper `KernelParamSpec` resources

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
```diff
@@ -73,6 +74,12 @@ func (ctrl *KernelParamSpecController) Run(ctx context.Context, r controller.Run
 		case <-ctx.Done():
 			return nil
 		case <-r.EventCh():
+			ksppParams := map[string]struct{}{}
+
+			for _, param := range kspp.GetKernelParams() {
+				ksppParams[param.Key] = struct{}{}
```
smira, line 80: I would guess it's not the right key here? as KSPP params have sysctl names as keys?

pkg/machinery/kernel/kernel.go
smira: bumping this question, as I think the interface should be straightforward. we can keep old methods if we want to, but I would rather explicitly break backwards compatibility by e.g. renaming the fields in the `Param` structure

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
Unix4ever: yeah... I don't like the code still, trying to make it more reliable and straightforward.

pkg/machinery/kernel/kernel.go
Unix4ever: should be good now.

pkg/machinery/kernel/kernel.go
Unix4ever: I wonder if we actually have any other users of `machinery`. It looks like only `kernel.Write/Read/DeleteParam` is using that code. Only if that's some third party projects, but do you think it can be the case?

internal/app/machined/pkg/controllers/runtime/kernel_param_spec.go
Unix4ever: finally it looks ok.

Ulexus (approved):
> Reviewed 1 of 6 files at r2, 18 of 18 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4843)* status: all files reviewed, 7 unresolved discussions (waiting on @frezbo, @smira, @Ulexus, and @Unix4ever)

*[a discussion](https://reviewable.io/reviews/talos-systems/talos/4843#-Mx-TWc549QaEI7jo3FN:-Mx-TWc6D5_psMdzf3r4:b-vu2ym3) (no related file):*
I have to say that I don't love the idea of making the keys include the standard mounted path.  I understand that:
 - the key must be a string
 - just using the path makes the conversion easier
 - those paths haven't changed in all of Linux, and they are pretty unlikely to do so

But... they are surrogate values.  They represent an external standard, not an identifier or definition.  I'd rather see just "sysctl" and "sysfs" as the prefix instead of the path, which are still trivial to search-and-replace, but have the advantage of not including external information or tying us to "legacy" mount paths.

This is hardly a blocker, though, and the case could certainly be argued the other way.

---

smira:
```
applied configuration with a reboot: this config change can't be applied in immediate mode
diff:   &v1alpha1.Config{
  	ConfigVersion: "v1alpha1",
  	ConfigDebug:   false,
  	ConfigPersist: true,
  	MachineConfig: &v1alpha1.MachineConfig{
  		... // 12 identical fields
  		MachineTime:                 nil,
  		MachineSysctls:              {"kernel.kexec_load_disabled": "1"},
- 		MachineSysfs:                nil,
+ 		MachineSysfs:                map[string]string{"device.system.cpu.cpu0.cpufreq.scaling_governor": "performance"},
  		MachineRegistries:           {RegistryMirrors: {"127.0.0.1:5005": &{MirrorEndpoints: {"http://172.20.0.1:5005"}}, "docker.io": &{MirrorEndpoints: {"http://172.20.0.1:5000"}}, "gcr.io": &{MirrorEndpoints: {"http://172.20.0.1:5003"}}, "ghcr.io": &{MirrorEndpoints: {"http://172.20.0.1:5004"}}, ...}},
  		MachineSystemDiskEncryption: nil,
  		... // 4 identical fields
  	},
  	ClusterConfig: &{ClusterID: "LQxd0Yh18Gun0sWE6nv7JU5BVu8eWUjSUL7d4s9J9XA=", ClusterSecret: "AqpBWFoEKetHTKfaN6e/PWTgxH3GA6/wlNasL5iBpxw=", ControlPlane: &{Endpoint: &{URL: &{Scheme: "https", Host: "172.20.0.1:6443"}}}, ClusterName: "talos-default", ...},
  }
```

we need to mark `.machine.sysfs` as immediately appliable

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -239,6 +239,10 @@ var (
 		"net.ipv4.ip_forward": "0",
 	}
 
+	machineSysfsExample = map[string]string{
+		"device.system.cpu.cpu0.cpufreq.scaling_governor": "performance",
```
smira, line 243:
```suggestion
		"devices.system.cpu.cpu0.cpufreq.scaling_governor": "performance",
```

(wrong path)

Unix4ever:
> But... they are surrogate values. They represent an external standard, not an identifier or definition. I'd rather see just "sysctl" and "sysfs" as the prefix instead of the path, which are still trivial to search-and-replace, but have the advantage of not including external information or tying us to "legacy" mount paths.

I did that at first, but the code was way to messy...

smira (commented):
:+1: but I can't approve my own PR lol

PR #4847, opened 2022-01-23
Etc localtime symlink

### LINKED ISSUE #4840: provide `/etc/localtime` symlink
## Feature Request

Talos is always UTC timezone, but some apps rely on presence of the `/etc/localtime`. It should be a symlink to the directory `/usr/share/zoneinfo/Etc/UTC` with the binary file provided for UTC.

sauterp:
# Issue

Fixes #4840 

I tried to take a first step towards an implementation for #4840 
I'm trying to undestand how Talos is put together and this issue seemed beginner-friendly to me :)
If you could give some feedback I'd be happy to complete it.

## What? (description)

As I understand the way to do this would be to create a new resource type EtcSymlink(similar to EtcFile) which is then used by the EtcFileController to setup a symlink.

## Why? (reasoning)

This approach could be reused later to create other symlinks or it could even be exposed for users to configure what `/etc/localtime` links to.

## Questions

An aspect I am unsure about is, how to generate or include the `/usr/share/zoneinfo/Etc/UTC` file(RFC 8536). With `talosctl list` I verified that these files don't exist yet in talos 0.14. On my fedora machine `/usr/share/zoneinfo` contains files for many timezones. But I guess the minimalist approach here would be to only include the `/usr/share/zoneinfo/Etc/UTC` file in talos. What do you think?

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4847)


smira:
first of all, thanks for your PR!

for this particular case, as the symlink is essentially static and doesn't ever change, we can create it in the read-only rootfs of Talos

as you said, we can take binary `UTC` zoneinfo file, put into the proper location under `/usr/share/zoneinfo` and create a symlink

the easiest way is modifying `Dockerfile` around the following lines: https://github.com/talos-systems/talos/blob/58eb3600fc44dc2fccaa82322207291ffd807205/Dockerfile#L381-L391

Please note there's `arm64` section below which should be changed exactly same way.

We can put the `zoneinfo` under `hack/` directory in Talos in copy it in like we do for containerd configuration.

Dockerfile
```diff
@@ -379,6 +379,9 @@ COPY --from=machined-build-amd64 /machined /rootfs/sbin/init
 COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
 RUN cleanup.sh /rootfs
 RUN mkdir -pv /rootfs/{boot,etc/cri/conf.d/hosts,usr/local/share,mnt,system,opt}
+RUN mkdir -pv /rootfs/usr/share/zoneinfo/Etc
+COPY --chmod=0644 hack/zoneinfo/Etc/UTC /rootfs/usr/share/zoneinfo/Etc/UTC
+RUN ln -s /etc/localtime /rootfs/usr/share/zoneinfo/Etc/UTC
```
smira, line 384: I guess it should be: (not tested)

```suggestion
RUN ln -s /rootfs/etc/localtime /usr/share/zoneinfo/Etc/UTC
```

smira:
@sauterp I cleaned up the symlink a bit

smira:
```bash
$ talosctl -n 172.20.0.2 ls -l /etc/
NODE         MODE         UID   GID   SIZE(B)   LASTMOD           NAME
172.20.0.2   drwxr-xr-x   0     0     262       Feb  7 15:40:32   .
172.20.0.2   Lrwxrwxrwx   0     0     8         Feb  7 15:40:32   ca-certificates -> /etc/ssl
172.20.0.2   d---------   0     0     19        Feb  7 15:42:57   cni
172.20.0.2   drwxr-xr-x   0     0     34        Feb  7 15:40:32   containerd
172.20.0.2   drwxr-xr-x   0     0     52        Feb  7 15:40:32   cri
172.20.0.2   -rw-r--r--   0     0     0         Feb  7 15:40:32   extensions.yaml
172.20.0.2   -rw-r--r--   0     0     179       Feb  7 15:42:56   hosts
172.20.0.2   drwxr-xr-x   0     0     75        Feb  7 15:40:32   iscsi
172.20.0.2   drwxr-xr-x   0     0     72        Feb  7 15:40:32   isns
172.20.0.2   d---------   0     0     108       Feb  7 15:43:52   kubernetes
172.20.0.2   Lrwxrwxrwx   0     0     27        Feb  7 15:40:32   localtime -> /usr/share/zoneinfo/Etc/UTC
172.20.0.2   drwxr-xr-x   0     0     67        Feb  7 15:40:32   lvm
172.20.0.2   -r--r--r--   0     0     32        Feb  7 15:42:56   machine-id
172.20.0.2   -rw-r--r--   0     0     213       Feb  7 15:42:55   os-release
172.20.0.2   Lrwxrwxrwx   0     0     8         Feb  7 15:40:32   pki -> /etc/ssl
172.20.0.2   -rw-r--r--   0     0     38        Feb  7 15:42:56   resolv.conf
172.20.0.2   drwxr-xr-x   0     0     28        Feb  7 15:40:32   ssl
```

smira (approved):

PR #4881, opened 2022-01-26
fix: handle 404 errors from AWS IMDS correctly

smira:
Looks like despite the documentation, AWS IMDS library is using
different error wrapping on 404s.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4881)


smira:
verified manually that it fixes the maintenance mode on AWS

PR #4902, opened 2022-01-31
chore: update k8s upgrade message

frezbo:
Update k8s upgrade message

Before:

```bash
❯ talosctl upgrade-k8s --to 1.23.3
automatically detected the lowest Kubernetes version 1.23.2
checking for resource APIs to be deprecated in version 1.23.3
discovered master nodes ["192.168.15.20"]
discovered worker nodes []
updating "kube-apiserver" to version "1.23.3"
 > "192.168.15.20": starting update
 > update kube-apiserver: 1.23.2 -> 1.23.3
 > "192.168.15.20": machine configuration patched
 > "192.168.15.20": waiting for API server state pod update
 < "192.168.15.20": successfully updated
updating "kube-controller-manager" to version "1.23.3"
 > "192.168.15.20": starting update
 > update kube-controller-manager: 1.23.2 -> 1.23.3
 > "192.168.15.20": machine configuration patched
  > "192.168.15.20": waiting for API server state pod update
```

each component in the message used to display `waiting for API server state pod update` which is confusing when a k8s component other than kube apiserver is being updated.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4902)

### DIFF
--- a/pkg/cluster/kubernetes/talos_managed.go
+++ b/pkg/cluster/kubernetes/talos_managed.go
@@ -200,7 +200,7 @@ func upgradeStaticPodOnNode(ctx context.Context, cluster UpgradeProvider, option
 	}
 
 	options.Log(" > %q: machine configuration patched", node)
-	options.Log(" > %q: waiting for API server state pod update", node)
+	options.Log(" > %q: waiting for %s pod update", node, service)
 
 	var expectedConfigVersion string
 


smira:
the intention was to say that we wait for the pod state to be reflected in the API server state; as it's a static pod, kubelet eventually pushes the state to the API server, but I agree that the message is confusing

PR #4904, opened 2022-01-31
feat: extract firmware part of system extensions into initramfs

### LINKED ISSUE #4816: system extensions: special handling of `/lib/firmware`
## Feature Request

Contents of `/lib/firmware` should be present during initial boot within `initramfs` (when the kernel attempts to load the firmware) and in the rootfs (when e.g. kernel module is loaded).

We should install system extensions containing `/lib/firmware` into `initramfs` skipping `sqshfs` stage.

Talos `init` should skip cleaning up `/lib/firmware` and it should bind mount read-only (would it work?) into final `rootfs`, so that `/lib/firmware` is available and identical in both `initramfs` and Talos `rootfs`.

### Description

Remove `bnc2x` from the rootfs, it most probably doesn't work.

smira:
Fixes #4816

This changes the way system extensions are packaged into the squashfs
images: `/lib/firmware` is now moved out of the future squashfs images
and becomes part of `initramfs` to make firmware available in the early
boot.

Talos will bind-mount `/lib/firwmare` into rootfs as well, so it will be
available in the rootfs as well.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4904)

### DIFF
--- a/internal/pkg/extensions/pull.go
+++ b/internal/pkg/extensions/pull.go
@@ -77,7 +77,7 @@ func (puller *Puller) PullAndMount(ctx context.Context, registryConfig config.Re
 			return err
 		}
 
-		mounts, err := snapshotService.View(ctx, path, chainID.String())
+		mounts, err := snapshotService.Prepare(ctx, path, chainID.String())
 		if err != nil {
 			return err
 		}
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -360,7 +360,7 @@ local integration_provision_tests_track_2 = Step("provision-tests-track-2", priv
 
 local integration_gvisor = Step("e2e-gvisor", target="e2e-qemu", privileged=true, depends_on=[load_artifacts], environment={
         "SHORT_INTEGRATION_TEST": "yes",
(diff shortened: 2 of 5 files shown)


internal/pkg/extensions/pull.go
```diff
@@ -77,7 +77,7 @@ func (puller *Puller) PullAndMount(ctx context.Context, registryConfig config.Re
 			return err
 		}
 
-		mounts, err := snapshotService.View(ctx, path, chainID.String())
+		mounts, err := snapshotService.Prepare(ctx, path, chainID.String())
```
smira, line 80: as installer now modifies system extension by moving out `/lib/firmware`, we need to make snapshot read-write

Ulexus (approved):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4904)* status: all files reviewed, 1 unresolved discussion (waiting on @Ulexus)

Ulexus (approved):
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/4904)* status: all files reviewed, 1 unresolved discussion (waiting on @smira)

PR #4908, opened 2022-02-01
feat: allow link scope routes in the machine config

### LINKED ISSUE #4897: Add to route link local hosts
## Feature Request

Case: I have 2 network interfaces on the node. One interface for meta-server, second for local private network with NAT.
Cloud provider sends to me default gateway on first interface. But I want to use second as default network.

```yaml
      - interface: eth0
        dhcp: true
      - interface: eth1
        dhcp: true
        dhcpOptions:
          routeMetric: 512
```

I can change the route metric on the second interface, but in this case I cannot reach the meta-server anymore.

## Proposal

Set link local route tables

```yaml
      - interface: eth0
        dhcp: true
        dhcpOptions:
          routeMetric: 2048
        routes:
          - network: 169.254.169.254/32
            metric: 1024
      - interface: eth1
        dhcp: true
```

I did not set the gateway in the ```routes``` path. In this case Talos sets set the route like:

```shell
ip ro add 169.254.169.254 scope link dev eth0
```

smira:
Fixes #4897

They were supported internally, but never properly exposed in the
machine configuration.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4908)


smira:
cc @sergelogvinov

frezbo (approved):

PR #4993, opened 2022-02-16
feat: add low source port capabilities

sergelogvinov:
This feature allow to us use low source port <1024 to make a http calls.

# Pull Request

## Acceptance

- [x] cloud integration tests

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4993)

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -9,18 +9,23 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
+	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"time"
 
+	"github.com/hashicorp/go-cleanhttp"
 	"github.com/talos-systems/go-retry/retry"
 )
 
 const b64 = "base64"
 
 type downloadOptions struct {
-	Headers map[string]string
-	Format  string
+	Headers    map[string]string
+	Format     string
+	LowSrcPort bool
 
 	ErrorOnNotFound      error
 	ErrorOnEmptyResponse error
@@ -54,6 +59,14 @@ func WithFormat(format string) Option {
 func WithHeaders(headers map[string]string) Option {
 	return func(d *downloadOptions) {
 		d.Headers = headers
+	}
+}
+
+// WithLowSrcPort sets low source port to download
+// the config.
+func WithLowSrcPort() Option {
+	return func(d *downloadOptions) {
+		d.LowSrcPort = true
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Ulexus:
So... why?

Ulexus:
I the idea is that only privileged clients can retrieve user-data on Scaleway.

sergelogvinov:
Usually network is not ready then Talos trying to reach meta-server. 
instance.NewMetadataAPI() - it can be failed sometimes.

I think "low source port" feature can be useful with another cloud platform.

Ulexus:
Sorry, how does this help with a network-not-ready condition?

sergelogvinov:
> Sorry, how does this help with a network-not-ready condition?

download package does it for us.

pkg/download/download.go
```diff
@@ -130,6 +142,24 @@ func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, e
 func download(req *http.Request, dlOpts *downloadOptions) (data []byte, err error) {
 	client := &http.Client{}
 
+	if dlOpts.LowSrcPort {
+		port := 100 + rand.Intn(512)
+
+		localTCPAddr, tcperr := net.ResolveTCPAddr("tcp", ":"+strconv.Itoa(port))
```
smira, line 148: I don't see though how this would handle already used ports?

pkg/download/download.go
```diff
@@ -130,6 +142,24 @@ func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, e
...
+		if tcperr != nil {
+			return nil, retry.ExpectedError(fmt.Errorf("resolving source tcp address: %s", tcperr.Error()))
+		}
+
+		client = &http.Client{
+			Transport: &http.Transport{
```
smira, line 154: the worst thing which I hate a lot in Go, is that actually `nil` `Transport` is not an empty transport - there's a lot of good defaults set explicitly. we need to preserve them somehow

pkg/download/download.go
```diff
@@ -130,6 +142,24 @@ func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, e
...
 
+	if dlOpts.LowSrcPort {
+		port := 100 + rand.Intn(512)
+
+		localTCPAddr, tcperr := net.ResolveTCPAddr("tcp", ":"+strconv.Itoa(port))
+		if tcperr != nil {
+			return nil, retry.ExpectedError(fmt.Errorf("resolving source tcp address: %s", tcperr.Error()))
+		}
+
+		client = &http.Client{
```
smira, line 153: probably makes more sense to override the `client.Transport =` vs.  overriding whole thing?

pkg/download/download.go
smira: we should actually be using https://github.com/hashicorp/go-cleanhttp here to build "clean" client and tweak things after that

we could handle port reuse in the dialer

pkg/download/download.go
sergelogvinov: No. I do not. If port is busy it returns error, and Download() retry the request.

pkg/download/download.go
```diff
@@ -130,6 +143,34 @@ func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, e
...
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			ForceAttemptHTTP2:     true,
+			DisableKeepAlives:     true,
+			MaxIdleConnsPerHost:   -1,
+		}
```
sergelogvinov, line 171: I've found similar here:

internal/pkg/containers/image/resolver.go:162
internal/app/machined/main.go:46

smira: yep, exactly, but you can see it sets a lot of weird defaults

smira: so the idea is to get `cleanhttp.DefaultTransport()` and modify it accordingly. It will require some careful typecasting though (unfortunately)

smira (approved):

PR #4999, opened 2022-02-16
fix: prefer logical on merging link specs

smira:
This solves a case when lower layer (platform) defines `bond0` as
logical interface properly, and upper layer (configuration) defines only
some part of the config (e.g. VIP).

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/4999)

### DIFF
--- a/pkg/machinery/resources/network/link_spec.go
+++ b/pkg/machinery/resources/network/link_spec.go
@@ -66,8 +66,9 @@ var (
 //
 //nolint:gocyclo
 func (spec *LinkSpecSpec) Merge(other *LinkSpecSpec) error {
-	if spec.Logical != other.Logical {
-		return fmt.Errorf("mismatch on logical for %q: %v != %v", spec.Name, spec.Logical, other.Logical)
+	// prefer Logical, as it is defined for bonds/vlans, etc.
+	if other.Logical {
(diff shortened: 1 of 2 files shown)


pkg/machinery/resources/network/link_spec.go
```diff
@@ -66,8 +66,9 @@ var (
 //
 //nolint:gocyclo
 func (spec *LinkSpecSpec) Merge(other *LinkSpecSpec) error {
-	if spec.Logical != other.Logical {
-		return fmt.Errorf("mismatch on logical for %q: %v != %v", spec.Name, spec.Logical, other.Logical)
+	// prefer Logical, as it is defined for bonds/vlans, etc.
```
frezbo, line 69: is it possible to have tests for this?

smira: yes, I can add tests, even though the code here is relatively trivial, we probably can do some higher level in the actual controller which merges the specs

frezbo (approved):

pkg/machinery/resources/network/link_spec.go
smira: added the test, thanks for the review!

frezbo (approved):

PR #5005, opened 2022-02-16
feat: store audit logs to disk

### LINKED ISSUE #5000: kube-apiserver audit log changes
## Feature Request

There are few recommended settings for the kube-apiserver's audit log settings which we
should consider:
 - `audit-log-maxsize = 100` (currently `50`)
 - `audit-log-maxbackup = 10` (currently `3`)
 - `audit-log-maxage = 10` (number of days' worth of log files to keep before deleting)

Additionally, a common requirement for audit logs is that they be stored to
file.  Presently, they are sent to stdout, which works well for tying into an
external logging system, but does not meet all logging requirements.
Ideally, we should be storing the audit logs to non-volatile local storage in a
way which:
 - does not lend itself to mounting from arbitrary Kubernetes workloads
 - does not present a risk of denial of service due to disk overflow

`audit-log-path = <something>` (pointing to a real directory; currently, this is `stdout`)

I view this latter as being a much less universal option which should probably _not_ be enabled by default.  Perhaps, in fact, this is best left up to a machine config patch to the apiserver.

### Description

These items address STIG vulnerabilities:
 - v-242465
 - v-242462

Ulexus:
Instead of bundling the apiserver audit logs with the rest of the
apiserver logs, we should store them separately to file, assuring
reasonable defaults for retention and rotation.

Fixes #5000

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5005)

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -38,6 +38,14 @@ Of particular note is that Talos now sets the `kernel.panic` reboot interval to 
 If your kubelet fails to start after the upgrade, please check the `kubelet` logs to determine the problem.
 """
 
+    [notes.auditlog]
+        title = "API Server audit logs"
+        description="""\
+`kube-apiserver` is now configured to store its audit logs separately from the `kube-apiserver` standard logs and directly to file.
+The `kube-apiserver` will maintain the rotation and retirement of these logs, which are stored in `/var/log/audit/`.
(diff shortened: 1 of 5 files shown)


smira (approved):
lgtm, but I think we should mention in the `hack/release.toml` that the audit log now is in a different location, and change is applied on upgrade

JAORMX (approved):

frezbo (approved):

PR #5013, opened 2022-02-18
gridsome version updated & algolia search added

mohsin309:
# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5013)


website/src/main.js
```diff
@@ -39,4 +39,25 @@ export default function(Vue, { router, head, isClient, appOptions }) {
...
+    type: 'text/javascript',
+    body: true,
+    src: 'https://cdn.jsdelivr.net/npm/@algolia/algoliasearch-netlify-frontend@1/dist/algoliasearchNetlify.js'
+  })
+  head.script.push({
+    type: 'text/javascript',
+    body: true,
+    innerHTML: `algoliasearchNetlify({
+    appId: 'W8IEQZJ792',
+    apiKey: 'b16fa302b5c8c43816d11ecc7da702d2',
```
andrewrynhard, line 57: Should this be secret?

mohsin309: No, we don't need to make it secret because it is only against the specific app it can't be used in other apps Thanks!

andrewrynhard: Not sure how they would know how to distinguish one app from another if all they have is the API key.

mohsin309: Okay this apId and secret are generated against a specific app like in our case 'Talos' from the Algolia Search dashboard so once these keys are added then Algolia dashboard detects the App but if you say I can add this in env so where you place the key share me the process.env.key_name I will add Thanks

andrewrynhard: Right. But the API key _is_ the identifier I imagine. Putting it in the env is a good idea. We will need to generate the key again since it was exposed.

mohsin309: Sure can you share the file name where i need to place appId & apiKey so after that i will access in code with its key name Thanks

mohsin309: E.g process.env.GA_ID or process.env.SITE_URL

andrewrynhard: I can set an env var in netlify. If you can code it to use process.env I can the rest up.

mohsin309: Okay here are the keys with their values you need to set in .env 

1- GRIDSOME_ALGOLIA_APP_ID  = W8IEQZJ792

2- GRIDSOME_ALGOLIA_API_KEY  = b16fa302b5c8c43816d11ecc7da702d2

3- GRIDSOME_ALGOLIA_SITE_ID  =  4ce8d530-06a7-45d7-bc2a-0484644882df

so I will access like as process.env.GRIDSOME_ALGOLIA_APP_ID e.t.c

Let me know then Thanks!

mohsin309: I have commented the code with secret keys for Algolia Search kindly review Thanks!

andrewrynhard (approved):

andrewrynhard (approved):

mohsin309:
Hi sir will you merge these changes to Master because master is still not updated yet i need to run the build Thanks waiting

website/src/main.js
```diff
@@ -5,38 +5,62 @@ import "~/assets/css/main.css";
...
+        next();
     });
-    next();
-  });
 
-  head.script.push({
-    src: "/js/asciinema-player.js",
-    body: true,
-  });
+    head.script.push({
```
Unix4ever, line 38: Is it possible to import the module using npm and embed it into the page during the static build to avoid loading it dynamically on each page redraw?

andrewrynhard: Good catch.

mohsin309: image

Here is the above code which I have written which cannot be used as npm and code is already approved kindly merge this with master Thanks

andrewrynhard: @mohsin309 Your code is fine ... we want the JS to be handled by webpack so that each page load won't have to pull resources.

andrewrynhard (changes requested):
We want the JS in webpack.

website/src/main.js
andrewrynhard: @Unix4ever found https://www.npmjs.com/package/gridsome-plugin-algolia#usage

Unix4ever: I would say that the main problem with your approach is that the dependencies get scattered around the code base.
If we rely on webpack to get dependencies then it is better to consistently use it.
It will be easier to update libraries that way as well.

mohsin309:
Hi Algolia added to plugins kindly check that and merge with master thanks 

And kindly update env with  below keys 

1- ALGOLIA_APP_ID = W8IEQZJ792

2- ALGOLIA_API_KEY = b16fa302b5c8c43816d11ecc7da702d2

3- ALGOLIA_SITE_ID = 4ce8d530-06a7-45d7-bc2a-0484644882df

so I will access like as process.env.ALGOLIA_SITE_ID e.t.c

Let me know then Thanks!

website/src/main.js
Unix4ever: Shouldn't everything here and below be removed as you now load this module using webpack?

mohsin309: I don't understand can you elaborate please

Unix4ever: If you have installed the module through webpack, why do you still need to load it manually here?
I would expect it to work without `head.link.push`, `head.script.push` and so on.

mohsin309: Kindly check this now Thanks

Unix4ever (approved):

website/src/main.js
Unix4ever: yeah... Looks correct now. Thank you.

mohsin309: Thanks to you really learning from you I wish to be a part of your team it will be a great chance for me to learn things in depth Thanks

Ulexus (approved):

andrewrynhard (changes requested):
Looks like `gridsome build` is broken.

mohsin309:
image

Hi, @andrewrynhard @Unix4ever  can you please help me I have added the alogia search in plugins in gridsome.config.js map of undefined error getting on gridsome build I have tried but have not found any solution yet. Thanks

andrewrynhard:
Looks like this is likely caused by upgrading gridsome:

```
5:55:58 PM: (node:1552) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
5:55:58 PM: (node:1559) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'disconnect' of null
5:55:58 PM:     at a.$route (assets/js/app.42d580ad.js:19761:21)
5:55:58 PM:     at un.run (/opt/build/repo/website/node_modules/vue/dist/vue.runtime.common.prod.js:6:27543)
5:55:58 PM:     at sn (/opt/build/repo/website/node_modules/vue/dist/vue.runtime.common.prod.js:6:25554)
5:55:58 PM:     at Array.<anonymous> (/opt/build/repo/website/node_modules/vue/dist/vue.runtime.common.prod.js:6:12260)
5:55:58 PM:     at Wt (/opt/build/repo/website/node_modules/vue/dist/vue.runtime.common.prod.js:6:11661)
5:55:58 PM: (node:1559) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
```

^ This is a warning in our current version.

steverfrancis:
We're abandoning this approach for a new hugo approach with a new dev. :-)
@andrew you can close/backout whatever one does with rejected PRs

PR #5022, opened 2022-02-20
docs: update Hetzner instructions [on hold]

choffmeister:
The old instructions using a snapshot to start from are deprecated. As confirmed by one of the contributors at Hetzner (see [here](https://github.com/hetznercloud/hcloud-go/issues/192#issuecomment-1046206839)) servers created from snapshots will not be able to get information from the new metadata endpoints. The next version of Hetzners container storage interface will rely on these endpoints and hence will not work anymore if servers are created from snapshots.

I have provided a quick script that bootstraps a Hetzner server with Talos installed and Talos config applied without relying on the use of snapshots.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5022)

### DIFF
--- a/website/content/docs/v0.15/Cloud Platforms/hetzner.md
+++ b/website/content/docs/v0.15/Cloud Platforms/hetzner.md
@@ -13,117 +13,32 @@ There are two options to upload your own.
 1. Run an instance in rescue mode and replace the system OS with the Talos image
 2. Use [Hashicorp packer](https://www.packer.io/docs/builders/hetzner-cloud) to prepare an image
 
-### Rescue mode
+### CLI
 
-Create a new Server in the Hetzner console.
-Enable the Hetzner Rescue System for this server and reboot.
-Upon a reboot, the server will boot a special minimal Linux distribution designed for repair and reinstall.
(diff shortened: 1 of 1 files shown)


sergelogvinov:
This is very sad news. The bootstrap time will increased.  And can be failed...

sergelogvinov:
I think, better to add it as third solution.

and use ```kexec``` to launch the Talos. Here ansible example https://github.com/sergelogvinov/ansible-collection-talos/blob/main/roles/talos-boot/tasks/main.yml#L41

choffmeister:
Maybe this PR should be put on hold until it finally resolved, if this really will be the case. So far the change mentioned is not yet released by Hetzner. Still hoping that they will not apply this IMHO not very senseful decision, to just block out old/snapshotted servers from fundamental cloud metadata endpoints.

sanmai-NL:
@choffmeister Can you mark this as a 'Draft'? That would allow for better filtering over 'on hold' in the title. Thanks!

sergelogvinov:
I think this PR can be closed now. Hetznel unlocked metadata for all images a few months ago...

PR #5038, opened 2022-02-22
feat: support admission control configuration and Pod Security admission

### LINKED ISSUE #5003: enable PSS by default, enforce baseline policy
## Feature Request

See https://kubernetes.io/docs/concepts/security/pod-security-standards/

Drop PSP as we have it today, enforce baseline policy by default, but keep `kube-system` as unrestricted. 

Do a scream test and see what breaks.

smira:
Fixes #5003

This implements a way to configure API server admission plugins via
Talos machine configuration.

If Pod Security admission is enabled, default cluster-wide policy is
generated which enforces baseline policy.

Policy can be overridden per-namespace.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5038)


Ulexus (approved):
> Reviewed 24 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/5038)* status: all files reviewed, 1 unresolved discussion (waiting on @smira)

*[a discussion](https://reviewable.io/reviews/talos-systems/talos/5038#-MwYXBfv-Sa8COQH3D0M:-MwYXBfv-Sa8COQH3D0N:b-pgnl6h) (no related file):*
The defaults appear to be installed by the config generation, which seems like the best pathway... but could we have a migration/upgrade section of the doc, as well?

---

frezbo (approved):

smira:
> The defaults appear to be installed by the config generation, which seems like the best pathway... but could we have a migration/upgrade section of the doc, as well?

Right now Pod Security is not enabled yet by default, the plan is to target that for the next version of Talos. I will add a link to What's New section once we prepare the release

PR #5045, opened 2022-02-24
feat: add a way to override kubelet configuration via machine config

### LINKED ISSUE #4629: Kubelet configuration available via machine configuration
Kubelet command line flags are getting deprecated, and new way to use the config file.

Talos today accepts only `extraArgs`, so we need a better way to fine-tune kubelet.

We don't want to import whole `KubeletConfig` as a Go structure into the machine configuration as this will cause unnecessary dependency. We would rather present `KubeletConfig` as a `map[string]interface{}`, we skip any validation on Talos side, and simply unmarshal raw data into the `KubeletConfig` before starting the `kubelet`. Kubelet will validate its config and error out if it's invalid.

smira:
Fixes #4629

Note: some fields are enforced by Talos and are not overridable.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5045)

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
@@ -14,15 +14,16 @@ import (
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/hashicorp/go-multierror"
 	"go.uber.org/zap"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/component-base/config/v1alpha1"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 
 	v1alpha1runtime "github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	"github.com/talos-systems/talos/pkg/argsbuilder"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
+	"github.com/talos-systems/talos/pkg/machinery/kubelet"
 	"github.com/talos-systems/talos/pkg/machinery/resources/k8s"
 )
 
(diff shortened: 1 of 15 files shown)


JAORMX (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -191,49 +195,80 @@ func (ctrl *KubeletSpecController) Run(ctx context.Context, r controller.Runtime
...
+	for field, message := range map[string]string{
+		"staticPodPath":  "staticPodPath can't be overridden",
+		"authentication": "authentication can't be overridden",
+		"authorization":  "authorization can't be overridden",
+		"cgroupRoot":     "cgroup configuration can't be overridden",
+		"systemCgroups":  "cgroup configuration can't be overridden",
+		"kubeletCgroups": "cgroup configuration can't be overridden",
+		"port":           "port can't be overridden",
+		"apiVersion":     "apiVersion can't be overridden",
+		"kind":           "kind can't be overridden",
```
JAORMX, line 212: Should these be documented?

JAORMX (changes requested):

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
smira: yes, they should.

this is a work in progress PR, it's not finished yet

internal/app/machined/pkg/controllers/k8s/kubelet_spec_test.go
```diff
@@ -184,3 +241,102 @@ func (suite *KubeletSpecSuite) TearDownTest() {
...
+				"Port": "xyz",
+			},
+			expectedErr: "error unmarshalling extra kubelet configuration: strict decoding error: unknown field \"API\", unknown field \"Port\", unknown field \"foo\"",
+		},
+		{
+			name: "wrong field type",
+			extraConfig: map[string]interface{}{
+				"oomScoreAdj": "v1",
+			},
+			expectedErr: "error unmarshalling extra kubelet configuration: unrecognized type: int32",
```
frezbo, line 265: shouldn't it say `string`  instead?

smira: it's not us in this case, it's the way Kubernetes converter unmarshaller the error, and we have no control over it

frezbo: okay, it was confusing though :sweat_smile:

Ulexus (approved):
> Reviewed 6 of 11 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/talos-systems/talos/5045)* status: all files reviewed, 2 unresolved discussions (waiting on @JAORMX and @smira)

frezbo (approved):

PR #5046, opened 2022-02-25
fix: set route to metaserver

sergelogvinov:
Set default route to meta-server, which exists only on eth0 interface.

This changes affect dual network setup. 
In dual network solution you have to set 169.254.42.42/32 route by your self.
After that you will lost ipv6 auto configuration.

So we can set up route 169.254.42.42/32 automatically.

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5046)


smira (approved):

frezbo (approved):
Does this need any doc changes?

sergelogvinov:
> Does this need any doc changes?

I do not think so. This route helps to reach the meta server. It will work with one/dual network interface setup.
Any ideas/thoughts?

PR #5060, opened 2022-03-01
fix: align list and diskusage command flags with their Linux analogs

Unix4ever:
Fixes: https://github.com/talos-systems/talos/issues/3018

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5060)

### DIFF
--- a/cmd/talosctl/cmd/talos/diskusage.go
+++ b/cmd/talosctl/cmd/talos/diskusage.go
@@ -41,7 +41,7 @@ var duCmd = &cobra.Command{
 			}
 
 			stream, err := c.DiskUsage(ctx, &machineapi.DiskUsageRequest{
-				RecursionDepth: recursionDepth,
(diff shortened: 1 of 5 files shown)


cmd/talosctl/cmd/talos/diskusage.go
```diff
@@ -41,7 +41,7 @@ var duCmd = &cobra.Command{
 			}
 
 			stream, err := c.DiskUsage(ctx, &machineapi.DiskUsageRequest{
-				RecursionDepth: recursionDepth,
+				RecursionDepth: recursionDepth + 1,
```
smira, line 44: this will cause some fun inconsistencies if `talosctl` and Talos versions dont' match, but I think it's better to fix things now rather than keep them broken forever.

smira (approved):

PR #5065, opened 2022-03-02
chore: re-enable kexec and default to UEFI booting in tests

### LINKED ISSUE #4947: kexec fails randomly with initramfs corruption
## Feature Request

Depending on `xz` parameters, `kexec` with initramfs either fails all the time, or fails randomly.

The issue manifests itself by the kernel printing error about `initramfs` decompression failure, which might be followed either by the failure to mount squashfs completely (`loop0: invalid argument`) or later stage squashfs corruption.

Same `initramfs` when booted via GRUB or PXE works just fine, so the problem is not with the `initramfs` image, but somewhere around the `kexec_file_load` interface.

### Description

Potential fixes:

* find the root cause and fix it
* try switching to `zstd` compression (unlikely to help from earlier testing)
* try using `kexec-tools` (capabilities ???)

See also #4941 #4940

smira:
Fixes #4947

It turns out there's something related to boot process in BIOS mode
which leads to initramfs corruption on later `kexec`.

Booting via GRUB is always successful.

Problem with kexec was confirmed with:

* direct boot via QEMU
* QEMU boot via iPXE (bundled with QEMU)

The root cause is not known, but the only visible difference is the
placement of RAMDISK with UEFI and BIOS boots:

```
[    0.005508] RAMDISK: [mem 0x312dd000-0x34965fff]
```

or:

```
[    0.003821] RAMDISK: [mem 0x711aa000-0x747a7fff]
```

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -370,9 +370,9 @@ local integration_cilium = Step("e2e-cilium-1.9.10", target="e2e-qemu", privileg
         "WITH_CONFIG_PATCH": '[{"op": "replace", "path": "/cluster/network/podSubnets", "value": ["10.0.0.0/8"]}]', # use Pod CIDRs as hardcoded in Cilium's quick-install
         "IMAGE_REGISTRY": local_registry,
 });
-local integration_uefi = Step("e2e-uefi", target="e2e-qemu", privileged=true, depends_on=[integration_cilium], environment={
+local integration_uefi = Step("e2e-bios", target="e2e-qemu", privileged=true, depends_on=[integration_cilium], environment={
         "SHORT_INTEGRATION_TEST": "yes",
-        "WITH_UEFI": "true",
+        "WITH_UEFI": "false",
         "IMAGE_REGISTRY": local_registry,
 });
 local integration_disk_image = Step("e2e-disk-image", target="e2e-qemu", privileged=true, depends_on=[integration_uefi], environment={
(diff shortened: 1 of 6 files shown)


.drone.jsonnet
```diff
@@ -370,9 +370,9 @@ local integration_cilium = Step("e2e-cilium-1.9.10", target="e2e-qemu", privileg
         "WITH_CONFIG_PATCH": '[{"op": "replace", "path": "/cluster/network/podSubnets", "value": ["10.0.0.0/8"]}]', # use Pod CIDRs as hardcoded in Cilium's quick-install
         "IMAGE_REGISTRY": local_registry,
 });
-local integration_uefi = Step("e2e-uefi", target="e2e-qemu", privileged=true, depends_on=[integration_cilium], environment={
+local integration_uefi = Step("e2e-bios", target="e2e-qemu", privileged=true, depends_on=[integration_cilium], environment={
```
smira, line 373: todo: need to rename variable

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -812,7 +812,7 @@ func init() {
 	createCmd.Flags().StringVar(&nodeDiskImagePath, "disk-image-path", "", "disk image to use")
 	createCmd.Flags().BoolVar(&applyConfigEnabled, "with-apply-config", false, "enable apply config when the VM is starting in maintenance mode")
 	createCmd.Flags().BoolVar(&bootloaderEnabled, "with-bootloader", true, "enable bootloader to load kernel and initramfs from disk image after install")
-	createCmd.Flags().BoolVar(&uefiEnabled, "with-uefi", false, "enable UEFI on x86_64 architecture (always enabled for arm64)")
+	createCmd.Flags().BoolVar(&uefiEnabled, "with-uefi", true, "enable UEFI on x86_64 architecture (always enabled for arm64)")
```
flokli, line 815:
```suggestion
	createCmd.Flags().BoolVar(&uefiEnabled, "with-uefi", true, "enable UEFI on x86_64 architecture")
```

frezbo (approved):

PR #5097, opened 2022-03-09
feat: bump boot partition size to 1000 MiB

### LINKED ISSUE #5096: bump /boot partition size

smira:
With system extensions, size of the `initramfs` might increase
significantly. With 1000 MiB `/boot`, as we store `A` and `B` boot
directories, we have 500 MiB for each Talos boot (size of the kernel and
initramfs).

Fixes #5096

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5097)

### DIFF
--- a/cmd/installer/pkg/raw.go
+++ b/cmd/installer/pkg/raw.go
@@ -12,7 +12,7 @@ import (
 
 const (
 	// RAWDiskSize is the minimum size disk we can create.
-	RAWDiskSize = 546
+	RAWDiskSize = 1246
 )
 
 // CreateRawDisk creates a raw disk by invoking the `dd` command.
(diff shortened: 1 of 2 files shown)


frezbo (approved):

cmd/installer/pkg/raw.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const (
 	// RAWDiskSize is the minimum size disk we can create.
-	RAWDiskSize = 546
+	RAWDiskSize = 1246
```
smira, line 15: also proportionally bumped the disk size of the images

JAORMX (approved):
Makes sense

PR #5129, opened 2022-03-14
feat: pause the boot process on some failures instead of rebooting

Unix4ever:
Some failures can be fixed by updating the machine configuration.
Now `userDisks` and `userFiles` do not make Talos to enter into reboot
loop but pause for 35 minutes.

Additionally, `apid` and `machined` are now started right after
containerd is up and running.

That makes it possible for the operator to connect to the node using
talosctl and fix the config.

Fixes: https://github.com/talos-systems/talos/issues/4669

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5129)

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -343,6 +343,10 @@ func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*ma
 func (s *Server) Bootstrap(ctx context.Context, in *machine.BootstrapRequest) (reply *machine.BootstrapResponse, err error) {
 	log.Printf("bootstrap request received")
 
+	if !s.Controller.Runtime().IsBootstrapAllowed() {
+		return nil, status.Error(codes.FailedPrecondition, "bootstrap is not available yet")
+	}
+
 	if s.Controller.Runtime().Config().Machine().Type() == machinetype.TypeWorker {
 		return nil, status.Error(codes.FailedPrecondition, "bootstrap can only be performed on a control plane node")
 	}
@@ -1891,7 +1895,16 @@ func (s *Server) EtcdSnapshot(in *machine.EtcdSnapshotRequest, srv machine.Machi
 }
 
 // EtcdRecover implements the machine.MachineServer interface.
+//nolint:gocyclo
 func (s *Server) EtcdRecover(srv machine.MachineService_EtcdRecoverServer) error {
+	if _, err := os.Stat(filepath.Dir(constants.EtcdRecoverySnapshotPath)); err != nil {
+		if os.IsNotExist(err) {
+			return status.Error(codes.FailedPrecondition, "etcd service is not ready for recovery yet")
+		}
+
+		return err
+	}
+
 	if err := s.checkControlplane("etcd recover"); err != nil {
 		return err
 	}
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -252,7 +252,10 @@ func run() error {
 	}
 
 	// Start the machine API.
-	system.Services(c.Runtime()).LoadAndStart(&services.Machined{Controller: c})
+	system.Services(c.Runtime()).LoadAndStart(
+		&services.Machined{Controller: c},
+		&services.APID{},
+	)
 
 	// Boot the machine.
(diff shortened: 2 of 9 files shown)


internal/app/machined/main.go
```diff
@@ -241,6 +242,23 @@ func run() error {
 		log.Printf("controller runtime finished")
 	}()
 
+	var eg errgroup.Group
```
JAORMX, line 245: should it be `eg.WithContext` [1]? That way context is cancelled and propagated appropriately too

[1] https://pkg.go.dev/golang.org/x/sync/errgroup

Unix4ever: yeah, that's a good idea, thanks.

JAORMX (approved):

internal/app/machined/main.go
```diff
@@ -241,6 +242,25 @@ func run() error {
 		log.Printf("controller runtime finished")
 	}()
 
+	var eg *errgroup.Group
+
+	eg, ctx = errgroup.WithContext(ctx)
+
+	eg.Go(func() error {
+		svc := &services.Containerd{}
+		if err = system.WaitForService(system.StateEventUp, svc.ID(c.Runtime())).Wait(ctx); err != nil {
```
smira, line 251: I believe `apid` service has a dependency on the `containerd` service already

Also I feel like we can start both after the `initialize` sequence, starting before `initialize` makes little sense to me. 

I also believe we don't need `errgroup` here

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1848,3 +1847,26 @@ func StopDBus(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc
...
+	timeout time.Duration) func(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
+	return func(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
+		f, name := callback(seq, data)
+
+		return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
+			err := f(ctx, logger, r)
+			if err != nil {
+				logger.Printf("%s failed, rebooting in %.0f minutes. You can use talosctl apply-config or talosctl edit mc to fix the issues, error:\n%s", name, timeout.Minutes(), err)
+
+				timer := time.NewTimer(time.Minute * 5)
```
smira, line 1861: nit: `defer timer.Stop()`

pkg/machinery/constants/constants.go
```diff
@@ -515,6 +515,9 @@ const (
 	// BootTimeout is the timeout to run all services.
 	BootTimeout = 35 * time.Minute
 
+	// FailurePauseTimeout is the timeout for the sequencer failures which can be fixed by updating the machine config.
```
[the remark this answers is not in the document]
Unix4ever: yeah... Makes sense. Bumped it.

smira:
I can still failures in the integration tests: `apid` goes up during the `install` phase, and our bootstrap code assumes node is up, and fails to bootstrap properly, as the node goes into a reboot loop. We either need to disable apid during the install phase (?) or we should handle such case

Unix4ever:
yeah... I wonder if it's easier to disable it during the install phase.

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1891,7 +1895,12 @@ func (s *Server) EtcdSnapshot(in *machine.EtcdSnapshotRequest, srv machine.Machi
 }
 
 // EtcdRecover implements the machine.MachineServer interface.
+//nolint:gocyclo
 func (s *Server) EtcdRecover(srv machine.MachineService_EtcdRecoverServer) error {
+	if _, err := os.Stat("/var/lib"); err != nil {
```
smira, line 1900:
```suggestion
	if _, err := os.Stat(filepath.Dir(constants.EtcdRecoverySnapshotPath)); err != nil {
```

internal/app/machined/main.go
```diff
@@ -251,8 +251,13 @@ func run() error {
 		return err
 	}
 
-	// Start the machine API.
-	system.Services(c.Runtime()).LoadAndStart(&services.Machined{Controller: c})
+	go func() {
```
smira, line 254: we don't need a goroutine here, `Start` is non-blocking, it just schedules the service to be loaded

smira (approved):

PR #5136, opened 2022-03-15
feat: relax extensions file structure validation

smira:
* allow empty directories (I see no harm in having them)
* allow symlinks 

See also https://github.com/talos-systems/extensions/pull/20

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5136)


internal/pkg/extensions/validate.go
```diff
@@ -72,7 +72,16 @@ func (ext *Extension) validateContents() error {
...
 		if d.Type().Type() == os.ModeSymlink {
-			return fmt.Errorf("symlinks are not allowed: %q", itemPath)
+			var dest string
+
+			dest, err = os.Readlink(path)
+			if err != nil {
+				return err
+			}
+
+			if strings.IndexByte(dest, '/') != -1 || dest == "." || dest == ".." {
```
frezbo, line 82: a symlink like this could potentially skip this test `ln -s ./../notes.txt foo`?

frezbo (changes requested):

internal/pkg/extensions/validate.go
smira: nope, as we don't allow slashes for now, so symlink could only be a filename like `a.so` -> `a.so`

internal/pkg/extensions/validate.go
```diff
@@ -88,24 +83,10 @@ func (ext *Extension) validateContents() error {
 		}
 
 		// no special files
-		if !d.IsDir() && !d.Type().IsRegular() {
+		if !d.IsDir() && !d.Type().IsRegular() && d.Type().Type() != os.ModeSymlink {
```
frezbo, line 86: `d.Type().Type()` intentional?

frezbo (changes requested):

internal/pkg/extensions/validate.go
smira: and it's covered by tests!

frezbo (approved):

PR #5139, opened 2022-03-16
fix: invert the condition to skip kubelet kernel checks

smira:
We should skip the checks on container platforms, as Talos has no way to
enforce conditions on the host kernel.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5139)


frezbo (approved):

Ulexus:
I don't know how I screwed that up, after going through all the trouble to carefully explain the logic.

Ulexus (approved):

PR #5145, opened 2022-03-18
fix: ignore pod CIDRs for kubelet node IPs

smira:
I'm not sure how I haven't noticed that before, but that is easily
reproducible with virtual IP moving between the nodes: Talos incorrectly
assumes that pod IPs might be valid kubelet node IPs, and this might
lead to unexpected results if the kubelet node IP is picked to be equal
to pod CIDR.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5145)

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/nodeip_config.go
+++ b/internal/app/machined/pkg/controllers/k8s/nodeip_config.go
@@ -89,6 +89,15 @@ func (ctrl *NodeIPConfigController) Run(ctx context.Context, r controller.Runtim
 
 				spec.ExcludeSubnets = nil
 
+				// filter out Pod & Service CIDRs, they can't be kubelet IPs
+				spec.ExcludeSubnets = append(
+					append(
+						spec.ExcludeSubnets,
+						cfgProvider.Cluster().Network().PodCIDRs()...,
+					),
+					cfgProvider.Cluster().Network().ServiceCIDRs()...,
+				)
+
 				// filter out any virtual IPs, they can't be node IPs either
 				for _, device := range cfgProvider.Machine().Network().Devices() {
 					if device.VIPConfig() != nil {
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/k8s/nodeip_config.go
```diff
@@ -89,6 +89,9 @@ func (ctrl *NodeIPConfigController) Run(ctx context.Context, r controller.Runtim
 
 				spec.ExcludeSubnets = nil
 
+				// filer out Pod CIDRs, they can't be kubelet IPs
```
frezbo, line 92:
```suggestion
				// filter out Pod CIDRs, they can't be kubelet IPs
```

frezbo:
I had seen this a lot in the logs and just ignored it :sweat_smile: :facepalm:

internal/app/machined/pkg/controllers/k8s/nodeip_config.go
```diff
@@ -89,6 +89,9 @@ func (ctrl *NodeIPConfigController) Run(ctx context.Context, r controller.Runtim
...
+				spec.ExcludeSubnets = append(spec.ExcludeSubnets, cfgProvider.Cluster().Network().PodCIDRs()...)
```
frezbo, line 93: would this exclude service IP's being broadcasted as ARP IP's? I've seen the same when the service IP's are broadcasted by kube-vip

smira: nope, but let me add that, safe to exclude service CIDRs as well

frezbo: the service CIDR's could be different from the ARP range though, but I don;t see any easy way to ignore this

Ulexus (approved):

PR #5160, opened 2022-03-21
chore: add day-two tests for e2e-qemu

TimJones:
Uses Sidero Labs d2ctl to install a small sample of
common cluster services to ensure the Talos cluster
is capable.

---
This change is [Reviewable](https://reviewable.io/reviews/talos-systems/talos/5160)

### DIFF
--- a/hack/test/day-two/config.yaml
+++ b/hack/test/day-two/config.yaml
@@ -1,0 +1,31 @@
+charts:
+  loki:
+    namespace: loki
+    repo: https://grafana.github.io/helm-charts
+    chart: loki-stack
+    valuesPath: hack/test/day-two/loki-values.yaml      
+
+  metallb:
+    namespace: metallb
+    repo: https://metallb.github.io/metallb
+    chart: metallb
+    valuesPath: hack/test/day-two/metallb-values.yaml      
+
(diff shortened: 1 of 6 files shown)


hack/test/day-two/config.yaml
```diff
@@ -0,0 +1,31 @@
...
+    chart: ingress-nginx
+    depends:
+      - metallb
+
+  rook:
+    namespace: rook-ceph
+    repo: https://charts.rook.io/release
+    chart: rook-ceph
+
+  ceph:
```
frezbo, line 26: it makes now sense to update the `e2e-qemi-csi` tests and use the charts for install, and move the tests somewhere with day2 tests

frezbo: https://github.com/talos-systems/talos/blob/master/hack/test/e2e.sh#L228

TimJones: Once I've confirmed the day-two tests work and cover the same case, I will remove the `e2e-qemu-csi` tests. I just didn't want to break anyone using them in the mean time :)

frezbo (approved):

smira (commented):
:tada:

PR #5202, opened 2022-03-24
docs: update docker guide for MacOS and Windows

andrewrynhard:
Adds a note to use the `--endpoint` flag on MacOS and Windows when using
`talosctl cluster create`.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5202)

### DIFF
--- a/website/content/v1.0/local-platforms/docker.md
+++ b/website/content/v1.0/local-platforms/docker.md
@@ -26,7 +26,13 @@ Further, when running on a Mac in docker,  due to networking limitations, VIPs a
 Creating a local cluster is as simple as:
 
 ```bash
-talosctl cluster create --wait
+talosctl cluster create
+```
+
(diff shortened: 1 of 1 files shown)


smira:
This should be automatic iirc.. I think it was, not sure if we broke that

andrewrynhard:
Didn’t work on windows at least.

steverfrancis:
On Mac it is not necessary to specify --endpoint.
Just tested with v1.0.0-beta.2

andrewrynhard:
> On Mac it is not necessary to specify --endpoint. Just tested with v1.0.0-beta.2

Sounds like we need same functionality on Windows then.

smira:
I will do a PR to fix this for windows, easy fix.

andrewrynhard:
> I will do a PR to fix this for windows, easy fix.

Cool. I will close this.

PR #5205, opened 2022-03-25
docs: fix extension service `path` dependency

frezbo:
Fix extension service `path` dependency

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5205)


smira:
we don't need to backport docs, as they're deployed from `master` always

frezbo:
Closing since docs are always from main branch

PR #5224, opened 2022-03-28
chore: use go:embed instead of ldflags

### LINKED ISSUE #5138: use //go:embed instead of ldflags -X
benefits:

* ldflags `-X` are not covered by Go backwards compatiblity
* builds not using our `Makefile` would still work

we can generate that embedded file in the buildkit as part of the build

for the git checkout, file should contain latest released tag with suffix `-git` as SHA

DmitriyMV:
Generate separate file for each variable and assign them during go build using go:embed instead of using ldflags -X.

Resolves #5138

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5224)


smira:
`nit`: it should be `chore:`, not `feat:`, as there are no features introduced for end users

Dockerfile
```diff
@@ -209,6 +226,11 @@ RUN --mount=type=cache,target=/.cache go generate ./...
 RUN goimports -w -local github.com/talos-systems/talos ./
 RUN gofumpt -w ./
 
+FROM build AS fix_embed
```
smira, line 229: nit: `fix-embed` (we don't use underscores)

Makefile
```diff
@@ -103,6 +92,11 @@ COMMON_ARGS += --build-arg=GO_BUILDFLAGS="$(GO_BUILDFLAGS)"
 COMMON_ARGS += --build-arg=GO_LDFLAGS="$(GO_LDFLAGS)"
 COMMON_ARGS += --build-arg=http_proxy=$(http_proxy)
 COMMON_ARGS += --build-arg=https_proxy=$(https_proxy)
+COMMON_ARGS += --build-arg=EMBED_NAME=$(NAME)
```
smira, line 95: once again nit, but I feel our convention was to use same variable names across Dockerfile/Makefile, so it might make sense to drop `EMBED_` prefix

smira (approved):
lgtm, just couple of nits

andrewrynhard:
I know I am late to the game here but I am curious if we haven't consider generating a YAML formatted file and embedding that? The individual files generated feels strange to me.

Question: Will every PR have to update these files?

Makefile
DmitriyMV: Fixed

DmitriyMV:
> I know I am late to the game here but I am curious if we haven't consider generating a YAML formatted file and embedding that? 

Separating into different files makes sense from compile time verification perspective. It will refuse to build if some of the files are missing. YAML doesn't give us this, so we risk panics if some of the info failed to properly generate.

> Will every PR have to update these files?

No - only on new releases or changes to image naming and such. Full commit SHA and TAG generation and embedding is delayed until later in Dockerfile.

Dockerfile
DmitriyMV: Missed that. Going to fix in the desk Dockerfile update.

PR #5238, opened 2022-03-29
docs: code block copying

frezbo:
Enable code blocks copying

Ref: https://aaronluna.dev/blog/add-copy-button-to-code-blocks-hugo-chroma/

Supersedes #5235

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5238)

### DIFF
--- a/website/config.toml
+++ b/website/config.toml
@@ -207,19 +207,19 @@ url = "https://twitter.com/SideroLabs"
 desc = "Development takes place here!"
 icon = "fab fa-github"
 name = "GitHub"
-url = "https://github.com/talos-systems/talos"
+url = "https://github.com/siderolabs/talos"
 
 [[menu.main]]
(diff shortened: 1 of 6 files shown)


Unix4ever (approved):

frezbo:
image

Updated css to show the content type too

smira:
should we also enable syntax highlighting and pull fixes from #5236 ?

andrewrynhard:
> should we also enable syntax highlighting and pull fixes from #5236 ?

Good question. If this is already a feature of the theme then we should use that instead of maintaining out own IMHO. We just got out of that mess 😄.

frezbo:
We already have syntax highlighting from whatever hugo uses

rsmitty (approved):

smira (approved):

PR #5268, opened 2022-03-31
release(v1.1.0-alpha.0): prepare release

smira:
This is the official v1.1.0-alpha.0 release.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5268)


frezbo (approved):

api/machine/machine.proto
```diff
@@ -82,12 +82,12 @@ message ApplyConfigurationRequest {
   bytes data = 1;
   // replaced by mode
   bool on_reboot = 2 [
-    (common.remove_deprecated_field) = "v1.1",
```
smira, line 85: bumped the deprecation, as at the very least we need to have it for the upgrade tests, and I think we should drop these pretty important things couple of cycles later

PR #5278, opened 2022-04-01
docs: override sitemap.xml to only include latest results

rsmitty:
This PR overrides the sitemap to only include, for example, v1.0 results
if that's our specified latest version. This should fix Google's
indexing of our docs to only include the latest results.


TimJones (approved):

smira (approved):

frezbo (approved):

steverfrancis:
The current sitemap at https://www.talos.dev/sitemap.xml has entries like this:
`<url>
<loc>/v0.12/introduction/</loc>
<lastmod>2022-03-18T08:45:03-04:00</lastmod>
</url>`
Google search console reject this sitemap: "Sitemap can be read, but has errors"

Looks like the <loc> field should specify the complete url:

From [https://www.sitemaps.org/protocol.html#locdef](https://www.sitemaps.org/protocol.html#locdef)
`<loc>| required | URL of the page. This URL must begin with the protocol (such as http) and end with a trailing slash, if your web server requires it. This value must be less than 2,048 characters.`


Looks like this change doesn't correct that - the <loc> is still just the  .Permalink, AFAICT...

rsmitty:
Yep, good call @steverfrancis. Some more color around this in the kubeflow docs here: https://github.com/kubeflow/website/issues/507

I went ahead and added talos.dev to our template. Will submit this same PR to sidero when this is in and we can confirm Google's parsing of our sitemap.

PR #5281, opened 2022-04-01
refactor: change the stages for embed files generation

smira:
The main goal to avoid cache invalidation on `tag` changes: by pushing
the tag down the build after the generate step, avoid invalidating build
cache on each tag change (`-dirty`, rebasing, etc.)

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5281)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -137,23 +137,6 @@ ENV PROTOTOOL_CACHE_PATH /.cache/prototool
 ARG SOURCE_DATE_EPOCH
 ENV SOURCE_DATE_EPOCH ${SOURCE_DATE_EPOCH}
 WORKDIR /src
-ARG NAME
-ARG SHA
-ARG USERNAME
-ARG REGISTRY
-ARG TAG
-ARG ARTIFACTS
-ARG PKGS
-ARG EXTRAS
-RUN mkdir -p pkg/machinery/gendata/data && \
-    echo -n ${NAME} > pkg/machinery/gendata/data/name && \
-    echo -n ${SHA} > pkg/machinery/gendata/data/sha && \
-    echo -n ${USERNAME} > pkg/machinery/gendata/data/username && \
-    echo -n ${REGISTRY} > pkg/machinery/gendata/data/registry && \
-    echo -n ${EXTRAS} > pkg/machinery/gendata/data/extras && \
-    echo -n ${PKGS} > pkg/machinery/gendata/data/pkgs && \
-    echo -n ${TAG} > pkg/machinery/gendata/data/tag && \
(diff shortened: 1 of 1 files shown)


frezbo (approved):

Dockerfile
```diff
@@ -226,11 +209,36 @@ RUN --mount=type=cache,target=/.cache go generate ./...
...
+
+FROM scratch AS embed
+COPY --from=embed-generate /src/pkg/machinery/gendata/data /pkg/machinery/gendata/data
+
+FROM embed-generate AS embed-abbrev-generate
 ARG ABBREV_TAG
 RUN echo -n "undefined" > pkg/machinery/gendata/data/sha && \
     echo -n ${ABBREV_TAG} > pkg/machinery/gendata/data/tag
 
+FROM scratch AS embed-abbrev
```
DmitriyMV, line 239: I'm unsure about this stage - the only reason I'm seeing is that we can use `COPY / /` in dependent stages, but since there is only one dependent stage for each branch I think it adds unnecessary level of indirection.

Dockerfile
```diff
@@ -226,11 +209,36 @@ RUN --mount=type=cache,target=/.cache go generate ./...
...
+    echo -n ${NAME} > pkg/machinery/gendata/data/name && \
+    echo -n ${SHA} > pkg/machinery/gendata/data/sha && \
+    echo -n ${USERNAME} > pkg/machinery/gendata/data/username && \
+    echo -n ${REGISTRY} > pkg/machinery/gendata/data/registry && \
+    echo -n ${EXTRAS} > pkg/machinery/gendata/data/extras && \
+    echo -n ${PKGS} > pkg/machinery/gendata/data/pkgs && \
+    echo -n ${TAG} > pkg/machinery/gendata/data/tag && \
+    echo -n ${ARTIFACTS} > pkg/machinery/gendata/data/artifacts
+
+FROM scratch AS embed
```
DmitriyMV, line 231: As with stage below, I'm also unsure about this.

DmitriyMV (commented):
Small nitpicks but nothing critical.

Dockerfile
smira: the reason I'm doing that is that it "hides" the details on which files should be copied to a single place; if later one we decide to move `pkg/machinery/gendata` to another location, we won't have to update each place 

plus it improves the caching - we make sure we don't depend on whole contents of `build`, just whatever was generated as the output, as this is what we depend on (when `COPY` is used, it will be invalidated each time `--from` changes, not taking into account what we're copying)

DmitriyMV: This is fair, I guess.

DmitriyMV (approved):

PR #5329, opened 2022-04-11
feat: migrate to go 1.18

DmitriyMV:
Increase go.mod version from 1.17 to 1.18 in all projects. Update Makefile to use latest tooling. Fix golangci by disable nolintlint for now.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5329)


.golangci.yml
```diff
@@ -155,6 +156,7 @@ linters:
     - maintidx
     - nestif
     - nilnil # we return "nil, nil"
+    - nolintlint
```
smira, line 159: just curious, what's the issue there?

smira (approved):

.golangci.yml
DmitriyMV: It shows a lot of false positives for some reason.

PR #5332, opened 2022-04-11
feat: build Talos with support for x86-64-v2 microarchitecture

smira:
See https://github.com/golang/go/wiki/MinimumRequirements#microarchitecture-support

This relies on new Go 1.18 feature to use more efficient x86-64
instructions.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5332)


frezbo (approved):
I hope this doesn;t break anyone running talos on super old hardware

smira:
> I hope this doesn;t break anyone running talos on super old hardware

yep, but they can do their own builds though? and it should benefit more users

PR #5333, opened 2022-04-11
feat: generic PoC in resource/cluster

DmitriyMV:
Add PoC implementation of generic typed resource. This implementation relies on phantom types, because passing callbacks to initializer is not reliable at all.

While this implementation looks promising at the first glance, it contains several drawbacks:
1. Phantom types look weird in Go. While they are indeed useful,  it adds a lot of "type info noise" to type signature.
2. We need to provide custom implementation of String and ResourceDefinition for each variant of resource type. And resulting types look weird, because they are essentially used for their methods, and never for their data.
3. Resulting code looks a lot more complicated.

I'm not sure about this implementation, but its the third iteration of my attempt to make generic Resource type.
First one was based on callbacks, and didn't work reliable.
The second used TypedResource and used a weird mix of embedding and "overriding".
While this code is not exactly Go'ish it doesn't look as weird as previous versions. Still, I'm unsure if we should do this.

CC @smira @Unix4ever

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5333)


pkg/machinery/resources/cluster/identity.go
```diff
@@ -18,10 +18,7 @@ const IdentityType = resource.Type("Identities.cluster.talos.dev")
 const LocalIdentity = resource.ID("local")
 
 // Identity resource holds node identity (as a member of the cluster).
-type Identity struct {
-	md   resource.Metadata
-	spec IdentitySpec
-}
+type Identity struct{}
```
frezbo, line 21: was removing the fields intentional?

DmitriyMV:
Closed for #5366

PR #5337, opened 2022-04-13
feat: add `dry-run` flag in `apply-config` and `edit` commands

Unix4ever:
Dry run prints out config diff, selected application mode without
changing the configuration.


frezbo:
shouldn't this work for `patch` too

internal/app/maintenance/server/server.go
```diff
@@ -83,6 +83,14 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 		},
 	}
 
+	if in.DryRun {
+		reply.Messages[0].ModeDetails = `dry run summary:
+node is running in mainenance mode
+applied config will be used after the reboot (skipped in dry run)`
```
smira, line 89: This is not quite true (`after a reboot`), as for disk image based setups there will be no reboot.

I would rather say that there's no existing machine configuration to compare with.

smira:
we need to have an integration test for it!

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -177,5 +179,6 @@ or 'notepad' for Windows.`,
 func init() {
 	editCmd.Flags().StringVar(&editCmdFlags.namespace, "namespace", "", "resource namespace (default is to use default namespace per resource)")
 	helpers.AddModeFlags(&editCmdFlags.Mode, editCmd)
```
smira, line 181: I might be wrong, but feels like `--dry-run` for `talosctl edit mc` is a bit weird, as you make changes, save them, and basically changes are lost?  (we can keep it for consistency sake still)

But at the same time `--dry-run` for `talosctl patch mc` might be interesting, as it gives us the ability to see the patch being applied, and it's easy to remove `--dry-run` after that to actually apply the changes

smira:
> we need to have an integration test for it!

bumping this :)

Unix4ever:
> bumping this :)

I've added a test already

hack/release.toml
```diff
@@ -85,6 +85,13 @@ The policy is part of the Talos machine configuration, and it can be modified to
         description="""\
 Talos is built for x86-64 architecture with support for [x86-64-v2 microarchitecture level](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels),
 so Talos no longer runs on processors supporting only baseline `x86-64` microarchitecture (before 2009).
+"""
+
+    [notes.apply-config]
+        title = "Apply Config `--dry-run`"
+        description="""\
+The commands `talosctl apply-config` and `talosctl edit mc` now support `--dry-run` flag.
```
smira, line 93: nit: add `talosctl patch mc` here

smira (approved):

PR #5344, opened 2022-04-13
feat: update etcd to 3.5.3

smira:
See https://github.com/etcd-io/etcd/releases/tag/v3.5.3

This release should contain a fix for data consistency issue when etcd
is killed under high load.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5344)


frezbo (approved):
do we need to disable the check we added earlier, i think i saw it's default now (have to confirm)

smira:
> do we need to disable the check we added earlier, i think i saw it's default now (have to confirm)

I believe it's  not default in 3.5, but I might be wrong

frezbo:
> > do we need to disable the check we added earlier, i think i saw it's default now (have to confirm)
> 
> I believe it's not default in 3.5, but I might be wrong

Should we set this `experimental-corrupt-check-time` too https://github.com/etcd-io/etcd/issues/9190#issue-290574193=

smira:
> Should we set this `experimental-corrupt-check-time` too [etcd-io/etcd#9190 (comment)](https://github.com/etcd-io/etcd/issues/9190#issue-290574193)

I'm concerned with the performance implications of that, and etcd maintainers say it's significant for big deployments.

PR #5356, opened 2022-04-15
fix: enable long timestamps for xfs

smira:
This "fixes" the message like:

```
xfs filesystem being mounted at /var supports timestamps until 2038 (0x7fffffff)
```

We should support Talos beyond 2038, even if we switch to a different
filesystem type by 2038 :)

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5356)


frezbo (approved):
:partying_face:

Unix4ever (approved):

PR #5369, opened 2022-04-18
docs: improve rook-ceph description

TimJones:
Improve the Rook Ceph documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5369)

### DIFF
--- a/website/content/v1.0/kubernetes-guides/configuration/ceph-with-rook.md
+++ b/website/content/v1.0/kubernetes-guides/configuration/ceph-with-rook.md
@@ -95,10 +95,11 @@ ceph-filesystem        rook-ceph.cephfs.csi.ceph.com   Delete          Immediate
 
 ## Talos Linux Considerations
 
-It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (be default `dataDirHostPath` is set to `/var/lib/rook`) which under Talos Linux is **ephemeral**.
-This makes cluster management a little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the ephemeral partition is wiped.
+It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (by default `dataDirHostPath` is set to `/var/lib/rook`).
+If running only a single `mon` instance, cluster management is little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the partition that stores the `/var` [file system]({{< relref "../../learn-more/architecture#the-file-system" >}}) is wiped, but the `--preserve` option of [`talosctl upgrade`]({{< relref "../../reference/cli#talosctl-upgrade" >}}) will ensure that doesn't happen.
 
-When performing maintenance on a Talos Linux node with a Rook Ceph cluster, it is imperative that care be taken to maintain the health of the Ceph cluster, for instance when upgrading the Talos Linux version.
+By default, Rook configues Ceph to have 3 `mon` instances, in which case the data stored in `dataDirHostPath` can be regenerated from the other `mon` instances.
+So when performing maintenance on a Talos Linux node with a Rook Ceph cluster (e.g. upgrading the Talos Linux version), it is imperative that care be taken to maintain the health of the Ceph cluster.
 Before upgrading, you should always check the health status of the Ceph cluster to ensure that it is healthy.
 
 ```shell
--- a/website/content/v1.1/kubernetes-guides/configuration/ceph-with-rook.md
+++ b/website/content/v1.1/kubernetes-guides/configuration/ceph-with-rook.md
@@ -95,10 +95,11 @@ ceph-filesystem        rook-ceph.cephfs.csi.ceph.com   Delete          Immediate
 
 ## Talos Linux Considerations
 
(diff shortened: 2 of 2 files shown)


website/content/v1.0/kubernetes-guides/configuration/ceph-with-rook.md
```diff
@@ -95,8 +95,8 @@ ceph-filesystem        rook-ceph.cephfs.csi.ceph.com   Delete          Immediate
 
 ## Talos Linux Considerations
 
-It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (be default `dataDirHostPath` is set to `/var/lib/rook`) which under Talos Linux is **ephemeral**.
-This makes cluster management a little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the ephemeral partition is wiped.
+It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (be default `dataDirHostPath` is set to `/var/lib/rook`).
```
frezbo, line 98:
```suggestion
It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (by default `dataDirHostPath` is set to `/var/lib/rook`).
```

website/content/v1.0/kubernetes-guides/configuration/ceph-with-rook.md
```diff
@@ -95,8 +95,8 @@ ceph-filesystem        rook-ceph.cephfs.csi.ceph.com   Delete          Immediate
...
+This makes cluster management a little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the partition that stores the `/var` [file system]({{< relref "../../learn-more/architecture#the-file-system" >}}) is wiped, but the `--preserve` option of [`talosctl upgrade`]({{< relref "../../reference/cli#talosctl-upgrade" >}}) can be used to avoid this.
```
frezbo, line 99: I think we should mention that if there is a odd number of OSD, the preserve flag could be avoided and wait for rook to mark oSD back as healthy after upgrade or manually mark the OSD's as paused (I'm not sure what;s the right ceph wording is here)

TimJones: Number of OSDs doesn't really matter as much as capacity & pool replica count.

In fact, Rook is smart enough to rebuild the mon data stored in the `/var` when the node comes back up. In the guide itself I never use `--preserve` when doing an upgrade since Ceph marks the OSDs as 'down' and not 'out', so when they come back 'up' only the changes have to be synced back for the OSD to be brought back 'in' and the cluster to be 'healthy' again.

Using `--preserve`, like with a non-HA cluster, is only really needed if we only have one `mon`.

TimJones: Ahhh, you mean if there are multiple `mon`s, yeah, that makes more sense! I'll add it.

smira: I think by default it always does three mons, so Rook/Ceph won't even deploy with less than 3 worker (schedulable) nodes

TimJones: For sure. It's a balance between what is a consideration for Rook/Ceph or Talos Linux.

frezbo (approved):

smira (approved):

website/content/v1.0/kubernetes-guides/configuration/ceph-with-rook.md
```diff
@@ -95,10 +95,11 @@ ceph-filesystem        rook-ceph.cephfs.csi.ceph.com   Delete          Immediate
 
 ## Talos Linux Considerations
 
-It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (be default `dataDirHostPath` is set to `/var/lib/rook`) which under Talos Linux is **ephemeral**.
-This makes cluster management a little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the ephemeral partition is wiped.
+It is important to note that a Rook Ceph cluster saves cluster information directly onto the node (by default `dataDirHostPath` is set to `/var/lib/rook`).
+If running only a single `mon` instance, cluster management is little bit more involved, as any time a Talos Linux node is reconfigured or upgraded, the partition that stores the `/var` [file system]({{< relref "../../learn-more/architecture#the-file-system" >}}) is wiped, but the `--preserve` option of [`talosctl upgrade`]({{< relref "../../reference/cli#talosctl-upgrade" >}}) will ensure that doesn't happen.
```
Timoses, line 99: > is **a** little bit more involved
(without the 'a' it sounds odd to my non-native English ears)

Timoses, line 99: > reconfigured or upgraded, the partition [...] is wiped, [...]

Is that correct in general, that partitions are wiped for ALL reconfiguration scenarios (even in those where no node reboot is needed)?

Timoses:
Apart from comments, LGTM

PR #5375, opened 2022-04-18
fix: provide logger to the etcd snapshot restore

smira:
With update of the client library to 3.5.3, etcd library started using
the logger, so using `nil` isn't fine anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5375)

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -42,6 +42,7 @@ import (
 	"github.com/talos-systems/talos/internal/pkg/etcd"
 	"github.com/talos-systems/talos/pkg/argsbuilder"
 	"github.com/talos-systems/talos/pkg/conditions"
+	"github.com/talos-systems/talos/pkg/logging"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -650,7 +651,7 @@ func (e *Etcd) argsForControlPlane(ctx context.Context, r runtime.Runtime) error
 
 // recoverFromSnapshot recovers etcd data directory from the snapshot uploaded previously.
 func (e *Etcd) recoverFromSnapshot(hostname, primaryAddr string) error {
-	manager := snapshot.NewV3(nil)
+	manager := snapshot.NewV3(logging.Wrap(log.Writer()))
```
frezbo, line 654: `log.Writer` on purpose?, the others used `os.Stderr`

frezbo (approved):

internal/app/machined/pkg/system/services/etcd.go
smira: yep, this one is _inside_ machined, so it should log properly to the console (`log.Writer` is set up for that)

other uses are in the CLI

PR #5378, opened 2022-04-18
Docs get machine config

### LINKED ISSUE #5296: Docs: Add to docs: Get controlplane config after interactive setup?
### Discussed in https://github.com/siderolabs/talos/discussions/5282

<div type='discussions-op-text'>

<sup>Originally posted by **jetersen** April  2, 2022</sup>
Is it possible to retrieve the config after interactive setup?

Or should I simply reinstall with non interactive?</div>

TimJones:
Add a note on how machine configuration can be retrieved
from the node, after e.g. interactive setup.

Closes #5296

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5378)

### DIFF
--- a/website/content/v1.1/talos-guides/configuration/editing-machine-configuration.md
+++ b/website/content/v1.1/talos-guides/configuration/editing-machine-configuration.md
@@ -28,6 +28,9 @@ Each of these commands can operate in one of four modes:
 > Note: applying change on next reboot (`--mode=staged`) doesn't modify current node configuration, so next call to
 > `talosctl edit machineconfig --mode=staged` will not see changes
 
+Additionally, there is also `talosctl get machineconfig`, which retrieves the current node configuration API resource and contains the machine configuration in the `.spec` field.
+It can be used to modify the configuration locally before being applied to the node.
+
 The list of config changes allowed to be applied immediately in Talos {{< release >}}:
 
 * `.debug`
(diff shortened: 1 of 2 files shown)


website/content/v1.1/talos-guides/configuration/editing-machine-configuration.md
```diff
@@ -28,6 +28,9 @@ Each of these commands can operate in one of four modes:
 > Note: applying change on next reboot (`--mode=staged`) doesn't modify current node configuration, so next call to
 > `talosctl edit machineconfig --mode=staged` will not see changes
 
+Additionally, there is also `talosctl get machineconfig`, which retrieves the current node configuration API resource and contains the machine configuration in the `.spec` field.
```
smira, line 31: I'm torn on this one. If you try to extract the `.spec` field with `yq`, it does slight reformatting, so what you get is not byte-by-byte what you submitted.

On other hand we have `talosctl read /system/state/config.yaml` which does exactly what you submitted, but this one looks weird. 

It's more of a question I'm struggling with myself, rather than related to the PR :)

smira (approved):

website/content/v1.1/talos-guides/configuration/editing-machine-configuration.md
TimJones: I struggled with this one too, and that may be an indicator that we need to figure out a better way of doing this.

In the context of the original conversation and page topic, I don't think a byte-for-byte copy is really warranted or needed, especially if the user has gone through interactive setup as they haven't "submitted" anything themselves.

frezbo: or we could have a new `talosctl machineconfig` similar to `talosctl kubeconfig` which just wraps the reading of `/system/state/config,yaml`

PR #5381, opened 2022-04-19
feat: add support for equinix metal events

rsmitty:
This PR gets us closer to being a supported option on Equinix Metal. We
now look for a new kernel arg `em.events_url` and will send events there
based on the state of our install. These events are then presented to
the user in the GUI or via calls to the EM API.

*** Note: this is currently a WIP as the EM folks aren't yet ready to
test this out ***

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5381)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -48,6 +48,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/adv"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/platform"
 	perrors "github.com/talos-systems/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/events"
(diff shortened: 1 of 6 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -86,9 +86,13 @@ const (
 	// KernelParamPanic is the kernel parameter name for specifying the time to wait until rebooting after kernel panic (0 disables reboot).
 	KernelParamPanic = "panic"
 
-	// KernelParamSideroLink is the kernel paramater name to specify SideroLink API endpoint.
+	// KernelParamSideroLink is the kernel parameter name to specify SideroLink API endpoint.
 	KernelParamSideroLink = "siderolink.api"
 
+	// KernelParamEquinixMetalEvents is the kernel parameter name to specify the Equinix Metal phone home endpoint.
+	// This param is injected by Equinix Metal and depends on the device ID and datacenter.
+	KernelParamEquinixMetalEvents = "em.events_url"
```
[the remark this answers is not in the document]
rsmitty: nope, has to be injected as a kernel param as far as I understand.

smira (commented):
on a more of an overview comment, I wonder if we could make this code more generic so that in the future we could make more platforms support events.

so if we could abstract away a bit the events themselves to be not specific to the EM, and make a generic function which tries to send event to the platform if it implements the required interface, it will cut the duplication and also allow future expansion

rsmitty:
Yeah that's probably do-able. Let me think through what it looks like and we can walk through it tomorrow.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -470,6 +472,18 @@ func LoadConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFu
...
 				logger.Println("machine configuration not found; starting maintenance service")
 
+				// nb: we treat maintenance mode as an "activate"
+				// event b/c the user is expected to be able to
+				// interact with the system at this point.
+				platform.FireEvent(
+					ctx,
+					r.State().Platform(),
+					platform.Event{
+						Type:    equinixmetal.EventTypeActivate,
```
smira, line 482: I feel these type constants should be in common `platform` package?

As it seems weird that common code imports equinixmetal specifically? 

We can for now pick EquinixMetal names for the constants, and add translation layer later as we have other platforms with event support.

rsmitty: hrmm, yeah that seems like it makes sense. let me take a stab at that.

smira (approved):

PR #5384, opened 2022-04-19
fix: detect lingering mounts in the installer correctly

smira:
Not sure how and when it got broken, but we're looking for mounts for
the blockdevice (like `/dev/vda`), while the actual mount info contains
the partition device (like `/dev/vda6`).

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5384)


frezbo (approved):

smira:
thanks to @frezbo for finding this bug!

PR #5391, opened 2022-04-20
docs: add canonical link ref

### LINKED ISSUE #5336: Docs: identify canonical URLs
## Feature Request
Identify canonical pages explicitly.

### Description
Google still returns stale documentation URLs, for old versions.
We only have the current versions in the sitemap, so while google generally seems to make the current version canonical - it does not usually identify old versions as the "same" page, so they can still show up in the search results. Sometimes in preference to the current version, presumably as there are external backlinks to old pages...

So, we need to explicitly identify the current version as the canonical page, so that the backlinks SEO weight transfer to the current page.
So with each new release of the documentation, we need to update all old version pages, and link to the latest version.
https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-bananapi_m64#rel-canonical-link-method

Note: we cannot link to /latest/ as that is a redirect - have to link to the explicit page.

So something like:
for each filename (e.g. "bananapi_m64") in current version of docs, do:
    find all files of same name in all old versions
            update the head:
            <link rel="canonical" href="https://www.talos.dev/v1.0/single-board-computers/bananapi_m64/" />
done

TimJones:
Adds a canonical link tag to doc pages
to help SEO find the current version of
documentation.

Closes #5336

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5391)

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,5 +1,5 @@
 [build]
-command = "cd website && hugo --gc --minify"
+command = "cd website && hugo --gc --minify --baseURL $DEPLOY_PRIME_URL"
 publish = "website/public"
 
 [build.environment]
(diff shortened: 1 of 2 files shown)


frezbo (approved):
not sure what i does :sweat_smile: , but I believe this fixes it

PR #5403, opened 2022-04-21
chore: allow mounting files from the host

frezbo:
Allow mounting files from host into extension services as per the [OCI
spec](https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts)

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5403)

### DIFF
--- a/internal/app/machined/pkg/system/services/extension.go
+++ b/internal/app/machined/pkg/system/services/extension.go
@@ -6,6 +6,7 @@ package services
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/services/extension.go
```diff
@@ -102,8 +103,13 @@ func (svc *Extension) Runner(r runtime.Runtime) (runner.Runner, error) {
 	}
 
 	for _, mount := range svc.Spec.Container.Mounts {
-		if err := os.MkdirAll(mount.Source, 0o700); err != nil {
-			return nil, err
+		// if file already exists skip
```
smira, line 106: I would rewrite it with less nesting:

```suggestion
if _, err := os.Stat(mount.Source); err == nil {
   // already exists, skip
   continue
} else if !errors.Is(err, os.ErrNotExist) {
   return nil, err
}

if err := os.MkdirAll(mount.Source, 0o700); err != nil {
   return nil, err
}
```

smira (approved):

PR #5408, opened 2022-04-21
feat: increase aio-max-nr and inotify.max_user_instances

### LINKED ISSUE #5175: sysctls for I/O performance
## Feature Request

In testing heavy I/O loads with Ceph, a couple sysctls came up which should probably be bumped by default.  We already bumped inotify recently, but it seems a bit more may be in order.

`fs.aio-max-nr` = 1048576
`fs.inotify.max_user_instances` = 8192

DmitriyMV:
Increase values:
- fs.aio-max-nr to 1048576 (for Ceph\Veritas\other storages)
- fs.inotify.max_user_instances to 8192 (since the usual 512 is too small today's needs)

There is no need to adjust fs.inotify.max_user_watches since it's set dynamically during startup by kernel.

Closes #5175

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5408)

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/kernel_param_defaults.go
+++ b/internal/app/machined/pkg/controllers/runtime/kernel_param_defaults.go
@@ -139,7 +139,11 @@ func (ctrl *KernelParamDefaultsController) getKernelParams() []*kernel.Param {
 		// configs inotify.
 		{
 			Key:   "proc.sys.fs.inotify.max_user_instances",
-			Value: "512",
+			Value: "8192",
+		},
+		{
+			Key:   "proc.sys.fs.aio-max-nr",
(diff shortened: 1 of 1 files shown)


smira (approved):
I believe there are some tests which need to be updated

DmitriyMV:
@frezbo Judging by https://github.com/torvalds/linux/commit/92890123749bafc317bbfacbe0a62ce08d78efb7 and since Talos requires 5.15 as minimal kernel version, we don't actually need to set `fs.inotify.max_user_watches` [since](https://stackoverflow.com/questions/535768/) it's set [automatically](https://www.kernel.org/doc/Documentation/sysctl/fs.txt).

PR #5440, opened 2022-04-25
docs: reformat config reference

TimJones:
Update the configuration reference documentation
to show field information in a tabular format.

Example of the tabular layout:
image

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5440)

### DIFF
--- a/pkg/machinery/config/encoder/markdown.go
+++ b/pkg/machinery/config/encoder/markdown.go
@@ -17,18 +17,11 @@ import (
 )
 
 var markdownTemplate = `
-{{ define "fieldExamples" }}
-Examples:
-
-{{ range $example := .Examples }}
-{{ yaml $example.GetValue $.Name }}
-{{ end }}
-{{ end }}
-
 {{ .Description }}
 {{- $anchors := .Anchors -}}
 {{- $tick := "` + "`" + `" -}}
 {{ range $struct := .Structs }}
+---
 ## {{ $struct.Type }}
 {{ if $struct.Description -}}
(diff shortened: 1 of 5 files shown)


smira:
I don't know whether it was intentional or not, but inline examples are now missing, e.g. for `MachineConfig` `sysctls` field (and all other fields). Without these examples it's hard to see what to put as the values of these fields as they don't have any specific sub-section.

smira:
I wonder if we should follow table with some examples? or have per-field example below the table? I don't know what is the best, just throwing out ideas.

TimJones:
Now with expandable, syntax-highlighted examples!!
Screenshot_20220426_163229

pkg/machinery/config/encoder/markdown.go
```diff
@@ -195,9 +158,18 @@ func encodeYaml(in interface{}, name string) string {
 		lines[i] = strings.TrimRight(line, " ")
 	}
 
-	return fmt.Sprintf("``` yaml\n%s```", strings.Join(lines, "\n"))
+	return fmt.Sprintf("{{< highlight yaml >}}\n%s{{< /highlight >}}", strings.Join(lines, "\n"))
```
TimJones, line 161: A possible issue here is that this is coupling the outputted markdown file with Hugo's rendering engine.

frezbo:
> Now with expandable, syntax-highlighted examples!! Screenshot_20220426_163229

that is some vodoo magic :mage:

smira (approved):
:heart:

PR #5451, opened 2022-04-27
fix: don't mount D-Bus socket via mount under recursive bind mount

smira:
`/var/run` was mounted from `/run`, and D-Bus socket to `/var/run/dbus/`
path, so when the container is stopped, container mounts are removed,
but on the host side mount propagates back, so D-Bus socket gets
propagated back to the host `/run`, and on the next kubelet restart
process continues adding even more mount levels exponentially.
Eventually on kubelet restart kernel resources are exhausted and the
node freezes.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5451)


frezbo:
good catch :+1:

PR #5455, opened 2022-04-27
feat: add bond slaves ordering

### LINKED ISSUE #5207: bond interface order is not preserved from machine configuration to resources
E.g. in the machine configuration:

```yaml
- name: bond0
  bond:
     interfaces:
        - eth1
        - eth0
```

When this gets converted to internal networking resources, order of links becomes alphabetical (based on resource IDs), and actual link order will be `eth0`, `eth1`

DmitriyMV:
Before this change, we didn't preserve bonded interfaces ordering, which caused problems in some scenarios. Fix this by remembering their position in the original config.

Fixes #5207.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5455)

### DIFF
--- a/internal/app/machined/pkg/controllers/network/link_spec_test.go
+++ b/internal/app/machined/pkg/controllers/network/link_spec_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/cosi-project/runtime/pkg/state/impl/inmem"
 	"github.com/cosi-project/runtime/pkg/state/impl/namespaced"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	"github.com/talos-systems/go-retry/retry"
 	"golang.zx2c4.com/wireguard/wgctrl/wgtypes"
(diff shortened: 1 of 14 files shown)


internal/app/machined/pkg/controllers/network/link_spec_private_test.go
```diff
@@ -0,0 +1,5 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package network_test
```
smira, line 5: remove it?

internal/app/machined/pkg/controllers/network/link_spec_test.go
```diff
@@ -371,24 +372,30 @@ func (suite *LinkSpecSuite) TestBond() {
...
-		Logical:     true,
-		MasterName:  bondName,
+		Name:    dummy1Name,
+		Type:    nethelpers.LinkEther,
+		Kind:    "dummy",
+		Up:      true,
+		Logical: true,
+		BondSlave: network.BondSlave{
+			MasterName: bondName,
+			SlaveIndex: 0,
```
smira, line 397:
```suggestion
			SlaveIndex: 1,
```

So that the order of enslaving is deterministic?

smira (approved):
with small nits

internal/app/machined/pkg/controllers/network/link_spec_test.go
DmitriyMV: Done.

internal/app/machined/pkg/controllers/network/link_spec_private_test.go
DmitriyMV: Done.

PR #5456, opened 2022-04-27
fix: allow graceful node shutdown to be overridden

smira:
The problem is that these values needs to be set to zero if the kubelet
feature gate is disabled, so we can't assume that we can override zero
value with the proper config, so we have to do an extra check on the
supplied configuration.

Also creates KB article on disabling this feature gate.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5456)

### DIFF
--- a/website/content/v1.1/learn-more/knowledge-base.md
+++ b/website/content/v1.1/learn-more/knowledge-base.md
@@ -1,0 +1,21 @@
+---
+title: "Knowledge Base"
+weight: 1999
+description: "Recipes for common configuration tasks with Talos Linux."
+---
+
+## Disabling `GracefulNodeShutdown` on a node
+
+Talos Linux enables [Graceful Node Shutdown](https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown) Kubernetes feature by default.
+
+If this feature should be disabled, modify the `kubelet` part of the machine configuration with:
+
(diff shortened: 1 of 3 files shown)


website/content/v1.1/learn-more/knowledge-base.md
```diff
@@ -0,0 +1,21 @@
...
+
+Talos Linux enables [Graceful Node Shutdown](https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown) Kubernetes feature by default.
+
+If this feature should be disabled, modify the `kubelet` part of the machine configuration with:
+
+```yaml
+machine:
+  kubelet:
+    extraArgs:
+      feature-gates: GracefulNodeShutdown=false
```
frezbo, line 17: i believe disabling the feature gate is enough

frezbo (approved):

website/content/v1.1/learn-more/knowledge-base.md
smira: it's not - that's the problem, kubelet fails if it finds non-zero settings for graceful node shutdown and the feature gate is disabled, that's what I wasted half of the day on :-D

frezbo: weird :sweat_smile:

PR #5460, opened 2022-04-28
chore: disable integration test with Cilium

smira:
The version we used is broken now:

```
  Warning  FailedCreatePodSandBox  6s     kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "506a948e4fcccf9941108c794f3f5020cd30bda55f276400232cf36e2272fb45": plugin type="loopback" failed (add): incompatible CNI versions; config is "1.0.0", plugin supports ["0.1.0" "0.2.0" "0.3.0" "0.3.1" "0.4.0"]
```

Newer versions can't be installed via the manifest, we definitely need a
better way to test with Cilium, but for now disabling the test is the
only way.


frezbo (approved):

smira:
I'll put this on hold until it's clear what's up with containerd 1.6.3, as e2e-aws is also broken, as it's using Calico.

PR #5461, opened 2022-04-28
docs: provide example on using config generation package

smira:
There were many discussions on creating native Talos providers for TF,
Pulumi, etc., but there's no documented idiomatic way to use our
machinery package to generate the config. This PR tries to fill this
gap.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5461)

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/generate/example_test.go
+++ b/pkg/machinery/config/types/v1alpha1/generate/example_test.go
@@ -1,0 +1,102 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package generate_test
+
+import (
+	"log"
+	"os"
+
+	"github.com/talos-systems/talos/pkg/machinery/config"
+	v1alpha1 "github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/generate"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+	"github.com/talos-systems/talos/pkg/machinery/constants"
+)
+
+//nolint:wsl
+func Example() {
+	// This is an example of generating a set of machine configuration files for multiple
+	// nodes of the cluster from a single cluster-specific cluster.
+
+	// Input values for the config generation:
+
+	// * cluster name and Kubernetes control plane endpoint
+	clusterName := "test-cluster"
+	controlPlaneEndpoint := "https://kubernetes.example.com:6443"
+
+	// * Kubernetes version to install, using the latest here
+	kubernetesVersion := constants.DefaultKubernetesVersion
+
+	// * version contract defines the version of the Talos cluster configuration is generated for
+	//   generate package can generate machine configuration compatible with current and previous versions of Talos
+	targetVersion := "v1.0"
+
+	// parse the version contract
+	var (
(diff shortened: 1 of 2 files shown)


smira:
image

pkg/machinery/config/types/v1alpha1/generate/example_test.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// generate the machine config for each node of the cluster using the secrets
+	for _, node := range []string{"machine1", "machine2"} {
+		var cfg *v1alpha1.Config
+
+		// generate the machine config for the node, using the right machine type:
+		// * machine.TypeConrolPlane for control plane nodes
+		// * machine.TypeWorker for worker nodes
+		cfg, err = generate.Config(machine.TypeControlPlane, input)
```
DmitriyMV, line 69: Can't we create `cfg` on assignment and just shadow `err`?

pkg/machinery/config/types/v1alpha1/generate/example_test.go
```diff
@@ -0,0 +1,102 @@
...
+			log.Fatalf("failed to generate config for node %q: %s", node, err)
+		}
+
+		// config can be tweaked at this point to add machine-specific configuration, e.g.:
+		cfg.MachineConfig.MachineInstall.InstallDisk = "/dev/sdb"
+
+		// marshal the config to YAML
+		var marshaledCfg []byte
+
+		marshaledCfg, err = cfg.Bytes()
```
DmitriyMV, line 80: Can't we create `marshaledCfg` on assignment and just shadow `err`?

smira: linters start complaining, and I didn't want to do shadowing in both cases

smira: for the sake of the example, having explicit types is better imho for readability

DmitriyMV (approved):

frezbo:
I think we should make this more easier to use later on. In the pulumi provider I still had to copy a lot of code wrt registry mirrors parsing, validations etc https://github.com/frezbo/pulumi-provider-talos/blob/main/provider/pkg/provider/provider.go#L347=

frezbo (approved):

pkg/machinery/config/types/v1alpha1/generate/example_test.go
```diff
@@ -0,0 +1,102 @@
...
+		// generate the machine config for the node, using the right machine type:
+		// * machine.TypeConrolPlane for control plane nodes
+		// * machine.TypeWorker for worker nodes
+		cfg, err = generate.Config(machine.TypeControlPlane, input)
+		if err != nil {
+			log.Fatalf("failed to generate config for node %q: %s", node, err)
+		}
+
+		// config can be tweaked at this point to add machine-specific configuration, e.g.:
+		cfg.MachineConfig.MachineInstall.InstallDisk = "/dev/sdb"
```
frezbo, line 75: isn't it easier to use `generate.WithInstallDisk()`

smira: it is, but I wanted to give an example of config "patching". I first wanted to see if I should do our JSON patches, but the code looks so weird that I decided to do direct modification of the structure.

smira:
> I think we should make this more easier to use later on. In the pulumi provider I still had to copy a lot of code wrt registry mirrors parsing, validations etc https://github.com/frezbo/pulumi-provider-talos/blob/main/provider/pkg/provider/provider.go#L347=

yes, while trying to write this example I realized how weird this path looks like with the Input generation (why?) followed by config generation. but we need something for now that we can improve and also it serves as an "API" test, as it compiles with every commit, so we know we haven't broken the external API.

PR #5472, opened 2022-05-01
chore: refactor remaining resources into typed.Resource

DmitriyMV:
Refactor remaining resources into typed.Resource. Exceptions are:
- MachineConfig
- MachineType
- LinkRefresh
- LinkStatus
all of which contain additional methods, and cannot be simply reworked into new resource framework.

StaticPod and StaticPodStatus are also absent from this PR, because they result in e2e errors which are going to be resolved in the next PR.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5472)


smira (approved):

smira:
> * LinkRefresh
> * LinkStatus

these two resources additional methods should be simply moved to the `*Spec` type, that should be an easy change

PR #5487, opened 2022-05-03
fix: talosctl exits with error if gen option and --input-dir flags are combined

### LINKED ISSUE #2275: talosctl cluster create: fail if gen options with --input-dir
## Feature Request

There are many config generation options in `talosctl cluster create` which make no sense with `--input-dir` as they're going to be ignored. 

### Description

Fail with error message is any gen options are used with `--input-dir`. This might confuse users as arguments are silently ignored.

sauterp:
## Changes

The user will get an error message and talosctl aborts if `talosctl cluster create` is called with gen options and the --input-dir flag. Previously the gen options would simply be ignored without warning.

## Acceptance

Fixes #2275

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5487)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -21,6 +21,7 @@ import (
 	humanize "github.com/dustin/go-humanize"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"github.com/talos-systems/go-blockdevice/blockdevice/encryption"
 	"github.com/talos-systems/go-procfs/procfs"
 	talosnet "github.com/talos-systems/net"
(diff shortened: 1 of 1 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -127,6 +158,14 @@ var createCmd = &cobra.Command{
 	Short: "Creates a local docker-based or QEMU-based kubernetes cluster",
 	Long:  ``,
 	Args:  cobra.NoArgs,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		definedGenFlag := checkForDefinedGenFlag(cmd.Flags())
+		if definedGenFlag != "" {
```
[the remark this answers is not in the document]
smira: we could probably move this check closer to the code which checks for `input-dir` flag presence vs. config generqtion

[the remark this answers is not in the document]
sauterp: I tried to do the check where you suggest, but there was a problem. That check requires access to the flags. If I use createCmd.Flags() inside the create() function I get an "initialization cycle" because create() is used inside the initialization of createCmd.
create() only accepts a context.Context as input, so I considered passing in the flags there, but that would mean using a type assertion inside create() when I get the flags out again from the context.Context.
That's why I decided to do it in the PreRunE.

Now I changed the signature of create, since it's only used exactly once and the context.Background() happens inside create.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -128,12 +158,14 @@ var createCmd = &cobra.Command{
...
 }
 
 //nolint:gocyclo,cyclop
-func create(ctx context.Context) (err error) {
+func create(flags *pflag.FlagSet) (err error) {
+	ctx := context.Background()
+
 	if masters < 1 {
 		return fmt.Errorf("number of masters can't be less than 1")
 	}
```
DmitriyMV, line 171: This introduces big but subtle change - `cli.WithContext` contains a logic which will cancel passed context if there was `SIGINT` or `SIGTERM` signal. Your codes doesn't do that, so it will no longer properly shutdown in case of signal.

I suggest you change this code to:
```go
		return cli.WithContext(context.Background(), func(ctx context.Context) error {
			create(ctx, cmd.Flags())
		})
	},
}

//nolint:gocyclo,cyclop
func create(ctx context.Context, flags *pflag.FlagSet) (err error) {
	if masters < 1 {
		return fmt.Errorf("number of masters can't be less than 1")
	}
```

DmitriyMV (changes requested):

cmd/talosctl/cmd/mgmt/cluster/create.go
sauterp: I didn't know that! thanks

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -445,6 +483,7 @@ func create(ctx context.Context) (err error) {
 		}
 
 		genOptions = append(genOptions, generate.WithEndpointList(endpointList))
+
```
DmitriyMV, line 486: Minor nit - do we really need empty line here? Lints complaining?

DmitriyMV (approved):

PR #5489, opened 2022-05-03
chore: refactor LinkRefresh and LinkStatus into typed.Resource

DmitriyMV:
From #5472 Andrey comments, this commit changes LinkRefresh and LinkStatus into typed.Resource by moving Bump and Physical methods to *Spec types.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5489)

### DIFF
--- a/pkg/machinery/resources/network/link_refresh.go
+++ b/pkg/machinery/resources/network/link_refresh.go
@@ -7,6 +7,7 @@ package network
 import (
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
 )
 
 // LinkRefreshType is type of LinkRefresh resource.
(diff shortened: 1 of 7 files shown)


pkg/machinery/resources/network/link_refresh.go
```diff
@@ -19,57 +21,46 @@ const LinkRefreshType = resource.Type("LinkRefreshes.net.talos.dev")
...
-type LinkRefresh struct {
-	md   resource.Metadata
-	spec LinkRefreshSpec
-}
+type LinkRefresh = typed.Resource[LinkRefreshSpec, LinkRefreshRD]
 
 // LinkRefreshSpec describes status of rendered secrets.
 type LinkRefreshSpec struct {
-	Generation int `yaml:"generation"`
+	Generation *int `yaml:"generation"`
```
smira, line 28: I might be missing something, but I thought `TypedSpec()` returns a pointer to the spec? 

so it should be fine to do `(*LinkRefreshSpec) Bump` ?

DmitriyMV: Not really. `TypedSpec()` indeed returns a pointer, but we still need to defined method with non-pointer receiver, since generic constraints expects method to have a non-pointer receiver. Basically `(*LinkRefreshSpec) Bump` does not compile.

smira: Sounds like a limitation to me, why should it care... 

I wonder if two values of LinkRefresh resource are still equal after DeepCopy? I hope they are, as it's using reflect.DeepEqual behind the scenes

DmitriyMV: I tried to explain the limitation here: https://go.dev/play/p/NObWzQao8cR

PR #5497, opened 2022-05-04
chore: refactor StaticPod and StaticPodStatus into typed.Resource

DmitriyMV:
This two required some additional attention and were split into separate branch from #5472.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5497)

### DIFF
--- a/pkg/machinery/resources/network/node_address.go
+++ b/pkg/machinery/resources/network/node_address.go
@@ -42,7 +42,9 @@ type NodeAddressSpec struct {
 
 // DeepCopy generates a deep copy of NodeAddressSpec.
 func (spec NodeAddressSpec) DeepCopy() NodeAddressSpec {
-	return spec
+	return NodeAddressSpec{
+		Addresses: append([]netaddr.IPPrefix(nil), spec.Addresses...),
+	}
 }
 
 // NewNodeAddress initializes a NodeAddress resource.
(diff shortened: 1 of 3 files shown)


DmitriyMV:
So the root problem was that old `StaticPodSpec.MarshalYAML` was defined on pointer receiver. That is
```
func (spec *StaticPodSpec) MarshalYAML() (interface{}, error) {
```

With typed.Resource the new code `Spec() interface{}` returns `StaticPodSpec{}` and not `*StaticPodSpec{}` which is bad if you are trying to call any methods on returned `interface{}` using type assertion, since you can't call methods which expect pointer receiver on non-pointer interface.

Minimal repro/example: https://go.dev/play/p/PLxBYfXoLCf

frezbo:
good catch

DmitriyMV:
> good catch

Big thanks to @smira for saying that `MarshalYAML` happen on empty interface.

pkg/machinery/resources/network/node_address.go
```diff
@@ -42,7 +42,9 @@ type NodeAddressSpec struct {
 
 // DeepCopy generates a deep copy of NodeAddressSpec.
 func (spec NodeAddressSpec) DeepCopy() NodeAddressSpec {
-	return spec
+	return NodeAddressSpec{
+		Addresses: append([]netaddr.IPPrefix(nil), spec.Addresses...),
```
smira, line 46: :+1:

smira (approved):

frezbo (approved):

PR #5498, opened 2022-05-04
feat: add more hardware information to the link status resources

Unix4ever:
The new fields will be used by the network device selector.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5498)

### DIFF
--- a/internal/app/machined/pkg/controllers/network/link_status_test.go
+++ b/internal/app/machined/pkg/controllers/network/link_status_test.go
@@ -7,10 +7,12 @@ package network_test
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"math/rand"
 	"net"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/controllers/network/link_status_test.go
```diff
@@ -126,6 +128,59 @@ func (suite *LinkStatusSuite) assertNoInterface(id string) error {
 }
 
 func (suite *LinkStatusSuite) TestLoopbackInterface() {
```
smira, line 130: feels like test name should be swapped with the next test?

pkg/machinery/nethelpers/device.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	if ueventContents == "" {
+		return &DeviceInfo{}, nil
+	}
+
+	device := &DeviceInfo{}
+
+	for _, line := range strings.Split(ueventContents, "\n") {
+		parts := strings.Split(line, "=")
```
smira, line 62: https://pkg.go.dev/strings#Cut is a better way here

smira (commented):
can we also add some example output to https://www.talos.dev/v1.0/learn-more/networking-resources/ ?

internal/app/machined/pkg/controllers/network/link_status_test.go
Unix4ever: lol, not sure how that happened 🙃

smira (approved):

PR #5529, opened 2022-05-10
feat: print the status of services during boot

### LINKED ISSUE #5449: print services Talos is waiting to start in the boot sequence
## Feature Request

When Talos is waiting for all services to start, it doesn't say which services it's waiting for.

### Description

We already have a Condition which does this job for service dependencies.

sauterp:
If the a service hangs in the boot sequence will not finish. To inform
users we poll the collection of services in the StartAllServices task
regularly and log if their status has changed.

Fixes #5449

# Pull Request

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5529)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -708,7 +708,26 @@ func StartAllServices(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 		ctx, cancel := context.WithTimeout(ctx, constants.BootTimeout)
 		defer cancel()
 
-		return conditions.WaitForAll(all...).Wait(ctx)
+		aggregateCondition := conditions.WaitForAll(all...)
+
+		errChan := make(chan error)
+
+		go func() {
+			errChan <- aggregateCondition.Wait(ctx)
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -707,7 +707,35 @@ func StartAllServices(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
...
+			select {
+			case err := <-errChan:
+				return err
+			case <-ticker.C:
+			}
+
+			latestMessage := aggregateCondition.String()
+
+			if latestMessage != lastMessage {
+				logger.Printf("%s", latestMessage)
```
smira, line 734: as an idea we discussed during the call: should we repeat this message every N seconds (say, 15 seconds), even if the message hasn't changed?

as there are many messages scrolling through dmesg at early boot, it might be easier to spot the right message this way

sauterp: makes sense, that way the user will also see if a service hangs, since it never disappears from the list. 15s sounds good, I'll try that

smira (approved):

PR #5535, opened 2022-05-11
fix: load kubelet system service in StartAllServices task

### LINKED ISSUE #5527: waiting for all services task sometimes skips the `kubelet` service
## Bug Report

The problem is that the task `StartAllServices` lists all services in the service runner, while the `services.Kubelet` service is added to the list asynchronously by the `KubeletServiceController`, so depending on a boot order, it might be that the kubelet service is already in the list or not.

### Description

Steps to reproduce:

1. Boot Talos.
2. Make `kubelet` fail on start with `talosctl edit mc`, add e.g. `extraArgs: \nkey: value` to the `machine.kubelet` section to make kubelet fail on start due to unsupported argument.
3. Reboot Talos.
4. Talos either correctly blocks on `StartAllServices` task and doesn't proceed further, or it might skip it through and get to the next task.

Suggested fix is to load `services.Kubelet` into the service runner in the `StartAllServices` task, as `Load`ing twice is no-op.

### Environment

- Platform:

sauterp:
# Pull Request

## Description

We found a race condition in the StartAllServices task, where the task
would start, before kubelet was loaded as a system service. It's
possible that the task completes without the kubelet service becoming "up". As a fix
we load the kubelet service again in StartAllServices.

Fixes #5527

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5535)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -672,6 +672,7 @@ func StartAllServices(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 
 		svcs.Load(
 			&services.CRI{},
+			&services.Kubelet{},
 		)
 
 		switch t := r.Config().Machine().Type(); t {


smira (approved):
:+1:

PR #5542, opened 2022-05-11
fix: ignore failures to dial wireguard client

smira:
This is required when running Talos in a container when host OS kernel
doesn't have support for Wireguard.

The failure is deferred so that controller actually fails when it needs
the wireguard client, but if wireguard is not used, controller continues
running.

This should fix errors when running Talos on Docker/WSL:

```
[talos] 2022/05/11 08:52:28 controller failed {"component": "controller-runtime", "controller": "network.LinkStatusController", "error": "error creating wireguard client: setsockopt: protocol not available"}
[talos] 2022/05/11 08:50:48 controller failed {"component": "controller-runtime", "controller": "kubespan.ManagerController", "error": "error creating wireguard client: setsockopt: protocol not available"}
[talos] 2022/05/11 08:50:32 controller failed {"component": "controller-runtime", "controller": "network.LinkSpecController", "error": "error creating wireguard client: setsockopt: protocol not available"}
```

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5542)

### DIFF
--- a/internal/app/machined/pkg/controllers/kubespan/manager.go
+++ b/internal/app/machined/pkg/controllers/kubespan/manager.go
@@ -133,12 +133,13 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 		ctrl.PeerReconcileInterval = DefaultPeerReconcileInterval
 	}
 
-	wgClient, err := ctrl.WireguardClientFactory()
-	if err != nil {
-		return fmt.Errorf("error creating wireguard client: %w", err)
-	}
+	var wgClient WireguardClient
 
-	defer wgClient.Close() //nolint:errcheck
+	defer func() {
+		if wgClient != nil {
+			wgClient.Close() //nolint:errcheck
+		}
+	}()
 
 	var rulesMgr RulesManager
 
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/kubespan/manager.go
```diff
@@ -207,6 +208,13 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			continue
 		}
 
+		if wgClient == nil {
```
frezbo, line 211: isn;t it always nil?

frezbo (approved):

internal/app/machined/pkg/controllers/kubespan/manager.go
smira: this is inside a `for` loop, so it should be `nil` only the first time it enters the loop, and on subsequent loops it will be already initialized

we're simply deferring initialization to the moment we need `wgCllient` for real: e.g. if KubeSpan is not enabled, no need even to create it

frezbo: the github diff was confusing. Thanks :+1:

PR #5544, opened 2022-05-11
fix: reset certificate SANs on update

smira:
This affects both API server and Talos API cert SANs.

Before the fix, SANs accumulated changes over time, so even if the
hostname changes, old hostname is still kept in SANs. Even though it
shouldn't be a problem in general, it is confusing as after reboot list
will be reset back to expected value.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5544)

### DIFF
--- a/internal/app/machined/pkg/controllers/secrets/kubernetes_cert_sans_test.go
+++ b/internal/app/machined/pkg/controllers/secrets/kubernetes_cert_sans_test.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"net"
 	"net/url"
+	"reflect"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/controllers/secrets/kubernetes_cert_sans_test.go
```diff
@@ -137,6 +140,54 @@ func (suite *KubernetesCertSANsSuite) TestReconcile() {
...
+					"foo.example.com",
+					"kubernetes",
+					"kubernetes.default",
+					"kubernetes.default.svc",
+					"kubernetes.default.svc.cluster.remote",
+					"localhost",
+					"some.other.url",
+				}
+
+				if !reflect.DeepEqual(spec.DNSNames, expectedDNSNames) {
```
frezbo, line 183: is this checking if it's reset?

smira: it's checking that if we remove some name from the input list, it will disappear from the final generated list of cert SANs.

we mutate the input and verify that output is correct

that's a regression test - if we remove .Reset() call, the test will fail

frezbo (approved):

Unix4ever (approved):

PR #5402, opened 2022-04-20
chore: extension spec update

frezbo:
Extension spec update

Signed-off-by: Noel Georgi <git@frezbo.dev>

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5402)

PR #5550, opened 2022-05-12
feat: extend the extension service spec with container security options

### LINKED ISSUE #5411: extend extension service spec to configure container security
## Feature Request

See #5402 as an example (this should be configurable)

We might also want to keep read-only rootfs configurable as well.

sauterp:
# Pull Request

We extend the extension service spec with three security options,
WithWriteableSysfs, WithMaskedPaths, WithReadonlyPaths

Fixes #5411

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5550)


frezbo (commented):
this also needs an update to the extension service docs

internal/app/machined/pkg/system/services/extension_test.go
```diff
@@ -0,0 +1,35 @@
...
+				},
+			},
+		},
+	}
+
+	// when
+	actual := svc.GetOCIOptions()
+
+	// then
+	assert.Len(t, actual, 9)
```
frezbo, line 34: isn't it better to explicitly validate the options, if in the future an option got added and removed, this would still pass

internal/app/machined/pkg/system/services/extension_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// TestGetOCIOptions ensures that no oci option spec is created for the options MaskedPaths and ReadonlyPaths if they are set to nil. This is backward compatible behavior.
+func TestGetOCIOptions(t *testing.T) {
+	// given
+	svc := &services.Extension{
+		Spec: &extservices.Spec{
+			Container: extservices.Container{
+				Security: extservices.Security{
+					MaskedPaths:   nil,
```
frezbo, line 23: I don't think we want masked paths to be nil by default, since containerd sets up default maskedpaths, so i think we need  it the other way around, @smira did I understand this this right?

frezbo (changes requested):

internal/app/machined/pkg/system/services/extension.go
```diff
@@ -95,6 +95,35 @@ func (svc *Extension) DependsOn(r runtime.Runtime) []string {
 	return deps
 }
 
+// GetOCIOptions gets all OCI options from an Extension.
+func (svc *Extension) GetOCIOptions() []oci.SpecOpts {
```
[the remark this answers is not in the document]
sauterp: I tried that, but I wrote a unit test for it in the "services" package. then when I ran the linter it said, I need to use "services_test". But when I do that, I loose access to the unexported methods of services.

internal/app/machined/pkg/system/services/extension_test.go
sauterp: The options are function references, I couldn't find a way to validate them.

smira: I think you can't compare them, as these are functions. But what we could do is to convert them options back to the OCI spec and assert on the OCI spec itself - whether some fields hold default value or an overridden value.

In fact, we can do a pretty extensive unit-test here, as we don't have any other tests.

internal/app/machined/pkg/system/services/extension_test.go
sauterp: I tried to emulate the case where the user didn't specify any maskedPaths in the yaml spec. In that case MaskedPaths here will be nil and as a result we want to fall back to the default MaskedPaths of containerd like you said. That's why I want to test, that the MaskedPaths option doesn't get added to the list.

frezbo: so how i do set empty masked paths like https://github.com/siderolabs/talos/pull/5402/files#diff-58056f2700d6a6e74200d7eaf5e8e2466f0eb2f368a4a1d540efb1a357fa6c1fR145, becuase by default talos won't set them and just using

```yaml
security:
  maskedPaths: []
```

feels like it's confusing. Maybe `withEmptyMaskedPaths`, I'm not sure which is better

smira: I think specifying empty `maskedPaths: []` is fine, we just need to document that default value is non-empty and OCI default

internal/app/machined/pkg/system/services/extension.go
```diff
@@ -95,6 +95,35 @@ func (svc *Extension) DependsOn(r runtime.Runtime) []string {
...
+	ociOpts := []oci.SpecOpts{
+		oci.WithRootFSPath(filepath.Join(constants.ExtensionServicesRootfsPath, svc.Spec.Name)),
+		oci.WithRootFSReadonly(),
```
smira, line 102: can we also make this option configurable please?

so that we have `.container.security.writeableRootfs`: by default we enable `oci.WirthRootfsReadonly()`, but if configured, we skip this option

smira (commented):
:+1: we just need to update the docs: https://www.talos.dev/v1.1/advanced/extension-services/

sauterp:
> +1 we just need to update the docs: https://www.talos.dev/v1.1/advanced/extension-services/
done

smira (approved):

Unix4ever (approved):

website/content/v1.1/advanced/extension-services.md
```diff
@@ -72,6 +72,21 @@ The section `mounts` uses the standard OCI spec:
...
+```yaml
+maskedPaths:
+  - "/should/be/masked"
+readonlyPaths:
+  - "/path/that/should/be/readonly"
+  - "/another/readonly/path"
+writeableRootfs: true
+```
+
+The rootfs is readonly by default unless `writeableRootfs: true` and masked paths will be mounted to /dev/null.
```
frezbo, line 88: Are masked paths are mounted to `/dev/null` (just a question)

frezbo: Thanks, :+1: , TIL

PR #5563, opened 2022-05-12
feat: implement DeepCopy using code generation

DmitriyMV:
Refactor every typed.Resource except KubeletConfigSpec to use deep-copy tool for generating DeepCopy method.

KubeletConfigSpec is excluded because its DeepCopy method is not trivial.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5563)


Unix4ever (approved):
Will just throw in an idea that it would also be cool to generate some unit test check for each type that will verify deepcopy creates an identical object.
Just to ensure we don't introduce resources for which deepcopy won't work. But maybe that's an overkill :man_shrugging:

DmitriyMV:
I'll think about it.

pkg/machinery/resources/cluster/affiliate.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
 )
 
+//go:generate deep-copy -type AffiliateSpec -type IdentitySpec -type MemberSpec -type ConfigSpec -header-file ../../../../hack/boilerplate.txt -o deep_copy.generated.go .
```
smira, line 16: I feel this commend should go to `cluster.go` as it has nothing to do with the `Affiliate`, but rather for all types in this directory. 

I know I'm late to the party, but still ;)

And it make sense to sort types by name to make it easier to see which one is there, and which one should be added

PR #5587, opened 2022-05-16
feat: add 'etcd membership is consistent across nodes' health check

### LINKED ISSUE #5552: cluster health check: etcd membership is consistent across nodes
Pull the list of etcd members from each control plane node, list of members should be identical: hostname, ID, flags.

CACCPT code: https://github.com/siderolabs/cluster-api-control-plane-provider-talos/blob/master/controllers/etcd.go#L19

DmitriyMV:
Add new health check which waits for all etcd members. Closes #5552.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5587)

### DIFF
--- a/pkg/cluster/check/etcd.go
+++ b/pkg/cluster/check/etcd.go
@@ -1,0 +1,97 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package check
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+
+	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
+	"github.com/talos-systems/talos/pkg/machinery/client"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+)
+
+// EtcdConsistentAssertion checks that etcd membership is consistent across nodes.
+//nolint:gocyclo
+func EtcdConsistentAssertion(ctx context.Context, cluster ClusterInfo) error {
+	cli, err := cluster.Client()
+	if err != nil {
+		return err
+	}
+
+	nodes := append(cluster.NodesByType(machine.TypeInit), cluster.NodesByType(machine.TypeControlPlane)...)
+	nodesCtx := client.WithNodes(ctx, nodes...)
+
+	resp, err := cli.EtcdMemberList(nodesCtx, &machineapi.EtcdMemberListRequest{})
+	if err != nil {
+		return err
+	}
+
+	type data struct {
+		hostname  string
+		id        uint64
+		isLearner bool
+	}
+
+	knownMembers := map[data]struct{}{}
+
+	messages := resp.GetMessages()
+	if len(messages) == 0 {
+		return errors.New("no messages returned")
+	}
+
+	sort.Slice(messages, func(i, j int) bool {
+		return messages[i].GetMetadata().GetHostname() < messages[j].GetMetadata().GetHostname()
+	})
+
+	for i, message := range messages {
+		if i == 0 {
+			// Fill data using first message
+			for _, member := range message.Members {
+				knownMembers[data{member.Hostname, member.Id, member.IsLearner}] = struct{}{}
+			}
+
+			continue
+		}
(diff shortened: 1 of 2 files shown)


pkg/cluster/check/etcd.go
```diff
@@ -0,0 +1,63 @@
...
+
+// EtcdConsistentAssertion checks that etcd membership is consistent across nodes.
+//nolint:gocyclo
+func EtcdConsistentAssertion(ctx context.Context, cluster ClusterInfo) error {
+	cli, err := cluster.Client()
+	if err != nil {
+		return err
+	}
+
+	resp, err := cli.EtcdMemberList(ctx, &machineapi.EtcdMemberListRequest{})
```
smira, line 22: this way we're talking to a random node in the cluster.

what we really want to do is to talk to all control plane nodes: `client.WithNodes(ctx, cluster.NodesByType(machine.TypeInit, machine.TypeControlPlane)...)`

this way we get responses from all control plane nodes

then each response line is a single member coming from one of the nodes, node info which delivered the response will be in message.Metadata, we should build a list for each node and they should match

smira: this code might help: https://github.com/siderolabs/talos/blob/5ec4e9096e98c8bf5e27b1a78171498dbc94b23c/cmd/talosctl/cmd/talos/etcd.go#L95-L129

it's more complicated than we need (it supports talking without node context, we don't need it here, in our case member.Hostname should be always set to the node we were talking to)

smira: and yep, we should get a message back from each node we requested with `client.WithNodes`

pkg/cluster/check/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+
+	type data struct {
+		hostname  string
+		id        uint64
+		isLearner bool
+	}
+
+	knownMembers := map[data]struct{}{}
+
+	for i, message := range resp.GetMessages() {
```
smira, line 39: it makes sense to sort `Messages` based on the `Metadata.Hostname`, so that error reported is consistent, otherwise it prints different error each poll, as the first message is different each time

pkg/cluster/check/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+		actualMembers := len(message.Members)
+
+		if actualMembers != expectedMembers {
+			return fmt.Errorf("%s: expected to have %d members, got %d", node, expectedMembers, actualMembers)
+		}
+
+		// check that member list is the same on all nodes
+		for _, member := range message.Members {
+			if _, found := knownMembers[data{member.Hostname, member.Id, member.IsLearner}]; !found {
+				return fmt.Errorf("%s: found extra etcd member %s", node, member.Hostname)
```
smira, line 63:
```suggestion
				return fmt.Errorf("%s: found unexpected etcd member %s", node, member.Hostname)
```

pkg/cluster/check/etcd.go
DmitriyMV: Done!

pkg/cluster/check/etcd.go
```diff
@@ -0,0 +1,93 @@
...
+			if _, found := knownMembers[data{member.Hostname, member.Id, member.IsLearner}]; !found {
+				return fmt.Errorf("%s: found unexpected etcd member %s", node, member.Hostname)
+			}
+		}
+	}
+
+	return nil
+}
+
+func mapCollect[M ~map[K]V, Z any, K comparable, V any](m M, fn func(K, V) Z) []Z {
```
smira, line 79: moar generic! :rofl:

smira (approved):

pkg/cluster/check/etcd.go
```diff
@@ -0,0 +1,99 @@
...
+	if len(messages) == 0 {
+		return errors.New("no messages returned")
+	}
+
+	sort.Slice(messages, func(i, j int) bool {
+		return messages[i].GetMetadata().GetHostname() < messages[j].GetMetadata().GetHostname()
+	})
+
+	for i, message := range messages {
+		fmt.Println("hostname", message.GetMetadata().GetHostname())
```
smira, line 52: remove me! it's breaking the checks

smira (changes requested):

smira (approved):

PR #5598, opened 2022-05-18
chore: go-smbios updates

frezbo:
go-smbios library updates

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5598)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,6 +52,7 @@ require (
 	github.com/gdamore/tcell/v2 v2.5.1
 	github.com/gizak/termui/v3 v3.1.0
 	github.com/godbus/dbus/v5 v5.1.0
+	github.com/golang/mock v1.6.0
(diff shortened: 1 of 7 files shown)


pkg/machinery/go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/talos-systems/crypto v0.3.5
 	github.com/talos-systems/go-blockdevice v0.3.2-0.20220405134045-d9c3a2738861
 	github.com/talos-systems/go-debug v0.2.1
+	github.com/talos-systems/go-smbios v0.2.0
```
smira, line 24: I believe it's not used in `machinery`, so we don't need it here

frezbo: yup, it was from the new files i added, not part of this commit

PR #5603, opened 2022-05-19
chore: add go.work

frezbo:
Add go.work to work with `talos` and `pkg/machinery`

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5603)


smira (approved):
good thing is that it won't be pulled into the actual build in `Dockerfile`, so it's pure for local development

PR #5626, opened 2022-05-24
fix: improve error message when creating cluster

### LINKED ISSUE #5625: Empty talosconfig confuses `talosctl`
### Description

"Cleaning up" talosconfig file by emptying it (something like `echo > ~/.talos/config`) will cause subsequent commands to throw unhelpful errors as the talosconfig file exists but without the expected yaml structure.

We should be able to catch this and at least throw a better error.

### Logs

```shell
$ sudo -E _out/talosctl-linux-amd64 cluster create  --name dev --provisioner=qemu --registry-mirror docker.io=http://10.5.0.1:5000 --registry-mirror k8s.gcr.io=http://10.5.0.1:5001 --registry-mirror quay.io=http://10.5.0.1:5002 --registry-mirror gcr.io=http://10.5.0.1:5003 --registry-mirror ghcr.io=http://10.5.0.1:5004 --registry-mirror 10.5.0.1:5005=http://10.5.0.1:5005 --install-image=127.0.0.1:5005/siderolabs/installer:v1.1.0-alpha.2-24-g1156daac2 --with-bootloader=false --extra-uefi-search-paths=/usr/share/ovmf/x64
generating PKI and tokens
created controlplane.yaml
created worker.yaml
creating state directory in "/home/tim/.talos/clusters/dev"
creating network dev
creating load balancer
creating dhcpd
creating master nodes
creating worker nodes
EOF
```

TimJones:
Add extra context to error message when unable to properly
open the talos config file when creating a cluster.

Closes #5625 

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5626)

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -698,7 +698,7 @@ func postCreate(ctx context.Context, clusterAccess *access.Adapter) error {
 func saveConfig(talosConfigObj *clientconfig.Config) (err error) {
 	c, err := clientconfig.Open(talosconfig)
 	if err != nil {
-		return err
+		return fmt.Errorf("error opening talos config: %s", err)
 	}
 
 	renames := c.Merge(talosConfigObj)


frezbo (approved):

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -698,7 +698,7 @@ func postCreate(ctx context.Context, clusterAccess *access.Adapter) error {
 func saveConfig(talosConfigObj *clientconfig.Config) (err error) {
 	c, err := clientconfig.Open(talosconfig)
 	if err != nil {
-		return err
```
smira, line 701: @TimJones we missed it, but @DmitriyMV noticed it: `%s` should be `%w` to allow proper wrapping

TimJones: @smira @DmitriyMV Thanks, fixed in #5629

PR #5638, opened 2022-05-25
feat: include azure public IPs from standard LB SKU into extips

### LINKED ISSUE #4150: [azure] support Standard SKU public IP discovery
## Feature Request

It's a long story, but to use Talos on Azure with the Azure CCM, we need to use Standard SKU public IPs.  Unfortunately, standard SKU public IPs do not get reported in instance metadata.  That means we have to discovery them in some other way.

Two ways that I see exist:
 - use the `netdiscover` method which we use with kubespan
 - query the Azure API (hopefully this doesn't mean importing their entire Go library in to Talos!)

Doc mentioning lack of metadata support for Standard SKU IPs: https://docs.microsoft.com/en-us/azure/virtual-network/public-ip-addresses#standard

sergelogvinov:
Add public IP from Azure Standard SKU LB to the list of external IP (for
proper certificate generation).

part of #4150

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5638)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/talos-systems/go-procfs/procfs"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
```diff
@@ -241,9 +275,30 @@ func (a *Azure) NetworkConfiguration(ctx context.Context, ch chan<- *runtime.Pla
...
+
+	lbConfig, err := download.Download(ctx, AzureLoadbalancerEndpoint,
+		download.WithHeaders(map[string]string{"Metadata": "true"}),
+		download.WithErrorOnNotFound(errors.ErrNoHostname),
+		download.WithErrorOnEmptyResponse(errors.ErrNoHostname))
+	if err != nil && !stderrors.Is(err, errors.ErrNoHostname) {
 		return err
 	}
 
+	if err = json.Unmarshal(lbConfig, &loadBalancerAddresses); err != nil {
```
smira, line 293: if we fail to download, should we skip the unmarshal?

sergelogvinov: Yep, I added more details on error.

smira (approved):

sergelogvinov:
typo fix )

PR #5645, opened 2022-05-26
feat: enable version API in maintenance mode

smira:
This is useful to find Talos version as it got booted (e.g. to generate
proper machine configuration).

There's a security concern that version API might return sensitive
information via public API. At the same time Talos version can be
guessed by looking at the output of other APIs, e.g. resource type list
(`talosctl get rd`), which changes with every minor version.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5645)

### DIFF
--- a/cmd/talosctl/cmd/talos/version.go
+++ b/cmd/talosctl/cmd/talos/version.go
@@ -24,6 +24,7 @@ var versionCmdFlags struct {
 	clientOnly   bool
 	shortVersion bool
 	json         bool
+	insecure     bool
 }
 
 // versionCmd represents the `talosctl version` command.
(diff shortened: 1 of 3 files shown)


frezbo (approved):
do we need to expose this for the cli, just only on the server side is enough right?

smira:
> do we need to expose this for the cli, just only on the server side is enough right?

server side is enough, but I think we should also let others use it?

frezbo:
> > do we need to expose this for the cli, just only on the server side is enough right?
> 
> server side is enough, but I think we should also let others use it?

I was thinking this was only going to be exposed over the siderolink, so normal api would still need a valid auth.

I've conflicting thoughts on exposing version, on one hand it's good, but on another if a version info could leak information about an security vuln. At least for now this is fine I believe, unless we have a very compelling need to hide the version info.

smira:
Reworked the PR to limit maintenance mode to SideroLink IPs only.

Unix4ever (approved):

PR #5651, opened 2022-05-26
fix: unmarshal HardwareAddr without stdlib help

smira:
Stdlib `net.ParseMAC` does lots of validations, but some hardware addrs
we can see (on logical interfaces) are not valid, so parse MACs in a
simple way.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5651)

### DIFF
--- a/pkg/machinery/nethelpers/hwaddr.go
+++ b/pkg/machinery/nethelpers/hwaddr.go
@@ -4,7 +4,11 @@
 
 package nethelpers
 
-import "net"
+import (
+	"bytes"
+	"encoding/hex"
+	"net"
+)
 
 // HardwareAddr wraps net.HardwareAddr for YAML marshaling.
 type HardwareAddr net.HardwareAddr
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

pkg/machinery/nethelpers/hwaddr.go
```diff
@@ -16,12 +20,17 @@ func (addr HardwareAddr) MarshalText() ([]byte, error) {
 
 // UnmarshalText implements text.Unmarshaler interface.
 func (addr *HardwareAddr) UnmarshalText(b []byte) error {
-	mac, err := net.ParseMAC(string(b))
+	rawHex := bytes.ReplaceAll(b, []byte(":"), []byte(""))
```
frezbo, line 23: should we add a test to make sure we didn't break anything

frezbo (approved):

pkg/machinery/nethelpers/hwaddr.go
smira: we already have a test for it (in previous PR): we marshal/unmarshal `LinkStatus` and compare

Issue #3272, opened 2021-03-08
config URLs should have variable substitution


Ulexus:
## Feature Request

We should support some limited variable substitution for our `talos.config` URLs.  This would allow more dynamic deployments with simpler bootloader requirements. 

For instance:

```
  talos.config=https://config.mydomain.com/?mac=${MAC}&uuid=${UUID}
```

Potential variable substitutions:
 - MAC address (of interface from which config is being requested... or maybe always eth0)
 - DMI-sourced serial number
 - UUID
 - hostname (if this is supplied by the DHCP server, for instance)

rgl:
As-of talos 1.0.0, it seems machined is already trying to parse the `talos.config` kernel argument. At least, in my metal setup, I'm configuring the argument as `http://pandora.talos.test/generic?mac=08-00-27-00-00-00` (which is handled by matchbox) and `machined` seems to complain about it in the logs:

```
[talos] fetching machine config from: "http://pandora.talos.test/generic?mac=08-00-27-00-00-00"
[talos] unsupported query parameter: "mac"
```

I was not expecting that talos would do any interpretation of the URL. If this is really intended, please:

1. provide a escaping mechanism, so I can prevent talos from touching the URL.
2. document this behavior in, at least, the [`talos.config` kernel argument documentation page](https://www.talos.dev/v1.0/reference/kernel/#talosconfig).

smira:
iPXE URL example:

```
http://{{ .Endpoint }}:{{ .Port }}/ipxe?uuid=${uuid}&mac=${net${idx}/mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial}&arch=${buildarch}
```

sauterp:
To obtain the Hostname and MAC address, we need to process resources that might not be available yet or there might be an error. Therefore the substitution can fail. Should this fail the boot sequence?

smira:
Re-opening #5771 as there's one more change we need to make in a subsequent change: we need to try to substitute variables on each download attempt, so that if the node hostname changes for example, we use the new hostname in the new download attempt. This requires a bit of restructuring of the retry in the download function.

PR #5661, opened 2022-05-28
feat: add varible substitution to talos.config

djcrabhat:
Signed-off-by: djcrabhat <djcrabhat@sosimplerecords.com>

# Pull Request

Passing the `talos.config` url through a templating engine.

## What? (description)

Implementing #3272 

## Why? (reasoning)

By doing token substitution on the `talos.config` downloadUrl with [text/template](https://pkg.go.dev/text/template), talos can provide can provide mode runtime information to a machine config serving endpoint, like a MAC address in a querystring or URL path part to check against a pre-approved list of interfaces.  As @Ulexus put it, "This would allow more dynamic deployments with simpler bootloader requirements."

## Acceptance


internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -94,6 +116,45 @@ func PopulateURLParameters(downloadURL string, getSystemUUID func() (string, err
 	return u.String(), nil
 }
 
+func getMacAddr() (string, error) {
+	ifen0, en0Err := net.InterfaceByName("en0")
```
frezbo, line 120: is `en0` the first interface name assigned by kernel?

djcrabhat: Oh, that's a great point.  I must admit, I thought of this pretty naively, just developing on my macbook, not even considering of the context it'd actually be running in on the metal.

I _believe_ the en0 to be the modern convention for the first logical ethernet device.  Great catch, I should've written a e2e test to cover this so actually ensure this works as expected.

smira:
thanks for your PR, we got a bit different implementation in #5771, and we're going to do one more pass on it before the relase

PR #5680, opened 2022-06-01
fix: wait for `/var` to be mounted in kubelet service controller

smira:
This is a cosmetic fix: when `KubeletServiceController` tries to write
files to `/etc/kubernetes` before `/var` mounted, it would fail.
Controller will be restarted, but each restart involves a backoff on
each restart which gets longer with each restart.

On the first boot, or when EPHEMERAL is encrypted, mounting might take
considerable time (seconds), so during that time controller might enter
such long backoff timeout that it will delay whole boot sequence - it
won't finish before `kubelet` is started.

By waiting for `EPHEMERAL` to be mounted before starting the controller
we eliminate long backoff cycles.

Also fix a bug when `StartAllServices` task might start a kubelet early
(before `KubeletServiceController` is actually going to start it).

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5680)

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -731,19 +731,24 @@ func StartAllServices(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 
 		svcs := system.Services(r)
 
+		// load the kubelet service, but don't start it;
+		// KubeletServiceController will start it once it's ready.
 		svcs.Load(
-			&services.CRI{},
 			&services.Kubelet{},
 		)
 
+		serviceList := []system.Service{
+			&services.CRI{},
+		}
+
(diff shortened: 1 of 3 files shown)


frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -731,19 +731,24 @@ func StartAllServices(seq runtime.Sequence, data interface{}) (runtime.TaskExecu
 
 		svcs := system.Services(r)
 
+		// load the kubelet service, but don't start it;
+		// KubeletServiceController will start it once it's ready.
 		svcs.Load(
-			&services.CRI{},
```
frezbo, line 735: why is the CRI service removed?

frezbo: ohh, nevermind, saw it below

frezbo (approved):

PR #5682, opened 2022-06-03
feat: add 'etcd members should be control plane nodes' health check

utkuozdemir:
Add new health check which checks if the etcd members match control plane nodes. Closes siderolabs#5553.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5682)

### DIFF
--- a/pkg/cluster/check/default.go
+++ b/pkg/cluster/check/default.go
@@ -29,6 +29,13 @@ func DefaultClusterChecks() []ClusterCheck {
 			}, 5*time.Minute, 5*time.Second)
 		},
 
+		// wait for etcd members to be the control plane nodes
+		func(cluster ClusterInfo) conditions.Condition {
+			return conditions.PollingCondition("etcd members to be control plane nodes", func(ctx context.Context) error {
+				return EtcdControlPlaneNodesAssertion(ctx, cluster)
+			}, 5*time.Minute, 5*time.Second)
+		},
+
 		// wait for apid to be ready on all the nodes
 		func(cluster ClusterInfo) conditions.Condition {
 			return conditions.PollingCondition("apid to be ready", func(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


pkg/cluster/check/default.go
```diff
@@ -29,6 +29,13 @@ func DefaultClusterChecks() []ClusterCheck {
 			}, 5*time.Minute, 5*time.Second)
 		},
 
+		// wait for etcd members to be the control plane nodes
+		func(cluster ClusterInfo) conditions.Condition {
+			return conditions.PollingCondition("etcd members to be control plane nodes", func(ctx context.Context) error {
```
frezbo, line 34: `checking for etcd members to be control plane nodes` i think makes more sense?

utkuozdemir: This part of the string is prepended by `waiting for ` by another place so that the final string renders as `waiting for etcd members to be control plane nodes`. See the other checks, they all use the same form.

utkuozdemir: Ah you mean removing the `the`?

frezbo: ahh, that makes sense then, just wanted to confirm

pkg/cluster/check/etcd.go
```diff
@@ -95,3 +133,24 @@ func sliceCollect[S ~[]V, V any, R any](slc S, fn func(V) R) []R {
 
 	return r
 }
+
+func containsAll[S ~[]V, V comparable](input S, check S) bool {
```
smira, line 137: nit: it might be more descriptive to use set semantics here?

(and probably arguments should be swapped, subset << set), this is easier than to understand `containsAll` - what should contain what? :)

```suggestion
func isSubset[S ~[]V, V comparable](set S, subset S) bool {
```

smira (approved):
some nits

PR #5700, opened 2022-06-06
chore: bump dependencies

smira:
dependabot + go-mod-outdated

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5700)


api/resource/resource.proto
```diff
@@ -33,6 +33,7 @@ message Metadata {
   google.protobuf.Timestamp created = 8;
   google.protobuf.Timestamp updated = 9;
   repeated string finalizers = 6;
+  map<string, string> labels = 10;
```
DmitriyMV, line 36: Huh? Should't it be in labels PR?

smira: Talos has its own gRPC API for COSI for historical reasons. I'm not really introducing labels into Talos here, but this change is required to make the code compile (as Metadata proto structure should implement COSI interface).

frezbo: so this change is from the recently added label selectors in COSI, right?

frezbo (approved):

api/resource/resource.proto
smira: yep, as COSI got bumped, it has now labels in COSI gRPC API (vs. Talos gRPC resources API), so it requires Talos Metadata proto to have these labels

Unix4ever (approved):

Unix4ever:
Looks like there are some places which rely on `Spec` being a value, not a pointer.

smira:
> Looks like `LinkSpec` controller is broken as it expects the spec to be value not a pointer.

thanks, yep, fixed now

smira:
fixed one more problem with grpc-proxy and new gRPC library

smira:
and one more issue with `Spec()` being a pointer now, but the code gets better

PR #5701, opened 2022-06-06
chore: fix rook-ceph resource use

TimJones:
Reduce Rook-Ceph cluster resource use to something more
fitting a simple set of integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5701)

### DIFF
--- a/hack/test/day-two/ceph-values.yaml
+++ b/hack/test/day-two/ceph-values.yaml
@@ -1,0 +1,8 @@
+cephClusterSpec:
+  resources:
+    mgr:
+      limits: {}
+      requests: {}
+    osd:
+      limits: {}
+      requests: {}
--- a/hack/test/day-two/config.yaml
+++ b/hack/test/day-two/config.yaml
@@ -31,5 +31,6 @@ charts:
     podSecurityLevel: privileged
     repo: https://charts.rook.io/release
     chart: rook-ceph-cluster
+    valuesPath: hack/test/day-two/ceph-values.yaml
     depends:
       - rook


hack/test/day-two/ceph-values.yaml
```diff
@@ -0,0 +1,16 @@
...
+    mgr:
+      limits:
+        cpu: 500m
+        memory: 512Mi
+      requests:
+        cpu: 250m
+        memory: 256Mi
+    osd:
+      limits:
+        cpu: 1000m
```
frezbo, line 12: i would just remove the limit for cpu in tests, it just throttles the app un-necessarily

frezbo (approved):

smira:
> /promote integration-qemu-day-two

that was probably too early, it allows to promote only when initial pipeline succeeds @TimJones

TimJones:
> > /promote integration-qemu-day-two
> 
> that was probably too early, it allows to promote only when initial pipeline succeeds @TimJones

Thanks @smira, I thought the initial pipeline was green but I must have been looking at the wrong check.

hack/test/day-two/ceph-values.yaml
TimJones: If the tests pass with the limit, I'll re-test without the limit and compare.
Thanks.

hack/test/day-two/ceph-values.yaml
```diff
@@ -0,0 +1,8 @@
+cephClusterSpec:
+  resources:
+    mgr:
+      limits: {}
+      requests: {}
+    osd:
+      limits: {}
+      requests: {}
```
utkuozdemir, line 8: It probably should be like this:

```suggestion
    mgr: {}
    osd: {}
```

TimJones: Unfortunately, the Rook operator threw some errors when the CephCluster CRD was configured that way. I'll try and dig out the logs for you.

PR #5702, opened 2022-06-06
chore: add generic methods and use them 1/n

DmitriyMV:
Things like ToSet, Keys etc...

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5702)

### DIFF
--- a/pkg/machinery/generic/maps/maps.go
+++ b/pkg/machinery/generic/maps/maps.go
@@ -1,0 +1,143 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package maps contains the generic functions for maps.
+package maps
+
+// NOTE(DmitriyMV): I tried to implement this generic functions to be as perfomant as possible.
+// However, I couldn't find a way to do it, since Go (1.18 at the time of writing) cannot inline closures if (generic)
+// function, which accepts the closure, was not inlined itself.
+// And inlining budget of 80 is quite small, since most of it is going towards closure call.
+// Somewhat relevant issue: https://github.com/golang/go/issues/41988
+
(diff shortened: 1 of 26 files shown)


pkg/machinery/generic/maps/maps.go
```diff
@@ -0,0 +1,64 @@
...
+
+	for k, v := range m {
+		r = append(r, fn(k, v))
+	}
+
+	return r
+}
+
+// Map applies the function f to each element of the map and returns a new map with the results.
+func Map[M ~map[K]V, K comparable, V any, K1 comparable, V1 any](m M, fn func(K, V) (K1, V1)) map[K1]V1 {
```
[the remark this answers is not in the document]
utkuozdemir: Also `.Filter()`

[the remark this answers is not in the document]
DmitriyMV: Done, I think?

pkg/machinery/generic/maps/maps.go
```diff
@@ -0,0 +1,109 @@
...
+	for k, v := range m {
+		if f(k, v) {
+			r[k] = v
+		}
+	}
+
+	// No reason to return empty map if there are no elements.
+	if len(r) == 0 {
+		return nil
+	}
```
utkuozdemir, line 90: Is this a common practice in Go? (asking because in Java it was an anti pattern).

Here they don't do it either: https://github.com/samber/lo/blob/master/map.go#L26

smira: it saves a bit of resources, as `nil` is smaller than an empty map, but in other aspects they're equivalent for read operations

utkuozdemir: Caller site will need a nil check, right? Not sure if it's worth it - can be a source of bugs, no?

smira: reading from `nil` map is same as reading from an empty map

utkuozdemir: got it, didn't know, thanks.

PR #5713, opened 2022-06-07
feat(apiserver): disallow anonymous requests by default

rio:
This is inline with CIS guidelines. Otherwise the kube-apiserver will pass along the request with the group
set to `system:unauthenticated`. This will expose anything that is allowed by the `system:public-info-viewer`
and `system:discovery` cluster roles.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5713)

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -300,9 +300,11 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 	}
 
 	builder := argsbuilder.Args{
-		"admission-control-config-file":      filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml"),
-		"advertise-address":                  "$(POD_IP)",
-		"allow-privileged":                   "true",
+		"admission-control-config-file": filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml"),
+		"advertise-address":             "$(POD_IP)",
+		"allow-privileged":              "true",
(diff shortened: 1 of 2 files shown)


smira:
thanks for your PR, I will do a little fixup to make it pass the CI!

smira (approved):

rio:
woops, sorry. Didn't notice you fixing stuff. Hoop that all the force pushing didn't overwrite some stuff.

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -300,9 +300,11 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
...
 
 	builder := argsbuilder.Args{
-		"admission-control-config-file":      filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml"),
-		"advertise-address":                  "$(POD_IP)",
-		"allow-privileged":                   "true",
+		"admission-control-config-file": filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml"),
+		"advertise-address":             "$(POD_IP)",
+		"allow-privileged":              "true",
+		// Do not accept anonymous requests by default. Otherwise the kube-apiserver will set the request's group to system:unauthenticated exposing endpoints like /version etc.
+		"anonymous-auth":                     "false",
```
frezbo, line 307: don't we have a test that validates all default options?

frezbo (approved):

Unix4ever (approved):

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
smira: nope... too many options? :)

PR #5718, opened 2022-06-08
feat: extend node definition used in health checks

utkuozdemir:
Introduce `cluster.NodeInfo` to represent the basic info about a node which can be used in the health checks. This information, where possible, will be populated by the discovery service in following PRs. Part of siderolabs#5554.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5718)

### DIFF
--- a/internal/integration/api/discovery.go
+++ b/internal/integration/api/discovery.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/netip"
 	"strings"
 	"time"
 
@@ -20,6 +21,8 @@ import (
 
 	"github.com/talos-systems/talos/internal/integration/base"
 	"github.com/talos-systems/talos/pkg/machinery/client"
+	"github.com/talos-systems/talos/pkg/machinery/generic/maps"
+	"github.com/talos-systems/talos/pkg/machinery/generic/slices"
 	"github.com/talos-systems/talos/pkg/machinery/resources/cluster"
 	"github.com/talos-systems/talos/pkg/machinery/resources/kubespan"
 )
@@ -43,7 +46,7 @@ func (suite *DiscoverySuite) SetupTest() {
 	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 15*time.Second)
 
 	// check that cluster has discovery enabled
-	node := suite.RandomDiscoveredNode()
+	node := suite.RandomDiscoveredNodeInternalIP()
 	suite.ClearConnectionRefused(suite.ctx, node)
 
 	nodeCtx := client.WithNodes(suite.ctx, node)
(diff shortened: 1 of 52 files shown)


internal/integration/api/discovery.go
```diff
@@ -66,18 +66,18 @@ func (suite *DiscoverySuite) TearDownTest() {
 //
 //nolint:gocyclo
 func (suite *DiscoverySuite) TestMembers() {
-	nodes := suite.DiscoverNodes(suite.ctx)
+	nodes := suite.DiscoverNodeInternalIPs(suite.ctx)
 	expectedTalosVersion := fmt.Sprintf("Talos (%s)", suite.Version)
 
-	for _, node := range nodes.Nodes() {
+	for _, node := range nodes {
```
smira, line 72: we could have used here the full list of IPs and match as a subset of member addresses

pkg/cluster/check/kubernetes.go
```diff
@@ -66,10 +69,7 @@ func K8sAllNodesReportedAssertion(ctx context.Context, cluster ClusterInfo) erro
 		}
 	}
 
-	sort.Strings(expectedNodes)
-	sort.Strings(actualNodes)
-
-	if reflect.DeepEqual(expectedNodes, actualNodes) {
+	if isSubset(allExpectedNodeIPs, actualNodes) {
```
smira, line 72: this feels like it should be changed, as we match on any expected address of the node, while later we assert that expected IP [0] was actually matched

pkg/cluster/check/kubernetes.go
```diff
@@ -79,13 +79,17 @@ func K8sAllNodesReportedAssertion(ctx context.Context, cluster ClusterInfo) erro
...
+func K8sFullControlPlaneAssertion(ctx context.Context, cl ClusterInfo) error {
+	clientset, err := cl.K8sClient(ctx)
 	if err != nil {
 		return err
 	}
 
-	expectedNodes := append(cluster.NodesByType(machine.TypeInit), cluster.NodesByType(machine.TypeControlPlane)...)
+	expectedNodes := append(cl.NodesByType(machine.TypeInit), cl.NodesByType(machine.TypeControlPlane)...)
+	expectedNodeIPs := sliceCollect(expectedNodes, func(v cluster.NodeInfo) string {
+		return v.IPs[0].String()
```
smira, line 90: should we collect all addresses here as well as verify that `actualNodes` is a subset of `allExpectedNodeIPs` ?

in fact what we want to verify that every `Node` matches some `NodeInfo`, I wonder if that can be extracted? comparing list of addresses is a bit more weak check than matching `Node` and `NodeInfo`.

smira (commented):
lgtm, with some minor notes!

pkg/cluster/check/kubernetes.go
utkuozdemir: I introduced a function, `assertNodes` which does assertion by comparing expectedNodes<>actualNodes, asserting that they are the same set of nodes (by their internalips) + for each node, asserting that the actual nodes IPs are a **subset** of the expected one. Can you have a look into that?

pkg/cluster/check/kubernetes.go
utkuozdemir: I introduced a function, `assertNodes` which does assertion by comparing expectedNodes<>actualNodes, asserting that they are the same set of nodes (by their internalips) + for each node, asserting that the actual nodes IPs are a **subset** of the expected one. Can you have a look into that?

smira (approved):
:cool:

PR #5744, opened 2022-06-13
feat: add ctest package and base for test suite

DmitriyMV:
This change adds ctest package which adds DefaultSuite and helper functions.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5744)


Unix4ever (approved):
:cool: Would be great to have it somewhere in the separate module to use it in Arges :)

utkuozdemir (approved):
LGTM

DmitriyMV:
@Unix4ever 

> Would be great to have it somewhere in the separate module to use it in Arges

The problem is that it uses `logging.Wrap` which is provided by `github.com/talos-systems/talos/pkg/logging`. Extracting the logging package seems unnecessary. Providing the callback to create logging instance will force us to use that callback everywhere. We could use `generic type factory` trick we use for Resources but I don't know if its a good idea to continue our abuse of generics.

PR #5746, opened 2022-06-13
wip: feat: use discovery service on cluster health checks

utkuozdemir:
Query the discovery service to fetch the node list and use the results in health checks. Closes siderolabs#5554.


pkg/cluster/check/discovery.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// NewDiscoveredClusterInfo returns a new cluster.Info populated from the discovery service.
+func NewDiscoveredClusterInfo(members []clussterres.Member) (cluster.Info, error) {
+	m, err := membersToNodeInfoMap(members)
+	if err != nil {
+		return nil, err
+	}
+
+	nodes := slices.FlatMap(maps.Values(m), identity[[]cluster.NodeInfo])
```
DmitriyMV, line 42: I think we can use lambda here?

utkuozdemir: you mean just inlining it as func?

DmitriyMV: Yup

utkuozdemir: Done (wish go had arrow syntax e.g. `t -> t`)

internal/app/machined/internal/server/v1alpha1/v1alpha1_cluster.go
```diff
@@ -118,46 +108,154 @@ func (cl *clusterState) NodesByType(t machine.Type) []cluster.NodeInfo {
...
-		var allNodeInfos []cluster.NodeInfo
-		allNodeInfos = append(allNodeInfos, controlPlaneNodeInfos...)
-		allNodeInfos = append(allNodeInfos, workerNodeInfos...)
+	result := make([]clusterres.Member, list.Len())
 
-		cl.nodeInfos = allNodeInfos
-		cl.nodeInfosByType = map[machine.Type][]cluster.NodeInfo{
-			machine.TypeControlPlane: controlPlaneNodeInfos,
-			machine.TypeWorker:       workerNodeInfos,
+	for i := 0; i < list.Len(); i++ {
```
DmitriyMV, line 237: I think it would be better to use iterator here.

internal/app/machined/internal/server/v1alpha1/v1alpha1_cluster.go
```diff
@@ -118,46 +108,154 @@ func (cl *clusterState) NodesByType(t machine.Type) []cluster.NodeInfo {
...
+
+	list, err := safe.StateList[*clusterres.Member](ctx, res, resource.NewMetadata(clusterres.NamespaceName, clusterres.MemberType, "", resource.VersionUndefined))
+	if err != nil {
+		return nil, err
+	}
 
-		var allNodeInfos []cluster.NodeInfo
-		allNodeInfos = append(allNodeInfos, controlPlaneNodeInfos...)
-		allNodeInfos = append(allNodeInfos, workerNodeInfos...)
+	result := make([]clusterres.Member, list.Len())
```
DmitriyMV, line 235: Preallocate and use append below maybe?

internal/app/machined/internal/server/v1alpha1/v1alpha1_cluster.go
```diff
@@ -118,46 +108,154 @@ func (cl *clusterState) NodesByType(t machine.Type) []cluster.NodeInfo {
...
+}
+
+func k8sNodeToNodeInfo(node *corev1.Node) (*cluster.NodeInfo, error) {
+	if node == nil {
+		return nil, nil
+	}
+
+	var internalIP netip.Addr
+
+	var ips []netip.Addr
```
DmitriyMV, line 200: Maybe we can preallocate here using `len(node.Status.Address)`?

pkg/cluster/check/discovery.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+
+		result[machineType] = append(result[machineType], *nodeInfo)
+	}
+
+	return result, nil
+}
+
+func memberToNodeInfo(member *clussterres.Member) (*cluster.NodeInfo, error) {
+	ips, err := stringsToNetipAddrs(slices.Map(member.TypedSpec().Addresses, netaddrIPToString))
```
DmitriyMV, line 70: Maybe ambda?

pkg/cluster/check/discovery.go
```diff
@@ -0,0 +1,98 @@
...
+		IPs:        ips,
+	}, nil
+}
+
+func netaddrIPToString(ip netaddr.IP) string {
+	return ip.String()
+}
+
+func stringsToNetipAddrs(ips []string) ([]netip.Addr, error) {
+	result := make([]netip.Addr, len(ips))
```
DmitriyMV, line 86:
```suggestion
	result := make([]netip.Addr, 0, len(ips))
```
?

pkg/cluster/check/discovery.go
```diff
@@ -0,0 +1,98 @@
...
+
+	for i, ip := range ips {
+		parsed, err := netip.ParseAddr(ip)
+		if err != nil {
+			return nil, err
+		}
+
+		result[i] = parsed
```
DmitriyMV, line 94:
```suggestion
		result = append(result, parsed)
```
?

utkuozdemir: what's the benefit of doing this vs the way it currently is?

DmitriyMV: Because you are not updating values in slices - you are creating new ones. In theory compiler can optimize `preallocate + append` by not zeroing memory during allocation stage. It's also more clearly translates your intentions. And gives you less chances to mess up with indexes, which is one of the [two fundamental CS problems](https://twitter.com/codinghorror/status/506010907021828096) :D

DmitriyMV (approved):

PR #5747, opened 2022-06-14
fix: enable orderly poweroff in hyper-v on Azure

### LINKED ISSUE #5641: Azure shutdown process
### Description

Talos does not shutdown properly. 

hv_utils module receive shutdown  event from hyper-v. 
It call the function [orderly_poweroff](https://github.com/torvalds/linux/blob/master/drivers/hv/hv_util.c#L165-L168) 
which runs [/sbin/poweroff](https://github.com/torvalds/linux/blob/master/kernel/reboot.c#L420) in user space.

Talos does not have this command. Thats why shutdown process failed. ```reboot: Failed to start orderly shutdown: forcing the issue```

### Logs

```
172.16.136.6: kern:    info: [2022-05-25T13:07:50.681476988Z]: hv_utils: Shutdown request received - graceful shutdown initiated
172.16.136.6: kern: warning: [2022-05-25T13:07:50.689541988Z]: reboot: Failed to start orderly shutdown: forcing the issue
172.16.136.6: kern:    info: [2022-05-25T13:07:50.698726988Z]: ACPI: PM: Preparing to enter system sleep state S5
172.16.136.6: kern:   emerg: [2022-05-25T13:07:50.709230988Z]: reboot: Power down
```

### Environment

- Talos version: master branch
- Platform: azure

sauterp:
Previously Talos would not shutdown gracefully if hyper-v issued the
'perform_shutdown' call. Said call would execute '/sbin/poweroff' which
did not exist in Talos. We hardlink machined to '/sbin/poweroff' and
make it send a shutdown API call to PID 1 machined.

I verified that my solution works by creating a Talos node on Azure and stopping it through the Azure Portal.

Fixes #5641

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5747)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -419,6 +419,9 @@ COPY --from=pkg-util-linux-amd64 /lib/libmount.* /rootfs/lib/
 COPY --from=pkg-kmod-amd64 /usr/lib/libkmod.* /rootfs/lib/
 COPY --from=pkg-kernel-amd64 /lib/modules /rootfs/lib/modules
 COPY --from=machined-build-amd64 /machined /rootfs/sbin/init
+# the orderly_poweroff call by the kernel will call '/sbin/poweroff'
+RUN ln /rootfs/sbin/init /rootfs/sbin/poweroff
+RUN chmod +x /rootfs/sbin/poweroff
 # NB: We run the cleanup step before creating extra directories, files, and
 # symlinks to avoid accidentally cleaning them up.
 COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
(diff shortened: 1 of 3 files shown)


internal/app/poweroff/main.go
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
frezbo, line 1: we probably need a way to test this in QEMU if possible

sauterp: I tested it in QEMU by modifying the shutdown call like this:
https://github.com/siderolabs/talos/commit/c5384c46a9c04d594a54686712762edde09d361a

(obviously this is not in the PR)

Dockerfile
```diff
@@ -419,6 +419,8 @@ COPY --from=pkg-util-linux-amd64 /lib/libmount.* /rootfs/lib/
 COPY --from=pkg-kmod-amd64 /usr/lib/libkmod.* /rootfs/lib/
 COPY --from=pkg-kernel-amd64 /lib/modules /rootfs/lib/modules
 COPY --from=machined-build-amd64 /machined /rootfs/sbin/init
+RUN ln /rootfs/sbin/init /rootfs/sbin/poweroff
```
utkuozdemir, line 422: Should we add a comment above this line to briefly explain why we do this?

Unix4ever (approved):

PR #5758, opened 2022-06-17
feat: add support for configuring network bridges

utkuozdemir:
Allows configuring bridge interfaces (similar to bonds). Closes siderolabs#5414.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5758)


internal/app/machined/pkg/controllers/network/link_spec.go
```diff
@@ -392,6 +392,73 @@ func (ctrl *LinkSpecController) syncLink(ctx context.Context, r controller.Runti
...
+					return fmt.Errorf("error encoding bridge attributes for %q: %w", link.TypedSpec().Name, err)
+				}
+
+				// bring bridge down
+				if err = conn.Link.Set(&rtnetlink.LinkMessage{
+					Family: existing.Family,
+					Type:   existing.Type,
+					Index:  existing.Index,
+					Flags:  0,
+					Change: unix.IFF_UP,
```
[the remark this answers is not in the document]
utkuozdemir: I took it as-is from the bond code but now that I look into it a bit, here's what it does: 
Change (`Change`) the flag `unix.IFF_UP` to the value of `Flags` which is `0`. Setting `Flags: 1` would bring it up. I tested it and verified.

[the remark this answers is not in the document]
frezbo: awesome, that explains it, weird Unix flags :sweat_smile:

frezbo:
also need to do a `make docs`

utkuozdemir:
> also need to do a `make docs`

Aha missed that one, done now.

frezbo (approved):
:tada:

internal/app/machined/pkg/adapters/network/bridge_master_spec.go
```diff
@@ -0,0 +1,63 @@
...
+// BridgeMasterSpec adapter provides encoding/decoding to netlink structures.
+//
+//nolint:revive,golint
+func BridgeMasterSpec(r *network.BridgeMasterSpec) bridgeMaster {
+	return bridgeMaster{
+		BridgeMasterSpec: r,
+	}
+}
+
+type bridgeMaster struct {
```
Unix4ever, line 23: Any reason to not make it public and make `revive` linter happy here?

Unix4ever (approved):

internal/app/machined/pkg/adapters/network/bridge_master_spec.go
utkuozdemir: Just did it. There wasn't any reason other than it being c/p from `bondSpec` :). Don't know what was the reason for @smira to do it that way on bonds.

internal/app/machined/pkg/controllers/network/link_spec.go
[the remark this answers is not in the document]
smira: afaik `Change` is a mask on `Flags`, so it indicates which bits of `Flags` should be updated: we only touch `IFF_UP` flag, but keep other flags intact

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2041,6 +2050,16 @@ type Bond struct {
...
 }
 
+// Bridge contains the various options for configuring a bridge interface.
+type Bridge struct {
+	//   description: The interfaces that make up the bond.
+	BridgedInterfaces []string `yaml:"interfaces"`
+	//   description: |
+	//     A bridge option.
+	//     Please see the official kernel documentation.
+	BridgeSTPEnabled bool `yaml:"stpEnabled,omitempty"`
```
smira, line 2060: it looks like `STP` might have many more options so, it might make sense to structure the config accordingly:

```yaml
bridge:
  stp:
     enabled: true
```

so we could add more settings in the future

also it feels like STP should be enabled by default, so I would rather do:

```yaml
bridge:
  stp:
     disabled: false
```

smira: (we can still keep more natural logic in `provider.go` interfaces, this is just for YAML zero value being sane default)

smira: or another, more natural option be `Enabled *bool` with `nil` being `true`

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -523,6 +523,15 @@ func (d *Device) Bond() config.Bond {
 	return d.DeviceBond
 }
 
+// Bridge implements the MachineNetwork interface.
+func (d *Device) Bridge() config.Bridge {
```
smira, line 527: looks like we don't have any validation in `v1alpha1_validation.go` with matching tests for it

we basically we need to do validation as discussed in the original ticket: e.g. same link can't be enslaved twice to a bridge and a bond, etc.

smira: or that one shouldn't assign addresses to bridges, etc.

pkg/machinery/resources/network/link.go
```diff
@@ -50,6 +50,11 @@ type BondMasterSpec struct {
 	PeerNotifyDelay uint32                        `yaml:"peerNotifyDelay,omitempty"`
 }
 
+// BridgeMasterSpec describes bridge settings if Kind == "bridge".
+type BridgeMasterSpec struct {
+	STPEnabled bool `yaml:"stpEnabled,omitempty"`
```
smira, line 55: same as in  the machine config, we might better structure it as `STP STPSpec ...` so that we can add more STP settings

smira (commented):
this is really great work, I left some comments after doing a single pass, I need to look more into `link_config.go` (note for myself)

internal/app/machined/pkg/adapters/network/bridge_master_spec.go
smira: I did it this way to have a better looking API: I don't need to make this type public, as it has no use outside of this package. The only public API is the function and the value it returns has a couple of methods.

Linters sometimes don't see the full picture. Yes, it's weird in general to return private types, but in this specific case it makes sense, as it's an adapter which extends a resource with additional methods.

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
utkuozdemir: I changed it to be the `Enabled *bool` as you suggested.

pkg/machinery/resources/network/link.go
utkuozdemir: Done.

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
utkuozdemir: I added some validations and tests for them. Feel free to have another look. 

I didn't add

> one shouldn't assign addresses to bridges

because it is a valid use case.

utkuozdemir:
> also need to update release notes

This won't make it to 1.1 release, so I guess I have to wait till we release 1.1.0 then rebase & update release notes. (also as a self-note)

@smira @frezbo pls correct me if I'm wrong on this or feel free to suggest alternatives.

smira:
> > also need to update release notes
> 
> This won't make it to 1.1 release, so I guess I have to wait till we release 1.1.0 then rebase & update release notes. (also as a self-note)
> 
> @smira @frezbo pls correct me if I'm wrong on this or feel free to suggest alternatives.

no need to wait... you can update `hack/release.toml`, I will clean it up before the alpha release. Stable 1.1.0 release uses notes from the `release-1.1` branch

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -875,6 +884,29 @@ func (b *Bond) PeerNotifyDelay() uint32 {
 	return b.BondPeerNotifyDelay
 }
 
+// Enabled implements the config.STP interface.
+func (s *STP) Enabled() bool {
+	if s.STPEnabled == nil {
```
smira, line 889: we can also check for `nil` here, and simplify places where we read this value:

```suggestion
	if s == nil || s.STPEnabled == nil {
```

so that we don't have to check for `STP()` being nil, and do `device.STP().Enabled()` always

utkuozdemir: makes sense, changed

smira (approved):
:cool:

frezbo (approved):

PR #5771, opened 2022-06-20
feat: add support for variable substitution in talos.config kernel parameter

### LINKED ISSUE #3272: config URLs should have variable substitution

sauterp:
# Pull Request

The URL to fetch the configuration for a talos node is given by the
talos.config kernel parameter. We add support for 4 variables ${uuid},
${serial}, ${mac} and ${hostname} which substitute the device UUID,
DMI-sourced serial number, MAC address of the first network interface to
be up and the hostname respectively.

Fixes #3272

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5771)


internal/app/machined/pkg/controllers/network/platform_config_test.go
```diff
@@ -616,7 +616,7 @@ func (mock *platformMock) Name() string {
 	return "mock"
 }
 
-func (mock *platformMock) Configuration(context.Context) ([]byte, error) {
+func (mock *platformMock) Configuration(context.Context, state.State) ([]byte, error) {
```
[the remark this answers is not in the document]
sauterp: yea, I try not to break existing APIs but in this case I didn't find another way, plus the function is only used in exactly one place.

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal_test.go
```diff
@@ -48,19 +94,252 @@ func TestPopulateURLParameters(t *testing.T) {
...
 				assert.EqualError(t, err, tt.expectedError)
 			} else {
-				assert.Equal(t, output, tt.expectedURL)
+				assert.Equal(t, tt.expectedURL, output)
 			}
 		})
 	}
 }
+
+type mockState struct {
```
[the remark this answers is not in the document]
sauterp: I used a mock because I wanted to test the behaviour in [getResource](https://github.com/siderolabs/talos/blob/ffe94876ed7f6c629ea8e9ec5441bbbade972712/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go#L169-L201), where it first calls state.State.List which returns nothing and later calls state.State.WatchKind. If I populate a real state.State I couldn't simulate this. Or maybe I didn't find a way. Is there an example where we do something similar?

smira:
* in cosi-project/runtime, add safe.WatchKind, and probably safe.Watch (?)
* unify smbios.New() calls to use cached version (via sync.Once) in hardware info controller
* expose smbios SystemInformation as a resource in hardware controller (UUID, Serial Number, Manufacturer, ...)
* rework this PR to consume all variable substitution values as resources from the state
* don't mock the state
* use state.WithBootstrapContents to WatchKind to avoid doing `List`, filter out events which are not relevant (only handle created/updated events)
* use safe.WatchKind wrapper and generics to eliminate type assertions
* parse URLs and replace fields in the parsed representation
* parse variable placeholders and match variable names case insensitive way
* treat `uuid=` arg as `uuid=${uuid}` to keep backwards compatibility

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -64,28 +72,77 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
 	if err != nil {
-		return "", fmt.Errorf("failed to parse %s: %w", constants.KernelParamConfig, err)
+		return "", fmt.Errorf("failed to parse %s: %w", populatedURL, err)
 	}
 
 	values := u.Query()
 
+	substitute := func(key string, getFunc func(ctx context.Context, r state.State) (string, error)) error {
+		for qKey, qValues := range values {
+			if len(qValues) > 0 {
```
smira, line 96: to reduce nesting level we can do `if len(qValues) == 0 { continue }` and un-nest the block below

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -64,28 +72,77 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
+				qVal := qValues[0]
+
+				varOfKey := keyToVar(key)
```
smira, line 99: these three assignments can be pushed out of the loop

in fact, it almost seems like there might be a benefit of declaring a replacement as a small struct which will initialize itself once and has a method to replace itself in the string, fetching its value

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -64,28 +72,77 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
+				varOfKeyExpr := regexp.QuoteMeta(varOfKey)
+				matcher := regexp.MustCompile(`(?i)` + varOfKeyExpr)
+
+				if matcher.MatchString(qVal) {
```
smira, line 103: similar to the above, probably it's easier to read if we `continue` if it doesn't match

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -64,28 +72,77 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
+					val, err := getFunc(context.Background(), r)
```
smira, line 104: should use `ctx` here!

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -64,28 +72,77 @@ func (m *Metal) Configuration(ctx context.Context) ([]byte, error) {
...
+	if err := substitute(constants.HostnameKey, getHostname); err != nil {
+		return "", err
+	}
+
+	// Although the UUID can be substituted via the ${uuid} variable, we keep the behavior that the empty uuid= parameter is filled in with the UUID. This is backwards compatible.
 	for key, qValues := range values {
-		switch key {
-		case "uuid":
-			// don't touch uuid field if it already has some value
+		if key == constants.UUIDKey { // don't touch uuid field if it already has some value
```
smira, line 137: I keep thinking we can handle this in the function above as a special case for `uuid=` - if it exists and empty, treat it as `uuid=${UUID}`. it looks strange we tread it different way?

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -94,24 +151,101 @@ func PopulateURLParameters(downloadURL string, getSystemUUID func() (string, err
...
+			if val != "" {
+				return val, nil
+			}
+		}
+	}
+}
+
+func getAndCheckUUID(r *hardwareResource.SystemInformation) string {
+	sysInfo := r.TypedSpec()
+	if sysInfo != nil {
```
smira, line 192: I don't think Spec can ever be `nil`, pointer is returned to allow modifications, but resource hold spec by value

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal_test.go
```diff
@@ -5,18 +5,68 @@
...
+	"github.com/talos-systems/talos/pkg/machinery/nethelpers"
+	"github.com/talos-systems/talos/pkg/machinery/resources/hardware"
+	"github.com/talos-systems/talos/pkg/machinery/resources/network"
 )
 
+func generateMAC() string {
+	const macLen = 6
+
+	bytes := make([]byte, macLen)
+	if _, err := rand.Read(bytes); err != nil {
```
smira, line 32: I know it's a test, but doing `rand.Read()` from `math/rand` might set a bad precedent for secure random data.

I don't know if we really want a random one each time? or a fixed value would be sufficient?

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/export_test.go
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package metal
+
+var (
+	GetMacAddress = getMACAddress
```
smira, line 8: do we still need these?

pkg/machinery/resources/hardware/system_information.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// SystemInformationType is type of SystemInformation resource.
+const SystemInformationType = resource.Type("SystemInformation.hardware.talos.dev")
+
+// SystemInformation resource holds node SystemInformation information.
+type SystemInformation = typed.Resource[SystemInformationSpec, SystemInformationRD]
+
+// SystemInformationSpec represents the system information obtained from smbios.
+type SystemInformationSpec struct {
```
smira, line 20: we should register new resource in the `v1alpha2_state.go` to make it work with `talosctl get`

internal/pkg/encryption/keys/nodeid.go
```diff
@@ -7,7 +7,7 @@ package keys
 import (
 	"fmt"
 
-	"github.com/talos-systems/go-smbios/smbios"
+	"github.com/talos-systems/talos/internal/app/machined/pkg/controllers/hardware"
```
smira, line 10: this should be a linter error - we import `internal/` package from public `pkg/`.

In fact I think we need to move `GetSMBIOSInfo()` function to some other shared package.

smira (commented):
some small things, but we also need:

* docs - we have a document on kernel arguments
* hack/release.toml note

website/content/v1.2/reference/kernel.md
```diff
@@ -83,6 +83,25 @@ Several of these are enforced by the Kernel Self Protection Project [KSPP](https
...
+  
+  The following example
+
+  `http://example.com/metadata?h=${hostname}&m=${mac}&s=${serial}&u=${uuid}`
+  
+  may translate to
+  
+  `http://example.com/metadata?h=myTestHostname&m=52%3A2f%3Afd%3Adf%3Afc%3Ac0&s=0OCZJ19N65&u=40dcbd19-3b10-444e-bfff-aaee44a51fda`
+  
+  For backwards compatibility we have a insert the UUID into the query parameter `uuid` if its value is empty. As in
```
frezbo, line 102:
```suggestion
  For backwards compatibility we insert the system UUID into the query parameter `uuid` if its value is empty. As in
```

frezbo (approved):

PR #5775, opened 2022-06-20
test: fix health endpoint cli test when discovery is disabled

utkuozdemir:
We skip the client-side health endpoint test that relies on the discovery service if the discovery service is not enabled for the cluster. Related to siderolabs#5554.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5775)

### DIFF
--- a/internal/integration/cli/health.go
+++ b/internal/integration/cli/health.go
@@ -12,8 +12,11 @@ import (
 	"regexp"
 	"strings"
 
+	"gopkg.in/yaml.v3"
+
 	"github.com/talos-systems/talos/internal/integration/base"
 	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+	"github.com/talos-systems/talos/pkg/machinery/resources/cluster"
 )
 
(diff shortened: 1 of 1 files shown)


frezbo (approved):

internal/integration/cli/health.go
```diff
@@ -79,6 +82,13 @@ func (suite *HealthSuite) TestClientSideWithExplicitNodes() {
 //
 //nolint:gocyclo
 func (suite *HealthSuite) TestClientSideWithDiscovery() {
+	discoveryEnabled, err := suite.isDiscoveryEnabled()
+	suite.Require().NoError(err)
+
+	if !discoveryEnabled {
+		suite.T().Skip()
```
smira, line 89: one small nit here: we can use `Skipf()` with some meaningful message to indicate why we skipped to make it more obvious to someone looking at the test output

smira (approved):

internal/integration/cli/health.go
utkuozdemir: makes sense, done.

PR #5776, opened 2022-06-20
test: fix CLI nodes discovery without provisioner data

smira:
When integration tests run without data from Talos provisioner (e.g.
against AWS/GCP), it should work only with `talosconfig` as an input.

This specific flow was missing filling out `infoWrapper` properly.

Clean up things a bit by reducing code duplication.


frezbo (approved):

utkuozdemir (approved):

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -461,14 +460,9 @@ func (e *Etcd) argsForInit(ctx context.Context, r runtime.Runtime) error {
...
 
-	p, err := platform.CurrentPlatform()
-	if err != nil {
-		return err
-	}
-
 	var upgraded bool
 
-	if p.Mode() != runtime.ModeContainer {
+	if r.State().Platform().Mode() != runtime.ModeContainer {
```
smira, line 465: it's a small unrelated change which got in, we don't need to build platform once again, as it's available via runtime state

PR #5792, opened 2022-06-22
fix: siderlink api assume port 443 with https schema

TimJones:
If no port is supplied for the SideroLink API endpoint and the https
schema is used, then assume port 443 is wanted.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5792)


sauterp (approved):
lgtm

smira (approved):

PR #5804, opened 2022-06-25
feat: delete etcd member by id

### LINKED ISSUE #4896: Etcd delete members by ID
## Feature Request

Currently you can remove the bad member by ```hostname```. But if some thing wrong happened in join stage. The etcd cluster does not known the member name yet. And you cannot delete it from the cluster.

Test case: add the new etcd member with different version (older).

### Description

Try to delete node first by name, and then by id

```shell
talosctl etcd remove-member <hostname/member-id>
```

sergelogvinov:
It helps delete the member with the empty hostname (uninitialized member).

close #4896

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5804)

### DIFF
--- a/cmd/talosctl/cmd/talos/etcd.go
+++ b/cmd/talosctl/cmd/talos/etcd.go
@@ -44,7 +44,7 @@ var etcdLeaveCmd = &cobra.Command{
 }
 
 var etcdMemberRemoveCmd = &cobra.Command{
-	Use:   "remove-member <hostname>",
+	Use:   "remove-member <hostname/id>",
 	Short: "Remove the node from etcd cluster",
 	Long: `Use this command only if you want to remove a member which is in broken state.
 If there is no access to the node, or the node can't access etcd to call etcd leave.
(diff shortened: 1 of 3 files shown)


smira:
I like the change, but I wonder if we should make it more explicit on the API side - have two params in the API for the node name and member ID? And in the CLI we could do the check to guess whether it's an ID or node name.

smira:
> > I like the change, but I wonder if we should make it more explicit on the API side - have two params in the API for the node name and member ID? And in the CLI we could do the check to guess whether it's an ID or node name.
> 
> Very good idea. But I can't understand how to make it. The node name can be [a-z][0-9] which converts to the integer(id) as well.
> 
> Idea: receive the member list, when compare the strings and define the type of input?

we could also provide specific response when member is not found for example, and use that to fall back to member IDs. 

or we could have a flag for member IDs.

sergelogvinov:
solution https://github.com/siderolabs/talos/pull/6417

PR #5806, opened 2022-06-25
fix: etcd peer sans list

sergelogvinov:
Before generate the peer certificate, we will check IP existence on the node.

## Why? (reasoning)

The node does not have all IP addresses when we create Etcd certificate. In this case, peer cert does not have an advertised IP.

## Acceptance

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5806)

### DIFF
--- a/internal/app/machined/pkg/controllers/secrets/api.go
+++ b/internal/app/machined/pkg/controllers/secrets/api.go
@@ -129,7 +129,7 @@ func (ctrl *APIController) Run(ctx context.Context, r controller.Runtime, logger
 	}
 }
 
-//nolint:gocyclo,cyclop
+//nolint:gocyclo,cyclop,dupl
 func (ctrl *APIController) reconcile(ctx context.Context, r controller.Runtime, logger *zap.Logger, isControlplane bool) error {
 	inputs := []controller.Input{
 		{
--- a/internal/app/machined/pkg/controllers/secrets/etcd.go
+++ b/internal/app/machined/pkg/controllers/secrets/etcd.go
@@ -7,6 +7,7 @@ package secrets
 import (
 	"context"
 	"fmt"
+	stdlibnet "net"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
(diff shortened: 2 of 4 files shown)


frezbo:
isn't this already fixed by https://github.com/siderolabs/talos/pull/5135

sergelogvinov:
Hello, thank for your reply. 

Unfortunately, isn't work for me. I have two interfaces (public/private). The second one appears a little late (more than 1 second). I use etcd.subnet to check IP existence on the node. If not - observe the node IPs again.
In dmesg logs, I sow that Talos tried to create a cert 3 times after this changes (_no address matched the provided subnet_). 

And I've removed the DNS record from **peer** certs because Talos uses only IPs to create member list (p2p communication). Etcd won't check reverse DNS lookup (https://etcd.io/docs/v3.3/op-guide/security/#notes-for-tls-authentication https://github.com/etcd-io/etcd/pull/7767). And any other IP from the certificate which not used at all. Peer SAN needs only one IP - advertised IP. 

Plus - an unauthorized client wouldn't know all node IPs. 

The server etcd certificate (for client connections) more complicated - I kept it as is.

smira:
I understand the issue, but I feel the fix should be more complicated. 

As I see it, peer certificate should only contain IPs from the etcd subnet?

sergelogvinov:
> I understand the issue, but I feel the fix should be more complicated.
> 
> As I see it, peer certificate should only contain IPs from the etcd subnet?

Yep, only one IP - advertised ip, and sometimes i think about 127.0.0.1/::1 should be removed too...

smira:
I hope with #6012 merge in, the etcd issues should be resolved in a bit different way.

* etcd certs still include all addresses, and they are updated on the fly as etcd is able to reload them.
* Talos waits for the configured etcd subnet to be available as an address on the node before starting etcd.

smira:
@sergelogvinov please let us know if you see any improvement with `master`, we should also cut next `alpha` today or tomorrow.

sergelogvinov:
Thank you.

Peer cert has my second IP address now.
```
# talosctl --talosconfig _cfgs/talosconfig --nodes 172.16.0.11 version                                                           
Client:
        Tag:         v1.1.0
        SHA:         d55a1871
        Built:       
        Go version:  go1.18.3
        OS/Arch:     darwin/arm64
Server:
        NODE:        172.16.0.11
        Tag:         v1.2.0-alpha.2
        SHA:         be351dcb
        Built:       
        Go version:  go1.19
        OS/Arch:     linux/amd6

# openssl s_client -showcerts -connect XXX.XXX.XXX.XXX:2380 |  openssl x509 -tex | grep 172.16.0.11
# echo $?                                                                                        
0
```

smira:
awesome, more fixes are coming for listen address as well, and advertised subnets, it should be even better soon!

PR #5818, opened 2022-06-27
fix: filtering of static pods and optimize fetching

utkuozdemir:
When we query kubelet API to populate the StaticPodStatuses, instead of checking for ownerReferences to be empty, we check the annotation "kubernetes.io/config.source" value so we avoid including standalone pods (that are regular pods but not part of a replicaset).

We also optimize their fetching by avoiding to unmarshal the fields we do not need.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5818)


smira (commented):
lgtm, but I think we should remove those TODO comments

utkuozdemir:
yep, messed it up

utkuozdemir:
Correct one is https://github.com/siderolabs/talos/pull/5819

PR #5819, opened 2022-06-27
fix: filter static pods correctly and optimize fetching

utkuozdemir:
When we query kubelet API to populate the StaticPodStatuses, instead of checking for ownerReferences to be empty, we check the annotation "kubernetes.io/config.source" value so we avoid including standalone pods (that are regular pods but not part of a replicaset).

We also optimize their fetching by avoiding to unmarshal the fields we do not need.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5819)

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_static_pod_controller.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_static_pod_controller.go
@@ -317,13 +317,13 @@ func (ctrl *KubeletStaticPodController) refreshPodStatus(ctx context.Context, r 
 	podsSeen := map[string]struct{}{}
 
 	for _, pod := range podList.Items {
-		if pod.OwnerReferences != nil {
+		if pod.Metadata.Annotations.ConfigSource != "file" {
 			continue
 		}
 
 		pod := pod
 
-		statusID := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+		statusID := fmt.Sprintf("%s/%s", pod.Metadata.Namespace, pod.Metadata.Name)
 
 		podsSeen[statusID] = struct{}{}
 
--- a/pkg/kubernetes/kubelet/kubelet.go
+++ b/pkg/kubernetes/kubelet/kubelet.go
@@ -7,6 +7,7 @@ package kubelet
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


pkg/kubernetes/kubelet/kubelet.go
```diff
@@ -65,10 +66,43 @@ func NewClient(nodename string, clientCert, clientKey, caPEM []byte) (*Client, e
...
+	Namespace   string      `json:"namespace"`
+	Annotations Annotations `json:"annotations"`
+}
 
-	return &podList, err
+// Annotations are the annotations on a pod.
+type Annotations struct {
+	// ConfigMapSource indicates where the resource is coming from.
+	// Its value is "file" for static pods and "api" for resources came from kube-apiserver.
+	ConfigSource string `json:"kubernetes.io/config.source"`
```
frezbo, line 107: I have not seen this annotation when listing pods, is this something only returned when querying the raw `/pods` api?

frezbo: nevrmind, i just queried the `/api/v1/pods` api and I see it

internal/app/machined/pkg/controllers/k8s/kubelet_static_pod_controller.go
```diff
@@ -317,13 +317,13 @@ func (ctrl *KubeletStaticPodController) refreshPodStatus(ctx context.Context, r
 	podsSeen := map[string]struct{}{}
 
 	for _, pod := range podList.Items {
-		if pod.OwnerReferences != nil {
+		if pod.Metadata.Annotations.ConfigSource != "file" {
```
frezbo, line 320: do we need to check if annotations is `nil` (I'm not sure if in some weird case it's not populated?)

utkuozdemir: well, it seems it'll at least have:
```
        "annotations": {
          "kubernetes.io/config.seen": "2022-06-27T14:22:09.108356948Z",
          "kubernetes.io/config.source": "api"
        },
```

but we can check just to be on the safe side.

frezbo (approved):

smira (approved):

PR #5824, opened 2022-06-28
docs: recommend setting "host" Processor Type on proxmox

sauterp:
Talos v1.1 is unable to boot in a Proxmox VM with "kvm64" set as its
Processor Type. We add an instruction in the Proxmox Guide for setting
Processor Type to "host". Talos v1.1 requires the x86-64-v2
microarchitecture, which is very common today.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5824)


smira:
It actually started with v1.0, so we should update docs for v1.0/v1.1 and v1.2 (future)

frezbo (approved):

PR #5837, opened 2022-06-29
feat: implement strategic merge machine config patching

smira:
This format is much easier to understand when compared to JSON patches,
it allows for more patch validation, and it should provide better user
experience.

This just implements the config merge, but it doesn't yet hook it up to
any CLI utility, so no user-facing docs.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5837)


pkg/machinery/config/merge/merge_test.go
```diff
@@ -0,0 +1,254 @@
...
+			left: Config{},
+			right: Config{
+				Slice: []Struct{},
+			},
+			expected: Config{
+				Slice: []Struct{},
+			},
+		},
+		{
+			name: "custom slice",
```
Unix4ever, line 235: Should we also have a case when `left` is not settable?

Unix4ever (approved):
:cool:

pkg/machinery/config/merge/merge_test.go
smira: I don't think it's possible actually with the structs we have... it's like merging into an `int`, for structs/maps/slices, we always have a way to set the left-hand part. I added the checks mostly to avoid panics, but it should never happen under normal operations

PR #5842, opened 2022-06-30
refactor: use pointer types more in machine config structs

smira:
There should be no functional change with this PR.

The primary driver is supporting stragetic merge configuration patches.
For such type of patches machine config should be loaded from incomplete
fragments, so it becomes critically important to distinguish between a
field having zero value vs. field being set in YAML.

E.g. with following struct:

```go
struct { AEnabled *bool `yaml:"a"` }
```

It's possible to distinguish between:

```yaml
a: false
```

and no metion of `a` in YAML.

Merging process trewats zero values as "not set" (skips them when
merging), so it's important to allow overriding value to explicit
`false`.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5842)


frezbo (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_apiserverconfig.go
```diff
@@ -40,7 +40,7 @@ func (a *APIServerConfig) Env() Env {
 
 // DisablePodSecurityPolicy implements the config.APIServer interface.
 func (a *APIServerConfig) DisablePodSecurityPolicy() bool {
-	return a.DisablePodSecurityPolicyConfig
+	return a.DisablePodSecurityPolicyConfig == nil || *a.DisablePodSecurityPolicyConfig
```
smira, line 43: wrong!

pkg/machinery/config/types/v1alpha1/v1alpha1_discoveryconfig.go
```diff
@@ -10,12 +10,12 @@ import (
 )
 
 // Enabled implements the config.ClusterDiscovery interface.
-func (c ClusterDiscoveryConfig) Enabled() bool {
-	return c.DiscoveryEnabled
+func (c *ClusterDiscoveryConfig) Enabled() bool {
+	return c.DiscoveryEnabled != nil && *c.DiscoveryEnabled
```
Unix4ever, line 14: Would suggest using a helper function like:

```
func IsTrue(value *bool) bool {
  return value != nil && *value
}
```

smira: we could even do a generic function for `*T`... I will try that... or we even had one

smira: https://github.com/siderolabs/go-pointer/blob/master/pointer.go#L13

smira: updated

Unix4ever (approved):
:cool:

Issue #5849, opened 2022-07-01
disable hardware.SystemInfoController when running talos in container mode


frezbo:
disable hardware.SystemInfoController when running talos in container mode

PR #5850, opened 2022-07-01
chore: disable systeminfo controller in container

frezbo:
Disable systeminfo controller in container mode

Fixes: #5849

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5850)

### DIFF
--- a/internal/app/machined/pkg/controllers/hardware/system.go
+++ b/internal/app/machined/pkg/controllers/hardware/system.go
@@ -15,13 +15,15 @@ import (
 	"go.uber.org/zap"
 
 	hwadapter "github.com/talos-systems/talos/internal/app/machined/pkg/adapters/hardware"
+	runtimetalos "github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
 	pkgSMBIOS "github.com/talos-systems/talos/internal/pkg/smbios"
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -62,48 +64,51 @@ func (ctrl *SystemInfoController) Run(ctx context.Context, r controller.Runtime,
...
-			return err
-		}
-
-		ctrl.SMBIOS = s
-	}
-
-	const systemInfoID = "systeminformation"
-	if err := r.Modify(ctx, hardware.NewSystemInformation(systemInfoID), func(res resource.Resource) error {
-		hwadapter.SystemInformation(res.(*hardware.SystemInformation)).Update(&ctrl.SMBIOS.SystemInformation)
+	if ctrl.V1Alpha1Mode != runtimetalos.ModeContainer {
```
smira, line 67: we can make it dead simple - if `ctrl.V1Alpha1Mode == runtimetalos.ModeContainer { return nil }` in the beginning of the method. Controller will stop and it won't be restarted.

frezbo: fixed :+1:

Unix4ever (approved):

PR #5869, opened 2022-07-06
feat: more circular pkg from internal to public pkg

DmitriyMV:
This PR moves circular package from internal packages to the public pkg folder as the package is stable enough.

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5869)


pkg/circular/circular_test.go
```diff
@@ -17,15 +17,15 @@ import (
 	"github.com/stretchr/testify/suite"
 	"golang.org/x/time/rate"
 
-	"github.com/talos-systems/talos/internal/pkg/circular"
+	circular2 "github.com/talos-systems/talos/pkg/circular"
```
smira, line 20: wth? :)

smira (approved):

pkg/circular/circular_test.go
DmitriyMV: Refactoring didn't work as good as I expected... One instance of it I fixed, but other slipped.

PR #5875, opened 2022-07-08
fix: use correct etcd cert path

smira:
Remove code which is no longer needed.

Taken from #5806

Co-authored-by: Serge Logvinov <serge.logvinov@sinextra.dev>

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5875)

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -242,12 +242,6 @@ func (e *Etcd) HealthSettings(runtime.Runtime) *health.Settings {
 
 //nolint:gocyclo
 func generatePKI(ctx context.Context, r runtime.Runtime) (err error) {
-	// remove legacy etcd PKI directory to handle upgrades with `--preserve` to Talos 0.12
-	// TODO: remove me in Talos 0.13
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -574,8 +568,8 @@ func (e *Etcd) argsForControlPlane(ctx context.Context, r runtime.Runtime) error
 		"listen-peer-urls":                   "https://" + net.FormatAddress(listenAddress) + ":2380",
 		"listen-client-urls":                 "https://" + net.FormatAddress(listenAddress) + ":2379",
 		"client-cert-auth":                   "true",
-		"cert-file":                          constants.KubernetesEtcdPeerCert,
```
frezbo, line 577: why were we using the peer cert instead of the actual cert here?

frezbo (approved):

internal/app/machined/pkg/system/services/etcd.go
smira: it was a bug, but in fact peer and server certs were generated with same settings, so nobody noticed.. we have same CA for server and peer connections, so they just work

PR #5876, opened 2022-07-08
fix: stabilize etcd join and promote sequences

smira:
There were two issues with using discovery service for join and promote:

* on join, that resulted in joining too fast which triggers race bugs in
  etcd cert generation (to be fixed as separate PR)
* on promote, Talos has to connect to non-learner member of the cluster
  which is somehow "automatic" with Kuberentes discovery, as it only
  lists `kube-apiserver` running which is up only when etcd on the same
  node is healthy. etcd client doesn't allow to avoid learner members,
  as even getting a member list from a learner doesn't work (to be fixed
  as a separate PR)

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5876)

### DIFF
--- a/internal/pkg/etcd/etcd.go
+++ b/internal/pkg/etcd/etcd.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"math/rand"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Unix4ever (commented):
`s/stablizie/stabilize` in the commit name :)

Unix4ever (approved):

sauterp (approved):

internal/pkg/etcd/etcd.go
```diff
@@ -96,6 +113,9 @@ func NewClientFromControlPlaneIPs(ctx context.Context, resources state.State) (c
 		endpoints[i] = net.FormatAddress(endpoints[i]) + ":2379"
 	}
 
+	// Shuffle endpoints to establish random order on each call to avoid patterns based on sorted IP list.
+	rand.Shuffle(len(endpoints), func(i, j int) { endpoints[i], endpoints[j] = endpoints[j], endpoints[i] })
```
frezbo, line 117: :+1:

frezbo (approved):

PR #5877, opened 2022-07-08
fix: correct SANs for etcd certs

smira:
I would like to rewrite whole cert generation process, but for now a few
fixes:

* client cert doesn't need any SANs
* peer cert should contain only non-localhost SANs
* server cert same as before (localhost + addresses)

See https://etcd.io/docs/v3.5/op-guide/security/ for details.


smira:
> * server cert same as before (localhost + addresses)

we need localhost to access etcd on the same node from Talos, also `kube-apiserver` is configured to talk to localhost always

frezbo:
> See https://etcd.io/docs/v3.5/op-guide/security/ for details.

reminded me, should we also start enabled rbac for talos/k8s access to etcd?

frezbo (approved):

frezbo:
also we don;t have tests for these?

smira:
> also we don;t have tests for these?

yep, only integration ones... I plan to rewrite this properly as a controller next, I just wanted to test new settings with old code first.

PR #5879, opened 2022-07-08
feat: bump k8s to v0.25.0-beta.0

frezbo:
Bump k8s to v0.25.0-beta.0

---
This change is [Reviewable](https://reviewable.io/reviews/siderolabs/talos/5879)


frezbo:
Updated:
- k8s.io/component-base path update - https://github.com/kubernetes/kubernetes/pull/105797

Skipped changes:
- renaming of `k8s.gcr.io` to `registry.k8s.io` is skipped now, since the pause image that's defined in containerd repo is still using `k8s.gcr.io`
- PodSecurity admission plugin version is still kept at `pod-security.admission.config.k8s.io/v1alpha1` since talos 1.2 has to support n-2 k8s versions


TODO:
- PodSecurityPolicy admission plugin is removed in 1.25, this probably needs some handling in machine config
- stop applying the "node-role.kubernetes.io/master:NoSchedule" taint to control plane nodes for new clusters. Remove the taint from existing control plane nodes during upgrades? (kubeadm does this)

smira:
> PodSecurityPolicy admission plugin is removed in 1.25, this probably needs some handling in machine config

we already default to disabling PSP, we can remove it completely with 1.27 I guess?

internal/integration/provision/upgrade.go
```diff
@@ -76,8 +76,8 @@ const (
 	stableRelease   = "v1.1.0" // or soon-to-be-stable
 	// The current version (the one being built on CI) is DefaultSettings.CurrentVersion.
 
-	previousK8sVersion = "1.23.6" // constants.DefaultKubernetesVersion in the previousRelease
-	stableK8sVersion   = "1.24.2" // constants.DefaultKubernetesVersion in the stableRelease
+	previousK8sVersion = "1.24.2"         // constants.DefaultKubernetesVersion in the previousRelease
```
smira, line 79: this shouldn't be updated, Talos 1.0.6 ships with 1.23.6, it doesn't even support 1.24

smira (approved):
:+1:

smira:
> Should i remove it from machineconfig all together?

please, no, Talos validates machine config on unknown keys, so we can't do that this way... at least let's not pull the plug right now

frezbo:
Will try to do PR's to calico and aws cloud controller rather than implementing hacks

frezbo:
Calico PR https://github.com/projectcalico/calico/pull/6370

smira:
btw, there's already alpha.3 available ;)

internal/pkg/discovery/registry/kubernetes.go
```diff
@@ -94,7 +94,7 @@ func AffiliateFromNode(node *v1.Node) *cluster.AffiliateSpec {
 	}
 
 	// Machine type is derived from node roles.
-	_, labelMaster := node.Labels[constants.LabelNodeRoleMaster]
+	_, labelMaster := node.Labels[constants.LabelNodeRoleControlPlane]
```
smira, line 97: I guess this should stay for backwards compatibility? or it should be removed, as control plane label is checked below

pkg/kubernetes/kubernetes.go
```diff
@@ -234,15 +234,12 @@ func (h *Client) LabelNodeAsMaster(ctx context.Context, name string, taintNoSche
...
 
-	n.Labels[constants.LabelNodeRoleMaster] = ""
 	n.Labels[constants.LabelNodeRoleControlPlane] = ""
 
 	taintIndex := -1
 
-	// TODO: with K8s 1.21, add new taint LabelNodeRoleControlPlane
-
 	for i, taint := range n.Spec.Taints {
-		if taint.Key == constants.LabelNodeRoleMaster {
```
smira, line 245: I guess we should match on both here - in case of an upgrade, the taint will be `master`, and we want to convert it to taint `control-plane`

hack/release.toml
```diff
@@ -120,6 +120,7 @@ vlan=eth1.5:eth1 ip=172.20.0.2::172.20.0.1:255.255.255.0::eth1.5:::::
 * Linux: 5.15.59
 * Flannel 0.19.1
 * containerd 1.16.7
+* Kubernetes: v1.25.0-beta.0
```
smira, line 123: I think we need to add a note on no longer labeling control plane nodes with `master` label, and changing the taint to be `control-plane` now. This is a breaking change for some workloads

smira (approved):
`s/chore:/feat:/` :)

internal/pkg/discovery/registry/kubernetes.go
frezbo: i think this one is fine, since on upgrade it would get a new label and will still work

smira: oh I mean that if you look on lines 97-98 and below what we do is we check for either `master` or `control-plane` labels, so check for `master` label can be dropped

frezbo: ahh yes, saw that while fixing other issues.

api/machine/machine.proto
```diff
@@ -1027,6 +1027,7 @@ message ClusterConfig {
   ControlPlaneConfig control_plane = 2;
   ClusterNetworkConfig cluster_network = 3;
   bool allow_scheduling_on_masters = 4;
+  bool allow_scheduling_on_control_planes = 5;
```
smira, line 1030: I think we can simply rename here, as protobuf doesn't care about field names

pkg/kubernetes/kubernetes.go
```diff
@@ -234,30 +234,30 @@ func (h *Client) LabelNodeAsMaster(ctx context.Context, name string, taintNoSche
...
 	}
 
-	n.Labels[constants.LabelNodeRoleMaster] = ""
 	n.Labels[constants.LabelNodeRoleControlPlane] = ""
+	// TODO: frezbo
+	// remove all references to `constants.LabelNodeRoleMaster` when Talos supports k8s v1.27
+	delete(n.Labels, constants.LabelNodeRoleMaster)
 
-	taintIndex := -1
+	newTaints := make([]corev1.Taint, 0)
```
smira, line 242: nit:

```suggestion
	newTaints := make([]corev1.Taint, 0, len(n.Spec.Taints))
```

smira (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -977,7 +977,17 @@ type ClusterConfig struct {
 	//     - yes
 	//     - false
 	//     - no
+	//
+	// Deprecated: Use `AllowSchedulingOnControlPlanes` instead.
```
smira, line 981: we can instead drop the whole docblock here and replace it with:

```
	// docgen:nodoc
```

this way it will disappear from 1.2 docs altogether, which imho makes sense

smira (approved):

Issue #5894, opened 2022-07-11
generating `secrets.yaml` from Kubernetes `/etc/kubernetes`


smira:
`talosctl gen secrets --from-kubernetes-pki=/etc/kubernetes --kubernetes-bootstrap-token=....`

The tool will copy out all the PKI it can find and generate new PKI for Talos-specific secrets.

PR #5898, opened 2022-07-12
feat: gen secrets from kubernetes pki dir

utkuozdemir:
This PR allows the ability to generate `secrets.yaml` (`talosctl gen secrets`) using a Kubernetes PKI directory path (e.g. `/etc/kubernetes/pki`) as input. Also introduces the flag `--kubernetes-bootstrap-token` to be able to set a static Kubernetes bootstrap token to the generated `secrets.yaml` file instead of a randomly-generated one. Closes siderolabs/talos#5894.


internal/integration/cli/pki.go
```diff
@@ -0,0 +1,70 @@
...
+	//go:embed "testdata/pki/sa.key"
+	pkiSAKey []byte
+	//go:embed "testdata/pki/etcd/ca.crt"
+	pkiEtcdCACrt []byte
+	//go:embed "testdata/pki/etcd/ca.key"
+	pkiEtcdCAKey []byte
+)
+
+func createTempPKIDir() (string, error) {
+	dir, err := os.MkdirTemp("", "talos-gen-test-")
```
smira, line 31: we can do `suite.T().TempDir()` which has automatic cleanup

pkg/machinery/config/types/v1alpha1/generate/generate.go
```diff
@@ -208,41 +208,26 @@ func NewSecretsBundle(clock Clock, opts ...GenOption) (*SecretsBundle, error) {
...
 	}
 
+	return NewSecretsBundleWithOptions(clock, options.VersionContract, "", "")
+}
+
+// NewSecretsBundleWithOptions creates secrets bundle,
+// generating all secrets or reading from the input options if provided.
+//
+//nolint:gocyclo
+func NewSecretsBundleWithOptions(clock Clock, versionContract *config.VersionContract, pkiDir, bootstrapToken string) (*SecretsBundle, error) {
```
smira, line 218: from the API perspective it feels like it should be a different method.

we have `NewSecretsBundle` which gives us all fresh unique PKI
we can have `NewSecretsBundleFromKubernetesPKI` which imports whatever it can from Kubernetes PKI and fills in the missing stuff. 

The implementation might be that we have a method which fills in missing secrets (generates them), and the `NewSecretsBundle` just calls it, while `NewSecretsBundleFromKubernetesPKI` first fills in what it can, and that calls that common shared method?

Right now it feels a bit of a hard to follow code, and this code is very critical to be reviewable

internal/integration/cli/gen.go
```diff
@@ -209,19 +208,62 @@ func (suite *GenSuite) TestSecrets() {
...
+// TestSecretsWithPKIDirAndToken ...
+func (suite *GenSuite) TestSecretsWithPKIDirAndToken() {
+	path := "/tmp/secrets-with-pki-dir-and-token.yaml"
+
+	dir, err := createTempPKIDir()
+	suite.Assert().NoError(err)
+
+	defer os.RemoveAll(dir) //nolint:errcheck
+
+	suite.RunCLI([]string{
```
smira, line 233: :+1: on having these tests

internal/integration/cli/pki.go
utkuozdemir: done

pkg/machinery/config/types/v1alpha1/generate/generate.go
utkuozdemir: I did it the way you suggested, you can have another look.

frezbo (approved):
Also needs something in the release notes and also a knowledgeable docs

frezbo (approved):

utkuozdemir:
Its documentation will be covered in a separate PR.

smira (approved):
:+1:

PR #5899, opened 2022-07-12
feat: resubstitute talos.config url variables on retry

### LINKED ISSUE #3272: config URLs should have variable substitution

sauterp:
The download of the talos configuration may fail because the substituted
information like the hostname may not be initialized yet. Therefore we
retry the download and resubstitute the variables each time.

Fixes #3272

# Pull Request

## Acceptance

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -26,6 +26,7 @@ type downloadOptions struct {
 	Headers    map[string]string
 	Format     string
 	LowSrcPort bool
+	Endpoint   string
 
 	ErrorOnNotFound      error
 	ErrorOnEmptyResponse error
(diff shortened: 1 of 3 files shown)


pkg/download/download.go
```diff
@@ -57,73 +58,82 @@ func WithFormat(format string) Option {
...
 // WithErrorOnEmptyResponse provides specific error to return when response is empty.
 func WithErrorOnEmptyResponse(e error) Option {
-	return func(d *downloadOptions) {
+	return func(d *DownloadOptions) {
 		d.ErrorOnEmptyResponse = e
 	}
 }
 
 // Download downloads a config.
 func Download(ctx context.Context, endpoint string, opts ...Option) (b []byte, err error) {
```
smira, line 89: feels like the endpoint is ignored now?

pkg/download/download.go
```diff
@@ -22,20 +22,21 @@ import (
 
 const b64 = "base64"
 
-type downloadOptions struct {
+type DownloadOptions struct {
```
smira, line 25: I don't think we need to have this public, we could rather do an option which provides a function which returns the endpoint.

Something like `WithEndpointFunc(func(ctx context.Context) (string, error)`

smira: So unless `WithEndpointFunc()` is specified, we use the `endpoint` arg to keep backwards compat

frezbo (approved):

Unix4ever (approved):

Issue #4470, opened 2021-10-29
Kubelet fails to use new control plane endpoint if changed


andrewrynhard:
### Description

I used the wrong IP in setting up the cluster and changed it with `edit mc --immediate`. The kubelet still tries to use this address, as well as the `KubernetesPushController`.

### Logs

`145.40.102.223` is the old IP:

```
139.178.84.183: E1029 16:48:37.061293    5085 controller.go:144] failed to ensure lease exists, will retry in 7s, error: Get "https://145.40.102.223:6443/apis/coordination.k8s.io/v1/namespaces/kube-node-lease/leases/flashmob-01?timeout=10s": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

```
139.178.84.183: 2021-10-29T16:48:02.212Z ERROR controller failed {"component": "controller-runtime", "controller": "cluster.KubernetesPushController", "error": "error pushing to Kubernetes registry: failed to get node \"flashmob-01\": Get \"https://145.40.102.223:6443/api/v1/nodes/flashmob-01?timeout=30s\": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"}
```

### Environment

- Talos version: 0.13.0
- Platform: Equinix Metal

PR #5902, opened 2022-07-13
fix: update kubelet kubeconfig when cluster control plane endpoint changes

utkuozdemir:
Overwrite cluster's server URL in the kubeconfig file used by kubelet when the cluster control plane endpoint is changed in machineconfig, so that kubelet doesn't lose connectivity to kube-apiserver.

Closes siderolabs/talos#4470.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
@@ -8,8 +8,10 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"text/template"
@@ -21,6 +23,7 @@ import (
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/tools/clientcmd"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
 
 	runtimetalos "github.com/talos-systems/talos/internal/app/machined/pkg/runtime"
@@ -184,6 +187,14 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 			}
 		}
 
+		err = updateKubeconfig(secretSpec.Endpoint)
+
+		// there is no problem if the file does not exist yet since
+		// it will be created by kubelet on start with correct endpoint
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+
 		if err = ctrl.V1Alpha1Services.Start("kubelet"); err != nil {
 			return fmt.Errorf("error starting kubelet service: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
```diff
@@ -0,0 +1,138 @@
...
+			return fmt.Errorf("error getting config: %w", err)
+		}
+
+		endpoint := machineConfig.(*config.MachineConfig).Config().Cluster().Endpoint()
+
+		if err = conditions.WaitForKubeconfigReady(constants.KubeletKubeconfig).Wait(ctx); err != nil {
+			return err
+		}
+
+		if err = updateKubeconfigControlPlaneEndpoint(endpoint.String()); err != nil {
```
smira, line 74: do we need to restart the kubelet to pick up the changes? I believe it will be indirectly restarted by another controller

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("couldn't read server from kubeconfig")
+	}
+
+	if server == newEndpoint {
+		return nil
+	}
+
+	clusterInfo["server"] = newEndpoint
+
+	err = kubeconfig.Truncate(0)
```
smira, line 127: I think this might not be safe to do while the kubelet is running... as a proposal, should we rather do it in another controller?

https://github.com/AlekSi/talos/blob/2a2eba4705e0fb8ba057a753028173b79dd53af5/internal/app/machined/pkg/controllers/k8s/kubelet_service.go#L181-L190

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
```diff
@@ -0,0 +1,138 @@
...
+// Name implements controller.Controller interface.
+func (ctrl *KubeletKubeconfigController) Name() string {
+	return "k8s.KubeletKubeconfigController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *KubeletKubeconfigController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
```
smira, line 37: https://github.com/AlekSi/talos/blob/2a2eba4705e0fb8ba057a753028173b79dd53af5/pkg/machinery/resources/secrets/kubelet.go#L26-L34 

this is a proper resource for the endpoint, so that we don't depend on machine config everywhere

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
utkuozdemir: Yep, it gets restarted. That's why I initially added (now obsolete, an oversight) the comment you commented on.

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
utkuozdemir: You suggest removing this controller completely, right?

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
utkuozdemir: Do you suggest adding it as a new field to it? Because right now it doesn't have it.

smira: Endpoint is the first field there unless I'm missing something

utkuozdemir: oh right, totally missed that one 🤦

internal/app/machined/pkg/controllers/k8s/kubelet_kubeconfig.go
utkuozdemir: Moved it into `KubeletServiceController` with a bit different logic.

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -184,9 +188,33 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
...
+		// try updating kubeconfig while kubelet is not running
+		upToDate, err := ctrl.updateKubeconfig(ctx, secretSpec.Endpoint)
+		if err != nil {
+			return fmt.Errorf("error updating kubeconfig: %w", err)
+		}
+
 		if err = ctrl.V1Alpha1Services.Start("kubelet"); err != nil {
 			return fmt.Errorf("error starting kubelet service: %w", err)
 		}
+
```
smira, line 200: discussed in call, we probably don't need this section, only lines 191-196

PR #5915, opened 2022-07-15
refactor: remove bootstrap sequence

smira:
Refactor things to remove the bootstrap sequence, this should help with
the task of sequencer concurrency changes and immediate API feedback.


Unix4ever (approved):

sauterp (approved):

frezbo (approved):
is this some sort of breaking change?

smira:
> is this some sort of breaking change?

nope, it should be invisible to the user, as it works same way as before

Issue #5293, opened 2022-04-04
enable seccomp profile by default


smira:
## Feature Request

See e.g. https://betterprogramming.pub/secure-your-kubernetes-cluster-with-seccomp-9403ecf831b2

We can enable seccomp default profile to be applied by default to all Kubernetes workloads, this should improve overall security.

We should also ship the audit profile and provide instructions on building custom seccomp profiles.

We should provide a way to inject additional seccomp profiles to the system (via machine config?).

JAORMX:
I'd really dig this enhancement!

@smira @frezbo also, if y'all are interested in the SPO, I'm one of the maintainers. We'd be happy to answer questions on the #security-profiles-operator channel in the kubernetes slack instance.

PR #5917, opened 2022-07-17
feat: follow KEP-2568 non-root enhancements

nberlee:
# Pull Request
~~WIP: works for new clusters, haven't tested upgrades (likely this breaks because of existing file permissions...?)~~
EDIT: tested and it just works, even with --preserve files are recreated with correct permission

~~Question: Should I include the seccomp yaml for the control-plane?~~

## What? (description)
Follow KEP-2568 (https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2568-kubeadm-non-root-control-plane). Seperation UIDs, Adding GIDs, allowPriviledgeEscalation: false,drop unneeded capabilities and add seccomp CP manifest changes.

~~Seccomp is excluded as i think it has a dependency on https://github.com/siderolabs/talos/issues/5293 ? Also I have no idea how to test that.~~

## Why? (reasoning)
Mainly better compatibility (this KEP is scheduled to be implented for kubeadm 1.25) + security:
* `allowPrivilegeEscalation: false`: adds [no_new_privs](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) flag on the container process
* `Capabilities: { Drop: [ "ALL" ] }`: Capabilities are a more fine-grained permissions model, and all capabilities should be dropped from a pod, with only those required added back. (NET_BIND_SERVICE is needed for kube-apiserver as it has setcap flags)
*  `RunAsUser` & `RunAsGroup`: Assigning an unique ID helps with limiting the file exposure of secrets
## Acceptance

- [X] you ran conformance (`make conformance`) (GPG identity fails)

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -423,12 +423,26 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 								v1.ResourceMemory: apiresource.MustParse("512Mi"),
 							},
 						},
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: pointer.To(false),
+							Capabilities: &v1.Capabilities{
+								Drop: []v1.Capability{"ALL"},
+								// kube-apiserver binary has cap_net_bind_service=+ep set.
+								// It does not matter if ports < 1024 are configured, the setcap flag causes a capability dependency.
+								// https://github.com/kubernetes/kubernetes/blob/5b92e46b2238b4d84358451013e634361084ff7d/build/server-image/kube-apiserver/Dockerfile#L26
+								Add: []v1.Capability{"NET_BIND_SERVICE"},
+							},
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
 					},
 				},
 				HostNetwork: true,
 				SecurityContext: &v1.PodSecurityContext{
 					RunAsNonRoot: pointer.To(true),
-					RunAsUser:    pointer.To[int64](constants.KubernetesRunUser),
+					RunAsUser:    pointer.To[int64](constants.KubernetesAPIServerRunUser),
+					RunAsGroup:   pointer.To[int64](constants.KubernetesAPIServerRunGroup),
 				},
 				Volumes: append([]v1.Volume{
 					{
@@ -568,12 +582,22 @@ func (ctrl *ControlPlaneStaticPodController) manageControllerManager(ctx context
 								v1.ResourceMemory: apiresource.MustParse("256Mi"),
 							},
 						},
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: pointer.To(false),
+							Capabilities: &v1.Capabilities{
+								Drop: []v1.Capability{"ALL"},
+							},
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
 					},
 				},
 				HostNetwork: true,
 				SecurityContext: &v1.PodSecurityContext{
 					RunAsNonRoot: pointer.To(true),
-					RunAsUser:    pointer.To[int64](constants.KubernetesRunUser),
+					RunAsUser:    pointer.To[int64](constants.KubernetesControllerManagerRunUser),
+					RunAsGroup:   pointer.To[int64](constants.KubernetesControllerManagerRunGroup),
 				},
 				Volumes: append([]v1.Volume{
 					{
@@ -678,12 +702,22 @@ func (ctrl *ControlPlaneStaticPodController) manageScheduler(ctx context.Context
 								v1.ResourceMemory: apiresource.MustParse("64Mi"),
 							},
 						},
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: pointer.To(false),
+							Capabilities: &v1.Capabilities{
+								Drop: []v1.Capability{"ALL"},
+							},
+							SeccompProfile: &v1.SeccompProfile{
+								Type: v1.SeccompProfileTypeRuntimeDefault,
+							},
+						},
 					},
 				},
 				HostNetwork: true,
 				SecurityContext: &v1.PodSecurityContext{
 					RunAsNonRoot: pointer.To(true),
-					RunAsUser:    pointer.To[int64](constants.KubernetesRunUser),
+					RunAsUser:    pointer.To[int64](constants.KubernetesSchedulerRunUser),
+					RunAsGroup:   pointer.To[int64](constants.KubernetesSchedulerRunGroup),
 				},
 				Volumes: append([]v1.Volume{
 					{
--- a/internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
+++ b/internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
@@ -93,11 +93,15 @@ func (ctrl *RenderConfigsStaticPodController) Run(ctx context.Context, r control
 		for _, pod := range []struct {
 			name      string
 			directory string
+			uid       int
+			gid       int
 			configs   []configFile
 		}{
 			{
 				name:      "kube-apiserver",
 				directory: constants.KubernetesAPIServerConfigDir,
+				uid:       constants.KubernetesAPIServerRunUser,
+				gid:       constants.KubernetesAPIServerRunGroup,
 				configs: []configFile{
 					{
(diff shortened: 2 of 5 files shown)


internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -423,12 +423,23 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 								v1.ResourceMemory: apiresource.MustParse("512Mi"),
 							},
 						},
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: pointer.To(false),
+							Capabilities: &v1.Capabilities{
+								Drop: []v1.Capability{"ALL"},
+								// kube-apiserver binary has cap_net_bind_service=+ep set.
```
frezbo, line 430: I remember seeing a fix in k8s/container+runc where the `cap_net_bind_service=+ep` when set on the binary is honored and there is not need to explicitly set `NET_BIND_SERVICE`.

frezbo: okay, that was not a fix, it was a KEP, [KEP-2763](https://github.com/kubernetes/enhancements/pull/2757/), it's  strange that this issue still persists

pkg/machinery/constants/constants.go
```diff
@@ -232,8 +232,23 @@ const (
...
+	KubernetesAPIServerRunUser = 65534
+
+	// KubernetesAPIServerRunGroup defines GID to run the API Server.
+	KubernetesAPIServerRunGroup = 65534
+
+	// KubernetesControllerManagerRunUser defines UID to the Controller Manager.
+	KubernetesControllerManagerRunUser = 65535
+
+	// KubernetesControllerManagerRunGroup defines GID to run the Controller Manager.
+	KubernetesControllerManagerRunGroup = 65535
```
frezbo, line 245: any reason why everything just can run as 65535. I guess the security worry is that different k8s api components could potentially access stuff owned by other component

internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
```diff
@@ -93,11 +93,15 @@ func (ctrl *RenderConfigsStaticPodController) Run(ctx context.Context, r control
...
 			name      string
 			directory string
+			uid       int
+			gid       int
 			configs   []configFile
 		}{
 			{
 				name:      "kube-apiserver",
 				directory: constants.KubernetesAPIServerConfigDir,
+				uid:       constants.KubernetesAPIServerRunUser,
```
frezbo, line 103: I guess this is for the user defined pods (I could be wrong here). If that's the case, we should not use defaults here

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
nberlee: Please tell me where I did go wrong :)
I've created a build without the `Add: [ NET_BIND_SERVICE ]`

```
$ talosctl cluster create --image  ghcr.io/nberlee/talos:v1.1.0-alpha.2-155-ga64941c76-dirty  --wait
generating PKI and tokens
downloading ghcr.io/nberlee/talos:v1.1.0-alpha.2-155-ga64941c76-dirty
creating network talos-default
creating master nodes
creating worker nodes
renamed talosconfig context "talos-default" -> "talos-default-8"
waiting for API
bootstrapping cluster
waiting for etcd to be healthy: OK
waiting for etcd members to be consistent across nodes: OK
waiting for apid to be ready: OK
waiting for kubelet to be healthy: OK
◱ waiting for all nodes to finish boot sequence: nodes ["10.5.0.2"] are still in boot sequence
context deadline exceeded
$ talosctl read -n 10.5.0.2 /var/log/pods/kube-system_kube-apiserver-talos-default-master-1_1c0b00f83b9263e8c3da34656ecec18b/kube-apiserver/2.log
2022-07-17T17:11:37.972419441Z stderr F exec /usr/local/bin/kube-apiserver: operation not permitted
```

frezbo: sorry if I confused you, i just linked an old KEP that I've seen, seems it's not completely fixed as i understand. Just a general observation, would have been nice to have it honor ambient capabilities.

internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
nberlee: These are configuration manifests (in this case only `admission-control-config.yaml`). Which must be to be readable by the APIServer pod.

As they are mounted in the pod:
```
  - hostPath:
      path: /system/config/kubernetes/kube-apiserver
      type: ""
    name: config
 ```
 
I have effectively only changed it from `constants.KubernetesRunUser` to `constants.KubernetesAPIServerRunUser` see: https://github.com/siderolabs/talos/pull/5917/files/a64941c7613289343e46385dc23c637190be2787#diff-c8111a0cf068d2938606c06f94680cc646d70d2aa4c2f32339aed1e409f076d3L131

File: pkg/machinery/constants/constants.go (hunk 2 above)
Comment (nberlee) [reply pkg/machinery/constants/constants.go:245]: Indeed, it add an extra layer of defense. Practically this protection layers acts in two fold:
1) it protects against containerd/runc/kubelet( for kubelet would not be a first!) security mount bugs which allows for more file access on the host.
2) it adds an extra layer of defense as a container escape would not be enough the extract more secrets (because of the filesystem owner). This is because you are still the same user on the host when you escape a container.

In both cases the process userid is still evaluated when access files.

Comment (nberlee) [reply pkg/machinery/constants/constants.go:245]: And a less important argument: it follows the KEP (kubeadm is doing it too)

File: internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go (hunk 3 above)
Comment (frezbo) [reply internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go:103]: Thanks for proving me wrong, I thought this was for user defined pods :sweat_smile:

File: internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go (hunk 1 above)
Comment (nberlee) [reply internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go:430]: As I understand  KEP-2763 made ambient capabilities work. And made things worse in this case.
 In this case, we do not care if its working. We actually do not want to honor the filecap, (as we do not need it)
 
 But because ambient capabilities (ie filecap) are honored, and was not impemented as a seperate `ambientCapabilities:`, we now have to allow  `NET_BIND_SERVICE`.

See the example which was previous behavior we would like in our case after `Now the if we update the image and create the pod:-` in [KEP-2763](https://github.com/vinayakankugoyal/enhancements/blob/0cfb55c5b6dd4a0b511fd17701d86a778817b455/keps/sig-security/2763-ambient-capabilities/README.md#motivation)

**Comment by nberlee** [comment]
So regarding my question in the PR description: `Should I include the seccomp yaml for the control-plane?`

Everything still works in QEMU when i add 
```
							SeccompProfile: &v1.SeccompProfile{
								Type: v1.SeccompProfileTypeRuntimeDefault,
							},
```

As it is part of the KEP-2568, should I add it? I am asking as there is already an issue regarding this for all containers: https://github.com/siderolabs/talos/issues/5293 and I am not sure if there is something on the OS level that still needs to be done... 
If there is, I think it's actually bad to add this. If not, I am happy to add this to this PR or a new one.

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
frezbo: yup, makes sense, I wrongly assumed k8s already implemented setting the ambient capabilities by now

frezbo (approved):

frezbo:
> So regarding my question in the PR description: `Should I include the seccomp yaml for the control-plane?`
> 
> Everything still works in QEMU when i add
> 
> ```
> 							SeccompProfile: &v1.SeccompProfile{
> 								Type: v1.SeccompProfileTypeRuntimeDefault,
> 							},
> ```
> 
> As it is part of the KEP-2568, should I add it? I am asking as there is already an issue regarding this for all containers: #5293 and I am not sure if there is something on the OS level that still needs to be done, or why the issue is hold back... If there is, I think it's actually bad to add this. If not, I am happy to add this to this PR or a new one.

I'll be looking into adding the seccomp support this week and see if there's anything extra needed.

frezbo:
I guess you can add it, since kubelet will not enforce it as long as the feature-gate/cli flags are set, so I think it's okay to do this in this PR

nberlee:
> I guess you can add it, since kubelet will not enforce it as long as the feature-gate/cli flags are set, so I think it's okay to do this in this PR

done

frezbo (approved):

smira (approved):
thanks!

yep, `--preserve` is fine as all the mounted files are created on `tmpfs` anyways, so they are thrown away with reboot.

PR #5932, opened 2022-07-18
feat: implement packet capture API

smira:
This uses the `go-packet` library with native bindings for the packet
capture (without `libpcap`). This is not the most performant way, but it
allows us to avoid CGo.

There is a problem with converting network filter expressions (like
`tcp port 3222`) into BPF instructions, it's only available in C
libraries, but there's a workaround with `tcpdump`.


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -2044,6 +2048,97 @@ func (s *Server) GenerateClientConfiguration(ctx context.Context, in *machine.Ge
 	return reply, nil
 }
 
+// PacketCapture performs packet capture and streams the pcap file.
+//
+//nolint:gocyclo
+func (s *Server) PacketCapture(in *machine.PacketCaptureRequest, srv machine.MachineService_PacketCaptureServer) error {
+
+	log.Printf("in PacketCapture")
```
frezbo, line 2056: debug

smira: yeah, lots of debug still :) it's heavy WIP

cmd/talosctl/cmd/talos/pcap.go
```diff
@@ -0,0 +1,196 @@
...
+ BPF filter can be applied, but it has to compiled to BPF instructions first using tcpdump.
+ Correct link type should be specified for the tcpdump: EN10MB for Ethernet links and RAW
+ for e.g. Wireguard tunnels:
+
+   talosctl pcap -i eth0 --bpf-filter "$(tcpdump -dd -y EN10MB 'tcp and dst port 80')"
+
+   talosctl pcap -i kubespan --bpf-filter "$(tcpdump -dd -y RAW 'port 50000')"
+
+As packet capture is transmitted over the network, it is recommended to filter out the Talos API traffic,
+e.g. by excluding packets with the port 50000.
```
frezbo, line 64: probably good to have an example that filters traffic on port 50000

frezbo (approved):
:cool:

cmd/talosctl/cmd/talos/pcap.go
smira: also there might be some fun if the request is forwarded between apids, but still port 50000.

smira: I'm torn on that... We can only build a filter with tcpdump on client side. We need to know capture data type (-y). Client should have tcpdump installed.

PR #5936, opened 2022-07-18
fix: skip bond itself when matching interface (Equinix Metal)

smira:
This fixes a problem when platform network configuration might have
already been applied from the cached on disk representation, and in that
base e.g. `bond0` MAC is same as `eth0`, so Talos might mistakenly pick
up `bond0` as a slave to itself instead of `eth0`.


frezbo (approved):
do we need a test?

smira:
> do we need a test?

this one is tricky to test properly, as it reads the list of network interfaces while running, so we would need to mock that.

we have a test which basically relies on the `else` clause there which works if the MAC is not found

Issue #5450, opened 2022-04-27
sequence concurrency changes


smira:
## Feature Request

When boot sequence isn't done, we can't upgrade the node (we should be able to upgrade the node always).

### Description

We should have a way to abort the `boot` sequence once another high-priority sequence is launched (reboot, reset, upgrade). 

We want to allow running only a single sequence at a time.

(bootstrapping `etcd` is an exception today - it runs as a tiny sequence in parallel with boot sequence - can we refactor that or make it the only exception)? only a single bootstrap etcd request should be running at a time on a single node

### Aborting a sequence

All sequences have a `context.Context` associated with them. We can cancel the context and check for context cancellation in between the task runs (and in phase runners). 

We can have a small graceful timeout (say 10 seconds) for the sequence to terminate, and if doesn't terminate (e.g. it's stuck in a task), we can ignore that.

### Taking over sequences

| what is running (columns) what is requested (rows) | boot | reboot | reset | upgrade |
|----------------------------------------------------|------|--------|-------|---------|
| reboot                                             | Y    | Y      | Y     | N       |
| reset                                              | Y    | N      | N     | N       |
| upgrade                                            | Y    | N      | N     | N       |

### Integration Testing

We should test that those `Y` cells should be covered with tests (add some jitter to verify that it works any time into the running sequence).

PR #5943, opened 2022-07-19
feat: allow running only one sequence at a time

Unix4ever:
Fix `Talos` sequencer to run only a single sequence at the same time.
Sequences priority was updated. To match the table:

| what is running (columns) what is requested (rows) | boot | reboot | reset | upgrade |
|----------------------------------------------------|------|--------|-------|---------|
| reboot                                             | Y    | Y      | Y     | N       |
| reset                                              | Y    | N      | N     | N       |
| upgrade                                            | Y    | N      | N     | N       |

With a small addition that `WithTakeover` is still there.
If set, priority is ignored.

This is mainly used for `Shutdown` sequence invokation.
And if doing apply config with reboot enabled.

Fixes: https://github.com/siderolabs/talos/issues/5450


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -243,6 +269,16 @@ func (c *Controller) Unlock() bool {
 }
 
 func (c *Controller) run(ctx context.Context, seq runtime.Sequence, phases []runtime.Phase, data interface{}) error {
+	c.ctxMutex.Lock()
```
smira, line 272: I think we should also add handling for canceled context is between phases,  so if the context is canceled, it shouldn't run the next phase?

my assumption is not all tasks actually use `ctx`, so they might keep running if the context is canceled.

smira: (this is not the right place to add this check, I was just trying to add comment somewhere in this file)

Unix4ever: I think phases loop already handles that properly.
https://github.com/siderolabs/talos/blob/master/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go#L296

While tasks loop doesn't and tasks are running in parallel. Can add something there if it makes sense

smira (approved):

sauterp (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -243,6 +269,16 @@ func (c *Controller) Unlock() bool {
...
+	c.runningSequence = seq
```
smira, line 273: is there any reason we set `runningSequence` here, while other pieces are set in `Run()` ?

as we check for `runningSequence` in `Run()`, does it make sense to set the `runningSequence` there as well? and we should set it somehow atomically with a decision to run the sequence

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -92,59 +118,53 @@ func (c *Controller) Run(ctx context.Context, seq runtime.Sequence, data interfa
...
 		}
 
-		defer c.Unlock()
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	defer c.Unlock()
```
smira, line 157: there is a gap here with the flow and decision making:

1. first we check whether we can take over the running sequence (while holding `c.ctxMutex`)
2. after the decision is made we attempt to acquire another lock with `c.TryLock`, but beween 1 and 2 there might be any change to the running sequence, or two concurrent `Run()` might both make a decision to take over (while they can't takeover each other).
3. same with canceling the running sequence - it looks like several concurrent `Run()` calls can do that.


It feels to me we should have a much broader and proper lock around the decision to run the sequence, cancel the running, etc. In other words, if this `Run()` is going to take over a running sequence, it should do it so that no other goroutine can do the same. 

The smallest possible change is to re-check the conditions after making a final decisions, but this is not clear for me.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -316,6 +352,25 @@ func (c *Controller) runPhase(ctx context.Context, phase runtime.Phase, seq runt
 
 	var eg errgroup.Group
 
+	doneCh := make(chan struct{})
+
+	eg.Go(func() error {
+		done := 0
+
+		for {
```
smira, line 360: why do we need this?

it feels like it's re-doing what `errgroup` should already be doing

Unix4ever: to cancel waiting for errgroup if the context is canceled and any of the underlying task doesn't cancel when the context is canceled.

smira: if the task doesn't cancel, we can't cancel either way, as the task is still running?

we should in fact use https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext, as it automatically cancels the context when any of the sub-goroutines fails.

Unix4ever: oh, I didn't know that this thing exists. Will update

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -480,3 +465,51 @@ func waitForUSBDelay() (err error) {
...
+
+	ctx, c.cancelCtx = context.WithCancel(ctx)
+	c.runningSequence = seq
+
+	return ctx
+}
+
+func (c *Controller) tryTakeOver(ctx context.Context, seq runtime.Sequence, opts runtime.ControllerOptions) error {
+	return retry.Constant(time.Minute*1, retry.WithUnits(time.Millisecond*100)).RetryWithContext(ctx, func(ctx context.Context) error {
+		if c.cancelCtx != nil && c.runningSequence != -1 {
```
smira, line 489: should this read be locked?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -480,3 +465,51 @@ func waitForUSBDelay() (err error) {
...
+			canTakeOver = canTakeOver || opts.Takeover
+
+			if !canTakeOver {
+				return runtime.ErrLocked
+			}
+
+			c.cancelCtx()
+		}
+
+		if c.TryLock() {
```
smira, line 509: we could probably have something simple as a lock, e.g. `chan` with buffer size of 1.

as long as there's a "token" there, you can grab it and run a sequence; once done, return token back to the pool

takeOver is all about cancelling the running sequence (if we are allowed to), and once cancelled, we grab the token to actually run the sequence

at the same time we want whoever cancelled the sequence (took it over) to run its sequence, so this probably becomes more tricky on this side.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -32,13 +30,11 @@ import (
 // Controller represents the controller responsible for managing the execution
 // of sequences.
 type Controller struct {
-	r  *Runtime
-	s  *Sequencer
+	S  runtime.Sequencer
```
smira, line 33: suggestion: keep them private and do tests with `package v1alpha1`

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -77,74 +74,19 @@ func NewController() (*Controller, error) {
...
-			})
-			if err != nil {
-				return err
-			}
-		} else if c.TryLock() {
-			c.Runtime().Events().Publish(&machine.SequenceEvent{
-				Sequence: seq.String(),
-				Action:   machine.SequenceEvent_NOOP,
-				Error: &common.Error{
-					Code:    common.Code_LOCKED,
```
smira, line 128: looks like we dropped this completely?

Unix4ever: it's there: https://github.com/siderolabs/talos/pull/5943/files/0ead5453fffc6414d5134b78edb6c47b1a449eeb#diff-1e941df6ba5f23ed7e30cd639a93eff4eb463ad248013757b7f13e89d0faaaf8R100

smira (approved):

Issue #5834, opened 2022-06-29
Node joining timing issue can cause double join and wrong node name


Davincible:
## Bug Report
Upon booting, Talos sets its own hostname in the format of `talos-<IPv4 | IPv6>`, only later in the boot process it sets it's hostname according to either DHCP, or static config. 

The node joining process happens separately from this, and the chance exists that Talos joins the node to the Kubernetes cluster prior to setting the correct hostname, essentially creating a race condition on the hostname.

Whether a node joins k8s prior to setting the correct hostname, or after, determines if a node joins correctly or not, this is entirely up to chance, and often also goes right, but not always. 

On the event that a node joins prior to having set its correct hostname, it joins under the initial, wrong hostname, after which it receives its right hostname, and then tries to rejoin under the new hostname. This causes two nodes to show up in `kubectl get nodes`, one of which has an incorrect hostname, and thus doesn't actually exist in the cluster.

### Futher Observations
In the case of a premature join, the CSR requestor _can_ (not always) be set as `system:node:talos-209-151-154-201` instead of the regular `system:bootstrap:lgq9yx`.

The hostname always gets set 3 times, as found by grepping the dmesg logs.

### Possible Fix
Check if the hostname is correctly set prior to starting the joining process.

#### Kubectl nodes

```bash
$ kubectl get nodes -owide
NAME                        STATUS     ROLES                  AGE   VERSION   INTERNAL-IP       EXTERNAL-IP   OS-IMAGE         KERNEL-VERSION   CONTAINER-RUNTIME
talos-194-113-73-83         NotReady   <none>                 10m   v1.23.2   194.113.73.83     <none>        Talos (v1.1.0)   5.15.48-talos    containerd://1.6.6
talos-209-151-151-243       NotReady   <none>                 10m   v1.23.2   209.151.151.243   <none>        Talos (v1.1.0)   5.15.48-talos    containerd://1.6.6
talos-209-151-154-201       NotReady   <none>                 10m   v1.23.2   209.151.154.201   <none>        Talos (v1.1.0)   5.15.48-talos    containerd://1.6.6
test-cluster-cp-1-us-nyc1   Ready      control-plane,master   10m   v1.23.2   194.113.72.215    <none>        Talos (v1.1.0)   5.15.48-talos    containerd://1.6.6
[...]

Issue #5896, opened 2022-07-11
stable default hostname based on machine-id


smira:
Talos provides default hostname based on the IP address, e.g. `talos-192-168-1-1`. This hostname is not stable, as it changes with the IP address changes.

Proposal: replace it with the hostname based on stable ID, e.g. `machine-id` (32 chars): `talos-<machine-id>`. The benefit is that this hostname won't change unless the node is fully wiped.

ohauer:
That's a nice idea!
I learned it the hard way with a different OS and also run in trouble with certs assigned to IP addresses as the other OS does ipconfig release during reboot.
Havent't looked what happens with talos and certs in case the IP change, but remember from the other solution nodes registerd twice with both IP's which leeds to several other issues, since then Im using only static IP over all nodes (even with my talos test cluster)

smira:
Consider using shorter name by taking less characters, e.g. 6 chars a-z + 0.9: `36^6 = 2176782336`

smira:
This should be disabled by default in the machine config, and enabled only for v1.2.0+ machine configs.

smira:
thanks, good point on not exposing machine-id. I think if we take a part of machine-id, it should be fine as well, as exposing only 4 bytes out of 16 bytes should be non-identifying still. 

Also this hostname will change with node wipe, so if the node is reinstalled, it gets new machine-id and new default hostname.

PR #5947, opened 2022-07-19
fix: regenerate kubelet certs when hostname changes

utkuozdemir:
Clear the kubelet certificates and kubeconfig when hostname changes so that on next start, kubelet goes through the bootstrap process and new certificates are generated and the node is joined to the cluster with the new name.

Fixes siderolabs/talos#5834.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
@@ -7,7 +7,9 @@ package k8s
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/base64"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -30,6 +32,7 @@ import (
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system"
 	"github.com/talos-systems/talos/internal/app/machined/pkg/system/services"
 	"github.com/talos-systems/talos/pkg/machinery/constants"
+	"github.com/talos-systems/talos/pkg/machinery/generic/slices"
 	"github.com/talos-systems/talos/pkg/machinery/resources/files"
 	"github.com/talos-systems/talos/pkg/machinery/resources/k8s"
 	runtimeres "github.com/talos-systems/talos/pkg/machinery/resources/runtime"
@@ -187,11 +190,16 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 			}
 		}
 
-		err = updateKubeconfig(secretSpec.Endpoint)
+		// refresh certs only if we are managing the node name (not overridden by the user)
+		if cfgSpec.ExpectedNodename != "" {
+			err = ctrl.refreshKubeletCerts(cfgSpec.ExpectedNodename)
+			if err != nil {
+				return err
+			}
+		}
 
-		// there is no problem if the file does not exist yet since
-		// it will be created by kubelet on start with correct endpoint
-		if err != nil && !errors.Is(err, os.ErrNotExist) {
+		err = updateKubeconfig(secretSpec.Endpoint)
+		if err != nil {
 			return err
 		}
 
@@ -281,9 +289,13 @@ func (ctrl *KubeletServiceController) writeConfig(cfgSpec *k8s.KubeletSpecSpec) 
 	return ioutil.WriteFile("/etc/kubernetes/kubelet.yaml", buf.Bytes(), 0o600)
 }
 
-// updateKubeconfig updates the kubeconfig of kubelet with the given endpoint.
+// updateKubeconfig updates the kubeconfig of kubelet with the given endpoint if it exists.
 func updateKubeconfig(newEndpoint *url.URL) error {
 	config, err := clientcmd.LoadFromFile(constants.KubeletKubeconfig)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -168,6 +178,17 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 
 		secretSpec := secret.(*secrets.Kubelet).TypedSpec()
 
+		hostname, err := r.Get(ctx, resource.NewMetadata(network.NamespaceName, network.HostnameSpecType, network.HostnameID, resource.VersionUndefined))
+		if err != nil {
+			if state.IsNotFoundError(err) {
+				continue
+			}
+
+			return fmt.Errorf("error getting secrets: %w", err)
```
frezbo, line 187:
```suggestion
			return fmt.Errorf("error getting hostname: %w", err)
```

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -297,3 +324,86 @@ func updateKubeconfig(newEndpoint *url.URL) error {
...
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+
+	if err != nil {
+		return err
+	}
+
+	// clear the pki directory
+	for _, entry := range dir {
```
frezbo, line 360: do we need to go over each entry or is it okay to remove the whole `constants.KubeletPKIDir`?

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -297,3 +324,86 @@ func updateKubeconfig(newEndpoint *url.URL) error {
...
+	raw, err := os.ReadFile(filepath.Join(constants.KubeletPKIDir, "kubelet.crt"))
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	for {
```
frezbo, line 386: is this because the file contains multiple certs?

utkuozdemir: yep, one is CA and other is the server one.

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
utkuozdemir: hm, it seemed to be necessary but that was before I added `os.Remove(constants.KubeletKubeconfig)` afterwards. let me give it another try now.

utkuozdemir: yep, it works fine when deleted. updated the code.

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -201,6 +212,17 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 	}
 }
 
+// extractHostname extracts the hostname from the kubelet args.
+func (ctrl *KubeletServiceController) extractHostname(kubeletArgs []string) (string, error) {
```
frezbo, line 216: i wonder what happens if a user set this to empty?

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
utkuozdemir: Good point. It worked, but it was removing certs each time controller is run because of hostname mismatch each time. I changed it to return the system hostname as fallback now, which avoids recreating the certs when `--hostname-override=""`.

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -201,6 +212,25 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
...
+
+			break
+		}
+	}
+
+	if hostname != "" {
+		return hostname, nil
+	}
+
+	return os.Hostname()
```
frezbo, line 231: is it safe the get the hostname for the os package whereas we get it from the hostname resource elsewhere?

utkuozdemir: I talked to @smira just now and we will do it differently - we'll add a new field to the `kubeletspec`.

Davincible:
Haven't looked into/tested these changes, but if I understand correctly from the description it allows a node to properly (re)join upon getting a new hostname. Does it then also delete the old node in k8s with the wrong hostname, when it joins the second time under the right hostname?

Please correct me if I'm misunderstanding anything

Davincible:
@utkuozdemir ah I see. That's a good start.

I have no clue about the Talos internals but perhaps it would be possible to only start the joining process after the hostname is obtained (either statically or through DHCP) to circumvent the issue entirely. Would definitely improve the user experience

utkuozdemir:
@Davincible 

This fix mainly focuses on keeping the node at least "alive" on a hostname change.

We have [this issue](https://github.com/siderolabs/talos/issues/5896) to stabilize the default hostname against IP changes.

We also plan to do something similar to what you suggested (recording the first hostname as stable one).

smira (approved):
we should promote this at least to e2e

frezbo (approved):

internal/integration/api/update-hostname.go
```diff
@@ -0,0 +1,150 @@
...
+
+	nodeInternalIP := suite.RandomDiscoveredNodeInternalIP(machine.TypeWorker)
+
+	nodeCtx := client.WithNodes(suite.ctx, nodeInternalIP)
+
+	node, err := suite.GetK8sNodeByInternalIP(suite.ctx, nodeInternalIP)
+	suite.Require().NoError(err)
+
+	if strings.HasSuffix(node.Name, ".ec2.internal") {
+		suite.T().Skip("aws does not support hostname changes")
```
utkuozdemir, line 73: @smira what do you think about this?

smira: :+1:

smira (approved):

PR #5950, opened 2022-07-21
feat: set stable default hostname based on machine-id

utkuozdemir:
Use machine-id as the source for the default hostname (e.g. `talos-2gd-76y`) instead of DHCP-assigned IP (e.g. `talos-172-20-0-2`). This way, DHCP IP changes won't impact the hostname. Defaults to true for Talos version >=1.2.

Closes siderolabs/talos#5896.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/hostname_config.go
+++ b/internal/app/machined/pkg/controllers/network/hostname_config.go
@@ -6,17 +6,20 @@ package network
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"strings"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/martinlindhe/base36"
 	"github.com/siderolabs/go-pointer"
 	"github.com/talos-systems/go-procfs/procfs"
 	"go.uber.org/zap"
 
 	talosconfig "github.com/talos-systems/talos/pkg/machinery/config"
+	"github.com/talos-systems/talos/pkg/machinery/resources/cluster"
 	"github.com/talos-systems/talos/pkg/machinery/resources/config"
 	"github.com/talos-systems/talos/pkg/machinery/resources/network"
 )
@@ -46,6 +49,12 @@ func (ctrl *HostnameConfigController) Inputs() []controller.Input {
 			ID:        pointer.To(network.NodeAddressDefaultID),
 			Kind:      controller.InputWeak,
 		},
+		{
+			Namespace: cluster.NamespaceName,
+			Type:      cluster.IdentityType,
+			ID:        pointer.To(cluster.LocalIdentity),
+			Kind:      controller.InputWeak,
+		},
 	}
 }
 
@@ -61,7 +70,7 @@ func (ctrl *HostnameConfigController) Outputs() []controller.Output {
 
 // Run implements controller.Controller interface.
 //
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (ctrl *HostnameConfigController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
 	for {
 		select {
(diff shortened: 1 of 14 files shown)


internal/app/machined/pkg/controllers/network/hostname_config.go
```diff
@@ -97,7 +105,36 @@ func (ctrl *HostnameConfigController) Run(ctx context.Context, r controller.Runt
 			defaultAddr = addrs.(*network.NodeAddress) //nolint:errcheck,forcetypeassert
 		}
 
-		specs = append(specs, ctrl.getDefault(defaultAddr))
+		if cfgProvider != nil && cfgProvider.Machine().Features().StableHostnameEnabled() {
```
smira, line 108: we should wait for the `cfgProvider != nil` before we make a decision on which hostname to use, otherwise Talos will always first do `talos-192-168-1-1` and then replace it with stable machine id

internal/app/machined/pkg/controllers/network/hostname_config.go
```diff
@@ -172,6 +209,21 @@ func (ctrl *HostnameConfigController) apply(ctx context.Context, r controller.Ru
...
 }
 
+func (ctrl *HostnameConfigController) getStableDefault(machineID []byte) (*network.HostnameSpecSpec, error) {
+	id := string(machineID)
+
+	if len(id) < 6 {
+		return nil, fmt.Errorf("machine ID is too short: %s", id)
+	}
+
+	hostname := "talos-" + id[0:6]
```
smira, line 219: machineID is 0-9a-f, I think too little values for our purpose.

I would propose to do `talos-XXX-XXX` where each `X` is `a-z0-9`, this gives us 36^6 different hostnames. 

in current scheme we have 16^6, which is way smaller.

pkg/machinery/config/types/v1alpha1/v1alpha1_features.go
```diff
@@ -12,3 +12,12 @@ func (f *FeaturesConfig) RBACEnabled() bool {
 
 	return *f.RBAC
 }
+
+// StableHostnameEnabled implements config.Features interface.
+func (f *FeaturesConfig) StableHostnameEnabled() bool {
+	if f.StableHostname == nil {
```
smira, line 18: we have a shortcut for this: https://github.com/siderolabs/go-pointer/blob/master/pointer.go#L12-L21

smira (commented):
looks good, and we should also add a note to `hack/release.toml`

internal/app/machined/pkg/controllers/network/hostname_config.go
frezbo, line 219: should we make it easier for users to figure out which hostname has which ip by storing this data in a resource maybe, nothing as a blocker for this PR, just some ideas

internal/app/machined/pkg/controllers/network/hostname_config.go
utkuozdemir: Wait like `cfgProvider != nil { continue }`? Or just get `ctrl.getDefault(defaultAddr)` under the `if`? Because former changes the current behavior of the controller.

smira: don't do any default hostname until we get machine config... this will allow us to avoid these endless hostname changes

smira: (not `continue`, just skip the default hostname until cfgProvider is around)

internal/app/machined/pkg/controllers/network/hostname_config.go
utkuozdemir: Got it, so I guess first 6 chars of `lower(base36(nodeID))` with `-` in the middle.

Do you think hashing (sha256 or something) the nodeID beforehand  make sense? To avoid any kind of potential length issues etc.

smira: we can do hashing to prevent a leak of node identity via hostname (which might be sent out to the DHCP server)

utkuozdemir: Yep, I thought so 👍

utkuozdemir: Hm, I'm not sure if it is worth it tbh - if the cluster is there, they can do `kubectl get node -owide`. If not, bunch of `talosctl get`s would be enough. @smira any inputs?

internal/app/machined/pkg/controllers/network/hostname_config.go
utkuozdemir: done

internal/app/machined/pkg/controllers/network/hostname_config.go
smira: I don't think we can make it any easier... you can do `taloctl get members` to see the mapping

smira (approved):

PR #5957, opened 2022-07-22
fix: data race in packet capture (part 2)

smira:
The `PacketSource` interface is racy, as it provides a channel to read
packets from, while packets are read in a (invisible) goroutine, so
closing the capture handle creates a data race with reading.

Unwrap that goroutine into an explicit loop to avoid the race.


frezbo (approved):
:cool:

PR #5959, opened 2022-07-22
docs: Explain updating k8s components.

mrichar1:
# Pull Request

## What?
Explain that editing cluster config may also require a k8s-upgrade and component restart step,
with link to 'upgrading kubernetes' docs.

Updated these docs to explain that version-number can be preserved.

Signed-off-by: Matthew Richardson <M.Richardson@ed.ac.uk>

## Why? (reasoning)
It is not currently clear that editing the `cluster` part of the machine config also requires these extra steps before the components will be running with the new config.

## Acceptance


website/content/v1.1/talos-guides/configuration/editing-machine-configuration.md
```diff
@@ -48,6 +48,8 @@ The list of config changes allowed to be applied immediately in Talos {{< releas
 * `.machine.kernel`
 * `.machine.registries` (CRI containerd plugin will not pick up the registry authentication settings without a reboot)
 
+Changes to `.cluster` configuration for Kubernetes components will not take effect until the Kubernetes cluster is [upgraded]({{< relref "../../kubernetes-guides/upgrading-kubernetes" >}}) and the components are restarted - either manually or via a reboot.
```
smira, line 51: I think we might need more careful wording here.

There are two types of components configured in `.cluster` section:

* control plane components running as static pods (API server, controller manager and scheduler) - changes to these components are applied immediately on the fly
* other settings which in fact update bootstrap manifests - manifests themselves, CNI, kube-proxy configuration - these are not applied automatically, only missing resources are created, but existing ones are not updated (and never removed) - these require a Kubernetes upgrade to be applied

PR #5970, opened 2022-07-25
feat: refactor disk size matcher to be compatible with DeepEqual

### LINKED ISSUE #5860: install disk size matcher breaks the equality check on the machine config
## Bug Report

See https://github.com/siderolabs/talos/blob/86a0a7bdf70d318bed2143d65784faae6f9125d4/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go#L1142-L1148

### Description

As a function closure is stored inside the struct, it breaks the check on equality (that is, `reflect.DeepEqual` will report false for two disk matchers being equal internally).

Proper fix will be to store parsed expression as an operator + bytes, keeping the function generated on the fly when needed to perform the actual match.

### Environment

- Platform:

DmitriyMV:
Replace Matcher field with Matcher method and store Op and size data directly in InstallDiskSizeMatcher.

Closes #5860.


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1264,13 +1239,46 @@ func (m *InstallDiskSizeMatcher) UnmarshalYAML(unmarshal func(interface{}) error
...
 	return nil
 }
 
+// Matcher is a method that can handle some custom disk matching logic.
+func (m *InstallDiskSizeMatcher) Matcher(d *disk.Disk) bool {
+	return m.MatchData.Compare(d)
+}
+
+// MatchData contains data for comparison - Op and Size.
+// docgen:nodoc
```
DmitriyMV, line 1256: @smira I'm not sure about this line, is this correct usage?

smira:
```suggestion
//
//docgen:nodoc
```

this should be good style-wise, actually docgen is @Unix4ever baby

utkuozdemir (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
DmitriyMV: Done

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1264,13 +1239,46 @@ func (m *InstallDiskSizeMatcher) UnmarshalYAML(unmarshal func(interface{}) error
...
 
+// Matcher is a method that can handle some custom disk matching logic.
+func (m *InstallDiskSizeMatcher) Matcher(d *disk.Disk) bool {
+	return m.MatchData.Compare(d)
+}
+
+// MatchData contains data for comparison - Op and Size.
+//docgen:nodoc
+type MatchData struct {
+	Op   string
```
smira, line 1258: nit, but as this file contains lots of data structures, should we make it more specific? 

```suggestion
type InstallDiskSizeMatchData struct {
```

DmitriyMV: I agree, done!

smira (approved):

Unix4ever (approved):

PR #5975, opened 2022-07-26
feat: enable default seccomp profile for kubelet

### LINKED ISSUE #5293: enable seccomp profile by default

frezbo:
Enable the default seccomp profile provided by the container runtime

Ref: https://kubernetes.io/docs/tutorials/security/seccomp/

Fixes #5293 


frezbo:
Controller tests remaining

frezbo:
> Controller tests remaining

Ready for review

smira:
I think this should be a `feat:` as it's really a feature.

internal/app/machined/pkg/controllers/seccomp/seccomp.go
```diff
@@ -0,0 +1,169 @@
...
+			return fmt.Errorf("error listing seccomp profiles: %w", err)
+		}
+
+		for _, res := range list.Items {
+			if _, ok := touchedIDs[res.Metadata().ID()]; !ok {
+				if err := r.Destroy(ctx, res.Metadata()); err != nil {
+					return fmt.Errorf("error deleting seccomp profile: %w", err)
+				}
+
+				if err := os.Remove(filepath.Join(ctrl.SeccompProfileDirectory, res.Metadata().ID())); err != nil {
```
smira, line 163: we should probably ignore `not found` here to handle cases when the resource was created, but the file wasn't

smira: also, as `EPHEMERAL` persists across reboots, but resources do not persist, I think it makes sense to explicitly clean up seccomp profiles in this directory which we don't have in the resources.

internal/app/machined/pkg/controllers/seccomp/seccomp.go
```diff
@@ -0,0 +1,169 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package seccomp
```
smira, line 5: idk, whether it makes sense to keep that under `k8s.`? or make something more generic, like `cri.`? it feels there won't be much under `seccomp.` :)

internal/app/machined/pkg/controllers/seccomp/seccomp.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		break
+	}
+
+	// normal reconcile loop
+	if err := r.UpdateInputs([]controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineConfigType,
```
smira, line 97: currently this controller will always run on any `machineconfig` change. this is not the biggest problem, but we can avoid it by introducing an intermediate resource & controller which translate `MachineConfig` into a specific `k8s.SeccompProfileConfig` resource. This way this controller will watch `k8s.SeccompProfileConfig`, and it will not be triggered on machine config changes which do not contain changes to seccomp profiles

internal/app/machined/pkg/controllers/seccomp/seccomp.go
```diff
@@ -0,0 +1,169 @@
...
+				}
+
+				return nil
+			}); err != nil {
+				return err
+			}
+
+			touchedIDs[profile.Name()] = struct{}{}
+
+			if err = ioutil.WriteFile(filepath.Join(ctrl.SeccompProfileDirectory, profile.Name()), value, 0o644); err != nil {
```
smira, line 146: as I understand it, seccomp profiles might be used by any time for containers, so writing a file like this might temporariliy render the data unusable (as file is truncated and re-written).

what we can do is to wrap this write into two-phase step: first check if the file already exists, if it does exist, try to read it; if the content matches, we can skip any writes to the file

pkg/machinery/resources/seccomp/seccomp_profile.go
```diff
@@ -0,0 +1,47 @@
...
+//nolint:lll
+//go:generate deep-copy -type ProfileSpec -header-file ../../../../hack/boilerplate.txt -o deep_copy.generated.go .
+
+// ProfileType is type of Etcd resource.
+const ProfileType = resource.Type("SeccompProfiles.seccomp.talos.dev")
+
+// Profile represents the last Profile stats snapshot.
+type Profile = typed.Resource[ProfileSpec, ProfileRD]
+
+// ProfileSpec represents the last CPU stats snapshot.
```
smira, line 22: doc block is outdated

pkg/machinery/resources/seccomp/seccomp_profile.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+)
+
+//nolint:lll
+//go:generate deep-copy -type ProfileSpec -header-file ../../../../hack/boilerplate.txt -o deep_copy.generated.go .
+
+// ProfileType is type of Etcd resource.
```
smira, line 16: and same here - doc block is outdated

smira (commented):
this certainly needs a `hack/release.toml` note

do we need to try this out e.g. with Rook? would it work with default seccomp profile?

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: `k8s` doesn't seem quite right, since maybe a system extension could use it :thinking: or some other runtime, it's definitely something machine specific. Maybe makes more sense to be under cri/files

frezbo:
> do we need to try this out e.g. with Rook? would it work with default seccomp profile?

Our rook integration test passed

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: that's a good point, I'll also test that wiping ephemeral re-creates the files on next boot.

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: fixed

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: fixed, and added tests

pkg/machinery/resources/seccomp/seccomp_profile.go
frezbo: fixed

pkg/machinery/resources/seccomp/seccomp_profile.go
frezbo: fixed

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: fixed

internal/app/machined/pkg/controllers/seccomp/seccomp.go
frezbo: tested and verfied

frezbo:
Tested Cilium also works fine

internal/app/machined/pkg/controllers/cri/cri_test.go
```diff
@@ -0,0 +1,100 @@
...
+	wg      sync.WaitGroup
+
+	//nolint:containedctx
+	ctx       context.Context
+	ctxCancel context.CancelFunc
+
+	seccompProfileDirectory string
+}
+
+func (suite *CRISuite) SetupTest() {
```
smira, line 39: there's new `ctest` framework by @DmitriyMV 

example: https://github.com/siderolabs/talos/blob/ae1bec59e9ce1a9b5f3a2f6a040a16bb26d991a7/internal/app/machined/pkg/controllers/secrets/kubernetes_cert_sans_test.go#L33-L42

we could use it here to give good example and also have better code

smira (approved):
:cool:

internal/app/machined/pkg/controllers/cri/cri_test.go
frezbo: oh that's nice, updating

internal/app/machined/pkg/controllers/cri/cri_test.go
```diff
@@ -0,0 +1,35 @@
...
+	seccompProfileDirectory string
+}
+
+func TestSeccompProfileSuite(t *testing.T) {
+	seccompProfiesDirectory := t.TempDir()
+
+	suite.Run(t, &CRISuite{
+		DefaultSuite: ctest.DefaultSuite{
+			AfterSetup: func(suite *ctest.DefaultSuite) {
+				suite.Require().NoError(suite.Runtime().RegisterController(&cri.SeccompProfileController{}))
```
frezbo, line 27: this feels like starting both controllers for a single controller test, wdyt @smira

internal/app/machined/pkg/controllers/cri/seccomp_profile_file_test.go
```diff
@@ -0,0 +1,106 @@
...
+		auditJSONSeccompProfileAfterUpdate, err := os.Stat(suite.seccompProfileDirectory + "/audit.json")
+		if err != nil {
+			if errors.Is(err, os.ErrNotExist) {
+				return retry.ExpectedError(err)
+			}
+
+			return err
+		}
+
+		suite.Eventually(func() bool {
```
frezbo, line 90: just checking the file seems to fail and had to use `suite.Eventually`. Seems using `filepath.WalkDir` is a little slow

internal/app/machined/pkg/controllers/cri/cri_test.go
frezbo: fixed

PR #5982, opened 2022-07-28
refactor: make apid stop gracefully and be stopped late

smira:
This fixes apid and machined shutdown sequences to do graceful stop of
gRPC server with timeout.

Also sequences are restructured to stop apid/machined as late as
possible allowing access to the node while the long sequence is running
(e.g. upgrade or reset).

### DIFF
--- a/internal/app/apid/main.go
+++ b/internal/app/apid/main.go
@@ -7,8 +7,12 @@ package apid
 import (
 	"context"
 	"flag"
+	"fmt"
 	"log"
+	"os/signal"
 	"regexp"
+	"syscall"
+	"time"
 
 	"github.com/cosi-project/runtime/api/v1alpha1"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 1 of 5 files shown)


internal/app/apid/main.go
```diff
@@ -46,39 +50,48 @@ func runDebugServer(ctx context.Context) {
...
+	defer cancel()
+
 	log.SetFlags(log.Lshortfile | log.Ldate | log.Lmicroseconds | log.Ltime)
 
 	rbacEnabled = flag.Bool("enable-rbac", false, "enable RBAC for Talos API")
 
 	flag.Parse()
 
-	go runDebugServer(context.TODO())
+	go runDebugServer(ctx)
```
frezbo, line 68: any reason we called this debugserver?

smira: because it's really debug server enabled `WITH_DEBUG=1`, it basically provides e.g. Go profiling

frezbo (approved):

PR #5986, opened 2022-07-28
chore: add gotagsrewrite tool and use it to add tags to resources

### LINKED ISSUE #5941: Automatic rewriting of struct tags for Spec types
We should automate the the process of adding `protobuf:"<n>"` tags to target structures. For that we should get a tool which should do the following:
- [ ] Identify which structs it needs to update
- [ ] Add `protobuf:"<n>"` to the exported fields of those structs, where `<n>` is an increasing number starting from 1. It should not add this tags to private fields.
- [ ] Update existing structures which already contain `protobuf:"<n>"` on some of the fields. Starting `<n>` in that case should be biggest `<n>` +1.
- [ ] Do not touch fields with `protobuf:"-"` which should be ignored during encoding/decoding process.

DmitriyMV:
chore: add gotagsrewrite tool and use it to add tags to resources

This commit adds gotagsrewrite tool, which is used to add `protobuf:"<n>"` tags to structs with `//gotagsrewrite:gen` comment. This will be used in conjunction with github.com/siderolabs/protoenc.

Closes #5941


hack/gotagsrewrite/main.go
```diff
@@ -0,0 +1,312 @@
...
+	tokenSet := token.NewFileSet()
+
+	for _, path := range paths {
+		parsedAST, err := parseAst(tokenSet, path) //nolint:govet
+		if err != nil {
+			return fmt.Errorf("failed to parse AST for %s: %w", path, err)
+		}
+
+		structs := filterStructWithFields(parsedAST, func(n *ast.TypeSpec, t *ast.StructType) bool {
+			return strings.HasSuffix(n.Name.Name, "SpecSpec") && isCapitalCase(n.Name.Name)
```
smira, line 58: I feel this doesn't cover _all_ of the structs we need to handle. Plus there might be embedded structs into the ones we analyze?

we can also do some kind of docblock tag to tag the specs, and anything which is embedded should be also covered?

frezbo:
don't we possible need to add protobuf tags to other projects too, if that's the case doesn't it make sense to make the cli generic?

smira:
> don't we possible need to add protobuf tags to other projects too, if that's the case doesn't it make sense to make the cli generic?

it's not clear right now, but if we do some kind of `//protostructtags` docblock, this will be generic enough

hack/gotagsrewrite/main.go
smira: we could also look for whatever uses `typed.Resource[<Spec>]`

DmitriyMV: > we can also do some kind of docblock tag to tag the specs

I think it's a best approach, since its easy enough and quite flexible. It requires _some_ manual work, but nothing particularly hard.

DmitriyMV:
@smira @frezbo I updated PR description - I think `//gotagsrewrite:gen` is a good way forward.

hack/gotagsrewrite/main.go
frezbo: > we could also look for whatever uses typed.Resource[<Spec>]

i think we then need to slowly update all controllers to use this, since the old ones doesn't used the typed.Resource

smira (approved):

Issue #5889, opened 2022-07-11
etcd join and certificate fixes


smira:
fix a number of problems observed with etcd today:

1. join sequence - joining with the right hostname and certs
2. promotion to full member - figure out avoiding learner members
3. certificate generation - generate in the controller and update in the runtime as etcd supports reloading

PR #5988, opened 2022-07-29
fix: iterate over etcd members endpoints for member promotion

smira:
This uses all available (potential) etcd endpoints, which includes the
member being promoted as well. We avoid failures by iterating over the
list of endpoints on each attempt to make sure each and every endpoint
is tried.

Part of #5889

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -305,7 +305,7 @@ func buildInitialCluster(ctx context.Context, r runtime.Runtime, name, ip string
 		retry.WithErrorLogging(true),
 	).RetryWithContext(ctx, func(ctx context.Context) error {
 		var (
-			peerAddrs = []string{"https://" + net.FormatAddress(ip) + ":2380"}
+			peerAddrs = []string{"https://" + nethelpers.JoinHostPort(ip, +constants.EtcdPeerPort)}
 			resp      *clientv3.MemberListResponse
 		)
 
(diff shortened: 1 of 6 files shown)


smira:
Also consistently use constants/clean up etcd peer URL construction.

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -600,24 +600,60 @@ func (e *Etcd) recoverFromSnapshot(hostname, primaryAddr string) error {
...
-		defer client.Close() //nolint:errcheck
+		if len(endpoints) == 0 {
+			return retry.ExpectedErrorf("no endpoints")
+		}
+
+		// try to iterate all available endpoints in the time available for an attempt
+		for i := 0; i < len(endpoints); i++ {
+			select {
+			case <-ctx.Done():
+				return retry.ExpectedError(ctx.Err())
```
frezbo, line 627: is this expected, retrying when the context is done?

frezbo (approved):

internal/app/machined/pkg/system/services/etcd.go
smira: so I was trying to do the following logic here:

* `ctx` in the retry loop  (in `RetryContext`) version is actually set with the deadline of the attempt timeout (i.e. 15 seconds in this case)
* so if I return not expected error, `RetryContext` loop will abort
* what I want to do: during those 15 seconds allocated for an attempt, I want to try as many different etcd endpoints as I can

Issue #5980, opened 2022-07-28
Talos API access through an authentication proxy


smira:
- server CA validation: might be nothing, if the server CA is publicly trusted
- several client authentication schemes:
   - (as it is today): client cert + key
   - Basic Auth
   - no auth (for development purposes)
   - other types auth (TBD)
 - optional `cluster` field to target a specific cluster behind a proxy
 - endpoint becomes endpoint of the authentication proxy

`talosconfig` today:

```yaml
    talos-default-457:
        endpoints:
            - 172.20.0.2
        ca: LS0tLS1CRUdJTiBDRV
        crt: LS0tLS1CRUd
        key: LS0t
```

`talosconfig` for going through an auth proxy:

```yaml
    talos-via-proxy:
        endpoints:
            - [https://]auth.proxy:453  # scheme is optional, `grpc://` for HTTP/2 h2c
        ca: LS0tLS1CRUdJTiBDRV # optional
        auth:
           basic:
              username: XYZ
              password: UIO
        cluster: my-lovely-cluster-1
```

`talosctl` new flags:

- `--cluster` flag

stereobutter:
Currently `talosctl` has port 50000 hardcoded for accessing the api. Even without additional options for client authentication it would be useful if `endpoints` could be fully specified by the user e.g. `talosctl stats --endpoints my-cluster-behind-a-loadbalancer:443  --nodes my-node`

smira:
This will be available as part of the work here.

PR #6004, opened 2022-08-01
feat: read talosconfig from secrets directory

utkuozdemir:
Similar to the way kubectl reads kubeconfig, we attempt to load talosconfig file from multiple locations. If the file exists under `/var/run/secrets/talos.dev/config`, we load with higher priority before falling back to `~/.talos/config`. This will allow talosctl to be able to access Talos API from inside a pod when talosconfig is mounted into `/var/run/secrets/talos.dev/config`, similar to the way Kubernetes service account tokens work.

Part of siderolabs/talos#5980.

### DIFF
--- a/pkg/machinery/client/config/path.go
+++ b/pkg/machinery/client/config/path.go
@@ -27,6 +27,17 @@ func GetDefaultPath() (string, error) {
 		return path, nil
 	}
 
+	talosSAPath := filepath.Join(constants.ServiceAccountMountPath, constants.ServiceAccountTalosconfigFilename)
+
+	_, err := os.Stat(talosSAPath)
+	if err != nil && !os.IsNotExist(err) && !os.IsPermission(err) {
+		return "", err
+	}
+
+	if err == nil {
+		return talosSAPath, nil
+	}
+
 	talosDir, err := GetTalosDirectory()
 	if err != nil {
 		return "", err
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -704,6 +704,12 @@ const (
 
 	// KubernetesTalosProvider is the name of the Talos provider as a Kubernetes label.
 	KubernetesTalosProvider = "talos.dev"
+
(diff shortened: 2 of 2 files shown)


pkg/machinery/client/config/path.go
```diff
@@ -27,6 +27,17 @@ func GetDefaultPath() (string, error) {
...
 	}
 
+	talosSAPath := filepath.Join(constants.ServiceAccountMountPath, constants.ServiceAccountTalosconfigFilename)
+
+	_, err := os.Stat(talosSAPath)
+	if err != nil && !os.IsNotExist(err) {
+		return "", err
+	}
+
+	if err == nil {
```
frezbo, line 37: isn't this redundant as the previous if block handle it? :thinking:

frezbo (approved):

pkg/machinery/client/config/path.go
utkuozdemir: Previous block only handles it if it is something else than `NotExist` error. Here, it can still be `NotExist`. If it exists, we return. Otherwise we let it fallback to `~/.talos/config`.

frezbo: ahh I missed that completely :sweat_smile: :+1:

pkg/machinery/client/config/path.go
```diff
@@ -27,6 +27,17 @@ func GetDefaultPath() (string, error) {
 		return path, nil
 	}
 
+	talosSAPath := filepath.Join(constants.ServiceAccountMountPath, constants.ServiceAccountTalosconfigFilename)
+
+	_, err := os.Stat(talosSAPath)
+	if err != nil && !os.IsNotExist(err) {
```
smira, line 33: we should ignore more errors here, e.g. it could be 'pemission denied' if the `/var` is not readable for the user.

I would argue we should ignore all errors? but at the very least permission denied should be in the list

if there's some other error reading `/var` we can stop here and never read the `~/.talos/config`

utkuozdemir: I also added checking for permission errors.

PR #6005, opened 2022-08-01
fix: keep entire vlan id when parsing cmdline

EirikAskheim:
# Pull Request

## What? (description)
When parsing vlan id from cmdline only the last digit is kept. For example parsing vlan=eth0.169:eth0, vlan id gets set to 9 instead of 169.
## Acceptance


internal/app/machined/pkg/controllers/network/cmdline.go
```diff
@@ -265,12 +265,11 @@ func ParseCmdlineNetwork(cmdline *procfs.Cmdline) (CmdlineNetworking, error) {
 			return settings, fmt.Errorf("malformed vlan commandline argument: %s", *vlanSettings)
 		}
 
-		numStart := strings.LastIndexFunc(vlanName, func(c rune) bool { return c >= '0' && c <= '9' })
```
frezbo, line 268: just changing `LastIndexFunc` to `IndexFunc` should be sufficient I believe

EirikAskheim: Format is like "eth0:169" or something like that. Won't IndexFunc find the 0 in eth0? I just tested and test fails with NetworkLinkSpecs: ([]network.LinkSpecSpec) <nil>

frezbo (approved):

frezbo:
would you mind signing off the commit? You can do with `git commit -s --amend --no-edit` and doing a `git push --force-with-lease`. Don't worry about the gpg check failure. It would be nice if you could rebase (else we can fix that while signing the commit for final merge).

smira (approved):

Issue #5342, opened 2022-04-13
talosctl reset doesn't work on not fully joined nodes


smira:
`talosctl reset` should be able to wipe a node even if the Kubernetes control plane is not available, or some services failed to start. It shouldn't fail on shutdown sequence.

Sequence concurrency changes: #5450 

We should make sure that reset sequence can skip steps which are not applicable (e.g. no way stop pods if CRI is not running/healthy, and try not to fail on the steps which are not critical).

Before actually wiping partitions (or wiping a disk), try to use the code from the reboot machined/main.go handler which kills all processes and unmounts all filesystems.

> Note: in integration tests we can drop the check on 'boot sequence done' before attempting a reset.

> We can also try to implement a test which resets a node after a reboot (or reset) at some random interval before node fully boots/joins.

szinn:
Perhaps reset should also support "--insecure"?

smira:
There should be nothing to wipe in maintenance mode (only Talos installation if booted from disk image).

But our concern was that as maintenance mode is wide open to the world, anything could connect to the port and issue a wipe which doesn't feel solid.

szinn:
In this case a config has been applied to the node and if the machine reboots, it continues on with that config. reset doesn't work unless a control plane node comes up to relay commands to it

smira:
I'm not sure which scenario are you talking about? Wiping a worker node when control plane nodes are down? Wiping a node when Kubernetes is not fully up?

szinn:
It is when a config has been applied to a worker node but no control plane nodes ever come up. So the worker is sitting around waiting for a control plane to come up to join and won't respond to a reset given to it via talosctl directly.

smira:
It should still respond to something like `talosctl reset -e <WORKER> -n <WORKER> reset` (setting the endpoint directly to the node in question).

Unix4ever:
Additional thoughts on the reset flow:
- we need to disable running sequence in parallel.
- need to have a way to cancel currently running sequence.
- need to have `BootstrapStatus` resource or something similar. Right now reset will try to stop drain and cordon the node. It should know when it should skip this step (of course we can check if the folders exist but it's not a great approach).

szinn:
The reset can be done by specifying -n node -e node.

It wasn't working for me earlier because talosctl only had the control planes as endpoints and the control planes hadn't been able to be started, thus the worker node was not reachable. Specifying -n worker_node -e worker_node does reset it.

PR #6008, opened 2022-08-01
fix: make reset work even if the node is not bootstrapped/not joined

Unix4ever:
Fixes: https://github.com/siderolabs/talos/issues/5342

Now the sequencer is smart enough to skip `LeaveEtcd` and
`CordonAndDrain` node if the node is not fully joined to the cluster
yet.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
@@ -284,15 +284,15 @@ func (*Sequencer) Reset(r runtime.Runtime, in runtime.ResetOptions) []runtime.Ph
 		phases = phases.AppendWhen(
 			in.GetGraceful(),
 			"drain",
-			CordonAndDrainNode,
+			taskErrorHandler(logError, CordonAndDrainNode),
 		).AppendWhen(
 			in.GetGraceful(),
 			"cleanup",
-			RemoveAllPods,
+			taskErrorHandler(logError, RemoveAllPods),
 		).AppendWhen(
 			!in.GetGraceful(),
 			"cleanup",
-			StopAllPods,
+			taskErrorHandler(logError, StopAllPods),
 		).Append(
 			"dbus",
 			StopDBus,
@@ -301,7 +301,10 @@ func (*Sequencer) Reset(r runtime.Runtime, in runtime.ResetOptions) []runtime.Ph
 			"leave",
 			LeaveEtcd,
 		).AppendList(
-			stopAllPhaselist(r, withKexec),
+			phaseListErrorHandler(logError, stopAllPhaselist(r, withKexec)...),
+		).Append(
+			"forceCleanup",
+			ForceCleanup,
 		).AppendWhen(
 			len(in.GetSystemDiskTargets()) == 0,
 			"reset",
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -33,6 +33,7 @@ import (
 	"github.com/talos-systems/go-blockdevice/blockdevice/partition/gpt"
 	"github.com/talos-systems/go-blockdevice/blockdevice/util"
 	"github.com/talos-systems/go-cmd/pkg/cmd"
+	"github.com/talos-systems/go-cmd/pkg/cmd/proc"
 	"github.com/talos-systems/go-kmsg"
 	"github.com/talos-systems/go-procfs/procfs"
 	"github.com/talos-systems/go-retry/retry"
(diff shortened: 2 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1998,3 +1998,37 @@ func pauseOnFailure(callback func(runtime.Sequence, interface{}) (runtime.TaskEx
 		}, name
 	}
 }
+
+func canLeaveEtcd() bool {
+	etcdID := (&services.Etcd{}).ID(nil)
```
smira, line 2003: I feel we should move the check into the task if possible; first of all, it makes it closer to the action in time, second, we might try to ignore failures (?)

we could e.g. check for the existence of etcd data directory for example

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1998,3 +1998,37 @@ func pauseOnFailure(callback func(runtime.Sequence, interface{}) (runtime.TaskEx
...
+		case events.StateFailed:
+			return true
+		}
+	}
+
+	return false
+}
+
+func canCordonNode() bool {
+	if _, err := os.Stat("/var/lib/kubelet/pki/kubelet-client-current.pem"); err != nil {
```
smira, line 2027: cordoning might fail with other failures, e.g. if the API server is not responsive, so once again we should check more thoroughly ignoring the errors as needed

smira (commented):
we should also do some error handling around stopping/removing all pods

we should also add as the very last step before actually wiping the disk:

https://github.com/siderolabs/talos/blob/cf620d4733d4bd4c243213676caf7c5fe6a20d9d/internal/app/machined/main.go#L156-L163

this should attempt to stop/unmount everything

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Unix4ever: I've made it ignore `Timeout` and `ServerTimeout` errors.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
Unix4ever: Moved, added folder check (same as in CACPPT)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -301,7 +301,7 @@ func (*Sequencer) Reset(r runtime.Runtime, in runtime.ResetOptions) []runtime.Ph
 			"leave",
 			LeaveEtcd,
 		).AppendList(
-			stopAllPhaselist(r, withKexec),
+			withErrorHandler(stopAllPhaselist(r, withKexec), forceCleanup),
```
smira, line 304: this feels wrong to me, as it adds `forceCleanup` to each task hiding the original error?

I meant adding `forceCleanup` as a step here before actually resetting things

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1263,7 +1275,14 @@ func CordonAndDrainNode(seq runtime.Sequence, data interface{}) (runtime.TaskExe
 
 		defer kubeHelper.Close() //nolint:errcheck
 
-		return kubeHelper.CordonAndDrain(ctx, nodename)
+		err = kubeHelper.CordonAndDrain(ctx, nodename)
+		if _, ok := data.(runtime.ResetOptions); ok {
+			if apierrors.IsTimeout(err) || apierrors.IsServerTimeout(err) {
```
smira, line 1280: I guess this is not exhaustive list of errors, it might e.g. "connection refused" easily? 

should we in general report, but ignore errors here?

smira (commented):
we still don't have a handler around CRI stopping all pods

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1249,6 +1251,15 @@ func UnmountSystemDiskBindMounts(seq runtime.Sequence, data interface{}) (runtim
 // k8s.io namespace.
 func CordonAndDrainNode(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
+		// skip not exist error as it means tha node hasn't fully joined yet
```
smira, line 1254:
```suggestion
		// skip not exist error as it means that the node hasn't fully joined yet
```

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1998,3 +2065,34 @@ func pauseOnFailure(callback func(runtime.Sequence, interface{}) (runtime.TaskEx
...
+		for i, task := range phase.Tasks {
+			phase.Tasks[i] = taskErrorHandler(handler, task)
+		}
+	}
+
+	return phases
+}
+
+func logError(name string, err error, logger *log.Logger) error {
+	logger.Printf("WARNING: task %s failed: %s", name, err)
```
smira, line 2095: nit, but I think `logger` already has task name in its prefix set

smira (approved):

Issue #4422, opened 2021-10-22
accessing Talos API from Kubernetes workloads


smira:
- [ ] Talos auth controller/mutating webhook which based on pod annotations injects a configmap volume mount which contains generated (and refreshed) `talosconfig` with RBAC tied to Kubernetes RBAC for the Service Account
- [x] service for the Talos API in Kubernetes: endpoints, service which talks to endpoints
- [ ] map Talos API RBAC to Kubernetes RBAC (use CRDs?)
- [ ] have a way to limit Talos RBAC roles user can get, default to empty (disabled by default, requires Kubernetes RBAC to be set up properly)
- [ ] final user experience should be that `talosctl` should work out of the box if the RBAC is configured

dimatha:
Hi Guys, I just wanted to know how far we are from reaching this goal? Thanks!

smira:
@dimatha no commitment yet, but we plan on delivering that in the next release 1.1

smira:
See also: https://github.com/siderolabs/rfcs/pull/9

PR #6010, opened 2022-08-02
feat: introduce Talos API access from Kubernetes

utkuozdemir:
We add a new CRD, `serviceaccounts.talos.dev` (with `tsa` as short name), and its controller which allows users to get a `Secret` containing a short-lived Talosconfig in their namespaces with the roles they need. Additionally, we introduce the `talosctl inject serviceaccount` command to accept a YAML file with Kubernetes manifests and inject them with Talos service accounts so that they can be directly applied to Kubernetes afterwards. If Talos API access feature is enabled on Talos side, the injected workloads will be able to talk to Talos API.

Closes siderolabs/talos#4422.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -14,6 +14,32 @@ preface = """\
 """
 
 [notes]
+    [notes.api-access-from-kubernetes]
+        title = "Talos API access from Kubernetes"
+        description = """\
+Talos now supports access to its API from within Kubernetes. It can be configured in the machine config as below:
+```yaml
+machine:
+  features:
+    kubernetesTalosAPIAccess:
+      enabled: true
+      allowedRoles:
+        - os:reader
+      allowedKubernetesNamespaces:
+        - kube-system
+```
+
+This feature introduces a new custom resource definition, `serviceaccounts.talos.dev`.
+Creating custom resources of this type will provide credentials to access Talos API from within Kubernetes.
+
+The new CLI subcommand `talosctl inject serviceaccount` can be used to configure Kubernetes manifests with Talos service accounts as below:
+```
+talosctl inject serviceaccount -f manifests.yaml > manifests-injected.yaml
+kubectl apply -f manifests-injected.yaml
+```
+
+See [documentation](https://www.talos.dev/v1.2/advanced/configuration/talos-api-access-from-k8s/) for more details.
+"""
 
     [notes.seccomp]
         title = "Seccomp Profiles"
--- a/internal/integration/api/serviceaccount.go
+++ b/internal/integration/api/serviceaccount.go
@@ -1,0 +1,303 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_api
+// +build integration_api
+
+package api
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/siderolabs/go-pointer"
+	"github.com/talos-systems/go-retry/retry"
+	corev1 "k8s.io/api/core/v1"
+	eventsv1 "k8s.io/api/events/v1"
+	kubeerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/talos-systems/talos/internal/integration/base"
+	machineapi "github.com/talos-systems/talos/pkg/machinery/api/machine"
+	"github.com/talos-systems/talos/pkg/machinery/client"
+	"github.com/talos-systems/talos/pkg/machinery/client/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+	"github.com/talos-systems/talos/pkg/machinery/constants"
+)
+
+var (
+	serviceAccountGVR = schema.GroupVersionResource{
+		Group:    constants.ServiceAccountResourceGroup,
+		Version:  constants.ServiceAccountResourceVersion,
+		Resource: constants.ServiceAccountResourcePlural,
+	}
+	secretGVR = schema.GroupVersionResource{
+		Group:    "",
+		Version:  "v1",
+		Resource: "secrets",
+	}
+)
+
+// ServiceAccountSuite verifies Talos ServiceAccount.
+type ServiceAccountSuite struct {
+	base.K8sSuite
+
+	ctx       context.Context //nolint:containedctx
+	ctxCancel context.CancelFunc
+}
+
+// SuiteName ...
+func (suite *ServiceAccountSuite) SuiteName() string {
+	return "api.ServiceAccountSuite"
+}
+
+// SetupTest ...
+func (suite *ServiceAccountSuite) SetupTest() {
+	// make sure API calls have timeout
+	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 5*time.Minute)
+}
+
+// TearDownTest ...
+func (suite *ServiceAccountSuite) TearDownTest() {
+	if suite.ctxCancel != nil {
+		suite.ctxCancel()
+	}
+}
+
+// TestValid tests Kubernetes service accounts.
+//
+//nolint:dupl
+func (suite *ServiceAccountSuite) TestValid() {
+	name := "test-valid"
+
+	err := suite.configureAPIAccess(true, []string{"os:reader"}, []string{"kube-system"})
+	suite.Assert().NoError(err)
+
+	_, err = suite.getCRD()
+	suite.Assert().NoError(err)
+
+	sa, err := suite.createServiceAccount("kube-system", name, []string{"os:reader"})
+	suite.Assert().NoError(err)
+
+	defer suite.DeleteResource(suite.ctx, serviceAccountGVR, "default", name) // nolint:errcheck
+
+	err = suite.WaitForEventExists(suite.ctx, "kube-system", func(event eventsv1.Event) bool {
+		return event.Regarding.UID == sa.GetUID() &&
+			event.Type == corev1.EventTypeNormal &&
+			event.Reason == "Synced"
+	})
+	suite.Assert().NoError(err)
+
+	secret, err := suite.waitForSecret("kube-system", name)
+	suite.Assert().NoError(err)
+
+	talosConfig := secret.Data["config"]
+
+	conf, err := config.FromBytes(talosConfig)
+	suite.Assert().NoError(err)
+
+	expectedServiceName := fmt.Sprintf(
+		"%s.%s",
+		constants.KubernetesTalosAPIServiceName,
+		constants.KubernetesTalosAPIServiceNamespace,
+	)
+	suite.Assert().Equal([]string{expectedServiceName}, conf.Contexts[conf.Context].Endpoints)
+
+	err = suite.DeleteResource(suite.ctx, serviceAccountGVR, "kube-system", name)
+	suite.Require().NoError(err)
+
+	err = suite.EnsureResourceIsDeleted(suite.ctx, 30*time.Second, secretGVR, "kube-system", name)
+	suite.Assert().NoError(err)
+}
+
+// TestNotAllowedNamespace tests Kubernetes service accounts in not allowed namespaces.
+//
+//nolint:dupl
+func (suite *ServiceAccountSuite) TestNotAllowedNamespace() {
+	name := "test-allowed-ns"
+
+	err := suite.configureAPIAccess(true, []string{"os:reader"}, []string{"kube-system"})
+	suite.Assert().NoError(err)
+
+	sa, err := suite.createServiceAccount("default", name, []string{"os:reader"})
+	suite.Assert().NoError(err)
+
+	defer suite.DeleteResource(suite.ctx, serviceAccountGVR, "default", name) // nolint:errcheck
+
+	err = suite.WaitForEventExists(suite.ctx, "default", func(event eventsv1.Event) bool {
+		return event.Regarding.UID == sa.GetUID() &&
+			event.Type == corev1.EventTypeWarning &&
+			event.Reason == "ErrNamespaceNotAllowed"
+	})
+	suite.Assert().NoError(err)
+}
+
+// TestNotAllowedRoles tests Kubernetes service accounts with not allowed roles.
+//
+//nolint:dupl
+func (suite *ServiceAccountSuite) TestNotAllowedRoles() {
+	name := "test-not-allowed-roles"
+
+	err := suite.configureAPIAccess(true, []string{"os:reader"}, []string{"kube-system"})
+	suite.Assert().NoError(err)
+
+	sa, err := suite.createServiceAccount("kube-system", name, []string{"os:admin"})
+	suite.Assert().NoError(err)
+
+	defer suite.DeleteResource(suite.ctx, serviceAccountGVR, "kube-system", name) // nolint:errcheck
+
+	err = suite.WaitForEventExists(suite.ctx, "kube-system", func(event eventsv1.Event) bool {
+		return event.Regarding.UID == sa.GetUID() &&
+			event.Type == corev1.EventTypeWarning &&
+			event.Reason == "ErrRolesNotAllowed"
+	})
+	suite.Assert().NoError(err)
+}
+
+// TestFeatureNotEnabled tests Kubernetes service accounts when API access feature is not enabled.
+//
+//nolint:dupl
+func (suite *ServiceAccountSuite) TestFeatureNotEnabled() {
+	name := "test-feature-not-enabled"
+
+	err := suite.configureAPIAccess(false, []string{"os:reader"}, []string{"kube-system"})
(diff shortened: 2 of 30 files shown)


internal/app/machined/pkg/controllers/kubeaccess/crd.go
```diff
@@ -0,0 +1,207 @@
...
+// Outputs implements controller.Controller interface.
+func (ctrl *CRDController) Outputs() []controller.Output {
+	return nil
+}
+
+// Run implements controller.Controller interface.
+//
+//nolint:gocyclo,cyclop
+func (ctrl *CRDController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	if err := r.UpdateInputs([]controller.Input{
```
Unix4ever, line 52: As these inputs are static, can we move them to `Inputs()` return?
Or add the comment why we need it in the `Run` call.

internal/app/machined/pkg/controllers/kubeaccess/crd.go
```diff
@@ -0,0 +1,207 @@
...
+
+				<-crdControllerErrCh
+
+				crdControllerCtxCancel = nil
+			}
+
+			continue
+		}
+
+		kubeSecretsResources, err := r.Get(ctx, resource.NewMetadata(secrets.NamespaceName, secrets.KubernetesType, secrets.KubernetesID, resource.VersionUndefined))
```
Unix4ever, line 133:
```suggestion
		kubeSecretsResources, err := safe.ReaderGet[*secrets.Kubernetes](ctx, r, resource.NewMetadata(secrets.NamespaceName, secrets.KubernetesType, secrets.KubernetesID, resource.VersionUndefined))
```

internal/app/machined/pkg/controllers/kubeaccess/crd.go
```diff
@@ -0,0 +1,207 @@
...
+			if !state.IsNotFoundError(err) {
+				return fmt.Errorf("error fetching kubernetes secrets: %w", err)
+			}
+
+			continue
+		}
+
+		kubeSecretsSpec := kubeSecretsResources.(*secrets.Kubernetes).TypedSpec()
+
+		osSecretsResource, err := r.Get(ctx, resource.NewMetadata(secrets.NamespaceName, secrets.OSRootType, secrets.OSRootID, resource.VersionUndefined))
```
Unix4ever, line 144: same as for `kubeSecret`

internal/app/machined/pkg/controllers/kubeaccess/crd.go
```diff
@@ -0,0 +1,207 @@
...
+			<-crdControllerErrCh
+
+			crdControllerCtxCancel = nil
+		}
+
+		var crdControllerCtx context.Context
+
+		crdControllerCtx, crdControllerCtxCancel = context.WithCancel(ctx) //nolint:govet
+
+		go func() {
```
Unix4ever, line 174: I think you can replace it with `errgroup.Group`:
```go
eg.Go(func() error {
  err := ctrl.runCRDController(...)
  if errors.is(err, context.Canceled) {
    return nil
  }

  return
}
```

Then when stopping (can also extract to common function):

```go
stopCrdController := func() error {
  if crdControllerCtxCancel != nil {
    crdControllerCtxCancel()
    crdControllerCtxCancel = nil

    return eg.Wait()
  }
}

...
if err = stopCrdController(); err != nil {
  return err
}
```

Not a big difference, but I guess using errgroups is preferred style wise :)

internal/app/machined/pkg/controllers/kubeaccess/crd.go
utkuozdemir: I did it this way because @smira did it the same way in `endpoint.go` - maybe he can comment on this.

internal/app/machined/pkg/controllers/kubeaccess/crd.go
utkuozdemir: Hm, nice, this was tricky so I took the it as-is from the `discovery_service.go` but this makes sense, let me try to adapt.

internal/app/machined/pkg/controllers/kubeaccess/talossa/crd_controller.go
```diff
@@ -0,0 +1,633 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package talossa
```
utkuozdemir, line 5: Note to reviewer: This is basically the same thing as https://github.com/kubernetes/sample-controller/blob/master/controller.go , except this one uses dynamic k8s client & unstructured to avoid code generation for the CRD.

internal/app/machined/pkg/controllers/kubeaccess/crd.go
utkuozdemir: As we discussed, unfortunately this does not work since we also use the err channel in select as well. But  👍 on reusing the common cancel code.

website/content/v1.2/advanced/talos-api-access-from-k8s.md
```diff
@@ -0,0 +1,157 @@
...
+          name: talos-secrets
+      tolerations:
+      - operator: Exists
+      volumes:
+      - name: talos-secrets
+        secret:
+          secretName: talos-api-access-talos-secrets
+status: {}
+---
+apiVersion: talos.dev/v1alpha1
```
frezbo, line 112: since we are doing a serviceaccount, couldn't we skip doing volume mounts and use `serviceAccountName` in the pod spec?

utkuozdemir: We do not create k8s serviceaccounts, we have our own different serviceaccount with `serviceaccounts.talos.dev` type (easier if we call them "Talos Service Accounts"). We just provision secrets based on `serviceaccounts.talos.dev` custom resources.

utkuozdemir: To further clarify, I did not name the CRD `talosserviceaccounts` because `talos` is redundant - it is also in the group name `talos.dev`.

cmd/talosctl/cmd/mgmt/kube_inject.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil
+	},
+}
+
+func init() {
+	kubeInjectCmd.Flags().StringVarP(&kubeInjectCmdFlags.file, "file", "f", "-",
+		fmt.Sprintf("file with Kubernetes manifests to be injected with %s", constants.ServiceAccountResourceKind))
+	kubeInjectCmd.Flags().StringSliceVarP(&kubeInjectCmdFlags.roles, "roles", "r", []string{"os:reader"},
+		fmt.Sprintf("roles to add to the generated %s manifests", constants.ServiceAccountResourceKind))
+	addCommand(kubeInjectCmd)
```
smira, line 56: it might make sense to put this under `talosconfig`, i.e. `talosctl talosconfig kube-inject`

hack/release.toml
```diff
@@ -14,6 +14,32 @@ preface = """\
...
+    kubernetesTalosAPIAccess:
+      enabled: true
+      allowedRoles:
+        - os:reader
+      allowedKubernetesNamespaces:
+        - kube-system
+```
+
+This feature introduces a new custom resource definition, `serviceaccounts.talos.dev`,
+which can be created by the user with the Talos roles they require.
```
smira, line 33: I think this needs to be clarified. We cant users to create resources of this type, not CRD itself.

smira (commented):
I will keep reviewing this one, it's big!

hack/release.toml
utkuozdemir: Ah yes, let me try to rephrase this.

cmd/talosctl/cmd/mgmt/kube_inject.go
utkuozdemir: Hm, I think it can be too long that way. Also, `talosconfig` command would be kind of redundant. But if you are sure, it's fine by me.

utkuozdemir: Corrected it and hopefully made it more clear.

hack/release.toml
utkuozdemir: Corrected it and hopefully made it more clear.

internal/app/machined/pkg/controllers/kubeaccess/crd.go
```diff
@@ -0,0 +1,204 @@
...
+		osSecretsSpec := osSecretsResource.TypedSpec()
+
+		kubeconfig, err := clientcmd.BuildConfigFromKubeconfigGetter("", func() (*clientcmdapi.Config, error) {
+			return clientcmd.Load([]byte(kubeSecretsSpec.LocalhostAdminKubeconfig))
+		})
+		if err != nil {
+			return fmt.Errorf("error loading kubeconfig: %w", err)
+		}
+
+		stopCRDController()
```
frezbo, line 165: why is the controller being stopped here?

internal/integration/api/serviceaccount.go
```diff
@@ -0,0 +1,288 @@
...
+				return retry.ExpectedError(err)
+			}
+
+			return nil
+		})
+	}
+
+	return nil
+}
+
```
frezbo, line 285: shouldn't we have a test to check that the k8s `endpoint` has been created?

website/content/v1.2/advanced/talos-api-access-from-k8s.md
```diff
@@ -0,0 +1,157 @@
...
+        app: talos-api-access
+    spec:
+      containers:
+        - name: talos-api-access
+          image: alpine:3
+          command:
+            - sh
+            - -c
+            - |
+              wget -O /usr/local/bin/talosctl https://github.com/siderolabs/talos/releases/download/<talos version>/talosctl-linux-amd64
```
frezbo, line 58: we could use the hugo release variable here so that the talosctl version in the example doesn't need to be updated

frezbo (approved):

website/content/v1.2/advanced/talos-api-access-from-k8s.md
utkuozdemir: Doesn't `<talos version>` get updated automatically? I found it from the other places in docs.

internal/app/machined/pkg/controllers/kubeaccess/crd.go
utkuozdemir: Every time an event is received, we stop&start the controller so it is created with the new config (allowed roles & namespaces). Same idea as we do for kubelet: https://github.com/siderolabs/talos/blob/master/internal/app/machined/pkg/controllers/k8s/kubelet_service.go#L186-L190

An alternative would be to never stop the controller but pass the new variables in a concurrency-safe manner, but it didn't occur to me while writing it :)

internal/integration/api/serviceaccount.go
utkuozdemir: Good idea, I just added it to the wait block so it will fail the test if the feature is enabled but the service `talos.default` does not appear.

website/content/v1.2/advanced/talos-api-access-from-k8s.md
frezbo: cool, i thought it needed a syntax like `{{` maybe I'm wrong :sweat_smile:

frezbo (approved):

cmd/talosctl/cmd/mgmt/kube_inject.go
smira: my reason is the following: `talosctl kube-inject` is ambigious, as it's not clear what it injects, and sooner or later we might want another way of `kube-inject` (to inject something else into kubernetes pods). `talosctl talosconfig` is somewhat specific to the `talosconfig` (client config), so it falls under proper domain.

any other opinions?

utkuozdemir: Yes, this makes sense. Another idea would be to switch it to `inject` and make `talosconfig` a subcommand or so (or rather, `serviceaccount`, since it is the type that is being injected. Sth like `talosctl inject serviceaccount -f ...`). 

This way, later, we'd be able to introduce `talosctl inject somethingelse`...

pkg/machinery/client/config/path.go
```diff
@@ -5,43 +5,92 @@
...
 
 	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
+// Path represents a path to a configuration file.
+type Path struct {
+	// Path is the filesystem path of the config.
+	Path string
+	// AllowWrite is true if the path is allowed to be written.
+	AllowWrite bool
```
smira, line 20: nit: looks better?

```suggestion
	Writeable bool
```

internal/integration/api/serviceaccount.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	secretGVR = schema.GroupVersionResource{
+		Group:    "",
+		Version:  "v1",
+		Resource: "secrets",
+	}
+)
+
+// ServiceAccountSuite verifies Talos ServiceAccount.
+type ServiceAccountSuite struct {
```
smira, line 46: :heart:

internal/integration/api/serviceaccount.go
```diff
@@ -0,0 +1,293 @@
...
+	"github.com/talos-systems/talos/pkg/machinery/client/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+)
+
+var (
+	serviceAccountGVR = schema.GroupVersionResource{
+		Group:    "talos.dev",
+		Version:  "v1alpha1",
+		Resource: "serviceaccounts",
```
smira, line 36: should we use constants here?

internal/app/machined/pkg/controllers/kubeaccess/talossa/crd_controller.go
```diff
@@ -0,0 +1,633 @@
...
+
+	for i := 0; i < workers; i++ {
+		go wait.Until(func() {
+			t.runWorker(ctx)
+		}, time.Second, ctx.Done())
+	}
+
+	t.logger.Debug("started workers")
+	<-ctx.Done()
+	t.logger.Debug("shutting down workers")
```
smira, line 228: it doesn't look like we wait here for the workers to stop? (or does it happen implicitly via something else?)

smira (commented):
:cool: :1st_place_medal:

pkg/machinery/client/config/path.go
utkuozdemir: `Writeable` sounded to me like a filesystem permissions thing, while actually it is our policy and has nothing to do with FS. That's why I chose the name. But no strong opinions, can change.

internal/integration/api/serviceaccount.go
utkuozdemir: I prefer to not use variables in tests so that they would fail explicitly if we change a constant and I think that's ok&safer. String literals > vars in tests. Wdyt?

smira: I feel it might be some sort of confusion eventually. This way tests would fail if we change the constants, but probably tracking it down won't be trivial, and it might seem that we don't have constants for those value.

I would suggest to keep using constants, as if we change them, we know we changed them :)

internal/app/machined/pkg/controllers/kubeaccess/talossa/crd_controller.go
utkuozdemir: Yep, controller is stopped when the context passed to `.Run` is cancelled. Context cancellation causes deferred `t.queue.ShutDown()` on line 203 to run, which causes the loop `for t.processNextWorkItem(ctx)` on line 234 to terminate, which terminates the worker. I also did test locally to make sure that the workers actually are stopped.

internal/integration/api/serviceaccount.go
utkuozdemir: done

pkg/machinery/client/config/path.go
utkuozdemir: `WriteAllowed`, sounds good, updated it.

cmd/talosctl/cmd/mgmt/kube_inject.go
smira: yep, `talosctl inject serviceaccount` sounds good to me!

internal/app/machined/pkg/controllers/kubeaccess/talossa/crd_controller.go
smira: so my concern would be that we don't wait for the workers to  stop... so if they don't actually stop, we would never notice that and it might leave some extra/stale workers running.

smira:
Seems like the only thing left is renaming the command, and it should be gtg?

utkuozdemir:
> Seems like the only thing left is renaming the command, and it should be gtg?

Yes, will do it today. Also will have a look to the workers stopping.

cmd/talosctl/cmd/mgmt/kube_inject.go
utkuozdemir: done the change, feel free to have another look!

internal/app/machined/pkg/controllers/kubeaccess/talossa/crd_controller.go
utkuozdemir: I added a wait using a waitgroup and locally verified that workers actually terminate and we wait for them. Feel free to have another look.

smira (approved):
:cool: :1st_place_medal:

Issue #5668, opened 2022-05-30
Allow configuring etcd to listen only on specified IPs/interfaces


TrueGoric:
### Description
While specifying `cluster.etcd.subnet` allows for filtering requests, etcd still exposes client and server ports leaking information and potentially allowing 0-days to be exploited. There should be a configuration entry for explicitly setting up IPs/interfaces.

```
$ nmap -p1-65535 <domain-node-endpoint>
Starting Nmap 7.92 ( https://nmap.org ) at 2022-05-30 13:25 Central European Daylight Time
Nmap scan report for <domain-node-endpoint>(<node-endpoint-ipv4>)
Host is up (0.031s latency).
rDNS record for <node-endpoint-ipv4>: <node-endpoint-ipv4-rdns>
Not shown: 65529 closed tcp ports (reset)
PORT      STATE    SERVICE
25/tcp    filtered smtp
2379/tcp  open     etcd-client
2380/tcp  open     etcd-server
10250/tcp open     unknown
50000/tcp open     ibm-db2
50001/tcp open     unknown

Nmap done: 1 IP address (1 host up) scanned in 80.78 seconds
```

PR #6012, opened 2022-08-02
fix: use controllers/resources for etcd configuration

### LINKED ISSUE #5889: etcd join and certificate fixes

smira:
This extracts etcd configuration and finalized run arguments as
resources managed by controllers.

The biggest change in terms of UX is that Talos now waits for the etcd
configured subnet to be actually available before starting etcd.
Previously etcd quickly failed if the requested subnet was not available
on the host.

Coupled with other fixes (#5951, #5988), this should bring etcd
join/promote sequence back into proper shape.

I also reverted all temporary measures for discovering etcd endpoints,
now etcd join doesn't depend on Kubernetes (once again).

Fixes #5889


internal/app/machined/pkg/controllers/etcd/config_test.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+	}
+
+	machineConfig := config.NewMachineConfig(cfg)
+	suite.Require().NoError(suite.State().Create(suite.Ctx(), machineConfig))
+
+	suite.AssertWithin(3*time.Second, 100*time.Millisecond, ctest.WrapRetry(func(assert *assert.Assertions, require *require.Assertions) {
+		etcdConfig, err := safe.StateGet[*etcd.Config](suite.Ctx(), suite.State(), etcd.NewConfig(etcd.NamespaceName, etcd.ConfigID).Metadata())
+		if err != nil {
+			assert.NoError(err)
```
frezbo, line 61: :thinking: I didn't quite get this check here

internal/app/machined/pkg/controllers/etcd/spec.go
```diff
@@ -0,0 +1,149 @@
...
+		ips, err := net.IPAddrs()
+		if err != nil {
+			return fmt.Errorf("error listing IPs: %w", err)
+		}
+
+		listenAddress := netip.IPv4Unspecified()
+
+		for _, ip := range ips {
+			if ip.To4() == nil {
+				listenAddress = netip.IPv6Unspecified()
```
frezbo, line 117: shouldn't we be able to listen on both ipv4 and ipv6?

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -386,19 +391,14 @@ func (e *Etcd) argsForInit(ctx context.Context, r runtime.Runtime) error {
 		_, upgraded = meta.LegacyADV.ReadTag(adv.Upgrade)
 	}
 
-	primaryAddr, listenAddress, err := primaryAndListenAddresses(r.Config().Cluster().Etcd().Subnet())
-	if err != nil {
-		return fmt.Errorf("failed to calculate etcd addresses: %w", err)
-	}
-
 	// TODO(scm): see issue #2121 and description below in argsForControlPlane.
```
frezbo, line 394: seems this issue is already closed

frezbo (approved):

internal/app/machined/pkg/controllers/etcd/spec.go
smira: I believe listening on IPv6 implies listening on IPv4.

This issue is that if IPv6 is disabled, we can't listen on IPv6 `::`

internal/app/machined/pkg/controllers/etcd/config_test.go
smira: the idea is that `assert.` continues retrying, but `require.` stops retrying

so if we failed to get a resources, it's a retryable failure, but we also want not to proceed further as `etcdConfig` is `nil`

internal/app/machined/pkg/system/services/etcd.go
smira: ok, this one was actually closed by mistake... there's a new one #5668 which I pulled into the v1.2, as it's relatively easy to implement now. anyways, this comment is outdated

internal/app/machined/pkg/controllers/etcd/spec.go
```diff
@@ -0,0 +1,149 @@
...
+
+				break
+			}
+		}
+
+		// we use stdnet.IP here to re-use already existing functions in talos-systems/net
+		// once talos-systems/net is migrated to netaddr or netip, we can use it here
```
Unix4ever, line 124: Would be great to have a ticket for migration and link it as `TODO` here :)
but not a great deal.

Unix4ever (approved):

internal/app/machined/pkg/controllers/etcd/spec.go
smira: yep, I'm thinking about it... it will simplify testing a lot as well

pkg/machinery/resources/etcd/spec.go
```diff
@@ -0,0 +1,63 @@
...
+		Aliases:          []resource.Type{},
+		DefaultNamespace: NamespaceName,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "Name",
+				JSONPath: "{.name}",
+			},
+			{
+				Name:     "AdvertisedAddress",
+				JSONPath: "{.advertistedAddress}",
```
sauterp, line 59:
```suggestion
				JSONPath: "{.advertisedAddress}",
```

pkg/machinery/resources/etcd/spec.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Spec resource holds status of rendered secrets.
+type Spec = typed.Resource[SpecSpec, SpecRD]
+
+// SpecSpec describes (some) Specuration settings of etcd.
+//
+//gotagsrewrite:gen
+type SpecSpec struct {
+	Name              string            `yaml:"name" protobuf:"1"`
+	AdvertisedAddress netip.Addr        `yaml:"advertistedAddress" protobuf:"2"`
```
sauterp, line 29:
```suggestion
	AdvertisedAddress netip.Addr        `yaml:"advertisedAddress" protobuf:"2"`
```

sauterp (changes requested):

sauterp (approved):

PR #6020, opened 2022-08-04
refactor: use controllers/resources to feed trustd with data

### LINKED ISSUE #5863: rework trustd server certificate generation process
## Feature Request

`trustd` certificate generation should be reworked to match exactly how `apid` handles that: 

- build a resource in the `machined`  which contains automatically renewed and update cert 
- use that certificate in `trustd` (and reload it automatically) via COSI connection to the `machined`
- don't filter out SideroLink addresses

smira:
This is mostly same as the way `apid` consumes certificates generated by
`machined` via COSI API connection.

Service `trustd` consumes two resources:

* `secrets.Trustd` which contains `trustd` server TLS certificates and
  it gets refreshed as e.g. node IP changes
* `secrets.OSRoot` which contains Talos API CA and join token

This PR fixes an issue with `trustd` certs not always including all IPs
of the node, as previously `trustd` certs will only capture addresses of
the node at the moment of `trustd` startup.

Another thing is that refactoring allows to dynamically change API CA
and join token. This needs more work, but `trustd` should now pick up
changes without any additional changes.

Fixes #5863


internal/app/machined/pkg/controllers/secrets/trustd.go
```diff
@@ -0,0 +1,273 @@
...
+	return nil
+}
+
+func (ctrl *TrustdController) teardownAll(ctx context.Context, r controller.Runtime) error {
+	list, err := r.List(ctx, resource.NewMetadata(secrets.NamespaceName, secrets.TrustdType, "", resource.VersionUndefined))
+	if err != nil {
+		return err
+	}
+
+	// TODO: change this to proper teardown sequence
```
frezbo, line 264: this teardown seems fine to me :sweat_smile: , anything else extra needed here?

internal/app/trustd/internal/reg/reg.go
```diff
@@ -31,15 +34,20 @@ func (r *Registrator) Register(s *grpc.Server) {
 
 // Certificate implements the securityapi.SecurityServer interface.
 func (r *Registrator) Certificate(ctx context.Context, in *securityapi.CertificateRequest) (resp *securityapi.CertificateResponse, err error) {
+	osRoot, err := safe.StateGet[*secrets.OSRoot](ctx, r.Resources, resource.NewMetadata(secrets.NamespaceName, secrets.OSRootType, secrets.OSRootID, resource.VersionUndefined))
+	if err != nil {
+		return nil, err
+	}
+
 	// TODO: Verify that the request is coming from the IP addresss declared in
```
frezbo, line 42: a ticket for this maybe

internal/app/machined/pkg/controllers/secrets/trustd.go
smira: yep, probably I should remove it, it's an old comment I got via copy-paste

internal/app/trustd/internal/reg/reg.go
smira: this was this same way for a looong time, so proper answer I guess is to change the join flow to be based on some form of node attestation, but nothing changed in this PR

frezbo (approved):

pkg/machinery/resources/secrets/trustd.go
```diff
@@ -0,0 +1,94 @@
...
+
+// TrustdCertsSpec describes etcd certs secrets.
+//
+//gotagsrewrite:gen
+type TrustdCertsSpec struct {
+	CA     *x509.PEMEncodedCertificateAndKey `yaml:"ca" protobuf:"1"` // only cert is passed, without key
+	Server *x509.PEMEncodedCertificateAndKey `yaml:"server" protobuf:"2"`
+}
+
+// NewTrustd initializes a Etc resource.
```
Unix4ever, line 35:
```suggestion
// NewTrustd initializes a Trustd resource.
```

Unix4ever (approved):

Issue #5847, opened 2022-07-01
allow Kubernetes node to be not registered


smira:
## Feature Request

Talos has some boot sequence actions which depend on the node being registered (e.g. labeling the node, uncordoning, etc.): we should provide a config option to disable node registration (kubelet `--register-node=false`) and skip those boot sequence actions.

smira:
With `--register-node=false` `kubelet` eventually fails as the node is not registered.

Q: can we make it run forever in 'not registered' state?

frezbo:
Kubelet has the option to run without registering to api server and just allow it to only support static pods:

```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
staticPodPath: "/etc/kubernetes/manifests"
authentication:
  anonymous:
    enabled: false
  webhook:
    enabled: false
authorization:
  mode: AlwaysAllow
clusterDNS:
  - 10.89.0.1
```
An example config is shown below, if the cni plugins are not installed kubelet logs would complain about missing cni's but static pods with `hostNetwork: true` would still work.

Kubelet can then be started as normal.

```bash
/usr/local/bin/kubelet --config=/opt/kubelet/kubelet-config.yaml --container-runtime=remote --container-runtime-endpoint=unix:///run/containerd/containerd.sock
```

PR #6021, opened 2022-08-04
feat: support skipping node registration

frezbo:
This PR supports skipping node registration from K8s.

This is an adavnced use case and only needs to be used in special cases.
In this mode Kubelet only runs the static pods.

Fixes: #5847

Operations that will be broken:

- `talosctl cluster create` would eventually timeout since it expects
  nodes to be registered.
- `talosctl health` since it expects nodes to be registered.
- `talosctl upgrade-k8s` since it expects nodes to be registered. Static
  pods can still be updated by editing the machine config..


internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -163,7 +166,7 @@ func (ctrl *KubeletSpecController) Run(ctx context.Context, r controller.Runtime
 			return fmt.Errorf("error merging arguments: %w", err)
 		}
 
-		kubeletConfig, err := NewKubeletConfiguration(cfgSpec.ClusterDNS, cfgSpec.ClusterDomain, cfgSpec.ExtraConfig, cfgSpec.DefaultRuntimeSeccompEnabled)
+		kubeletConfig, err := NewKubeletConfiguration(cfgSpec)
```
frezbo, line 169: do i need to pass as?
```suggestion
		kubeletConfig, err := NewKubeletConfiguration(cfgSpec.DeepCopy())
```

frezbo: the map gets modified here: https://github.com/siderolabs/talos/pull/6021/files#diff-c1756580a1897c92944f52223cd04cfd6279aabfafc71c7bfcd30a0904e6b807R323

smira (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
smira: looks like it's fine? as we don't use `cfgSpec` further, and what we get back as a resource is always a copy

Issue #6011, opened 2022-08-02
VLANs Do Not Have `dhcpOptions` Exposed


dcplaya:
## Bug Report

I am unable to set any `dhcpOptions` for VLANs. My uses case specifically is `routeMetric`.
In reference to [this](https://taloscommunity.slack.com/archives/CMARMBC4E/p1659453543705179) Slack thread

### Environment

- Talos version: `v1.1.1` & `v1.1.2`
- Kubernetes version: ` v1.24.3`
- Platform: All, but my platforms specifically are Proxmox VMs and Intel NUC11s. Both are managed by my Sidero cluster.

PR #6022, opened 2022-08-04
feat: support dhcp options for vlan

frezbo:
Add `DHCPOptions` for VLAN device.

Fixes: #6011


smira (approved):

smira:
I would argue it's a `feat:`, as it updates machine config

frezbo:
> I would argue it's a `feat:`, as it updates machine config

it felt trivial, I'll update the commit message

PR #6043, opened 2022-08-08
auth proxy WIP

sauterp:
# Pull Request

## Acceptance


cmd/talosctl/cmd/talos/root.go
```diff
@@ -60,6 +62,18 @@ func WithClientNoNodes(action func(context.Context, *client.Client) error) error
 				opts = append(opts, client.WithContextName(Cmdcontext))
 			}
 
+			ctx = metadata.AppendToOutgoingContext(ctx, "runtime", "Talos")
```
smira, line 65: we should probably figure out a way to generalize this... probably need a call with @Unix4ever

cmd/talosctl/cmd/talos/root.go
```diff
@@ -60,6 +62,18 @@ func WithClientNoNodes(action func(context.Context, *client.Client) error) error
...
+
+			// TODO should this be configurable?
```
smira, line 67: this should be done in the client library based on `talosconfig`

pkg/machinery/client/client.go
```diff
@@ -182,7 +183,18 @@ func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.Cl
 		// Such possibilities include SRV records, multiple IPs from A and/or AAAA
 		// records, and descriptive TXT records which include things like load
 		// balancer specs.
-		target = fmt.Sprintf("dns:///%s:%d", net.FormatAddress(endpoints[0]), constants.ApidPort)
+		endpoint := endpoints[0]
```
smira, line 186: this is the wrong place actually. 

we need to handle it same way for the case when `len(endpoints) > 1` and `len(endpoints) == 1`

pkg/machinery/client/client.go
```diff
@@ -182,7 +183,18 @@ func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.Cl
...
+
+		// TODO make port 443 the default for https://
+		ind := strings.Index(endpoint, "://")
```
smira, line 189: we should try to use `url.Parse` here imho and handle the rest.

pkg/machinery/client/client.go
```diff
@@ -221,7 +233,63 @@ func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.Cl
 
 	dialOpts = append(dialOpts, opts...)
 
-	return grpc.DialContext(ctx, target, dialOpts...)
+	creds := BasicAuth{
+		auth: c.options.configContext.Auth.Basic.Username + ":" + c.options.configContext.Auth.Basic.Password,
```
smira, line 237: why here? unconditional?

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -62,9 +63,21 @@ func (r *roundRobinResolver) start() error {
 	var addrs []resolver.Address //nolint:prealloc
 
 	for _, a := range strings.Split(r.target.Endpoint, ",") { //nolint:staticcheck
+		addr := net.FormatAddress(a)
```
smira, line 66: we should stop using this one and use `stdnet.JoinHostPort` instead of this and `fmt.Sprintf(%s:%d)`

pkg/machinery/client/client.go
sauterp: I'm trying to go step by step, compiling and trying. once I can successfully connect, authenticate and get a response I want to make it conditional etc.

smira:
@sauterp can this be closed now?

PR #6053, opened 2022-08-09
feat: support custom ports for connecting to apid from talosctl

sauterp:
# Pull Request

Users can now add a port suffix to the endpoints used by talosctl. Either
in the CLI flag or the ~/.talos/config. The default port is still 50000.


pkg/machinery/client/client.go
```diff
@@ -171,6 +173,14 @@ func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.Cl
 
 	var target string
 
+	for i, endpoint := range endpoints {
+		if net.AddressContainsPort(endpoint) {
+			endpoints[i] = net.FormatAddress(endpoint)
```
smira, line 178: please let's try to avoid using this package, as it's confusing, and current behavior is wrong 

we should instead do the following: https://pkg.go.dev/net#SplitHostPort

pkg/machinery/client/client.go
```diff
@@ -171,6 +173,14 @@ func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.Cl
...
+		} else {
+			endpoints[i] = gonet.JoinHostPort(net.FormatAddress(endpoint), strconv.Itoa(constants.ApidPort))
```
smira, line 180: this is wrong for IPv6 - `net.FormatAddress` adds square brackets, and another square brackets in `JoinHostPort`, so just leave `net.JoinHostPort`

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -62,9 +64,21 @@ func (r *roundRobinResolver) start() error {
 	var addrs []resolver.Address //nolint:prealloc
 
 	for _, a := range strings.Split(r.target.Endpoint, ",") { //nolint:staticcheck
+		addr := net.FormatAddress(a)
```
smira, line 67: not totally clear why can't we prepare `endpoint:port` before we pass down to this function, so that we don't have to append ports here?

smira (commented):
parsing/adding ports should be extracted to a function and covered with unit-tests including IPv4, IPv6 and hostnames

pkg/machinery/client/client.go
```diff
@@ -302,7 +314,7 @@ func NewClient(cfg *tls.Config, endpoints []string, port int, opts ...grpc.DialO
 	c = &Client{}
 
 	// TODO(smira): endpoints[0] should be replaced with proper load-balancing
-	c.conn, err = grpc.DialContext(context.Background(), fmt.Sprintf("%s:%d", net.FormatAddress(endpoints[0]), port), opts...)
+	c.conn, err = grpc.DialContext(context.Background(), fmt.Sprintf("%s:%d", endpoints[0], port), opts...)
```
smira, line 317: this became wrong now: it should use `net.JoinHostPort`, but better if we drop the method altogether. It is deprecated for a long time now.

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -61,10 +62,19 @@ type roundRobinResolver struct {
 func (r *roundRobinResolver) start() error {
 	var addrs []resolver.Address //nolint:prealloc
 
-	for _, a := range strings.Split(r.target.Endpoint, ",") { //nolint:staticcheck
+	for _, addr := range strings.Split(r.target.Endpoint, ",") { //nolint:staticcheck
```
smira, line 65: how can we get to 'no port' scenario here?

sauterp: you're right, that wouldn't happen now.

pkg/machinery/client/client.go
```diff
@@ -302,7 +307,7 @@ func NewClient(cfg *tls.Config, endpoints []string, port int, opts ...grpc.DialO
 	c = &Client{}
```
smira, line 307: can we please drop `NewClient` method altogether?

pkg/machinery/client/client.go
```diff
@@ -167,7 +169,10 @@ func New(ctx context.Context, opts ...OptionFunc) (c *Client, err error) {
 
 // GetConn creates new gRPC connection.
 func (c *Client) GetConn(ctx context.Context, opts ...grpc.DialOption) (*grpc.ClientConn, error) {
```
smira, line 171: let's make this method private, iirc it's not being used outside of this module, and it doesn't make sense to keep it public

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -58,13 +60,30 @@ type roundRobinResolver struct {
 	port   int
 }
 
+func EnsureEndpointsHavePorts(endpoints []string, defaultPort int) ([]string, error) {
+	for i, endpoint := range endpoints {
```
smira, line 64: this can be written with `slices.Map` function in a bit easier way

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -58,13 +60,30 @@ type roundRobinResolver struct {
...
+		if !sideronet.AddressContainsPort(endpoint) {
```
smira, line 65: like I put in my previous comment, let's use `net.SplitHostPort` to attempt to split a port. if it succeeds, we don't need to do anything, otherwise we do `JoinHostPort`

pkg/machinery/client/client.go
sauterp: While we're at it, it looks to me like NewClientContextAndCredentialsFromConfig and NewClientContextAndCredentialsFromParsedConfig are also depreceated and unused. Should we delete those aswell or do other repositories/users depend on it?

sauterp: a quick search of github says they aren't used anywhere

pkg/machinery/client/resolver/roundrobin.go
sauterp: I saw your comment and tried it, but `net.SplitHostPort` returns an error if there is no port and this function can also return other errors, so the only way to distinguish them would be a weird string comparison.

smira: no big deal, if it returns an error, there's no port, and that's all we need to know. we don't need an error here

sauterp: done

pkg/machinery/client/client.go
sauterp: done

pkg/grpc/gen/remote.go
```diff
@@ -20,10 +20,13 @@ import (
...
 
-var trustdResolverScheme string
+var (
+	trustdResolverScheme string
+	resolverDefaultPort  int = constants.TrustdPort
+)
 
 func init() {
-	trustdResolverScheme = resolver.RegisterRoundRobinResolver(constants.TrustdPort)
+	trustdResolverScheme = resolver.RegisterRoundRobinResolver(resolverDefaultPort)
```
smira, line 29: why do we need to pass port the resolver? if we append the port outside the resolver??

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -58,13 +64,35 @@ type roundRobinResolver struct {
...
+		host, _, err := net.SplitHostPort(serverName)
+		if err == nil {
+			serverName = host
+		}
+
 		addrs = append(addrs, resolver.Address{
-			ServerName: a,
-			Addr:       fmt.Sprintf("%s:%d", net.FormatAddress(a), r.port),
+			ServerName: serverName,
+			Addr:       sideronet.FormatAddress(addr),
```
smira, line 95: please, please, please, don't use this method. it's not needed here, and it's going to break any correctly formatted endpoint.

smira (changes requested):

frezbo (approved):

pkg/machinery/client/resolver/roundrobin.go
```diff
@@ -5,30 +5,28 @@
...
+	"github.com/talos-systems/talos/pkg/machinery/generic/slices"
 )
 
-// RegisterRoundRobinResolver registers round-robin gRPC resolver for specified port and returns scheme to use in grpc.Dial.
-func RegisterRoundRobinResolver(port int) (scheme string) {
-	scheme = fmt.Sprintf(roundRobinResolverScheme, port)
+// RoundRobinResolverScheme is a scheme to use in grpc.Dial for the round-robin gRPC resolver.
+// This resolver requires that all endpoints have a port appended.
+// To ensure this, use EnsureEndpointsHavePorts before constructing a connection string.
+const RoundRobinResolverScheme = "roundrobin"
```
smira, line 21: I believe etcd or something else might use a similar name, probably it might be safer to use `talosroundrobin` here.

as gRPC resolver schemes share a common namespace, we should play nice and choose prefixed names

smira (approved):

Issue #5499, opened 2022-05-04
tracking progress of "action" API calls


smira:
## Feature Request

In the `talosctl` CLI track progress of the operations:

* reboot
* reset
* shutdown
* upgrade

.. make the UX better by providing the progress of the operation (10% done, failed, etc.), tracking the progress over the reboot as well (if possible) by waiting for the API to be available after an upgrade.

Tracking progress is the default behavior.

Additional flags:

* `--no-wait` to restore current fire-and-forget behaviour
* `--debug` to stream `dmesg` from the involved nodes for debugging issues

(NB: this should work with multiple `--nodes`).

### Description

Idea: for upgrade, apply-config (in reboot mode) and reboot we should wait for the node to reboot and come back.

smira:
- we can use Events API to track the progress
- `apid` should be kept running as long as possible, possibly it should have a chance to flush the data before stopping
- we can add `actor ID` to the sequence runner and that gets embedded into the event
- these API calls should return `actor ID` to the client so that client can filter out events on actor ID

PR #6061, opened 2022-08-10
feat: add actor ID to events & emit an initial empty event

utkuozdemir:
Add a new field `actorID` to the events and populate it with a UUID for the lifecycle actions `reboot`, `reset`, `upgrade` and `shutdown`. This actor ID will be present on all events emitted by this triggered action. We can use this ID later on the client side to be able to track triggered actions.

Also introduce a new event type, `EmptyEvent`, and emit it on the events streaming GRPC endpoint when a client connects. The purpose of this event is to signal to the client that the event streaming has actually started.

Server-side part of siderolabs/talos#5499.


api/machine/machine.proto
```diff
@@ -248,12 +249,14 @@ message EventsRequest {
   int32 tail_events = 1;
   string tail_id = 2;
   int32 tail_seconds = 3;
+  string actor_id = 4;
```
smira, line 252:
```suggestion
  string with_actor_id = 4; 
```

smira (commented):
do we need to change the Talos API client as well? to wait for the first "null" event? to support filtering on actor ID?

utkuozdemir:
> do we need to change the Talos API client as well? to wait for the first "null" event? to support filtering on actor ID?

I thought I would make those changes part of the second (client-side) PR as `EventsWatchV2` and do not change the behavior of the existing `EventsWatch` function.

Or do we want to freeze the client API as well when before we cut the first beta?

api/machine/machine.proto
utkuozdemir: done

frezbo (approved):

smira:
> Or do we want to freeze the client API as well when before we cut the first beta?

Usually we do, but probably it will be still fine as a special case here.

smira (approved):

utkuozdemir:
> > Or do we want to freeze the client API as well when before we cut the first beta?
> 
> Usually we do, but probably it will be still fine as a special case here.

👍 - it won't be a breaking change anyway, just a new function.

PR #6062, opened 2022-08-10
feat: support patching the machine config in the `apply-config` cmd

Unix4ever:
Fixes: https://github.com/siderolabs/talos/issues/6045

`talosctl apply-config` now supports `--config-patch` flag that takes
machine config patches as the input.

### DIFF
--- a/internal/integration/cli/apply-config.go
+++ b/internal/integration/cli/apply-config.go
@@ -1,0 +1,54 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_cli
+// +build integration_cli
+
+package cli
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/talos-systems/talos/internal/integration/base"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+)
+
+// ApplyConfigSuite verifies dmesg command.
+type ApplyConfigSuite struct {
+	base.CLISuite
+}
+
+// SuiteName ...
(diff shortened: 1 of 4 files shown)


frezbo (approved):
:dancers:

internal/integration/cli/apply-config.go
```diff
@@ -0,0 +1,51 @@
...
+
+	suite.Require().NoError(os.WriteFile(patchPath, []byte(patch), 777))
+
+	data, _ := suite.RunCLI([]string{"read", "--nodes", suite.RandomDiscoveredNodeInternalIP(), "/system/state/config.yaml"})
+
+	configPath := filepath.Join(suite.T().TempDir(), "config.yaml")
+
+	suite.Require().NoError(os.WriteFile(configPath, []byte(data), 777))
+
+	suite.RunCLI([]string{"apply-config", "--nodes", suite.RandomDiscoveredNodeInternalIP(), "--config-patch", fmt.Sprintf("@%s", patchPath), "-f", configPath})
```
smira, line 46: we should probably patch the same node we read config from, not _another_ random node? :)

Unix4ever: lol, oops. WIll fix :D

internal/integration/cli/apply-config.go
```diff
@@ -0,0 +1,53 @@
...
+
+// TestApplyWithPatch verifies that .
+func (suite *ApplyConfigSuite) TestApplyWithPatch() {
+	patch := `---
+cluster:
+  apiServer:
+    extraArgs:
+      logging-format: text`
+
+	node := suite.RandomDiscoveredNodeInternalIP()
```
smira, line 36: I think it should be machine.TypeControlPlane, otherwise the patch doesn't make much sense?

smira (approved):

internal/integration/cli/apply-config.go
Unix4ever: yeah... makes sense. Fixed

PR #6069, opened 2022-08-11
feat: enable talos client to connect to Talos through an auth proxy

### LINKED ISSUE #5980: Talos API access through an authentication proxy

sauterp:
# Pull Request
Talos client can connect to Talos API via a proxy with basic auth.
Additionally it is now optional to specify a TLS CA,key or crt. Optionally
Developers can build talosctl with WITH_DEBUG=1 to allow insecure
connections when http:// endpoints are specified.

Fixes #5980

### DIFF
--- a/cmd/talosctl/cmd/talos/root.go
+++ b/cmd/talosctl/cmd/talos/root.go
@@ -37,13 +37,14 @@ var (
 	Endpoints   []string
 	Nodes       []string
 	Cmdcontext  string
+	Cluster     string
 )
 
 const pathAutoCompleteLimit = 500
 
 // WithClientNoNodes wraps common code to initialize Talos client and provide cancellable context.
 //
-// WithClientNoNodes doesn't set any node information on request context.
+// WithClientNoNodes doesn't set any node information on the request context.
 func WithClientNoNodes(action func(context.Context, *client.Client) error) error {
 	return cli.WithContext(
 		context.Background(), func(ctx context.Context) error {
@@ -65,6 +66,10 @@ func WithClientNoNodes(action func(context.Context, *client.Client) error) error
 				opts = append(opts, client.WithEndpoints(Endpoints...))
 			}
 
+			if Cluster != "" {
+				opts = append(opts, client.WithCluster(Cluster))
+			}
+
 			c, err := client.New(ctx, opts...)
 			if err != nil {
 				return fmt.Errorf("error constructing client: %w", err)
@@ -77,6 +82,8 @@ func WithClientNoNodes(action func(context.Context, *client.Client) error) error
 	)
 }
 
+var errConfigContext = fmt.Errorf("failed to resolve config context")
+
 // WithClient builds upon WithClientNoNodes to provide set of nodes on request context based on config & flags.
 func WithClient(action func(context.Context, *client.Client) error) error {
 	return WithClientNoNodes(
@@ -84,7 +91,7 @@ func WithClient(action func(context.Context, *client.Client) error) error {
 			if len(Nodes) < 1 {
 				configContext := c.GetConfigContext()
 				if configContext == nil {
-					return fmt.Errorf("failed to resolve config context")
+					return errConfigContext
 				}
 
 				Nodes = configContext.Nodes
--- a/pkg/machinery/client/client.go
+++ b/pkg/machinery/client/client.go
@@ -10,21 +10,16 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
-	"crypto/tls"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
-	"strings"
 	"time"
 
 	cosiv1alpha1 "github.com/cosi-project/runtime/api/v1alpha1"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/cosi-project/runtime/pkg/state/protobuf/client"
-	grpctls "github.com/talos-systems/crypto/tls"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/emptypb"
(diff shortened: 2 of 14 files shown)


pkg/machinery/client/connection.go
```diff
@@ -0,0 +1,183 @@
...
+	tlsConfig = &tls.Config{}
+
+	if err := c.resolveConfigContext(); err != nil {
+		return nil, fmt.Errorf("failed to resolve configuration context: %w", err)
+	}
+
+	basicAuth := c.options.configContext.Auth.Basic
+	if basicAuth != nil {
+		dialOpts = append(dialOpts, WithGRPCBasicAuth(basicAuth.Username, basicAuth.Password))
+		if shouldInsecureConnectionsBeAllowed(endpoints) {
```
smira, line 63: why are we taking a shortcut here? 

we set the basic auth, and that should be it.

whether to use insecure or TLS-based connection should be decided below

pkg/machinery/client/connection.go
```diff
@@ -0,0 +1,183 @@
...
+			return c.makeConnection(ctx, target, insecure.NewCredentials(), dialOpts)
+		}
+	}
+
+	caBytes, err := getCA(c.options.configContext)
```
smira, line 68: lines 68-91 here seems to be building `credentials.TransportCredentials` and we can extract this to a separate (testable) function. 

in this function we can also return `Insecure` creds if it's a debug build and insecure creds were requested

pkg/machinery/client/grpc_connection_wrapper.go
```diff
@@ -0,0 +1,50 @@
...
+package client
+
+import (
+	"context"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+type GRPCConnectionWrapper struct {
```
smira, line 14: we can embed the connection into a struct

```go
type grpcConnectionWrapper struct {
   *grpc.ClientConn
   
    clusterName string
}
```

this way all methods of `ClientConn` will be exposed via the wrapper by default, but we can override the methods we need (`Invoke` & `NewStream`). This way we can drop `Close`

also this struct shouldn't be public, as we don't need it in the  external API

pkg/machinery/client/grpc_connection_wrapper.go
```diff
@@ -0,0 +1,50 @@
...
+	clusterName string
+
+	conn *grpc.ClientConn
+}
+
+func NewGRPCConnectionWrapper(clusterName string, conn *grpc.ClientConn) *GRPCConnectionWrapper {
```
smira, line 20:
```suggestion
func newGRPCConnectionWrapper(clusterName string, conn *grpc.ClientConn) *GRPCConnectionWrapper {
```

pkg/machinery/client/grpc_connection_wrapper.go
```diff
@@ -0,0 +1,50 @@
...
+	ctx = metadata.AppendToOutgoingContext(ctx, "runtime", "Talos")
+
+	if c.clusterName != "" {
+		ctx = metadata.AppendToOutgoingContext(ctx, "context", c.clusterName)
+	}
+
+	return ctx
+}
+
+func (c *GRPCConnectionWrapper) Close() error {
```
smira, line 48: drop it (see above)

smira:
as an overall thought: we need more tests

cmd/talosctl/cmd/talos/root.go
```diff
@@ -94,6 +97,15 @@ func WithClient(action func(context.Context, *client.Client) error) error {
 				return fmt.Errorf("nodes are not set for the command: please use `--nodes` flag or configuration file to set the nodes to run the command against")
 			}
 
+			if Cluster == "" {
```
smira, line 100: this should go to `WithClientNoNodes`

cmd/talosctl/cmd/talos/root.go
```diff
@@ -94,6 +97,15 @@ func WithClient(action func(context.Context, *client.Client) error) error {
...
+				configContext := c.GetConfigContext()
+				if configContext == nil {
+					return configContextErr
+				}
+
+				Cluster = configContext.Cluster
```
smira, line 106: this `Cluster` varialbe if set should go to `client.New` as an option: `WithCluster(Cluster)`.

pkg/machinery/client/connection.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (c *Client) makeConnection(ctx context.Context, target string, creds credentials.TransportCredentials, dialOpts []grpc.DialOption) (*GRPCConnectionWrapper, error) {
+	dialOpts = append(dialOpts,
+		grpc.WithTransportCredentials(creds),
+		grpc.WithInitialWindowSize(65535*32),
+		grpc.WithInitialConnWindowSize(65535*16))
+
+	conn, err := grpc.DialContext(ctx, target, dialOpts...)
+
+	return NewGRPCConnectionWrapper(c.GetConfigContext().Cluster, conn), err
```
smira, line 104: `Cluster` should also come (as an optional override) from `Client` `Options`

pkg/machinery/client/connection.go
```diff
@@ -0,0 +1,192 @@
...
+	caBytes, err := getCA(configContext)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get CA: %w", err)
+	}
+
+	if len(caBytes) > 0 {
+		tlsConfig.RootCAs = x509.NewCertPool()
+		tlsConfig.ClientCAs = x509.NewCertPool()
+
+		if ok := tlsConfig.ClientCAs.AppendCertsFromPEM(caBytes); !ok {
```
smira, line 83: `.ClientAuth` and `.ClientCAs` are fields which are used by the server-side `tls.Config`, they shouldn't be set here: https://pkg.go.dev/crypto/tls#Config

pkg/machinery/client/client.go
```diff
@@ -138,6 +124,24 @@ func (c *Client) GetEndpoints() []string {
 	return nil
 }
 
+// GetClusterName returns the client's endpoints from the override set with WithEndpoints
```
smira, line 127: comment has to be updated

smira (approved):
:cool:

frezbo (approved):

PR #6070, opened 2022-08-11
fix: introduce 'routed' NodeAddresses and use them in kubelet

smira:
Same change will be done for the etcd in a separate PR.

The idea is to introduce a subset of `current` addresses: `routed`
addresses don't include external IPs (like AWS), as they are not on the
node, and excludes SideroLink IPs (as these are not routeable).

Reimplement `kubelet` nodeIP selection based on the new resources
removing the reliance on `net.IPAddrs`.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/node_address_test.go
+++ b/internal/app/machined/pkg/controllers/network/node_address_test.go
@@ -114,6 +114,7 @@ func (suite *NodeAddressSuite) TestDefaults() {
 					[]string{
 						network.NodeAddressDefaultID,
 						network.NodeAddressCurrentID,
+						network.NodeAddressRoutedID,
 						network.NodeAddressAccumulativeID,
 					}, func(r *network.NodeAddress) error {
 						addrs := r.TypedSpec().Addresses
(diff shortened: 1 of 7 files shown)


internal/app/machined/pkg/controllers/network/node_address_test.go
```diff
@@ -243,32 +248,47 @@ func (suite *NodeAddressSuite) TestFilters() {
 						case network.NodeAddressCurrentID:
 							if !reflect.DeepEqual(
 								addrs,
-								ipList("1.2.3.4/32 10.0.0.1/8 25.3.7.9/32 2001:470:6d:30e:4a62:b3ba:180b:b5b8/64"),
+								ipList("1.2.3.4/32 10.0.0.1/8 25.3.7.9/32 2001:470:6d:30e:4a62:b3ba:180b:b5b8/64 fdae:41e4:649b:9303:7886:731d:1ce9:4d4/128"),
```
frezbo, line 251: assert.Equal doesn't work here?

frezbo (approved):

internal/app/machined/pkg/controllers/network/node_address_test.go
smira: it's an "old" test, we can't use assert, as it's not compatible with `retry`, as we want to return an error instead of recording a test failure

sauterp (approved):

PR #6075, opened 2022-08-12
feat: track progress of action API calls

utkuozdemir:
Track the progress of the long-running actions `reboot`, `reset`, `upgrade` and `shutdown` on the client side by default, unless `--no-wait=true` is specified.

Use the events API to follow the events using the actor ID of the action and display it using an stderr reporter with a spinner.

Closes siderolabs/talos#5499.

The UX looks like this:
reboot


frezbo:
We'll need some release notes too

utkuozdemir:
> We'll need some release notes too

Yep, this is draft, still on it.

smira (commented):
I'll keep reviewing, but I need to switch gears to another issue, so pushing what I have

cmd/talosctl/cmd/talos/reboot.go
```diff
@@ -24,28 +28,62 @@ var rebootCmd = &cobra.Command{
...
+
+		postCheckFn := func(ctx context.Context, c *client.Client) error {
+			_, err := c.Disks(ctx)
+
+			return err
+		}
+
+		err := action.NewTracker(&GlobalArgs, action.MachineReadyEventFn, rebootGetActorID, postCheckFn, rebootCmdFlags.debug).Run()
+		if err != nil {
+			os.Exit(1)
```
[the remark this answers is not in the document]
smira: if we need to hide the error message, we can do that e.g. by doing some wrapping and checking for the wrapping when the error is printed in `root.go`

doing `os.Exit(1)` means that we don't close the connections properly, `defer`s don't run, etc.

cmd/talosctl/cmd/talos/shutdown.go
```diff
@@ -24,17 +28,49 @@ var shutdownCmd = &cobra.Command{
...
+					return fmt.Errorf("error executing shutdown: %s", err)
+				}
+
+				return nil
+			})
+		}
+
+		err := action.NewTracker(&GlobalArgs, action.StopAllServicesEventFn, shutdownGetActorID, nil, shutdownCmdFlags.debug).Run()
+		if err != nil {
+			os.Exit(1)
```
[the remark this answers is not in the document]
utkuozdemir: it's to avoid the error message to be printed by cobra, since it is already printed as a (colored) status by the reporter.

cmd/talosctl/cmd/talos/reboot.go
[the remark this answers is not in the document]
utkuozdemir: ah, good one, yes let me do that instead.

cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -0,0 +1,236 @@
...
+			node := node
+
+			var (
+				dmesg *circular.Buffer
+				err   error
+			)
+
+			if a.debug {
+				dmesg, err = circular.NewBuffer()
+				dmesg.GetReader()
```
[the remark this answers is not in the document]
utkuozdemir: nope, mistake

cmd/talosctl/pkg/talos/action/node.go
```diff
@@ -0,0 +1,321 @@
...
+		if eventResult.Error != nil {
+			a.update(reporter.Update{
+				Message: fmt.Sprintf("Error: %v", eventResult.Error),
+				Status:  reporter.StatusError,
+			})
+
+			return eventResult.Error
+		}
+
+		if eventResult.Event.ActorID == actorID {
```
smira, line 253: should we ignore events with `ActorID != actorID` ? as we enter this function

cmd/talosctl/pkg/talos/action/node.go
```diff
@@ -0,0 +1,321 @@
...
+			Status:  reporter.StatusRunning,
+		})
+
+	case *machineapi.ServiceStateEvent:
+		a.update(reporter.Update{
+			Message: fmt.Sprintf("ServiceStateEvent service: %v message: %v healthy: %v", msg.GetService(), msg.GetMessage(), msg.GetHealth().GetHealthy()),
+			Status:  reporter.StatusRunning,
+		})
+
+	case *machineapi.AddressEvent:
```
smira, line 307: this is more of a hacky event which we used to broadcast machine addresses, it gets sent periodically, idk, probably we should ignore it?

cmd/talosctl/pkg/talos/action/node.go
```diff
@@ -0,0 +1,321 @@
...
+		a.update(reporter.Update{
+			Message: fmt.Sprintf("AddressEvent hostname: %v addresses: %v", msg.GetHostname(), msg.GetAddresses()),
+			Status:  reporter.StatusRunning,
+		})
+
+	default:
+		a.update(reporter.Update{
```
smira, line 314: probably we should ignore unknown events? as we simply report them, new events might be of no use?

cmd/talosctl/pkg/talos/action/node.go
```diff
@@ -0,0 +1,321 @@
...
+			}
+		}
+	}
+}
+
+func (a *nodeTracker) handleEvent(event client.Event) error {
+	switch msg := event.Payload.(type) {
+	case *machineapi.PhaseEvent:
+		a.update(reporter.Update{
+			Message: fmt.Sprintf("PhaseEvent phase: %s action: %v", msg.GetPhase(), msg.GetAction()),
```
smira, line 266: nit, but we try to make our console messages lowercase, probably we could drop completely event name here and in other places?

```suggestion
			Message: fmt.Sprintf("phase: %s action: %v", msg.GetPhase(), msg.GetAction()),
```

utkuozdemir: done

cmd/talosctl/pkg/talos/action/node.go
utkuozdemir: done

cmd/talosctl/pkg/talos/action/node.go
utkuozdemir: done

smira:
There's something broken:

image

The node successfully rebooted and it was available for a long time:

```
$ talosctl -n 172.20.0.2 version
Client:
	Tag:         v1.2.0-alpha.2-44-g3dc3dfb06
	SHA:         3dc3dfb0
	Built:       
	Go version:  go1.19
	OS/Arch:     linux/amd64
Server:
	NODE:        172.20.0.2
	Tag:         v1.2.0-alpha.2-46-ge4710016f
	SHA:         e4710016
	Built:       
	Go version:  go1.19
	OS/Arch:     linux/amd64
	Enabled:     RBAC
```

smira:
one more bug: if the sequence fails, it should fail waiting

image

(easy way to reproduce - upgrade to old Talos, it will fail validating new machine config and upgrade is aborted)

smira:
> one more bug: if the sequence fails, it should fail waiting

this is how sequence failure event looks like iirc, so we should process that

pkg/machinery/client/events.go
```diff
@@ -137,6 +138,8 @@ type EventResult struct {
 	Event Event
 	// Err is the error that occurred.
 	Error error
+
+	Done bool
```
smira, line 142: seems to be unused?

smira (approved):
do we need both `.gifs`? I'm not sure why it is in the `wireguard-guide`

pkg/machinery/client/events.go
utkuozdemir: was accidental, removed

utkuozdemir:
> do we need both `.gifs`? I'm not sure why it is in the `wireguard-guide`

also was mistake, now removed.

website/content/v1.2/reference/cli.md
```diff
@@ -2191,11 +2197,13 @@ talosctl upgrade [flags]
 ### Options
 
 ```
+      --debug          debug operation from kernel logs. --no-wait is set to false when this flag is set
   -f, --force          force the upgrade (skip checks on etcd health and members, might lead to data loss)
   -h, --help           help for upgrade
   -i, --image string   the container image to use for performing the install
   -p, --preserve       preserve data
   -s, --stage          stage the upgrade to perform it after a reboot
+      --wait           wait for the operation to complete, tracking its progress. always set to true when --debug is set
```
frezbo, line 2206: the gif shows the `--no-wait` flag, and we have the `--wait` flag

utkuozdemir: oh good one, yes, video is old now. I'll re-record

utkuozdemir: just re-recorded it

PR #6076, opened 2022-08-12
chore: add proto definitions for common types

DmitriyMV:
This commit adds proto definitions for this types;
- *url.URL
- netaddr.IP
- netaddr.IPPort
- netaddr.IPPrefix
- *x509.PEMEncodedKey
- *x509.PEMEncodedCertificateAndKey


api/resource/resource.proto
```diff
@@ -92,3 +92,30 @@ message WatchResponse {
   Resource definition = 3;
   Resource resource = 4;
 }
+
+message URL {
```
smira, line 96: should we move this to e.g. `api/resource/common.proto`, as whole `resource.proto` is deprecated now

PR #6079, opened 2022-08-12
feat: add new event watch function and return lifecycle action responses on API

utkuozdemir:
Add a new function EventsWatchV2 that blocks until receiving the first event then switches to non-blocking mode.

Also add new API functions to return responses of the lifecycle actions `reboot`, `reset` and `shutdown`.

Required for the client-side part of siderolabs/talos#5499.


go.work.sum
```diff
@@ -284,54 +297,42 @@ gopkg.in/square/go-jose.v2 v2.5.1 h1:7odma5RETjNHWJnR32wx8t+Io4djHE1PqxCFx3iiZ2w
...
 k8s.io/api v0.24.2/go.mod h1:AHqbSkTm6YrQ0ObxjO3Pmp/ubFF/KuM7jU+3khoBsOg=
-k8s.io/api v0.25.0-rc.0 h1:JzQmjOfEAkOyCzDyFTayWB3nCQeS4TcCldrl2Q+417g=
-k8s.io/api v0.25.0-rc.0/go.mod h1:H8qkBzdcUEFJEhgA14ip3PpzUFN0XXPgnlDnu/KaDf0=
 k8s.io/apimachinery v0.24.2 h1:5QlH9SL2C8KMcrNJPor+LbXVTaZRReml7svPEh4OKDM=
 k8s.io/apimachinery v0.24.2/go.mod h1:82Bi4sCzVBdpYjyI4jY6aHX+YCUchUIrZrXKedjd2UM=
-k8s.io/apimachinery v0.25.0-rc.0 h1:k3n1k01fFHrTsRHM8xt+z9yOpak7KmO/pbmD0/tAkzs=
-k8s.io/apimachinery v0.25.0-rc.0/go.mod h1:qMx9eAk0sZQGsXGu86fab8tZdffHbwUfsvzqKn4mfB0=
 k8s.io/apiserver v0.24.2 h1:orxipm5elPJSkkFNlwH9ClqaKEDJJA3yR2cAAlCnyj4=
 k8s.io/apiserver v0.24.2/go.mod h1:pSuKzr3zV+L+MWqsEo0kHHYwCo77AT5qXbFXP2jbvFI=
-k8s.io/apiserver v0.25.0-rc.0 h1:PCLYM37Cq9SEbb6jFhA1E7NpfW6GddG2AzgECMU5jlw=
```
smira, line 295: should not be comitted? idk... it feels weird it now removes 0.25... this file is weird :-D

pkg/machinery/client/events.go
```diff
@@ -123,6 +123,84 @@ func (c *Client) EventsWatch(ctx context.Context, watchFunc func(<-chan Event),
...
+	// Err is the error that occurred.
+	Error error
+}
+
+// EventsWatchV2 watches events of a single node and wraps the Events by providing a simpler interface.
+// It blocks until the first (empty) event is received, then spawns a goroutine that sends events to the given channel.
+// EventResult objects sent into the channel contain either the errors or the received events.
+//
+//nolint:gocyclo
+func (c *Client) EventsWatchV2(ctx context.Context, node string, ch chan<- EventResult, opts ...EventsOptionFunc) error {
```
smira, line 139: we probably don't want a `node` here, it should be part of the context

go.work.sum
utkuozdemir: Yes I wasn't sure. It keeps changing on my local and I'm unsure if I should commit it. I just assumed that Go knows what it is doing...

pkg/machinery/client/events.go
utkuozdemir: Oh right, forgot that

pkg/machinery/client/client.go
```diff
@@ -416,14 +416,21 @@ func (c *Client) Reset(ctx context.Context, graceful, reboot bool) (err error) {
...
 // ResetGeneric implements the proto.MachineServiceClient interface.
-func (c *Client) ResetGeneric(ctx context.Context, req *machineapi.ResetRequest) (err error) {
+func (c *Client) ResetGeneric(ctx context.Context, req *machineapi.ResetRequest) error {
+	_, err := c.ResetGenericGetResponse(ctx, req)
+
+	return err
+}
+
+// ResetGenericGetResponse resets the machine and returns the response.
+func (c *Client) ResetGenericGetResponse(ctx context.Context, req *machineapi.ResetRequest) (*machineapi.ResetResponse, error) {
```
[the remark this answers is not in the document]
utkuozdemir: yes, better, done on all

pkg/machinery/client/events.go
utkuozdemir: done

go.work.sum
smira: yeah, same change for me, so let's commit it ;)

utkuozdemir: ok, pushed one more time, bringing it back.

pkg/machinery/client/events.go
```diff
@@ -123,6 +123,86 @@ func (c *Client) EventsWatch(ctx context.Context, watchFunc func(<-chan Event),
...
+
+	go func() {
+		for {
+			event, err := stream.Recv()
+			if err != nil {
+				if err == io.EOF || StatusCode(err) == codes.Canceled {
+					return
+				}
+
+				ch <- EventResult{Error: fmt.Errorf("error while watching events: %w", err)}
```
smira, line 169: I would probably wrap all `ch <-` sends with `select` on `ctx.Done()` to be on the safe side - if the caller walked away and cancelled the context, there's not much we can do, it might not be consuming the events anymore

smira (approved):

PR #6081, opened 2022-08-12
feat: add event actor id to client api and events cmd

utkuozdemir:
Add the missing actor id on the event and a way to filter by it to the events cli command.

Related to siderolabs/talos#5499.

### DIFF
--- a/pkg/machinery/client/events.go
+++ b/pkg/machinery/client/events.go
@@ -47,6 +47,13 @@ func WithTailDuration(dur time.Duration) EventsOptionFunc {
 	}
 }
 
+// WithActorID sets up Watcher to return events with the specified actor ID.
+func WithActorID(actorID string) EventsOptionFunc {
+	return func(opts *machineapi.EventsRequest) {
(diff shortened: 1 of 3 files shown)


pkg/machinery/client/events.go
```diff
@@ -47,6 +47,13 @@ func WithTailDuration(dur time.Duration) EventsOptionFunc {
 	}
 }
 
+// WithTailActorID sets up Watcher to return events with the specified actor ID.
```
smira, line 50:
```suggestion
// WithActorID sets up Watcher to return events with the specified actor ID.
```

smira (approved):

pkg/machinery/client/events.go
utkuozdemir: Not sure about this because it's in the client package and `client.WithActorID` doesn't tell much. All other tail builder functions follow the `WithTail*` naming. `WithTailEvents`, `WithTailID`, `WithTailDuration`.

PR #6094, opened 2022-08-15
chore: validate that etcd ca is not empty

frezbo:
Validate that the etcd CA is not empty in machine configuration.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
@@ -43,6 +43,8 @@ var (
 
 	// Security.
 
+	// ErrEmptyKeyCert denotes that crypto key/cert combination should not be empty.
+	ErrEmptyKeyCert = errors.New("key/cert combination should not be empty")
 	// ErrInvalidCert denotes that the certificate specified is invalid.
 	ErrInvalidCert = errors.New("certificate is invalid")
 	// ErrInvalidCertType denotes that the certificate type is invalid.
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -117,6 +119,10 @@ func (c *Config) Validate(mode config.RuntimeMode, options ...config.ValidationO
 
 	switch c.Machine().Type() {
 	case machine.TypeInit, machine.TypeControlPlane:
+		if c.Cluster().Etcd().CA() == nil {
```
smira, line 122: we have `ClusterConfig.Validate()` up there on line 86

frezbo: i didn't want to validate for non control plane machine since there's no etcd there, does that makes sense?

smira: we can pass the machine type to the `Validate()` function, but I think it still make sense to validate `EtcdConfig` in a single place in the code

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -792,10 +794,14 @@ func (k *KubeletConfig) Validate() ([]string, error) {
 	return nil, result.ErrorOrNil()
 }
 
-// Validate kubelet configuration.
+// Validate etcd configuration.
 func (e *EtcdConfig) Validate() error {
 	var result *multierror.Error
 
+	if e.CA() == nil {
```
smira, line 801: I don't remember if it runs for workers? on workers there's no requirement on etcd... or whole etcd config is nil?

frezbo: yeh, it's nil by default, unless someone adds those fields manually, in that case the other existing checks also fail

smira (approved):

PR #6100, opened 2022-08-15
chore: update API descriptors

smira:
Re-generate protobuf API descriptors in preparation for 1.2.0-beta.


frezbo (approved):
shouldn't the build fail if this is not up-to-date? Or is it only checking for deprecations/breaking changes in CI. Also if the creation of `api.descriptors` reproducible?

smira:
> shouldn't the build fail if this is not up-to-date? 

we should update this one per minor release, so that we make sure that during 1.3.0 cycle we don't break backwards compat with 1.2.0


> Or is it only checking for deprecations/breaking changes in CI.

yes, it verifies that current `.proto` files are backwards compatible with `api.descriptors`

as our `.proto` files are frozen now, and we're moving towards 1.3.0 cycle.

>  Also if the creation of `api.descriptors` reproducible?

yes

Issue #6098, opened 2022-08-15
Talosctl edit mc goes on a loop in the EDITOR if the save is not valid


frezbo:
## Bug Report

If the edits made to the Talos machineconfig is not valid, the EDITOR goes in a loop and the only way out is forcibly killing the talosctl process

PR #6102, opened 2022-08-15
fix: talosctl edit mc loop

frezbo:
Fixes re-opening editor forever when using `talosctl edit mc`.
Also fixes the temp dir getting filled up with temporary files created
for editing machine config.

Fixes: #6098

### DIFF
--- a/cmd/talosctl/cmd/talos/edit.go
+++ b/cmd/talosctl/cmd/talos/edit.go
@@ -7,8 +7,10 @@ package talos
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
+	"os"
 	"runtime"
 	"strings"
 	"time"
@@ -36,7 +38,10 @@ var editCmdFlags struct {
 
 //nolint:gocyclo
 func editFn(c *client.Client) func(context.Context, string, resource.Resource, error) error {
-	var lastError string
+	var (
+		path      string
+		lastError string
+	)
 
 	edit := editor.NewDefaultEditor([]string{
 		"TALOS_EDITOR",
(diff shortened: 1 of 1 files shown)


frezbo:
Code mostly borrowed from https://github.com/kubernetes/kubectl/blob/4da03973dd2fcd4645f20ac669d8a73cb017ff39/pkg/cmd/util/editor/editoptions.go#L315

cmd/talosctl/cmd/talos/edit.go
```diff
@@ -83,11 +88,7 @@ func editFn(c *client.Client) func(context.Context, string, resource.Resource, e
 			}
 
 			if lastError != "" {
-				lines := strings.Split(lastError, "\n")
-
-				_, err = w.Write([]byte(
-					fmt.Sprintf("# \n# %s\n", strings.Join(lines, "\n# ")),
```
smira, line 89: I think this had some `CRLF` magic for Windows due to the special writer being used which is not handled by the new function

frezbo: it's the same code as the old one

frezbo: https://github.com/siderolabs/talos/blob/e3c96117536a7dc55fb2f5817b1631819b7c6534/cmd/talosctl/cmd/talos/edit.go#L77 it's here

smira: ah... I see, sorry, we're still using the `w.Write`... I somehow looks at the signature and ignored that

smira (approved):

PR #6103, opened 2022-08-15
fix: surround `cancelCtx` with the mutex

Unix4ever:
Looks like `cancelCtx` access from the different goroutines wasn't
protected.


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_priority_lock.go
```diff
@@ -63,6 +64,9 @@ func (lock *PriorityLock[T]) setRunningPriority(seq T, cancelCtx context.CancelF
 		lock.cancelCtx()
 	}
 
+	lock.cancelCtxMu.Lock()
```
smira, line 67: I don't think we need yet another mutex.

Our `lock.mu` should already cover the access to that field.

I thought if `getRunningPriority` should also return `cancelCtx` ?

smira (approved):
:cool:

PR #6106, opened 2022-08-15
feat: update etcd advertised peer addresses on the fly

### LINKED ISSUE #6080: Update etcd advertised peer addresses
Part of #5668 story: as the advertised peer URLs might change, we might want to update them in real time using the etcd members API for each node.

Need to figure out this node ID and then update the members list accordingly.

smira:
This allows to update the member information (for the current node) with
new advertised peer URLs as the config changes.

E.g. if the node IP changes, this will update the peer URLs for the
member accordingly.

At the same time any member update requires quorum, so changing IPs can
only be done on node-by-node basis.

If there are no changes to advertised peer URLs, controller does
nothing.

Talos node might still need a reboot to update the listen addresses, as
these are not handled automatically for now.

Fixes #6080


frezbo (approved):
i guess testing this needs an embedded etcd?

smira:
> i guess testing this needs an embedded etcd?

yep, we need an actual IP change... I tested this manually by moving all etcd members to different IPs on the same subnet

PR #6107, opened 2022-08-16
fix: correct link to api access from k8s

ammmze:
# Pull Request

## What? (description)

Correct link to the Talos API access from Kubernetes.

## Why? (reasoning)

The previous link went to a non-existent page.

## Acceptance

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -82,7 +82,7 @@ talosctl inject serviceaccount -f manifests.yaml > manifests-injected.yaml
 kubectl apply -f manifests-injected.yaml
 ```
 
-See [documentation](https://www.talos.dev/v1.2/advanced/configuration/talos-api-access-from-k8s/) for more details.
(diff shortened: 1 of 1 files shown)


ammmze:
Looking at the conform.yaml and the conform policy code, it looks like that `conform/commit/gpg-identity` is expected to fail unless I am a member of the `siderolabs` Organization? So I guess that presumably means I'll need one you folks to co-author my commit?

frezbo (approved):

frezbo:
> Looking at the conform.yaml and the conform policy code, it looks like that `conform/commit/gpg-identity` is expected to fail unless I am a member of the `siderolabs` Organization? So I guess that presumably means I'll need one you folks to co-author my commit?

yes, we require all commits to be signed by one of our devs, so we do re-signing and push to HEAD for all community PR's

Unix4ever (approved):

PR #6112, opened 2022-08-16
fix: handle grub config being empty in the `Revert` function

Unix4ever:
Looks like it returns nil if it doesn't exist and the code doesn't
handle it properly.


internal/app/machined/pkg/runtime/v1alpha1/bootloader/meta.go
```diff
@@ -117,6 +117,10 @@ func (m *Meta) Revert() (err error) {
 		return err
 	}
 
+	if conf == nil {
```
frezbo, line 120: seems, this was missed in this file, all other places checked if the conf was nil :+1:

frezbo (approved):

PR #6113, opened 2022-08-17
fix: make 'ca', 'crt' and 'key' flags optional for 'talosctl config add'

sauterp:
As the 'ca', 'crt' and 'key' parameters are now optional for the talos
client, requiring them for the 'talosctl config add' command no longer
makes sense.

### DIFF
--- a/cmd/talosctl/cmd/talos/config.go
+++ b/cmd/talosctl/cmd/talos/config.go
@@ -23,7 +23,6 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 
 	"github.com/talos-systems/talos/cmd/talosctl/pkg/talos/helpers"
-	"github.com/talos-systems/talos/pkg/cli"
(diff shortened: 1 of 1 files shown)


frezbo (approved):

cmd/talosctl/cmd/talos/config.go
```diff
@@ -157,25 +156,33 @@ var configAddCmd = &cobra.Command{
...
+			}
+
+			newContext.Crt = base64.StdEncoding.EncodeToString(crtBytes)
 		}
 
-		newContext := &clientconfig.Context{
-			CA:  base64.StdEncoding.EncodeToString(caBytes),
-			Crt: base64.StdEncoding.EncodeToString(crtBytes),
-			Key: base64.StdEncoding.EncodeToString(keyBytes),
+		if configAddCmdFlags.key != "" {
```
frezbo, line 179: shouldn't we enforce if either key or cert is set?

frezbo (approved):

PR #6128, opened 2022-08-22
fix: stable default hostname bias

smira:
When converting to base36 a 256-bit number there's a bias in the
first character of the base36 encoding, as 256-bit number never fits
perfectly base 36 number.

To give an example, when converting 4-digit binary number to decimal,
the first digit of the decimal number will be [0..3], while the
second digit won't be biased:

```
0000 -> 00
0001 -> 01
...
0111 -> 15
1000 -> 16
...
1111 -> 31
```

Same issue happens when going from e.g. base16 to base36.

Stable hostnames were biased towards having a digit as the first
character.

The fix is to skip the first character of the base36 representation, and
also we don't need to convert all 256 bits to base36, if we use only 6
characters, we can save some CPU resources by taking only 8 bytes
instead of full 32 bytes.


frezbo (approved):

utkuozdemir (approved):
👍

PR #6130, opened 2022-08-22
fix: don't wait for the hostname in maintenance mode

### LINKED ISSUE #6119: talos 1.2.0a1 does not boot up to maintenance mode
### Description
Booting the iso of talos get stuck after the waiting for network address and hostname to be ready.
This happens since the 1.2.0-alpha1 version (I.e. the problem occurs with alpha2 and beta0 too).
Alpha0 version boot normaly and so does version 1.1.2.
First screen shot is the boot with 1.2.0alpha0 which boot instantly to maintenance state.
Second screenshot is the boot with the 1.2.0beta0 stuck. This screeshot has been taken after waiting for 5-6 minutes.
In both case, the same vm has been booted with different iso. dhcp server was the same.

### Logs
There is no log since the apid is not reachable. However I've been able to take screenshot of beta0 booting.
boot-1 2 0a0-ok
boot1 2 0b0-ko

### Environment

- Talos version: 1.2.0alpha1, alpha2, beta0
- Kubernetes version: N/A
- Platform: VMWare and bare metal (iso and ova).

smira:
Fixes #6119

With new stable default hostname feature, any default hostname is
disabled until the machine config is available.

Talos enters maintenance mode when the default config source is empty,
so it doesn't have any machine config available at the moment
maintenance service is started.

Hostname might be set via different sources, e.g. kernel args or via
DHCP before the machine config is available, but if all these sources
are not available, hostname won't be set at all.

This stops waiting for the hostname, and skips setting any DNS names in
the maintenance mode certificate SANs if the hostname is not available.

Also adds a regression test via new `--disable-dhcp-hostname` flag to
`talosctl cluster create`.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -397,6 +397,14 @@ local integration_kubespan = Step("e2e-kubespan", target="e2e-qemu", privileged=
         "IMAGE_REGISTRY": local_registry,
         "WITH_CONFIG_PATCH": '[{"op": "replace", "path": "/cluster/discovery/registries/kubernetes/disabled", "value": false}]', # use Kubernetes discovery backend
 });
+local integration_default_hostname = Step("e2e-default-hostname", target="e2e-qemu", privileged=true, depends_on=[integration_kubespan], environment={
+        # regression test: make sure Talos works in maintenance mode when no hostname is set
+        "SHORT_INTEGRATION_TEST": "yes",
+        "IMAGE_REGISTRY": local_registry,
+        "VIA_MAINTENANCE_MODE": "true",
+        "DISABLE_DHCP_HOSTNAME": "true",
+});
+
 local integration_qemu_encrypted_vip = Step("e2e-encrypted-vip", target="e2e-qemu", privileged=true, depends_on=[load_artifacts], environment={
         "WITH_DISK_ENCRYPTION": "true",
         "WITH_VIRTUAL_IP": "true",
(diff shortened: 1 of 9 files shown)


frezbo (approved):

hack/test/e2e-qemu.sh
```diff
@@ -63,6 +63,23 @@ case "${WITH_CONTROL_PLANE_PORT:-false}" in
     ;;
 esac
 
+case "${VIA_MAINTENANCE_MODE:-false}" in
+  false)
+    ;;
+  *)
+    # apply config via maintenance mode
+    QEMU_FLAGS="${QEMU_FLAGS} --disable-injecting-config --with-apply-config"
```
sbskas, line 71: e2e test seems to be failing because of unknown parameter --disable-injecting-config.

smira: yep, thanks, fix is coming :)

.drone.jsonnet
```diff
@@ -397,6 +397,14 @@ local integration_kubespan = Step("e2e-kubespan", target="e2e-qemu", privileged=
         "IMAGE_REGISTRY": local_registry,
         "WITH_CONFIG_PATCH": '[{"op": "replace", "path": "/cluster/discovery/registries/kubernetes/disabled", "value": false}]', # use Kubernetes discovery backend
 });
+local integration_default_hostname = Step("e2e-default-hostname", target="e2e-qemu", privileged=true, depends_on=[integration_kubespan], environment={
```
utkuozdemir, line 400: awesome

utkuozdemir (approved):
LGTM

PR #6136, opened 2022-08-23
chore: limit GOMAXPROCS for Talos services

### LINKED ISSUE #5971: limit GOMAXPROCS in Talos Go processes
Having high GOMAXPROCS (e.g. on 32+ on multi-core systems) might be a source of inefficiency for processes using less than maximum number of cores.

Talos's machined, apid, trustd should not ever use more than 2-4 cores in any real-life scenario.

So the proposal is to set `runtime.Numprocs` to the value:

* machined: max(numcpus, 4)
* apid: max(numcpus, 2)
* trustd: max(numcpus, 2)

smira:
Fixes #5971

### DIFF
--- a/pkg/startup/maxprocs.go
+++ b/pkg/startup/maxprocs.go
@@ -1,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package startup
+
+import (
+	"log"
+	"runtime"
+)
+
+// LimitMaxProcs limits the GOMAXPROCS to the number specified.
+func LimitMaxProcs(maxProcs int) {
+	curProcs := runtime.GOMAXPROCS(0)
+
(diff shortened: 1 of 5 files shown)


rsmitty (approved):

pkg/startup/maxprocs.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// LimitMaxProcs limits the GOMAXPROCS to the number specified.
+func LimitMaxProcs(maxProcs int) {
+	curProcs := runtime.GOMAXPROCS(0)
+
+	if curProcs > maxProcs {
+		runtime.GOMAXPROCS(maxProcs)
+
+		log.Printf("limited GOMAXPROCS to %d", maxProcs)
```
frezbo, line 19: should we also pass in the service name?

frezbo (approved):

pkg/startup/maxprocs.go
smira: it is printed to the service log actually, so it should be clear which service is that, e.g. `talosctl logs apid`

PR #6139, opened 2022-08-23
chore: move from inet.af/netaddr to net/netip and go4.org/netipx

### LINKED ISSUE #6007: netaddr -> netip refactoring
Refactor (if possible) Talos to use our own fork of `netaddr` which is compatible with `netip` package.

DmitriyMV:
Closes #6007


DmitriyMV:
There is only one instance of `netaddr` remaining - in [internal/app/machined/pkg/controllers/runtime/kmsg_log_test.go:24](https://github.com/siderolabs/talos/blob/master/internal/app/machined/pkg/controllers/runtime/kmsg_log_test.go#L24). That's because siderolink has talos in deps, so we need to update talos first.

go.mod
```diff
@@ -269,6 +269,7 @@ require (
 	go.starlark.net v0.0.0-20200306205701-8dd3e2ee1dd5 // indirect
 	go.uber.org/multierr v1.8.0 // indirect
 	go4.org/intern v0.0.0-20220617035311-6925f38cc365 // indirect
+	go4.org/netipx v0.0.0-20220812043211-3cc044ffd68d // indirect
```
smira, line 272: should it be direct dependency?

pkg/machinery/nethelpers/std_netaddr.go
```diff
@@ -6,24 +6,25 @@ package nethelpers
 
 import (
 	"net"
+	"net/netip"
 
-	"inet.af/netaddr"
```
smira, line 10: nit: this file might be renamed to `std_netip.go`

smira (approved):
also due to massive changes, it might make sense to `/promote integration`

go.mod
DmitriyMV: Rerun `go mod tidy` - should be good now.

pkg/machinery/nethelpers/std_netaddr.go
DmitriyMV: Done!

PR #6151, opened 2022-08-24
fix: bump rtnetlink to 1.2.2

smira:
This updates release-1.2 directly, as master was already bumped.

This fixes compatibility with host Linux 5.19. Talos comes with Linux
5.15, so it's not a problem usually, but when running Talos in Docker
when host Linux kernel is 5.19+ this becomes an issue.


smira:
See https://github.com/jsimonetti/rtnetlink/pull/146

frezbo (approved):

rsmitty (approved):

PR #6165, opened 2022-08-29
feat: add structprotogen tool

### LINKED ISSUE #6078: Generate `.proto` files from Go structs.
Now that we have implemented and added Go-structs 2 protobuf encoder/decoder we should add proto generator. 

Relevant:
- https://github.com/siderolabs/talos/issues/4417
- https://github.com/siderolabs/talos/issues/4416

DmitriyMV:
This commit adds structprotogen tool which is used to generate proto file from Go structs.

Closes #6078.


DmitriyMV:
I'm not sure about `website/content/v1.2/reference/api.md` at all.

api/definitions/definitions.proto
```diff
@@ -0,0 +1,895 @@
+syntax = "proto3";
```
smira, line 1: looks like we have two copies of this file?

should we put it under `resource/*` ?

api/definitions/definitions.proto
```diff
@@ -0,0 +1,895 @@
...
+
+package definitions;
```
smira, line 3: I think proto files should mirror as much as possible structure of the resources, so that messages has more meaningful names, e.g. `cluster.proto` with `cluster/*` resources? and there should be a package `talos.cluster`, so that structure name is `talos.cluster.AffiliateSpec` for example

DmitriyMV: Splitting files is okay, but dependencies is the thing I really don't want to implement, since it would require us to implement dependency graph. Are were sure we want it?

DmitriyMV: A since proto doesn't handle cyclic dependencies you can't just generate one stab and reuse it everywhere.

api/definitions/definitions.proto
DmitriyMV: Is it okay to put it there? I mean we usually store it in `/api`. `make docs` expect `.proto` files to be there.

smira: I mean under `/api/resource/*`, as we already have it

api/definitions/definitions.proto
```diff
@@ -0,0 +1,895 @@
...
+import "google/protobuf/timestamp.proto";
+
+// AffiliateSpec describes Affiliate state.
+message ClusterAffiliateSpec {
+  string node_id = 1;
+  repeated common.NetIP addresses = 2;
+  string hostname = 3;
+  string nodename = 4;
+  string operating_system = 5;
+  int64 machine_type = 6;
```
smira, line 17: this is an interesting one... as it should be an `enum` type actually, and things like this one which are constants in Go with enumeration

we can create it as `message`-local enum to avoid dependencies, something like:

```proto
message A {
   enum B {
      A = 0;
      B = 1;
   }
   
   B enum_field = 3;
}
```

api/definitions/definitions.proto
smira: I don't see how there can be any cyclic dependencies between resources in different packages. Can you provide an example?

api/definitions/definitions.proto
DmitriyMV: Done!

DmitriyMV:
Redone proto generator from the start. It's now properly handles types, including enum types (which are hardcoded for now).

Dockerfile
```diff
@@ -178,6 +182,23 @@ RUN --mount=type=cache,target=/.cache prototool format --overwrite --protoc-bin-
...
+RUN gofumpt -w ./pkg/
+WORKDIR /src/pkg/machinery
+RUN --mount=type=cache,target=/.cache go generate ./...
+RUN gotagsrewrite .
+RUN goimports -w -local github.com/talos-systems/talos ./
+RUN gofumpt -w ./
+
+FROM go-generate AS gen-proto-go
+WORKDIR /src/
+RUN structprotogen github.com/talos-systems/talos/pkg/machinery/... /api/resource/definitions/
```
smira, line 200: should we add `--mount=type=cache,target=/.cache` to re-use build cache?

I wonder if we need same for `gotagsrewrite` ?

website/content/v1.2/reference/api.md
```diff
@@ -24,6 +24,147 @@ description: Talos gRPC API reference.
   
     - [File-level Extensions](#common/common.proto-extensions)
   
+- [resource/definitions/cluster/cluster.proto](#resource/definitions/cluster/cluster.proto)
```
smira, line 27: I guess this should be reverted (changes only for 1.3)?

api/resource/definitions/network/network.proto
```diff
@@ -0,0 +1,300 @@
+syntax = "proto3";
+
+package network;
```
smira, line 3: I think we need longer fully-qualified package names: https://developers.google.com/protocol-buffers/docs/style

e.g. `package talos.resource.definitions.network`

website/content/v1.2/reference/api.md
DmitriyMV: What do you mean by reverted? This is 1.3 doc generated, so shouldn't this be included?

smira (approved):
:tada:

Dockerfile
DmitriyMV: gotagsrewrite doesn't use type information, only plain AST so it doesn't need build cache.

As for this - I dunno, it does use type checking in but I'm unsure if it uses go build cache. Will check.

website/content/v1.2/reference/api.md
smira: I mean we shouldn't touch 1.2 docs anymore. 1.2 docs should be frozen until we decide to do some form of the backport.

DmitriyMV: Oh, right! I removed leftovers.

api/resource/definitions/network/network.proto
DmitriyMV: Done!

Dockerfile
DmitriyMV: I checked - it does use `~/.cache` but I didn't notice any performance hit even with  empty `~/.cache/go-build` directory.

PR #6167, opened 2022-08-29
fix: always abort the maintenance service

smira:
I hit this bug when one the API calls got hanging, and submitting the
machine config with `apply-config` never takes the node out of
maintenance mode, as `.GracefulStop()` may hang forever waiting for all
the calls to finish.

This way we always abort at some timeout and stop the server forcefully.

### DIFF
--- a/internal/app/maintenance/main.go
+++ b/internal/app/maintenance/main.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"net"
 	"net/netip"
+	"time"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

frezbo (approved):

smira:
> > submitting the machine config with apply-config never takes the node out of maintenance mode, as .GracefulStop() may hang forever waiting for all the calls to finish.
> 
> don;t we have the iso test that applies the machine config in maintenance mode?

yeah, we do, it was a specific condition which I triggered there with a "hanging" API call

frezbo:
> yeah, we do, it was a specific condition which I triggered there with a "hanging" API call

do we need an integration test later to catch these?

smira:
> > yeah, we do, it was a specific condition which I triggered there with a "hanging" API call
> 
> do we need an integration test later to catch these?

It was a bug in COSI (https://github.com/cosi-project/runtime/pull/140), so I don't know if we can actually trigger that without a bug.

PR #6170, opened 2022-08-29
fix: limit apid backoff max delay

smira:
This fixes a case when a node is rebooted, and connection via another
endpoint apid "caches" a connection error even when the node is up.

E.g. this command:

```
talosctl -e IP1 -n IP2 version
```

If node `IP2` is rebooted, `apid` at `IP1` might enter long backoff loop
and return an error still when `IP2` is actually up.


frezbo (approved):

utkuozdemir:
👍

PR #6183, opened 2022-08-31
fix: properly handle `configContext` being `nil` in Talos client

Unix4ever:
Client crashes if you try using it in the unixSocket mode.


frezbo (approved):

Unix4ever:
.ь

Issue #6193, opened 2022-09-01
kube-proxy rbac


frezbo:
## Bug Report

Fixup kube-proxy rbac

PR #6197, opened 2022-09-01
fix: drop kube-system SA default binding

smira:
This is not needed anymore, it's a leftover from bootkube times.


frezbo:
Fixes: #6193

PR #6202, opened 2022-09-02
docs: fork docs for v1.3

smira:
Now master docs are generate for v1.3.0.


hack/release.toml
```diff
@@ -19,117 +18,6 @@ preface = """\
     [notes.updates]
         title = "Component Updates"
         description="""\
```
frezbo, line 20: the flannel update should be here right?

smira: I have two PRs, I can't do both at once :-D

PR #6203, opened 2022-09-02
fix: prevent panic on health check if a member has no IPs

utkuozdemir:
If a member has no IP addresses, prevent cluster health checks from failing with a panic by checking for the length of member IPs and not assuming there's always at least 1 IP.


frezbo (approved):

pkg/cluster/check/discovery.go
```diff
@@ -72,6 +73,10 @@ func memberToNodeInfo(member *clussterres.Member) (cluster.NodeInfo, error) {
 		return cluster.NodeInfo{}, err
 	}
 
+	if len(ips) == 0 {
+		return cluster.NodeInfo{}, fmt.Errorf("no IP address found for member: %s", member.Metadata().String())
```
smira, line 77: I think `member.Metadata().String()` is too cryptic, we can use `ID()` which should be a hostname/nodename

smira (approved):

Issue #6206, opened 2022-09-02
Add support for setting kernel module parameters


mdallaire:
## Feature Request
It is currently not possible to set kernel module parameters in talos.
### Description
I want to use the OSS nvidia modules with my consumer grade GPUs. Since the consumer grade GPUs are not officially supported by the nvidia OSS drivers I get the following at boot:

```
[talos] boot sequence: done: 28.819444293s
10.0.20.203: kern: warning: [2022-09-02T19:03:16.922855838Z]: NVRM: Open nvidia.ko is only ready for use on Data Center GPUs.
10.0.20.203: kern: warning: [2022-09-02T19:03:16.923935838Z]: NVRM: To force use of Open nvidia.ko on other GPUs, see the
10.0.20.203: kern: warning: [2022-09-02T19:03:16.924966838Z]: NVRM: 'OpenRmEnableUnsupportedGpus' kernel module parameter described
10.0.20.203: kern: warning: [2022-09-02T19:03:16.926121838Z]: NVRM: in the README.
10.0.20.203: kern: warning: [2022-09-02T19:03:17.235131838Z]: NVRM: GPU 0000:01:00.0: RmInitAdapter failed! (0x63:0x0:1905)
10.0.20.203: kern: warning: [2022-09-02T19:03:17.236812838Z]: NVRM: GPU 0000:01:00.0: rm_init_adapter failed, device minor number
```
Extract from [the readme](https://us.download.nvidia.com/XFree86/Linux-x86_64/515.65.01/README/kernel_open.html) 

> Use of the open kernel modules on GeForce and Workstation GPUs should be considered alpha-quality in this release due to the missing features listed above. To enable use of the open kernel modules on GeForce and Workstation GPUs, set the "NVreg_OpenRmEnableUnsupportedGpus" nvidia.ko kernel module parameter to 1. E.g.,
>
>   ` modprobe nvidia NVreg_OpenRmEnableUnsupportedGpus=1`
>or, in an /etc/modprobe.d/ configuration file:
>
>    `options nvidia NVreg_OpenRmEnableUnsupportedGpus=1`

PR #6209, opened 2022-09-05
feat: support setting kernel parameters

frezbo:
Support setting kernel parameters via machine config.

Fixes: #6206


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2549,4 +2549,7 @@ type KernelModuleConfig struct {
 	// description: |
 	//   Module name.
 	ModuleName string `yaml:"name"`
+	// description: |
+	//   Module parameters.
```
smira, line 2553: we might want to add a note here that changing module parameters requires a reboot

smira (approved):
:cool:

PR #6219, opened 2022-09-05
fix: automatically discard VIPs for etcd advertised addresses

### LINKED ISSUE #6210: etcd shouldn't advertise VIPs by default

smira:
Fixes #6210

Refactored the code a bit to support excludes and default configuration.

Etcd should never advertise VIPs, as VIPs are managed by etcd elections.

### DIFF
--- a/internal/app/machined/pkg/controllers/etcd/config.go
+++ b/internal/app/machined/pkg/controllers/etcd/config.go
@@ -92,12 +92,27 @@ func (ctrl *ConfigController) Run(ctx context.Context, r controller.Runtime, log
 			return fmt.Errorf("error getting machine config: %w", err)
 		}
 
-		if err = safe.WriterModify(ctx, r, etcd.NewConfig(etcd.NamespaceName, etcd.ConfigID), func(status *etcd.Config) error {
-			status.TypedSpec().AdvertiseValidSubnets = machineConfig.Config().Cluster().Etcd().AdvertisedSubnets()
-			status.TypedSpec().ListenValidSubnets = machineConfig.Config().Cluster().Etcd().ListenSubnets()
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/etcd/config.go
```diff
@@ -92,12 +92,27 @@ func (ctrl *ConfigController) Run(ctx context.Context, r controller.Runtime, log
 			return fmt.Errorf("error getting machine config: %w", err)
 		}
 
-		if err = safe.WriterModify(ctx, r, etcd.NewConfig(etcd.NamespaceName, etcd.ConfigID), func(status *etcd.Config) error {
-			status.TypedSpec().AdvertiseValidSubnets = machineConfig.Config().Cluster().Etcd().AdvertisedSubnets()
-			status.TypedSpec().ListenValidSubnets = machineConfig.Config().Cluster().Etcd().ListenSubnets()
+		if err = safe.WriterModify(ctx, r, etcd.NewConfig(etcd.NamespaceName, etcd.ConfigID), func(cfg *etcd.Config) error {
+			cfg.TypedSpec().AdvertiseValidSubnets = machineConfig.Config().Cluster().Etcd().AdvertisedSubnets()
+			cfg.TypedSpec().AdvertiseExcludeSubnets = nil
```
frezbo, line 97: excludesubnets already defined earlier?

frezbo (approved):

internal/app/machined/pkg/controllers/etcd/config.go
smira: this makes sure we start with an empty list, otherwise `append()` below will keep appending more and more copies of the same VIP

frezbo: https://github.com/siderolabs/talos/blob/dce923f7471a73623c7bc99e376e51a53401faae/pkg/machinery/resources/etcd/config.go#L30 i was not sure if it was defined, checked and confirmed.

PR #6220, opened 2022-09-05
chore: bump dependencies

smira:
dependabot


hack/images.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
frezbo, line 1: part of this or used for release?

frezbo (approved):

hack/images.sh
smira: yep, I kind of added it by accident, but it makes sense to get it pushed, as I used it for couple of months

Unix4ever (approved):

PR #6235, opened 2022-09-08
fix: list COSI APIs for the apid authenticator

smira:
As APIs were not listed explicitly, access with `os:reader` was denied by default, while it should have been checked down in the access filter.

### DIFF
--- a/internal/app/machined/pkg/system/services/machined.go
+++ b/internal/app/machined/pkg/system/services/machined.go
@@ -75,6 +75,12 @@ var rules = map[string]role.Set{
 	"/resource.ResourceService/Get":   role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/List":  role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/Watch": role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/Create":     role.MakeSet(role.Admin),
+	"/cosi.resource.State/Destroy":    role.MakeSet(role.Admin),
+	"/cosi.resource.State/Get":        role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/List":       role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/Update":     role.MakeSet(role.Admin),
+	"/cosi.resource.State/Watch":      role.MakeSet(role.Admin, role.Reader),
 
 	"/storage.StorageService/Disks": role.MakeSet(role.Admin, role.Reader),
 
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/system/services/machined.go
```diff
@@ -75,6 +75,12 @@ var rules = map[string]role.Set{
 	"/resource.ResourceService/Get":   role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/List":  role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/Watch": role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/Create":     role.MakeSet(role.Admin, role.Reader),
```
frezbo, line 78: create for `os:reader`?

smira: yep, it will be denied in the resource filter; this line basically says "allow with any role", but on other hand probably better to remove reader here

smira: yep, fixed them all, thanks

internal/app/machined/pkg/system/services/machined.go
```diff
@@ -75,6 +75,12 @@ var rules = map[string]role.Set{
 	"/resource.ResourceService/Get":   role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/List":  role.MakeSet(role.Admin, role.Reader),
 	"/resource.ResourceService/Watch": role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/Create":     role.MakeSet(role.Admin),
+	"/cosi.resource.State/Destroy":    role.MakeSet(role.Admin),
+	"/cosi.resource.State/Get":        role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/List":       role.MakeSet(role.Admin, role.Reader),
+	"/cosi.resource.State/Update":     role.MakeSet(role.Admin),
+	"/cosi.resource.State/Watch":      role.MakeSet(role.Admin, role.Reader),
```
frezbo, line 83: watch implies secrets too right?

frezbo (approved):

internal/app/machined/pkg/system/services/machined.go
smira: yep, but we have a check down before we get to actual resources to make a decision on per-resource types

internal/integration/cli/config.go
```diff
@@ -127,7 +127,7 @@ func (suite *TalosconfigSuite) TestNew() {
 			adminOpts: []base.RunOption{base.StdoutShouldMatch(regexp.MustCompile(`MachineConfig`))},
 			readerOpts: []base.RunOption{
 				base.ShouldFail(),
-				base.StdoutEmpty(),
+				base.StdoutShouldMatch(regexp.MustCompile(`\QNODE   NAMESPACE   TYPE   ID   VERSION`)),
```
frezbo, line 130: so now we have empty tables output?

frezbo (approved):

internal/integration/cli/config.go
smira: yep, as in fact client _can_ fetch the resource definition (as it's allowed), but server denies listing actual resource

PR #6236, opened 2022-09-08
feat: update CoreDNS to 1.9.4

smira:
See https://github.com/coredns/coredns/blob/master/notes/coredns-1.9.4.md

### DIFF
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -311,7 +311,7 @@ const (
 	CoreDNSImage = "docker.io/coredns/coredns"
 
 	// DefaultCoreDNSVersion is the default version for the CoreDNS.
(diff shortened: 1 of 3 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -311,7 +311,7 @@ const (
 	CoreDNSImage = "docker.io/coredns/coredns"
 
 	// DefaultCoreDNSVersion is the default version for the CoreDNS.
-	DefaultCoreDNSVersion = "1.9.3"
+	DefaultCoreDNSVersion = "1.9.4"
```
frezbo, line 314: we should renovatify this :sweat_smile:

smira: :+1:

PR #6237, opened 2022-09-08
feat: check apid client certificate extended key usage

smira:
This is enabled via a machine config feature/version contract, as `talosconfig` certificate generated previously didn't have proper key usage set, so we need to keep backwards compatibility on upgrades.

New v1.3+ clusters will include this check.

This check prevents even potential mis-use of server certificates as a client certificate.

### DIFF
--- a/internal/app/apid/main.go
+++ b/internal/app/apid/main.go
@@ -6,10 +6,12 @@ package apid
 
 import (
 	"context"
+	"crypto/x509"
 	"flag"
 	"fmt"
 	"log"
 	"os/signal"
+	"reflect"
 	"regexp"
 	"syscall"
 	"time"
@@ -34,8 +36,6 @@ import (
 	"github.com/talos-systems/talos/pkg/startup"
 )
 
-var rbacEnabled *bool
-
 func runDebugServer(ctx context.Context) {
 	const debugAddr = ":9981"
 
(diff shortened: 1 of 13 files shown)


internal/app/apid/main.go
```diff
@@ -215,3 +221,23 @@ func apidMain() error {
 
 	return errGroup.Wait()
 }
+
+func verifyExtKeyUsage(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+	if len(verifiedChains) == 0 {
+		return fmt.Errorf("no verified chains")
+	}
+
+	certs := verifiedChains[0]
```
frezbo, line 230: does this always have a slice of slice with exactly one cert?

internal/app/apid/main.go
```diff
@@ -215,3 +221,23 @@ func apidMain() error {
 
 	return errGroup.Wait()
 }
+
+func verifyExtKeyUsage(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
```
frezbo, line 225: `rawCerts` not used?

internal/app/apid/main.go
smira: we check for chains length above, so at least one chain is there (it should always be there, as we require client cert)

so we walk through one chain and make sure there's a leaf cert which has proper key usage.

in practice there should be a single chain

internal/app/apid/main.go
smira: yes, it's a callback from stdlib, so we have to have expected signature

frezbo (approved):

internal/app/apid/main.go
frezbo: > in practice there should be a single chain

should we fail for some reason if there's more than one?

smira: I think if there's a single verified chain which has proper client cert, we should consider that to be positive check result - it means that client presented at least a single valid client certificate with proper key usage; if there's another chain, it doesn't matter if it has even wrong cert, as there's a valid one

Unix4ever (approved):

PR #6240, opened 2022-09-09
fix: kill the task processes when cleaning up stale task

smira:
The bug was triggered by `containerd` crash (restart), in this case runner receives an error as if the process exited. Runner tries to restart the container, but as the container is still running, attempt to delete the task would fail.

With this change Talos always tries to kill the running container and waits for the container to terminate.

The error message when the bug was triggered looks like:

```
service[kubelet](Waiting): Error running Containerd(kubelet), going to restart forever: failed to clean up task "kubelet": task must be stopped before deletion: running: failed precondition
```

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd.go
@@ -124,7 +124,7 @@ func (c *containerdRunner) Close() error {
 
 // Run implements runner.Runner interface
 //
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (c *containerdRunner) Run(eventSink events.Recorder) error {
 	defer close(c.stopped)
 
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd.go
```diff
@@ -137,6 +137,24 @@ func (c *containerdRunner) Run(eventSink events.Recorder) error {
 	// attempt to clean up a task if it already exists
 	task, err = c.container.Task(c.ctx, nil)
 	if err == nil {
+		var s <-chan containerd.ExitStatus
+
+		s, err = task.Wait(c.ctx)
```
frezbo, line 142: do we need some timeout, like in a case of reboot/reset and the task hangs forever, or do we have a top level context that cancels this?

smira: it should be the `<-c.stop` receive below which will be triggered if something tries to stop the service

`Wait` itself is asynchronous - it returns immediately

frezbo (approved):

PR #6253, opened 2022-09-12
fix: stop containers more thoroughly

smira:
Don't skip pods which are not ready, try still to stop containers inside not ready pod sandboxes.

Re-enable the test with Canal CNI (upstream Calico got fixed).

### DIFF
--- a/internal/pkg/cri/pods.go
+++ b/internal/pkg/cri/pods.go
@@ -13,6 +13,8 @@ import (
 	"google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+
+	"github.com/talos-systems/talos/pkg/machinery/constants"
 )
 
 // RunPodSandbox creates and starts a pod-level sandbox. Runtimes should ensure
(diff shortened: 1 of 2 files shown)


internal/pkg/cri/pods.go
```diff
@@ -182,18 +178,22 @@ func stopAndRemove(ctx context.Context, stopAction StopAction, client *Client, p
...
 
 		g.Go(func() error {
-			log.Printf("%s container %s/%s:%s", action, pod.Metadata.Namespace, pod.Metadata.Name, container.Metadata.Name)
-
 			// TODO(andrewrynhard): Can we set the timeout dynamically?
-			if criErr := client.StopContainer(ctx, container.Id, 30); criErr != nil {
-				if grpcstatus.Code(criErr) == codes.NotFound {
-					return nil
-				}
+			if container.State == runtimeapi.ContainerState_CONTAINER_RUNNING || container.State == runtimeapi.ContainerState_CONTAINER_UNKNOWN {
```
frezbo, line 182: do we have any potential other states in between as reported by cri, maybe a container is still restarting/not healthy?

internal/pkg/cri/pods.go
```diff
@@ -213,14 +213,16 @@ func stopAndRemove(ctx context.Context, stopAction StopAction, client *Client, p
 		return err
 	}
 
-	if err = client.StopPodSandbox(ctx, pod.Id); err != nil {
-		if grpcstatus.Code(err) == codes.NotFound {
-			return nil
-		}
+	if pod.State == runtimeapi.PodSandboxState_SANDBOX_READY {
```
frezbo, line 216: what if a pod failed to start due to cni error, is it okay to go ahead?

smira: we just don't try to stop a pod if it's not ready, as if it's not ready, it got stopped already

internal/pkg/cri/pods.go
smira: these are two different states, there's a sandbox (pod) state and a container state. 

we are in fact interested in stopping containers, as they hold the mounts, etc. 

we stop the pods as well, but that mostly means tearing down CNI stuff

frezbo (approved):

PR #6269, opened 2022-09-14
docs: update docs for pod security

paucampana:
add new section to see how to disable admission control in control plane

Signed-off-by: Pau Campana <pau.campanya.soler@gmail.com>

# Pull Request

Add a section for how admission control can be disabled in the control plane

## Why? (reasoning)

It is not clear how it could be disabled.

## Acceptance

### DIFF
--- a/website/content/v1.2/learn-more/knowledge-base.md
+++ b/website/content/v1.2/learn-more/knowledge-base.md
@@ -84,3 +84,15 @@ machine:
 ```
 
 > Note: Talos needs to be upgraded for the `extraKernelArgs` to take effect.
+
+## Disable `admissionControl` on control plane nodes
+
+Talos Linux enables admission control in the API Server by default.
+
+Although it is not recommended from a security point of view, admission control can be removed by patching your control plane machine configuration:
+
+```bash
+talosctl gen config \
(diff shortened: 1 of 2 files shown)


smira (commented):
I'm torn on whether we keep it documented here. I don't want to give a bad advice.

Other people might have different ideas, I thought about having it in a Knowledge Base?

But I would really like people to tag namespaces with different policies if they need to vs. disabling it altogether.

frezbo:
> But I would really like people to tag namespaces with different policies if they need to vs. disabling it altogether.

Agree on this, this default policies prevent a lot of mis-configurations or potential privilege escalations. It easy to add exceptions while keeping the policy enabled.

frezbo:
> In our case for example, we are using [gatekeeper](https://github.com/open-policy-agent/gatekeeper).

that's a valid use case, let's get this moved over to the knowledgebase then

paucampana:
> > In our case for example, we are using [gatekeeper](https://github.com/open-policy-agent/gatekeeper).
> 
> that's a valid use case, let's get this moved over to the knowledgebase then

Added in knowledge base section

smira (approved):
thank you!

PR #6275, opened 2022-09-15
feat: update etcd to v3.5.5

smira:
See https://github.com/etcd-io/etcd/releases/tag/v3.5.5


frezbo (approved):
should we add the new flags?

`--experimental-compact-hash-check-enabled --experimental-compact-hash-check-time`

smira:
> should we add the new flags?
> 
> `--experimental-compact-hash-check-enabled --experimental-compact-hash-check-time`

this is tricky... as we allow etcd version to be set in the machine config.

frezbo:
> > should we add the new flags?
> > `--experimental-compact-hash-check-enabled --experimental-compact-hash-check-time`
> 
> this is tricky... as we allow etcd version to be set in the machine config.

ahh yes, but at the same time a lot of users probably run a single node cp, I'm not sure what's the right approach

smira:
> --experimental-compact-hash-check-enabled --experimental-compact-hash-check-time

so what I will do here: in this PR I will leave it as it was, so that is backportable to 1.2.x, and in 1.3.x we will enable the check

PR #6276, opened 2022-09-15
feat: enable etcd consistency hashcheck

smira:
This will be only enabled for Talos v1.3.x.


internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -409,6 +409,7 @@ func (e *Etcd) argsForInit(ctx context.Context, r runtime.Runtime, spec *etcdres
 		"peer-trusted-ca-file":               constants.EtcdCACert,
 		"experimental-initial-corrupt-check": "true",
 		"experimental-watch-progress-notify-interval": "5s",
+		"experimental-compact-hash-check-enabled":     "true",
```
frezbo, line 412: should we add a machineconfig validation such that we check if etcd is set to < 3.5.0 since it's overridable?

frezbo: let's go ahead with this.

PR #6300, opened 2022-09-20
chore: add ice drivers

frezbo:
Add Intel ice drivers and bump kernel to 5.15.68


smira (approved):
do we need to do same firmware on arm64?

bump of kernel version in constants.go is missing... and probably hack/release.toml (but I have it in my PR, I can probably fix it up there)

frezbo:
> do we need to do same firmware on arm64?

I'm not sure, i'll add it just in case.

frezbo:
> > do we need to do same firmware on arm64?
> 
> I'm not sure, i'll add it just in case.

This increases the initramfs size by 632KB when uncompressed.

smira:
> > > do we need to do same firmware on arm64?
> > 
> > 
> > I'm not sure, i'll add it just in case.
> 
> This increases the initramfs size by 632KB when uncompressed.

I would go this way for the sake of consistency... There might be arm64 servers with ICE? on other hand it's pretty rare

smira (approved):

PR #6319, opened 2022-09-23
feat: custom audit policy

sergelogvinov:
# Pull Request

## Why? (reasoning)

Unfortunately, the default audit policy file does not cover all cases. 
So I've added the resource `AuditPolicyConfigs` to solve it.

```yaml
# machineconfig
cluster:
  apiServer:
    auditPolicy:
      apiVersion: audit.k8s.io/v1
      kind: Policy
      omitStages:
        - RequestReceived
      rules:
        - level: Metadata
```

```sh
# talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.11 get AuditPolicyConfigs.kubernetes.talos.dev -oyaml
node: 192.168.10.11
metadata:
    namespace: controlplane
    type: AuditPolicyConfigs.kubernetes.talos.dev
    id: audit-policy
    version: 1
    owner: config.K8sControlPlaneController
    phase: running
    created: 2022-09-22T16:30:52Z
    updated: 2022-09-22T16:30:52Z
spec:
    config:
        apiVersion: audit.k8s.io/v1
        kind: Policy
        omitStages:
            - RequestReceived
        rules:
            - level: Metadata
```

```sh
# talosctl --talosconfig _cfgs/talosconfig --nodes 192.168.10.11 read /system/config/kubernetes/kube-apiserver/auditpolicy.yaml
apiVersion: audit.k8s.io/v1
kind: Policy
metadata:
  creationTimestamp: null
omitStages:
- RequestReceived
rules:
- level: Metadata
```

## Acceptance


smira (commented):
:heart:

smira:
@sergelogvinov please let me know if you need help with this one

sergelogvinov:
> @sergelogvinov please let me know if you need help with this one

If it is to easily understand what is wrong here, it will be nice ) 
Unfortunately, I cannot decide by logs...

smira (approved):

internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
```diff
@@ -139,7 +160,7 @@ func (ctrl *RenderConfigsStaticPodController) Run(ctx context.Context, r control
 
 		if err = r.Modify(ctx, k8s.NewConfigStatus(k8s.ControlPlaneNamespaceName, k8s.ConfigStatusStaticPodID), func(r resource.Resource) error {
 			r.(*k8s.ConfigStatus).TypedSpec().Ready = true
-			r.(*k8s.ConfigStatus).TypedSpec().Version = admissionRes.Metadata().Version().String()
+			r.(*k8s.ConfigStatus).TypedSpec().Version = admissionRes.Metadata().Version().String() + auditRes.Metadata().Version().String()
```
frezbo, line 163: adding up versions?

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1569,6 +1581,13 @@ type APIServerConfig struct {
 	//   examples:
 	//     - value: admissionControlConfigExample
 	AdmissionControlConfig AdmissionPluginConfigList `yaml:"admissionControl,omitempty"`
+	//   description: |
+	//     Configure the API server audit policy.
+	//
+	//     Defaults to the audit policy shown in the example.
+	//   examples:
+	//     - value: auditPolicyConfigDefault
+	AuditPolicyConfig Unstructured `yaml:"auditPolicy,omitempty"`
```
frezbo, line 1590: we'd need to sill generate the existing default configs in `init` and `worker` files if none is provided,

internal/app/machined/pkg/controllers/k8s/render_config_static_pods.go
smira: these are strings, so technically concatenating them. bumping the version here bumps the static pod manifest definition, so that kubelet reloads the pod

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
smira: it defaults to existing default policy if not set, so we could print it into the default config... idk, both ways are valid

frezbo: i think it makes more sense to add to the generated config since it's now a fieldd in the machineconfig, less confusion, rather than having a default. Ahh on second thoughts I think it's fine. Just need to reflect in release notes. I'm little conflicted here :sweat_smile:

smira: we can do both actually... nothing stops us from doing both: for backwards compat, we inject the default if missing, and also make it explicit in the generated machine config

frezbo: > we can do both actually... nothing stops us from doing both: for backwards compat, we inject the default if missing, and also make it explicit in the generated machine config

sounds like a plan :+1:

smira: updated this way!

frezbo (approved):

sergelogvinov:
Thank you 👍

PR #6326, opened 2022-09-25
feat: use rel=canonical in the head of docs pages

myndzi:
# Pull Request

## What? (description)
Adds `<link rel="canonical">` tags to page headers

## Why? (reasoning)
Google is indexing and preferring old versions of the docs. Specifying a canonical target will hint to Google that newer versions of the page are preferable to surface to searchers.

In SEO terms, this is preferable to setting `noindex` on pages -- old pages will still be indexed and appear in search results (e.g. for queries like `inurl:talos.dev/v1.0`, but ranking signals for old docs pages will be "forwarded to" the new ones when new versions get published, and Google should generally surface the most recent docs when people perform generic searches (e.g. `talos linux iscsi`)

It is peaceful to include a self-referencing rel=canonical tag.

If this is not behavior you want, feel free to reject/close the PR with no explanation :) 

(It seemed unfortunate to edit a file inside of what appears to be a vendor directory, but the docsy docs themselves suggest that this is kosher -- at least for hooks -- and it seemed the cleanest way to do it)

## Acceptance

### DIFF
--- a/website/themes/docsy/layouts/partials/head.html
+++ b/website/themes/docsy/layouts/partials/head.html
@@ -11,6 +11,8 @@
 {{ else -}}
 <meta name="robots" content="noindex, nofollow">
 {{ end -}}
+{{ $latest_doc := partial "doc_latest_version.html" . -}}
+<link rel="canonical" href="{{ $latest_doc | safeURL }}" />
 
 {{ partialCached "favicons.html" . }}
 <title>


steverfrancis:
@rsmitty we already do this now, yes?

rsmitty:
yes, as far as I understand, this is already being done with https://github.com/siderolabs/talos/pull/5391/files

rsmitty:
image

Looking at the 1.1 docs, at appears this has what's needed unless I'm mistaken about something.

TimJones:
My original PR should be generating this, but it looks as if the whole `head-end.html` hook isn't being rendered at all hence the loss of the canonical links. I will look into it and try to solve there.

steverfrancis:
We can see that the google search provided (e.g. `talos linux iscsi`) was returning the v0.11 page - although that page did have the rel canonical link - google is just ignoring it.

I've requested reindexing and validation of those explicit pages, and our sitemap, so that should (May?) help - but we already have the right code, thanks!

myndzi:
Ah, I didn't get the notification for the ping -- sorry!

Strange, I didn't see where in the code this was added but I can indeed see from "view source" that it seems to be present in the current version of e.g. the v1.1 docs (the search you gave gives an example). It just seemed that every search I perform lands me on old docs versions, so I'm not entirely sure what the deal is then. I'll check up on something and see if there's an easy explanation...

PR #6333, opened 2022-09-26
feat: implement 'permanent addr' in link statuses

smira:
Permanent address is only available for physical links, and it might be different from the 'hardware address': when bonding, 'hardware address' gets overridden from the bond master, while 'permanent address' still shows MAC of the interface.

This part of the fix for incorrect bonding issue on Equinix Metal.


frezbo (approved):
:+1: took me a while to figure out the change :sweat_smile:

PR #6340, opened 2022-09-27
feat: jsonpath filter for talosctl get outputs

### LINKED ISSUE #6109: talosctl --output jsonpath
## Feature Request

Support jsonpath output.

### Description

As kubectl, I'd like to have -o jsonpath output filter.
It helps me to get resource values from api.

Example:
```shell
talosctl get etcdsecrets.secrets.talos.dev -ojsonpath="{.spec.etcdPeer.crt}" | base64 --decode
```

sauterp:
We add a filter to the `talosctl get` command that allows users to specify a jsonpath filter. Now they can reduce the information that is printed to only the parts they are interested in.

Fixes #6109

### DIFF
--- a/cmd/talosctl/cmd/talos/output/output.go
+++ b/cmd/talosctl/cmd/talos/output/output.go
@@ -7,11 +7,14 @@ package output
 
 import (
 	"fmt"
+	"os"
+	"strings"
 
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/resource/meta"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/spf13/cobra"
+	"k8s.io/client-go/util/jsonpath"
 )
 
 // Writer interface.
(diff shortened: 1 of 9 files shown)


smira (commented):
we definitely need an integration test for `talosctl get` in the CLI section for `jsonpath` output!

cmd/talosctl/cmd/talos/output/output.go
```diff
@@ -23,19 +25,25 @@ type Writer interface {
...
+	writer := os.Stdout
+
+	switch {
+	case format == "table":
+		return NewTable(writer), nil
+	case format == "yaml":
+		return NewYAML(writer), nil
+	case format == "json":
+		return NewJSON(writer), nil
+	case strings.HasPrefix(format, "jsonpath"):
```
smira, line 37:
```suggestion
	case strings.HasPrefix(format, "jsonpath="):
```

prefix should be same as the index expression below, otherwise it might be off by one error

internal/integration/cli/jsonpath.go
```diff
@@ -0,0 +1,52 @@
...
+	)
+}
+
+// TestGetComplexPropertyWithJSONPath verifies that the jsonpath filter to the get command can return JSON.
+func (suite *JSONPathSuite) TestGetComplexPropertyWithJSONPath() {
+	node := suite.RandomDiscoveredNodeInternalIP()
+
+	const jsonMetadataRegex = `\{\s*"created":\s".*",\s*"id":\s".*",\s*\s*"namespace":\s".*",\s*"owner":\s".*",\s*"phase":\s".*",\s*"type":\s".*",\s*"updated":\s".*",\s*"version":\s\d\n\}`
+
+	suite.RunCLI([]string{"get", "--nodes", node, "etcfilestatus", "--output", `jsonpath='{.metadata}'`},
```
smira, line 44: can we have something with multiple results? 

I don't know if it's a valid ask, but we could have used `manifests` resource on control plane nodes? It contains a list of items, so something with `[*]` might work there.

smira (approved):

PR #6343, opened 2022-09-28
release(v1.3.0-alpha.0): prepare release

smira:
This is the official v1.3.0-alpha.0 release.


hack/release.toml
```diff
@@ -6,7 +6,7 @@ github_repo = "siderolabs/talos"
 match_deps = "^github.com/((talos-systems|siderolabs)/[a-zA-Z0-9-]+)$"
 
 # previous release
-previous = "v1.2.2"
+previous = "v1.2.0"
```
DmitriyMV, line 9: Just curious - why is it v1.2.0 and not v1.2.2?

DmitriyMV (approved):

hack/release.toml
smira: Right now the latest is v1.2.3, as we release 1.2.4, it will be 1.2.4, and so on.

Issue #6324, opened 2022-09-24
network route mtu option


sergelogvinov:
## Feature Request

Set MTU size for the route.

### Description

For instance, you have two data-centers and link between (gre/vxvlan and other methods).
Inside datacenter better to use 1500/9000 MTU size, but to reach the nodes from another - we need to use lower MTU size 1400.

This option solves this problem.

```
# machine config
interfaces:
    - interface: eth0 # The interface name.
      addresses:
        - 192.168.2.1/24
      routes:
        - network: 0.0.0.0/0
          gateway: 192.168.2.1
          metric: 1024 
        - network:192.168.1.0/24
          gateway: 192.168.2.1
          mtu: 1400  <-- new value
```

nberlee:
I like this FR for a different use-case: Only allow jumbo frames to my Ceph cluster subnet, use 1500 for the rest. Although Path MTU discovery takes effectively care of this, PMTUD is sub-optimal at best.

smira:
should be fairly easy to implement

PR #6349, opened 2022-09-29
feat: support mtu for routes

frezbo:
Support setting MTU for routes.

Fixes: #6324


internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -203,7 +203,9 @@ func (ctrl *RouteSpecController) syncRoute(ctx context.Context, r controller.Run
 				existing.Protocol == uint8(route.TypedSpec().Protocol) &&
 				existing.Attributes.OutIface == linkIndex && existing.Attributes.Priority == route.TypedSpec().Priority &&
 				(value.IsZero(route.TypedSpec().Source) ||
-					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) {
+					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) &&
```
frezbo, line 206: it seems there are other route parameters where we don't check if the spec matches completely, so updates are just no-op, until a reboot

frezbo:
manually ran the route_spec test `go test -exec sudo -run '^TestRouteSpecSuite$' github.com/talos-systems/talos/internal/app/machined/pkg/controllers/network`

pkg/machinery/resources/network/route_spec.go
```diff
@@ -39,6 +39,7 @@ type RouteSpecSpec struct {
 	Flags       nethelpers.RouteFlags    `yaml:"flags" protobuf:"10"`
 	Protocol    nethelpers.RouteProtocol `yaml:"protocol" protobuf:"11"`
 	ConfigLayer ConfigLayer              `yaml:"layer" protobuf:"12"`
+	MTU         uint32                   `yaml:"mtu" protobuf:"13"`
```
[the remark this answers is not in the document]
sauterp: yes, re-arranging protobuf numbers would break the API.

sauterp (approved):

frezbo:
tested and confirmed by the following steps:

`ip route show` on a standard qemu test environment from a pod running in `hostNetwork: true`:

```bash
bash-5.1# ip route show
default via 10.5.0.1 dev eth0 src 10.5.0.3 metric 1024 
10.5.0.0/24 dev eth0 proto kernel scope link src 10.5.0.3 
10.244.1.0/24 via 10.244.1.0 dev flannel.1 onlink
```

applying the patch below:

```bash
❯ _out/talosctl-linux-amd64 -n 10.5.0.3 patch mc -p '[{"op": "add", "path": "/machine/network/interfaces/0/routes", "value": [{"network": "172.16.0.0/16", "mtu": 1400}, {"network": "172.18.0.0/16"}]}]'
patched MachineConfigs.config.talos.dev/v1alpha1 at the node 10.5.0.3
Applied configuration without a reboot
```

checking the routes after the patch

```bash
bash-5.1# ip route show
default via 10.5.0.1 dev eth0 src 10.5.0.3 metric 1024 
10.5.0.0/24 dev eth0 proto kernel scope link src 10.5.0.3 
10.244.1.0/24 via 10.244.1.0 dev flannel.1 onlink 
172.16.0.0/16 dev eth0 proto static scope link metric 1024 mtu 1400 
172.18.0.0/16 dev eth0 proto static scope link metric 1024
```

smira (commented):
I'm from mobile, so I might have missed itz but looks like we don't have RouteStatus updated for MTU?

internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -254,6 +256,9 @@ func (ctrl *RouteSpecController) syncRoute(ctx context.Context, r controller.Run
 				OutIface: linkIndex,
 				Priority: route.TypedSpec().Priority,
 				Table:    uint32(route.TypedSpec().Table),
+				Metrics: &rtnetlink.RouteMetrics{
```
smira, line 259: should we send 0 MTU or should we skip it completely? I haven't looked into the kernel code to verify, but feels like it's an optional message, so we could omit it completely if MTU == 0 ?

internal/app/machined/pkg/runtime/v1alpha1/platform/equinixmetal/testdata/expected.yaml
```diff
@@ -72,6 +72,7 @@ routes:
       flags: ""
       protocol: static
       layer: platform
+      mtu: 0
```
smira, line 75: we could add `"yaml:"mtu,omitempty"` to avoid this, and as it's rarely being used, that might make sense

smira:
> tested and confirmed by the following steps:

we should implement reading MTU in the `route_status.go` controller, and then we can verify with `talosctl get routes -o yaml`

internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -203,7 +203,9 @@ func (ctrl *RouteSpecController) syncRoute(ctx context.Context, r controller.Run
 				existing.Protocol == uint8(route.TypedSpec().Protocol) &&
 				existing.Attributes.OutIface == linkIndex && existing.Attributes.Priority == route.TypedSpec().Priority &&
 				(value.IsZero(route.TypedSpec().Source) ||
-					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) {
+					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) &&
+				(value.IsZero(route.TypedSpec().MTU) ||
```
DmitriyMV, line 207: `value.IsZero` is good when you have structures which, because of Go parser, you cannot compare using `==`. That is - you cannot do `if myVal == MyStruct{} {` but for some reason you can do `return myval == MyStruct{}` or assign the result.

For simple type like ints, strings and others I think it is better to check using `==`.

internal/app/machined/pkg/controllers/network/route_spec_test.go
```diff
@@ -222,6 +223,7 @@ func (suite *RouteSpecSuite) TestDefaultRoute() {
 					netip.Prefix{}, netip.MustParseAddr("127.0.11.2"), func(route rtnetlink.RouteMessage) error {
 						suite.Assert().Nil(route.Attributes.Dst)
 						suite.Assert().EqualValues(1048576, route.Attributes.Priority)
+						suite.Assert().Equal(uint32(1300), route.Attributes.Metrics.MTU)
```
DmitriyMV, line 226: You can use EqualValues here.

internal/app/machined/pkg/controllers/network/route_status.go
```diff
@@ -122,6 +122,10 @@ func (ctrl *RouteStatusController) Run(ctx context.Context, r controller.Runtime
 				status.Protocol = nethelpers.RouteProtocol(route.Protocol)
 				status.Flags = nethelpers.RouteFlags(route.Flags)
 
+				if route.Attributes.Metrics != nil {
+					status.MTU = route.Attributes.Metrics.MTU
+				}
```
smira, line 127: also need an `else` clause:

```go
} else { 
   status.MTU = 0
}
```
if the MTU is removed (and attributes are nil), we should reset the value to 0

internal/app/machined/pkg/controllers/network/route_spec.go
smira: also it might matter when going from MTU set to some value to the case when MTU should be unset (0).

internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -266,6 +271,7 @@ func (ctrl *RouteSpecController) syncRoute(ctx context.Context, r controller.Run
 			zap.String("gateway", gatewayStr),
 			zap.Stringer("table", route.TypedSpec().Table),
 			zap.String("link", route.TypedSpec().OutLinkName),
+			zap.Uint32("mtu", route.TypedSpec().MTU),
```
smira, line 274: this might be confusing if the MTU is zero, I think we shouldn't log it

internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -203,7 +203,15 @@ func (ctrl *RouteSpecController) syncRoute(ctx context.Context, r controller.Run
...
-					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) {
+					existing.Attributes.Src.Equal(route.TypedSpec().Source.AsSlice())) &&
+				(func(in uint32, existing *rtnetlink.RouteMessage) bool {
+					if existing.Attributes.Metrics != nil {
+						return in == existing.Attributes.Metrics.MTU
+					}
+
+					// zero is default value
+					return in == 0
+				}(route.TypedSpec().MTU, existing)) {
```
smira, line 214: for debugging purposes we might add `old_mtu`/`new_mtu` to the `logger.Debug` at line 217/225

internal/app/machined/pkg/controllers/network/route_spec.go
smira: we could probably simplify this a bit with something like this before `if`:

```go
var existingMTU uint32

if existing.Attributes.Metrics != nil {
  existingMTU = existing.Attributes.Metrics.MTU
}


...

|| existingMTU == route.TypedSpec().MTU
```

smira (approved):

PR #6352, opened 2022-10-02
feat: pin iface ip by mac address in cmdline

sergelogvinov:
Set interface ip address by mac.

```cmdline: ip=172.20.0.2::172.20.0.1:255.255.255.0::enx001122aabbcc```

Iface mac is 001122aabbcc.
Source: https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

# Pull Request

## Acceptance

### CODE REVIEW

**Comment by sergelogvinov** [comment]
+ update documentation

**Review by smira — APPROVED** [review]
I will do a little fixup, but this looks good to me!

File: internal/app/machined/pkg/controllers/network/cmdline.go [hunk 1]
```diff
@@ -163,6 +163,25 @@ func ParseCmdlineNetwork(cmdline *procfs.Cmdline) (CmdlineNetworking, error) {
...
+			for _, iface := range ifaces {
+				ifaceMAC := strings.ReplaceAll(iface.HardwareAddr.String(), ":", "")
+				if ifaceMAC == mac {
+					settings.LinkName = iface.Name
+
+					break
+				}
+			}
+
+			if strings.HasPrefix(settings.LinkName, "enx") {
```
frezbo, line 180: need the if check again? since it's a;ready checking the same here https://github.com/siderolabs/talos/pull/6352/files#diff-9834688c7ece82096be9a1e985a0bff6a1889753c2a27e7829a159728f282bc3R167, 

seems this would always fail :thinking: , is `settings.LinkName` coming from machine config?

sergelogvinov: The interface with MAC cannot be found, so I've checked it again...

smira: This can be rewritten, but the idea is that if we haven't done the substitution, there was no such MAC, so fail

frezbo: my bad, my brain wasn't grokking the code right yesterday :sweat_smile:

frezbo (approved):

PR #6360, opened 2022-10-04
chore: publish ami's with imds v2 enabled

frezbo:
Publish AMI's with IMDS v2 enabled.

Ref: https://aws.amazon.com/about-aws/whats-new/2022/10/amazon-machine-images-support-instance-metadata-service-version-2-default/


smira (commented):
:+1: 

:cool:

smira (approved):

PR #6364, opened 2022-10-04
feat: expose Talos node labels as a machine configuration field

### LINKED ISSUE #6301: expose Talos node labels as a machine configuration field
Node labels is a very powerful concept. At the moment node labels can only be passed as an extra argument to the `kubelet` which accepts node labels as a comma-separated string, and this makes it almost impossible to patch the machine config with multiple labels.

Proposal: expose labels as a field in the machine configuration, make Talos apply labels on boot (as it does for e.g. controlplane nodes).

(this will be still limited to the labels available to the kubelet role)

sauterp:
We add the `nodeLabels` key to the machine config to allow users to add node labels to the kubernetes Node object. The NodeLabels controller reads the nodeLabels from the machine config and applies them via the kubernetes API.
talosctl will throw an unknown keys error if `edit mc` is called on a node with this change.

Fixes #6301


internal/app/machined/pkg/controllers/config/node_labels.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	labelIDsToDelete, labelsToAdd, labelsToChange := ctrl.computeChangesToMake(stateNodeLabels, mcNodeLabels)
+
+	// We are making a compromise here.
+	// Since we don't have transactions in COSI, we can't make sure that both the labels on the kubernetes Node
+	// aswell as in the COSI state are the same in case writing one of them fails.
+	// To make sure labels are always applied to kubernetes we write those first and update the COSI state second.
+	// In case of an error while writing to COSI state, the labels will be picked up on the next reconcile loop,
+	// but will be marked as not created by Talos.
```
utkuozdemir, line 268: Not suggesting to change it but, couldn't we instead use a well-known prefix like `talos.dev/` so all labels managed by Talos would require to be in the form `talos.dev/some-label: value`? It would simplify detecting the owner, we probably wouldn't even need a dedicated resource?

If we don't want the prefix, then I think we could do something simple as, the labels on machineconfig would always win over (overwrite) the other sources.

smira: I think we can't limit our users to use labels of some form, they might need to set labels the way they want it

internal/app/machined/pkg/controllers/config/node_labels.go
```diff
@@ -0,0 +1,334 @@
...
+func (ctrl *NodeLabelsController) Name() string {
+	return "config.NodeLabelsController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *NodeLabelsController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineConfigType,
```
smira, line 47: such controller should never be hanging on the machine config itself, but rather on an interim resource: MachineConfig -> NodeLabelsConfig -> ...

internal/app/machined/pkg/controllers/config/node_labels.go
```diff
@@ -0,0 +1,334 @@
...
+	logger *zap.Logger,
+) error {
+	stateNodeLabels, err := getStateNodeLabels(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	labelIDsToDelete, labelsToAdd, labelsToChange := ctrl.computeChangesToMake(stateNodeLabels, mcNodeLabels)
+
+	// We are making a compromise here.
```
smira, line 263: we don't need to make a compromise here necessarily

the problem of figuring out "ownership" is more or less already solved in Talos e.g. for network addresses

the supposed way is to implement a proper finalizer flow on "desired" label resources - this controller should set its own finalizer asap, and then before letting the label resource to be torn down (when the label is removed from the config), the controller will make sure the label is removed from the Kubernetes node

in other words, ownership is determined by configuration. 

with the node labels removal is not trivial though, as Kubernetes might be down (while Linux kernel can't be for addresses), so controller might not be able to remove the label from the node, while after a reboot Talos will lose track of what labels need to be removed

so we either say we don't support label removal at all, or we say that label removal is not fully guaranteed, it works 99% of the time (if Kubernetes is up)

internal/app/machined/pkg/controllers/config/node_labels.go
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		if err := ctrl.reconcileWithK8s(ctx, r, logger); err != nil {
+			return err
+		}
+	}
+}
+
+func getNodeName(ctx context.Context, r controller.Runtime) (string, error) {
+	nodenameResource, err := r.Get(ctx, resource.NewMetadata(k8s.NamespaceName, k8s.NodenameType, k8s.NodenameID, resource.VersionUndefined))
```
smira, line 129: we can use `safe.` wrappers here and in other places to reduce the boilerplate

internal/app/machined/pkg/controllers/config/node_labels_config.go
```diff
@@ -0,0 +1,78 @@
...
+			Kind:      controller.InputWeak,
+		},
+	}
+}
+
+// Outputs implements controller.Controller interface.
+func (ctrl *NodeLabelsConfigController) Outputs() []controller.Output {
+	return []controller.Output{
+		{
+			Type: config.NodeLabelsConfigType,
```
smira, line 44: it's not consistent in the codebase right now, but looks like most of other `k8s.` configuration resources are in the `k8s/` package

internal/app/machined/pkg/controllers/config/node_labels.go
```diff
@@ -0,0 +1,436 @@
...
+
+	"github.com/talos-systems/talos/pkg/kubernetes"
+	talosconfig "github.com/talos-systems/talos/pkg/machinery/config"
+	"github.com/talos-systems/talos/pkg/machinery/config/types/v1alpha1/machine"
+	"github.com/talos-systems/talos/pkg/machinery/resources/config"
+	"github.com/talos-systems/talos/pkg/machinery/resources/k8s"
+	"github.com/talos-systems/talos/pkg/machinery/resources/secrets"
+)
+
+// NodeLabelsK8sClientIface is the minimal interface needed fo manage the node labels.
```
smira, line 26: can I get the tl;dr on what is the design of this controller?

if the labels owned by Talos are kept as annotations in the Kubernetes `Node` resource, it should be fairly straightforward to sync whatever labels present in the machine config with the `Node` resource?

sauterp: The controller first synchronizes the labels from the machine config and from kubernetes to the COSI state. I wanted to have the labels from other sources also in Talos so the user can see which labels are already "occupied". Here I had to take care that labels which are removed on k8s from another source get removed in the Talos state aswell.
In a second step all the labels from the COSI state are synchronized to k8s if they are owned by Talos. I made sure that no labels that aren't owned by Talos are overwritten or removed. Here I decided to add and remove each label in an individual patch, so that if there is a permission error for one of the labels, all the other labels still get applied.

smira: do we actually need to go that far? 

we could present all node labels, by 'Watch'ing the `Node` resource, but in the end Talos should only remove/update labels which are recorded as managed by Talos in the annotation. 

so there's definitely 'Status' part of that we could do, that is the list of labels present today

there's also 'Spec' part which syncs labels from the (machine) config to the node

I'm also thinking if we should split 'NodeLabel' spec into separate resources? this would allow to inject e.g. controlplane label based on machine type.

sauterp: makes sense, I split them up into NodeLabelStatus and NodeLabelSpec resources. The code should be less convoluted now.

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -104,8 +104,11 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
 			V1Alpha1Mode: ctrl.v1alpha1Runtime.State().Platform().Mode(),
 		},
 		&config.MachineTypeController{},
-		&config.K8sAddressFilterController{},
```
smira, line 107: nit: these should be sorted properly

internal/app/machined/pkg/controllers/k8s/node_label_spec.go
```diff
@@ -0,0 +1,326 @@
...
+	GetK8sClientFunc func(ctx context.Context, r controller.Runtime) (NodeLabelsK8sClientIface, error)
+}
+
+// Name implements controller.Controller interface.
+func (ctrl *NodeLabelSpecController) Name() string {
+	return "k8s.NodeLabelSpecController"
+}
+
+// KubernetesClientDependencies holds the dependencies needed for a controller to create a kubernetes client on control plane or worker nodes.
+var KubernetesClientDependencies = []controller.Input{
```
smira, line 43: I'm still lost on the complexity here... why do we have so many controllers changing things?

let's pick a simple version which will do its job (two controllers):

1. convert machine config to a list of node labels to have (this should be NodeLabelConfig, or NodeLabelSpec to be consistent with naming).
2. controller which watches NodeLabelConfig and on change builds Kubernetes client, gets `Node`, and compares config (spec) with the actual node state. If label is being added, add it to the special annotation which says that it's managed by Talos; if label is being updated, and it's already in the annotation, update it, otherwise error out; if the label is being removed, and it's already in the annotation, remove it and remove from annotation, otherwise error out.

Done. This is pretty good solution with the only downside it doesn't handle the case if someone goes and updates the labels themselves.

I would probably stop here, but if we don't want to stop, then we need another controller:

1. Controller which watches/polls Node and publishes NodeLabelStatus.
2. Feed NodeLabelStatus to the controller which manages labels, this way (without even reading NodeLabelStatus) it will reconcile things on each node label change.

pkg/machinery/resources/k8s/node_label_status.go
```diff
@@ -0,0 +1,81 @@
...
+// NodeLabelStatusType is the type.
+const NodeLabelStatusType = resource.Type("NodeLabelStatuses.k8s.talos.dev")
+
+// NodeLabelStatusSpec represents a label that's attached to a Talos node.
+//
+//gotagsrewrite:gen
+type NodeLabelStatusSpec struct {
+	Key            string `yaml:"key" protobuf:"1"`
+	Value          string `yaml:"value" protobuf:"2"`
+	CreatedByTalos bool   `yaml:"createdByTalos" protobuf:"3"`
```
smira, line 25:
```suggestion
	Managed bool   `yaml:"managed" protobuf:"3"`
```

api/resource/definitions/config/config.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+package talos.resource.definitions.config;
+
+option go_package = "github.com/talos-systems/talos/pkg/machinery/api/resource/definitions/config";
+
+// NodeLabelSpec represents a label that's attached to a Talos node.
+message NodeLabelSpec {
```
smira, line 8: should be cleaned up?

smira (approved):

utkuozdemir (approved):
LGTM

PR #6370, opened 2022-10-06
feat: use readonly containers

andrewrynhard:


internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
```diff
@@ -67,6 +68,9 @@ func CurrentPlatform() (p runtime.Platform, err error) {
 		platform = *p
 	}
 
+	flag.StringVar(&platform, "platform", "", "the platform")
+	flag.Parse()
```
andrewrynhard, line 72: Add the ability to supply a flag to `machined` specifying the platform. It makes using the Talos container feel cleaner.

internal/pkg/mount/overlay.go
```diff
@@ -8,20 +8,20 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var Overlays = []string{
+	"/etc/kubernetes",
+	"/etc/cni",
+	"/usr/libexec/kubernetes",
+	"/usr/etc/udev",
+	"/opt",
+}
```
andrewrynhard, line 17: Export these so we can have them defined in one place.

pkg/provision/providers/docker/node.go
```diff
@@ -84,24 +86,35 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
...
+			Target: path,
+		})
+	}
+
+	for _, path := range append(talosmount.Overlays, "/var", "/system/state") {
+		mounts = append(mounts, mount.Mount{
+			Type:   mount.TypeVolume,
+			Target: path,
+		})
+	}
```
andrewrynhard, line 107: Mimic the `overlay` and other persistent mounts we create in non-container mode.

pkg/provision/providers/docker/node.go
```diff
@@ -84,24 +86,35 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
...
+
 	hostConfig := &container.HostConfig{
 		Privileged:  true,
 		SecurityOpt: []string{"seccomp:unconfined"},
 		Resources: container.Resources{
 			NanoCPUs: nodeReq.NanoCPUs,
 			Memory:   nodeReq.Memory,
 		},
+		ReadonlyRootfs: true,
+		Mounts:         mounts,
```
andrewrynhard, line 117: Use `Mounts` instead of volumes as we have better control over the mount type.

api/resource/definitions/cluster/cluster.proto
```diff
@@ -53,4 +53,3 @@ message MemberSpec {
   talos.resource.definitions.enums.MachineType machine_type = 4;
   string operating_system = 5;
 }
-
```
andrewrynhard, line 56: `make fmt` did this

pkg/machinery/constants/constants.go
```diff
@@ -780,3 +780,12 @@ const (
 	HostnameKey     = "hostname"
 	MacKey          = "mac"
 )
+
+// Overlays is the set of paths to create overlay mounts for.
+var Overlays = []string{
```
andrewrynhard, line 785: I hate that this is here. I had to pull it out of the `mount` package since darwin builds are incompatible with that package. Open to other ideas. I did this temporarily (hopefully) to keep things moving.

Dockerfile
```diff
@@ -120,7 +120,8 @@ RUN --mount=type=cache,target=/.cache go install github.com/planetscale/vtprotob
     && mv /go/bin/protoc-gen-go-vtproto /toolchain/go/bin/protoc-gen-go-vtproto
 RUN --mount=type=cache,target=/.cache go install golang.org/x/vuln/cmd/govulncheck@latest \
     && mv /go/bin/govulncheck /toolchain/go/bin/govulncheck
-RUN curl -sfL https://github.com/uber/prototool/releases/download/v1.10.0/prototool-Linux-x86_64.tar.gz | tar -xz --strip-components=2 -C /toolchain/bin prototool/bin/prototool
+RUN --mount=type=cache,target=/.cache go install github.com/uber/prototool/cmd/prototool@v1.10.0 \
+    && mv /go/bin/prototool /toolchain/go/bin/prototool
```
andrewrynhard, line 124: Makes running native `arm64` builds possible.

pkg/provision/providers/docker/node.go
```diff
@@ -84,24 +85,35 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
...
+			Target: path,
+		})
+	}
+
+	for _, path := range append(constants.Overlays, "/var", "/system/state") {
+		mounts = append(mounts, mount.Mount{
+			Type:   mount.TypeVolume,
+			Target: path,
+		})
+	}
```
andrewrynhard, line 106: Mimic the overlay and other persistent mounts we create in non-container mode.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -842,7 +842,7 @@ func MountOverlayFilesystems(seq runtime.Sequence, data interface{}) (runtime.Ta
 // SetupSharedFilesystems represents the SetupSharedFilesystems task.
 func SetupSharedFilesystems(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
-		targets := []string{"/", "/var/lib/kubelet", "/etc/cni", "/run"}
+		targets := []string{"/", "/var", "/etc/cni", "/run"}
```
smira, line 845: this seems like a big change... not sure if we need that?

pkg/provision/providers/docker/node.go
smira: overlay mounts should technically persist between reboots?

smira: ah, scratch that, they're proper volumes here

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
andrewrynhard: This line specifically or the whole PR?

This line specifically should match the mounts we create. Otherwise this will fail since `/var/lib/kubelet` is no longer a mount.

smira: ok, I think I figured it out, this code actually runs _only_ for container mode, I had an assumption it runs always

smira:
as you are on it, does make sense to drop a simple `docker run` command somewhere in the knowledge base section of our docs (or anywhere else?) on how to run Talos now manually (without `talosctl`), so that people can also translate that into a matching Kubernetes pod spec. As we're changing the mounts, this might affect whoever is running Talos is Kubernetes as a container.

pkg/provision/providers/docker/node.go
```diff
@@ -139,7 +151,7 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
 
 		hostConfig.PortBindings = generatedPortMap.portBindings
 
-		containerConfig.Volumes[constants.EtcdDataPath] = struct{}{}
+		hostConfig.Mounts = append(hostConfig.Mounts, mount.Mount{Type: mount.TypeVolume, Target: constants.EtcdDataPath})
```
andrewrynhard, line 154: This isn't actually need now that `/var` is a volume.

smira (approved):

frezbo (approved):

PR #6388, opened 2022-10-10
feat: use URL-based manifests to present static pods to the kubelet

### LINKED ISSUE #5494: use URL-based manifests to present static pods to the kubelet
See https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/#pods-created-via-http for details.

Controllers can simply start serving on localhost a tiny web server which lists the virtual list of static pods to present.

For the compatibility with previous versions of Talos, we should probably still support static pod path, but anything Talos-managed should be automatically cleaned up from that path on boot and replaced with web-based manifest.

sauterp:
Previously static pod manifests were written to and read from a folder on the disk. We add a controller that cleans up the default static pod manifests on the disk and serves them as a PodList manifest via HTTP. The to the manifest is injected into the kubelet. File based static pod manifests are still supported.

Fixes #5494


pkg/machinery/resources/k8s/static_pod_list.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/talos-systems/talos/pkg/machinery/proto"
+)
+
+// StaticPodListType is type of StaticPodList resource.
+const StaticPodListType = resource.Type("StaticPodLists.kubernetes.talos.dev")
```
smira, line 17: I'm torn on that, but feels like we actually don't need this resource.

As soon as the request goes to the static pod HTTP server, it could list `StaticPods` and build a `v1.PodList` in memory and serve that back to the `kubelet`.

pkg/machinery/resources/k8s/static_pod_list_url.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/talos-systems/talos/pkg/machinery/proto"
+)
+
+// StaticPodListURLType is type of StaticPodListURL resource.
+const StaticPodListURLType = resource.Type("StaticPodListURLs.kubernetes.talos.dev")
```
smira, line 17: Feels like probably not a good name, `StaticPodServerStatus` ? 

if we have a static pod HTTP server, this provides a status of it, i.e. once the resource exists, server is running and the URL/endpoint of it is available here.

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -166,7 +172,7 @@ func (ctrl *KubeletSpecController) Run(ctx context.Context, r controller.Runtime
 			return fmt.Errorf("error merging arguments: %w", err)
 		}
 
-		kubeletConfig, err := NewKubeletConfiguration(cfgSpec)
+		kubeletConfig, err := NewKubeletConfiguration(cfgSpec, getStaticPodURL(ctx, r, logger))
```
smira, line 175: I think we shouldn't create with empty static pod URL, as this means restarting the kubelet. we can pull the HTTP server status before this call and skip reconcile if it's not ready yet

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -243,6 +262,7 @@ func NewKubeletConfiguration(cfgSpec *k8s.KubeletConfigSpec) (*kubeletconfig.Kub
 		Kind:       "KubeletConfiguration",
 	}
 	config.StaticPodPath = constants.ManifestsDirectory
```
smira, line 264: we should remove/disable the static pod path.

we need to think about the upgrade sequence here, probably it should be new machine config value protected by the version contract

smira: in fact, this URL can be put in another controller, which builds `KubeletConfig`

internal/app/machined/pkg/controllers/k8s/static_pod_list.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+}
+
+func (ctrl *StaticPodListController) serveStaticPodSpecs(ctx context.Context, r controller.Runtime, logger *zap.Logger) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			if err := ctrl.startServer(ctx, r, logger); err != nil {
```
smira, line 123: controllers are not allowed to touch `r controller.Runtime` from another goroutine, so this needs to be rewritten not to use goroutines. 

the HTTP server should be started in main goroutine, while `.Serve` might run in additional goroutine.

in general, probably server should start and stop with the controller

internal/app/machined/pkg/controllers/k8s/kubelet_config.go
```diff
@@ -73,36 +91,51 @@ func (ctrl *KubeletConfigController) Run(ctx context.Context, r controller.Runti
...
 
-				kubeletConfig.ClusterDomain = cfgProvider.Cluster().Network().DNSDomain()
-				kubeletConfig.ExtraArgs = cfgProvider.Machine().Kubelet().ExtraArgs()
-				kubeletConfig.ExtraMounts = cfgProvider.Machine().Kubelet().ExtraMounts()
-				kubeletConfig.ExtraConfig = cfgProvider.Machine().Kubelet().ExtraConfig()
-				kubeletConfig.CloudProviderExternal = cfgProvider.Cluster().ExternalCloudProvider().Enabled()
-				kubeletConfig.DefaultRuntimeSeccompEnabled = cfgProvider.Machine().Kubelet().DefaultRuntimeSeccompProfileEnabled()
-				kubeletConfig.SkipNodeRegistration = cfgProvider.Machine().Kubelet().SkipNodeRegistration()
+		if len(kubeletConfig.ClusterDNS) == 0 {
+			var addrs []net.IP
```
smira, line 119: this can be removed now with `:=` below

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -60,6 +60,12 @@ func (ctrl *KubeletSpecController) Inputs() []controller.Input {
 			ID:        pointer.To(k8s.KubeletID),
 			Kind:      controller.InputWeak,
 		},
+		{
```
smira, line 63: this should be removed now, as the value is read from the `kubeletconfig`

internal/app/machined/pkg/controllers/k8s/static_pod_server.go
```diff
@@ -0,0 +1,177 @@
...
+		case err := <-serverError:
+			return fmt.Errorf("http server closed unexpectedly: %w", err)
+		case sendPodListChan := <-buildPodListChan:
+			staticPodList, err := buildPodList(ctx, r)
+			if err != nil {
+				logger.Error("error building static pod list", zap.Error(err))
+			}
+
+			sendPodListChan <- staticPodList
+		case <-r.EventCh():
```
smira, line 84: the logic here seems to be more complicated than it should be.

on reconcile event `<-r.EventCh()` we can rebuild the pod list in `[]byte` and it keep it e.g. in variable with a mutex around it; webserver can grab the mutex, get the slice and write it to the wire. (we don't need to rebuild it in other cases, as it hasn't changed).

channel ping-pong here is fine, but seems to be an overcomplication for what we need

pkg/machinery/config/contract.go
```diff
@@ -147,3 +147,8 @@ func (contract *VersionContract) ApidExtKeyUsageCheckEnabled() bool {
 func (contract *VersionContract) APIServerAuditPolicySupported() bool {
 	return contract.Greater(TalosVersion1_2)
 }
+
+// KubeletEnableManifestsDirectoryFlagSupported returns true if kube-apiserver custom audit policy is supported.
```
smira, line 151: comment is out of sync

pkg/machinery/config/contract.go
```diff
@@ -147,3 +147,8 @@ func (contract *VersionContract) ApidExtKeyUsageCheckEnabled() bool {
...
+func (contract *VersionContract) KubeletEnableManifestsDirectoryFlagSupported() bool {
```
smira, line 152:
```suggestion
func (contract *VersionContract) KubeletManifestsDirectoryDisabled() bool {
```

I thtink this communicates the intention better - we disable the manifests directory.

pkg/machinery/config/types/v1alpha1/generate/init.go
```diff
@@ -73,6 +73,10 @@ func initUd(in *Input) (*v1alpha1.Config, error) {
 		machine.MachineKubelet.KubeletDefaultRuntimeSeccompProfileEnabled = pointer.To(true)
 	}
 
+	if in.VersionContract.KubeletEnableManifestsDirectoryFlagSupported() {
```
smira, line 76: same change should be done to `worker.go`

smira (commented):
also need a note in `hack/release.toml`

mostly nits, lgtm

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1104,6 +1104,15 @@ type KubeletConfig struct {
...
 	KubeletSkipNodeRegistration *bool `yaml:"skipNodeRegistration,omitempty"`
+	//   description: |
+	//     The `enableManifestsDirectory` field configures the kubelet to get static pod manifests from the /etc/kubernetes/manifests directory.
+	//     It's recommended to configure static pods with the "pods" key instead.
+	//   values:
+	//     - true
+	//     - yes
+	//     - false
+	//     - no
+	KubeletEnableManifestsDirectory *bool `yaml:"enableManifestsDirectory,omitempty"`
```
smira, line 1115: for backwards compatibility (when Talos 1.2 is upgraded to Talos 1.3) we need to keep this "on" by default:

* either we call it `DisableManifestsDirectory`
* or we change the `v1alpha1_provider` to make `true` the option when the option is `nil`

internal/app/machined/pkg/controllers/k8s/kubelet_config.go
```diff
@@ -73,36 +91,49 @@ func (ctrl *KubeletConfigController) Run(ctx context.Context, r controller.Runti
...
+			modifyKubeletConfig(cfgProvider, staticPodListURL),
+		); err != nil {
+			return fmt.Errorf("error modifying KubeletConfig resource: %w", err)
+		}
+	}
+}
 
-				kubeletConfig.Image = cfgProvider.Machine().Kubelet().Image()
+func getStaticPodListURL(ctx context.Context, r controller.Runtime) (string, error) {
+	staticPodURLRes, err := r.Get(ctx, resource.NewMetadata(k8s.NamespaceName, k8s.StaticPodServerStatusType, k8s.StaticPodServerStatusResourceID, resource.VersionUndefined))
```
smira, line 102:
```suggestion
	staticPodURLRes, err := safe.ReaderGet[*k8s.StaticPodServerStatus](ctx, r, resource.NewMetadata(k8s.NamespaceName, k8s.StaticPodServerStatusType, k8s.StaticPodServerStatusResourceID, resource.VersionUndefined))
```

use type-safe version to avoid typecasting below

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -242,7 +242,12 @@ func NewKubeletConfiguration(cfgSpec *k8s.KubeletConfigSpec) (*kubeletconfig.Kub
 		APIVersion: kubeletconfig.SchemeGroupVersion.String(),
 		Kind:       "KubeletConfiguration",
 	}
-	config.StaticPodPath = constants.ManifestsDirectory
+
+	if cfgSpec.EnableManifestsDirectory {
+		config.StaticPodPath = constants.ManifestsDirectory
+	}
```
smira, line 248:
```suggestion
	} else {
	    config.StaticPodPath = ""
        }
```

this is required to make sure if manifest directory is disabled in the config, we update the config properly

internal/app/machined/pkg/controllers/k8s/static_pod_server.go
```diff
@@ -0,0 +1,177 @@
...
+		url := fmt.Sprintf("http://%s", listener.Addr().String())
+
+		r.(*k8s.StaticPodServerStatus).TypedSpec().URL = url
+
+		return nil
+	}); err != nil {
+		return nil, nil, fmt.Errorf("error modifying StaticPodListURL resource: %w", err)
+	}
+
+	serverError := make(chan error)
```
smira, line 170: I will probably provide a capacity of 1 here:

```suggestion
	serverError := make(chan error, 1)
```

As the server might be shutdown outside of this function, it might be that the controller is no longer listening on `serverError` channel rendering the goroutine stuck forever

smira (commented):
lgtm, one more round of nits :)

PR #6392, opened 2022-10-10
fix: skip protobuf full unmarshaling for some talosctl commands

smira:
The following commands:

* `talosctl get`
* `talosctl edit mc`, `talosctl patch mc`

Now these commands don't fully unmarshal resources from protobuf representation, but rather use YAML representation.

This allows `talosctl` version to be out of sync with Talos version.

Still other commands do full unmarshaling (e.g. `talosctl upgrade-k8s`), so `talosctl` should match Talos to avoid issues.

### DIFF
--- a/cmd/talosctl/cmd/talos/edit.go
+++ b/cmd/talosctl/cmd/talos/edit.go
@@ -18,6 +18,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/spf13/cobra"
 	"google.golang.org/protobuf/types/known/durationpb"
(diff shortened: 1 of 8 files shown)


frezbo (approved):
does it makes sense to have an integration test with old talosctl just to make sure we don;t marshal/unmarshal  errors in the future?

smira:
> does it makes sense to have an integration test with old talosctl just to make sure we don;t marshal/unmarshal errors in the future?

that is a dream of mine for a long time... not sure how exactly we can pull it through, but we should definitely test backwards compatibility

PR #6405, opened 2022-10-13
fix: kubespan mtu size

sergelogvinov:
Kubespan creates package size more than MTU external interface size.

This PR adds capabilities to change MTU size through machine config. And sets MTU of the default kubespan route.

# Pull Request

## Acceptance


frezbo (approved):

internal/app/machined/pkg/controllers/kubespan/manager.go
```diff
@@ -413,12 +413,18 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
...
+			mtu = constants.KubeSpanLinkMTU
+		}
+
 		for _, spec := range []network.RouteSpecSpec{
 			{
 				Family:      nethelpers.FamilyInet4,
 				Destination: netip.Prefix{},
 				Source:      netip.Addr{},
 				Gateway:     netip.Addr{},
+				MTU:         mtu,
```
smira, line 427: this makes sense, if we route to the KubeSpan link, we set the MTU

internal/app/machined/pkg/controllers/kubespan/manager.go
```diff
@@ -474,6 +481,7 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 				spec.Kind = "wireguard"
 				spec.Up = true
 				spec.Logical = true
+				spec.MTU = mtu
```
smira, line 484: 1420 should be the default for Wireguard link, but we can set it explicitly

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2483,6 +2483,10 @@ type NetworkKubeSpan struct {
 	//   forced to go via KubeSpan (even if Wireguard peer connection is not up), or traffic can go directly
 	//   to the peer if Wireguard connection can't be established.
 	KubeSpanAllowDownPeerBypass *bool `yaml:"allowDownPeerBypass,omitempty"`
+	// description: |
+	//   The kubespan's MTU size.
+	//   Defaul value of Wireguard interface is 1420.
```
smira, line 2488:
```suggestion
	//   Default value is 1420.
```

smira (approved):

internal/app/machined/pkg/controllers/kubespan/manager.go
sergelogvinov: Unfortunately, it always should be set.

Case: default external interface mtu=1500, etcd (or other local daemons) creates packet size=1500, it routes to the wireguard link, which adds extra bytes...
As the result: we send packet > 1500 out of the external interface.

internal/app/machined/pkg/controllers/kubespan/manager.go
sergelogvinov: I think, default wireguard mtu size can be changed.
I want to make sure it will be pre-defend size... because the same size used in route table.

smira: yep, it's correct size, it's `1500 - (size of IPv6 + UDP headers)`

internal/app/machined/pkg/controllers/kubespan/manager.go
smira: yep, I saw the same, and this is a nice fix!

Issue #6223, opened 2022-09-06
etcd: support operations by member ID


smira:
Using `etcd` member hostname (name) as a way to identify members of the cluster is not reliable, as node hostname might change, while etcd label hasn't been updated yet.

Also hostname is missing for members which haven't joined yet.

We should use member ID all the time as this identifier is stable.

> Alternative idea: use "stable hostname" for etcd (i.e. name based on machine-id). 

We need following changes:

* support an API to get etcd member ID for the Talos node (e.g. exposed as a resource): member ID can be derived from the etcd API response or probably by reading directly the etcd database
* in Talos internal operations, e.g. `leave cluster`, we should use member ID, not the hostname
* support (and prefer) member ID over hostnames in CLI operations (see also #4896).

smira:
First iteration:

- make a resource 'EtcdMember` which contains two fields:
   - `MemberID` if known (read from the database)
   - `Joining` flag (true if the node hasn't started etcd yet and it's in a join sequence)

PR #6417, opened 2022-10-17
feat: use the etcd member id for etcd operations instead of hostname

sauterp:
We add a controller that provides the etcd member id as a resource and change the etcd related commands to support member ids next to hostnames.

Fixes: #6223


pkg/machinery/resources/etcd/member.go
```diff
@@ -0,0 +1,105 @@
...
+
+// ResourceDefinition implements typed.ResourceDefinition interface.
+func (MemberRD) ResourceDefinition(resource.Metadata, MemberSpec) meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             MemberType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: NamespaceName,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "MEMBER_ID",
```
smira, line 58:
```suggestion
				Name:     "Member ID",
```

pkg/machinery/resources/etcd/member.go
```diff
@@ -0,0 +1,105 @@
...
+				JSONPath: "{.memberID}",
+			},
+			{
+				Name:     "JOINING",
```
smira, line 62:
```suggestion
				Name:     "Joining",
```

internal/app/machined/pkg/controllers/etcd/member.go
```diff
@@ -0,0 +1,169 @@
...
+			return fmt.Errorf("error getting etcd service resource: %w", err)
+		}
+
+		switch etcdService.Metadata().Phase() {
+		case resource.PhaseRunning:
+			if err := ctrl.modifyLocalEtcdMember(ctx, r, m, etcdService.TypedSpec().Running); err != nil {
+				return err
+			}
+
+			if err := r.AddFinalizer(ctx, etcdService.Metadata(), ctrl.Name()); err != nil {
```
smira, line 85: adding finalizer to the service resource doesn't mean a lot, as service resources just watch service events and reflect the state; in other words, it won't block etcd service from being stopped

pkg/machinery/resources/etcd/member.go
```diff
@@ -0,0 +1,105 @@
...
+	member, err := safe.ReaderGet[*Member](
+		ctx,
+		s,
+		NewMember(NamespaceName, LocalMemberID).Metadata(),
+	)
+	if err != nil {
+		return "", fmt.Errorf("failed to get local etcd member ID: %w", err)
+	}
+
+	return member.TypedSpec().MemberID, nil
```
smira, line 80: this should check for `joining` I guess, otherwise it might return 0 instead

internal/app/machined/pkg/controllers/etcd/member.go
```diff
@@ -0,0 +1,169 @@
...
+		memberIDUint64, err = ctrl.GetLocalMemberID(ctx)
+		if err != nil {
+			return fmt.Errorf("error getting etcd local member ID: %w", err)
+		}
+	}
+
+	memberID := etcd.FormatMemberID(memberIDUint64)
+
+	if err := safe.WriterModify(ctx, r, m, func(status *etcd.Member) error {
+		if memberIDUint64 != 0 {
```
smira, line 136: not sure what 'joining' means fully, but joining members don't have a member ID, so we should set it to zero if it's joining? or drop joining completely (as it doesn't seem to be used)

internal/pkg/etcd/etcd.go
```diff
@@ -148,53 +149,75 @@ func validateMemberHealth(ctx context.Context, memberURIs []string) (err error)
...
 	if err != nil {
-		return fmt.Errorf("failed to remove member %d: %w", *id, err)
+		return fmt.Errorf("failed to remove member %d: %w", id, err)
+	}
+
+	return nil
+}
+
+// RemoveMemberByMemberIDStr removes the member from the etcd cluster.
+func (c *Client) RemoveMemberByMemberIDStr(ctx context.Context, memberID string) error {
```
smira, line 207: I feel like we should have explicit methods for member ID/hostname, it feels much better.

internal/pkg/etcd/etcd.go
```diff
@@ -148,53 +149,75 @@ func validateMemberHealth(ctx context.Context, memberURIs []string) (err error)
...
 
 // LeaveCluster removes the current member from the etcd cluster and nukes etcd data directory.
-func (c *Client) LeaveCluster(ctx context.Context) error {
-	hostname, err := os.Hostname()
-	if err != nil {
-		return err
-	}
-
-	if err = c.RemoveMember(ctx, hostname); err != nil {
+func (c *Client) LeaveCluster(ctx context.Context, memberID string) error {
```
smira, line 152: if this method goes that far as stopping the services, it might find local member ID itself? otherwise someone might call it instead of the 'RemoveMember'

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1754,7 +1755,7 @@ func (s *Server) EtcdRemoveMember(ctx context.Context, in *machine.EtcdRemoveMem
 
 	ctx = clientv3.WithRequireLeader(ctx)
 
-	if err = client.RemoveMember(ctx, in.Member); err != nil {
```
smira, line 1757: I'm almost positive we should have separate (new) API which removes by member ID, and use it in proper way, while keeping old API deprecated.

cmd/talosctl/cmd/talos/etcd.go
```diff
@@ -44,15 +45,29 @@ var etcdLeaveCmd = &cobra.Command{
...
+			// first, try to parse argument as member ID
+			if memberID, err := etcdresource.ParseMemberID(args[0]); err == nil {
+				err = c.EtcdRemoveMemberByID(ctx, &machine.EtcdRemoveMemberByIDRequest{
+					MemberId: memberID,
+				})
+
+				// in the unlikely event that the hostname parses as member ID, try to proceed with remove by hostname
+				if client.StatusCode(err) != codes.NotFound {
+					return err
+				}
```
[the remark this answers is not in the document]
smira: sorry, probably I missed something, but I don't understand your comment.

`client.StatusCode(err)` returns a code always, even for `nil` error (`codes.Unknown`). We want to try `RemoveMemberByHostname` only if `codes.NotFound` is returned explicitly - i.e. the call succeeded, but the member wasn't found by ID.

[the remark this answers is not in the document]
Unix4ever: ah right... If `err` is nil, it will return err as well and exit the function.

smira (approved):

Unix4ever (approved):

PR #6424, opened 2022-10-18
fix: lowercase nodename

smira:
Kubernetes always lowercases whatever nodename is given to the kubelet, so we should do the same, otherwise Talos looks for a `Node` with uppercase letter which is never going to be registered.


smira:
thanks to @steverfrancis for finding this bug

frezbo (approved):

PR #6425, opened 2022-10-18
fix: regenerate cert on node labeling retry

smira:
We have a high timeout for node labeling retry, while the generated temporary Kubernetes PKI has a lifetime of 10 minutes, which means after 10 minutes any call would fail with `Unauthorized`.

Fix that by pulling in the PKI generation into the retry loop, this way cert is always refresh, and it also reacts to machine config changes (e.g. if the endpoint got changed).

Also it helps with the retries if the DNS updates or any other changes like that.

thanks to @steverfrancis for finding this issue


frezbo (approved):
:cool:

PR #6432, opened 2022-10-21
feat: implement Talos API auth using SideroV1 signatures

smira:
This uses https://github.com/siderolabs/go-api-signature as an actual implementation.

PGP keys generated will be stored under `~/.local/share/talos/keys` for Linux.


frezbo (approved):
:cool:

PR #6433, opened 2022-10-21
chore: update kernel with BTF support

smira:
This pulls in:

* https://github.com/siderolabs/pkgs/pull/612
* https://github.com/siderolabs/pkgs/pull/606

### DIFF
--- a/internal/integration/provision/upgrade.go
+++ b/internal/integration/provision/upgrade.go
@@ -394,6 +394,10 @@ func (suite *UpgradeSuite) setupCluster() {
 		)
 	}
 
+	genOptions = append(genOptions, generate.WithSysctls(map[string]string{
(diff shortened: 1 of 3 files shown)


frezbo (approved):

internal/integration/provision/upgrade.go
```diff
@@ -394,6 +394,10 @@ func (suite *UpgradeSuite) setupCluster() {
 		)
 	}
```
smira, line 396: I had to add this (disabling kexec in upgrade tests), as otherwise tests are broken as VMs get stuck in random places. not sure what we can do about it, whether it's QEMU only, or that is a real issue

smira: created #6443 to track this

Unix4ever (approved):

PR #6442, opened 2022-10-24
chore: bump dependencies

Unix4ever:
Via renovate bot.


go.mod
```diff
@@ -100,7 +100,7 @@ require (
 	github.com/siderolabs/siderolink v0.2.0
 	github.com/spf13/cobra v1.6.0
 	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.8.0
```
smira, line 103: there _should_ be more here, as at least AWS SDK should be bumped.

`make go-mod-outdated` lists everything

Unix4ever: should be good now

smira: I don't see any changes

smira (approved):

Issue #6362, opened 2022-10-04
Consider using secretbox instead of aescbc for etcd encryption


Preisschild:
### Description

Talos currently uses the `aescbc` provider for etcd encryption. 

The official documentation does not recommend using this anymore due to it being vulnerable to padding oracle attacks and thus rates it as "Weak". A newer option would be `secretbox`, which is currently rated "Strong" in the Kubernetes documentation

**More Information**
- https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#providers

PR #6448, opened 2022-10-25
feat: support encryption with secretbox

sauterp:
We add support for encryption with secretbox. While AESCBC is still supported secretbox will take precedence if both are configured. Secretbox is not the default encryption for new clusters.

Fixes: #6362

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -111,6 +111,22 @@ machine:
 ```
 """
 
+    [notes.secretbox]
+        title = "Encryption with secretbox"
+        description = """\
+By default new clusters will use secretbox for encryption instead of AESCBC.
+If both are configured secretbox will take precedence.
+Old clusters may keep using AESCBC.
+To enable secretbox you may add an encryption secret at `cluster.secretboxEncryptionSecret`.
+You should keep `aescbcEncryptionSecret` however, even if secretbox is enabled older data will still be encrypted with AESCBC.
+
+How to generate the secret:
+
+```bash
+dd if=/dev/random of=/dev/stdout bs=32 count=1 | base64
+```
+"""
+
 [make_deps]
 
     [make_deps.tools]
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
@@ -882,11 +882,20 @@ type ClusterConfig struct {
 	//       value: '"wlzjyw.bei2zfylhs2by0wd"'
 	BootstrapToken string `yaml:"token,omitempty"`
 	//   description: |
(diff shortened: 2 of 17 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -882,12 +882,21 @@ type ClusterConfig struct {
 	//       value: '"wlzjyw.bei2zfylhs2by0wd"'
 	BootstrapToken string `yaml:"token,omitempty"`
 	//   description: |
-	//     The key used for the [encryption of secret data at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/).
+	//     A key used for the [encryption of secret data at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/).
 	//   examples:
 	//     - name: Decryption secret example (do not use in production!).
+	//     Enables encryption with AESCBC.
 	//       value: '"z01mye6j16bspJYtTB/5SFX8j7Ph4JXxM2Xuu4vsBPM="'
 	ClusterAESCBCEncryptionSecret string `yaml:"aescbcEncryptionSecret"`
```
smira, line 890:
```suggestion
	ClusterAESCBCEncryptionSecret string `yaml:"aescbcEncryptionSecret,omitempty"`
```

we should make both `omitempty`, one for aesthetic reasons, but the new one should be omitempty always, as otherwise the config generate for Talos <= 1.2 will be broken due to unknown YAML key

smira (commented):
lgtm, we should have a note in `hack/release.toml` detailing the following:

* new clusters for Talos 1.3.0 use secretbox by default
* old clusters keep using aescbc
* if one wants to upgrade to secretbox, they should add secretbox secret to the machine config which will activate secretbox for _new_ data, while still allowing to use aescbc to read old data

hack/release.toml
```diff
@@ -109,6 +109,23 @@ machine:
...
+If both are configured secretbox will take precedence.
+Old clusters may keep using AESCBC.
+To enable secretbox you may add an encryption secret at `cluster.secretboxEncryptionSecret`.
+You should keep `aescbcEncryptionSecret` however, even if secretbox is enabled older data will still be encrypted with AESCBC.
+
+Eg:
+
+```yaml
+cluster:
+  secretboxEncryptionSecret: "z01mye6j16bspJYtTB/5SFX8j7Ph4JXxM2Xuu4vsBPM="
```
smira, line 127: let's drop the example or make it invalid, probably better to offer a way to generate the token, like `dd if=/dev/random of=/dev/stdout bs=32 count=1 | base64`

as otherwise people will copy paste straight into their machine configs :(

smira (approved):

PR #6450, opened 2022-10-26
fix: serve static pod files on 127.0.0.1 instead of localhost

sauterp:
Previously we were serving the static pod files on localhost which assumes DNS. We change that to 127.0.0.1.


frezbo (approved):

internal/app/machined/pkg/controllers/k8s/static_pod_server.go
```diff
@@ -136,7 +136,7 @@ func (ctrl *StaticPodServerController) createServer(ctx context.Context, r contr
 		}
 	})
 
-	listener, err := net.Listen("tcp", "localhost:0")
```
smira, line 139: this is good, but I guess the real issue is that it doesn't correctly restart the server if it fails to listen.

PR #6459, opened 2022-10-28
fix: generate correct Flannel config for IPv6-only clusters

### LINKED ISSUE #6427: kube-flannel fails to start in IPv6-only cluster
## Bug Report

kube-flannel fails to start when `cluster.network.podSubnets` contains no IPv4 network (IPv6 only) resulting in node condition `NetworkUnavailable`.

### Description

Talos creates net-conf.json for flannel ([here](https://github.com/siderolabs/talos/blob/18c377a4d1ce046b310e3609033e9c1f39f9337b/internal/app/machined/pkg/controllers/k8s/templates.go#L515)) and sets `EnableIPv6: true` if an IPv6 network is configured. Setting `EnableIPv4` however is never set and defaults to true so that flannel complains about IPv4 being enabled with no IPv4 network set and fails to start. Manually editing the configmap and adding `EnableIPv4: false` fixes the problem.

### Logs

```
kube-flannel W1019 18:31:46.439858       1 client_config.go:614] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
kube-flannel E1019 18:31:47.517699       1 main.go:224] Failed to create SubnetManager: error parsing subnet config: Please define a correct Network parameter in the flannel config
```

### Environment

- Talos version: v1.2.5 65fab60c
- Kubernetes version: v1.25.2
- Platform: tested in qemu amd64, most probably affects all

smira:
See https://github.com/flannel-io/flannel/blob/master/Documentation/configuration.md#ipv6-only

Fixes #6427

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/manifest_test.go
+++ b/internal/app/machined/pkg/controllers/k8s/manifest_test.go
@@ -292,6 +292,28 @@ func (suite *ManifestSuite) TestReconcileIPv6() {
 
 	v, _, _ = unstructured.NestedString(service.Object, "spec", "ipFamilyPolicy") //nolint:errcheck
 	suite.Assert().Equal("SingleStack", v)
+
+	r, err = suite.state.Get(
+		suite.ctx,
+		resource.NewMetadata(
+			k8s.ControlPlaneNamespaceName,
+			k8s.ManifestType,
+			"05-flannel",
+			resource.VersionUndefined,
+		),
+	)
+	suite.Require().NoError(err)
+
+	manifest = r.(*k8s.Manifest) //nolint:errcheck,forcetypeassert
+	suite.Assert().Len(k8sadapter.Manifest(manifest).Objects(), 5)
+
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/k8s/manifest_test.go
```diff
@@ -292,6 +292,28 @@ func (suite *ManifestSuite) TestReconcileIPv6() {
 
 	v, _, _ = unstructured.NestedString(service.Object, "spec", "ipFamilyPolicy") //nolint:errcheck
 	suite.Assert().Equal("SingleStack", v)
+
+	r, err = suite.state.Get(
```
DmitriyMV, line 296: `safe.StateGet`?

smira: copy-paste from the above, so decided to keep same style

DmitriyMV (approved):

frezbo (approved):

PR #6471, opened 2022-11-01
chore: replace talos-systems Go modules with siderolabs

smira:
This the first step towards replacing all import paths to be based on `siderolabs/` instead of `talos-systems/`.

All updates contain no functional changes, just refactorings to adapt to the new path structure.


cmd/talosctl/pkg/mgmt/helpers/wireguard.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	talosnet "github.com/talos-systems/net"
+	talosnet "github.com/siderolabs/net"
```
frezbo, line 13: sideronet :stuck_out_tongue_winking_eye:

frezbo (approved):

PR #6476, opened 2022-11-02
chore: rename talos-systems/talos to siderolabs/talos

smira:
There's a cyclic dependency on siderolink library which imports talos machinery back. We will fix that after we get talos pushed under a new name.


frezbo:
finally :heart:

smira:
nope, not that easy lol, cyclic dependency with siderolink, I will see if I can do something :hankey:

frezbo (approved):

PR #6478, opened 2022-11-03
fix: filter up duplicate IPs out of NodeAddresses

sergelogvinov:
Node can have two IPv6 of the same addresses:
* IPv6/64
* IPv6/128

In this case, the node will advertise two of the same IP:PORT endpoints. Which adds more time to create/recover a p2p (kubespan) connection.

# Pull Request

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/cluster/local_affiliate.go
+++ b/internal/app/machined/pkg/controllers/cluster/local_affiliate.go
@@ -193,7 +193,6 @@ func (ctrl *LocalAffiliateController) Run(ctx context.Context, r controller.Runt
 					spec.OperatingSystem = fmt.Sprintf("%s (%s)", version.Name, version.Tag)
 
 					nodeIPs := addresses.(*network.NodeAddress).TypedSpec().IPs()
-
 					spec.Addresses = make([]netip.Addr, 0, len(nodeIPs))
 
 					for _, ip := range nodeIPs {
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/cluster/local_affiliate.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/go-pointer"
 	"go.uber.org/zap"
+	"gopkg.in/typ.v4/slices"
```
smira, line 18: we try to stick with our own package `github.com/siderolabs/gen/slices`

internal/app/machined/pkg/controllers/cluster/local_affiliate.go
```diff
@@ -202,6 +203,11 @@ func (ctrl *LocalAffiliateController) Run(ctx context.Context, r controller.Runt
 							continue
 						}
 
+						if slices.Contains(spec.Addresses, ip) {
```
smira, line 206: NodeAddresses shouldn't contain duplicates in the first place. if it does, let's better fix `IPs()` method on the `NodeAddress` spec resource (see line 196 above)

internal/app/machined/pkg/controllers/cluster/local_affiliate.go
```diff
@@ -230,7 +236,13 @@ func (ctrl *LocalAffiliateController) Run(ctx context.Context, r controller.Runt
 								continue
 							}
 
-							endpoints = append(endpoints, netip.AddrPortFrom(ip, constants.KubeSpanDefaultPort))
+							endpoint := netip.AddrPortFrom(ip, constants.KubeSpanDefaultPort)
```
smira, line 239: not sure why this makes sense if IPs are already unique

sergelogvinov:
Thank you. Now, much cleaner...

smira (approved):
refactored the code a bit to benefit from uniqueness for all callers of `IPs()` method

Issue #6483, opened 2022-11-07
Ext4 partition not detected when mounted as MachineDisk


d7415:
## Bug Report
In the [docs](https://github.com/siderolabs/talos/blob/release-1.2/website/content/v1.2/reference/configuration.md?plain=1#L247) for MachineDisk (hidden when viewing the [published website](https://www.talos.dev/v1.2/reference/configuration/#machinedisk), but present in the source) it says
> Note that the partitioning and formating is done only once, if and only if no existing partitions are found.

When I mounted a disk with an ext4 partition it was replaced with an empty XFS partition. When I mounted a disk with an existing XFS partition the contents were retained. I'm not sure whether this applies to any non-XFS partition or just ext4.

This was briefly discussed in [Slack](https://taloscommunity.slack.com/archives/CMARMBC4E/p1667775540535899) before posting.

### Environment

- Talos version: v1.2.6
- Kubernetes version: v1.25.3
- Platform: amd64 (KVM)

PR #6484, opened 2022-11-07
docs: specify that only XFS partitions are detected

d7415:
# Pull Request

## What? (description)
Update the docs to specify that only XFS partitions are detected when added in a `MachineDisk`. It is possible that other partition types are accepted, but it appears that ext4 is not.

I also fixed a minor typo in the same sentence.

## Why? (reasoning)
#6483 

This should be documented or fixed. I've taken the easier route and changed the documentation, but feel free to close this if you'd rather handle it another way.

## Acceptance

- Conformance complained about number of commits and not recognising my signature.
- `make fmt` changed a lot of unrelated files (removing newlines at the end of files) so I ignored those.
- Unit tests don't like my set up, but given it's only documentation and the rest passed, hopefully this is ok.


smira (approved):
thank you!

PR #6485, opened 2022-11-09
feat: add timeout to cli action tracking, track by default & refactor

utkuozdemir:
Add a timeout of 15 minutes to the trackable CLI actions reboot, reset, shutdown and upgrade and refactor the action tracking.
Make waiting for these operations the default behavior (set `--wait` to `true` by default).

### DIFF
--- a/cmd/talosctl/cmd/talos/track.go
+++ b/cmd/talosctl/cmd/talos/track.go
@@ -1,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package talos
+
+import (
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+type trackableActionCmdFlags struct {
+	wait    bool
+	debug   bool
+	timeout time.Duration
+}
+
+func (f *trackableActionCmdFlags) addTrackActionFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&f.wait, "wait", true, "wait for the operation to complete, tracking its progress. always set to true when --debug is set")
+	cmd.Flags().BoolVar(&f.debug, "debug", false, "debug operation from kernel logs. --no-wait is set to false when this flag is set")
+	cmd.Flags().DurationVar(&f.timeout, "timeout", 30*time.Minute, "time to wait for the operation is complete if --debug or --wait is set")
+}
--- a/cmd/talosctl/pkg/talos/action/tracker.go
+++ b/cmd/talosctl/pkg/talos/action/tracker.go
@@ -11,10 +11,11 @@ import (
 	"os"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/mattn/go-isatty"
+	"github.com/siderolabs/gen/containers"
 	"github.com/siderolabs/gen/maps"
 	"github.com/siderolabs/go-circular"
 	"golang.org/x/sync/errgroup"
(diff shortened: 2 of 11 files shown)


cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -189,6 +191,38 @@ func (a *Tracker) Run() error {
...
+
+	err = multierror.Append(err, eg.Wait())
+
+	if !a.debug {
+		return err
+	}
+
+	var failedNodes []string
+
+	failedNodesToDmesgs.Range(func(key, value any) bool {
```
DmitriyMV, line 203: Maybe we can replace `sync.Map` with `github.com/siderolabs/gen/containers.ConcurrentMap`?

cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -189,6 +191,38 @@ func (a *Tracker) Run() error {
...
+		failedNodes = append(failedNodes, key.(string))
+
+		return true
+	})
+
+	if len(failedNodes) > 0 {
+		sort.Strings(failedNodes)
+
+		fmt.Printf("console logs for nodes %v:\n", failedNodes)
```
DmitriyMV, line 212:
```suggestion
		fmt.Printf("console logs for nodes %q:\n", failedNodes)
```

Having " is nice here.

cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -189,6 +191,38 @@ func (a *Tracker) Run() error {
...
+
+		for _, node := range failedNodes {
+			dmesgReaderRaw, _ := failedNodesToDmesgs.Load(node)
+			dmesgReader := dmesgReaderRaw.(io.Reader) //nolint:errcheck
+
+			_, copyErr := io.Copy(os.Stdout, dmesgReader)
+			if copyErr != nil {
+				fmt.Printf("%v: failed to print debug logs: %v\n", node, copyErr)
```
DmitriyMV, line 220:
```suggestion
				fmt.Printf("%q: failed to print debug logs: %v\n", node, copyErr)
```

Minor nit about quotes.

utkuozdemir: not sure about this one because I see that for nodes, in more places we use `fmt.Printf("%s: ...", node, ...)` instead of `%q`

cmd/talosctl/pkg/talos/action/tracker.go
utkuozdemir: nice 👍

DmitriyMV (approved):
Several minor nits, but everything else seems good.

cmd/talosctl/cmd/talos/track.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/spf13/cobra"
+)
+
+type trackableActionCmdFlags struct {
+	wait    bool
+	debug   bool
+	timeout time.Duration
+}
+
+func addTrackActionFlags(cmd *cobra.Command, flags *trackableActionCmdFlags) {
```
DmitriyMV, line 19: I think this can be `trackableActionCmdFlags` method

cmd/talosctl/pkg/talos/action/tracker.go
utkuozdemir: done

cmd/talosctl/pkg/talos/action/tracker.go
utkuozdemir: done

cmd/talosctl/pkg/talos/action/tracker.go
DmitriyMV: Ok - both options work for me

cmd/talosctl/cmd/talos/track.go
```diff
@@ -0,0 +1,23 @@
...
+type trackableActionCmdFlags struct {
+	wait    bool
+	debug   bool
+	timeout time.Duration
+}
+
+func (f *trackableActionCmdFlags) addTrackActionFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&f.wait, "wait", true, "wait for the operation to complete, tracking its progress. always set to true when --debug is set")
+	cmd.Flags().BoolVar(&f.debug, "debug", false, "debug operation from kernel logs. --no-wait is set to false when this flag is set")
+	cmd.Flags().DurationVar(&f.timeout, "timeout", 15*time.Minute, "time to wait for the operation is complete if --debug or --wait is set")
```
smira, line 22: with bare metal nodes, it might take 10-15 minutes just to through BIOS boot (if kexec is disabled), so I think we should bump the default timeout to ~30 minutes.

cmd/talosctl/pkg/talos/action/tracker.go
smira: I like `%q` when it comes to empty strings, as it's confusing with `%v`

smira (approved):

cmd/talosctl/cmd/talos/track.go
utkuozdemir: done

Issue #5937, opened 2022-07-18
Kubespan shouldn't report 'external' IPs as node addresses


smira:
When using "external" IPs (cloud-assigned public IPs which are not assigned to the node networking stack), KubeSpan shouldn't report those as node addresses reachable over KubeSpan.

The problem with external IPs is that if the packet for external IP reaches the proper node, it will be dropped, as the node doesn't have that address assigned to the node. Those external IPs are only reachable when going over public Internet.

Issue #4898, opened 2022-01-29
KubeSpan and whitelist/blacklist


sergelogvinov:
## Feature Request

Make possible to set whitelist/blacklist subnets to limit advertisement IPs for the kubespan. Like Talos already has for kubelet/etcd...

Use cases:
1. do not route traffic to vip address through WireGuard.
2. do not use subnets without internet access.
3. do not use IPs from ```dummy``` interfaces.

It increases the network recovery process.

Issue #4369, opened 2021-10-13
KubeSpan and trusted zones


sergelogvinov:
## Feature Request

Add capabilities to set CIDR subnet which disable wireguard encryptions.
Node to node encryption adds latency between pods on different nodes.
If you have huge traffic inside one datacenter and mtu 9000 it could potentially affect the service. 

It will be nice to have this feature...

smira:
1. Filtering subnets which are sent over KubeSpan (basically limit the packet masking process, today KubeSpan targets all node addresses in the cluster except itself).
2. Filtering endpoints which can be used to establish a Wireguard tunnel (today we use all node addresses plus endpoints learned from other connections).

smira:
Machine config example:

```yaml
machine:
  network:
     kubespan:
       enabled: true
       filterSubnets:
          destinations: # traffic to the subnets should go directly bypassing KubeSpan
             - 10.0.0.0/8
             - "!10.2.0.0/16"
           endpoints: # endpoints will be filtered before being attempted
             - 192.168.35.0/24
             - 172.20.0.0/16
```

Filtering is local to each machine, so proper filtering should be set up on each node. Two nodes should have symmetric filters for destinations, otherwise traffic might not flow correctly.

PR #6486, opened 2022-11-10
feat: kubespan node endpoints filter

sergelogvinov:
This feature allows us to use only IPv4 or IPv6 stack to reach the peers. Also, it can help to not share the node-specific IPs, which cannot be accessible at all.

part of #5937, #4898, #4369

I've renamed the key filterSubnets -> filters. Because I believe it will have more options inside... 

# Pull Request

## Acceptance


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2505,6 +2505,15 @@ type NetworkKubeSpan struct {
...
 	KubeSpanMTU *uint32 `yaml:"mtu,omitempty"`
+	// description: |
+	//   KubeSpan filters.
+	KubeSpanFilters *KubeSpanFilters `yaml:"filters,omitempty"`
+}
+
+// KubeSpanFilters struct describes KubeSpan filters.
+type KubeSpanFilters struct {
+	// description: KubeSpanFiltersEndpoints list allowed node' IPs to make p2p connections.
+	KubeSpanFiltersEndpoints *[]string `yaml:"endpoints,omitempty"`
```
smira, line 2516: nit: as it stands out, you don't need a pointer to slice, as slice itself will be `nil` if it's not specified.

sergelogvinov:
I've added tests and example to use.

smira (approved):

PR #6492, opened 2022-11-11
test: fix assertion on reboot test

utkuozdemir:
Fix the reboot test assertion after the changed formatting.


smira (approved):
let's first promote this to `integration-qemu` before merging it.

utkuozdemir:
> let's first promote this to `integration-qemu` before merging it.

Yes, seems I promoted to the wrong one on the previous PR hence didn't catch it: https://github.com/siderolabs/talos/pull/6488#issuecomment-1310904483

Issue #6522, opened 2022-11-19
talosctl get machineconfig && apply failed to parse config: missing kind


vorburger:
## Bug Report

As a naive first time noob user, I would have expected the following to work in an OS which "manages machines [like any other Kubernetes CRD](https://github.com/vorburger/LearningKubernetes-CodeLabs/blob/develop/docs/crd.md)"  - even though I see 😈  from reading https://www.talos.dev/v1.2/talos-guides/configuration/editing-machine-configuration/ this may not be the currently intended/supported "flow", but I thought I would raise it here anyway, just FYI, if you want to consider supporting this in the future (eventually, not blocking; I can probably just use `taloctl edit machineconfig` in the meantime):

```
$ talosctl get machineconfig
NODE           NAMESPACE   TYPE            ID         VERSION
192.168.1.123   config      MachineConfig   v1alpha1   1

$ talosctl get machineconfig v1alpha1 -o yaml >mach.yaml

$ talosctl apply -f mach.yaml --mode=no-reboot
error applying new configuration: 1 error occurred:
        * 192.168.1.98: rpc error: code = InvalidArgument desc = failed to parse config: missing kind

$ talosctl apply machineconfig -f mach.yaml --mode=no-reboot
error applying new configuration: 1 error occurred:
        * 192.168.1.98: rpc error: code = InvalidArgument desc = failed to parse config: missing kind
```

BTW that ID `v1alpha1` is surprising... perhaps that should actually be `192.168.1.123`?

### Environment

- Talos version: [`talosctl version --nodes`]

      Client:
              Tag:         v1.2.5
              SHA:         65fab60c
              Built:
              Go version:  go1.19.2
              OS/Arch:     linux/amd64
      Server:
              NODE:        192.168.1.123
              Tag:         v1.2.5
              SHA:         65fab60c
              Built:
              Go version:  go1.19.2
              OS/Arch:     linux/amd64
              Enabled:     RBAC

- Kubernetes version: [`kubectl version --short`]

      Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
      Client Version: v1.25.3
      Kustomize Version: v4.5.7
      Server Version: v1.25.3

@andrewrynhard FYI

vorburger:
That `missing kind` can, of course, be fixed by inserting this at the start of `mach.yaml`, but perhaps that should automatically be there:

    kind: MachineConfigs.config.talos.dev

That then leads to:

    error applying new configuration: 1 error occurred:
[...]

PR #6523, opened 2022-11-19
docs: Clarify talosctl apply-config & talosctl get machineconfig

vorburger:
@smira @andrewrynhard how about this, as a "docz fix" for #6522 ?


smira (approved):

smira (approved):
thank you!

PR #6539, opened 2022-11-21
feat: hcloud location properties

sergelogvinov:
HCloud starts share the region/availability-zone values.

https://github.com/hetznercloud/csi-driver/pull/293#issuecomment-1320045174

# Pull Request

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/metadata.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/metadata.go
@@ -7,32 +7,21 @@ package hcloud
 import (
 	"context"
 	stderrors "errors"
-	"log"
+	"fmt"
+	"strings"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
 	"github.com/siderolabs/talos/pkg/download"
 )
 
 const (
-	// HCloudExternalIPEndpoint is the local hcloud endpoint for the external IP.
-	HCloudExternalIPEndpoint = "http://169.254.169.254/hetzner/v1/metadata/public-ipv4"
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/metadata.go
```diff
@@ -63,37 +52,44 @@ type NetworkConfig struct {
...
 		return nil, err
 	}
 
-	log.Printf("fetching externalIP from: %q", HCloudExternalIPEndpoint)
+	// Origianl CCM/CSI uses first part of availability-zone to define region name.
+	// But metadata has different value.
+	// We will follow official behavior.
+	metadata.Region = strings.SplitN(metadata.AvailabilityZone, "-", 2)[0]
+
+	// if metadata.Region, err = getMetadataKey("region"); err != nil {
```
smira, line 86: should be removed?

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/hcloud/metadata.go
sergelogvinov: I was confused a little bit with it. 
I almost get used to using the region names like eu-central/us-east. 
But csi does not work well with that code names. I commented on it for the future. Maybe it helps to understand why we do not use the region value from the meta server.

smira: we prefer not to have commented out code... we can leave some comment explaining the intent, but not the code

sergelogvinov: gotcha 👍

PR #6545, opened 2022-11-23
Fix talosctl upgrade force flag parsing

alexandrem:
# Pull Request

## What? (description)

This PR fixes the `talosctl upgrade` command where the `--force` flag is incorrectly parsed and remains always false.

The upgrade force option is not available for users of `talosctl`, although the server implementation supports it.

## Why? (reasoning)

This seems to have been an oversight during some refactoring in the v1.2 release.

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/kubeconfig.go
+++ b/cmd/talosctl/cmd/talos/kubeconfig.go
@@ -22,11 +22,11 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/client"
 )
 
-var (
+var kubeconfigFlags struct {
 	force            bool
 	forceContextName string
 	merge            bool
-)
+}
 
 // kubeconfigCmd represents the kubeconfig command.
 var kubeconfigCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


frezbo:
@smira do we still need to support `--force` :thinking: ?

smira (approved):
thank you, it's definitely using wrong flag name!

smira:
you're right, in general `--force` flag skips some safety checks around `etcd`, but at the same time it's very rarely needed.

I wouldn't recommend upgrading control plane nodes when etcd is unhealthy, and single node clusters should definitely use `--preserve` (using `--force` won't upgrade the node).

at the same time what you said about cluster discovery disabled, I need to double check, it shouldn't affect upgrade process

smira:
@alexandrem  I need your DCO on the commit please, I can fix up other issues myself  (git commit --amend -s)

alexandrem:
@smira Just did.

We can continue the conversation about the cluster discovery behaviour with regard to the upgrade process in a separate issue.

PR #6550, opened 2022-11-24
fix: disable kexec on upgrades from pre-BTF kernel

### LINKED ISSUE #6443: investigate kexec failures with BTF kernel
## Feature Request

See https://github.com/siderolabs/talos/pull/6433/files#r1003463700

smira:
Enabling BTF in the kernel brakes kexec from pre-BTF kernel (e.g. when upgrading from 1.2.x to 1.3.x).

As there's no way to detect Talos version in the installer at the moment, use another way to detect whether BTF is enabled in the Talos version which is running right now.

Fixes #6443

### DIFF
--- a/cmd/installer/pkg/install/errata.go
+++ b/cmd/installer/pkg/install/errata.go
@@ -1,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package install
+
+import (
+	"log"
+	"os"
+
+	pkgkernel "github.com/siderolabs/talos/pkg/kernel"
+	"github.com/siderolabs/talos/pkg/machinery/kernel"
+)
+
+// errataBTF handles the case when kexec from pre-BTF kernel to BTF enabled kernel always fails.
+//
+// This applies to upgrades of Talos < 1.3.0 to Talos >= 1.3.0.
+func errataBTF() {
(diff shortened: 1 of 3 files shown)


internal/integration/provision/upgrade.go
```diff
@@ -398,10 +398,6 @@ func (suite *UpgradeSuite) setupCluster() {
 		)
 	}
 
-	genOptions = append(genOptions, generate.WithSysctls(map[string]string{
```
smira, line 401: removing workaround previously introduced in the tests

cmd/installer/pkg/install/errata.go
```diff
@@ -0,0 +1,33 @@
...
+		return
+	}
+
+	log.Printf("disabling kexec due to upgrade to the BTF enabled kernel")
+
+	if err = pkgkernel.WriteParam(&kernel.Param{
+		Key:   "proc.sys.kernel.kexec_load_disabled",
+		Value: "1",
+	}); err != nil {
+		log.Printf("failed to disable kexec: %s", err)
```
utkuozdemir, line 31: why not return all the way?

smira: I don't want to fail on upgrade e.g. if someone compiles a custom kernel with kexec disabled altogether.

I can't always rely on the presence of this field. Another might be upgrading from such an old version of Talos that it doesn't have kexec. So don't fail the upgrade if possible :)

utkuozdemir: got it, makes sense

utkuozdemir (approved):

PR #6551, opened 2022-11-25
feat: implement pre-flight checks in the installer

### LINKED ISSUE #6149: installer pre-checks
*Research*

Extend `installer` container pre-checks when upgrading Talos to check for:

* Talos upgrade path: is it safe to upgrade from vX.Y to vA.B
* Kubernetes upgrade path: is current version of Kubernetes vX.Y supported with new Talos version?

Running Talos should pass enough information to the installer container (how? stdin? args? mount API socket?) to figure out if it's safe to do an upgrade.

If current Talos doesn't support passing the information, installer should not do the checks.

We need a way to pass structured errors back to Talos (output error as JSON?).

smira:
Host Talos mounts machined socket for API access into the installer contianer (for upgrades).

Installer runs any check it might need to verify compatibility.

At the moment following checks are implemented:

* Talos version (whether upgrade from version X to Y is supported)
* Kubernetes version (whether Kubernetes version X is supported with Talos Y).

Fixes #6149

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -5,6 +5,7 @@
 package install
 
 import (
+	"context"
 	"fmt"
 	"log"
 
@@ -181,6 +182,10 @@ func (i *Installer) probeBootPartition() error {
 func (i *Installer) Install(seq runtime.Sequence) (err error) {
 	errataBTF()
 
+	if err = i.runPreflightChecks(seq); err != nil {
+		return err
+	}
+
 	if err = i.installExtensions(); err != nil {
 		return err
 	}
@@ -337,4 +342,23 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 	}
 
 	return nil
+}
+
+func (i *Installer) runPreflightChecks(seq runtime.Sequence) error {
+	if seq != runtime.SequenceUpgrade {
+		// pre-flight checks only apply to upgrades
+		return nil
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	checks, err := NewPreflightChecks(ctx)
+	if err != nil {
+		return fmt.Errorf("error initializing pre-flight checks: %w", err)
+	}
+
+	defer checks.Close() //nolint:errcheck
+
+	return checks.Run(ctx)
 }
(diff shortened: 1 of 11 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -338,3 +343,22 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
...
 }
+
+func (i *Installer) runPreflightChecks(seq runtime.Sequence) error {
+	if seq != runtime.SequenceUpgrade {
+		// pre-flight checks only apply to upgrades
+		return nil
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
utkuozdemir, line 354: why do we use a cancellable context here?

utkuozdemir (approved):
really cool feature

cmd/installer/pkg/install/install.go
smira: good question, right now there's no real reason, my only thought might be to cancel any streaming calls in progress (if we'd have any in the future)

pkg/machinery/compatibility/talos13/talos13.go
```diff
@@ -0,0 +1,26 @@
...
+var MajorMinor = [2]int{1, 3}
+
+// MinimumHostUpgradeVersion is the minimum version of Talos that can be upgraded to 1.3.
+var MinimumHostUpgradeVersion = version.Must(version.NewVersion("1.0.0"))
+
+// MaximumHostDowngradeVersion is the maximum (not inclusive) version of Talos that can be downgraded to 1.3.
+var MaximumHostDowngradeVersion = version.Must(version.NewVersion("1.5.0"))
+
+// BlacklistedHostUpgradeVersions are the versions of Talos that cannot be upgraded to 1.3.
+var BlacklistedHostUpgradeVersions = []*version.Version{}
```
frezbo, line 20:
```suggestion
var DenylistedHostUpgradeVersions = []*version.Version{}
```

frezbo (approved):

pkg/machinery/compatibility/talos13/talos13.go
frezbo: probably `DeniedHostUpgradeVersions` make sense

smira: yep, this looks better, thanks, I keep forgetting about it!

PR #6561, opened 2022-11-28
feat: use 'registry.k8s.io' for Kubernetes images

smira:
See https://kubernetes.io/blog/2022/11/28/registry-k8s-io-faster-cheaper-ga/


frezbo (commented):
I believe we should also remove the k8s.gcr.io -> registry.k8s.io mirror configuration on the machine config

smira:
> I believe we should also remove the k8s.gcr.io -> registry.k8s.io mirror configuration on the machine config

I thought about that, but that would be a breaking change for those upgrading from previous versions of Talos and still using `k8s.gcr.io`... But on other hand, as this only applies to new clusters :thinking: it might make sense

smira:
@frezbo updated removing registry mirror for 1.3+

frezbo (approved):
I hope there are no references of the `talosctl images` output in the offline docs

PR #6570, opened 2022-11-30
feat: add install script to improve talosctl installation 

steverfrancis:
This install script detects the platform and architecture, and downloads the correct talosctl, and checks the gpg checksums. It also installs and chmods the binary.

Signed-off-by: Steve Francis <steve.francis@talos-systems.com>

# Pull Request

## Acceptance

### DIFF
--- a/website/static/install
+++ b/website/static/install
@@ -1,0 +1,135 @@
+#!/usr/bin/env sh
+
+set -eu
+
+INSTALLPATH=${INSTALLPATH:-/usr/local/bin}
+
+happyexit() {
+  echo "Installed version:"
+  echo "$(talosctl version 2>/dev/null )"
+  echo ""
+  echo "Now run:"
+  echo ""
+  echo "  talosctl cluster create                  # install a local test cluster"
+  echo "  talosctl dashboard                       # If you have created a cluster, launch the dashboard"
+  echo ""
+  echo "Looking for more? Visit https://talos.dev/latest"
+  echo ""
+  exit 0
+}
+
+validate_checksum() {
+  filename=$1
+  url="https://github.com/siderolabs/talos/releases/latest/download/sha256sum.txt"
+  SHA=$(curl --proto '=https' --tlsv1.2 -sSfL "${url}" | grep "${filename}" | awk '{print $1 }')
+  echo ""
+  echo "Validating checksum..."
+
+  case $checksumbin in
+    *openssl)
+      checksum=$($checksumbin dgst -sha256 "${filename}" | sed -e 's/^.* //')
+      ;;
+    *shasum)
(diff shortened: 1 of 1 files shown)


website/static/install
```diff
@@ -0,0 +1,135 @@
+#!/bin/sh
```
andrewrynhard, line 1: `#!/usr/bin/env sh` could offer broader compatibility.

andrewrynhard (approved):
Minor nit.

website/static/install
```diff
@@ -0,0 +1,135 @@
...
+
+set -eu
+
+INSTALLPATH=/usr/local/bin
```
frezbo, line 5: I would make this user settable: 

for eg, many folks might not `~/.bin` or `~/.local/bin`: eg

`curl -SL https://raw.githubusercontent.com/siderolabs/talos/6f99922f681522066bc6aaae40efdd9ec728cc2d/website/static/install | INSTALLPATH=/tmp sh`

```suggestion
INSTALLPATH=${INSTALLPATH:-/usr/local/bin}
```

website/static/install
```diff
@@ -0,0 +1,135 @@
...
+
+happyexit() {
+  echo "Installed version:"
+  echo "$(talosctl version 2>/dev/null )"
```
frezbo, line 9: same here, use the full path:

```suggestion
  echo "$(${INSTALLPATH}/talosctl version 2>/dev/null )"
```

frezbo: that's a good point, for me when I tested the modifed version of the script with `/tmp` it actually picked up my normal talosctl install, if the user specifies a custom path I guess it's upto them to make sure it's in their search path I guess, or maybe we could just check if there's already a talosctl in normal path and warn them of a custom path

steverfrancis: I think that if people are setting INSTALLPATH, they know about search paths, and would want to see the version that is called first... ANd can adjust path if necessary.
So I'd rather say , as you said " if the user specifies a custom path I guess it's upto them to make sure it's in their search path "

website/static/install
```diff
@@ -0,0 +1,135 @@
...
+}
+
+OS=$(uname -s)
+arch=$(uname -m)
+cli_arch=""
+case $OS in
+  CYGWIN* | MINGW64*)
+    OS=windows-amd64.exe
+    ;;
+  Darwin | Linux | FreeBSD)
```
smira, line 52: I don't remember how WSL 2 reports itself, but it'd be nice to have it detected as Linux

website/static/install
steverfrancis, line 1:
```suggestion
#!/usr/bin/env sh```

steverfrancis (commented):
Implemented @andrewrynhard 's suggestion

smira (approved):

PR #6575, opened 2022-12-01
fix: ignore many more filesystems in IMA

### LINKED ISSUE #6553: Bad latency reading large files from mounted volumes
### Description
When I attempt to read large files (multiple GB) in volumes mounted into pods on a Talos cluster, the time to read just one byte out of the file is insanely long, and seems to scale with the size of the file. I've watched transfer metrics, and it looks like the entire file is getting read in at open time (!?!). None of the readahead settings I know of look out of whack (read_ahead_kb setting for the device is 128KB, filesystem readaheads look okay, but probably aren't relevant in light of the info in the next paragraph).

So far, it does not matter what filesystem is mounted, or what CSI the mount is using; I have replicated using RBD, CephFS, geesefs S3, OpenEBS lvm-localpv, and plain old built-in local volumes. hostPath mounts and local volume mounts on the system disk are the only mounts where I haven't been able to replicate this behavior.

I have not been able to replicate this issue when I installed the same version of k3s or k8s via kubeadm in the same environment, which is why I'm thinking this is a Talos issue.

### Logs

Haven't found anything in logs that seems relevant, although when I add an `strace` to the second `dd` in the reproducer below, the long hang occurs during the `openat()` call opening the file from the mounted volume. Let me know if there are any logs you would like me to provide, but hopefully the reproducer is simple enough that you can replicate in your environment.

### Environment

- Talos version: 1.2.7
- Kubernetes version: 1.25.4
- Platform: metal

### Reproducer

1. Spin up a QEMU cluster using `sudo --preserve-env=HOME talosctl cluster create --provisioner qemu --extra-disks 1 --extra-disks-size 20480`.
2. Deploy the PV/PVC/pod from the attached manifests.

[pvc.yaml.txt](https://github.com/siderolabs/talos/files/10096481/pvc.yaml.txt)
[alpine.yaml.txt](https://github.com/siderolabs/talos/files/10096482/alpine.yaml.txt)

3. Run `kubectl exec -it pod/alpine -- /bin/sh`
4. Run `dd if=/dev/zero of=/data/zerofile bs=4M count=4096` to create a 16GB file.
5. Run `time dd if=/data/zerofile of=/dev/null bs=1K count=1` to read the first KB of that file.
6. Observe that time to run second `dd` is very long.

### Sample reproducer output (on my dev laptop with QEMU VMs running on NVME)

```
[evan@nitrogen test]$ kubectl exec -it pod/alpine -- /bin/sh
/ # dd if=/dev/zero of=/data/zerofile bs=4M count=4096
4096+0 records in
4096+0 records out
/ # time dd if=/data/zerofile of=/dev/null bs=1K count=1
1+0 records in
1+0 records out
real	0m 46.57s
user	0m 0.00s
sys	0m 46.50s
/ # 
```

smira:
Fixes #6553

Talos itself defaults to XFS, so IMA measurements weren't done for Talos own filesystems. But many other solutions create by default ext4 filesystems, or it might be something mounted by other means.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -342,6 +342,20 @@ var rules = []string{
 	"dont_measure fsmagic=0xde5e81e4", // EFIVARFS_MAGIC
 	"dont_measure fsmagic=0x58465342", // XFS_MAGIC
 	"dont_measure fsmagic=0x794c7630", // OVERLAYFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x9123683e", // BTRFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x72b6",     // JFFS2_SUPER_MAGIC
+	"dont_measure fsmagic=0x4d44",     // MSDOS_SUPER_MAGIC
+	"dont_measure fsmagic=0x2011bab0", // EXFAT_SUPER_MAGIC
+	"dont_measure fsmagic=0x6969",     // NFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x5346544e", // NTFS_SB_MAGIC
+	"dont_measure fsmagic=0x9660",     // ISOFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x15013346", // UDF_SUPER_MAGIC
+	"dont_measure fsmagic=0x52654973", // REISERFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x137d",     // EXT_SUPER_MAGIC
+	"dont_measure fsmagic=0xef51",     // EXT2_OLD_SUPER_MAGIC
+	"dont_measure fsmagic=0xef53",     // EXT2_SUPER_MAGIC / EXT3_SUPER_MAGIC / EXT4_SUPER_MAGIC
+	"dont_measure fsmagic=0x00c36400", // CEPH_SUPER_MAGIC
+	"dont_measure fsmagic=0x65735543", // FUSE_CTL_SUPER_MAGIC
 	"measure func=MMAP_CHECK mask=MAY_EXEC",
 	"measure func=BPRM_CHECK mask=MAY_EXEC",
 	"measure func=FILE_CHECK mask=^MAY_READ euid=0",


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -338,10 +338,25 @@ var rules = []string{
 	"dont_measure fsmagic=0x43415d53", // SMACK_MAGIC
 	"dont_measure fsmagic=0x27e0eb",   // CGROUP_SUPER_MAGIC
 	"dont_measure fsmagic=0x63677270", // CGROUP2_SUPER_MAGIC
-	"dont_measure fsmagic=0x6e736673", // NSFS_MAGIC
+	"dont_measure fsmagic=0x6e736673", // NFS_MAGIC
 	"dont_measure fsmagic=0xde5e81e4", // EFIVARFS_MAGIC
 	"dont_measure fsmagic=0x58465342", // XFS_MAGIC
 	"dont_measure fsmagic=0x794c7630", // OVERLAYFS_SUPER_MAGIC
+	"dont_measure fsmagic=0x858458f6", // BTRFS_SUPER_MAGIC
```
frezbo, line 345: any links to where these magic numbers are from, the kernel docs doesn't seem to have ext4 ones, at least the one in the comment

smira: https://man7.org/linux/man-pages/man2/statfs.2.html

smira: ha... this one is wrong, I will fix it up, thanks for noticing that https://github.com/torvalds/linux/blob/master/include/uapi/linux/magic.h

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -338,10 +338,25 @@ var rules = []string{
 	"dont_measure fsmagic=0x43415d53", // SMACK_MAGIC
 	"dont_measure fsmagic=0x27e0eb",   // CGROUP_SUPER_MAGIC
 	"dont_measure fsmagic=0x63677270", // CGROUP2_SUPER_MAGIC
-	"dont_measure fsmagic=0x6e736673", // NSFS_MAGIC
+	"dont_measure fsmagic=0x6e736673", // NFS_MAGIC
```
TimJones, line 341: This was correct before, its the namespace filesystem.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: ok, now I'm more confused... I'll dig into that

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: you're right, I have something messed up, fixing the list

frezbo: cool, I thought the kernel docs also had a typo :sweat_smile:

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo: https://github.com/torvalds/linux/blob/v5.15/include/uapi/linux/magic.h we'd probably need this, since our kernel is 5.15

smira: I guess magics shouldn't change, they're derived from the filesystem superblock. should be better now

frezbo (approved):

frezbo:
:cool:

evandeaubl:
#6553 filer here. 😄 I notice in the list of excluded filesystems, FUSE isn't included. Is there a reason that can't be included as well, because a number of CSIs use FUSE under the covers, and I was reproducing the issue with one of those (https://github.com/yandexcloud/k8s-csi-s3).

frezbo:
> #6553 filer here. smile I notice in the list of excluded filesystems, FUSE isn't included. Is there a reason that can't be included as well, because a number of CSIs use FUSE under the covers, and I was reproducing the issue with one of those (https://github.com/yandexcloud/k8s-csi-s3).

good point, just missed, don't see any harm in including FUSE too

smira:
> #6553 filer here. smile I notice in the list of excluded filesystems, FUSE isn't included. Is there a reason that can't be included as well, because a number of CSIs use FUSE under the covers, and I was reproducing the issue with one of those (https://github.com/yandex-cloud/k8s-csi-s3).

thanks, just added that as well

Issue #6562, opened 2022-11-28
`talosctl gen config` to output a single config type


smira:
Add a flag to output a single config type (controlplane/worker/talosconfig) instead of a bundle, specify destination file (we can default to stdout).

smira:
Please note that `gen config` already supports creating reproducible base config from a secrets bundle, and add patches on top of it, so this issue is only about changing the output format.

andrewrynhard:
User wants something like:

```
talosctl gen config --from=base.yaml --config-patch=patch.yaml
```

`from` is an existing base machine configuration.

smira:
I don't like `gen config` with `--from`, how about `talosctl machineconfig patch -f base.yaml -p @patch1.yaml` ?

PR #6576, opened 2022-12-01
feat: add talosctl machineconfig patch command

utkuozdemir:
Add talosctl machineconfig patch command which accepts a machine config as input and a list of patches, applying the patches and writing the result to a file or to stdout.

Link `talosctl machineconfig gen` to `talosctl gen config`, so they work the same way.

Closes siderolabs/talos#6562.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/gen/config.go
+++ b/cmd/talosctl/cmd/mgmt/gen/config.go
@@ -73,28 +73,30 @@ var genConfigCmdFlags struct {
 	withSecrets             string
 }
 
-// genConfigCmd represents the `gen config` command.
-var genConfigCmd = &cobra.Command{
-	Use:   "config <cluster name> <cluster endpoint>",
-	Short: "Generates a set of configuration files for Talos cluster",
-	Long: `The cluster endpoint is the URL for the Kubernetes API. If you decide to use
+// NewConfigCmd builds the config generation subcommand with the given name.
+func NewConfigCmd(name string) *cobra.Command {
+	return &cobra.Command{
+		Use:   fmt.Sprintf("%s <cluster name> <cluster endpoint>", name),
+		Short: "Generates a set of configuration files for Talos cluster",
+		Long: `The cluster endpoint is the URL for the Kubernetes API. If you decide to use
 a control plane node, common in a single node control plane setup, use port 6443 as
 this is the port that the API server binds to on every control plane node. For an HA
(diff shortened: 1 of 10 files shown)


smira (approved):
:cool:

hack/release.toml
```diff
@@ -243,6 +243,26 @@ Talos now uses `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes containe
...
+
+Example:
+
+```bash
+talosctl machineconfig patch controlplane.yaml \
+  --patch '[{"op":"replace","path":"/cluster/clusterName","value":"patch1"}]' \
+  --patch @/path/to/patch2.json
+```
+
+Additionally, `talosctl machineconfig gen` subcommand is introduced as an alias to `talosctl gen config`.
```
utkuozdemir, line 265: @TimJones could you please have a look here regarding the wording?

cmd/talosctl/cmd/mgmt/machineconfig/patch.go
```diff
@@ -0,0 +1,73 @@
...
+
+			return nil
+		}
+
+		// write to file
+
+		parentDir := filepath.Dir(patchCmdFlags.output)
+
+		// Create dir path, ignoring "already exists" messages
+		if err := os.MkdirAll(parentDir, os.ModePerm); err != nil && !os.IsExist(err) {
```
DmitriyMV, line 59: Are we sure that we want to create folders with all `rwxrwxrwx` set?

cmd/talosctl/cmd/mgmt/gen/config.go
```diff
@@ -73,28 +73,31 @@ var genConfigCmdFlags struct {
...
 
-		switch genConfigCmdFlags.configVersion {
-		case "v1alpha1":
-			return writeV1Alpha1Config(args)
-		}
+			//nolint:gocritic
+			switch genConfigCmdFlags.configVersion {
+			case "v1alpha1":
+				return writeV1Alpha1Config(args)
+			}
```
DmitriyMV, line 96: Maybe we should return an error if `configVersion` string was incorrect?

DmitriyMV (approved):

cmd/talosctl/cmd/mgmt/gen/config.go
utkuozdemir: good point, added that

PR #6577, opened 2022-12-02
fix: introduce 'overridePath' setting and fix Talos resolver

smira:
There was inconsistency in the way `/v2` was appended to registry endpoint path between containerd (CRI) and Talos:

* Talos only appended `/v2` to empty paths
* containerd appended `/v2` if it's not the suffix already

Fix Talos to act same as containerd, and introduce a setting `overridePath` which stops both Talos and `containerd` from appending `/v2` (should be required with e.g. Harbor registry mirror).

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -265,6 +265,36 @@ talosctl machineconfig patch controlplane.yaml \
 Additionally, `talosctl machineconfig gen` subcommand is introduced as an alias to `talosctl gen config`.
 """
 
+    [notes.registry-mirrors]
+        title = "Registry Mirrors"
+        description = """\
+Talos had an inconsistency in the way registry mirror endpoints are handled when compared with `containerd` implementation:
+
+```yaml
+machine:
+    registries:
+        mirrors:
+            docker.io:
+                endpoints:
+                    - "https://mirror-registry/v2/mirror.docker.io"
+```
+
+Talos would use endpoint `https://mirror-registry/v2/mirror.docker.io`, while `containerd` would use `https://mirror-registry/v2/mirror.docker.io/v2`.
+This inconsistency is now fixed, and Talos uses same endpoint as `containerd`.
+
+New `overridePath` configuration is introduced to skip appending `/v2` both on Talos and containerd side:
+
+```yaml
+machine:
+    registries:
+        mirrors:
+            docker.io:
+                endpoints:
+                    - "https://mirror-registry/v2/mirror.docker.io"
+                overridePath: true
+```
+"""
+
 [make_deps]
 
     [make_deps.tools]
--- a/internal/pkg/containers/cri/containerd/hosts.go
+++ b/internal/pkg/containers/cri/containerd/hosts.go
@@ -47,24 +47,24 @@ func GenerateHosts(cfg config.Registries, basePath string) (*HostsConfig, error)
 		Directories: map[string]*HostsDirectory{},
 	}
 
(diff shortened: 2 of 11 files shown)


hack/release.toml
```diff
@@ -243,6 +243,36 @@ Talos now uses `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes containe
...
+```yaml
+machine:
+    registries:
+        mirrors:
+            docker.io:
+                endpoints:
+                    - "https://mirror-registry/v2/mirror.docker.io"
+```
+
+Talos will be using endpoint `https://mirror-registry/v2/mirror.docker.io`, while `containerd` will be using `https://mirror-registry/v2/mirror.docker.io/v2`.
```
TimJones, line 262:
```suggestion
Talos would use endpoint `https://mirror-registry/v2/mirror.docker.io`, while `containerd` would use `https://mirror-registry/v2/mirror.docker.io/v2`.
```

hack/release.toml
```diff
@@ -243,6 +243,36 @@ Talos now uses `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes containe
...
+This inconsistency is now fixed, and Talos uses same endpoint as `containerd`.
+
+New configuration was introduced to skip appending `/v2` both on Talos and containerd side:
```
TimJones, line 265:
```suggestion
New `overridePath` configuration was introduced to skip appending `/v2` both on Talos and containerd side:
```

hack/release.toml
```diff
@@ -243,6 +243,36 @@ Talos now uses `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes containe
 See [Kubernetes documentation](https://kubernetes.io/blog/2022/11/28/registry-k8s-io-faster-cheaper-ga/) for additional details.
 
 If using registry mirrors, or in air-gapped installations you may need to update your configuration.
+"""
+
+    [notes.registry-mirrors]
+        title = "Registry Mirrors"
+        description = """\
+Talos had a inconsistency in the way registry mirror endpoints are handled when compared with `containerd` implementation:
```
utkuozdemir, line 251:
```suggestion
Talos had an inconsistency in the way registry mirror endpoints are handled when compared with `containerd` implementation:
```

I think. Right @TimJones ?

TimJones: Yup! I missed that one!

TimJones (approved):

hack/release.toml
```diff
@@ -243,6 +243,36 @@ Talos now uses `registry.k8s.io` instead of `k8s.gcr.io` for Kubernetes containe
...
+        mirrors:
+            docker.io:
+                endpoints:
+                    - "https://mirror-registry/v2/mirror.docker.io"
+```
+
+Talos would use endpoint `https://mirror-registry/v2/mirror.docker.io`, while `containerd` would use `https://mirror-registry/v2/mirror.docker.io/v2`.
+This inconsistency is now fixed, and Talos uses same endpoint as `containerd`.
+
+New `overridePath` configuration was introduced to skip appending `/v2` both on Talos and containerd side:
```
frezbo, line 265: was introduced means it was there previously, right? I am not grokking this :sweat_smile:

DmitriyMV (approved):

hack/release.toml
smira: how do you say that? `is introduced` ?

frezbo (approved):

PR #6594, opened 2022-12-05
chore: bump kernel to 5.15.81

frezbo:
Bump kernel to [5.15.81](https://github.com/siderolabs/pkgs/pull/622)


Makefile
```diff
@@ -14,7 +14,7 @@ NAME = Talos
 
 ARTIFACTS := _out
 TOOLS ?= ghcr.io/siderolabs/tools:v1.3.0
-PKGS ?= v1.3.0
+PKGS ?= v1.3.0-1-g17ea5e6
```
smira, line 17: we should probably cut `v1.4.0-alpha.0` to avoid confusion

frezbo: fixed

smira (approved):

Issue #6444, opened 2022-10-24
upgrade-k8s should fail on incompatible machine config in 1.25


nberlee:
## Feature Request
upgrade-k8s should fail fast if a prerequisite for updating in the machine config is not met. 

upgrade-k8s takes pretty much care of everything, but it does not checks for prerequisites. Although its well documented in upgrade notes `Before update` on the website, upgrade-k8s should easily be able to asses if these prerequisites are met.

For example `disablePodSecurityPolicy: true` is needed for 1.25 and failing to set it results in a failing apiserver start.  With in the log line`"command failed" err="enable-admission-plugins plugin \"PodSecurityPolicy\" is unknown"`.

At that point the api on that server is down, which was pretty much preventable. (the enduser should have also read the documentation)

I am talking there about first party machine config, custom args like explicit `extraArgs` are of course out of scope as it would be unexpected if Talos had any opinion about those.

Issue #6599, opened 2022-12-05
Better `k8s-upgrade` checks for removed api-resources


frezbo:
## Feature Request

Better `k8s-upgrade` checks for removed api-resources

### Description

Make sure talosctl fails for resources that are going to be removed in the `upgradeTo` kubernetes version and is having the removed api version as the storage version in api server. Move these checks to the new `upgrade_checks.go`

smira:
After some discussion, we landed on the fact that `kube-apiserver` automatically handles API version translation, so the only concern is the resource type going out completely.

PR #6600, opened 2022-12-05
feat: add more checks for K8s upgrade

frezbo:
Add more checks for the Talos Kubernetes upgrade.

The removed api-server resources checks are kept as is, needs to be moved to the new checks as part of #6599.

Fixes: #6444

### DIFF
--- a/pkg/cluster/kubernetes/upgrade_checks.go
+++ b/pkg/cluster/kubernetes/upgrade_checks.go
@@ -1,0 +1,304 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package kubernetes
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/slices"
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/siderolabs/talos/pkg/machinery/client"
+	"github.com/siderolabs/talos/pkg/machinery/resources/k8s"
+)
+
+// K8sUpgradeChecks is a set of checks to run before upgrading k8s components.
+type K8sUpgradeChecks struct {
+	state             state.State
+	controlPlaneNodes []string
+	log               func(string, ...interface{})
+
+	upgradePath         string
+	upgradeVersionCheck map[string]componentChecks
+}
+
+// K8sComponentRemovedItemsError is an error type for removed items.
+type K8sComponentRemovedItemsError struct {
+	AdmissionFlags []K8sComponentItem
+	CLIFlags       []K8sComponentItem
+	FeatureGates   []K8sComponentItem
+}
+
+// K8sComponentItem represents a component item.
+type K8sComponentItem struct {
+	Node      string
+	Component string
+	Value     string
+}
+
+type componentChecks struct {
+	// feature gates are common to kube-apiserver, kube-controller-manager and kube-scheduler
+	removedFeatureGates []string
+	// checks specific to kube-apiserver
+	kubeAPIServerChecks apiServerCheck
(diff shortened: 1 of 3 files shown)


pkg/cluster/kubernetes/upgrade_checks.go
```diff
@@ -0,0 +1,354 @@
...
+		for _, node := range checks.controlPlaneNodes {
+			ctx = client.WithNode(ctx, node)
+
+			for _, id := range []string{k8s.APIServerID, k8s.ControllerManagerID, k8s.SchedulerID} {
+				resource, err := checks.talosClient.COSI.Get(ctx, resource.NewMetadata(k8s.NamespaceName, k8s.StaticPodType, id, resource.VersionUndefined))
+				if err != nil {
+					return err
+				}
+
+				staticPod, ok := resource.(*k8s.StaticPod)
```
smira, line 160: we can do `safe.StateGet[*k8s.StaticPod](ctx, checks.talosClient.COSI, ...`

pkg/cluster/kubernetes/upgrade_checks.go
```diff
@@ -0,0 +1,354 @@
...
+	kubeAPIServerChecks apiServerCheck
+	// checks specific to kube-controller-manager
+	kubeControllerManagerChecks k8sComponentCheck
+	// checks specific to kube-scheduler
+	kubeSchedulerChecks k8sComponentCheck
+}
+
+type apiServerCheck struct {
+	// removedAPIResources represent the Kuberenetes API resources that are removed in the upgrade version
+	removedAPIResources []string
```
smira, line 52: I don't see this being used?

And we had some prior code which did these checks?

pkg/cluster/kubernetes/upgrade_checks.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	if hasRemovedFeatureGates {
+		return errors.New(upgradeCheckMessages.String()), nil
+	}
+
+	return nil, nil
+}
+
+func checkK8sAPIServerRemovedAdmissionPlugins(component string, flags []string, removedAdmissionPlugins []string) (k8sUpgradeCheckErr, error) {
```
smira, line 298: I think this function and one above would benefit a lot from the unit-test

or, another option, if we make `upgrade checks` depend on `state.State` without the `talosClient`, we could build a unit-test for the whole thing, where we inject "fake" resources for static pods and verify that it correctly errors out on a problem

pkg/cluster/kubernetes/upgrade_checks.go
```diff
@@ -0,0 +1,349 @@
...
+			fmt.Fprintf(w, "%s\t%s\n", component, removedFeatureGate)
+		}
+	}
+
+	if err := w.Flush(); err != nil {
+		return nil, err
+	}
+
+	if hasRemovedFeatureGates {
+		return errors.New(upgradeCheckMessages.String()), nil
```
smira, line 292: I think this could be done a bit cleaner way. 

We could declare a custom error type, like:

```
type RemoveFeatureGatesError struct {
   Items []struct{
      Component string
       FeatureGate string
   }
}

func (e *RemoveFeatureGatesError) Error() string{
   var buf bytes.Buffer

   w := tabwriter.NewWriter(buf, 0, 0, 3, ' ', 0)
   fmt.Fprintf(w, "\nCOMPONENT\tREMOVED FEATURE GATE\n")
   
   
    ....
    return buf.String()
}
```

pkg/cluster/kubernetes/upgrade_checks_test.go
```diff
@@ -0,0 +1,209 @@
...
+							fmt.Sprintf("/usr/local/bin/%s", id),
+						},
+					},
+				},
+			},
+		}
+
+		suite.Require().NoError(resourceState.Create(ctx, cfg))
+
+		suite.Assert().NoError(
```
smira, line 54: I believe this check is not needed, as resource is immediately available after creation (same for the check on line 175)

smira (approved):

pkg/cluster/kubernetes/upgrade_checks_test.go
frezbo: for some reason I was thinking this was checking the staticpodstatus resource :facepalm:

PR #6603, opened 2022-12-06
fix: fix nil pointer panic and incorrect error output

DmitriyMV:
Currently `.Error()` call is panicking if `watchErr` is nil. Besides - we want to wrap errors the way we can unwrap them.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url.go
@@ -14,6 +14,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/hashicorp/go-multierror"
 
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 	hardwareResource "github.com/siderolabs/talos/pkg/machinery/resources/hardware"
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url.go
```diff
@@ -134,7 +137,9 @@ func getResource[T resource.Resource](ctx context.Context, r state.State, namesp
 
 			eventResource, err := event.Resource()
 			if err != nil {
-				watchErr = fmt.Errorf("%s; invalid resource in wrapped event: %w", watchErr.Error(), err)
+				watchErr = multierror.Append(watchErr, fmt.Errorf("invalid resource in wrapped event: %w", err))
```
smira, line 140: I totally don't understand what's going on in this code (not about your fixes), but more in general why we even have this `watchErr` in the first place

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url.go
```diff
@@ -119,7 +120,9 @@ func getResource[T resource.Resource](ctx context.Context, r state.State, namesp
 		select {
 		case <-watchCtx.Done():
 			err := fmt.Errorf("failed to determine %s of %s: %w", valName, typ, watchCtx.Err())
-			err = fmt.Errorf("%s; %w", err.Error(), watchErr)
+			if watchErr != nil {
+				err = multierror.Append(watchErr, err)
```
smira, line 124: why `watchErr` if we return `err` below?

DmitriyMV: Redone here https://github.com/siderolabs/talos/pull/6605

PR #6604, opened 2022-12-06
feat: add health check for minimal memory / disk size

### LINKED ISSUE #6467: Report when a control plane node does not meet requirements
## Feature Request
Have talosctl health report if minimum requirements are not met.
### Description
Currently, if you fire up a node with 1GB of RAM and try to make it a control plane, the node just gets stuck.
Talosctl reports:
```
waiting for all nodes to finish boot sequence: ...
waiting for all nodes to finish boot sequence: nodes ["45.32.84.84"] are still in boot sequence
```
and just sits there. Logs show
```
 "k8s.KubeletStaticPodController", "error": "error refreshing pod status: error fetching pod status: an error on the server (\"Authorization error (user=apiserver-kubelet-client, verb=get, resource=nodes, subresource=proxy)\") has prevented the request from succeeding"}
```

It would be helpful if talosctl health reported "Control plane nodes require at least 2GB of memory", if they do not have that.
Also would be good if `talosctl apply-config --file controlplane.yaml` did the same thing...

DmitriyMV:
This PR adds two additional checks which are performed during boot sequence and in `talosctl health`. They ensure that nodes have enough memory and disk.

- Boot check will print a warning if memory / disk size is not sufficient.
- Health check will fail if memory / disk size is not sufficient.

Closes #6467

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -26,8 +26,10 @@ import (
 	cgroupsv2 "github.com/containerd/cgroups/v2"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/dustin/go-humanize"
+	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	pprocfs "github.com/prometheus/procfs"
 	"github.com/siderolabs/go-blockdevice/blockdevice"
 	"github.com/siderolabs/go-blockdevice/blockdevice/partition/gpt"
 	"github.com/siderolabs/go-blockdevice/blockdevice/util"
@@ -73,6 +75,7 @@ import (
 	resourcefiles "github.com/siderolabs/talos/pkg/machinery/resources/files"
 	"github.com/siderolabs/talos/pkg/machinery/resources/k8s"
 	resourceruntime "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
+	"github.com/siderolabs/talos/pkg/minimal"
 	"github.com/siderolabs/talos/pkg/version"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/cluster/check/nodes.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+		hostname := msg.Metadata.Hostname
+
+		typ, ok := nodes[hostname]
+		if !ok {
+			return fmt.Errorf("unexpected node %q in response", hostname)
+		}
+
+		if msg.Metadata.Error != "" {
```
smira, line 57: this is not needed, `Memory` function of the client does error unwrapping removing any `Metadata.Error` responses and pushing them to `err error` response

pkg/cluster/check/nodes.go
```diff
@@ -0,0 +1,236 @@
...
+			resultErr = multierror.Append(resultErr, err)
+
+			continue
+		}
+
+		if totalMemory := msg.Meminfo.Memtotal * humanize.KiByte; totalMemory < minimum {
+			resultErr = multierror.Append(
+				resultErr,
+				fmt.Errorf(
+					"node %q has too small memory: %d MiB, minimum is %d MiB",
```
smira, line 77: @rsmitty I think this message should be rephrased, `too little memory` is more correct, but sounds weird?

pkg/cluster/check/nodes.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil, fmt.Errorf("no nodes with mounts found")
+	}
+
+	nodesMnts := map[string][]mntData{}
+
+	for _, msg := range diskResp.Messages {
+		switch {
+		case msg.Metadata == nil:
+			return nil, fmt.Errorf("no metadata in response")
+		case msg.Metadata.Error != "":
```
smira, line 219: same thing with error here

pkg/cluster/check/nodes.go
rsmitty: Maybe "node %q does not meet memory requirements..."?

pkg/minimal/limits.go
```diff
@@ -0,0 +1,38 @@
...
+// Memory returns the minimal/recommended amount of memory required to run the node.
+func Memory(typ machine.Type) (minimum, recommended uint64, err error) {
+	// We remove 100 MiB from the recommended memory to account for the kernel
+	switch typ { //nolint:exhaustive
+	case machine.TypeControlPlane, machine.TypeInit:
+		minimum = 2*humanize.GiByte - 100*humanize.MiByte
+		recommended = 4*humanize.GiByte - 100*humanize.MiByte
+
+	case machine.TypeWorker:
+		minimum = 2*humanize.GiByte - 100*humanize.MiByte
```
smira, line 25: same settings?

DmitriyMV: Copy paste error.

pkg/cluster/check/nodes.go
DmitriyMV: Fixed!

pkg/cluster/check/nodes.go
DmitriyMV: Took @rsmitty suggestion :)

pkg/cluster/check/nodes.go
DmitriyMV: Fixed!

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -26,8 +26,10 @@ import (
 	cgroupsv2 "github.com/containerd/cgroups/v2"
 	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/state"
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/dustin/go-humanize"
+	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/runtime-spec/specs-go"
+	procfs2 "github.com/prometheus/procfs"
```
utkuozdemir, line 32: maybe use a better name like `promprocfs` or something?

DmitriyMV: Now its caled pprocfs 😄

pkg/cluster/check/nodes.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+type mountData struct {
+	Source string
+}
+
+// ErrOldTalosVersion is returned when the node is running an old version of Talos.
+var ErrOldTalosVersion = fmt.Errorf("old Talos version")
+
+func getEphermalPartitionData(ctx context.Context, state state.State, nodeIP string) (mountData, error) {
```
utkuozdemir, line 180:
```suggestion
func getEphemeralPartitionData(ctx context.Context, state state.State, nodeIP string) (mountData, error) {
```

typo

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -681,6 +684,82 @@ func ValidateConfig(seq runtime.Sequence, data interface{}) (runtime.TaskExecuti
...
+		diskSize, err := disk.Size()
+		if err != nil {
+			return fmt.Errorf("failed to get ephemeral disk size: %w", err)
+		}
+
+		if minimum := minimal.DiskSize(); diskSize < minimum {
+			log.Println("WARNING: disk size is less than recommended")
+			log.Println("WARNING: Talos may not work properly")
+			log.Println("WARNING: minimum recommended disk size is", minimum/humanize.MiByte, "MiB")
+			log.Println("WARNING: current total disk size is", diskSize/humanize.MiByte, "MiB")
```
utkuozdemir, line 753: I am overall confused with what we want to do if the checks a. fail b. conditions not met.

Here we just print warnings, but if we fail to read the disk size for example, then we return an error. Shouldn't we be consistent here and either warn or fail on all cases?

Question also applies to the health checks - there we fail explicitly, but here we only warn. What's the expected behavior here?

smira (approved):

PR #6628, opened 2022-12-12
fix: allow empty dnsDomain in machine config

### LINKED ISSUE #6619: `dnsDomain` doesn't have default value as documented
### Description

`dnsDomain` in the cluster network config is documented as having the default value of `cluster.local`; but if I omit it from my talos config then I get the error mesage:

### Logs
```
1 error occurred:
	* "" is not a valid DNS name
```
### Environment

- Talos version: v1.2.7

smira:
This field has a default value, but validation was prohibiting empty value.

Fixes #6619


frezbo (approved):

james-callahan:
I think there's still a bug here.
Leaving `dnsDomain` empty results in a core-dns config with:
```
        kubernetes  in-addr.arpa ip6.arpa {
            pods insecure
            fallthrough in-addr.arpa ip6.arpa
        }
```

i.e. missing `cluster.local`; which results in all in-cluster lookups failing

PR #6647, opened 2022-12-18
fix: kubespan mss clamping

sergelogvinov:
Second step to solve the wireguard mtu issue.

The standard ICMP-method to discover optimal MTU size is not working with wireguard. All icmp packages cannot reach the sender because the OS cannot understand who it was. As a result we have an ICMP black hole.

We can switch on the net.ipv4.tcp_mtu_probing but it will affect all containers. I do not think this is a good idea. We will decrease the MSS of tcp packages which go through the tunnel.

```shell
# nft -n -a list table inet talos_kubespan
...
	chain kubespan_outgoing { # handle 2
		type route hook output priority 0; policy accept;
		meta mark & 0x00000060 == 0x00000020 accept # handle 15
		oifname "lo" accept # handle 16
		ip daddr @kubespan_targets_ipv4 tcp flags & (0x2 | 0x4) == 0x2 tcp option maxseg size > 1260 tcp option maxseg size set 1260 # handle 19
		ip daddr @kubespan_targets_ipv4 meta mark set meta mark & 0xffffffdf | 0x00000040 accept # handle 20
		ip6 daddr @kubespan_targets_ipv6 tcp flags & (0x2 | 0x4) == 0x2 tcp option maxseg size > 1240 tcp option maxseg size set 1240 # handle 21
		ip6 daddr @kubespan_targets_ipv6 meta mark set meta mark & 0xffffffdf | 0x00000040 accept # handle 22
	}
```
Reference:
* https://blog.cloudflare.com/path-mtu-discovery-in-practice/
* https://www.ietf.org/rfc/rfc4821.txt

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/kubespan/nftables.go
+++ b/internal/app/machined/pkg/controllers/kubespan/nftables.go
@@ -11,11 +11,12 @@ import (
 	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"go4.org/netipx"
+	"golang.org/x/sys/unix"
 )
 
 // NfTablesManager manages nftables outside of controllers/resources scope.
 type NfTablesManager interface {
-	Update(*netipx.IPSet) error
+	Update(ips *netipx.IPSet, mtu uint32) error
 	Cleanup() error
 }
 
@@ -54,6 +55,7 @@ type nfTablesManager struct {
 	MarkMask     uint32
 
 	currentSet *netipx.IPSet
+	currentMTU uint32
 
 	// nfTable is a handle for the KubeSpan root table
 	nfTable *nftables.Table
@@ -66,16 +68,17 @@ type nfTablesManager struct {
 }
 
 // Update the nftables rules based on the IPSet.
-func (m *nfTablesManager) Update(desired *netipx.IPSet) error {
-	if m.currentSet != nil && m.currentSet.Equal(desired) {
+func (m *nfTablesManager) Update(desired *netipx.IPSet, mtu uint32) error {
+	if m.currentSet != nil && m.currentSet.Equal(desired) && m.currentMTU == mtu {
 		return nil
 	}
 
-	if err := m.setNFTable(desired); err != nil {
+	if err := m.setNFTable(desired, mtu); err != nil {
 		return fmt.Errorf("failed to update IP sets: %w", err)
 	}
 
 	m.currentSet = desired
+	m.currentMTU = mtu
 
 	return nil
 }
@@ -129,7 +132,7 @@ func (m *nfTablesManager) tableExists() (bool, error) {
 	return foundExisting, nil
 }
 
-func (m *nfTablesManager) setNFTable(ips *netipx.IPSet) error {
+func (m *nfTablesManager) setNFTable(ips *netipx.IPSet, mtu uint32) error {
 	c := &nftables.Conn{}
 
 	// NB: sets should be flushed before new members because nftables will fail
@@ -173,6 +176,22 @@ func (m *nfTablesManager) setNFTable(ips *netipx.IPSet) error {
 
 	if err := c.AddSet(m.targetSet6, setElements6); err != nil {
 		return fmt.Errorf("failed to add IPv6 set: %w", err)
+	}
+
+	// meta ifname "lo" accept
+	ruleLo := []expr.Any{
+		// [ meta load oifname => reg 1 ]
+		&expr.Meta{Key: expr.MetaKeyOIFNAME, Register: 1},
+		// [ cmp eq reg 1 lo ]
+		&expr.Cmp{
+			Op:       expr.CmpOpEq,
+			Register: 1,
+			Data:     ifname("lo"),
+		},
+		// Accept the packet to stop the ruleset processing
+		&expr.Verdict{
+			Kind: expr.VerdictAccept,
+		},
 	}
 
 	// meta mark & 0x00000060 == 0x00000020 accept
@@ -219,6 +238,14 @@ func (m *nfTablesManager) setNFTable(ips *netipx.IPSet) error {
 		Exprs: ruleExpr,
 	})
 
+	// match lo interface
+	// accept and return without modifying the table or mark
+	c.AddRule(&nftables.Rule{
+		Table: m.nfTable,
+		Chain: outChain,
+		Exprs: ruleLo,
+	})
+
 	c.AddRule(&nftables.Rule{
 		Table: m.nfTable,
 		Chain: preChain,
@@ -234,7 +261,19 @@ func (m *nfTablesManager) setNFTable(ips *netipx.IPSet) error {
 	c.AddRule(&nftables.Rule{
 		Table: m.nfTable,
 		Chain: outChain,
+		Exprs: matchIPSetMSS(m.targetSet4, mtu, nftables.TableFamilyIPv4),
+	})
+
+	c.AddRule(&nftables.Rule{
+		Table: m.nfTable,
+		Chain: outChain,
 		Exprs: matchIPv4Set(m.targetSet4, m.InternalMark, m.MarkMask),
+	})
+
+	c.AddRule(&nftables.Rule{
+		Table: m.nfTable,
+		Chain: outChain,
+		Exprs: matchIPSetMSS(m.targetSet6, mtu, nftables.TableFamilyIPv6),
 	})
 
 	c.AddRule(&nftables.Rule{
@@ -258,16 +297,23 @@ func matchIPv6Set(set *nftables.Set, mark, mask uint32) []expr.Any {
 	return matchIPSet(set, mark, mask, nftables.TableFamilyIPv6)
 }
 
-func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamily) []expr.Any {
-	var (
-		offset uint32 = 16
-		length uint32 = 4
-	)
-
-	if family == nftables.TableFamilyIPv6 {
+func ipOffsetLength(family nftables.TableFamily) (offset uint32, length uint32) {
+	switch family { //nolint:exhaustive
+	case nftables.TableFamilyIPv4:
+		offset = 16
+		length = 4
+	case nftables.TableFamilyIPv6:
 		offset = 24
 		length = 16
+	default:
+		panic("unexpected IP family")
 	}
+
+	return offset, length
+}
+
+func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamily) []expr.Any {
+	offset, length := ipOffsetLength(family)
 
 	// ip daddr @kubespan_targets_ipv4 meta mark set meta mark & 0xffffffdf | 0x00000040 accept
 	return []expr.Any{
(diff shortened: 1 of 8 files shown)


smira:
this looks cool, couple of quick questions:

1. why match on `lo`? the traffic routed to `lo` shouldn't even hit these rules anyways?
2. how to build a test for it?

sergelogvinov:
1. I skip traffic to LO interface for optimization - local communication we do not need to manage/change it.
2. I have no idea how to test it. May be launch e2e test with kubespan. Worker nodes should flapping without this fix (kubelet regularly lost the connection).

Can we back-port it to the 1.3.x version? Currently the node-to-node communication is not working well...

smira:
> Can we back-port it to the 1.3.x version? Currently the node-to-node communication is not working well...

If we have a proper test, I'm more happy about it. I guess we'd need to lower node MTU and run some iperf across CNI from node to node?

sergelogvinov:
CF article recommends use this:
```shell
iptables -I INPUT -s <src_ip> -m length --length 1400:1500 -j DROP
```
to block big packages, but i am not sure it will work in one machine (soft bridge).  
GCP has less 1500 MTU size between AV-zones too.

sergelogvinov:
A few words:
GCP does MSS clamping too https://cloud.google.com/vpc/docs/mtu#vpn-mtu
also very common stings to google:
* `iptables ... -j TCPMSS --set-mss`
* `iptables ... -j TCPMSS --clamp-mss-to-pmtu`

https://tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.cookbook.mtu-mss.html
May be we need to change rule to `tcp tcp flags & (syn|rst) == syn`

sergelogvinov:
> > Can we back-port it to the 1.3.x version? Currently the node-to-node communication is not working well...
> 
> If we have a proper test, I'm more happy about it. I guess we'd need to lower node MTU and run some iperf across CNI from node to node?

I think we had miscommunication here. I thought we spoke about how make test in Build/CICD machine.
I will show how i did the manual tests:

Wireguard [overhead](https://lists.zx2c4.com/pipermail/wireguard/2017-December/002201.html) (MTU): 
* 20+8+4+4+8+16 = 60 (IPv4)
* 40+8+4+4+8+16 = 80 (IPv6)

TCP overhead (MSS):
* 1500(MTU) - 40(IP+TCP) = 1460 (IPv4)
* 1500(MTU) - 60(IP+TCP) = 1440 (IPv6)


```yaml
    kubespan:
      enabled: true
      mtu: 1300
```

After fix
```shell
# tcpdump -n -i kubespan  "tcp[tcpflags] & (tcp-syn) != 0"
07:10:11.757945 IP A.52700 > B.2380: Flags [S], seq 3439863491, win 64240, options [mss 1260,sackOK,TS val 2986858467 ecr 0,nop,wscale 7], length 0
07:10:11.757976 IP B.2380 > A.52700: Flags [S.], seq 1402417112, ack 3439863492, win 65160, options [mss 1260,sackOK,TS val 250299695 ecr 2986858467,nop,wscale 7], length 0
07:10:11.930731 IP B.37378 > C.2380: Flags [S], seq 187058504, win 64240, options [mss 1260,sackOK,TS val 463708175 ecr 0,nop,wscale 7], length 0
07:10:11.930731 IP B.53046 > A.2380: Flags [S], seq 3977751670, win 64240, options [mss 1260,sackOK,TS val 250299868 ecr 0,nop,wscale 7], length 0
07:10:11.938398 IP C.2380 > B.37378: Flags [S.], seq 3452280647, ack 187058505, win 65160, options [mss 1260,sackOK,TS val 386238012 ecr 463708175,nop,wscale 7], length 0
07:10:11.939196 IP A.2380 > B.53046: Flags [S.], seq 3584111533, ack 3977751671, win 65160, options [mss 1260,sackOK,TS val 2986858648 ecr 250299868,nop,wscale 7], length 0

# tcpdump -n -i lo "tcp[tcpflags] & (tcp-syn) != 0"
07:10:38.166255 IP 127.0.0.1.36912 > 127.0.0.1.50001: Flags [S], seq 132867157, win 65495, options [mss 65495,sackOK,TS val 3011546279 ecr 0,nop,wscale 7], length 0
07:10:38.166271 IP 127.0.0.1.50001 > 127.0.0.1.36912: Flags [S.], seq 2324732834, ack 132867158, win 65483, options [mss 65495,sackOK,TS val 3011546280 ecr 3011546279,nop,wscale 7], length 0
07:10:38.991148 IP 127.0.0.1.38398 > 127.0.0.1.8091: Flags [S], seq 1323883606, win 65495, options [mss 65495,sackOK,TS val 3011547104 ecr 0,nop,wscale 7], length 0
07:10:38.991174 IP 127.0.0.1.8091 > 127.0.0.1.38398: Flags [S.], seq 77588273, ack 1323883607, win 65483, options [mss 65495,sackOK,TS val 3011547104 ecr 3011547104,nop,wscale 7], length 0
07:10:39.723864 IP 127.0.0.1.37162 > 127.0.0.1.50000: Flags [S], seq 2212319204, win 65495, options [mss 65495,sackOK,TS val 3011547837 ecr 0,nop,wscale 7], length 0
07:10:39.723889 IP 127.0.0.1.50000 > 127.0.0.1.37162: Flags [S.], seq 3221349603, ack 2212319205, win 65483, options [mss 65495,sackOK,TS val 3011547837 ecr 3011547837,nop,wscale 7], length 0
07:10:39.991068 IP 127.0.0.1.38400 > 127.0.0.1.8091: Flags [S], seq 1604474215, win 65495, options [mss 65495,sackOK,TS val 3011548104 ecr 0,nop,wscale 7], length 0

# tcpdump -n -i eth0 "ip[2:2] > 1200"
07:38:41.161954 IP A.51820 > B.51820: UDP, length 1332
07:38:41.162296 IP A.51820 > B.51820: UDP, length 1332
07:38:41.162298 IP A.51820 > B.51820: UDP, length 1332
07:38:41.162299 IP A.51820 > B.51820: UDP, length 1332
07:38:41.369850 IP B.51820 > A.51820: UDP, length 1332
07:38:41.369861 IP B.51820 > A.51820: UDP, length 1332
```

External traffic (A,B,C) uses MSS = 1260, local communications - MSS = 65495
Encrypted traffic = 1332 - 32byte extra kubespan MTU (4-byte type, 4-byte key index, 8-byte nonce, 16-byte authentication tag)

# nftables

to create nft-rules, i used:
```shell
nft --debug=netlink add rule meh tcp_flags 'tcp flags & (syn|rst) == syn tcp option maxseg size > 1460 tcp option maxseg size set 1460'
```

smira:
I think I'm still missing something. I create a QEMU cluster with MTU of 1000 bytes.

I set the KubeSpan MTU to 920.

I ran `iperf` in TCP mode both over host network and pod network (with Flannel CNI), and I see no issues with MSS whatsoever. I wonder if there's something else here?

```
bash-5.2# iperf -c 10.244.3.2 -m -e -d
------------------------------------------------------------
Server listening on TCP port 5001 with pid 10882
Read buffer size:  128 KByte (Dist bin width=16.0 KByte)
MSS size 536 bytes
TCP window size:  128 KByte (default)
------------------------------------------------------------
------------------------------------------------------------
Client connecting to 10.244.3.2, TCP port 5001 with pid 10882 (1 flows)
Write buffer size: 131072 Byte
MSS size 536 bytes
TOS set to 0x0 (Nagle on)
TCP window size: 16.0 KByte (default)
------------------------------------------------------------
[  1] local 10.244.1.2%eth0 port 38884 connected with 10.244.3.2 port 5001 (sock=4) on 2023-01-13 18:25:05 (UTC)
[  2] local 10.244.1.2%eth0 port 5001 connected with 10.244.3.2 port 51454 (sock=5) on 2023-01-13 18:25:05 (UTC)
[ ID] Interval        Transfer    Bandwidth       Reads=Dist
[  2] 0.00-10.03 sec   560 MBytes   468 Mbits/sec  35754=18148:17045:169:127:59:113:26:67
[ ID] Interval        Transfer    Bandwidth       Write/Err
[  1] 0.00-10.04 sec   468 MBytes   391 Mbits/sec  3746/0
```

smira:
what I see though is that packets when using CNI are fragmented, as they get encapsulated by Flannel once again, and this is too many layers of encapsulation:

```
PACKET: 104 bytes, wire length 104 cap length 104 @ 2023-01-13 18:29:03.500305265 +0000 UTC
- Layer 1 (20 bytes) = IPv4	{Contents=[..20..] Payload=[..84..] Version=4 IHL=5 TOS=0 Length=104 Id=1174 Flags= FragOffset=112 TTL=64 Protocol=UDP Checksum=7500 SrcIP=172.20.0.6 DstIP=172.20.0.5 Options=[] Padding=[]}
- Layer 2 (84 bytes) = Fragment	84 byte(s)

PACKET: 916 bytes, wire length 916 cap length 916 @ 2023-01-13 18:29:03.500305295 +0000 UTC
- Layer 1 (20 bytes) = IPv4	{Contents=[..20..] Payload=[..896..] Version=4 IHL=5 TOS=0 Length=916 Id=1175 Flags=MF FragOffset=0 TTL=64 Protocol=UDP Checksum=64142 SrcIP=172.20.0.6 DstIP=172.20.0.5 Options=[] Padding=[]}
- Layer 2 (896 bytes) = Fragment	896 byte(s)
```

smira:
Even if I force MSS to some (high) value, it still works without issues:

```
bash-5.2# iperf -c 10.244.3.2 -m -e -d -M 1500
------------------------------------------------------------
Server listening on TCP port 5001 with pid 14168
Read buffer size:  128 KByte (Dist bin width=16.0 KByte)
MSS req size 1500 bytes (per TCP_MAXSEG)
TCP window size:  128 KByte (default)
------------------------------------------------------------
------------------------------------------------------------
Client connecting to 10.244.3.2, TCP port 5001 with pid 14168 (1 flows)
Write buffer size: 131072 Byte
MSS req size 1500 bytes (per TCP_MAXSEG)
TOS set to 0x0 (Nagle on)
TCP window size: 16.0 KByte (default)
------------------------------------------------------------
[  1] local 10.244.1.2%eth0 port 35576 connected with 10.244.3.2 port 5001 (sock=4) on 2023-01-13 18:39:48 (UTC)
[  2] local 10.244.1.2%eth0 port 5001 connected with 10.244.3.2 port 56658 (sock=5) on 2023-01-13 18:39:48 (UTC)
[ ID] Interval        Transfer    Bandwidth       Reads=Dist
[  2] 0.00-10.03 sec   579 MBytes   484 Mbits/sec  39373=21651:17150:153:113:72:136:20:78
[ ID] Interval        Transfer    Bandwidth       Write/Err
[  1] 0.00-10.04 sec   495 MBytes   414 Mbits/sec  3961/0
```

So I feel there's something else here into the play

smira:
wow, with Linux 6.1 we can't have MTU < 1280 for Wireguard, as KubeSpan is IPv6 https://lore.kernel.org/wireguard/20190321033638.1ff82682@natsu/t/

smira:
I don't see any difference with `iperf` with or without the patch in this PR (with Flannel). @sergelogvinov if you are using a different CNI or a different scenario, this might help me to reproduce.

sergelogvinov:
It is very hard to catch this issue on local machine.
kvm/qemu virtualization uses paravirtualization network. And it does some tricks. Guest VM sends many small tcp packages and network interface merge it in one. On host machine you can notice VM packages 4-8k bytes, even if bridge has mtu=1500. 

The main problem here is node-to-node communications. Kubelet creates the socket (by default mss calculates based on main mtu interface - 40 = 1500-40). But kubespan route it to the wireguard, and it adds more bytes. And at the end the package will have >1500 size and drops between server/datacenters. 
Pod-to-pod communications works fine. Container mtu size less than main interface useably or you can customize it through  CNI plugin.

https://cloud.google.com/vpc/docs/mtu#handling_of_packets_that_exceed_mtu 
GCP does it for you automatically.
```
For some network paths, Google Cloud performs MSS clamping, lowering a sender's advertised MSS value in a SYN packet or a recipient's advertised MSS value in a SYN-ACK packet.
```

smira:
> The main problem here is node-to-node communications. Kubelet creates the socket (by default mss calculates based on main mtu interface - 40 = 1500-40). But kubespan route it to the wireguard, and it adds more bytes. And at the end the package will have >1500 size and drops between server/datacenters. Pod-to-pod communications works fine. Container mtu size less than main interface useably or you can customize it through CNI plugin.

Interesting, does this mean that control plane endpoint itself is over KubeSpan as well? Is it using Talos VIP feature?

sergelogvinov:
Currently, I have a christmas tree, all kubelet periodically lost the kube-api and when recovery by themself. 
And it affects not only kubelet, it affects all hostnode-containers...

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -324,6 +363,130 @@ func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamil
...
+		length uint32 = 4
+		ifmtu  uint16
+		mss    uint16
+	)
+
+	switch {
+	case mtu == 0:
+		ifmtu = 1500
+	case mtu < 576: // Minimum IPv4 mtu size
+		ifmtu = 576 - 40
```
sergelogvinov, line 378: bug here..
```
ifmtu = 576
```

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -324,6 +363,130 @@ func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamil
...
+func matchIPSetMSS(set *nftables.Set, mtu uint32, family nftables.TableFamily) []expr.Any {
+	var (
+		offset uint32 = 16
+		length uint32 = 4
+		ifmtu  uint16
+		mss    uint16
+	)
+
+	switch {
+	case mtu == 0:
```
smira, line 375: not sure why `0` can be reached here? 

and default MTU should be 1420.

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -324,6 +363,130 @@ func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamil
...
+			Offset:         2,
+			Len:            2,
+			Op:             expr.ExthdrOpTcpopt,
+		},
+	}
+}
+
+func ifname(n string) []byte {
+	b := make([]byte, 16)
+	copy(b, []byte(n+"\x00"))
```
smira, line 485:
```suggestion
	copy(b, []byte(n))
```

as `b` is all zeroes, there should be no reason to append `0`

internal/app/machined/pkg/controllers/kubespan/nftables.go
smira: looks like with Linux 6.1 minimum MTU will be 1280 https://lore.kernel.org/wireguard/20190321033638.1ff82682@natsu/t/

smira:
@sergelogvinov I can take a look at trying to improve the code a bit and add the tests if that works

internal/app/machined/pkg/controllers/kubespan/nftables.go
sergelogvinov: mtu=0 is default value, maybe it was the reason of that.

smira:
I did some cleanups and fixes:

* MTU size now defaults in the config provider
* added a check for min MTU size
* code refactoring
* fixed the code to update nftables rules when MTU (MSS) changes

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -175,6 +178,22 @@ func (m *nfTablesManager) setNFTable(ips *netipx.IPSet) error {
 		return fmt.Errorf("failed to add IPv6 set: %w", err)
 	}
 
+	// meta iifname "lo" accept
```
frezbo, line 181:
```suggestion
	// meta ifname "lo" accept
```

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -258,17 +297,24 @@ func matchIPv6Set(set *nftables.Set, mark, mask uint32) []expr.Any {
...
+func ipOffsetLength(family nftables.TableFamily) (offset uint32, length uint32) {
+	switch family { //nolint:exhaustive
+	case nftables.TableFamilyIPv4:
+		offset = 16
+		length = 4
+	case nftables.TableFamilyIPv6:
 		offset = 24
 		length = 16
+	default:
+		panic("unexpected IP family")
```
frezbo, line 309: should we really panic here?

internal/app/machined/pkg/controllers/kubespan/nftables.go
```diff
@@ -324,6 +370,119 @@ func matchIPSet(set *nftables.Set, mark, mask uint32, family nftables.TableFamil
...
 }
 
+func matchIPSetMSS(set *nftables.Set, mtu uint32, family nftables.TableFamily) []expr.Any {
+	offset, length := ipOffsetLength(family)
+
+	var mss uint16
+
+	switch family { //nolint:exhaustive
+	case nftables.TableFamilyIPv4:
+		mss = uint16(mtu) - 40 // TCP + IPv4 overhead
```
frezbo, line 380: maybe a comment on what 40 and 60 means would be helpful

frezbo (approved):

internal/app/machined/pkg/controllers/kubespan/nftables.go
smira: yes, we call this code (no user input), and it's tested, so if it panics, something went wrong

internal/app/machined/pkg/controllers/kubespan/nftables.go
smira: isn't this what I added on this line?

internal/app/machined/pkg/controllers/kubespan/nftables.go
smira: fixed

internal/app/machined/pkg/controllers/kubespan/nftables.go
frezbo: missed it :sweat_smile:

smira (approved):

Issue #3989, opened 2021-07-27
add context delete option


Ulexus:
## Feature Request

Especially since adding contexts can be automated, we should add a context remove CLI option, as well.

`talosctl context remove XXXX`

Ideally, this should allow glob patterns, too:  `talosctl context remove talos-default-*`.

hobyte:
Love this Idea

It would be nice, if this can delete the kubernetes context too

murtaza-u:
There's already the `talosctl config add <myctx>` command to add new contexts. It would make much more sense to plan the new sub-command as `talosctl config remove <myctx>`.

DmitriyMV:
@murtaza-u are you going to open pull request?

murtaza-u:
@DmitriyMV Yeah...shortly.

DmitriyMV:
Just wanted to say, that we probably want confirmation dialogue for each match individually.

murtaza-u:
Yeah that actually makes sense. The way I have it setup is that there's only one confirmation dialogue for all the matches. Any suggestion on how we should go about designing this? May be a flag like `--step` which steps over each match one by one.

DmitriyMV:
`noconfirm` should be enough when you want to delete every match without confirmations. For completeness there could probably be a `dry` mode which will basically show what will be deleted.

DmitriyMV:
Closed in #6671

PR #6671, opened 2022-12-22
New sub-command `talosctl config remove` to remove contexts

murtaza-u:
## What? (description)

Adds a new sub-command to `talosctl config`. It takes in the context to be deleted as argument and supports glob matching.

A local flag `--noconfirm|-y` can be passed to bypass the confirmation prompt.

It also supports dry run by passing the `--dry-run` flag similar to `apply-config` and `edit` commands.

Example:

```
$ talosctl config remove 'ctx-*'
Remove context ctx-a(*)? (y/N): n
Remove context ctx-b? (y/N): y
Remove context ctx-c? (y/N): y
```

## Why? (reasoning)

`talosctl` has a command to add contexts (`talosctl config add myctx`) but it was missing one to remove contexts.

## Acceptance

References #3989

### DIFF
--- a/cmd/talosctl/cmd/talos/config.go
+++ b/cmd/talosctl/cmd/talos/config.go
@@ -10,6 +10,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"os"
 	"sort"
@@ -19,6 +20,7 @@ import (
 	"time"
 
 	"github.com/dustin/go-humanize"
+	"github.com/ryanuber/go-glob"
 	"github.com/siderolabs/gen/maps"
 	"github.com/spf13/cobra"
 	"google.golang.org/protobuf/types/known/durationpb"
@@ -184,6 +186,89 @@ var configAddCmd = &cobra.Command{
 
 		return nil
 	},
+}
+
+// configRemoveCmdFlags represents the `config remove` command flags.
+var configRemoveCmdFlags struct {
+	noconfirm bool
+	dry       bool
+}
+
+// configRemoveCmd represents the `config remove` command.
+var configRemoveCmd = &cobra.Command{
+	Use:   "remove <context>",
+	Short: "Remove contexts",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		pattern := args[0]
+		if pattern == "" {
+			return fmt.Errorf("no context specified")
+		}
+
+		c, err := clientconfig.Open(GlobalArgs.Talosconfig)
+		if err != nil {
+			return fmt.Errorf("error reading config: %w", err)
+		}
+
+		if len(c.Contexts) == 0 {
+			return errors.New("no contexts defined")
+		}
+
+		matches := sortInPlace(maps.Keys(
+			maps.Filter(c.Contexts, func(context string, _ *clientconfig.Context) bool {
+				return glob.Glob(pattern, context)
+			}),
+		))
+		if len(matches) == 0 {
+			return fmt.Errorf("no contexts matched %q", pattern)
+		}
+
+		// we want to prevent file updates in case there were no changes
+		noChanges := true
+
+		for _, match := range matches {
+			if match == c.Context {
+				fmt.Fprintf(
+					os.Stderr,
+					"skipping removal of current context %q, please change it to another before removing\n",
+					match,
+				)
+
+				continue
+			}
+
+			if !configRemoveCmdFlags.noconfirm {
+				prompt := fmt.Sprintf("remove context %q", match)
+
+				if !helpers.Confirm(prompt + "?") {
+					continue
+				}
+			} else {
+				fmt.Fprintf(os.Stderr, "removing context %q\n", match)
+			}
+
+			noChanges = false
+			delete(c.Contexts, match)
+		}
+
+		if configRemoveCmdFlags.dry || noChanges {
+			return nil
+		}
+
+		err = c.Save(GlobalArgs.Talosconfig)
+		if err != nil {
+			return fmt.Errorf("error writing config: %w", err)
+		}
+
+		return nil
+	},
+	ValidArgsFunction: CompleteConfigContext,
+}
+
+func sortInPlace(slc []string) []string {
+	sort.Slice(slc, func(i, j int) bool { return slc[i] < slc[j] })
+
+	return slc
 }
 
 func checkAndSetCrtAndKey(configContext *clientconfig.Context) error {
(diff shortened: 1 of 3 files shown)


DmitriyMV:
Okay, so I did various small additions:
- It will not skip the current context if it's not the only one in contexts.
- It will remove the config file if you are trying to remove the last context.

DmitriyMV (approved):

murtaza-u:
> It will remove the config file if you are trying to remove the last context.

Just wanted to say that we should reconsider deleting the config file. What if in future Talos decides to store more than just contexts in its config file?

cmd/talosctl/cmd/talos/config.go
```diff
@@ -186,6 +188,106 @@ var configAddCmd = &cobra.Command{
...
+
+		for _, match := range matches {
+			if match == c.Context {
+				fmt.Fprintf(os.Stderr, "Skipping removal of the current context %q, please change it to another before removing\n", match)
+
+				continue
+			}
+
+			if !configRemoveCmdFlags.noconfirm {
+				prompt := "Remove context " + match
```
smira, line 248: we never capitalize prompts/output of the CLI

cmd/talosctl/cmd/talos/config.go
```diff
@@ -186,6 +188,106 @@ var configAddCmd = &cobra.Command{
...
+				if match == c.Context {
+					prompt += "(*)"
+				}
+
+				if !helpers.Confirm(prompt + "?") {
```
smira, line 253: we should already have such function somewhere... probably around the config merging

cmd/talosctl/cmd/talos/config.go
```diff
@@ -186,6 +188,106 @@ var configAddCmd = &cobra.Command{
...
+					continue
+				}
+			} else {
+				fmt.Fprintln(os.Stderr, "Removing context", match)
```
smira, line 257: same here and other places - 
```suggestion
				fmt.Fprintln(os.Stderr, "removing context", match)
```

cmd/talosctl/cmd/talos/config.go
```diff
@@ -186,6 +188,106 @@ var configAddCmd = &cobra.Command{
...
+}
+
+// configRemoveCmd represents the `config remove` command.
+var configRemoveCmd = &cobra.Command{
+	Use:   "remove <context>",
+	Short: "Remove a context",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		pattern := args[0]
+		if pattern == "" {
```
murtaza-u, line 204: Unncessary check. `cobra.ExactArgs(1)` will make sure an argument is passed.

cmd/talosctl/cmd/talos/config.go
```diff
@@ -186,6 +188,106 @@ var configAddCmd = &cobra.Command{
...
+			}
+
+			return removeConfig(c)
+		}
+
+		noChanges := true
+
+		for _, match := range matches {
+			if match == c.Context {
+				fmt.Fprintf(os.Stderr, "Skipping removal of the current context %q, please change it to another before removing\n", match)
```
murtaza-u, line 242: Lets leave this decision to the user. The subtle '*' succeeding the current context acts as a warning in itself.

cmd/talosctl/cmd/talos/config.go
murtaza-u: Actually I checked there isn't any function that does this. The closest thing I could find is this (located in cmd/talos/kubeconfig.go)

```go
func askOverwriteOrRename(prompt string) (kubeconfig.ConflictDecision, error) {
	reader := bufio.NewReader(os.Stdin)

	for {
		fmt.Printf("%s [(r)ename/(o)verwrite]: ", prompt)

		response, err := reader.ReadString('\n')
		if err != nil {
			return "", err
		}

		switch strings.ToLower(strings.TrimSpace(response)) {
		case "overwrite", "o":
			return kubeconfig.OverwriteDecision, nil
		case "rename", "r":
			return kubeconfig.RenameDecision, nil
		}
	}
}
```

cmd/talosctl/cmd/talos/config.go
murtaza-u: I need confirmation on this one before pushing new changes. How do you feel about this @DmitriyMV ?

cmd/talosctl/cmd/talos/config.go
DmitriyMV: Not exactly. You can still do `""`.

cmd/talosctl/cmd/talos/config.go
DmitriyMV: Sorry for the delay. We don't want to remove the current context, because `talosctl` it starts to produce errors like `error constructing client: failed to determine endpoints` which is not exactly descriptive. Granted, it will produce the same error message if there is no config at all, but that is at least easier to debug.

We can do two things. Either:
a) Do not allow the removal of the current context.
b) Switch to another context (which one?).

But we cannot leave current context blank if other contexts are present.

DmitriyMV:
@murtaza-u 

> Just wanted to say that we should reconsider deleting the config file. What if in future Talos decides to store more than just contexts in its config file?

Thats looking a bit too far, but since file is getting recreated by `talosctl` anyway I see no reason to do this either.

murtaza-u:
> @murtaza-u
> 
> > Just wanted to say that we should reconsider deleting the config file. What if in future Talos decides to store more than just contexts in its config file?
> 
> Thats looking a bit too far, but since file is getting recreated by `talosctl` anyway I see no reason to do this either.

Exactly. My point was that such small details get forgotten as time goes by and may lead to unexpected results.

I've made the suggested/discussed changes, kindly review it.

cmd/talosctl/cmd/talos/config.go
murtaza-u: Did not see your changes and forced pushed my own, sorry for that.

Skipping the current context (option a) is a much better option rather than switching to another one. Also the following code will fail if all the contexts have already been deleted.

```go
		if !noChanges && c.Context == "" && len(c.Contexts) > 0 {
			// we ensure that there is always a current context if there
			// are any contexts

			// sortInPlace receives `nil` if len(c.Contexts) == 0
			c.Context = sortInPlace(maps.Keys(c.Contexts))[0]
		}
```

DmitriyMV (approved):

DmitriyMV:
@murtaza-u this is now merged. Congrats!

murtaza-u:
> @murtaza-u this is now merged. Congrats!

Thank you! This was my first PR at Talos. I'm Grateful for all the help.

PR #6679, opened 2022-12-23
feat: update Linux 6.1.1, containerd 1.6.14

smira:
Bumps tools/pkgs/extras to the latest.

Bumps Go modules.

Enables adaptive capacity for COSI state.


internal/app/machined/pkg/adapters/perf/cpu.go
```diff
@@ -41,8 +40,22 @@ func (a cpu) Update(stat *procfs.Stat) {
 		}
 	}
 
-	translateListOfCPUStat := func(in []procfs.CPUStat) []perf.CPUStat {
-		return slices.Map(in, translateCPUStat)
+	translateListOfCPUStat := func(in map[int64]procfs.CPUStat) []perf.CPUStat {
```
frezbo, line 43: why did we drop `slices.Map`?

frezbo (approved):

internal/app/machined/pkg/adapters/perf/cpu.go
smira: because update to prometheus/procfs changed the type for CPU stats:

```go
[]procfs.CPUStat
```

became

```go
map[int64]procfs.CPUStat
```

This I guess makes more sense, as some CPUs might be missing. 

But at the same time we can't change our resources/APIs, so we need to convert back to slice, which is tricky in this case.

DmitriyMV, line 43: Since it's no longer a simple lambda-like function maybe we should promote it and `translateCPUStat` to standalone global functions?

DmitriyMV: Will probably need a way to show that CPU Core data is missing.

smira: we keep it same as way as before - it will be all zeroes (that's what procfs would generate before)

DmitriyMV (approved):

internal/app/machined/pkg/adapters/perf/cpu.go
smira: no, it's different actually in both places - it translates to different data structures

DmitriyMV: Ah, okay!

PR #6689, opened 2022-12-27
feat: ipv6 on gcp

sergelogvinov:
Introduce ipv6 to the google cloud.

It also can work with dhcpv6 is on.
But the route receives through RA packages which not working.

# Pull Request

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
+	"github.com/siderolabs/talos/pkg/machinery/nethelpers"
 	"github.com/siderolabs/talos/pkg/machinery/resources/network"
 	runtimeres "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
 )
@@ -32,7 +33,9 @@ func (g *GCP) Name() string {
 }
 
 // ParseMetadata converts GCP platform metadata into platform network config.
-func (g *GCP) ParseMetadata(metadata *MetadataConfig) (*runtime.PlatformNetworkConfig, error) {
+//
+//nolint:gocyclo
+func (g *GCP) ParseMetadata(metadata *MetadataConfig, interfaces []NetworkInterfaceConfig) (*runtime.PlatformNetworkConfig, error) {
 	networkConfig := &runtime.PlatformNetworkConfig{}
 
 	if metadata.Hostname != "" {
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
```diff
@@ -71,6 +68,81 @@ func (g *GCP) ParseMetadata(metadata *MetadataConfig) (*runtime.PlatformNetworkC
...
+				continue
+			}
+
+			ipPrefix, err := netip.ParsePrefix(ipv6addr)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse ip address: %w", err)
+			}
+
+			if !strings.HasPrefix(ipv6addr, "fd20:") {
+				publicIPs = append(publicIPs, strings.SplitN(ipv6addr, "/", 2)[0])
```
smira, line 103: we shouldn't expose same address as both assigned to the interface and public IP.

public IP (or external IP how we call it) is only a hack for 1:1 NAT, as the address is not assigned to the interfaces, but it is in fact there via 1:1 NAT. so assigning the address to the interface is all we need

internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/metadata.go
```diff
@@ -61,16 +79,26 @@ func (g *GCP) getMetadata(context.Context) (*MetadataConfig, error) {
...
 	meta.Preempted, err = metadata.Get("instance/scheduling/preemptible")
 	if err != nil {
 		return nil, err
 	}
 
 	return &meta, nil
 }
+
+func (g *GCP) getNetworkMetadata(ctx context.Context) ([]NetworkInterfaceConfig, error) {
+	metadataNetworkConfigDl, err := download.Download(ctx, GCPInterfacesEndpoint,
```
smira, line 91: can we use the Google metadata package for it? 

Looks like we could with `metadata.Get("instance/network-interfaces/?recursive=true&alt=json")` and decode the result as JSON. Let's not use this ugly custom download requests :)

internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
sergelogvinov: Yep, node can have many public accessible IPs (subnets). Only platform knows witch one is for node. 

I think the best way to add all main public accessible IP to the array externalIPs. I noticed that behavior on different platforms. IP exists on the node and meta-server expose it as external IP (1:1 NAT)

smira: Hold on, IPv6 addresses are all assigned to the interface (e.g. `eth0`). They _don't_ need to be exposed as `publicIPs`, as they're already on the node.

IPv4 addresses are different - private IP is assigned to the interface, while public IP is not known to the node, so we put as public IP to issue proper certs. 

But there's no reason to do that for IPv6

sergelogvinov: yep, IPv6 have to exist on the node...
I've used externalIP to understand witch one is real(main) ipv6 on the node. And it should be accessible from the internet.

I've removed this code.

smira (approved):
thank you!

PR #6692, opened 2022-12-28
fix: implement upgrade version checks for Talos 1.4

smira:
I missed that before cutting 1.4.0-alpha.0 release, which means unfortunately that Talos 1.4.0-alpha.0 can't upgrade Talos 1.4.0-alpha.0.

### DIFF
--- a/pkg/machinery/compatibility/talos_version.go
+++ b/pkg/machinery/compatibility/talos_version.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/siderolabs/talos/pkg/machinery/api/machine"
 	"github.com/siderolabs/talos/pkg/machinery/compatibility/talos13"
+	"github.com/siderolabs/talos/pkg/machinery/compatibility/talos14"
 )
 
 // TalosVersion embeds Talos version.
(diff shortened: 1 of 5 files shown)


pkg/machinery/compatibility/talos_version.go
```diff
@@ -38,24 +39,35 @@ func (v *TalosVersion) String() string {
...
-			}
-		}
+	if host.version.Core().GreaterThanOrEqual(maxHostDowngradeVersion) {
+		return fmt.Errorf("host version %s is too new to downgrade to Talos %s", host.version.String(), v.version.String())
+	}
 
-		return nil
-	default:
-		return fmt.Errorf("upgrades to version %s are not supported", v.version.String())
+	for _, blacklisted := range deniedHostUpgradeVersions {
```
frezbo, line 66:
```suggestion
	for _, deniedHostUpgradeVersion := range deniedHostUpgradeVersions {
```

frezbo (approved):

pkg/machinery/compatibility/talos_version.go
smira: fixed this, and also another one `blacklisted`

PR #6694, opened 2022-12-28
feat: include Kubernetes controlplane endpoint as one of the endpoints

smira:
These endpoints are used for workers to find the addresses of the controlplane nodes to connect to `trustd` to issue certificates of `apid`.

These endpoints today come from two sources:

* discovery service data
* Kubernetes API server endpoints

This PR adds to the list static entry based on the Kubernetes control plane endpoint in the machine config.

E.g. if the loadbalancer is used for the controlplane endpoint, and that loadbalancer also proxies requests for port 50001 (trustd), this static endpoint will provide workers with connectivity to trustd even if the discovery service is disabled, and Kubernetes API is not up.

If this endpoint doesn't provide any trustd API, Talos will still try other endpoints.

Talos does server certificate validation when calling trustd, so including malicious endpoints doesn't cause any harm, as malicious endpoint can't provider proper server certificate.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/static_endpoint.go
+++ b/internal/app/machined/pkg/controllers/k8s/static_endpoint.go
@@ -1,0 +1,92 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package k8s
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/netip"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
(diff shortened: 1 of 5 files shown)


smira:
e.g. in `talosctl cluster create` environment:

```shell
$ talosctl -n 172.20.0.5 get endpoints
NODE         NAMESPACE      TYPE       ID               VERSION   ADDRESSES
172.20.0.5   controlplane   Endpoint   controlplane     1         ["172.20.0.1"]
172.20.0.5   controlplane   Endpoint   discovery        3         ["172.20.0.2","172.20.0.3","172.20.0.4"]
172.20.0.5   controlplane   Endpoint   kube-apiserver   1         ["172.20.0.2","172.20.0.3","172.20.0.4"]
```

The endpoint 172.20.0.1 gets injected, but it doesn't change anything, as there's no trustd on that address.

frezbo (approved):

PR #6723, opened 2023-01-10
feat: add ISO wipe GRUB boot option

smira:
This simply uses existing `talos.experimental.wipe` kernel flag to wipe the currently installed Talos.

This allows to use Talos ISO to reset the system disk and revert back to maintenance mode.


TimJones (approved):

smira:
This was tested manually with QEMU.

Issue #6681, opened 2022-12-24
`modules.builtin` empty after adding NVIDIA drivers to installer image


rtrox:
### Description

Currently, the way the installer target in the dockerfile is set up, it copies files in from any `customization`s, overwriting any files present with the same name. This is problematic for customizations like the NVIDIA Proprietary drivers, as the Pkgfile which generates those drivers generates a blank `modules.builtin` file in /lib/modules. As such, when you follow the [NVIDIA Proprietary Drivers Instructions](https://www.talos.dev/v1.2/talos-guides/configuration/nvidia-gpu-proprietary/) to the letter, the generated image will have the NVIDIA drivers, but all other built-ins (such as iptables) will be unavailable.

### Logs

No relevant logs to speak of, but I confirmed this in a few ways:
1. Simplest: `talosctl read /lib/modules/5.15.83-talos/modules.builtin` returns no output on a node upgraded with an nvidia image.
2. Less Simple: I downloaded the default installer, and the compiled nvidia installer, and compared by expanding and unsquashing the initramfs in each image. In the nvidia built installer, `modules.builtin` was empty.
3. Confirmation: See the Workaround below, which fixed the inability to utilize iptables by editing the customization image prior to build.

Can definitely give a more complete repro if needed.

### Workaround
I worked around this by removing the `modules.builtin.*` files from my customization prior to build, like this:
```
FROM alpine:latest as customization_prep
COPY --from=ghcr.io/rtrox/nonfree-kmod-nvidia:v1.3.0-5-g6509d23 /lib/modules /tmp/lib/modules
RUN find /tmp/ -name modules.builtin* -exec rm {} \;

FROM scratch as customization
COPY --from=customization_prep /tmp/lib/modules /lib/modules

FROM ghcr.io/siderolabs/installer:v1.3.0
COPY --from=ghcr.io/rtrox/kernel:v1.3.0-5-g6509d23 /boot/vmlinuz /usr/install/${TARGETARCH}/vmlinuz
```
If working around this from scratch, it's likely easier to just remove this line from the `pkgs` repo prior to building:
https://github.com/siderolabs/pkgs/blob/6509d23595cf1862cd80728500b230fcff67ff98/nonfree/kmod-nvidia/pkg.yaml#L40

### Environment

- Talos version:
```
Server:
        NODE:        REDACT
        Tag:         v1.3.0
        SHA:         323e3ccd
        Built:
        Go version:  go1.19.4
        OS/Arch:     linux/amd64
[...]

PR #6751, opened 2023-01-18
feat: generate kernel module dependency tree

frezbo:
Run `depmod` during install/upgrades when extensions provide kernel modules and `modules.dep` needs to be re-generated. This also allows modules of same name from kernel to co-exist. Modules in `extras` folder takes precedence over `in-built` ones.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -652,6 +652,7 @@ RUN apk add --no-cache --update --no-scripts \
     bash \
     cpio \
     efibootmgr \
+    kmod \
     mtools \
     qemu-img \
     squashfs-tools \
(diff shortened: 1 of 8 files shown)


frezbo:
This should also fix #6681

smira (approved):
great work! :cool: 

I think we should have a follow-up to build an integration test with the following:

1. our base kernel has something built as a module (e.g. Mellanox stuff)
2. Talos loads Mellanox drivers by default for backwards compat
3. build something (new) in the kernel as a module, package it as an extension
4. in `e2e-extensions`, install this new extension, and load the module
5. verify that new module got successfully loaded

Issue #6705, opened 2023-01-03
Add generation of docs in Json schema format to support coding assistance


gberche-orange:
## Feature Request

As a talos configuration author
in order to get coding assistance while editing talos config files such as
* completion
* syntax check
* inline documentation
I need the talos documentation to also be generated in json schema format

### Description

Example screenshots of authoring experience using intellij IDEA IDE coding assistance with a POC of Json schema for Talos, which is available at https://github.com/siderolabs/talos/compare/main...orange-cloudfoundry:talos:json-schema?expand=1

Completion
![image](https://user-images.githubusercontent.com/4748380/210369287-28fd2bfa-f522-45dc-af4e-065e17b11589.png)
![image](https://user-images.githubusercontent.com/4748380/210369237-247f738c-101a-43dc-a6df-0d53704e16a5.png)

Static code analysis
![image](https://user-images.githubusercontent.com/4748380/210369978-bed56461-8391-4000-a9aa-df2978adb1b1.png)
![image](https://user-images.githubusercontent.com/4748380/210370027-a1708ddc-ee8c-456c-a351-633a61dae4fa.png)

Inline documentation
![image](https://user-images.githubusercontent.com/4748380/210370372-ee8a3b17-d952-481b-9603-b002a366ae5c.png)

### Additional details

I have the intuition that the existing docgen tool might be able to support automatic generation of [Json Schema](https://json-schema.org/) doc. There are many json schema versions, https://json-schema.org/specification-links.html with various implementations supporting them available at https://json-schema.org/implementations.html  including IDEs such as intellij https://www.jetbrains.com/help/idea/json.html which supports draft 4,6,7

![image](https://user-images.githubusercontent.com/4748380/210372453-d8ad3de8-4065-43f5-bc34-83a03944b232.png)

I'm unfortunately not fluent enough in go to directly make such contribution to codegen, but may help with testing or json schema syntax.

smira:
That's a great idea, thank you!

We were actually considering implementing this for exactly same purpose - providing code completion and inline help in the editor.

utkuozdemir:
@gberche-orange We have implemented the feature. Schema for the upcoming `v1.4` can be found on: https://www.talos.dev/v1.4/schemas/v1alpha1_config.schema.json

Please give it a try. Your feedback is welcome :)

gberche-orange:
Thanks @utkuozdemir !
I'll test and feedback next week when I'll back in office.

gberche-orange:
[...]

PR #6753, opened 2023-01-18
feat: create JSON schema for v1alpha1.Config

utkuozdemir:
Extend `docgen` tool to generate a JSON schema for `v1alpha1.Config` if a new optional cli arg is provided.

Extend the YAML-structured code comments on config fields to allow overriding the generated schema.

Add custom schemas for complex types.

Related to siderolabs/talos#6705.


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1454,6 +1504,9 @@ type TimeConfig struct {
 	//     Specifies the timeout when the node time is considered to be in sync unlocking the boot sequence.
 	//     NTP sync will be still running in the background.
 	//     Defaults to "infinity" (waiting forever for time sync)
+	//   schema:
+	//     type: string
+	//     pattern: ^[-+]?(((\d+(\.\d*)?|\d*(\.\d+)+)(ns|us|µs|ms|s|m|h))|0)+$
```
utkuozdemir, line 1509: Test of this pattern: https://regex101.com/r/xNTblv/2

frezbo: that's a lot of matches :cool:

utkuozdemir: yes, took them from Go's internal parse tests. It matches all except the ones fail due to overflow.

pkg/machinery/config/types/v1alpha1/v1alpha1_config_schema_test.go
```diff
@@ -0,0 +1,128 @@
...
+var (
+	//go:embed v1alpha1_config.schema.json
+	schemaData string
+
+	//go:embed testdata/schema/v1alpha1_config.tmpl.yaml
+	configTemplateData []byte
+)
+
+func TestSchemaValidation(t *testing.T) {
+	schema, err := validatejsonschema.CompileString("https://www.talos.dev/schemas/v1alpha1/config.json", schemaData)
```
smira, line 36: my only concern with this test is how we can keep it up to date, as it is based on static machine config. 

can we probably try to use `generate` package here and generate the machine config on the fly with the test, patching it as necessary for each test case?

smira (approved):
hard to validate the result lol, but this looks great to me! :cool:

hack/docgen/schema.go
```diff
@@ -0,0 +1,242 @@
...
+		Ref:     "#/$defs/Config",
+	}
+
+	definitions := make(map[string]*jsonschema.Schema, len(doc.Structs))
+
+	for _, st := range doc.Structs {
+		definitions[st.Name] = structToSchema(st)
+	}
+
+	schema.Definitions = definitions
```
DmitriyMV, line 207: I think we can use [slices.ToMap](https://github.com/siderolabs/gen/blob/main/slices/slices.go#L89) here directly

hack/docgen/schema.go
```diff
@@ -0,0 +1,242 @@
...
+type SchemaTypeInfo struct {
+	typeName string
+	ref      string
+}
+
+type SchemaDefinitionInfo struct {
+	typeInfo           SchemaTypeInfo
+	arrayItemsTypeInfo *SchemaTypeInfo
+	mapValueTypeInfo   *SchemaTypeInfo
+	enumValues         []any
```
DmitriyMV, line 54: Any reason for `[]any` and not `[]string`?

DmitriyMV: Ah - assigning to  to `jsonschema.Schema.Enum`.

hack/docgen/schema.go
```diff
@@ -0,0 +1,242 @@
...
+
+	if typeInfo.typeName != "" {
+		schema.Type = typeInfo.typeName
+	}
+
+	if typeInfo.ref != "" {
+		schema.Ref = typeInfo.ref
+	}
+
+	return schema
```
DmitriyMV, line 116: One style here and below will look better I think  - either `&schema` everywhere or `&jsonschema.Schema{}`.

DmitriyMV (approved):

hack/docgen/schema.go
utkuozdemir: done

hack/docgen/schema.go
utkuozdemir: good point, done

PR #6755, opened 2023-01-19
chore: preallocate disk images for QEMU VMs

smira:
This improves the performance of the I/O operations if the underlying filesystem supports it.

### DIFF
--- a/pkg/provision/providers/vm/disk.go
+++ b/pkg/provision/providers/vm/disk.go
@@ -7,6 +7,7 @@ package vm
 import (
 	"fmt"
 	"os"
+	"syscall"
 
 	"github.com/siderolabs/talos/pkg/provision"
 )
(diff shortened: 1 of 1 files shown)


pkg/provision/providers/vm/disk.go
```diff
@@ -49,15 +50,20 @@ func (p *Provisioner) CreateDisks(state *State, nodeReq provision.NodeRequest) (
 
 		defer diskF.Close() //nolint:errcheck
 
-		err = diskF.Truncate(int64(disk.Size))
+		if err = diskF.Truncate(int64(disk.Size)); err != nil {
+			return nil, err
+		}
+
+		if err = syscall.Fallocate(int(diskF.Fd()), 0, 0, int64(disk.Size)); err != nil {
+			fmt.Fprintf(os.Stderr, "WARNING: failed to allocate disk space for %q (size %d): %s", diskPath, disk.Size, err)
```
DmitriyMV, line 58: Maybe `preallocate` will sound better here, since allocate means we can't get required storage at all, and thats not the case I think.

pkg/provision/providers/vm/disk.go
```diff
@@ -49,15 +50,20 @@ func (p *Provisioner) CreateDisks(state *State, nodeReq provision.NodeRequest) (
...
+		}
+
 		diskPaths[i] = diskPath
 	}
 
 	if len(diskPaths) == 0 {
-		err = fmt.Errorf("node request must have at least one disk defined to be used as primary disk")
-
-		return
+		return nil, fmt.Errorf("node request must have at least one disk defined to be used as primary disk")
```
DmitriyMV, line 65: Maybe we should just drop named returns here? It looks kinda inconsistent  with whats written above in this function.

smira: I see named returns as documentation here, so I kept them, but all `return`s are explicit, as otherwise function returning `[]string, error` is confusing

pkg/provision/providers/vm/disk.go
smira: good point, fixed

pkg/provision/providers/vm/disk.go
DmitriyMV: Oh, okay!

DmitriyMV (approved):

frezbo (approved):

PR #6757, opened 2023-01-19
chore: bump deps

DmitriyMV:
For some reason `go-mod-outdated` didn't work for me, so I had to do this manually.


go.mod
```diff
@@ -30,21 +30,21 @@ require (
...
-	github.com/aws/aws-sdk-go v1.44.177
+	github.com/aws/aws-sdk-go v1.44.182
 	github.com/beevik/ntp v0.3.0
 	github.com/cenkalti/backoff/v4 v4.2.0
 	github.com/containerd/cgroups v1.0.4
 	github.com/containerd/containerd v1.6.15
 	github.com/containerd/typeurl v1.0.2
 	github.com/containernetworking/cni v1.1.2
-	github.com/containernetworking/plugins v1.1.1
+	github.com/containernetworking/plugins v1.2.0
```
smira, line 40: ok, we have CNI plugin bump coming in `pkgs` as well, that's good

internal/app/machined/pkg/controllers/runtime/events_sink.go
```diff
@@ -15,7 +16,6 @@ import (
 	"github.com/siderolabs/go-pointer"
 	"github.com/siderolabs/go-procfs/procfs"
 	"github.com/siderolabs/siderolink/api/events"
-	"go.uber.org/atomic"
```
smira, line 18: this looks werid

DmitriyMV: It's actually not - I removed unnecessary side dependency since it's already in std.

smira: oh, I read it backwards :-D

I guess it was too late for me

smira (approved):

PR #6769, opened 2023-01-23
fix: mark DigitalOcean anchor IP as scope link

### LINKED ISSUE #6760: DigitalOcean: drop 'anchorIPs' from platform network configuration
## Feature Request

According to the DO docs and our testing, they are not being used with Reserved IP and custom image droplets.

This adds another address to the node which is not routeable, and it has to be explicitly excluded for kubelet/etcd.

smira:
This excludes it out of the `NodeAddress`.

Needs extra testing to confirm that it actually still works as anchor IP.

Fixes #6760


smira:
Did some manual testing with DO, using Talos image marked as 'Custom OS' and 'Ubuntu'. Both work, including 'Reserved IP', but the connection in either case doesn't come via anchor IP. So making its scope to be link doesn't seem to affect Talos operations.

rsmitty (approved):

frezbo (approved):

PR #6781, opened 2023-01-24
fix: renovate config

frezbo:
Add proper `extractVersion` regex.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -13,7 +13,7 @@
                 "Makefile"
             ],
             "matchStrings": [
-                "# renovate: datasource=(?<datasource>.*?) depName=(?<depName>.+?)\\s.*_VERSION\\s+\\?=\\s+(?<currentValue>.+)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -17,7 +17,7 @@ TOOLS ?= ghcr.io/siderolabs/tools:v1.4.0-alpha.0-5-g763c1d9
 PKGS ?= v1.4.0-alpha.0-11-ge561dcb
 EXTRAS ?= v1.4.0-alpha.0-2-g8cb4792
 # renovate: datasource=github-tags depName=golang/go
-GO_VERSION ?= 1.19
+GO_VERSION ?= 1.19.2
```
smira, line 20: This is not the real Go version used, but for some extra steps which don't use tools. I would rather keep it without a patch release

smira (approved):

Makefile
frezbo: yes, but go seems to release images for the patch versions too, or I can update the version to check for major.minor only

frezbo: updated  to only check for major.minor

PR #6783, opened 2023-01-24
test: build integration test early in the pipeline

smira:
If we don't pre-build, it's getting built each time the `e2e-*` step runs, and we have some running in parallel.


Makefile
```diff
@@ -335,7 +335,7 @@ unit-tests-race: ## Performs unit tests with race detection enabled.
 	@$(MAKE) target-$@ TARGET_ARGS="--allow security.insecure" PLATFORM=linux/amd64
 
 $(ARTIFACTS)/$(INTEGRATION_TEST_DEFAULT_TARGET)-amd64:
-	@$(MAKE) local-$(INTEGRATION_TEST_DEFAULT_TARGET) DEST=$(ARTIFACTS) PLATFORM=linux/amd64 WITH_RACE=true NAME=Client
+	@$(MAKE) local-$(INTEGRATION_TEST_DEFAULT_TARGET) DEST=$(ARTIFACTS) PLATFORM=linux/amd64 WITH_RACE=true NAME=Client PUSH=false
```
frezbo, line 338: why no push? does it use the cache?

frezbo (approved):

Makefile
smira: it's more of a UX thing, `local-` can't push, and it will complain if `PUSH=true` is enabled

we do `make talosctl installer PUSH=true`, and each step which is not an image overrides `PUSH=false` so that whole line still can work. so you can have a single line with `PUSH=true`, but only those which support PUSH will actually push :)

PR #6805, opened 2023-01-30
feat: automatically load modules based on hardware info

frezbo:
Enable `udevd` to automatically load modules based on hardware info.


frezbo:
seems the build is failing. I guess the amlogic script needs something extra

pkg/machinery/constants/constants.go
```diff
@@ -814,6 +814,16 @@ var Overlays = []string{
 	"/etc/cni",
 	"/etc/kubernetes",
 	"/usr/libexec/kubernetes",
-	"/usr/etc/udev",
 	"/opt",
 }
+
+// DefaultDroppedCapabilities is the default set of capabilities to drop.
```
frezbo, line 820: importing `cap` package breaks darwin builds of talosctl, so using the string representation instead

frezbo:
Implemented as three separate PR's:

#6814 
#6822 
#6823 

Closing

PR #6809, opened 2023-02-02
fix: correctly expand parameters in the URL

### LINKED ISSUE #6807: Using ${hostname} in talos.config kernel parameter causes kernel panic
### Description

I am setting up Talos device provisioning via PXE boot and want to make use of the talos.config kernel parameter to download a config while booting with PXE. As soon as I add the `&hostname=${hostname}` part Talos crashes with a kernel panic.

` 
  KERNEL /Linux/Talos/vmlinuz-amd64
  append initrd=/Linux/Talos/initramfs-amd64 talos.platform=metal slab_nomerge pti=on init_on_alloc=1 init_on_free=1 earlyprintk=ttyS0 console=ttyS0 console=tty0 nvme_core.io_timeout=4294967295 printk.devmsg=on ima_template=ima-ng ima_appraise=fix ima_hash=sha512 talos.config=http://192.168.100.105:8080/configure?uuid=${uuid}&serial=${serial}&mac=${mac}&hostname=${hostname}
`
- A false config is usually identified properly and the node will be rebooted
- No trace at the webserver serving the configure endpoint when hostname is added to the querystring

### Logs

- Video has been shared with @smira 

### Environment

- Talos version: 1.3.2
- Kubernetes version: NA
- Platform: amd64 (Gen 1 VM in Hyper-V via PXE boot)

smira:
This fixes multiple issues:

* `log.Fatalf` in the machined code leads to kernel panic
* return URL if some expansion fails
* correctly handle destroyed event (wait for the next one)

Fixes #6807

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -51,7 +51,7 @@ func (m *Metal) Configuration(ctx context.Context, r state.State) ([]byte, error
 	getURL := func() string {
 		downloadEndpoint, err := PopulateURLParameters(ctx, *option, r)
 		if err != nil {
-			log.Fatalf("failed to populate talos.config fetch URL: %q ; %s", *option, err.Error())
+			log.Printf("failed to populate talos.config fetch URL: %q ; %s", *option, err.Error())
 		}
 
 		log.Printf("fetching machine config from: %q", downloadEndpoint)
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -51,7 +51,7 @@ func (m *Metal) Configuration(ctx context.Context, r state.State) ([]byte, error
 	getURL := func() string {
 		downloadEndpoint, err := PopulateURLParameters(ctx, *option, r)
 		if err != nil {
-			log.Fatalf("failed to populate talos.config fetch URL: %q ; %s", *option, err.Error())
+			log.Printf("failed to populate talos.config fetch URL: %q ; %s", *option, err.Error())
```
frezbo, line 54: would this retry?

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
smira: yes, but in non-obvious way - this function gets called from `download.Download` on each attempt, so the error is not fatal - next time if will be called if the download fails from partially expanded URL.

PR #6814, opened 2023-02-03
feat: use wrapper for starting processes

frezbo:
Use a wrapper for starting processes which can setup proper cgroups, OOMscore, and also drop capabilities for the process, then it calls `execve`.

The containerd tests is also fixed to support cgroups when running tests in buildkit. It used to pass previosuly as we did not error if cgroup setup failed.

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd_test.go
@@ -74,16 +74,14 @@ func (suite *ContainerdSuite) SetupSuite() {
 	suite.Require().NoError(os.Mkdir(rootDir, 0o777))
 
 	if cgroups.Mode() == cgroups.Unified {
-		var (
-			groupPath string
-			manager   *cgroupsv2.Manager
-		)
+		var manager *cgroupsv2.Manager
 
-		groupPath, err = cgroupsv2.NestedGroupPath(suite.tmpDir)
+		manager, err = cgroupsv2.NewManager(constants.CgroupMountPath, "/"+suite.T().Name(), &cgroupsv2.Resources{})
 		suite.Require().NoError(err)
 
-		manager, err = cgroupsv2.NewManager(constants.CgroupMountPath, groupPath, &cgroupsv2.Resources{})
-		suite.Require().NoError(err)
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd_test.go
```diff
@@ -74,16 +74,14 @@ func (suite *ContainerdSuite) SetupSuite() {
...
 
-		groupPath, err = cgroupsv2.NestedGroupPath(suite.tmpDir)
+		manager, err = cgroupsv2.NewManager(constants.CgroupMountPath, "/"+suite.T().Name(), &cgroupsv2.Resources{})
 		suite.Require().NoError(err)
 
-		manager, err = cgroupsv2.NewManager(constants.CgroupMountPath, groupPath, &cgroupsv2.Resources{})
-		suite.Require().NoError(err)
+		// when using buildkit runner, parent `cgroup.type` is set to `domain threaded`, so child cgroups have to explicitly specify
+		// `cgroup.type` to "threaded" https://www.kernel.org/doc/html/v5.0/admin-guide/cgroup-v2.html#threads
+		suite.Require().NoError(os.WriteFile(filepath.Join(constants.CgroupMountPath, suite.T().Name(), "cgroup.type"), []byte("threaded"), 0o644))
```
smira, line 84: :heart:

internal/app/wrapperd/main.go
```diff
@@ -0,0 +1,98 @@
...
+// Main is the entrypoint into /sbin/wrapperd.
+// nolint: gocyclo
+func Main() {
+	flag.StringVar(&name, "name", "", "process name")
+	flag.StringVar(&droppedCaps, "dropped-caps", "", "comma-separated list of capabilities to drop")
+	flag.StringVar(&cgroupPath, "cgroup-path", "", "cgroup path to use")
+	flag.IntVar(&oomScore, "oom-score", 0, "oom score to set")
+	flag.Parse()
+
+	if droppedCaps != "" {
```
smira, line 40: should we drop capabilities last, as we might not be able to adjust OOM score/enter cgroup with lowered caps?

smira (approved):

internal/app/wrapperd/main.go
frezbo: good point, I'll move that

PR #6820, opened 2023-02-05
fix: trackable action flag usage text. --no-wait does not exist

wind0r:
# Pull Request

## What? (description)
Fix --debug usage text. Flag --wait gets set to true

## Why? (reasoning)
It looks like while developing --no-wait got replaced with --wait, but the debug usage description did not get changed

## Acceptance


cmd/talosctl/cmd/talos/track.go
```diff
@@ -18,6 +18,6 @@ type trackableActionCmdFlags struct {
 
 func (f *trackableActionCmdFlags) addTrackActionFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&f.wait, "wait", true, "wait for the operation to complete, tracking its progress. always set to true when --debug is set")
-	cmd.Flags().BoolVar(&f.debug, "debug", false, "debug operation from kernel logs. --no-wait is set to false when this flag is set")
+	cmd.Flags().BoolVar(&f.debug, "debug", false, "debug operation from kernel logs. --wait is set to true when this flag is set")
```
frezbo, line 21: I believe `wait` is true by default since talos 1.2, right @utkuozdemir ?

smira: yes, it defaults to `--wait`, but still `--debug=true --wait=false` will imply `--wait=true`

utkuozdemir: ⬆️ yep. And `--no-wait` does not exist. So the change on the code is correct.

smira (approved):
thank you!

PR #6822, opened 2023-02-06
feat: use process wrapper for dropping capabilities

frezbo:
Use process wrapper introduced in #6814 to drop capabilities. This change also means the capabilities are dropped per process level and not for PID 1 (machined), which allows us to drop capabilities per process.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -42,7 +42,6 @@ import (
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"golang.org/x/sys/unix"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
-	"kernel.org/pub/linux/libs/security/libcap/cap"
 
 	installer "github.com/siderolabs/talos/cmd/installer/pkg/install"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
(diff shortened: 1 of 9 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -817,3 +817,9 @@ var Overlays = []string{
 	"/usr/etc/udev",
 	"/opt",
 }
+
+// DefaultDroppedCapabilities is the default set of capabilities to drop.
+var DefaultDroppedCapabilities = []string{
```
smira, line 822: we could make this `map[string]struct{}` for easier checks?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -295,39 +294,6 @@ func SetRLimit(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFun
 	}, "setRLimit"
 }
 
-// DropCapabilities drops some capabilities so that they can't be restored by child processes.
-func DropCapabilities(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
-	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
-		prop, err := krnl.ReadParam(&kernel.Param{Key: "proc.sys.kernel.kexec_load_disabled"})
```
smira, line 301: I think this check got dropped as well, it's mostly important for `cri`

internal/app/machined/pkg/system/services/udevd.go
```diff
@@ -84,6 +84,7 @@ func (c *Udevd) Runner(r runtime.Runtime) (runner.Runner, error) {
 		runner.WithLoggingManager(r.Logging()),
 		runner.WithEnv(env),
 		runner.WithCgroupPath(constants.CgroupSystemRuntime),
+		runner.WithDroppedCapabilities(constants.DefaultDroppedCapabilities),
```
smira, line 87: should udevd keep `SYS_MODULE`?

frezbo: no, I'll update than for the udevd PR

pkg/machinery/constants/constants.go
frezbo: I didn't quite get that

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo: fixed

pkg/machinery/constants/constants.go
smira: I mean if we change it from being a slice to being a map with capabilities as keys, we can simplify lookup

smira: I mean https://github.com/siderolabs/talos/pull/6822/files#diff-0be4290a97f76e2ccbaa28623fbfd909d32b6385dabe780a6acf080af7025cd2R24 here

frezbo: fixed

internal/app/machined/pkg/system/runner/runner.go
```diff
@@ -167,9 +167,15 @@ func WithCustomSeccompProfile(override func(*specs.LinuxSeccomp)) Option {
...
 }
 
-// WithBoundedCapabilities sets the list of capabilities to drop.
-func WithBoundedCapabilities(caps []string) Option {
+// WithDroppedCapabilities sets the list of capabilities to drop.
+func WithDroppedCapabilities(caps map[string]struct{}) Option {
+	droppedCaps := make([]string, 0, len(caps))
+
+	for cap := range caps {
+		droppedCaps = append(droppedCaps, cap)
```
smira, line 175: nit: https://pkg.go.dev/github.com/siderolabs/gen@v0.4.3/maps#Keys

smira (approved):

internal/app/machined/pkg/system/runner/runner.go
frezbo: fixed

PR #6823, opened 2023-02-06
feat: automatically load modules based on hw info

frezbo:
Automatically load kernel modules based on hardware info and modules alias info. udevd would automatically load modules based on HW information present.

Signed-off-by: Noel Georgi <git@frezbo.dev>

PR #6832, opened 2023-02-09
fix: wait for network and retry in platform get config funcs

smira:
Wait for the network before trying to access the metadata service.

Retry the calls when appropriate (most platforms use `download.Download` function which does proper retries).

Co-authored-by: Noel Georgi <git@frezbo.dev>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -201,7 +201,7 @@ RUN gofumpt -w ./
 
 FROM go-generate AS gen-proto-go
 WORKDIR /src/
-RUN structprotogen github.com/siderolabs/talos/pkg/machinery/... /api/resource/definitions/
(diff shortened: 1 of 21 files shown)


pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -18,7 +18,6 @@ import (
 	"github.com/cosi-project/runtime/pkg/safe"
```
frezbo, line 18: these got in :sweat_smile:

smira: yep, already removed these

frezbo (approved):

frezbo:
With this fix #6823 e2e also passes :+1:

PR #6855, opened 2023-02-16
fix: add support for a fallback '*' mirror configuration

smira:
Talos always supported that, but CRI config lacked support for it.

Now with recent containerd the new `_default` host is used as a fallback, so this re-enables the support and updates the docs.

See https://github.com/containerd/containerd/pull/8065


smira:
> hack/release.toml update too i guess?

good point!

frezbo (approved):

PR #6857, opened 2023-02-17
fix: docker talosctl cluster create provisioner

smira:
Recent Docker versions seem to have changed the API in the way container IP addresses are reported.

Also fix running Talos 1.3 image under talosctl 1.4.

### DIFF
--- a/pkg/provision/providers/docker/node.go
+++ b/pkg/provision/providers/docker/node.go
@@ -101,7 +101,8 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/provision/providers/docker/node.go
```diff
@@ -101,7 +101,8 @@ func (p *provisioner) createNode(ctx context.Context, clusterReq provision.Clust
 		})
 	}
```
frezbo, line 103: udev doesn't matter in docker, right?

frezbo (approved):

pkg/provision/providers/docker/node.go
smira: doesn't really, but Talos still does the write and fails

PR #6868, opened 2023-02-20
fix: use passed `--context` in `talosctl config` cmd

### LINKED ISSUE #6846: talosctl config endpoint/node ignores context
### Description

If a `talosconfig` file has more than one context and is used with commands `talosctl config endpoint` and `talosctl config node` the `--context` flag is ignored.

### Logs

```shell
❯ talosctl gen config foo https://foo.example.com:6443 --output "$(mktemp -d)"
generating PKI and tokens
Created /tmp/tmp.gi4UL2fvx4/controlplane.yaml
Created /tmp/tmp.gi4UL2fvx4/worker.yaml
Created /tmp/tmp.gi4UL2fvx4/talosconfig

❯ talosctl gen config bar https://bar.example.com:6443 --output "$(mktemp -d)"
generating PKI and tokens
Created /tmp/tmp.PRFTZEQase/controlplane.yaml
Created /tmp/tmp.PRFTZEQase/worker.yaml
Created /tmp/tmp.PRFTZEQase/talosconfig

# Merge the second talosconfig into the first
❯ talosctl --talosconfig /tmp/tmp.gi4UL2fvx4/talosconfig config merge /tmp/tmp.PRFTZEQase/talosconfig

# "Context" is 'bar'
❯ cat /tmp/tmp.gi4UL2fvx4/talosconfig
context: bar
contexts:
    bar:
        endpoints:
            - 127.0.0.1
        ca: ...
        crt: ...
        key: ...
    foo:
        endpoints:
            - 127.0.0.1
        ca: ...
        crt: ...
        key: ...

# Try to set nodes for 'foo'
❯ talosctl --talosconfig /tmp/tmp.gi4UL2fvx4/talosconfig --context foo config nodes 127.0.0.2

# See they were actually set for 'bar'
❯ cat /tmp/tmp.gi4UL2fvx4/talosconfig
context: bar
contexts:
    bar:
        endpoints:
            - 127.0.0.1
        nodes:
            - 127.0.0.2
        ca: ...
        crt: ...
        key: ...
    foo:
        endpoints:
            - 127.0.0.1
        ca: ...
        crt: ...
        key: ...
```

### Environment

- Talos version: `v.1.3.4`
- Kubernetes version: n/a
- Platform: Linux-amd64

DmitriyMV:
Use context from command line flags. Also some minor fixes.

Closes #6846

### DIFF
--- a/cmd/talosctl/cmd/talos/config.go
+++ b/cmd/talosctl/cmd/talos/config.go
@@ -77,7 +77,12 @@ var configEndpointCmd = &cobra.Command{
 			args[i] = strings.TrimSpace(args[i])
 		}
 
-		c.Contexts[c.Context].Endpoints = args
+		ctxData, err := getContextData(c)
+		if err != nil {
+			return err
+		}
+
+		ctxData.Endpoints = args
 		if err := c.Save(GlobalArgs.Talosconfig); err != nil {
 			return fmt.Errorf("error writing config: %w", err)
 		}
@@ -103,7 +108,12 @@ var configNodeCmd = &cobra.Command{
 			args[i] = strings.TrimSpace(args[i])
 		}
 
-		c.Contexts[c.Context].Nodes = args
+		ctxData, err := getContextData(c)
+		if err != nil {
+			return err
+		}
+
+		ctxData.Nodes = args
 		if err := c.Save(GlobalArgs.Talosconfig); err != nil {
 			return fmt.Errorf("error writing config: %w", err)
 		}
@@ -446,9 +456,12 @@ Certificate expires: {{ .CertTTL }} ({{ .CertNotAfter }})
 //
 //nolint:goconst
 func configInfoCommand(config *clientconfig.Config, now time.Time) (string, error) {
-	context := config.Contexts[config.Context]
+	cfgContext, err := getContextData(config)
+	if err != nil {
+		return "", err
+	}
 
-	b, err := base64.StdEncoding.DecodeString(context.Crt)
+	b, err := base64.StdEncoding.DecodeString(cfgContext.Crt)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/talos/config.go
```diff
@@ -127,8 +129,6 @@ var configContextCmd = &cobra.Command{
 			return err
 		}
 
-		c.Context = context
```
smira, line 130: isn't it what actually does the job in this command?

DmitriyMV: It is now moved [here](https://github.com/siderolabs/talos/pull/6868/files#diff-79095d8dc4c63f2ca9526001f07b86e5f486c035b2832847c6c3925618f6ad02R60).

smira: I don't think this is right, no command should change the current context except for this one.

DmitriyMV: `openConfigAndContext` receives empty context string in all cases except for `node`. Inside it verifies that `context` is empty and uses the one from the file. **Tim** wanted `talosctl config` to support `--context` flag, so we override Context field everywhere instead of only overriding it in `node`. That is actually quite consistent, no?

smira: Updating this field iirc changes the active context in talosconfig. Tim asked for the command to operate on the specified context, but not switch the context as a side effect. (I think that was it)

DmitriyMV: Oh, I understand now - the change will result in override context being set as **active** one and saved in `talosconfig` when we didn't want that. Reverted and pushed the different set of changes.

cmd/talosctl/cmd/talos/config.go
```diff
@@ -446,9 +456,9 @@ Certificate expires: {{ .CertTTL }} ({{ .CertNotAfter }})
 //
 //nolint:goconst
 func configInfoCommand(config *clientconfig.Config, now time.Time) (string, error) {
-	context := config.Contexts[config.Context]
```
smira, line 449: should this be using `getContextData` as well?

smira (approved):

PR #6869, opened 2023-02-20
refactor: extract parts of kubernetes libraries

smira:
The shared code is going out to the
github.com/siderolabs/go-kubernetes library.

The code will be used in Talos and other projects using same features.


smira:
depends on https://github.com/siderolabs/go-kubernetes/pull/2

frezbo (approved):

Issue #4791, opened 2022-01-12
put `talosctl dashboard` on a virtual tty


smira:
## Feature Request

After boot Talos should keep console logs on one tty9, and switch tty1 (0?) to display `talosctl dashboard` with summary view as a default.

### Description

* are we missing any kernel config?
* how to implement switching between virtual ttys?
* how to redirect console logs to another tty?

smira:
setting up virtual consoles: https://github.com/systemd/systemd/blob/main/src/vconsole/vconsole-setup.c

flokli:
This should probably only apply to servers with multiple (functional) ttys.

Headless VMs, or SBCs without a GPU should probably only show their console logs on the serial port (which has a too small "screen size" anyways to make the dashboard useful)

utkuozdemir:
https://github.com/siderolabs/talos/pull/6879

PR #6871, opened 2023-02-20
feat: start machined earlier & in maintenance mode

utkuozdemir:
Load & start machined earlier and in initialize sequence, so that it is possible to use its API over its unix socket in maintenance mode.

Additionally, do not return features from Version API  if a config is not yet available.

Related to siderolabs/talos#4791.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
@@ -77,6 +77,9 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 			CreateSystemCgroups,
 			CreateOSReleaseFile,
 		).Append(
+			"machined",
+			StartMachined,
+		).Append(
 			"config",
 			LoadConfig,
 		)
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/system/runner/goroutine/goroutine.go
```diff
@@ -89,14 +88,7 @@ func (r *goroutineRunner) wrappedMain() (err error) {
 	//nolint:errcheck
 	defer w.Close()
 
-	var writer io.Writer
-	if r.runtime.Config().Debug() {
```
frezbo, line 93: would we lose the debug ability here?

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -102,6 +105,9 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 		).Append(
 			"udevd",
 			StartUdevd,
+		).Append(
```
smira, line 108: we can merge the two phases (udevd & machined) into a single one, so that they start in parallel and we have better startup time ;)

internal/app/machined/pkg/system/runner/goroutine/goroutine.go
smira: yep, like with udevd, but it shouldn't be too crazy...

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
utkuozdemir: yep, done

PR #6878, opened 2023-02-22
feat: add grub option to drop to maintenance mode

### LINKED ISSUE #6842: GRUB option to drop to maintenance mode
1. `talos.experimental.wipe=system:EPHEMERAL,STATE` boot kernel arg
2. GRUB option to wipe like above ^^^ - update GRUB library to handle that

DmitriyMV:
- [x] Support `talos.experimental.wipe=system:EPHEMERAL,STATE` boot kernel arg
- [x] GRUB option to wipe like above
- [x] update GRUB library to handle that

Closes #6842

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/encode.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/encode.go
@@ -32,6 +32,14 @@ menuentry "{{ $entry.Name }}" {
   initrd {{ $entry.Initrd }}
 }
 {{ end -}}
+
+{{ $defaultEntry := index .Entries .Default -}}
+menuentry "Reset Talos installation and return to maintenance mode" {
+  set gfxmode=auto
+  set gfxpayload=text
+  linux {{ $defaultEntry.Linux }} {{ quote $defaultEntry.Cmdline }} talos.experimental.wipe=system:EPHEMERAL,STATE
+  initrd {{ $defaultEntry.Initrd }}
+}
 `
 
 // Write the grub configuration to the given file.
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
@@ -46,7 +46,7 @@ func (p PhaseList) AppendWhen(when bool, name string, tasks ...runtime.TaskSetup
 	return p
 }
 
-// AppendWithDeferredCheck appends a task to the phase list but skips the sequence if `check func()` returns `true` during execution.
+// AppendWithDeferredCheck appends a task to the phase list but skips the sequence if `check func()` returns `false` during execution.
 func (p PhaseList) AppendWithDeferredCheck(check func() bool, name string, tasks ...runtime.TaskSetupFunc) PhaseList {
 	p = append(p, runtime.Phase{
 		Name:      name,
@@ -118,6 +118,14 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 			},
 			"dashboard",
 			StartDashboard,
+		).AppendWithDeferredCheck(
+			func() bool {
+				wipe := procfs.ProcCmdline().Get(constants.KernelParamWipe).First()
+
+				return pointer.SafeDeref(wipe) != ""
+			},
+			"wipeDisks",
+			ResetSystemDiskPartitions,
 		).AppendWithDeferredCheck(
 			func() bool {
 				return r.State().Machine().Installed()
@@ -199,11 +207,6 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
 // This sequence should never be reached if an installation is not found.
 func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
 	phases := PhaseList{}
-
-	wipe := procfs.ProcCmdline().Get(constants.KernelParamWipe).First()
-	if wipe != nil && *wipe == "system" {
-		return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
-	}
 
 	phases = phases.AppendWhen(
 		r.State().Platform().Mode() != runtime.ModeContainer,
@@ -436,7 +439,7 @@ func (*Sequencer) StageUpgrade(r runtime.Runtime, in *machineapi.UpgradeRequest)
 }
 
 // MaintenanceUpgrade is the upgrade sequence in maintenance mode.
-func (*Sequencer) MaintenanceUpgrade(r runtime.Runtime, in *machineapi.UpgradeRequest) []runtime.Phase {
+func (*Sequencer) MaintenanceUpgrade(r runtime.Runtime, _ *machineapi.UpgradeRequest) []runtime.Phase {
 	phases := PhaseList{}
 
 	switch r.State().Platform().Mode() { //nolint:exhaustive
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -30,6 +30,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	pprocfs "github.com/prometheus/procfs"
+	"github.com/siderolabs/gen/slices"
 	"github.com/siderolabs/go-blockdevice/blockdevice"
 	"github.com/siderolabs/go-blockdevice/blockdevice/partition/gpt"
 	"github.com/siderolabs/go-blockdevice/blockdevice/util"
@@ -80,7 +81,7 @@ import (
 )
 
 // SetupLogger represents the SetupLogger task.
-func SetupLogger(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
+func SetupLogger(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
 		machinedLog, err := r.Logging().ServiceLog("machined").Writer()
 		if err != nil {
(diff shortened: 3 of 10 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1672,8 +1672,35 @@ func ResetUserDisks(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, st
...
+			if err != nil {
+				return fmt.Errorf("error reading partition table: %w", err)
+			}
+
+			_, err = target.Locate(pt)
+			if err != nil {
+				return fmt.Errorf("error locating partition with label %q: %w", target.Label, err)
+			}
+
+			if err = target.Format(); err != nil {
```
DmitriyMV, line 1691: Should we also drop and write partition here just like `ResetSystemDisk` does is Reset?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/encode.go
```diff
@@ -32,6 +32,14 @@ menuentry "{{ $entry.Name }}" {
   initrd {{ $entry.Initrd }}
 }
 {{ end -}}
+
+{{ $defaultEntry := index .Entries .Default -}}
+menuentry "Reset Talos installation and return to maintenance mode" {
+  set gfxmode=auto
+  set gfxpayload=text
+  linux {{ $defaultEntry.Linux }} {{ quote $defaultEntry.Cmdline }} talos.experimental.wipe=system:EPHEMERAL,STATE
```
utkuozdemir, line 40: Should we maybe use the existing constants for these?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: noooo! :)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -201,8 +202,22 @@ func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
 	phases := PhaseList{}
 
 	wipe := procfs.ProcCmdline().Get(constants.KernelParamWipe).First()
-	if wipe != nil && *wipe == "system" {
-		return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
+	if wipe != nil {
+		switch *wipe {
+		case "system":
+			return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
+		case "system:EPHEMERAL,STATE":
```
utkuozdemir, line 209: same here - constants?

smira (commented):
in any case we need a `hack/release.toml` section describing the change

new arg option(s) should be documented in the Kernel reference section of the docs for 1.4

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1672,8 +1672,35 @@ func ResetUserDisks(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, st
...
+			logger.Printf("successfully wiped %s disk", target)
+
+			return nil
+		}, "resetSystemDiskPartition"
+	}
+}
+
 // ResetSystemDiskSpec represents the task to reset the system disk by spec.
-func ResetSystemDiskSpec(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
+func ResetSystemDiskSpec(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, string) {
```
smira, line 1703: we could re-use this by passing appropriate options?

utkuozdemir (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
smira: in fact, we don't need to reboot here.. we could move this whole block to be significantly earlier, in the `Initialize` sequence right before `MountStatePartition` 

if we do a full wipe, we have to reboot, if we do selective wipe and we don't wipe BOOT, we can continue on - Talos will act on it.

e.g. we could write a task which calls existing tasks as needed?

DmitriyMV: That can be arranged actually...

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
DmitriyMV: Yes, I think I have an idea how to wrap `ResetSystemDiskSpec` in `ResetSystemDiskPartition`.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/encode.go
DmitriyMV: We no longer have this string anywhere but here.

DmitriyMV:
> in any case we need a hack/release.toml section describing the change

Done, I think?

smira:
> new arg option(s) should be documented in the Kernel reference section of the docs for 1.4

I still don't see this

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -201,8 +201,8 @@ func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
 	phases := PhaseList{}
 
 	wipe := procfs.ProcCmdline().Get(constants.KernelParamWipe).First()
-	if wipe != nil && *wipe == "system" {
-		return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
+	if wipePhases := parseWipeParamString(wipe); wipePhases != nil {
```
smira, line 204: I left a comment here in the previous version that we need to consider moving this earlier to the `Initialize` sequence (e.g. before `mountStatePartition`),  so that wipe can work always, and it should be considerably easier.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -287,6 +287,20 @@ func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
...
+		return nil
+	}
+
+	phases := PhaseList{}
+
+	if *wipe == "system" {
+		return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
+	}
+
+	return phases.Append("wipeSystemDisk", ResetSystemDiskPartitions(*wipe)).Append("reboot", Reboot)
```
smira, line 301: same way I had a comment here that we can skip `Reboot` if we move it up to the `Initialize` sequence if `BOOT` is not wiped

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -287,6 +287,20 @@ func (*Sequencer) Boot(r runtime.Runtime) []runtime.Phase {
...
 
+func parseWipeParamString(wipe *string) []runtime.Phase {
+	if wipe == nil || *wipe == "" {
+		return nil
+	}
+
+	phases := PhaseList{}
+
+	if *wipe == "system" {
+		return phases.Append("wipeSystemDisk", ResetSystemDisk).Append("reboot", Reboot)
```
smira, line 298: I think instead of doing all these surgery here, we could instead add a phase which runs only if the `wipe` parameter is set, and does nothing otherwise. we could call `Reboot` from the task itself as well? Reboot returns an error and aborts the sequencer iirc

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -118,6 +118,14 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 			},
 			"dashboard",
 			StartDashboard,
+		).AppendWithDeferredCheck(
```
DmitriyMV, line 121: @smira like this?

website/content/v1.4/reference/kernel.md
```diff
@@ -207,6 +207,7 @@ Resets the disk before starting up the system.
 Valid options are:
 
 * `system` resets system disk.
+* `system:EPHEMERAL,STATE` resets ephemeral and state partitions.
```
DmitriyMV, line 210: Oh, here it is for kernel section.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -118,6 +118,14 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
...
+			func() bool {
+				wipe := procfs.ProcCmdline().Get(constants.KernelParamWipe).First()
+
+				return pointer.SafeDeref(wipe) != ""
+			},
+			"wipeDisks",
+			ResetSystemDiskPartitions,
```
DmitriyMV, line 128: @smira like this?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
DmitriyMV: Moved just above `MountStatePartition`.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
DmitriyMV: Fixed. `ResetSystemDiskPartitions` becomes kinds complex, returning task function depending on `talos.experimental.wipe`, but maybe that's okay?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
DmitriyMV: Done.

DmitriyMV:
> > new arg option(s) should be documented in the Kernel reference section of the docs for 1.4
> 
> I still don't see this

Yup, sorry, added.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
smira: yep, if that actually works. and one thing - if `BOOT` gets wiped, we should better reboot for real

smira (approved):
:cool:

Issue #6841, opened 2023-02-13
interactive console: clean up the `talosctl dashboard` code


smira:
1. refactor existing code (TUI, data collection)
2. extract as a separate package usable in `talosctl` and in Talos itself

rsmitty:
Wanted to add (hope this is the right issue), but can we add the UUID of the machine to the console view?

utkuozdemir:
> Wanted to add (hope this is the right issue), but can we add the UUID of the machine to the console view?

Sure, this is the right issue for the time being.

utkuozdemir:
Done in [f55f5df7396b7073e75267c7e10a35814f1185c9](https://github.com/siderolabs/talos/pull/6879)

Continuation in https://github.com/siderolabs/talos/issues/6752

PR #6879, opened 2023-02-22
feat: move dashboard package & run it in tty2

utkuozdemir:
Move dashboard package into a common location where both Talos and talosctl can use it.

Add support for overriding stdin, stdout, stderr and ctt in process runner.

Create a dashboard service which runs the dashboard on /dev/tty2.

Redirect kernel messages to tty1 and switch to tty2 after starting the dashboard on it.

Related to siderolabs/talos#6841, siderolabs/talos#4791.


internal/app/dashboard/main.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Main is the entrypoint into dashboard.
+func Main() {
+	if err := dashboardMain(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func dashboardMain() error {
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
```
smira, line 36: I think we should have a way to disable `SIGINT` behavior when running on a tty - same as `q` button. There should be no way to quit the dashboard :)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -115,6 +116,45 @@ func SetupLogger(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionF
...
+
+	// we need a valid fd to any tty because ioctl requires it
+	tty0, err := os.OpenFile("/dev/tty0", os.O_RDWR, 0)
+	if err != nil {
+		return err
+	}
+
+	defer tty0.Close() //nolint: errcheck
+
+	if _, _, errno := syscall.Syscall(syscall.SYS_IOCTL, tty0.Fd(), 0x5606 /* VT_ACTIVATE */, uintptr(ttyNumber)); errno != 0 {
```
frezbo, line 135: maybe add this as a comment here https://www.kernel.org/doc/html/v5.18/fb/fbcon.html :shrug:

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -105,20 +110,82 @@ func (p *processRunner) build() (cmd *exec.Cmd, logCloser io.Closer, err error)
...
+		})
+	} else {
+		cmd.Stderr = writer
+	}
+
+	if p.opts.Ctty.IsPresent() {
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Setsid:  true,
+			Setctty: true,
+			Ctty:    1,
```
smira, line 162: should be the the value of `p.opts.Ctty` ?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
utkuozdemir: This doesn't have anything to do with framebuffer, but I can add this: https://man7.org/linux/man-pages/man2/ioctl_console.2.html

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -127,16 +194,20 @@ func (p *processRunner) run(eventSink events.Recorder) error {
 		defer reaper.Stop(notifyCh)
 	}
 
-	if err = cmd.Start(); err != nil {
+	if err = cmdWrapper.cmd.Start(); err != nil {
 		return fmt.Errorf("error starting process: %w", err)
 	}
 
-	eventSink(events.StateRunning, "Process %s started with PID %d", p, cmd.Process.Pid)
+	if cmdWrapper.afterStart != nil {
```
smira, line 201: this doesn't seem to be `nil` ever, and I guess it should be called even if `Start()` fails?

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -105,20 +110,82 @@ func (p *processRunner) build() (cmd *exec.Cmd, logCloser io.Closer, err error)
...
 //nolint:gocyclo
 func (p *processRunner) run(eventSink events.Recorder) error {
-	cmd, logCloser, err := p.build()
+	cmdWrapper, err := p.build()
 	if err != nil {
 		return fmt.Errorf("error building command: %w", err)
 	}
 
-	defer logCloser.Close() //nolint:errcheck
+	if cmdWrapper.afterTermination != nil {
```
smira, line 186: doesn't seem to be `nil` ever? we could simplify?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo: ahh, i though the VT_ACTIVATE was in the link , yeh man page would work

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -105,20 +110,82 @@ func (p *processRunner) build() (cmd *exec.Cmd, logCloser io.Closer, err error)
...
 	}
 
-	cmd.Stdout = writer
-	cmd.Stderr = writer
+	var afterStartFuncs []func()
+
+	if p.opts.StdinFile != "" {
+		stdin, err := os.Open(p.opts.StdinFile)
+		if err != nil {
+			return commandWrapper{}, err
```
smira, line 118: `w` is not closed on all these exit paths

internal/app/machined/pkg/system/services/dashboard.go
```diff
@@ -0,0 +1,69 @@
...
+	return restart.New(process.NewRunner(false, &runner.Args{
+		ID:          d.ID(r),
+		ProcessArgs: []string{"/sbin/dashboard"},
+	},
+		runner.WithLoggingManager(r.Logging()),
+		runner.WithEnv([]string{"TERM=linux"}),
+		runner.WithStdinFile(tty),
+		runner.WithStdoutFile(tty),
+		runner.WithCtty(1),
+		runner.WithOOMScoreAdj(-400),
```
smira, line 65: we should drop all capabilities while starting the dashboard, I think it doesn't need any.

it makes sense to put it under a cgroup of its own, also worth checking memory consumption of the dashboard

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -115,6 +116,45 @@ func SetupLogger(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionF
 	}, "setupLogger"
 }
 
+// switchConsole switches the active console to the specified tty.
+func switchConsole(ttyNumber int) error {
```
smira, line 120: let's please move it to some `internal/pkg/console`, this file is too big already.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
utkuozdemir: they are not nicely documented, I added some links to the relevant parts of the kernel source code both here and in `redirectKernelLogs`.

internal/app/machined/pkg/system/runner/process/process.go
frezbo, line 186: don;t we set the `afterTermination` always here https://github.com/siderolabs/talos/pull/6879/files#diff-439f2aec9d3b4e0f5549166ca67c6cb8735f72c68262955e90185b85f5f41befR173?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -115,6 +116,51 @@ func SetupLogger(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionF
...
+	if err != nil {
+		return err
+	}
+
+	defer tty0.Close() //nolint: errcheck
+
+	// VT_ACTIVATE: activate the specified virtual terminal
+	// https://man7.org/linux/man-pages/man2/ioctl_console.2.html
+	// https://github.com/torvalds/linux/blob/v6.2/include/uapi/linux/vt.h#L42
+	if _, _, errno := syscall.Syscall(syscall.SYS_IOCTL, tty0.Fd(), 0x5606 /* VT_ACTIVATE */, uintptr(ttyNumber)); errno != 0 {
```
smira, line 138: let's please use constants

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: I was just trying to make it more generic, just in case we set it to be nil later. But I will simplify.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
utkuozdemir: done

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: done

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: yep, my bad, fixed.

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: done

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
utkuozdemir: done

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: yep, fixed

internal/app/machined/pkg/system/runner/process/process.go
utkuozdemir: correct, fixed.

internal/app/dashboard/main.go
utkuozdemir: done, no way to exit now :D

internal/app/machined/pkg/system/services/dashboard.go
smira: this is still not addressed - we should drop _all_ capabilities I think. the way it is done now it drops only some.

we have `cap.Known()` which returns all of them

internal/app/dashboard/main.go
```diff
@@ -0,0 +1,52 @@
...
+func Main() {
+	if err := dashboardMain(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func dashboardMain() error {
+	startup.LimitMaxProcs(constants.DashboardMaxProcs)
+
+	log.Printf("UID: %d GID: %d", syscall.Getuid(), syscall.Getgid())
```
frezbo, line 37: i guess debug statement?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -792,6 +802,19 @@ func StartMachined(_ runtime.Sequence, _ interface{}) (runtime.TaskExecutionFunc
 	}, "startMachined"
 }
 
+// StartDashboard represents the task to start dashboard.
+func StartDashboard(_ runtime.Sequence, _ interface{}) (runtime.TaskExecutionFunc, string) {
+	return func(_ context.Context, _ *log.Logger, r runtime.Runtime) error {
+		ttyNumber := 2
```
frezbo, line 808: constants maybe?

internal/app/wrapperd/main.go
```diff
@@ -78,26 +83,45 @@ func Main() {
...
 			log.Fatalf("failed to set capabilities: %v", err)
 		}
 
-		if err := iab.SetProc(); err != nil {
+		if err = iab.SetProc(); err != nil {
 			log.Fatalf("failed to apply capabilities: %v", err)
 		}
 	}
 
+	if flagPassed(flagUID) {
```
frezbo, line 105: we could just use a normal flag and by default it would have a value of zero, so if not zero set it, makes it simple

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -106,6 +109,15 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 			"earlyServices",
 			StartUdevd,
 			StartMachined,
+		).AppendWithDeferredCheck(
+			func() bool {
+				disabledStr := procfs.ProcCmdline().Get(constants.KernelParamDashboardDisabled).First()
```
smira, line 114: we need a note in `hack/release.toml` that now we switch to tty2, and that logs run on tty1, and that dashboard can be disabled by doing this and that

in the docs, kernel param reference, we should add new kernel arg

smira (approved):

internal/app/wrapperd/main.go
utkuozdemir: done

hack/release.toml
```diff
@@ -77,6 +77,17 @@ machine:
         title = "Machine Configuration"
         description="""\
 Strategic merge config patches correctly support merging `.vlans` sections of the network interface.
+"""
+
+    [notes.dashboard]
```
frezbo, line 82: let's keep it disabled by default for sbc's

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
utkuozdemir: 👍 done, also disabled it by default on SBCs.

internal/app/machined/pkg/system/services/dashboard.go
utkuozdemir: done

Dockerfile
```diff
@@ -476,6 +476,8 @@ RUN ln /rootfs/sbin/init /rootfs/sbin/poweroff
 RUN chmod +x /rootfs/sbin/poweroff
 RUN ln /rootfs/sbin/init /rootfs/sbin/wrapperd
 RUN chmod +x /rootfs/sbin/wrapperd
+RUN ln /rootfs/sbin/init /rootfs/sbin/dashboard
```
frezbo, line 479: this should be for arm64 too

frezbo (approved):

Dockerfile
utkuozdemir: yes, missed that, now added

frezbo (approved):

Issue #6815, opened 2023-02-03
allow wiping user disks in Reset API


smira:
If enabled, wipes all writeable block devices on Reset.

It should ignore network blockdevices (e.g. iSCSI).

PR #6881, opened 2023-02-22
feat: introduce new flag in reset API that makes Talos reset user disks

Unix4ever:
Fixes: https://github.com/siderolabs/talos/issues/6815

Additionally, make it possible to run reset in maintenance mode: to enable a way for resetting system disk and remove all traces of Talos from it.

The new reset flow works in a separate sequence, changed disk probe lookup to check the boot partition instead of the ephemeral one.


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1593,7 +1594,13 @@ func ResetSystemDisk(seq runtime.Sequence, data interface{}) (runtime.TaskExecut
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
 		var dev *blockdevice.BlockDevice
 
-		dev, err = blockdevice.Open(r.State().Machine().Disk().Device().Name())
+		disk := r.State().Machine().Disk(disk.WithPartitionLabel(constants.BootPartitionLabel))
```
Unix4ever, line 1597: Changed reset to lookup the system disk by the `BootPartitionLabel`. It feels like this is a more appropriate way to do it.
As another option we can do lookup by ephemeral first then fall back to boot.

smira: I'm not sure what the consequences would be, but my guess we should always reset by EPHEMERAL, as this is the most critical one used everywhere. Imagine having an USB stick plugged in with no EPHEMERAL in it, but BOOT.

In maintenance mode, we should figure out if the Talos is installed, and if we do, we know the system disk.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1604,6 +1611,31 @@ func ResetSystemDisk(seq runtime.Sequence, data interface{}) (runtime.TaskExecut
 	}, "resetSystemDisk"
 }
 
+// ResetUserDisks represents the task to reset the user disks.
+func ResetUserDisks(seq runtime.Sequence, data interface{}) (runtime.TaskExecutionFunc, string) {
+	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
+		for _, disk := range r.Config().Machine().Disks() {
```
smira, line 1617: this looks like not what we were discussing - we should be resetting all disks but Talos system disk, like Sidero Agent does.

users might be running Rook/Ceph, and `Config().Machine().Disks()` will be empty

internal/app/maintenance/server/server.go
```diff
@@ -201,3 +201,36 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (reply
 
 	return reply, nil
 }
+
+// Reset resets the node.
+//
+//nolint:gocyclo
+func (s *Server) Reset(ctx context.Context, in *machine.ResetRequest) (reply *machine.ResetResponse, err error) {
+	actorID := uuid.New().String()
```
smira, line 209: this should be limited to SideroLink IPs, like e.g. `Version()` API

cmd/talosctl/cmd/talos/reset.go
```diff
@@ -113,6 +179,9 @@ func resetGetActorID(ctx context.Context, c *client.Client, req *machineapi.Rese
 func init() {
 	resetCmd.Flags().BoolVar(&resetCmdFlags.graceful, "graceful", true, "if true, attempt to cordon/drain node and leave etcd (if applicable)")
 	resetCmd.Flags().BoolVar(&resetCmdFlags.reboot, "reboot", false, "if true, reboot the node after resetting instead of shutting down")
+	resetCmd.Flags().BoolVar(&resetCmdFlags.insecure, "insecure", false, "reset using the insecure (encrypted with no auth) maintenance service")
+	resetCmd.Flags().Var(&resetCmdFlags.wipeMode, "wipe-mode", "disks reset mode")
```
smira, line 183:
```suggestion
	resetCmd.Flags().Var(&resetCmdFlags.wipeMode, "wipe-mode", "disk reset mode")
```

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -610,7 +611,50 @@ func (s *Server) Reset(ctx context.Context, in *machine.ResetRequest) (reply *ma
...
+
+		disks := slices.ToMap(diskList, func(disk *bddisk.Disk) (string, *bddisk.Disk) {
+			return disk.DeviceName, disk
+		})
+
+		systemDisk := s.Controller.Runtime().State().Machine().Disk()
+
+		// validate input
+		for _, deviceName := range in.GetUserDisksToWipe() {
+			if disks == nil {
```
smira, line 634: looks like this check is not needed, reading from `nil` map is fine

internal/app/maintenance/server/server.go
```diff
@@ -201,3 +204,64 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (reply
...
+
+	var dev *blockdevice.BlockDevice
+
+	disk := s.controller.Runtime().State().Machine().Disk(disk.WithPartitionLabel(constants.BootPartitionLabel))
+
+	if disk == nil {
+		return nil, fmt.Errorf("reset failed: Talos is not installed")
+	}
+
+	dev, err = blockdevice.Open(disk.Device().Name())
```
smira, line 232: should we also support user disks here? this might be needed to get the machine into a "clean" state

Unix4ever: good idea, added!

smira (approved):
:cool:

PR #6893, opened 2023-02-28
chore: handle grub option - "wipe"

DmitriyMV:
This PR ensures that we can handle third grub option - "wipe". We will use it in 1.4.

For #6842

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/constants.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/constants.go
@@ -13,6 +13,9 @@ const (
 	// BootB is a bootloader label.
 	BootB BootLabel = "B"
 
+	// BootReset is a bootloader label.
+	BootReset BootLabel = "Reset"
+
 	// ConfigPath is the path to the grub config.
 	ConfigPath = constants.BootMountPoint + "/grub/grub.cfg"
 )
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/constants.go
```diff
@@ -13,6 +13,9 @@ const (
 	// BootB is a bootloader label.
 	BootB BootLabel = "B"
 
+	// BootWipe is a bootloader label.
+	BootWipe BootLabel = "Wipe"
```
smira, line 17: My only concern is the name of the label, as it starts the user-visible string in GRUB menu.

What will be the final menu entry?

```
Reset Talos installation and return to maintenance mode
```

It will probably start with `Reset` ? I think we don't use the word `Wipe` often, and here we talk about a partial "wipe"

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/testdata/grub_parse_test.cfg
```diff
@@ -20,3 +20,10 @@ menuentry "B - v2" {
   linux /B/vmlinuz cmdline B
   initrd /B/initramfs.xz
 }
+
+menuentry "Wipe system disks" {
```
frezbo, line 24: as a beginner user this might be still confusing, most users expect reset to return to maintenance mode , maybe `Reset to maintenance` still doesn't sound clear I guess

DmitriyMV: This should't even be there -`Reset Talos installation and return to maintenance mode` is a new text. Updated PR.

utkuozdemir (approved):

frezbo (approved):

PR #6895, opened 2023-02-28
fix: remove "defer" leak in ResetUserDisks

DmitriyMV:
Also, print error if we failed to close the device.


frezbo (approved):

Unix4ever (approved):
thanks!

smira:
let's please hold on merging until #6894 is merged

PR #6896, opened 2023-02-28
fix: improve etcd leave on reset process

smira:
When removing a member from `etcd`, the server does a pre-check to make sure the member is connected to a quorum of other members, and the remove request might fail. Add a retry to wait for the etcd to be fully connected before giving up, as some parts of the reset flow alrady ran.

Also fix an issue which appears in the integration test, when `reset` is called early in the boot sequence when local etcd hasn't started fully yet.

### DIFF
--- a/internal/pkg/etcd/local.go
+++ b/internal/pkg/etcd/local.go
@@ -7,6 +7,7 @@ package etcd
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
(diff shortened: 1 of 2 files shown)


internal/pkg/etcd/local.go
```diff
@@ -16,10 +17,14 @@ import (
...
 func GetLocalMemberID(ctx context.Context, s state.State) (uint64, error) {
-	member, err := safe.ReaderGet[*etcd.Member](
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Minute)
+	defer cancel()
+
+	member, err := safe.StateWatchFor[*etcd.Member](
 		ctx,
 		s,
 		etcd.NewMember(etcd.NamespaceName, etcd.LocalMemberID).Metadata(),
+		state.WithEventTypes(state.Created),
```
frezbo, line 27: :+1:

frezbo (approved):

PR #6900, opened 2023-03-01
fix: redo assertHostnames in HostnameMergeSuite.TestMerge

DmitriyMV:
Use `rtestutils.AssertResources` for hostnames test.


DmitriyMV:
I updated as many tests as I could, but there is just too many of them for one go.

internal/app/machined/pkg/controllers/network/rtestutils/assert.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// AssertResources asserts on a resource list.
+func AssertResources[R ResourceWithRD](
+	ctx context.Context,
+	t *testing.T,
+	st state.State,
+	ids []resource.ID,
+	assertionFunc func(r R, assertion *assert.Assertions),
+	opts ...Option,
```
DmitriyMV, line 38: I'm not sure if we have any other places that require us to pass Namespace separately, and those I don't know if it make sense to move it into runtime project.

smira: I think certainly we shouldn't have a copy here

DmitriyMV: Okay - moved to https://github.com/cosi-project/runtime/pull/236

PR #6904, opened 2023-03-02
feat: add cilium e2e tests

frezbo:
Add cilium e2e tests. The existing cilium check was very old, update to latest cilium version and also add a test for KPR strict mode.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -416,17 +416,28 @@ local integration_provision_tests_track_2 = Step('provision-tests-track-2', priv
 local integration_extensions = Step('e2e-extensions', target='e2e-qemu', privileged=true, depends_on=[extensions_patch_manifest], environment={
   SHORT_INTEGRATION_TEST: 'yes',
   QEMU_MEMORY_WORKERS: '3072',
-  WITH_CONFIG_PATCH_FILE_WORKER: '_out/extensions-patch.json',
+  WITH_CONFIG_PATCH_WORKER: '@_out/extensions-patch.json',
   WITH_TEST: 'run_extensions_test',
   IMAGE_REGISTRY: local_registry,
 });
-local integration_cilium = Step('e2e-cilium-1.9.10', target='e2e-qemu', privileged=true, depends_on=[integration_extensions], environment={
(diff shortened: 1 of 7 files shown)


.drone.jsonnet
```diff
@@ -420,10 +420,17 @@ local integration_extensions = Step('e2e-extensions', target='e2e-qemu', privile
   WITH_TEST: 'run_extensions_test',
   IMAGE_REGISTRY: local_registry,
 });
-local integration_cilium = Step('e2e-cilium-1.9.10', target='e2e-qemu', privileged=true, depends_on=[integration_extensions], environment={
+local integration_cilium = Step('e2e-cilium', target='e2e-qemu', privileged=true, depends_on=[integration_extensions], environment={
```
smira, line 423: I feel like we should probably split `integration-misc` a bit, as it now might run for 1.5 hours with the `extensions` build... nothing pressing, probably `e2e-extensions` should be its own pipeline, as it got significantly more complex?

frezbo: yes, makes sense, should cilium also be it's own pipeline, also would make my testing easier :sweat_smile:

hack/test/e2e-qemu.sh
```diff
@@ -129,6 +129,16 @@ case "${WITH_CONFIG_PATCH_FILE_WORKER:-false}" in
     ;;
 esac
 
+case "${WITH_SKIP_BOOT_PHASE_FINISHED_CHECK:-no}" in
+  # using arrays here to preserve spaces properly in WITH_CONFIG_PATCH_FILE
```
smira, line 133: comment here doesn't match?

smira (approved):

.drone.jsonnet
smira: as I was testing this, I commented out everything but Cilium from the `integration-misc`, or make it the first one - the order doesn't matter

frezbo: > as it now might run for 1.5 hours with the extensions build

extensions e2e is a different pipeline, it's only part of intergration-misc as part of cron, should we change it to a separate cron?

smira: yes, probably

PR #6908, opened 2023-03-03
feat: add pod ip to kube-proxy spec

TimJones:
Exposes the pod IP as the `POD_IP` environment variable via the downward API in the kube-proxy pod for use in e.g. metrics-bind-addr.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/templates.go
+++ b/internal/app/machined/pkg/controllers/k8s/templates.go
@@ -134,6 +134,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
+          - name: POD_IP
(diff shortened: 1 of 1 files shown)


TimJones:
Should it also be added/created when upgrading from older version i.e. in the daemonsetUpgrader?

utkuozdemir (approved):

frezbo:
> Should it also be added/created when upgrading from older version i.e. in the daemonsetUpgrader?

`talosctl upgrade-k8s` should apply the new changes

smira:
> Should it also be added/created when upgrading from older version i.e. in the daemonsetUpgrader?

@TimJones it will be handled by the new upgrade process now (I have outstanding PR for it)

PR #6909, opened 2023-03-03
ci: move blob storage to azure for builds

rsmitty:
This PR moves blob storage to azure.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -270,13 +270,16 @@ local save_artifacts = {
   image: 'autonomy/build-container:latest',
   pull: 'always',
   environment: {
-    AWS_ACCESS_KEY_ID: { from_secret: 'rook_access_key_id' },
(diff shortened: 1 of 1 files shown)


.drone.jsonnet
```diff
@@ -267,16 +267,15 @@ local push_latest = {
 
 local save_artifacts = {
   name: 'save-artifacts',
-  image: 'autonomy/build-container:latest',
+  image: 'mcr.microsoft.com/azure-cli:latest',
```
frezbo, line 270: doesn't pur image already has azure-cli?

rsmitty: yeah it does. just missed it when I went looking earlier. thx.

rsmitty:
needs https://github.com/siderolabs/extensions/pull/128 merged first.

frezbo (approved):

PR #6913, opened 2023-03-05
sets default image, validates empty image

### LINKED ISSUE #6912: talosctl: upgrade image is not locally validated
## Bug Report

If you run `talosctl upgrade` without specifying an image, `talosctl` will happy run off and try to upgrade to the image "".
At a minimum, we should check for an empty image name locally, first.

Ideally, too, the `talosctl` version should have its default image set to the same version of Talos itself.
This would make upgrades much simpler:  upgrade `talosctl` then run upgrade.  No need to continually specify the image URL.

Ulexus:
Adds a default image URL and ensures that an empty image URL is not sent when calling `talosctl upgrade`.

Fixes #6912

# Pull Request

## Acceptance


cmd/talosctl/cmd/talos/upgrade.go
```diff
@@ -138,7 +144,7 @@ func upgradeGetActorID(ctx context.Context, c *client.Client) (string, error) {
 }
 
 func init() {
-	upgradeCmd.Flags().StringVarP(&upgradeCmdFlags.upgradeImage, "image", "i", "", "the container image to use for performing the install")
+	upgradeCmd.Flags().StringVarP(&upgradeCmdFlags.upgradeImage, "image", "i", fmt.Sprintf("%s:%s", constants.DefaultInstallerImage, version.Tag), "the container image to use for performing the install")
```
frezbo, line 147: this can be used instead:
```suggestion
	upgradeCmd.Flags().StringVarP(&upgradeCmdFlags.upgradeImage, "image", "i", fmt.Sprintf("%s/%s/installer:%s", gendata.ImagesRegistry, gendata.ImagesUsername, gendata.VersionTag), "the container image to use for performing the install")
```

gendata coming from: `"github.com/siderolabs/talos/pkg/machinery/gendata"`

Ulexus: Ah, nice!

frezbo (approved):

Ulexus:
Hmm.
This is a bit of a problem with the docs:  the docs are generated on each commit, and with this change, the default installer image will necessarily always be different.
Thus, the dirty check will _always_ fail.

Is there a known workaround for this problem?

Ulexus:
Changed to implicitly calculate the default image name, which is functionally identical but less ideal from the end-user perspective.

smira:
> Is there a known workaround for this problem?

we have this problem in one more place, but I think the tag is abbreviated, so it only changes after a release (when the new tag is pushed)

smira:
I refactored code a bit, so now default value is being used as intended.

frezbo (approved):

PR #6918, opened 2023-03-06
fix: get running and newest container logs

DmitriyMV:
During container restart, there is small window when two containers exist at the same time, exited and running.
We want to grab the running one. But regardless of this, we also want to grab the newest one, so even if there are two exited
containers, we will get the logs for the newest.

Co-authored-by: Utku Ozdemir <utku.ozdemir@siderolabs.com>


utkuozdemir (approved):

smira:
I don't think this is enough, sometimes you want the _exited_ one, as it prints the error right before exiting, so better to give user a way to see both 'new' and 'old' logs. One of the ideas might be to use the flag (like `kubectl logs -p`), or provide different IDs for old and new containers.

DmitriyMV:
Superseded by #6926

PR #6920, opened 2023-03-06
chore: support running eudev in debug mode

frezbo:
Support running eudev in debug mode.

### DIFF
--- a/internal/app/machined/pkg/system/services/udevd.go
+++ b/internal/app/machined/pkg/system/services/udevd.go
@@ -63,19 +63,25 @@ func (c *Udevd) DependsOn(r runtime.Runtime) []string {
 
 // Runner implements the Service interface.
 func (c *Udevd) Runner(r runtime.Runtime) (runner.Runner, error) {
-	// Set the process arguments.
-	args := &runner.Args{
-		ID: c.ID(r),
-		ProcessArgs: []string{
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/system/services/udevd.go
```diff
@@ -63,21 +63,27 @@ func (c *Udevd) DependsOn(r runtime.Runtime) []string {
...
 	if r.Config() != nil {
 		debug = r.Config().Debug()
 	}
 
+	processArgs := []string{
+		"/sbin/udevd",
+		"--resolve-names=never",
+	}
+
+	if debug {
```
smira, line 77: don't we start it before the machine config is acquired?

frezbo: yes, but once machine is up, it should have the debug mode, so next boot should provide this.

smira: but still we start before `loadConfig`, so unless a service is restarted, I don't see how it would work :shrug:

frezbo: ahh I see the problem, I'll close for now then :sweat_smile: , just wanted to see if it makes easier for others to debug, I usually build a new installer with the `--debug` added :sweat_smile:

PR #6926, opened 2023-03-07
chore: add container id to `talosctl -k containers` and `talosctl -k logs`

### LINKED ISSUE #6886: `talosctl logs` Prints logs from random dead pod
### Description
`talosctl logs` Prints logs from random dead pod when querying a crash looping static pod

Running the same command will give different historical logs (detectable by timestamps). Adding `--tail` or `-f` has the same behaviour. Seems like `-f` also does not stream log after pod restart.

### Logs
Not sure which logs to collect.

### Environment

    ```
  Client:
          Tag:         v1.3.5
          SHA:         03edf8c1
          Built:       
          Go version:  go1.19.6
          OS/Arch:     darwin/amd64
  Server:
          NODE:        192.168.2.205
          Tag:         v1.3.5
          SHA:         03edf8c1
          Built:       
          Go version:  go1.19.6
          OS/Arch:     linux/amd64
          Enabled:     RBAC
  ```
  N/A (playing with kube-apiserver, which is crash looping)
- Platform:
  kubevirt / OKD

DmitriyMV:
This PR adds first 12 symbols from container ID and adds them to `talosctl -k containers` each container output.
That way we can ensure that we get the logs from proper container even if there is a newer one.

Closes #6886

Co-authored-by: Utku Ozdemir <utku.ozdemir@siderolabs.com>


utkuozdemir:
Does `containers -k` command list the Exited containers as well? For example, it should now list 2 containers for a static pod that has crashed once - the Exited one and the Running one. Otherwise I cannot discover the ID of the crashed one so I cannot troubleshoot.

DmitriyMV:
@utkuozdemir 
> Does containers -k command list the Exited containers as well? 

Yes.

> For example, it should now list 2 containers for a static pod that has crashed at least once - the Exited one and the Running one. 

It showed both containers before too. Just with the same display name.

> Otherwise I cannot discover the ID of the crashed one so I cannot troubleshoot.

`talosctl containers -k` should show you both running and exited container, so you can access both of them.

DmitriyMV:
> Also, IMO, --previous would be a better user experience - just like kubectl does.

Yes, but it will also require us add the obtaining of exited containers into `talosctl support`.

smira (approved):

PR #6930, opened 2023-03-09
fix: upgrade-k8s to flag should not be required since there is a default

wind0r:
# Pull Request
## What? (description)
Make the --to flag optional since there is default value for that flag.

## Why? (reasoning)

talosctl upgrade-k8s says the --to flag has a default, still when calling without the --to flag the users gets told to that the required flag.

A different solution may would be to remove the default value.

```
talosctl upgrade-k8s -n server
required flag(s) "to" not set

Usage:
  talosctl upgrade-k8s [flags]

Flags:
      --dry-run           skip the actual upgrade and show the upgrade plan instead
      --endpoint string   the cluster control plane endpoint
      --from string       the Kubernetes control plane version to upgrade from
  -h, --help              help for upgrade-k8s
      --to string         the Kubernetes control plane version to upgrade to (default "1.26.1")
      --upgrade-kubelet   upgrade kubelet service (default true)

Global Flags:
      --cluster string       Cluster to connect to if a proxy endpoint is used.
      --context string       Context to be used in command
  -e, --endpoints strings    override default endpoints in Talos configuration
  -n, --nodes strings        target the specified nodes
      --talosconfig string   The path to the Talos configuration file. Defaults to 'TALOSCONFIG' env variable if set, otherwise '$HOME/.talos/config' and '/var/run/secrets/talos.dev/config' in order.
```

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/upgrade-k8s.go
+++ b/cmd/talosctl/cmd/talos/upgrade-k8s.go
@@ -12,7 +12,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
-	"github.com/siderolabs/talos/pkg/cli"
 	"github.com/siderolabs/talos/pkg/cluster"
 	k8s "github.com/siderolabs/talos/pkg/cluster/kubernetes"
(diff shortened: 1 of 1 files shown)


smira:
please add the DCO (`git commit --amend -s`), and some commit body, and it should be gtg, thanks

PR #6934, opened 2023-03-10
fix: easier upgrade to Talos 1.3 with custom CRI config

### LINKED ISSUE #6882: Sponsored: provide Talos 1.2.x release which accepts new containerd customization path
Problem: https://www.talos.dev/v1.3/talos-guides/upgrading-talos/#before-upgrade-to-v132

Provide another patch release of Talos 1.2.x which will accept (but ignore) the path `/etc/cri/conf.d/20-customization.part,` so that the upgrade flow is the following:
 
* upgrade to the latest 1.2.x
* patch machine config to provide same customizations for containerd under old `/var/cri/conf.d` path and new `/etc/cri/conf.d/20-customization.part` path
* on Talos 1.2.x, only `/var/cri/conf.d path` will be used
* upgrade to Talos 1.3.x
* after an upgrade `/var/cri/conf.d` path will be ignored, and new path `/etc/cri/conf.d` will be used
* drop old override under `/var/cri/conf.d`

smira:
This makes Talos 1.2 accept but ignore Talos 1.3 specific `/etc/cri/conf.d/20-customization.part` file, so that new config can be injected into running Talos 1.2 before upgrading to Talos 1.3.

Talos 1.3 will ignore the customization under `/var`, so it can be safely removed after the upgrade.

Fixes #6882

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -1080,6 +1080,10 @@ func WriteUserFiles(seq runtime.Sequence, data interface{}) (runtime.TaskExecuti
 				continue
 			}
 
+			if f.Path() == filepath.Join("/etc", constants.CRICustomizationConfigPart) {
+				continue
+			}
+
 			// Determine if supplied path is in /var or not.
 			// If not, we'll write it to /var anyways and bind mount below
 			p := f.Path()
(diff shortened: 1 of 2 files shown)


smira:
Tested the upgrade flow manually with the following config:

```yaml
machine:
    files:
      - content: |
          [metrics]
            address = "0.0.0.0:11234"
        path: /etc/cri/conf.d/20-customization.part
        op: create
      - content: |
          [metrics]
            address = "0.0.0.0:11234"        
        path: /var/cri/conf.d/metrics.toml
        op: create
```

Before and after upgrade metrics were available.

frezbo (approved):

PR #6946, opened 2023-03-13
docs: fix Talos version to use template

### LINKED ISSUE #6944: Docs reference a very old version (which doesn't actually work)
Took me a while to figure out why my testing cluster isn't coming up online. I've been following the instructions here:
https://www.talos.dev/v1.3/talos-guides/install/cloud-platforms/hetzner/
and they reference version 0.13
```variable "talos_version" {
  type    = string
  default = "v0.13.0"
}
```
Which doesn't really work
Maybe other people new with talos (trying it out like myself) could use from an update in that doc
Thanks

smira:
Fixes #6944


cristianrat:
very quick
:)
thank you

PR #6949, opened 2023-03-13
feat: update Flannel to v0.21.3

smira:
See https://github.com/flannel-io/flannel/releases/tag/v0.21.3


pkg/flannel/template.go
```diff
@@ -32,6 +32,13 @@ rules:
   - nodes/status
   verbs:
   - patch
+- apiGroups:
```
smira, line 35: this is auto-generated

frezbo (approved):

PR #6951, opened 2023-03-13
chore: update CI to tag azure buckets

rsmitty:
This PR updates CI to remove the immutability policy and tags the azure "containers" (aka buckets) with a ci=true tag. This will allow us to handle the deletion of buckets with the cloud-cleaner app. Also adds the override flag to batch upload, as that was actually failing silently on subsequent runs of the CI.


smira (approved):

smira (commented):
we will need same change for Kres as well (to update Omni)

Unix4ever (approved):

Issue #4721, opened 2021-12-22
Sponsored: configurable ephemeral partition


andrewrynhard:
### Description

We would like to be able to choose the size and the location of the ephemeral and maybe the /var partition
At this time we have recycled ESX machines, booting on SD card, with very big disks. I tried a few time to use the SD Card for Talos which is booting like the ESX kernel in RAM for the system partition
which is a great features of talos but not so great feature when it tried to run the container on the same SD Card which is really not meant for heavy IO use.
A way to give the talos an alternate partition to use for ephemeral would be a nice to have feature.

andrewrynhard:
Ideas:

  - It _might_ be possible to mount a user disk under somewhere under `/var`. We should test that and fix it if not.

We need more details on this issue. We will follow up later.

PR #6952, opened 2023-03-14
feat: support ephemeral partition size

jrcichra:
Allow ephemeral partition size to be specified in configuration. Users can limit the size of their ephemeral partition and partition the rest of the disk for other tools such as rook OSDs.

# Pull Request

Partially addresses: https://github.com/siderolabs/talos/issues/4721

This solution was what worked best for my use case (running a Rook OSD on the same drive as my Talos install) by making the partition size and subsequent partition resizing of the ephemeral partition configurable.

The default behavior remains the same, where the ephemeral partition will resize to the maximum remaining size of the drive.

## What? (description)
Allow ephemeral partition size to be specified in configuration. 

## Why? (reasoning)
Users can limit the size of their ephemeral partition and partition the rest of the disk for other tools such as rook OSDs.
## Acceptance

### DIFF
--- a/cmd/installer/cmd/install.go
+++ b/cmd/installer/cmd/install.go
@@ -78,6 +78,8 @@ func runInstallCmd() (err error) {
 		if config.Machine().Install().LegacyBIOSSupport() {
 			options.LegacyBIOSSupport = true
 		}
+
+		options.EphemeralSize = config.Machine().Install().EphemeralSize()
 	}
 
 	return install.Install(p, seq, options)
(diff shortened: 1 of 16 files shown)


andrewrynhard:
Hello @jrcichra! Appreciate the time you out into this. Unfortunately we have a different design we are working on to support something like this. I love contributions but we will have to close this one. Thanks again for the time you put into this. We should have something official in the Talos 1.5 release maybe.

PR #6957, opened 2023-03-14
refactor: change the interface of META

smira:
Use a global instance, handle loading/saving META in global context.

Deprecate legacy syslinux ADV, provide an easier interface for consumers.

Expose META as resources.

Fix the bootloader revert process (it was completely broken for quite a while :sad:).

This is a first step which mostly does preparation work, real changes will come in the next PRs:

* add APIs to write to META
* consume META keys for platform network config for `metal`
* custom key for URL `${code}`
* writing meta during the Talos image creation


cmd/installer/pkg/install/install.go
```diff
@@ -323,20 +323,19 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
...
+		var metaState *meta.Meta
 
-		if meta, err = bootloader.NewMeta(); err != nil {
+		if metaState, err = meta.New(context.Background(), nil); err != nil {
 			return err
 		}
 
-		//nolint:errcheck
-		defer meta.Close()
+		var ok bool
```
frezbo, line 332: why `ok` as a var definition instead of `inline` with `if ok, err :=`

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -168,6 +171,12 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
 			).Append(
 				"install",
 				Install,
+			).Append(
+				"loadMeta",
```
frezbo, line 175:
```suggestion
				"meta",
```

as used everywhere else

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -2409,6 +2397,25 @@ func CleanupLegacyStaticPodFiles(runtime.Sequence, any) (runtime.TaskExecutionFu
...
 
+// ReloadMeta reloads META partition after disk mount, installer run, etc.
+func ReloadMeta(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
+	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
+		err := r.State().Machine().Meta().Reload(ctx)
+		if os.IsNotExist(err) {
+			err = nil
+		}
+
+		return err
```
frezbo, line 2408:
```suggestion
		if !os.IsNotExist(err) {
			return err
		}

		return nil
```



seems easier to read

internal/app/machined/revert.go
```diff
@@ -0,0 +1,114 @@
...
+func revertBootloader() {
+	if err := revertBootloadInternal(); err != nil {
+		log.Printf("failed to revert bootloader: %s", err)
+	}
+}
+
+//nolint:gocyclo
+func revertBootloadInternal() error {
+	metaState, err := meta.New(context.Background(), nil)
+	if os.IsNotExist(err) {
```
frezbo, line 29:
```suggestion
	if !os.IsNotExist(err) {
		return err
	}
```

isn't this easier?

internal/pkg/meta/constants.go
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package meta
+
+const (
+	// Upgrade is the upgrade tag.
+	Upgrade = iota + 6
```
frezbo, line 9: what's the significance of the number `6`, or just that the lower numbers are used anywhere else?

internal/pkg/meta/meta.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	defer dev.Close() //nolint:errcheck
+
+	return dev.PartPath(constants.MetaPartitionLabel)
+}
+
+// Reload refreshes the META from the disk.
+func (meta *Meta) Reload(ctx context.Context) error {
+	path, err := meta.getPath()
```
frezbo, line 99: do we need locking here, if some controller uses this (haven;t looked at the whole changes yet, just thinking if controllers use this?)

frezbo (approved):

internal/pkg/meta/constants.go
smira: it came as I split the core ADV code and the constants used in Talos itself.

this way the constants keep same values as before, but we don't have constants which make zero sense for Talos (inherited from syslinux days)

internal/pkg/meta/meta.go
smira: controllers do not use it... but good point on locking... probably to be on the safe side I would rather use locks

internal/app/machined/revert.go
smira: this also works, but needs a check for `err != nil`, as `nil` will also satisfy the `!os.IsNotExist(err)` condition

smira: and here it is a bit different - we want to skip the revert if the META is not found, while we still return an error if we fail to read META

frezbo: > nil will also satisfy the !os.IsNotExist(err) condition

interesting, TIL

internal/pkg/meta/meta.go
smira: added locks

internal/app/machined/revert.go
smira: I changed the code a bit to make it easier to read

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: reworked similar to this

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
smira: fixed

PR #6958, opened 2023-03-15
feat: implement APIs to write to META

smira:
This allows to put keys to META partition.

META contents can be viewed with `talosctl get metakeys`.

There is not real usecase for it yet, but the next PRs will introduce two special keys which can be written:

* platform network config for `metal`
* `${code}` variable


website/content/v1.4/reference/cli.md
frezbo, line 1: do we need the `talosctl meta` to be public?

frezbo (approved):

website/content/v1.4/reference/cli.md
smira: I don't think it hurts, as we have APIs, we should have a way to interact with it?

utkuozdemir (approved):

Issue #6956, opened 2023-03-14
Talos 1.3.5 LinkSpecController segfaulting following Sidero Metal Deployment


AlvesAlexander:
### Description
Seeing the below segfault occur after deploying to a rack of 3x Dell PowerEdge r740 machines
### Logs
```
172.18.0.16: kern:    info: [2023-03-14T14:40:01.674091594Z]: init[1779]: segfault at 8 ip 0000000002617d86 sp 000000c001ff0c48 error 4 in init[400000+2510000]
172.18.0.16: kern:    info: [2023-03-14T14:40:01.805998594Z]: Code: 01 00 00 00 eb 52 48 8d 05 37 08 52 01 48 8b 9c 24 c8 0b 00 00 e8 ea 9c 49 fe 48 8b 8c 24 30 02 00 00 48 8b 51 10 48 8b 52 60 <48> 8b 72 08 48 8b 12 48 8b 58 18 48 39 70 20 74 07 b8 01 00 00 00
172.18.0.16: user: warning: [2023-03-14T14:40:02.057627594Z]: [talos] controller failed {"component": "controller-runtime", "controller": "network.LinkSpecController", "error": "controller \x5c"network.LinkSpecController\x5c" panicked: runtime error: invalid memory address or nil pointer dereference\x5cn\x5cngoroutine 488 [running]:\x5cnruntime/debug.Stack()\x5cn\x5ct/toolchain/go/src/runtime/debug/stack.go:24 +0x65\x5cngithub.com/cosi-project/runtime/pkg/controller/runtime.(*adapter).runOnce.func2()\x5cn\x5ct/.cache/mod/github.com/cosi-project/runtime@v0.2.1/pkg/controller/runtime/adapter.go:401 +0x5d\x5cnpanic({0x2d786e0, 0x57da970})\x5cn\x5ct/toolchain/go/src/runtime/panic.go:884 +0x212\x5cngithub.com/siderolabs/talos/internal/app/machined/pkg/controllers/network.(*LinkSpecController).syncLink(0x22?, {0x3b52c38, 0xc0009f2280}, {0x3b70900, 0xc000402080}, 0x1?, 0xc000b2b880, 0xc001ff1a38, 0xc001ff1a20, 0xc00125b500)\x5cn\x5ct/src/internal/app/machined/pkg/controllers/network/link_spec.go:227 +0x386\x5cngithub.com/siderolabs/talos/internal/app/machined/pkg/controllers/...
```
### Environment

[...]

PR #6959, opened 2023-03-15
fix: nil pointer exception in syncLink

DmitriyMV:
If link has no `Info` field we can't do anything meaningful, so we'll just log and skip.
Also fix race in test.

For #6956

### DIFF
--- a/internal/app/machined/pkg/controllers/network/link_spec.go
+++ b/internal/app/machined/pkg/controllers/network/link_spec.go
@@ -225,6 +225,16 @@ func (ctrl *LinkSpecController) syncLink(ctx context.Context, r controller.Runti
 		if existing != nil && link.TypedSpec().Logical {
 			replace := false
 
+			if existing.Attributes.Info == nil {
+				logger.Warn("requested logical link has no info, skipping sync",
+					zap.String("name", existing.Attributes.Name),
+					zap.Stringer("type", nethelpers.LinkType(existing.Type)),
+					zap.Uint32("index", existing.Index),
+				)
+
+				return nil
+			}
+
 			// if type/kind doesn't match, recreate the link to change it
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/network/link_spec.go
```diff
@@ -219,6 +219,16 @@ func (ctrl *LinkSpecController) syncLink(ctx context.Context, r controller.Runti
 	case resource.PhaseRunning:
 		existing := findLink(*links, link.TypedSpec().Name)
 
+		if existing != nil && existing.Attributes.Info == nil {
```
frezbo, line 222: would `existing.Attributes` ever be nil?

frezbo: nevermind, saw the comment in the other file

frezbo (approved):

internal/app/machined/pkg/controllers/network/link_spec.go
DmitriyMV: That's a good question: `findLink` actually looks into `link.Attributes.Name` and there is no nil pointer panic there in #6956. But `LinkMessage.UnmarshalBinary` doesn't make any promises :)

smira (approved):

smira:
hold on, why is this against `release-1.3`, it should be against `main` ?

smira:
> hold on, why is this against `release-1.3`, it should be against `main` ?

we try to avoid direct pushes to release branches unless it's specific to this release - this problem is not obviously specific to 1.3.x, so we should PR against `main`, and we will backport in the next release 1.3.7

DmitriyMV:
@smira done

smira (approved):

Issue #6963, opened 2023-03-15
`talosctl gen secrets` override existing file


frezbo:
## Bug Report

`talosctl gen secrets` override existing file

andrewrynhard:
Related: #5711

frezbo:
That should be fixed now

PR #6964, opened 2023-03-15
fix: add `--force` flag for `talosctl gen`

frezbo:
Error out if file(s) already exists and warn user to use
`--force` to overwrite.

Fixes: #6963

### DIFF
--- a/cmd/talosctl/cmd/mgmt/gen/config.go
+++ b/cmd/talosctl/cmd/mgmt/gen/config.go
@@ -70,7 +70,6 @@ var genConfigCmdFlags struct {
 	withDocs                bool
 	withClusterDiscovery    bool
 	withKubeSpan            bool
-	force                   bool
 	withSecrets             string
 }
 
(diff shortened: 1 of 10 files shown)


andrewrynhard:
does it make sense to have `--force` be part of `gen` and not `gen secrets` _and_ `gen config`?

cmd/talosctl/cmd/mgmt/gen/config.go
```diff
@@ -280,9 +279,14 @@ func validateFlags() error {
 	return err
 }
 
+// nolint:gocyclo
 func writeConfigBundle(configBundle *bundle.ConfigBundle, outputPaths configOutputPaths, commentsFlags encoder.CommentsFlags) error {
 	outputTypesSet := slices.ToSet(genConfigCmdFlags.outputTypes)
 
+	if err := validateFilesExists([]string{outputPaths.controlPlane, outputPaths.worker, outputPaths.talosconfig}); err != nil {
```
smira, line 286: I think this might be wrong it `outputPaths` is not complete/stdout. 

I would better keep the check in `writeToDestination`, or even use `writeToDestination` everywhere in `gen` function?

smira (approved):

Issue #6960, opened 2023-03-15
Talos Dashboard: Use resource watches for summary dashboard instead of get/list


utkuozdemir:
We do an excessive amount of get/list requests to populate the data used on the summary dashboard.

We should use resource watches instead.

https://github.com/siderolabs/talos/blob/main/internal/pkg/dashboard/datasource/node.go#L260

PR #6967, opened 2023-03-16
refactor: use resource watches on dashboard

utkuozdemir:
Instead of doing excessive get/list requests, do a watch per node in an infinite retry.

Additionally, refactor the dashboard code to make the various data listener namings more consistent and reorganize the packages.

Closes siderolabs/talos#6960.


internal/pkg/dashboard/apidata/source.go
```diff
@@ -85,23 +82,36 @@ func (source *API) run(dataCh chan<- *data.Data) {
 }
 
 //nolint:gocyclo,cyclop
-func (source *API) gather() *data.Data {
-	result := &data.Data{
+func (source *Source) gather() *Data {
```
utkuozdemir, line 85: Reverted API data gathering logic here to its previous shape, as we now don't need to do anymore an initial request to find out the set of nodes.

internal/pkg/dashboard/components/footer.go
```diff
@@ -19,13 +18,12 @@ import (
 type Footer struct {
 	tview.TextView
 
-	nodes             []string
-	selectedNodeIndex int
+	selectedNode string
```
utkuozdemir, line 21: Moved the node selection logic from the footer to the dashboard.go itself & made it simpler.

internal/pkg/dashboard/dashboard.go
```diff
@@ -46,14 +48,24 @@ const (
...
+
+// LogDataListener is a listener which is notified when a log line is received.
+type LogDataListener interface {
+	OnLogDataChange(node string, logLine string)
+}
+
+// NodeSetListener is a listener which is notified when the set of nodes changes.
+type NodeSetListener interface {
+	OnNodeSetChange(nodes []string)
 }
```
utkuozdemir, line 69: The idea here is to make the 3 different data sources and their listeners namings similar to each other.

internal/integration/api/apply-config.go
```diff
@@ -239,22 +239,18 @@ func (suite *ApplyConfigSuite) TestApplyConfigRotateEncryptionSecrets() {
 
 	encryption := machineConfig.MachineConfig.MachineSystemDiskEncryption
 
-	if encryption == nil {
+	if encryption == nil { //nolint:staticcheck
 		suite.T().Skip("skipped in not encrypted mode")
 	}
 
-	cfg := encryption.EphemeralPartition
```
smira, line 246: this looks weird, could it be that your `make lint` got wrong cache and now reporting wrong issues?

utkuozdemir: oh yes, must be that, how can I fix it?

internal/pkg/dashboard/resourcedata/resourcedata.go
```diff
@@ -0,0 +1,207 @@
...
+
+			return
+		case <-timer.C:
+		}
+	}
+}
+
+//nolint:gocyclo,cyclop
+func (source *Source) runResourceWatch(ctx context.Context, node string) error {
+	if node != "" {
```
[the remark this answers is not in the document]
utkuozdemir: I do it here? https://github.com/siderolabs/talos/pull/6967/files#diff-5a885872d568c5cd54266f334a33749652c59927321d5443c914dc469b9c014cR85-R89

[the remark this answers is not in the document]
smira: I would better move it into the function with `defer`, easier to understand the flow and handles panics as well

internal/integration/api/apply-config.go
smira: sometimes just `docket stop` for the buildx container is enough

sometimes `docker rm -f` for the buildx container helps (as it kills the cache)

you can also try `docker buildx prune`

smira (approved):

internal/integration/api/apply-config.go
utkuozdemir: yep, the last command solved it. reverted.

internal/pkg/dashboard/resourcedata/resourcedata.go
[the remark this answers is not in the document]
utkuozdemir: done

PR #6974, opened 2023-03-20
feat: netstat pod support

nberlee:
talosctl netstat -k show all host and non-hostnetwork-pods sockets/connections. talosctl netstat namespace/pod shows sockets/connections of a specific pod.

# Pull Request

## What? (description)
extend netstat to support kubernetes pods. You can either query a specific pod with the `namespace/pod` format as a parameter or lists all host + pod (non-hostnetworked) sockets using `-k` 
## Why? (reasoning)
* Container images should not ship with ss or netstat
* A unified view of all connections in your cluster is cool

## Acceptance


cmd/talosctl/cmd/talos/netstat.go
```diff
@@ -210,6 +248,12 @@ func printNetstat(response *machine.NetstatResponse) error {
 				}
 			}
 
+			if netstatCmdFlags.pods {
+				args = append(args, []interface{}{
```
frezbo, line 252: why `[]interface` when args is a list of strings?

frezbo (approved):
:cool:

cmd/talosctl/cmd/talos/netstat.go
nberlee: sre you thinking of the args inside `cobra.command`?
`args` is indeed a `[]string` there.

This is a builds a `args[]` array for the tab writer:
initialized here:
https://github.com/siderolabs/talos/blob/ae339a2024c2c7d33b491dfee7e980be2674736a/cmd/talosctl/cmd/talos/netstat.go#L212

used here: 
https://github.com/siderolabs/talos/blob/ae339a2024c2c7d33b491dfee7e980be2674736a/cmd/talosctl/cmd/talos/netstat.go#L268
as Fprintf takes a mix of types depending on the format string

smira:
I'm not sure, but is it possible to do a pure client-side implementation by grabbing a list of PIDs and then applying it as a filter to the existing `netstat` API?

nberlee:
I understand, you want to keep the APIs as clear as possible and not con-volute them with cross domain knowledge. Makes sense.

Yes I think its doable for me. But since pids are node specific, I need to distinguish them by adding a hostname, unless protobuf has some directed message per host feature...?

Guessing that is not the case, the relevant protobuf definition for `NetstatRequest` would look something like this:
```
...
  message NetNS {
    bool hostnetwork = 1;
    message PodPids {
      string hostname = 1;
      repeated uint32 pid = 2;
    }
    repeated PodPids podpids = 2;
  }
  NetNS netns = 4;
  ```
  
  And I need to add  `HostNetwork` to `message ContainerInfo`
  
Can you give me an estimate when you plan the release for 1.4.0 beta.0? As I guess that would end the merge window for this PR

smira:
Also I wonder if there's an easier way to filter sockets by a network namespace? I might be wrong, but it feels like it makes more sense to add a filter per network namespace.

nberlee:
@frezbo @smira Please review again (after TalosCon), now containers are client side.

> Also I wonder if there's an easier way to filter sockets by a network namespace? I might be wrong, but it feels like it makes more sense to add a filter per network namespace.

To be clear, this functionality is adding namespace socket results, not filtering them...

I also wonder who will filter sockets by a network namespace or even recognize cni of nsnet naming...

Its not about filtering, but actually adding namespaces sockets to the result. Its easy to read the link in `/var/run/netns` and iterate over all processes in /ns/net to find matching inodes. However, a network namespace name is not very user friendly and is not easy recognizable. 

It basically boils down to mappings in procfs:
* netns -> inodes
* processes -> inodes
* containers -> pids

I even tried syscalls to circumvent mappings of the above, however, syscall are really slower for network stats

nberlee:
Sorry for all the force pushes, I should have set it to draft after changing everything to the client side. 

I was in a loop:
for {
 thinking of something
 changing everything completely
 happy enough to push as I thought this is the best I can do
 after 10min, not really happy about the code
}

But i will now break for this until I get some feedback on where to improve :)

PR #6971, opened 2023-03-18
feat: add network config screen to dashboard

utkuozdemir:
Implement the network config screen with input forms to configure the initial node networking by writing a config to the META partition.

Closes siderolabs/talos#6961.

Preview:
dashboard-edit-network-config

PR #6980, opened 2023-03-21
fix: add uint32 type to Magic1 and Magic2

DmitriyMV:
Discovered in #6971. Go compiler cannot deduce proper type on 32bit architectures for those constants,
in `fmt.Print(f)` functions. Since we only compare them with uint32 variables, it makes sense to add proper
types to them.

### DIFF
--- a/internal/pkg/meta/internal/adv/talos/talos.go
+++ b/internal/pkg/meta/internal/adv/talos/talos.go
@@ -27,8 +27,8 @@ const (
 
 // Magic constants.
 const (
-	Magic1 = 0x5a4b3c2d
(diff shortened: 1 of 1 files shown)


frezbo (approved):
I'm not sure why this never failed before, of just that we started using this in fmt recently only?

smira:
> > I'm not sure why this never failed before, of just that we started using this in fmt recently only?
> 
> For some reason it wasn't failing till #6971, and that's the mystery :)

because this code got pulled in into `talosctl`, and we build it for `armv7`

smira (approved):

PR #6983, opened 2023-03-21
feat: update k8s to v1.27.0-beta.0

frezbo:
Update k8s to v1.27.0-beta.0


frezbo:
should I be running some integration test on this? Also what about the conformance?

smira:
> should I be running some integration test on this? Also what about the conformance?

you should `/promote` to `integration-provision` to test the k8s upgrade I think

smira (approved):

PR #6986, opened 2023-03-22
chore: bump deps

frezbo:
Bump deps


internal/pkg/tui/components/formbutton.go
```diff
@@ -74,6 +74,13 @@ func (b *FormModalButton) SetFinishedFunc(handler func(key tcell.Key)) tview.For
 	return b
 }
 
+// SetDisabled implements tview.FormItem.
```
frezbo, line 77: as per https://github.com/rivo/tview/commit/47b3275db48de684be18375672ea5c5671cad6e2

PR #6996, opened 2023-03-23
test: fix misprint in e2e scripts

smira:
This bug breaks `e2e-extensions`.


frezbo (approved):

frezbo:
not sure how it  worked previously :joy_cat:

PR #7011, opened 2023-03-27
feat: netstat pod support

nberlee:
# Pull Request

## What? (description)
extend netstat to support kubernetes pods. `talosctl netstat -k` show all host and non-hostnetwork pods sockets/connections. `talosctl netstat namespace/pod` shows sockets/connections of a specific pod + autocompletes in the shell.

### In comparison to #6974 :
Containers are identified as network namespaces by the Netstat api. This has an extra advantages as `-k` just means all namespaces in `/var/run/netns`, which simplifies the code at the client side. The lookup based on Pids is still available in the module, as it comes with slightly less overhead, although its not used in this PR.

## Why? (reasoning)
* Container images should not ship with ss or netstat
* A unified view of all connections in your cluster is cool

## Acceptance


internal/pkg/containers/cri/cri.go
```diff
@@ -225,6 +227,18 @@ func (i *inspector) buildPod(sandbox *runtimeapi.PodSandbox) (*ctrs.Pod, error)
 					pod.Containers[0].Digest = digest
 				}
 			}
+
+			if sandboxStatus.Linux.Namespaces.Options.Network == runtimeapi.NamespaceMode_POD {
+				if ns, ok := verboseInfo["runtimeSpec"].(map[string]interface{})["linux"].(map[string]interface{})["namespaces"].([]interface{}); ok {
```
frezbo, line 232: would `runtimeSpec.linux` be ever not set?

nberlee: Line 231 is logic which is not needed, and only there to skip unnecessary code.
I am happy to remove `if sandboxStatus.Linux.Namespaces.Options.Network` it is only there as on 233 starts a loop which will be pointless if the container is not network namespaced.

If your question is about the `if` on 232 itself, as this is unmarshaled json any assumption on how the json should look like could lead to a nil pointer reference in the future. (which is of course not worth stripping the if, also that why surrounding code like `verboseInfo["images")` and `verboseInfo["pod"]` have it.

Please let me know if I should remove it.

internal/pkg/containers/cri/cri.go
```diff
@@ -225,6 +227,18 @@ func (i *inspector) buildPod(sandbox *runtimeapi.PodSandbox) (*ctrs.Pod, error)
...
+					for _, n := range ns {
+						if nt, ok := n.(map[string]interface{}); ok && nt["type"] == "network" {
+							_, pod.Containers[0].NetworkNamespace = path.Split(nt["path"].(string))
```
nberlee, line 235: Just noticed this, I think this is safer:
```
							if netnsPath, ok := nt["path"].(string); ok {
								pod.Containers[0].NetworkNamespace = path.Base(netnsPath)

								break
							}
```

nberlee:
Ok added a  integration test, after some retries (force-pushes), I figured out to run e2e tests locally. 
Nice that is documented on the site, as I could not find the option with make help

And rediscovered (AGAIN :( ) that UDP sockets do not listen.

nberlee:
Ok, thats actually not the behavior of net-tools netstat.
`netstat -l` filters on  destination ip: 0.0.0.0 with port 0 and not on state.
This means that sockets with state CLOSE are also shown for udp and other protocols. (as there is no state LISTEN for stateless protocols like UDP) 

See:
udp: https://github.com/ecki/net-tools/blob/4030929bb6f3ee6f465c76869d7a49424df45d9e/netstat.c#L1316
raw: https://github.com/ecki/net-tools/blob/4030929bb6f3ee6f465c76869d7a49424df45d9e/netstat.c#L1413
tcp: https://github.com/ecki/net-tools/blob/4030929bb6f3ee6f465c76869d7a49424df45d9e/netstat.c#L1145

All not really related to this PR, but found in the test included, in order to make that PASS,
i fixed it in this commit also. 
Hope thats ok (some goes for the `--ipv6` `-6` which was assigned as the ipv4 bool flag.

smira (approved):
thank you :heart: 

very cool, did some small nitpicky fixes

internal/pkg/containers/cri/cri.go
smira: I modified this code to be less "panicky" to check for every layer in JSON

PR #7012, opened 2023-03-27
feat: allow writing initial META values into the image

### LINKED ISSUE #6999: META: writing during image creation
Allow writing `META` during image creation - when `imager` runs.

See if we can do something with ISO/PXE boot (?).

smira:
E.g. with the command:

```
make image-metal IMAGER_ARGS="--meta 0xc=abc --meta 0xd=abc"
```

This doesn't support ISO/PXE boot yet, it's going to come into the next PR.

Fixes #6999

### DIFF
--- a/cmd/installer/cmd/root.go
+++ b/cmd/installer/cmd/root.go
@@ -6,9 +6,11 @@
 package cmd
 
 import (
+	"encoding/base64"
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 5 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -331,8 +332,16 @@ func (i *Installer) Install(seq runtime.Sequence) (err error) {
 
 		var ok bool
 
-		if ok, err = metaState.SetTag(context.Background(), meta.Upgrade, string(i.Current)); !ok || err != nil {
-			return fmt.Errorf("failed to set upgrade tag: %q", i.Current)
+		if seq == runtime.SequenceUpgrade {
```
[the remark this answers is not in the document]
smira: above we check if any of two conditions is true, if true, we load META and prepare to flush it in the end

here we re-check the condition - we don't want to set the key if it's not upgrade

smira:
Same with ISO image, but internally the path is more complicated, and it uses kernel args to set environment variables which are passed to the installer container:

```
$ make iso IMAGER_ARGS="--meta 0xda=foobar"
$ talosctl -n 172.20.0.2 get meta
NODE         NAMESPACE   TYPE      ID     VERSION   VALUE
172.20.0.2   runtime     MetaKey   0xda   1         foobar
```

cmd/installer/cmd/root.go
```diff
@@ -23,9 +25,34 @@ var rootCmd = &cobra.Command{
...
+	}
+
+	return nil
+}
+
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
+	// Set defaults for flags from the environment variables.
+	if err := parseEnvironment(); err != nil {
```
frezbo, line 51: rename to setEnvironment, since it also sets meta?

smira: I mean it parses environment variables and sets flags (defaults) based on env variables, but probably name is not good one

frezbo (approved):

PR #7013, opened 2023-03-27
feat: environment variables via the kernel arguments

### LINKED ISSUE #6984: HTTP proxy variables for kernel arguments
## Feature Request

It would be helpful to add extra kernel command-line arguments so that Talos Linux is able to use an http proxy at the boot process before obtaining the `talos.config` content and before the network is configured.

Potentially, the variables could be named `http_proxy` and `https_proxy`.

### Description

In most of our environments egress capabilities are restricted, particularly during machine provisioning, where the out-of-band network that is used to fetch the boot artifact and boot the machine, also prevents communication to the outside world. It's not possible to reach out to any external http services, whether on the internet or even outside of the server failure domain.

We currently have to store the talos config content on a static file server located in the same network where the machine is located. To use Omni, it means having a L4 proxy server located in that environment that will proxy the requests towards the Omni endpoint, which we have to deploy in each failure domain.

smira:
Unify getting environment variables, support passing environment variables via kernel args.

Fixes #6984
See #6999

For META this will be used to pass environment variables to the installer for ISO images (or PXE booting).

### DIFF
--- a/internal/pkg/environment/environment.go
+++ b/internal/pkg/environment/environment.go
@@ -1,0 +1,44 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package environment provides a set of functions to get environment variables.
+package environment
+
+import (
+	"github.com/siderolabs/go-procfs/procfs"
+
+	"github.com/siderolabs/talos/pkg/machinery/config"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+)
+
+// Get the desired set of the environment variables based on kernel cmdline and machine config.
+//
+// The returned value is a list of strings in the form of "key=value".
+func Get(cfg config.Provider) []string {
+	return GetCmdline(procfs.ProcCmdline(), cfg)
+}
+
+// GetCmdline the desired set of the environment variables based on kernel cmdline.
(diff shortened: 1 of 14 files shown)


frezbo (approved):

internal/pkg/environment/environment.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for {
+		val := param.Get(idx)
+		if val == nil {
+			break
+		}
+
+		result = append(result, *val)
+		idx++
+	}
```
DmitriyMV, line 40:
```suggestion
	for idx := 0; ; idx++ {
		val := param.Get(idx)
		if val == nil {
			break
		}
		result = append(result, *val)
	}
```

internal/pkg/environment/environment.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if cfg != nil {
+		for k, v := range cfg.Machine().Env() {
+			result = append(result, fmt.Sprintf("%s=%v", k, v))
```
DmitriyMV, line 44:
```suggestion
			result = append(result, k+"="+v)
```

This should do the trick.

DmitriyMV (approved):
Few minor nits, but all good otherwise.

internal/pkg/environment/environment.go
smira: thanks, fixed!

internal/pkg/environment/environment.go
smira: thanks, fixed!

PR #7023, opened 2023-03-29
fix: add a link-scope route if the cmdline gateway is not reachable

### LINKED ISSUE #7020: Talos does not boot correctly in /32 subnets
### Description

I'm using a Hetzner server with Proxmox installed in routed mode as described [here](https://community.hetzner.com/tutorials/install-and-configure-proxmox_ve#example-of-the-guest-configuration-in-routed-setup)

To install Talos, the [documentation](https://www.talos.dev/v1.3/talos-guides/install/virtualized-platforms/proxmox/) describes the `ip=` kernel parameter, that I set according to the screenshot:

image

For the sake of privacy, I removed the IP addresses. Let's take 192.168.0.2 as the IP address of the gateway and 10.0.0.42 as the IP address of the Talos VM.

After booting, the system does not come up and isn't responding to ping. In tcpdump I can only see ARP requests coming from the vm host.

### Logs

talos_logs

### Environment

- Talos version: 1.3.6
- Kubernetes version: none, pre-install environment
- Platform: Hetzner/Proxmox

smira:
Fixes #7020

### DIFF
--- a/internal/app/machined/pkg/controllers/network/route_config.go
+++ b/internal/app/machined/pkg/controllers/network/route_config.go
@@ -192,27 +192,43 @@ func (ctrl *RouteConfigController) parseCmdline(logger *zap.Logger) (routes []ne
 			continue
 		}
 
-		var route network.RouteSpecSpec
-
-		route.Gateway = linkConfig.Gateway
-
-		if route.Gateway.Is6() {
-			route.Family = nethelpers.FamilyInet6
-		} else {
-			route.Family = nethelpers.FamilyInet4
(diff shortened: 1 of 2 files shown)


frezbo:
needs backport?

frezbo (approved):

smira:
> needs backport?

I feel it's a change we shouldn't backport... but I'm torn, idk

sergelogvinov:
I've had the same case. And I solve it by using IPv6... 
thank you for `link-scope route` ❤️

PR #7024, opened 2023-03-29
test: bump e2e-aws Calico version

smira:
Try updating to the latest one.


smira:
So the test passes with v1.27.0-beta.0, but not with v1.27.0-rc.0

Issue #6993, opened 2023-03-23
add `${code}` field to the dashboard


smira:
Could be part of the network config screen.

See #6989

PR #7029, opened 2023-03-30
feat: add config code entry screen to dashboard

utkuozdemir:
Implement a screen for entering/managing the config `${code}` variable.

Enable this screen only when the platform is `metal` and there is a `${code}` variable in the `talos.config` kernel cmdline URL query.

Additionally, remove the "Delete" button and its functionality from the network config screen to avoid users accidentally deleting PlatformNetworkConfig parts that are not managed by the dashboard.

Add some tests for the form data parsing on the network config screen.

Closes [#6993](https://github.com/siderolabs/talos/issues/6993).

Preview:

configcode


internal/app/dashboard/main.go
```diff
@@ -51,7 +54,30 @@ func dashboardMain() error {
...
+
+		if showConfigCodeTab() {
+			screens = append(screens, dashboard.ScreenConfigCode)
+		}
 	}
 
 	return dashboard.Run(adminCtx, c, dashboard.WithAllowExitKeys(false), dashboard.WithScreens(screens...))
 }
+
+func showConfigCodeTab() bool {
```
smira, line 66: :heart:

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url/variable.go
```diff
@@ -28,31 +28,29 @@ type Variable struct {
...
-			Key:   constants.HostnameKey,
-			Value: HostnameValue(),
-		},
-		{
-			Key:   constants.CodeKey,
-			Value: CodeValue(),
-		},
-	}
+// AllVariables is a map of all supported variables.
+var AllVariables = map[string]*Variable{
```
smira, line 32: as `Variable` embeds `Value` which contains matched value, we should always generate a fresh (empty) list here, so this should be still a `func()`

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/metal/url/variable.go
utkuozdemir: got it, done.

utkuozdemir:
> Since we have plans on making this much more generic and allowing for other variables besides `code` I feel the tab should be named something like "Config URL".

Changed it to be Config URL.

PR #7030, opened 2023-03-30
feat: introduce network probes

### LINKED ISSUE #6998: network health checks
Minimal initial implementation - no support for machine config (yet), but allow in platform network config.

The result of TCP checks should probably affect network status - connectivity check.

smira:
Network probes are configured with the specs, and provide their output as a status.

At the moment only platform code can configure network probes.

If any network probes are configured, they affect network.Status 'Connectivity' flag.

Example, create the probe:

```
talosctl -n 172.20.0.3 meta write 0xa '{"probes": [{"interval": "1s", "tcp": {"endpoint": "google.com:80", "timeout": "10s"}}]}'
```

Watch probe status:

```
$ talosctl -n 172.20.0.3 get probe
NODE         NAMESPACE   TYPE          ID                  VERSION   SUCCESS
172.20.0.3   network     ProbeStatus   tcp:google.com:80   5         true
```

With failing probes:

```
$ talosctl -n 172.20.0.3 get probe
NODE         NAMESPACE   TYPE          ID                  VERSION   SUCCESS
172.20.0.3   network     ProbeStatus   tcp:google.com:80   4         true
172.20.0.3   network     ProbeStatus   tcp:google.com:81   1         false
$ talosctl -n 172.20.0.3 get networkstatus
NODE         NAMESPACE   TYPE            ID       VERSION   ADDRESS   CONNECTIVITY   HOSTNAME   ETC
172.20.0.3   network     NetworkStatus   status   5         true      false          true       true
```

Fixes #6998


frezbo (approved):

frezbo:
> With failing probes:

shouldn't the `networkstatus` say false?

smira:
> > With failing probes:
> 
> shouldn't the `networkstatus` say false?

it does say false, copy-pasted wrong example :-D

internal/app/machined/pkg/controllers/network/internal/probe/probe.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+		} else {
+			firstIteration = false
+		}
+
+		err := runner.probe(ctx)
+		if err == nil {
+			if consecutiveFailures > 0 {
+				logger.Info("probe succeeded")
+			}
```
utkuozdemir, line 89: didn't get this - we log "probe succeeded" if there is a failure?

pkg/machinery/resources/network/status.go
```diff
@@ -49,7 +49,24 @@ func (StatusExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 		Type:             StatusType,
 		Aliases:          []resource.Type{"netstatus", "netstatuses"},
 		DefaultNamespace: NamespaceName,
-		PrintColumns:     []meta.PrintColumn{},
+		PrintColumns: []meta.PrintColumn{
```
utkuozdemir, line 52: 🆒

utkuozdemir (approved):

internal/app/machined/pkg/controllers/network/internal/probe/probe.go
smira: we log it if there were failures before, so that we don't log each success

Issue #7001, opened 2023-03-24
Talos not reloading/retriggering udev on boot


buroa:
## Bug Report

udev is not reloading/retriggering rules on boot

### Description

While trying to get my thunderbolt devices to authorize on a boot, I am using this udev rule:

```
  udev:
    rules:
      - ACTION=="add", SUBSYSTEM=="thunderbolt", ATTR{authorized}=="0", ATTR{authorized}="1"
```

However, it is not being applied on boot correctly. If we run after boot inside some privileged container:

```
root@w2:/# nsenter --mount=/proc/1/ns/mnt -- udevadm trigger --type=devices --action=add
root@w2:/# nsenter --mount=/proc/1/ns/mnt -- udevadm trigger --type=subsystems --action=add
```

We can then see this devices are running that rule correctly, which authorizes the thunderbolt devices.

### Logs

No logs really... I can just see that the device is not authorized as it should be with that udev rule:

```
talosctl -n 10.0.0.21 read /sys/bus/thunderbolt/devices/0-3/authorized
0
```

### Environment

- Talos version:

Client:
	Tag:         v1.3.1
	SHA:         4469ad12
	Built:
	Go version:  go1.19.4
	OS/Arch:     darwin/arm64
Server:
	NODE:        10.0.0.22
	Tag:         v1.4.0-alpha.2
	SHA:         594f27d8
	Built:
	Go version:  go1.20.1
	OS/Arch:     linux/amd64
	Enabled:     RBAC

- Kubernetes version: Server Version: v1.26.2
- Platform: Apple Mac Mini's (Intel 2018)

PR #7032, opened 2023-03-30
fix: udevd rules trigger

frezbo:
Restart udevd on adding custom rules where in the case the subsystems needs to be re-triggered.

Fixes: #7001

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
@@ -164,7 +164,7 @@ func (g *GCP) Configuration(ctx context.Context, r state.State) ([]byte, error) 
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/gcp/gcp.go
```diff
@@ -164,7 +164,7 @@ func (g *GCP) Configuration(ctx context.Context, r state.State) ([]byte, error)
 		return nil, err
 	}
 
-	log.Printf("fetching machine config from AWS")
+	log.Printf("fetching machine config from GCP metadata service")
```
smira, line 167: :rofl:

smira (approved):

frezbo:
Also fixes a typo in gcp code. This is more of a simple fix, I think the udevd rules is better handled as a controller :shrug:

Issue #7009, opened 2023-03-27
talosctl reset interactive stuck after reboot


smira:
the command seems to be stuck after a reboot, while the node is actually ok, and `talosctl` commands work:

```
$ talosctl -n 10.68.182.3 reset --system-labels-to-wipe=EPHEMERAL --reboot                                                                                                                                                                                                                                                    
◳ watching nodes: [10.68.182.3]
    * 10.68.182.3: unavailable, retrying...
```

the upgrade command seemed to work fine (?)

Issue #6966, opened 2023-03-16
`talosctl reset` returns invalid exit code on error


artuross:
## Bug Report
I'm resetting nodes via `talosctl reset --nodes <hostname>` before decomissioning servers. I use the hostname, because it's convenient and the cli supports it.

Sometimes, I get this error:

```sh
❯ talosctl --talosconfig talosconfig reset --graceful --wait --timeout 240s --nodes test
◰ watching nodes: [test]
    * test: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: lookup test on 10.10.10.10:53: no such host"
```

but the exit code is `0`. I think the behavior should be consistent with other `talosctl` commands where the exit code set to `1`.

To me it looks [caused by this line](https://github.com/siderolabs/talos/blob/main/cmd/talosctl/cmd/talos/reset.go#L134).

### Environment

- Talos version: `v1.3.5`
- Kubernetes version: `v1.26.0`
- Platform: `hetzner`

I'm running `talosctl` on an M1 Mac.

PR #7048, opened 2023-04-03
fix: improve action tracking post checks

utkuozdemir:
In the tracking of the `reset --reboot`, `reboot` and `upgrade` lifecycle commands, verify that the node was actually rebooted in the post check by comparing the pre- and post-check boot IDs.

In the `reset --reboot` post-check, try both maintenance and normal mode, since the reset might be issued to only remove `EPHEMERAL` partition, which will not put the node into the maintenance mode.

Fixes https://github.com/siderolabs/talos/issues/7009.

Additionally, if an action tracking fails, return the error instead of swallowing it. This way the command erminates with a non-zero exit code. Suppress the re-printing this error after the command was run.

Fixes https://github.com/siderolabs/talos/issues/6966.

### DIFF
--- a/cmd/talosctl/pkg/talos/action/tracker.go
+++ b/cmd/talosctl/pkg/talos/action/tracker.go
@@ -18,6 +18,7 @@ import (
 	"github.com/siderolabs/gen/containers"
 	"github.com/siderolabs/gen/maps"
 	"github.com/siderolabs/go-circular"
+	"github.com/siderolabs/go-retry/retry"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/backoff"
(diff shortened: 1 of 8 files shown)


cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -50,6 +51,20 @@ var (
...
+
+	// BootIDChangedPostCheckFn is a post check function that returns nil if the boot ID has changed.
+	BootIDChangedPostCheckFn = func(ctx context.Context, c *client.Client, preActionBootID string) error {
+		currentBootID, err := getBootID(ctx, c)
+		if err != nil {
+			return err
+		}
+
+		if preActionBootID == currentBootID {
+			return retry.ExpectedError(fmt.Errorf("didn't reboot yet"))
```
smira, line 63:
```suggestion
			return retry.ExpectedErrorf("didn't reboot yet")
```

smira (approved):

PR #7057, opened 2023-04-05
chore: update Go to 1.20.3

smira:
See https://go.dev/doc/devel/release#go1.20


hack/release.toml
```diff
@@ -26,7 +26,7 @@ preface = """\
 * CoreDNS: v1.10.1
 * Flannel: v0.21.4
 
-Talos is built with Go 1.20.2.
+Talos is built with Go 1.20.3.
```
frezbo, line 29: don't we drop the previous release notes, or is only after v1.4

frezbo (approved):

hack/release.toml
frezbo: makes sense :+1:

PR #7062, opened 2023-04-06
fix: quote ISO kernel args for GRUB

smira:
Use GRUB quoting function to the kernel args passed to Talos.

This fixes passing `${variable}` to `talos.config=` kernel argument.

Also fix a problem with `ONBUILD` being exected for `imager` image.


Dockerfile
```diff
@@ -706,7 +708,7 @@ ONBUILD RUN find /rootfs \
     && rm -rf /initramfs
 ONBUILD WORKDIR /
 
-FROM installer AS imager
```
smira, line 709: the `ONBUILD` instructions in `installer` were executed here, while it shouldn't have been

frezbo (approved):

DmitriyMV (approved):

PR #7101, opened 2023-04-18
docs: update documentation for Talos 1.3

hkraal:
## What? (description)

Mention the requirement of the nocloud image.

## Why? (reasoning)

It might be non obvious to newer users (that would be me), it might also save some hairs.


smira:
@hkraal could you please sign-off your commit? (DCO)

`git commit --amend -s` if using the CLI

hkraal:
Should be fixed, let me know if more is needed.

smira (approved):

PR #7115, opened 2023-04-20
chore: fix container image reproducibility

smira:
Also provide make targets to sign the images.


Dockerfile
```diff
@@ -663,6 +665,8 @@ COPY --from=pkg-grub / /
 COPY --from=unicode-pf2 /usr/share/grub/unicode.pf2 /usr/share/grub/unicode.pf2
 
 FROM alpine:3.17.2 AS installer-image
+ARG SOURCE_DATE_EPOCH
```
frezbo, line 668: `SOURCE_DATE_EPOCH` comes from buildkit?

smira: we inject it

.drone.jsonnet
```diff
@@ -575,6 +578,7 @@ local integration_pipelines = [
   Pipeline('cron-integration-qemu-race', default_pipeline_steps + [build_race, integration_qemu_race], [default_cron_pipeline]) + cron_trigger(['nightly']),
   Pipeline('cron-integration-qemu-csi', default_pipeline_steps + [integration_qemu_csi], [default_cron_pipeline]) + cron_trigger(['nightly']),
   Pipeline('cron-integration-images', default_pipeline_steps + [integration_images, integration_sbcs], [default_cron_pipeline]) + cron_trigger(['nightly']),
+  Pipeline('cron-integration-reproducibility-test', default_pipeline_steps + [integration_reproducibility_test], [default_cron_pipeline]) + cron_trigger(['nightly']),
```
frezbo, line 581: do we need this nighty where we built with a fresh cache, seems to aggressive, maybe weekly and before release

frezbo: at least until the ci cutover?

frezbo (approved):

.drone.jsonnet
smira: I'll see how long it takes in the CI. but in general whole nightly passes fine except for extensions which is basically launching a Talos build which waits for extensions build which in turn wants another whole machine (in terms of resource claims). I think proper fix is to lower resource claims for extensions build

Unix4ever (approved):

PR #7120, opened 2023-04-21
feat: implement distributed image store

smira:
This is PoC/experiment.

Each Talos node runs `registryd` component which acts both as a registry and a fan-out service. For local requests, `registryd` serves manifests/blobs from the containerd content storage. For incoming requests, `registryd` fans out requests to other nodes (cluster members), finding the first one which has the content.

I had to disable content store deduplication, as otherwise containerd drops original layers immediately.

One not fully solved question is how to inject `registryd`, what I did in my testing is to inject it as the endpoint in the registry mirror scheme, so if `registryd` has nothing, `containerd` falls back to "upstream" registry/mirror. There needs some work to be done to support it for `*` redirects.

There is unresolved issues with images protected by authorization. At the moment `registryd` never resolves tags (defers it to the upstream registry), but still it might deliver images without pull secrets given the proper digest.

How to secure `registryd` from access outside of the cluster?

### DIFF
--- a/internal/app/machined/pkg/system/services/registryd.go
+++ b/internal/app/machined/pkg/system/services/registryd.go
@@ -1,0 +1,479 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package services
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/netip"
+	"net/url"
+	"path"
+	"strconv"
+	"strings"
+	"time"
+
+	containerdapi "github.com/containerd/containerd"
+	"github.com/containerd/containerd/content"
+	"github.com/containerd/containerd/namespaces"
+	criconstants "github.com/containerd/containerd/pkg/cri/constants"
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/docker/errdefs"
+	"github.com/hashicorp/go-cleanhttp"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/siderolabs/gen/channel"
+	"github.com/siderolabs/gen/slices"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system/events"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system/health"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system/runner"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system/runner/goroutine"
+	"github.com/siderolabs/talos/pkg/conditions"
+	"github.com/siderolabs/talos/pkg/logging"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+	"github.com/siderolabs/talos/pkg/machinery/resources/cluster"
+)
+
+type registrydService struct {
+	logger     *zap.Logger
+	resources  state.State
+	client     *containerdapi.Client
+	httpClient *http.Client
+}
+
+// Main is an entrypoint to the API service.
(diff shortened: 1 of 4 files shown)


smira:
This requires something like in the machine config (first endpoint is `registryd`, second is my registry mirrors, unrelated):

```
            ghcr.io:
                # List of endpoints (URLs) for registry mirrors to use.
                endpoints:
                    - http://127.0.0.1:3172
                    - http://172.20.0.1:5004
```

smira:
Of course final solution should be opt-in, configurable with a single flag:

* reconfigure CRI not to drop image layers from content store
* reconfigure mirror endpoints to inject "registryd" endpoint as the first one

Open questions:

* images with pull credentials - need to dig more into that... how the auth is applied to blobs/manifests, what if the layer is shared? 
* securing access to registryd from outside of the cluster

internal/app/machined/pkg/system/services/registryd.go
```diff
@@ -0,0 +1,479 @@
...
+			logger.Error("failed to get content info", zap.Error(err))
+			w.WriteHeader(http.StatusInternalServerError)
+
+			return
+		}
+
+		break
+	}
+
+	if err != nil {
```
frezbo, line 225: this seems not needed, which err case is this handling?

smira: this handles the case when we get `IsNotFound` for all namespaces we checked for

ruifung:
> * reconfigure mirror endpoints to inject "registryd" endpoint as the first one

I just found this and had a few thoughts, how would this work if the config has something like this?

```
docker.io:
    overridePath: true
    endpoints:
      - https://harbor.example.com/v2/dockerhub/
```

And don't mirror endpoints require listing every single source separately?

smira:
@ruifung I don't think I got your question, but this is early PoC, not real implementation yet, so some details are not known

Ornias1993:
> > * reconfigure mirror endpoints to inject "registryd" endpoint as the first one
> 
> I just found this and had a few thoughts, how would this work if the config has something like this?
> 
> ```
> docker.io:
>     overridePath: true
>     endpoints:
>       - https://harbor.example.com/v2/dockerhub/
> ```
> 
> And don't mirror endpoints require listing every single source separately?

Afaik we can just use this as a forward for the image store?

onedr0p:
@smira I'm not sure if you saw this project or not but it works great on Talos. It seems like what you want to do here, maybe you'll find some ideas looking thru the source.

https://github.com/XenitAB/spegel

Ornias1993:
> @smira I'm not sure if you saw this project or not but it works great on Talos. It seems like what you want to do here, maybe you'll find some ideas looking thru the source.
> 
> https://github.com/XenitAB/spegel

Thats great software actually, thanks for the tip!

smira:
> @smira I'm not sure if you saw this project or not but it works great on Talos. It seems like what you want to do here, maybe you'll find some ideas looking thru the source.
> 
> https://github.com/XenitAB/spegel

yes, this was the inspiration, but probably more stuff we could do easier, but this is not done yet

phillebaba:
@smira is there something specific in Spegel that you do not want, which is causing you to implement your own embedded registry?

smira:
> @smira is there something specific in Spegel that you do not want, which is causing you to implement your own embedded registry?

it's not that Spegel has anything wrong, but rather it's a generic solution, while on Talos Linux we have more control and more information, e.g. we have the discovery data. So it should be easier to implement and run it on Talos. 

Also it's our philosophy to keep things simple for the end users, just flip the switch and you get  a distributed image cache.

phillebaba:
I agree with you, my thought was that Talos could embed Spegel the same way k3s does. You don't even have to use the libp2p router if you have some other way of routing the traffic. Most components are interfaces so it should be pretty easy to just replace the router with a custom implementation.

Ornias1993:
> I agree with you, my thought was that Talos could embed Spegel the same way k3s does. You don't even have to use the libp2p router if you have some other way of routing the traffic. Most components are interfaces so it should be pretty easy to just replace the router with a custom implementation.

I think this is a great suggestion, which is also easier to maintain.

Ornias1993:
> > @smira is there something specific in Spegel that you do not want, which is causing you to implement your own embedded registry?
> 
> it's not that Spegel has anything wrong, but rather it's a generic solution, while on Talos Linux we have more control and more information, e.g. we have the discovery data. So it should be easier to implement and run it on Talos.
> 
> Also it's our philosophy to keep things simple for the end users, just flip the switch and you get a distributed image cache.

I've done some implementation of Spegel now, and I have to say: It basically does precisely what you describe here... Its pretty much "apply and forget".

PrivatePuffin:
I'm pretty sure this isn't stale and @smira is still working/thinking-about this.

phillebaba:
Has there been any decision made on Spegel vs. implementing your own solution?

PrivatePuffin:
> Has there been any decision made on Spegel vs. implementing your own solution?

He literally said above they had already made a decision.

Issue #7066, opened 2023-04-09
Can't cold boot Talos on a T2 Intel Mac


dhess:
## Bug Report

Using a USB boot drive, and having disabled the system's Apple-signed boot restrictions, I'm able to install Talos v1.3.7 on a T2 Intel Mac (iMac Pro (2017), in this specific case) just fine, and everything at runtime works as expected. `kexec` warm reboots also work fine. But as soon as I cold boot the machine, I get the firmware's "missing boot folder" start-up screen. Holding the Option key during boot doesn't show the EFI partition as a choice, either &mdash; the only option I get is to boot into Internet recovery mode.

[rEFInd](http://www.rodsbooks.com/refind/) is a common solution to Linux booting issues on T2 Intel Macs, but even booting rEFInd from a USB drive or SD card doesn't give an option to boot Talos. rEFInd doesn't seem to recognize the EFI partition.

Booting into the Mac's recovery mode and running `refind-install` also doesn't work as expected. In this case, rEFInd's installer recognizes the EFI partition and tries to mount it (via `mount -t msdos`), but this fails with a cryptic `Invalid argument` message.

Is there something unusual about Talos's EFI partition?

One workaround is to boot from the USB drive that I used to install Talos, which seems to recognize the installed image and uses its config, but that isn't a good long-term solution since it would require updating the drive every time I also want to update the installed version of Talos.

Note that there was a similar issue reported in the Sidero Labs Community Slack last month, this time for a 2018 Mac mini: https://taloscommunity.slack.com/archives/CMARMBC4E/p1678790269057759

### Environment

- Talos version: v1.3.7
- Kubernetes version: v1.26.3
- Platform: iMac Pro (2017), using built-in (nvme) storage

smira:
We don't have any specific answer, you might need to dig into it. My only guess it's something with the way GRUB installs and sets up things.

dhess:
I've returned to this and while playing around a bit, I've discovered that Disk Utility in recovery mode claims that the EFI partition is too small, and that the disk needs to be repartitioned. Is there an easy way to change the EFI partition size used by the Talos installer without patching the source and rebuilding from scratch?

smira:
No, EFI size is hardcoded in Talos

tommy-skaug:
Bumping this as the workaround in the former PR doesn't work on v1.5.4 for me (if I'm not noobing about something with the build). Was there any progress made in regard to the mentioned 1.5.0 work?

```
[...]

PR #7132, opened 2023-04-25
fix: Increase EFI partition size to 200 MiB.

dhess:
Fixes: #7066

# Pull Request

## What? (description)

This PR increases the fixed EFI partition size to 200MiB. Apparently some EFI environments, including at least T2 Intel Macs, don't like sizes smaller than this; ref:

* https://www.linux.org/threads/efi-partition-sizes.35741/
* https://www.ctrl.blog/entry/esp-size-guide.html

## Why? (reasoning)

1. T2 Intel Macs can boot most Linuxes, but not Talos. I was able to boot the Talos v1.3.7 and v1.4.0 ISOs on a 2017 iMac Pro using a USB flash drive, and `kexec` reboots also worked, but warm & cold reboots did not: the Mac does not recognize the EFI partition. During debugging, I noted that the Mac's recovery mode Disk Utility reported that the EFI partition size (100MiB) was too small. After bumping the size to 200MiB and re-installing using the modified installer, Talos v1.4.0 now boots fine on the iMac Pro.
2. Anecdotally, there are also reports of issues with 100MiB EFI partitions on at least some other PCs, though I haven't personally experienced this (except on the 2017 iMac Pro as described above).
3. An additional 100MiB for better compatibility seems like a good tradeoff.

## Acceptance

### DIFF
--- a/internal/pkg/partition/constants.go
+++ b/internal/pkg/partition/constants.go
@@ -30,7 +30,7 @@ const (
 const (
 	MiB = 1024 * 1024
 
-	EFISize      = 100 * MiB
+	EFISize      = 200 * MiB
 	BIOSGrubSize = 1 * MiB
 	BootSize     = 1000 * MiB
 	MetaSize     = 1 * MiB


smira:
We had internal discussion on this issue. There will be some work done for Talos 1.5 for Trusted Boot which might lead to big changes to the whole boot process, and probably much bigger EFI partition.

On other hand, we don't want to put specific workarounds for a single hardware type to mainline Talos.

You can build Talos images with your local changes: https://www.talos.dev/v1.4/advanced/building-images/

buroa:
@smira Can we reopen this? 100MiB doesn't seem like much, if anything. I'm also sure this would benefit other devices that are not known at this time; so the _workaround_ (really?) wouldn't target "just one device".

frezbo:
> @smira Can we reopen this? The Trusted Boot got it's own EFI size; so it's still a default of 100MiB which doesn't seem like a large increase to 200MiB. I'm also sure this would benefit other devices that are not known at this time; so the _workaround_ (really?) wouldn't target "just one device".
> 
> [#7066 (comment)](https://github.com/siderolabs/talos/issues/7066#issuecomment-1789228904)

this is already fixed if you generate a UKI

buroa:
> > @smira Can we reopen this? The Trusted Boot got it's own EFI size; so it's still a default of 100MiB which doesn't seem like a large increase to 200MiB. I'm also sure this would benefit other devices that are not known at this time; so the _workaround_ (really?) wouldn't target "just one device".
> > [#7066 (comment)](https://github.com/siderolabs/talos/issues/7066#issuecomment-1789228904)
> 
> this is already fixed if you generate a UKI

So either I turn on SecureBoot or I don't get a increase in EFI?

frezbo:
> > > @smira Can we reopen this? The Trusted Boot got it's own EFI size; so it's still a default of 100MiB which doesn't seem like a large increase to 200MiB. I'm also sure this would benefit other devices that are not known at this time; so the _workaround_ (really?) wouldn't target "just one device".
> > > [#7066 (comment)](https://github.com/siderolabs/talos/issues/7066#issuecomment-1789228904)
> > 
> > 
> > this is already fixed if you generate a UKI
> 
> So either I turn on SecureBoot or I don't get a increase in EFI?

UKI doesn't need SecureBoot to be enforced, you can boot a UKI in non-secureboot mode also

buroa:
> > > > @smira Can we reopen this? The Trusted Boot got it's own EFI size; so it's still a default of 100MiB which doesn't seem like a large increase to 200MiB. I'm also sure this would benefit other devices that are not known at this time; so the _workaround_ (really?) wouldn't target "just one device".
> > > > [#7066 (comment)](https://github.com/siderolabs/talos/issues/7066#issuecomment-1789228904)
> > > 
> > > 
> > > this is already fixed if you generate a UKI
> > 
> > 
> > So either I turn on SecureBoot or I don't get a increase in EFI?
> 
> UKI doesn't need SecureBoot to be enforced, you can boot a UKI in non-secureboot mode also

Configuring UKI; generating more keys; etc sounds like a bigger pain than I'm willing to go through with just to get a bigger EFI partition size.

I'm not a go expert, but isn't there someway you can override the `partition.EFISize` constant on build using something like `GO_LDFLAGS="-X partition.EFISize=2097152"` when using `make installer`?

frezbo:
constants cannot be overridden, only variables

buroa:
> constants cannot be overridden, only variables

Can we make it configurable?

smira:
we do not plan to do so, but probably we'll offer a way to boot via UKIs without SecureBoot on UEFI systems soon. With GRUB, you get small EFI partition size, without GRUB, you get a bigger one.

PR #7133, opened 2023-04-25
feat: clean up (garbage collect) system images which are not referenced

### LINKED ISSUE #7121: Talos should clean up images in `system` namespace of CRI containerd
Talos pulls two images itself: `kubelet` and `etcd`, but never prunes them.

Kubelet/CRI doesn't see these images (as they're in a different namespace), but this is good, as they won't be pruned by mistake.

Talos should GC these images based on the fact they're not being used.

(this is only a problem if the image ref is changed, i.e. new image is pulled)

smira:
Fixes #7121

Talos pulls some images on its own (without CRI/kubelet) to the `system` namespace of the CRI containerd. These images are not visible to the CRI/kubelet, so we need to clean them up manually.

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
+++ b/internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
@@ -1,0 +1,272 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package runtime
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/namespaces"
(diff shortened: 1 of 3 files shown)


frezbo (approved):

smira:
Refactored adding unit-tests

internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
```diff
@@ -0,0 +1,272 @@
...
+	return "runtime.CRIImageGCController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *CRIImageGCController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: v1alpha1.NamespaceName,
+			Type:      v1alpha1.ServiceType,
+			ID:        pointer.To("cri"),
```
frezbo, line 61: no defined name for `cri` ?

frezbo (approved):

internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
smira: yeah, we have it as "cri" everywhere...

Issue #7117, opened 2023-04-20
[bug] regression: user disks requiring udev rules no longer mount


Ulexus:
## Bug Report
As of v1.4 (and specifically commit [ddb014cfdcb3630055d5b7be7f480080612dfc11](https://github.com/siderolabs/talos/commit/ddb014cfdcb3630055d5b7be7f480080612dfc11) (I believe), there appears to be a race condition between udev rules being processed and user disks being mounted.

### Description

Before v1.4.0, we had no problems.  Likewise, downgrading from v1.4.0 resolves the problems.

We have a machineconfig-supplied udev rule defined which establishes a symlink between an NVME device and a known, fixed name.  Additionally, a user disk mount is defined using that fixed name.

When the v1.4.0 system boots up, we see udev config being processed _after_ the mount attempt fails (despite the fact that udev rule application is supposed to occur _before_ mountUserDisks in the [sequencer](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go#L267-L275)).  While I describe this as a possible race condition, this has thus far failed _every_ time.

It is further strange, because the commit referenced above does not seem to do anything about concurrency; rather it just restarts `udevd`.

At first glance, one might assume that we need to add a WaitForService [here](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go#L790), like several other tasks have... but testing this locally did not resolve the problem.

### Environment

- Talos version: v1.4.0
- Kubernetes version: N/A
- Platform: AWS EC2, x86-64

smira:
Sean, thanks for the report, so the symlink eventually appears, but it's "too late"? 

I'm trying to understand whether `udevd` correctly re-processes the events and applies the custom udev rule. If we need to wait for that, that is easier to fix on Talos side (should be).

Ulexus:
Yes, the symlink does appear (inspection via talosctl verifies both that it is there and that it is correct), but apparently not by the time the mount is tried (it fails with a no such file error).

The weird thing for me is why it worked before but not now.  The only difference now that I see is that udevd is restarted immediately prior to the mount attempt.

frezbo:
I was reproducing this yesterday and it seems we've had more bugs with udevd anyways, just that somehow it never populated, working on a better fix.

frezbo:
[...]

PR #7139, opened 2023-04-26
fix: udevd rules trigger

frezbo:
Fix udevd not triggering rules properly.
This also fixes an issue with go-blockdevice not resolving symlinks.

Fixes: #7117


frezbo:
This can be tested by creating a `patch.yaml` as below:

```yaml
machine:
  udev:
    rules:
      - SUBSYSTEM=="block", KERNEL=="vdb*", SYMLINK+="mygreathdd%n"
  disks:
    - device: /dev/mygreathdd
      partitions:
        - mountpoint: /var/mnt/extras

```

then creating a local qemu cluster using:

```bash
sudo -E _out/talosctl-linux-amd64 cluster create --provisioner=qemu $REGISTRY_MIRROR_FLAGS --masters=1 --workers=1 --with-bootloader=true --extra-disks=1 --config-patch-worker @patch.yaml
```

Once cluster is up this can be verfied:

```bash
❯ talosctl -n 10.5.0.3 read /proc/mounts | grep mygreathdd
/dev/mygreathdd1 /var/mnt/extras xfs rw,noatime,attr2,inode64,logbufs=8,logbsize=32k,noquota 0 0

❯ talosctl -n 10.5.0.3 mounts | grep mygreathdd
10.5.0.3   /dev/mygreathdd1   5.30       0.07       5.23            1.34%          /var/mnt/extras

❯ talosctl -n 10.5.0.3 ls -l /dev | grep mygreathdd
10.5.0.3   Lrwxrwxrwx    0     0     3         Apr 26 18:50:30   mygreathdd -> vdb
10.5.0.3   Lrwxrwxrwx    0     0     4         Apr 26 18:50:30   mygreathdd1 -> vdb1
```

PR #7168, opened 2023-05-02
fix: set the static pod priority as values

smira:
API server takes care of setting priority for "regular" pods from priorityClassName, but nothing does that for static pods, so we have to specify the priotity explicitly for static pods.

This fixes the graceful node shutdown (kubelet) to stop non-critical pods before the api-server and friends (critical pods).

Also some logging fixes.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -31,6 +31,9 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/resources/v1alpha1"
 )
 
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
frezbo, line 1: tests doesn't need update?

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
smira: good point... I tested it to the death manually, but the tests don't test every field in the spec

PR #7173, opened 2023-05-04
fix: inhibit timer to follow kubelet timer

salkin:
ensure to wait as long as possibly given to kubelet shutdown timers. Related to fix of siderolabs#7138


smira:
nice catch, thank you!

by the way that long shutdown timeout was probably related to wrong shutdown order - apiserver dying too early, but anyways, a good fix

salkin:
Yes, I phased issue when I tried to add 5 minute kubelet shutdown timer, Talos though aborted at 1 min after starting kubelet graceful shutdown without this fix

smira:
Re-pushed with small fixes as #7176, as I can't push back to `nokia` fork.

Issue #7142, opened 2023-04-26
SideroLink controller: reconnect on connection down


smira:
SideroLink controller:

* watch Wireguard connection status, if it's down for > N seconds, retry the SideroLink flow -- connect to the API, get the endpoints, and reconfigure Wireguard

Issue #7143, opened 2023-04-26
SideroLink controller: refactor to have input config resource


smira:
Introduce a controller which populates SideroLink config resource based on kernel args (in the future we'll add another config source).

Make SideroLink controller consume that config resource instead of reading directly from the boot kernel parameters.

utkuozdemir:
Closed by https://github.com/siderolabs/talos/pull/7175.

PR #7175, opened 2023-05-04
feat: introduce siderolink config resource & reconnect

utkuozdemir:
Introduce a new resource, `SiderolinkConfig`, to store SideroLink connection configuration (api endpoint for now).

Introduce a controller for this resource which populates it from the Kernel cmdline.

Rework the SideroLink `ManagerController` to take this new resource as input and reconfigure the link on changes.

Additionally, if the siderolink connection is lost, reconnect to it and reconfigure the links/addresses.

Closes https://github.com/siderolabs/talos/issues/7142, https://github.com/siderolabs/talos/issues/7143.


internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -248,7 +255,92 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
...
+	return nil
+}
+
+// nodeUUID returns node UUID from the provided function or from SMBIOS.
+func (ctrl *ManagerController) nodeUUID() (string, error) {
+	if ctrl.NodeUUIDFunc != nil {
+		return ctrl.NodeUUIDFunc()
+	}
+
+	s, err := smbios.GetSMBIOSInfo()
```
frezbo, line 287: don't we have this info already popualted in the hardware info controller?

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -110,20 +115,18 @@ func parseAPIEndpoint(sideroLinkParam string) (apiEndpoint, error) {
...
 //
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
 	if ctrl.Cmdline == nil || ctrl.Cmdline.Get(constants.KernelParamSideroLink).First() == nil {
 		// no SideroLink command line argument, skip controller
 		return nil
 	}
 
-	s, err := smbios.GetSMBIOSInfo()
```
smira, line 120: we should better replace this with fetching from the resource https://github.com/siderolabs/talos/blob/cbf6dc1009ad47a2804774839e4e0301efa8ac78/pkg/machinery/resources/hardware/system_information.go#L28-L37

utkuozdemir: yep, makes sense. will do.

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -248,7 +255,92 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			return fmt.Errorf("error creating address spec: %w", err)
 		}
 
-		// all done, terminate controller
-		return nil
+		return ctrl.pollConnection(ctx, logger)
```
frezbo, line 258: wouldn't this re-create the `wgclient` every time, should we cache it?

utkuozdemir, line 258: Note: this logic will again change on the implementation of this, which I'm already working on: https://github.com/siderolabs/talos/issues/7143

utkuozdemir: This polling is kinda long-lived: as long as the connection stays alive, it'll keep using the same `wgClient`. I'm not sure if further caching is necessary tbh.

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -231,6 +234,10 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			return fmt.Errorf("error creating siderolink spec: %w", err)
 		}
 
+		if err = ctrl.clearSiderolinkAddresses(ctx, r); err != nil {
```
smira, line 237: we can pair on that, but I feel it might be cleaner, and more controller-like, if we introduce a formal `for{}` loop here which will set up the connection, check its status, and re-setup if the connection is down

internal/app/machined/pkg/controllers/siderolink/manager.go
smira: I posted below, but my point it's easier to get "right" structure now

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: Got it, you mean also without failing the controller to reconnect, right?

smira: well, to some extent. if the Wireguard connection goes down, it's not a failure, we should just re-setup and inject new endpoint/address.

Also one more thing I forgot to mention - when cleaning up `AddressSpec` resources, and it should probably clean them up always, it should only clean up its own (where `Metadata().Owner() == ctrl.Name()`)

utkuozdemir: @smira It seems the owner of the `AddressSpec` is `AddressMergeController` here - not sure what to do about it.

utkuozdemir: oh right 🤦 always forget about that.

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: switched to use it.

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: done

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: done

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: This changed a bit as well - now the client lives as long as the controller does not crash - so it is even longer lived now.

internal/app/machined/pkg/controllers/siderolink/config.go
```diff
@@ -0,0 +1,92 @@
...
+		{
+			Type: siderolink.ConfigType,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+// Run implements controller.Controller interface.
+func (ctrl *ConfigController) Run(ctx context.Context, r controller.Runtime, _ *zap.Logger) error {
+	if err := ctrl.updateConfig(ctx, r); err != nil {
```
smira, line 49: there's no need for this block, controller always starts with an event queued in `r.EventCh()`

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -5,40 +5,55 @@
...
 // ManagerController interacts with SideroLink API and brings up the SideroLink Wireguard interface.
 type ManagerController struct {
-	Cmdline *procfs.Cmdline
-
 	nodeKey wgtypes.Key
 }
 
+// NewManagerController creates a new ManagerController.
+func NewManagerController() (*ManagerController, error) {
+	nodeKey, err := wgtypes.GeneratePrivateKey()
```
smira, line 47: can we push it down to the first step of `ManagerController.Run()`? as the controller struct survives the restarts, we can see if the key is zero, and create a new one (once).

or we could simply use `sync.Once` as well.

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -110,59 +137,56 @@ func parseAPIEndpoint(sideroLinkParam string) (apiEndpoint, error) {
...
-
-	sideroLinkClient := pb.NewProvisionServiceClient(conn)
+	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
 			return nil
+		case <-ticker.C:
```
smira, line 160: there's a small problem here - the reconnect logic now runs on any change to inputs, and some of that might be fine, but not all of changes are fine - e.g. `NetworkStatus` changes should not force a reconnect probably.

to address this issue, you could split up the code a bit to switch inputs: first, wait for the network to be ready, then drop it from the inputs.

example can be found here: https://github.com/siderolabs/talos/blob/339986db9d3675b78ce0d268f799ad654862fb0f/internal/app/machined/pkg/controllers/k8s/kubelet_service.go#L74-L145

internal/app/machined/pkg/controllers/siderolink/config.go
utkuozdemir: Got it, was about to test that to reply @frezbo 's comment.

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: Yes, I was not very comfortable about that input. Splitting makes sense.

smira: controller-runtime allows to change inputs at any point in the lifetime of the controller, the only caveat is that it doesn't ever reset it back to `Inputs()` (that is only initial setting, when controller is created). 

once inputs are updated, it makes sense to `r.QueueReconcile()` to trigger at least a single run, but after that any change to inputs will trigger controller normal way.

internal/app/machined/pkg/controllers/siderolink/config.go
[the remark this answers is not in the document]
utkuozdemir: yes, I was unsure if it would receive an event on startup - turns out it does. Removed.

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: sure, done.

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: Updated to do exactly that.

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -110,80 +107,173 @@ func parseAPIEndpoint(sideroLinkParam string) (apiEndpoint, error) {
...
+		var transportCredentials credentials.TransportCredentials
+
+		if parsedEndpoint.Insecure {
+			transportCredentials = insecure.NewCredentials()
+		} else {
+			transportCredentials = credentials.NewTLS(&tls.Config{})
+		}
+
+		provision := func() (*pb.ProvisionResponse, error) {
+			conn, connErr := grpc.DialContext(ctx, parsedEndpoint.Host, grpc.WithTransportCredentials(transportCredentials))
```
smira, line 254: I would probably set here some timeout on the context, just to make sure we don't hang for too long, and better retry

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -196,10 +286,14 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			return fmt.Errorf("error parsing node address: %w", err)
 		}
 
-		if err = r.Modify(ctx,
-			network.NewLinkSpec(network.ConfigNamespaceName, network.LayeredID(network.ConfigOperator, network.LinkID(constants.SideroLinkName))),
-			func(r resource.Resource) error {
-				spec := r.(*network.LinkSpec).TypedSpec()
+		if err = ctrl.cleanup(ctx, r, linkSpec.Metadata()); err != nil {
```
smira, line 289: here it's more of a nit, but it'd be nice if we could not delete the `siderolink` link and addresses if they don't change.

that might need some refactoring once again to make sure things work correctly. 

I would try to imagine this controller outputting resources: LinkSpec & AddresSpec, and a cleanup which should follow it - remove owned LinkSpec & AddressSpec which are no longer touched by the controller.

One case which doesn't work right now is that if `siderolink.Config` is dropped, this controller should tear down the link & address

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -247,8 +341,51 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			}); err != nil {
 			return fmt.Errorf("error creating address spec: %w", err)
 		}
```
smira, line 343: it might be nice to add some logging here for "success" case, e.g. `logger.Info("siderolink connection configured", zap.String("endpoint", endpoint), ...`

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: > I would try to imagine this controller outputting resources: LinkSpec & AddresSpec, and a cleanup which should follow it - remove owned LinkSpec & AddressSpec which are no longer touched by the controller.

👍 

> One case which doesn't work right now is that if `siderolink.Config` is dropped, this controller should tear down the link & address

Actually we do that: https://github.com/siderolabs/talos/pull/7175/files#diff-37ca5158dd35364ca3b988bd5c946b7cb8488d0a595212a6dbff527072de904cR216-R218

Tested as well: https://github.com/siderolabs/talos/pull/7175/files#diff-cd9530d76c3a8d8f60adfb9f54af2b6feff998e01388da405db06ccf92bc51aaR148-R170

Or do you mean something else?

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: done

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: > I would try to imagine this controller outputting resources: LinkSpec & AddresSpec, and a cleanup which should follow it - remove owned LinkSpec & AddressSpec which are no longer touched by the controller.

This is also done now.

smira (approved):

internal/app/machined/pkg/controllers/siderolink/manager.go
utkuozdemir: done

PR #7186, opened 2023-05-05
fix: properly skip/cleanup controlplane configs for workers

smira:
This bug is pretty cosmetic, but it shows up as a wrong check when performing worker upgrade - Talos pretends it checks e.g. kube-apiserver version which doesn't make sense for workers.

There were two bugs in the code:

* check for machine type was done against `TypeWorker`, while `MachineType` resource is initially created as `TypeUnknown`
* the cleanup code was not implemented

As I touched the code, I updated controller and tests to use modern conventions.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/slices"
 	"github.com/siderolabs/go-pointer"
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/k8s/control_plane.go
```diff
@@ -121,10 +119,8 @@ func (ctrl *ControlPlaneController) Run(ctx context.Context, r controller.Runtim
 			return fmt.Errorf("error getting machine type: %w", err)
 		}
 
-		machineType := machineTypeRes.(*config.MachineType).MachineType()
-
-		if machineType == machine.TypeWorker {
-			if err = ctrl.teardownAll(ctx, r, logger); err != nil {
+		if !machineType.MachineType().IsControlPlane() {
+			if err = ctrl.teardownAll(ctx, r); err != nil {
```
frezbo, line 123: why would we need to delete the resources if it's not controlplane, it shouldn't be there in the first place, right?

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/control_plane.go
smira: I tried to explain it above, the previous check got into the state when resource was created for worker nodes while `MachineType` was in `TypeUnknown`. This code is more for completeness, once we support changing `MachineType` on the fly one day.

frezbo: that makes sense. cool

PR #7215, opened 2023-05-13
chore: use ctest.UpdateWithConflicts instead of plain UpdateWithConflicts

DmitriyMV:
More type-safety.

### DIFF
--- a/internal/app/machined/pkg/controllers/cri/seccomp_profile_test.go
+++ b/internal/app/machined/pkg/controllers/cri/seccomp_profile_test.go
@@ -123,9 +123,7 @@ func (suite *CRISeccompProfileSuite) TestReconcileSeccompProfile() {
 		},
 	})
 
-	ctest.UpdateWithConflicts(suite, cfg, func(mc *config.MachineConfig) error {
-		return nil
-	})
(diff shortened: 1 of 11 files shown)


internal/app/machined/pkg/controllers/cri/seccomp_profile_test.go
```diff
@@ -123,9 +123,7 @@ func (suite *CRISeccompProfileSuite) TestReconcileSeccompProfile() {
 		},
 	})
 
-	ctest.UpdateWithConflicts(suite, cfg, func(mc *config.MachineConfig) error {
-		return nil
-	})
+	ctest.UpdateWithConflicts(suite, cfg, func(mc *config.MachineConfig) error { return nil })
```
frezbo, line 126: this is a no-op right, since it's not updating anything

frezbo (approved):

internal/app/machined/pkg/controllers/cri/seccomp_profile_test.go
DmitriyMV: Yep - I think I was reformatting code and covered this too.

PR #7237, opened 2023-05-17
feat: create azure community gallery image version on release

ro11net:
Create Azure Community Gallery Image Version on release:
- Add /hack/cloud-image-uploader/azure.go
	- Upload vhd file to container for all architectures
	- Create managed disk from vhd file for all architectures
	- Create image version from managed disk for all architectures
- Modify /hack/cloud-image-uploader/main.go
	- Start Community Gallery processes concurently with AWS upload
- Modify /hack/cloud-image-uploader/options.go
	- Add additional Options for Community Gallery processes
- Modify .drone.jsonnet to use secrets for environment variables
	- The following secrets need to be created for this to work:
		- azure_subscription_id
		- azure_client_id
		- azure_client_secret
		- azure_tenant_id

# Pull Request

## Acceptance


hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package main
+
+import (
```
smira, line 7: this is a nit, but we need to apply general Go formatting here, probably easier to set up an editor to do that for you automatically (most modern IDEs do it out of the box).

also we probably need to set up linting process to lint such files.

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	subscriptionID       string
+	cred                 *azidentity.DefaultAzureCredential
+	authorizer           autorest.Authorizer
+	storageClientFactory *armstorage.ClientFactory
+	accountsClient       *armstorage.AccountsClient
+	providersClient      resources.ProvidersClient
+	locations         = make(map[string]Location)
+	resourceGroupName = "SideroGallery"
+	defaultRegion = "eastus"
+	galleryName = "SideroLabs"
```
smira, line 42: the bottom part should be `const` probably, while the top part of variables should better be part of the `AzureUploader` struct and used as members, this provides better scoping

In general, we shouldn't use global variables if possible

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	var err error
+
+	var g *errgroup.Group
+	g, ctx = errgroup.WithContext(ctx)
+
+	fmt.Println("Setting default creds")
+
+	err = setDefaultAzureCreds()
+	if err != nil {
+		log.Fatalf("Error: %v\n", err)
```
smira, line 73: this aborts the whole process in a non-graceful way, this and other cases should be fixed to:

```suggestion
		return fmt.Errorf("error setting (getting?) default Azure credentials: %w", err)
```

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	Options Options
+}
+
+func (azu *AzureUploader) AzureGalleryUpload(ctx context.Context) error {
+	var err error
+
+	var g *errgroup.Group
+	g, ctx = errgroup.WithContext(ctx)
+
+	fmt.Println("Setting default creds")
```
smira, line 69: let's stay consistent with `aws.go`, ans use `log.Printf`, and no uppercase first chracters

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		log.Printf("Error: %v\n", err)
+	}
+
+	source, err := os.Open(azu.Options.AzureImage(arch))
+	if err != nil {
+		return err
+	}
+
+	defer func(source *os.File) {
```
smira, line 133: this can be simplified to 

```suggestion
	defer source.Close() //nolint:errcheck
```

we're not really expecting errors here

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	totalSize := int64(reader.Len())
+
+	// Check if the file size is a multiple of 512 bytes
+	if totalSize%pageblob.PageBytes != 0 {
+		panic("The file size must be a multiple of 512 bytes")
+	}
+
+	_, err = pageBlobClient.Create(ctx, totalSize, nil)
+	if err != nil {
+		log.Printf("Error: %v\n", err)
```
smira, line 161: should abort/return error

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	providersClient.Authorizer = authorizer
+
+	return nil
+}
+
+func (azu *AzureUploader) createAzureDisk(ctx context.Context, templateFile string, arch string) error {
+
+	diskParameters := map[string]interface{}{
+		"disk_name": map[string]string{
+			"value": "talos-" + arch + "-" + azu.Options.Tag,
```
smira, line 255:
```suggestion
			"value": fmt.Sprintf("talos-%s-%s",  arch, azu.Options.Tag)
```

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	err := deployResourceFromTemplate(ctx, subscriptionID, templateFile, diskParameters, deploymentName)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+	}
+	return err
+}
+
+func (azu *AzureUploader) createAzureImageVersion(ctx context.Context, templateFile string, arch string) error {
+
+	var g *errgroup.Group
```
smira, line 282: this is not being used at all

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+		"targetRegions": map[string]interface{}{
+			"value": targetRegions,
+		},
+	}
+
+	deploymentName := "imgversion-talos-" + arch + "-" + azu.Options.Tag
+
+	err := deployResourceFromTemplate(ctx, subscriptionID, templateFile, versionParameters, deploymentName)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
```
smira, line 324: no need to print an error

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+		return err
```
smira, line 325: better to wrap an error here, 
```suggestion
		return fmt.Errorf("error creating image version: %w", err)
```

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+
+	reader := bytes.NewReader(data)
+	totalSize := int64(reader.Len())
+
+	// Check if the file size is a multiple of 512 bytes
+	if totalSize%pageblob.PageBytes != 0 {
+		panic("The file size must be a multiple of 512 bytes")
+	}
+
+	_, err = pageBlobClient.Create(ctx, totalSize, nil)
```
smira, line 159: this whole upload function with chunking seems to be really complicated, and this is all already implemented in the SDK

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob#example-package-Client_UploadStream

they have options for concurrency, page size, etc. do we really need to re-implement it?

hack/cloud-image-uploader/azure-image-version-template.json
```diff
@@ -0,0 +1,61 @@
...
+            "replicaCount": 1,
+            "excludeFromLatest": false,
+            "storageAccountType": "Standard_LRS",
+            "replicationMode": "Full"
+          },
+          "storageProfile": {
+            "osDiskImage": {
+              "hostCaching": "None",
+              "source": {
+                "id": "[concat('/subscriptions/7f739b7d-f399-4b97-9a9f-f1962309ee6e/resourceGroups/SIDEROGALLERY/providers/Microsoft.Compute/disks/', parameters('disk_name'))]"
```
smira, line 51: this mix of hardcoded and provided variables feels like a weird mix to me.

I think we could have these files, `//go:embed` them, and use Go templating to populate all values, this will be more readable/maintainable in the future.

hack/cloud-image-uploader/azure.go
ro11net: The link you shared uses the UploadStream function which will upload the image file as a Block Blob. The file must be uploaded as a Page Blob to used to create managed disks and images. There does not seem to be anything similar to UploadStream in the SDK for Page Blobs.

hack/cloud-image-uploader/azure-image-version-template.json
ro11net: The way these arm templates are written are exactly the way Azure wants them formatted. When deploymentsClient.BeginCreateOrUpdate is called, it uses the variables passed in with it to fill in the parameters. I can look into using go:embed, but maintaining these, only requires you to log on to the Portal and grab a new example template from an existing resource with updated API versions or syntax.

hack/cloud-image-uploader/azure.go
smira: ok, makes sense, in that case the code needs to be rewritten, and simplieid, as it's turned inside out:

* number of goroutines should be number of concurrent uploaders
* channel should contain "work" items - [start_offset, end_offset)
* each worker picks up the work from the channel, uploads and looks for the next check

There's no point to create 1000 goroutines and make them all wait on a semaphore.

hack/cloud-image-uploader/azure.go
```diff
@@ -0,0 +1,420 @@
...
+	if err != nil {
+		log.Printf("Error: %v\n", err)
+	}
+
+	// Maximum size of the request body for a single UploadPages call is 4 MiB
+	chunkSize := 4 * 1024 * 1024
+	chunks := int(math.Ceil(float64(totalSize) / float64(chunkSize)))
```
smira, line 166:
```suggestion
	chunks := int((totalSize + chunkSize -1)/chunkSize)
```

`chunkSize` should be a `const`

frezbo (approved):

smira:
> /m

hm... we haven't tested this in the CI this is a bit unexpected merge

PR #7238, opened 2023-05-18
chore: bump deps

DmitriyMV:
- github.com/benbjohnson/clock to v1.3.5
- cilium-cli to v0.14.3

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/BurntSushi/toml v1.2.1
 	github.com/aws/aws-sdk-go v1.44.264
 	github.com/beevik/ntp v1.0.0
-	github.com/benbjohnson/clock v1.3.4
+	github.com/benbjohnson/clock v1.3.5 // project archived on 2023-05-18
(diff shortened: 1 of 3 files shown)


smira (approved):

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/BurntSushi/toml v1.2.1
 	github.com/aws/aws-sdk-go v1.44.264
 	github.com/beevik/ntp v1.0.0
-	github.com/benbjohnson/clock v1.3.4
+	github.com/benbjohnson/clock v1.3.5 // project archived on 2023-05-18
```
frezbo, line 35: is this something of concern in the longer term?

frezbo (approved):

go.mod
DmitriyMV: Yes, there are still bugs in it, so we should probably look for something else at some point. People suggest [this](https://github.com/jonboulle/clockwork) but we'll have to investigate.

PR #7239, opened 2023-05-18
feat: enable xfs project quota support, kubelet feature

### LINKED ISSUE #7181: support mount options for EPHEMERAL partition/user disks
## Feature Request

E.g. allow to enable XFS quota tracking with `pquota` mount option.

smira:
This is controlled with a feature flag which gets enabled automatically for Talos 1.5+.

Fixes #7181

If enabled, configures kubelet to use project quotas to track xfs volume usage, which is much more efficient than doing `du` periodically.


smira:
This PR requires Talos version 1.5+ to properly activate the version contract, so we need to release v1.5.0-alpha.0 first.

PR #7243, opened 2023-05-18
chore: update default image sizes to 10GB for all "cloud" images

rsmitty:
This PR adds a flag to imager that allows for tweaking the size of the created disk. Additionally, it sets the default value of that created disk to 10GB, as most images are cloud images that fail when uploaded b/c it only picks up a 1GB disk currently. Also adds some conditionals to the makefile to make sure we set the default small value for metal images and SBCs.

### DIFF
--- a/cmd/installer/cmd/image.go
+++ b/cmd/installer/cmd/image.go
@@ -61,9 +61,7 @@ func runImageCmd() (err error) {
 
 	log.Printf("creating image for %s", p.Name())
 
-	log.Print("creating RAW disk")
-
-	img, err := pkg.CreateRawDisk()
+	img, err := pkg.CreateRawDisk(p, options.DiskSize)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


frezbo (approved):

hack/imager.sh
```diff
@@ -17,3 +17,16 @@ prepare_extension_images() {
 
     echo "${extensions_dir}"
 }
+
+# returns the value to use for the raw disk image generated by imager.
+# we use the smallest possible value for metal and sbc, a 10GB value for all cloud images.
+determine_image_disk_size() {
```
smira, line 23: I wonder if we should put the default into Go code, this way you have the ability still to run `imager` with `docker run` without specifying the size? 

this might be easier to support and find, as these will be constants in the code. 

iirc we have anyways big `switch` statement for each platform in the code, we could put the size there

(we can keep the arg as well to override the default)

rsmitty: We create the raw disk before the switch statement, but I do think there's a better way to do this. Gimme just a few.

PR #7249, opened 2023-05-19
chore: add a pipeline to test cloud-images step without a release

smira:
Also uncomment Azure uploader.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -188,6 +188,23 @@ local Pipeline(name, steps=[], depends_on=[], with_docker=true, disable_clone=fa
   depends_on: [x.name for x in depends_on],
 };
 
+local creds_env_vars = {
+  // AWS creds
+  AWS_ACCESS_KEY_ID: { from_secret: 'aws_access_key_id' },
+  AWS_SECRET_ACCESS_KEY: { from_secret: 'aws_secret_access_key' },
+  AWS_SVC_ACCT: { from_secret: 'aws_svc_acct' },
+  // Azure creds
+  AZURE_SVC_ACCT: { from_secret: 'azure_svc_acct' },
+  AZURE_SUBSCRIPTION_ID: { from_secret: 'azure_subscription_id' },
+  AZURE_CLIENT_ID: { from_secret: 'azure_client_id' },
+  AZURE_CLIENT_SECRET: { from_secret: 'azure_client_secret' },
+  AZURE_TENANT_ID: { from_secret: 'azure_tenant_id' },
+  // TODO(andrewrynhard): Rename this to the GCP convention.
+  GCE_SVC_ACCT: { from_secret: 'gce_svc_acct' },
+  PACKET_AUTH_TOKEN: { from_secret: 'packet_auth_token' },
+  GITHUB_TOKEN: { from_secret: 'ghcr_token' },  // Use GitHub API token to avoid rate limiting on CAPI -> GitHub calls.
+};
+
 // Default pipeline.
 
 local external_artifacts = Step('external-artifacts', depends_on=[setup_ci]);
@@ -516,6 +533,7 @@ local integration_qemu_csi = Step('e2e-csi', target='e2e-qemu', privileged=true,
 
 local integration_images = Step('images', target='images', depends_on=[load_artifacts], environment={ IMAGE_REGISTRY: local_registry });
 local integration_sbcs = Step('sbcs', target='sbcs', depends_on=[integration_images], environment={ IMAGE_REGISTRY: local_registry });
+local integration_cloud_images =  Step('cloud-images', depends_on=[integration_images], environment=creds_env_vars);
 
 local integration_reproducibility_test = Step('reproducibility-test', target='reproducibility-test', depends_on=[load_artifacts], environment={ IMAGE_REGISTRY: local_registry });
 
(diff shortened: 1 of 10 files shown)


.drone.jsonnet
```diff
@@ -189,6 +189,23 @@ local Pipeline(name, steps=[], depends_on=[], with_docker=true, disable_clone=fa
...
 };
 
+local creds_env_vars = {
+  // AWS creds
+  AWS_ACCESS_KEY_ID: { from_secret: 'aws_access_key_id' },
+  AWS_SECRET_ACCESS_KEY: { from_secret: 'aws_secret_access_key' },
+  AWS_SVC_ACCT: { from_secret: 'aws_svc_acct' },
+  // Azure creds
+  AZURE_SVC_ACCT: { from_secret: 'azure_svc_acct' },
+  AZURE_SUBSCRIPTION_ID: { from_secret: 'azure_subscription_id' },
```
ro11net, line 199: Is there anything else in addition to this for it to read the environment variables?
It appears to be failing because it cannot find the AZURE_SUBSCRIPTION_ID environment variable. I have verified the four secrets needed for the cloud images upload (azure_client_id, azure_client_secret, azure_subscription_id, azure_tenant_id) are in the siderolabs org secrets with Pull Request Read permissions.

smira: probably makes sense to track it down via the Makefile: https://github.com/siderolabs/talos/blob/ff11fd39c723a40c01abe6348f64b1f892856175/Makefile#L308-L316

ro11net: Yep, just found that. I'll add them to that.

smira:
@ro11net you can only promote when the build finishes, and PR should be rebased on top of `main` (this one is not rebased), and you do that with the PR comment like I did above

ro11net:
This time it failed because the new image size of 10GiB is too large for the P2 disk tier. I changed the disk template to match Azure's P3 tier and match the P3 template size of 16GiB.

ro11net:
New issue. In local testing I was using versions of int.int.int without realizing Azure's image version names are actually limited to: Major(int).Minor(int).Patch(int) format, where int is between 0 and 2,147,483,647 (inclusive). e.g: 1.0.0, 2018.12.1 etc.

If this was a release we could drop the `v` and put just the version number in there, but it's currently failing because it can't create a version name of `v1.5.0-alpha.0-4-xxxxxxxxxx` and dropping the other characters wouldn't differentiate between releases and tests run the pipeline. I'm going to think on this. I'm open to ideas.

smira:
> What do you think about using a separate private Compute Gallery for testing?
> 
> If we do that, we could drop all the extra characters aside from the version number and push the version being tested to the testing Gallery. I believe that could be accomplished by dropping the extra characters on the version and add the gallery name as an additional cli option to give us the capability to use separate galleries in testing vs production from the pipeline.

separate gallery makes sense for all releases which do not match 'vX.Y.Z' pattern, or even between release CI runs

Makefile
```diff
@@ -308,9 +308,10 @@ talosctl-cni-bundle: ## Creates a compressed tarball that includes CNI bundle fo
 .PHONY: cloud-images
 cloud-images: ## Uploads cloud images (AMIs, etc.) to the cloud registry.
 	@docker run --rm -v $(PWD):/src -w /src \
-		-e TAG=$(TAG) -e ARTIFACTS=$(ARTIFACTS) \
+		-e TAG=$(TAG) -e ARTIFACTS=$(ARTIFACTS) -e ABBREV_TAG=$(ABBREV_TAG) \
```
ro11net, line 311: Added ABBREV_TAG to environment variable here and in cloud-image-uploader.sh passed as an argument for the tag to be compared without the build metadata.

hack/cloud-image-uploader/azure-disk-template.json
```diff
@@ -45,7 +45,7 @@
             "storageAccountId": "[concat('/subscriptions/7f739b7d-f399-4b97-9a9f-f1962309ee6e/resourceGroups/SideroGallery/providers/Microsoft.Storage/storageAccounts/', parameters('storage_account'))]",
             "sourceUri": "[concat('https://siderogallery.blob.core.windows.net/images/talos/', parameters('vhd_name'))]"
           },
-          "diskSizeGB": 10,
+          "diskSizeGB": 8,
```
ro11net, line 48: Disk size is chaned to 8GB to be within the limits of the P2 Azure disk tier. The price of storing per disk doubles when moving to P3 (up to 16GB). The disk size of Talos cloud images was also changed to match in: [https://github.com/siderolabs/talos/pull/7252](https://github.com/siderolabs/talos/pull/7252).

hack/cloud-image-uploader/azure.go
```diff
@@ -238,13 +265,13 @@ uploadLoop:
 func (azu *AzureUploader) createAzureDisk(ctx context.Context, armTemplate []byte, arch string) error {
 	diskParameters := map[string]interface{}{
 		"disk_name": map[string]string{
-			"value": "talos-" + arch + "-" + azu.Options.Tag,
+			"value": fmt.Sprintf("talos-%s-%s%s", arch, azu.Options.AzureCoreTag, azu.Options.AzurePreRelease),
```
ro11net, line 268: Append `-prerelease` to the core version for the disk image if it is a prerelease to keep the source consistent if overwritten during tests. If it is not a pre-release, this value is set to an empty string above.

ro11net (approved):
Added steps to Azure install documentation for using Community Gallery images.

rsmitty:
needs a rebase, then :shipit:

smira:
> needs a rebase, then :shipit:

I can take care of that

PR #7254, opened 2023-05-21
(EXPERIMENT) feat: prompt openai api when talosctl command fails

utkuozdemir:
Ask OpenAI for suggestions with the debug info.

### DIFF
--- a/cmd/talosctl/main.go
+++ b/cmd/talosctl/main.go
@@ -6,17 +6,251 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
 	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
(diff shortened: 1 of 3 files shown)


cmd/talosctl/main.go
```diff
@@ -6,17 +6,251 @@
...
+func readTalosconfig() (map[string]any, error) {
+	talosConfig, err := config.Open("")
+	if err != nil {
+		return nil, err
+	}
+
+	for _, configContext := range talosConfig.Contexts {
+		configContext.CA = ""
+		configContext.Crt = ""
+		configContext.Key = ""
```
frezbo, line 241: might as well the tokens when using an omni downloaded talosctl

utkuozdemir: Yes, but this is nothing to be reviewed to be merged, rather about the rough idea.

PR #7264, opened 2023-05-23
refactor: split config.Provider into Config & Container

smira:
See #7230

This is a step towards preparing for multi-doc config.

Split the `config.Provider` interface into parts which have different implementation:

* `config.Config` accesses the config itself, it might be implemented by `v1alpha1.Config` for example
* `config.Container` will be a set of config documents, which implement validation, encoding, etc.

`Version()` method dropped, as it makes little sense and it was almost not used.

`Raw()` method renamed to `RawV1Alpha1()` to support legacy direct access to `v1alpha1.Config`, next PR will refactor more to make it return proper type.

There will be many more changes coming up.


cmd/talosctl/pkg/mgmt/helpers/wireguard.go
```diff
@@ -97,7 +97,7 @@ type WireguardConfigBundle struct {
 
 // PatchConfig generates config patch for a node and patches the configuration data.
 func (w *WireguardConfigBundle) PatchConfig(ip fmt.Stringer, cfg config.Provider) (config.Provider, error) {
-	config := cfg.Raw().(*v1alpha1.Config).DeepCopy()
+	config := cfg.RawV1Alpha1().(*v1alpha1.Config).DeepCopy()
```
frezbo, line 100: is it possible to make this more typed so that we can avoid type conversion in future PR's?

frezbo (approved):

cmd/talosctl/pkg/mgmt/helpers/wireguard.go
smira: yes, in the next PR, I need to move more things around to avoid import loop

PR #7278, opened 2023-05-25
chore: add ukify Go script

smira:
This is a port of ukify.py from systemd.


Dockerfile
```diff
@@ -135,6 +143,15 @@ RUN --mount=type=cache,target=/.cache cd /go/src/github.com/siderolabs/structpro
     && go build -o structprotogen . \
     && mv structprotogen /toolchain/go/bin/
 COPY --from=importvet /importvet /toolchain/go/bin/importvet
+COPY ./hack/ukify /go/src/github.com/siderolabs/ukify
+RUN --mount=type=cache,target=/.cache \
+    --mount=type=bind,source=pkg,target=/go/src/github.com/pkg \
```
frezbo, line 148: requires access to Talos repo for constants

Dockerfile
```diff
@@ -135,6 +143,14 @@ RUN --mount=type=cache,target=/.cache cd /go/src/github.com/siderolabs/structpro
...
+    cd /go/src/github.com/siderolabs/ukify \
+    && go build -o gen-certs ./gen-certs \
+    && CGO_ENABLED=1 go build -o ukify . \
```
frezbo, line 151: tpm simulator needs Cgo

Dockerfile
```diff
@@ -135,6 +143,14 @@ RUN --mount=type=cache,target=/.cache cd /go/src/github.com/siderolabs/structpro
     && go build -o structprotogen . \
     && mv structprotogen /toolchain/go/bin/
 COPY --from=importvet /importvet /toolchain/go/bin/importvet
+COPY ./hack/ukify /go/src/github.com/siderolabs/ukify
+RUN --mount=type=cache,target=/.cache \
+    --mount=type=bind,source=pkg,target=/go/src/github.com/pkg \
+    cd /go/src/github.com/siderolabs/ukify \
+    && go build -o gen-certs ./gen-certs \
```
smira, line 150: nit: should we make the binary name more specific to the purpose, e.g. `gen-uki-certs` ?

hack/ukify/measure/measure.go
```diff
@@ -0,0 +1,259 @@
...
+	sim, err := simulator.Get()
+	if err != nil {
+		return nil, fmt.Errorf("creating tpm2 simulator failed: %v", err)
+	}
+
+	defer sim.Close()
+
+	for _, section := range constants.OrderedSections() {
+		if file, ok := sectionData[section]; ok && file != "" {
+			if err := pcrExtent(sim, pcr, alg, append([]byte(section), []byte("\x00")...)); err != nil {
```
smira, line 73: this is better:

```suggestion
			if err := pcrExtent(sim, pcr, alg, append([]byte(section), 0)); err != nil {
```

but this might also change the underlying `section` slice... so to be on the safe side we could also do:

```suggestion
			if err := pcrExtent(sim, pcr, alg, append(append([]byte(nil), []byte(section)...), 0)); err != nil {
```

hack/ukify/measure/measure.go
```diff
@@ -0,0 +1,259 @@
...
+// signatureData returns the hashed signature digest and base64 encoded signature
+type signatureData struct {
+	Digest          string
+	SignatureBase64 string
+}
+
+// SectionData holds a map of Section to file path to the corresponding section
+type SectionsData map[constants.Section]string
+
+func calculatePCRBankData(pcr int, alg tpm2.Algorithm, sectionData SectionsData, privateKeyFile string) ([]bankData, error) {
```
smira, line 55: can we do some unit-test with known data here? to make sure it generates same hashes/signatures always?

smira (commented):
:cool: lgtm, I can't approve my own PR :)

smira (commented):
:cool: lgtm, I can't approve my own PR :)

hack/ukify/measure/measure.go
frezbo: yes, I'll add a code that actually invokes `sd-measure` and compare output (but that means we need to keep sd-measure in tools if we want to run this in CI? Or have both, one with testdata and expected output and another that can be run locally on a system that has swtpm and sd-measure.

frezbo (approved):

PR #7281, opened 2023-05-26
chore: registry.k8s.io for coredns image

EirikAskheim:
# Pull Request

## What? (description)

Replace docker.io with registry.k8s.io for the coredns image.

## Why? (reasoning)

registry.k8s.io is the official kubernetes registry. Dockerhub has a pull limit for unauthenticated image pulls and this can cause outages if you spend your entire quota.

For these reasons I think registry.k8s.io is a better default for most Talos users.

## Acceptance

### DIFF
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -325,11 +325,11 @@ const (
 	KubernetesSchedulerImage = "registry.k8s.io/kube-scheduler"
 
 	// CoreDNSImage is the enforced CoreDNS image to use.
-	CoreDNSImage = "docker.io/coredns/coredns"
+	CoreDNSImage = "registry.k8s.io/coredns/coredns"
 
 	// DefaultCoreDNSVersion is the default version for the CoreDNS.
 	// renovate: datasource=github-releases depName=coredns/coredns
-	DefaultCoreDNSVersion = "1.10.1"
+	DefaultCoreDNSVersion = "v1.10.1"
 
 	// LabelNodeRoleControlPlane is the node label required by a control plane node.
 	LabelNodeRoleControlPlane = "node-role.kubernetes.io/control-plane"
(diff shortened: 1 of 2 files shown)


smira:
Is there any documentation on `registry.k8s.io` mirroring `coredns`? 

The official Helm chart seems to be still using `docker.io`: https://github.com/coredns/helm/blob/03226f99ed78fe5018c528abaa69201b34688e71/charts/coredns/values.yaml#LL6C6-L6C6

EirikAskheim:
https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io#managing-kubernetes-container-registries

> Requirements
> The rule of thumb is that staging repositories should be used to host artifacts produced by code that is part of the Kubernetes project, as defined by presence in one of the [project-owned GitHub Organizations](https://git.k8s.io/community/github-management#project-owned-organizations).
> 
> In other words, code that is not part of the Kubernetes project should not have its artifacts hosted in staging repos. SIG K8s Infra may make exceptions to this policy on a case-by-case basis.
> 
> For example:
> 
> CRI-O is not part of the kubernetes project, it does not meet the requirements to get a staging repo
> While etcd and coredns are not part of the kubernetes project, we do bundle them with kubernetes as part of the release, so for this specific case are allowing a staging repo to host them (solely within the context of the kubernetes project)

https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io/images/k8s-staging-coredns

In the images.yaml specific images are mirrored to registry.k8s.io. So as I understand this, Talos can use registry.k8s.io as it's source for these images if it is always in lock step with Kubernetes as other versions does not get published to this image registry.

smira:
> In the images.yaml specific images are mirrored to registry.k8s.io. So as I understand this, Talos can use registry.k8s.io as it's source for these images if it is always in lock step with Kubernetes as other versions does not get published to this image registry.

thank you, it looks like mirroring for new version is done within some reasonable time, so it should be good to go.

PR #7289, opened 2023-05-29
refactor: rewrite code to include preliminary support for multi-doc

smira:
`config.Container` implements a multi-doc container which implements both `Container` interface (encoding, validation, etc.), and `Conifg` interface (accessing parts of the config).

Refactor `generate` and `bundle` packages to support multi-doc, and provide backwards compatibility.

Implement a first (mostly example) machine config document for SideroLink API URL.

Many places don't properly support multi-doc yet (e.g. config patches).


internal/app/machined/pkg/controllers/files/etcfile_test.go
```diff
@@ -153,11 +153,12 @@ func (suite *EtcFileSuite) TearDownTest() {
 	suite.ctxCancel()
 
 	suite.wg.Wait()
-
```
frezbo, line 156: these are not needed anymore?

pkg/machinery/config/bundle/bundle.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/config/configloader"
+	"github.com/siderolabs/talos/pkg/machinery/config/configpatcher"
+	"github.com/siderolabs/talos/pkg/machinery/config/encoder"
+	"github.com/siderolabs/talos/pkg/machinery/config/generate"
+	"github.com/siderolabs/talos/pkg/machinery/config/machine"
+)
+
+// Bundle defines a set of machine configuration files.
+type Bundle struct {
+	InitCfg         config.Provider
```
frezbo, line 25: can we drop this now?

pkg/machinery/config/types/siderolink/siderolink.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/config/internal/registry"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/meta"
+)
+
+//go:generate deep-copy -type ConfigV1Alpha1 -pointer-receiver -header-file ../../../../../hack/boilerplate.txt -o deep_copy.generated.go .
+
+// Kind is a siderolink config document kind.
+const Kind = "SiderolinkConfig"
```
frezbo, line 19:
```suggestion
const Kind = "SideroLinkConfig"
```

frezbo (approved):

internal/app/machined/pkg/controllers/files/etcfile_test.go
smira: in fact I made `go test -v ./...` pass on Talos repo without root - it skips the tests which are not applicable/can't run without root or some tools.

PR #7290, opened 2023-05-29
feat: use stable network interface names

### LINKED ISSUE #5681: Use predictable network interface names
## Feature Request
Use predictable network interface names (`eno1`, `eno2`, etc).

### Description
Talos currently uses the legacy scheme with interface names `eth0`, `eth1`, etc. However, the name of an interface is unpredictable, and can change after reboots. This caused me to try and bind a VIP to the wrong interface. Other projects that need to deal with this use predicable network interface names. Some related documentation about this:

https://github.com/harvester/harvester/issues/2073
https://wiki.debian.org/NetworkInterfaceNames
https://www.freedesktop.org/software/systemd/man/systemd.net-naming-scheme.html
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-understanding_the_predictable_network_interface_device_names
https://specs.openstack.org/openstack/fuel-specs/specs/8.0/network-interfaces-naming-schemes.html

### LINKED ISSUE #6756: allow MAC addresses to be used in the machine config where link names (ethX) are supported
## Feature Request

Device selectors are not available for bond slaves, and that makes hard to match a specific link.

smira:
Use `udevd` rules to create stable interface names.

Link controllers should wait for `udevd` to settle down, otherwise link rename will fail (interface should not be UP).

Fixes #5681
Fixes #6756


hack/udevd/80-net-name-slot.rules
frezbo, line 1: this file is already is already in udev, we just need to add it back https://github.com/siderolabs/pkgs/blob/main/eudev/pkg.yaml#L49

hack/release.toml
```diff
@@ -40,6 +40,25 @@ it is set to true for new clusters.
...
+
+* firmware/BIOS provided index numbers for on-board devices (example: `eno1`)
+* firmware/BIOS provided PCI Express hotplug slot index numbers (example: `ens1`)
+* physical/geographical location of the connector of the hardware (example: `enp2s0`)
+* interfaces's MAC address (example: `enx78e7d1ea46da`)
+
+The predictable network interface names features can be disabled by specifying `net.ifnames=0` in the kernel command line.
+Talos automatically adds the `net.ifnames=0` kernel argument when upgrading from Talos versions before 1.5.
+
+This change doesn't affect "cloud" platforms, like AWS, as Talos automatically adds `net.ifnames=0` to the kernel command line.
```
frezbo, line 61: is this a requirement from the cloud?

pkg/machinery/compatibility/talos_version.go
```diff
@@ -39,6 +39,15 @@ func (v *TalosVersion) String() string {
 	return v.version.String()
 }
 
+// DisablePredictableNetworkInterfaces returns true if predictable network interfaces should be disabled on upgrade.
+func (v *TalosVersion) DisablePredictableNetworkInterfaces() bool {
```
frezbo, line 43: a test for this maybe?

frezbo (approved):

hack/udevd/80-net-name-slot.rules
smira: nope, I modified it to support `enx`

hack/release.toml
smira: It's more about the way some of the clouds are configured - they provide index-based network metadata, so `eth%d` is way better there. on other hand most of them don't have really two network cards

internal/app/machined/pkg/controllers/network/link_spec.go
```diff
@@ -58,6 +53,19 @@ func (ctrl *LinkSpecController) Outputs() []controller.Output {
 //
 //nolint:gocyclo
 func (ctrl *LinkSpecController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	// wait for udevd to be healthy, which implies that all link renames are done
+	if err := v1alpha1.WaitForServiceHealthy(ctx, r, "udevd",
```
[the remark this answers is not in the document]
smira: I believe so, `udevd` does rename as part of the rule processing - it either renames or fails to do so

pkg/machinery/compatibility/talos_version.go
smira: added a test

.drone.jsonnet
```diff
@@ -295,6 +295,7 @@ local load_artifacts = {
     AZURE_TENANT: { from_secret: 'az_tenant' },
   },
   commands: [
+    'mkdir -p _out/',
```
frezbo, line 298: it doesn't create?

frezbo (approved):

Issue #7261, opened 2023-05-22
Building UEFI ISO for UKI


frezbo:
## Feature Request

Build an UEFI supported iso with UKI

Ref: http://www.panticz.de/create-uefi-iso

Sample code used to create iso:

```bash
#!/usr/bin/env bash

set -eux -o pipefail

ISO_DIR=$(mktemp -d)

ISO_SIZE="300M"

function cleanup() {
    set +ue
    rm -rf "${ISO_DIR}"
}

rm -f secureboot-poc.iso

trap cleanup SIGINT EXIT

mkdir -p "${ISO_DIR}/EFI/BOOT"
mkdir -p "${ISO_DIR}/EFI/Linux"

sbsign --key certs/uki-signing.key --cert certs/uki-signing.crt boot/systemd-bootx64.efi --output "${ISO_DIR}/EFI/BOOT/BOOTX64.efi"

# dd if=/dev/zero of="${ISO_DIR}/EFI/BOOT/efiboot.img" bs=1M count=100
# mkfs.msdos -F 32 -m "TalosISO" "${ISO_DIR}/EFI/BOOT/efiboot.img"
truncate -s "${ISO_SIZE}" "${ISO_DIR}/EFI/BOOT/efiboot.img"
parted -s "${ISO_DIR}/EFI/BOOT/efiboot.img" mklabel gpt
parted -s "${ISO_DIR}/EFI/BOOT/efiboot.img" unit MiB mkpart primary fat32 0% 100%
parted -s "${ISO_DIR}/EFI/BOOT/efiboot.img" set 1 boot on

mkfs.fat "${ISO_DIR}/EFI/BOOT/efiboot.img"
fatlabel "${ISO_DIR}/EFI/BOOT/efiboot.img" ESP

mmd -i "${ISO_DIR}/EFI/BOOT/efiboot.img" ::EFI
mmd -i "${ISO_DIR}/EFI/BOOT/efiboot.img" ::EFI/BOOT
mmd -i "${ISO_DIR}/EFI/BOOT/efiboot.img" ::EFI/Linux

./ukify.py vmlinuz-amd64 \
    initramfs-amd64.xz \
    --stub=boot/linuxx64.efi.stub \
    --uname=6.1.26-talos \
    --splash=boot/sidero.bmp \
    --os-release=@boot/os-release \
[...]

PR #7297, opened 2023-05-30
feat: add secureboot iso generation

frezbo:
This adds code to generate a SecureBoot enabled ISO.

Fixes: #7261

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -193,10 +193,12 @@ local Pipeline(name, steps=[], depends_on=[], with_docker=true, disable_clone=fa
 local external_artifacts = Step('external-artifacts', depends_on=[setup_ci]);
 local generate = Step('generate', target='generate docs', depends_on=[setup_ci]);
 local check_dirty = Step('check-dirty', depends_on=[generate, external_artifacts]);
-local build = Step('build', target='talosctl-all kernel initramfs installer imager talos _out/integration-test-linux-amd64', depends_on=[check_dirty], environment={ IMAGE_REGISTRY: local_registry, PUSH: true });
+local uki_certs = Step('uki-certs', depends_on=[setup_ci], environment={ PLATFORM: 'linux/amd64' });
+local build = Step('build', target='talosctl-all kernel initramfs installer imager talos _out/integration-test-linux-amd64', depends_on=[check_dirty, uki_certs], environment={ IMAGE_REGISTRY: local_registry, PUSH: true });
 local lint = Step('lint', depends_on=[build]);
 local talosctl_cni_bundle = Step('talosctl-cni-bundle', depends_on=[build, lint]);
 local iso = Step('iso', target='iso', depends_on=[build], environment={ IMAGE_REGISTRY: local_registry });
+local iso_uki = Step('iso-uki', target='iso-uki', depends_on=[build], environment={ IMAGE_REGISTRY: local_registry });
 local images_essential = Step('images-essential', target='images-essential', depends_on=[iso], environment={ IMAGE_REGISTRY: local_registry });
 local unit_tests = Step('unit-tests', target='unit-tests unit-tests-race', depends_on=[build, lint]);
 local e2e_docker = Step('e2e-docker-short', depends_on=[build, unit_tests], target='e2e-docker', environment={ SHORT_INTEGRATION_TEST: 'yes', IMAGE_REGISTRY: local_registry });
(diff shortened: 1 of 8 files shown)


.drone.jsonnet
```diff
@@ -193,7 +193,8 @@ local Pipeline(name, steps=[], depends_on=[], with_docker=true, disable_clone=fa
 
 local generate = Step('generate', target='generate docs', depends_on=[setup_ci]);
 local check_dirty = Step('check-dirty', depends_on=[generate]);
-local build = Step('build', target='talosctl-all kernel initramfs installer imager talos _out/integration-test-linux-amd64', depends_on=[check_dirty], environment={ IMAGE_REGISTRY: local_registry, PUSH: true });
+local uki_certs = Step('uki-certs', depends_on=[check_dirty], environment={ PLATFORM: 'linux/amd64' });
```
smira, line 196: does it affect _dirtiness_? do we not gitignore it?

hack/ukify/main.go
```diff
@@ -298,7 +298,15 @@ func run() error {
 		return err
 	}
 
-	return buildUKI(sdStub, output, sections)
+	if err := buildUKI(sdStub, output, sections); err != nil {
+		return err
+	}
+
+	if _, err := sbSign(output); err != nil {
```
smira, line 305: nit:

```suggestion
	_, err := sbSign(output)
	
	return err
```

cmd/installer/pkg/install/manifest.go
```diff
@@ -132,6 +132,14 @@ func NewManifest(label string, sequence runtime.Sequence, bootPartitionFound boo
 					Source:      fmt.Sprintf(constants.InitramfsAssetPath, opts.Arch),
 					Destination: filepath.Join(constants.BootMountPoint, label, constants.InitramfsAsset),
 				},
+				{
+					Source:      "/usr/install/amd64/systemd-boot.efi.signed",
+					Destination: "/boot/EFI/EFI/BOOT/BOOTX64.EFI",
```
smira, line 137: something temporary to be removed?

frezbo: I'm not sure if we should do this now or later, otherwise after installing booting off from this iso, it won't use UKI. Or we only care about iso generation here in this PR?

smira: but isn't this code common already, so it would affect any install of Talos?

my concern is that it doesn't use proper constants, it's hardcoded even if we don't use new secure boot/trusted boot.

smira: I understand that, but it needs to be done proper way - it should be optional, we should do a proper layout for the partitions, etc.

we need to detect and act differently depending on the way Talos got booted

frezbo: yeh, I'm not sure how yet, maybe we add a new kernel param and installer checks it?

.drone.jsonnet
smira: so we add a step, but nothing depends on it? should we just add it as a target to `generate` ? or run it in parallel, and make `check_dirty` depend on it?

smira: so we add a step, but nothing depends on it? should we just add it as a target to `generate` ? or run it in parallel, and make `check_dirty` depend on it?

cmd/installer/pkg/iso.go
```diff
@@ -42,3 +46,79 @@ func CreateISO(iso, dir string) error {
...
+
+	if err := os.MkdirAll(isoDir, 0o755); err != nil {
+		return err
+	}
+
+	defer os.RemoveAll(isoDir) // nolint:errcheck
+
+	efiBootImg := filepath.Join(isoDir, "efiboot.img")
+
+	if _, err := cmd.Run("dd", "if=/dev/zero", "of="+efiBootImg, "bs=1M", "count=120"); err != nil {
```
smira, line 63: if we want to create a file which is 120MiB, we can do it easier with go:

```go
   f, _ := os.Create(out)
   f.Seek(120 * MB, os.SeekStart)
   f.Truncate()
   f.Close()
```

hack/ukify/main.go
```diff
@@ -298,7 +298,13 @@ func run() error {
 		return err
 	}
 
-	return buildUKI(sdStub, output, sections)
+	if err := buildUKI(sdStub, output, sections); err != nil {
+		return err
+	}
+
+	_, err = sbSign(output)
```
smira, line 305: are we signing the whole UKI once again? :)

smira (approved):

Makefile
```diff
@@ -324,6 +324,13 @@ iso: ## Builds the ISO and outputs it to the artifact directory.
 		docker run --rm -e SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) -i $(REGISTRY_AND_USERNAME)/imager:$(IMAGE_TAG) iso --arch $$arch --tar-to-stdout $(IMAGER_ARGS) | tar xz -C $(ARTIFACTS)  ; \
 	done
 
+iso-uki: ## Builds UEFI only ISO which uses UKI and outputs it to the artifact directory.
```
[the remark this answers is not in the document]
frezbo: do we need it as a default for `make`?

hack/ukify/main.go
frezbo: yes, that's needed, I forgot earlier :sweat_smile: (the kernel signing is also needed along with this)

Issue #7191, opened 2023-05-08
In-cluster internal Kubernetes controlplane endpoint


smira:
Run `go-loadbalancer` based TCP loadbalancer on `localhost:<some-port>` with the following endpoints:

* cluster endpoint (from machine config)
* `localhost:6443` for cp nodes
* `talosctl get members` | `grep controlplane` -> control plane endpoints + `:6443`

(we need to know control plane port, which might not be available on workers (?))

We can use this endpoint for `kubeconfig-in-cluster` -> `kube-proxy` DaemonSet.

`kubelet` on worker nodes could use this Kubernetes endpoint (or even controlplane nodes).

`go-loadbalancer` runs the health checks against all endpoints, and picks only the ones which are reachable.

frezbo:
Should we also include the vip endpoint?

smira:
VIP endpoint is the controlplane endpoint, but we don't know about VIPs on the worker nodes (only controlplane endpoint).

ruifung:
I just saw this, and I'm curious if this will work independent of kube-proxy? I.e. for cilium in it's kube-proxy free mode which requires specifying the control plane endpoint and port.

frezbo:
> I just saw this, and I'm curious if this will work independent of kube-proxy? I.e. for cilium in it's kube-proxy free mode which requires specifying the control plane endpoint and port.

yes, that's one the use cases, you'd specify this endpoint for cilium and since it's localhost it would just happily work.

PR #7307, opened 2023-06-01
chore: add endpoints balancer controller

DmitriyMV:
This PR adds support for creating a list of API endpoints (each is pair of host and port).

It gets them from
- Machine config cluster endpoint.
- Localhost with LocalAPIServerPort if machine is control panel.
- Other affiliates if they are control panels.

For #7191


DmitriyMV:
Waits for https://github.com/siderolabs/discovery-api/pull/4 and https://github.com/siderolabs/discovery-client/pull/7

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
```diff
@@ -0,0 +1,184 @@
...
+				Port: uint32(machineConfig.Config().Cluster().LocalAPIServerPort()),
+			})
+		}
+
+		for it := safe.IteratorFromList(affiliates); it.Next(); {
+			affiliateSpec := it.Value().TypedSpec()
+
+			if identity.TypedSpec().NodeID == affiliateSpec.NodeID {
+				// skip adding ourselves to the list
+				continue
```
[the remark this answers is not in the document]
DmitriyMV: Solved.

DmitriyMV:
`/home/dmitriymv/.talos/clusters/talos-default/talos-default-controlplane-1.log`
image
for
image

pkg/machinery/resources/cluster/endpoints.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// EndpointsType is type of Endpoints resource.
```
smira, line 18: I feel we should move this resource to `k8s/`, as it's really specific there (with the loadbalancer itself later).

And I would it call it some more specific way, e.g. `APIServerEndpoints`, as we have many other endpoints as well.

pkg/machinery/resources/cluster/affiliate.go
```diff
@@ -82,6 +82,14 @@ type AffiliateSpec struct {
...
 	MachineType     machine.Type          `yaml:"machineType" protobuf:"6"`
 	KubeSpan        KubeSpanAffiliateSpec `yaml:"kubespan,omitempty" protobuf:"7"`
+	ControlPlane    *ControlPlane         `yaml:"controlPlane,omitempty" protobuf:"8"`
+}
+
+// ControlPlane describes ControlPlane data if any.
+//
+//gotagsrewrite:gen
+type ControlPlane struct {
+	Port int `yaml:"port" protobuf:"1"`
```
smira, line 92: same as with discovery data:

```suggestion
	APIServerPort int `yaml:"port" protobuf:"1"`
```

internal/app/machined/pkg/controllers/cluster/member.go
```diff
@@ -48,30 +50,28 @@ func (ctrl *MemberController) Outputs() []controller.Output {
...
 				// not a cluster member
 				continue
 			}
 
-			if err = r.Modify(ctx, cluster.NewMember(cluster.NamespaceName, affiliateSpec.Nodename), func(res resource.Resource) error {
-				spec := res.(*cluster.Member).TypedSpec()
+			if err = safe.WriterModify(ctx, r, cluster.NewMember(cluster.NamespaceName, affiliateSpec.Nodename), func(res *cluster.Member) error {
+				spec := res.TypedSpec()
 
 				spec.Addresses = slices.Clone(affiliateSpec.Addresses)
```
smira, line 76: I feel we should extract control plane's API server port from affiliates to members

internal/app/machined/pkg/controllers/cluster/local_affiliate.go
```diff
@@ -108,188 +114,204 @@ func (ctrl *LocalAffiliateController) Outputs() []controller.Output {
...
-			if err != nil && !state.IsNotFoundError(err) {
-				return fmt.Errorf("error getting kubespan identity: %w", err)
-			}
+			continue
+		}
 
-			kubespanConfig, err := safe.ReaderGetByID[*kubespan.Config](ctx, r, kubespan.ConfigID)
-			if err != nil && !state.IsNotFoundError(err) {
-				return fmt.Errorf("error getting kubespan config: %w", err)
+		machineConfig, err := safe.ReaderGetByID[*config.MachineConfig](ctx, r, config.V1Alpha1ID)
```
smira, line 185: I'm not sure we should _require_ machine config here - we can still generate a local affiliate, but we can't fill only local API server port if we don't have it.

internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
```diff
@@ -35,6 +37,13 @@ func (suite *LocalAffiliateSuite) TestGeneration() {
 
 	suite.Require().NoError(suite.runtime.RegisterController(&clusterctrl.LocalAffiliateController{}))
 
+	mc := config.NewMachineConfig(container.NewV1Alpha1(&v1alpha1.Config{
+		ConfigVersion: "v1alpha1",
+		MachineConfig: &v1alpha1.MachineConfig{},
+		ClusterConfig: &v1alpha1.ClusterConfig{},
+	}))
+	suite.Require().NoError(suite.state.Create(suite.ctx, mc))
```
smira, line 45: we should assert on some control plane apiserver port in the test?

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
```diff
@@ -0,0 +1,184 @@
...
+// Inputs implements controller.Controller interface.
+func (ctrl *EndpointsBalancerController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineTypeType,
+			ID:        pointer.To(config.MachineTypeID),
+			Kind:      controller.InputWeak,
+		},
+		safe.Input[*cluster.Affiliate](controller.InputWeak),
```
smira, line 41: should depend on `Member`, not `Affiliate`

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
```diff
@@ -0,0 +1,184 @@
...
+			endpoints = append(endpoints, cluster.Endpoint{
+				Host: "localhost",
+				Port: uint32(machineConfig.Config().Cluster().LocalAPIServerPort()),
+			})
+		}
+
+		for it := safe.IteratorFromList(affiliates); it.Next(); {
+			affiliateSpec := it.Value().TypedSpec()
+
+			if identity.TypedSpec().NodeID == affiliateSpec.NodeID {
```
smira, line 118: not sure if we need to skip oursleves here? does it bring any benefit?

internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
DmitriyMV: No - if we go with optional machine config in previous suggestion, we can drop this block.

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
```diff
@@ -0,0 +1,184 @@
...
+			logger.Info("removed endpoints resource", zap.String("id", res.Metadata().ID()))
+		}
+
+		r.ResetRestartBackoff()
+	}
+}
+
+func toPort(port string) uint32 {
+	if port == "" {
+		return 0
```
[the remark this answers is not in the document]
DmitriyMV: Will do!

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/siderolabs/gen/channel"
+	"github.com/siderolabs/go-pointer"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/talos/pkg/machinery/config/machine"
+	"github.com/siderolabs/talos/pkg/machinery/resources/cluster"
+	"github.com/siderolabs/talos/pkg/machinery/resources/config"
+)
+
+// EndpointsBalancerController watches v1alpha1.Config, updates discovery config.
```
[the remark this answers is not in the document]
DmitriyMV: Will do!

internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
smira: I mean we should have a test that verifies that controlplane block gets properly populated :)

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
DmitriyMV: I think it will be kinda strange if we use both public address and localhost in the same list, no?

.golangci.yml
```diff
@@ -131,7 +131,8 @@ linters-settings:
     replace-local: true
     replace-allow-list:
       - gopkg.in/yaml.v3
-      - inet.af/tcpproxy
+      - github.com/siderolabs/discovery-api
```
smira, line 134: can be removed now?

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+			endpoints = append(endpoints, k8s.APIServerEndpoint{
+				Host: "localhost",
+				Port: uint32(machineConfig.Config().Cluster().LocalAPIServerPort()),
+			})
+		}
+
+		for it := safe.IteratorFromList(members); it.Next(); {
+			memberSpec := it.Value().TypedSpec()
+
+			if identity.TypedSpec().NodeID == memberSpec.NodeID {
```
smira, line 119: `identity` can be `nil` here... not sure what is the fix, but I feel we can safely add ourselves to the list.

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+				// skip adding ourselves to the list
+				continue
+			}
+
+			if len(memberSpec.Addresses) > 0 && memberSpec.ControlPlane != nil {
+				endpoints = append(endpoints, k8s.APIServerEndpoint{
+					Host: memberSpec.Addresses[0].String(),
```
smira, line 126: we should add every address here, not the first one. the loadbalancer will remove endpoitns which don't work

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+					Port: uint32(memberSpec.ControlPlane.APIServerPort),
+				})
+			}
+		}
+
+		logger.Info("updating endpoints", zap.Stringers("endpoints", endpoints))
```
smira, line 132: this will be too verbose logging, and probably not even needed, as one can do `talosctl get <type> -w` to see all updates

smira (approved):
the most important one is adding all addresses of each cp member for endpoints

internal/app/machined/pkg/controllers/cluster/discovery_service.go
```diff
@@ -497,3 +489,15 @@ func specAffiliate(affiliate *pb.Affiliate, endpoints []*pb.Endpoint) cluster.Af
 
 	return result
 }
+
+func fromPlane(plane *pb.ControlPlane) *cluster.ControlPlane {
```
frezbo, line 493: `fromControlPlane` is better? :shrug:

internal/app/machined/pkg/controllers/cluster/local_affiliate.go
frezbo: i think this is still not resolved, correct?

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+// Inputs implements controller.Controller interface.
+func (ctrl *APIServerEndpointsController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineTypeType,
+			ID:        pointer.To(config.MachineTypeID),
+			Kind:      controller.InputWeak,
+		},
+		safe.Input[*cluster.Member](controller.InputWeak),
```
frezbo, line 42: is this something new?

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,185 @@
...
+		identity, err := safe.ReaderGetByID[*cluster.Identity](ctx, r, cluster.LocalIdentity)
+		if err != nil {
+			if !state.IsNotFoundError(err) {
+				return fmt.Errorf("error getting local identity: %w", err)
+			}
+
+			continue
+		}
+
+		members, err := safe.ReaderListAll[*cluster.Member](ctx, r)
```
frezbo, line 94: should we conditionally check this based on if discovery is enabled (kubespan/k8s registry)?

pkg/machinery/resources/k8s/apiserver_endpoints.go
```diff
@@ -0,0 +1,75 @@
...
+// APIServerEndpointsExtension provides auxiliary methods for APIServerEndpoints.
+type APIServerEndpointsExtension struct{}
+
+// ResourceDefinition implements [typed.Extension] interface.
+func (APIServerEndpointsExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
+	return meta.ResourceDefinitionSpec{
+		Type:             APIServerEndpointsType,
+		Aliases:          []resource.Type{},
+		DefaultNamespace: NamespaceName,
+		PrintColumns:     []meta.PrintColumn{},
```
frezbo, line 64: better add a user friendly print column for host and port

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
DmitriyMV: Yep - it will create `Input` calling `ResourceDefinition` on given type.

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
DmitriyMV: I'm not sure. If discovery is disabled, we will get an empty list and will simply skip adding from that list later on.

internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
```diff
@@ -129,77 +125,54 @@ func (suite *LocalAffiliateSuite) TestGeneration() {
...
-				netip.MustParseAddrPort("172.20.0.2:51820"),
-				netip.MustParseAddrPort("10.5.0.1:51820"),
-				netip.MustParseAddrPort("1.1.1.1:51820"),
-				netip.MustParseAddrPort("[2001:123:4567::1]:51820"),
-			}, spec.KubeSpan.Endpoints)
-
-			return nil
-		}),
-	))
+	ctest.AssertResource(suite, nodeIdentity.TypedSpec().NodeID, func(r *cluster.Affiliate, asrt *assert.Assertions) {
```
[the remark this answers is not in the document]
DmitriyMV: Done in `TestCPGeneration` test.

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
DmitriyMV: Both `DiscoveryServiceController` and `KubernetesPushController` get local id from `*cluster.Identity` (and do not allow to continue if it's nil, same logic here) so I don't know what else we can use in `MemberSpec` to identify current node.

The problem with adding ourselves to the list is from load balancing perspective - we have `localhost+port` already in the list, so if we add ourselves agin, we two times more likely to go to ourselves than to others, no?

internal/app/machined/pkg/controllers/cluster/endpoints_balancer.go
smira: I don't think it matters, if the address is local to the node, it works technically same way as `localhost`.

internal/app/machined/pkg/controllers/k8s/apiserver_endpoints.go
smira: we will talk to ourselves always, as it's anyways in the "local" tier, no matter which address we're using. 

what I mean is that in the controller above iirc `identity` might be `nil` if it's not found.

DmitriyMV:
image

PR #7308, opened 2023-06-01
feat: add network chaos to qemu development environment

ro11net:
Add flags for configuring the qemu bridge interface with chaos options:
- network-chaos-enabled
- network-jitter
- network-latency
- network-packet-loss
- network-packet-reorder
- network-packet-corrupt
- network-bandwidth

These flags are used in /pkg/provision/providers/vm/network.go at the end of the CreateNetwork function to first see if the network-chaos-enabled flag is set, and then check if bandwidth is set.  This will allow developers to simulate clusters having a degraded WAN connection in the development environment and testing pipelines.

If bandwidth is not set, it will then enable the other options.
- Note that if bandwidth is set, the other options such as jitter, latency, packet loss, reordering and corruption will not be used.  This is for two reasons:
	- Restriction the bandwidth can often intoduce many of the other issues being set by the other options.
	- Setting the bandwidth uses a separate queuing discipline (Token Bucket Filter) from the other options (Network Emulator) and requires a much more complex configuration using a Heirarchial Token Bucket Filter which cannot be configured at a granular enough level using the vishvananda/netlink library.

Adding both queuing disciplines to the same interface may be an option to look into in the future, but would take more extensive testing and control over many more variables which I believe is out of the scope of this PR.  It is also possible to add custom profiles, but will also take more research to develop common scenarios which combine different options in a realistic manner.

# Pull Request

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -154,6 +154,13 @@ var (
 	controlPlanePort           int
 	dhcpSkipHostname           bool
 	skipBootPhaseFinishedCheck bool
+	networkChaos               bool
+	jitter                     time.Duration
+	latency                    time.Duration
+	packetLoss                 float64
+	packetReorder              float64
+	packetCorrupt              float64
+	bandwidth                  int
 )
 
 // createCmd represents the cluster up command.
@@ -269,6 +276,13 @@ func create(ctx context.Context, flags *pflag.FlagSet) (err error) {
 		}
 	}
 
+	// Validate network chaos flags
+	if !networkChaos {
+		if jitter != 0 || latency != 0 || packetLoss != 0 || packetReorder != 0 || packetCorrupt != 0 || bandwidth != 0 {
+			return fmt.Errorf("network chaos flags can only be used with --with-network-chaos")
+		}
+	}
+
 	provisioner, err := providers.Factory(ctx, provisionerName)
 	if err != nil {
 		return err
@@ -296,6 +310,13 @@ func create(ctx context.Context, flags *pflag.FlagSet) (err error) {
 			},
 			DHCPSkipHostname:  dhcpSkipHostname,
 			DockerDisableIPv6: dockerDisableIPv6,
+			NetworkChaos:      networkChaos,
+			Jitter:            jitter,
+			Latency:           latency,
+			PacketLoss:        packetLoss,
+			PacketReorder:     packetReorder,
+			PacketCorrupt:     packetCorrupt,
+			Bandwidth:         bandwidth,
 		},
 
 		Image:         nodeImage,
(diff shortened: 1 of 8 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -933,6 +954,13 @@ func init() {
 	createCmd.Flags().IntVar(&controlPlanePort, controlPlanePortFlag, constants.DefaultControlPlanePort, "control plane port (load balancer and local API port)")
 	createCmd.Flags().BoolVar(&dhcpSkipHostname, "disable-dhcp-hostname", false, "skip announcing hostname via DHCP (QEMU only)")
 	createCmd.Flags().BoolVar(&skipBootPhaseFinishedCheck, "skip-boot-phase-finished-check", false, "skip waiting for node to finish boot phase")
+	createCmd.Flags().BoolVar(&networkChaos, networkChaosFlag, false, "enable to use network chaos parameters when creating a qemu cluster")
+	createCmd.Flags().IntVar(&jitter, jitterFlag, 0, "specify jitter (in ms) on the bridge interface when creating a qemu cluster (default: 0)")
+	createCmd.Flags().IntVar(&latency, latencyFlag, 0, "specify latency (in ms) on the bridge interface when creating a qemu cluster (default: 0)")
```
smira, line 959: in Go, every time-related thing is better represented as `time.Duration`, and the flag of the same type

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -933,6 +954,13 @@ func init() {
...
+	createCmd.Flags().IntVar(&packetLoss, packetLossFlag, 0, "specify percent of packet loss on the bridge interface when creating a qemu cluster (default: 0)")
```
smira, line 960: this is more of a nit, but I feel it might be easier to represent that as `float` and a fraction, i.e `[0, 1)' range.

pkg/provision/providers/vm/network.go
```diff
@@ -129,6 +131,65 @@ func (p *Provisioner) CreateNetwork(ctx context.Context, state *State, network p
...
+	// configure bridge interface with network chaos if flag is set
+	if network.NetworkChaos {
+		link, err := netlink.LinkByName(state.BridgeName)
+		if err != nil {
+			return fmt.Errorf("could not get link: %v", err)
+		}
+
+		fmt.Fprintln(options.LogWriter, "network chaos enabled on interface:", state.BridgeName)
+
+		if network.Bandwidth != 0 {
```
smira, line 143: I guess it makes sense to return an error if both bandwidth & any of jitter/latency is set.

smira:
@ro11net any updates for this PR? I could help to plug it into testing pipeline

pkg/provision/providers/vm/network.go
ro11net: I added a line that prints a log message saying "bandwidth is set. No other chaos options will be used." 

Another option could be to set two flags instead of one to enable network chaos. One for bandwidth (tbf) qdiscs and one for the other other (netem) qdiscs. Currently, you set the flag `--network-chaos-enabled` and then set your other flags such as `network-jitter`, `network-latency`, `network-packet-loss`, `network-packet-reorder`, `network-packet-corrupt`, `network-bandwidth` and if bandwidth is set, it will just not apply any of the other flags.

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -933,6 +954,13 @@ func init() {
 	createCmd.Flags().IntVar(&controlPlanePort, controlPlanePortFlag, constants.DefaultControlPlanePort, "control plane port (load balancer and local API port)")
 	createCmd.Flags().BoolVar(&dhcpSkipHostname, "disable-dhcp-hostname", false, "skip announcing hostname via DHCP (QEMU only)")
 	createCmd.Flags().BoolVar(&skipBootPhaseFinishedCheck, "skip-boot-phase-finished-check", false, "skip waiting for node to finish boot phase")
+	createCmd.Flags().BoolVar(&networkChaos, networkChaosFlag, false, "enable to use network chaos parameters when creating a qemu cluster")
+	createCmd.Flags().DurationVar(&jitter, jitterFlag, 5*time.Millisecond, "specify jitter on the bridge interface when creating a qemu cluster (default: 5ms)")
```
ro11net, line 958: Changed to time.Duration

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -933,6 +954,13 @@ func init() {
...
+	createCmd.Flags().DurationVar(&latency, latencyFlag, 35*time.Millisecond, "specify latency on the bridge interface when creating a qemu cluster (default: 35ms)")
+	createCmd.Flags().Float64Var(&packetLoss, packetLossFlag, 0.0, "specify percent of packet loss on the bridge interface when creating a qemu cluster. e.g. 50% = 0.50 (default: 0.0)")
```
ro11net, line 960: Changed to float so user can set the value as a fraction. e.g. 50% would be input as .50

smira (approved):

frezbo (approved):

Unix4ever (approved):

PR #7309, opened 2023-06-01
Update deploy-metrics-server.md

kasimon:
# Pull Request

## What

Add hint about csr pki approval.

## Why? (reasoning)

Enabling certificate rotation on a freshly installed cluster requires the kublet CSRs  to be manually approved. This requirement may not be obvious to people new to kubernetes like me who wonder, why for example `kubectl logs` fails after enabling certificate rotation. 

## Acceptance

### DIFF
--- a/website/content/v1.5/kubernetes-guides/configuration/deploy-metrics-server.md
+++ b/website/content/v1.5/kubernetes-guides/configuration/deploy-metrics-server.md
@@ -21,6 +21,13 @@ machine:
       rotate-server-certificates: true
 ```
 
+This may require manually approving the kubelets certificate signing requests in the cluster PKI:
+
+```shell
+kubectl get csr --sort-by=.metadata.creationTimestamp
+kubectl certificate approve csr-xxxxx
+```
+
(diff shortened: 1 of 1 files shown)


website/content/v1.5/kubernetes-guides/configuration/deploy-metrics-server.md
```diff
@@ -21,6 +21,13 @@ machine:
       rotate-server-certificates: true
 ```
 
+This may require manually approving the kubelets certificate signing requests in the cluster PKI:
```
smira, line 24: this feels to be confusing a bit, or should be its own section, as the notes below describe how to install auto-approver.

I don't think a manual way is sustainable in the long run, but it might be some Note or special block which explains how to do it manually, but better to find automated solution

kasimon: You are totally right. I somehow managed to overlook the second part of the page, which completely solves this problem. Will close this PR.

PR #7315, opened 2023-06-02
feat: add support for multi-doc strategic merge patching

### LINKED ISSUE #7312: Patching for multi-doc config

smira:
Other changes:

* renamed `version:` to `apiVersion:` in the multi-doc format, as this better matches Kubernetes objects
* introduced (not used at the moment) a concept of `NamedDocuments` (many documents for the same type)
* added container validation on not having duplicate documents
* JSON6902 now denies multi-doc config

Fixes #7312

### DIFF
--- a/pkg/machinery/config/configloader/internal/decoder/decoder.go
+++ b/pkg/machinery/config/configloader/internal/decoder/decoder.go
@@ -20,8 +20,8 @@ import (
 var ErrMissingKind = errors.New("missing kind")
 
 const (
-	// ManifestVersionKey is the string indicating a manifest's version.
-	ManifestVersionKey = "version"
+	// ManifestAPIVersionKey is the string indicating a manifest's version.
+	ManifestAPIVersionKey = "apiVersion"
 	// ManifestKindKey is the string indicating a manifest's kind.
 	ManifestKindKey = "kind"
 	// ManifestDeprecatedKeyMachine represents the deprecated v1alpha1 manifest.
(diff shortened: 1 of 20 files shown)


pkg/machinery/config/configloader/internal/decoder/decoder.go
```diff
@@ -20,8 +20,8 @@ import (
 var ErrMissingKind = errors.New("missing kind")
 
 const (
-	// ManifestVersionKey is the string indicating a manifest's version.
-	ManifestVersionKey = "version"
+	// ManifestAPIVersionKey is the string indicating a manifest's version.
+	ManifestAPIVersionKey = "apiVersion"
```
frezbo, line 24: isn't this a breaking change, i mean for existing config?

pkg/machinery/config/configpatcher/apply_test.go
```diff
@@ -54,7 +61,72 @@ func TestApply(t *testing.T) {
 			bytes, err := out.Bytes()
 			require.NoError(t, err)
 
-			assert.Equal(t, expected, bytes)
+			assert.Equal(t, string(expected), string(bytes))
```
frezbo, line 64: assert equal doesn't work with bytes?

frezbo (approved):

pkg/machinery/config/configloader/internal/decoder/decoder.go
smira: nope, we never ever actually supported multi-doc before

pkg/machinery/config/configpatcher/apply_test.go
smira: it does work, but the diff shown is for bytes, which is very hard to read (it does a hexdump diff), while our data is text, so if I cast to string, the diff is much more sane

PR #7319, opened 2023-06-05
chore: optimize image compression

smira:
Use `pigz` (parallel) instead of `gzip`.

Use `xz` compression `-0` instead of `-6`.

This has pros and cons:

* image size goes up (77M -> 79M) (+2.5%)
* image generation goes down 19s -> 10s (-50%).

### DIFF
--- a/cmd/installer/cmd/image.go
+++ b/cmd/installer/cmd/image.go
@@ -279,7 +279,7 @@ func tar(filename, src, dir string) error {
 }
 
 func gz(filename string) error {
-	if _, err := cmd.Run("gzip", "-6", filename); err != nil {
+	if _, err := cmd.Run("pigz", "-6", filename); err != nil {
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

frezbo (approved):
I guess this shouldn't affect reproducability

smira:
> I guess this shouldn't affect reproducability

Most of our disk images are not reproducible, because they contain `ctime` (creation time) in the created filesystem attributes which can't be controlled/changed.

frezbo:
> > I guess this shouldn't affect reproducability
> 
> Most of our disk images are not reproducible, because they contain `ctime` (creation time) in the created filesystem attributes which can't be controlled/changed.

I think some FS has an option, at least fat32 has `--invariant` option :sweat_smile:

PR #7320, opened 2023-06-05
chore: add default console args

frezbo:
Add default console args for UKI iso.

### DIFF
--- a/hack/ukify/main.go
+++ b/hack/ukify/main.go
@@ -170,6 +170,9 @@ func run() error {
 		return err
 	}
 
+	defaultCmdline.Append("console", "ttyS0")
+	defaultCmdline.Append("console", "tty0")
+
 	flag.StringVar(&sdStub, "sd-stub", "_out/linuxx64.efi.stub", "path to sd-stub")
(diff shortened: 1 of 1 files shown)


hack/ukify/main.go
```diff
@@ -165,6 +165,8 @@ func Measure(tempDir, kernel, signingKey string, sections []section) ([]section,
 func run() error {
 	defaultCmdline := procfs.NewCmdline("")
 	defaultCmdline.Append(talosconstants.KernelParamPlatform, "metal")
+	defaultCmdline.Append("console", "ttyS0)")
```
smira, line 168:
```suggestion
	defaultCmdline.Append("console", "ttyS0")
```

smira (approved):

hack/ukify/main.go
smira: and it also makes sense to append first these: https://github.com/siderolabs/talos/blob/f0fadcc9cd45bb077e73d905a74023311c500732/pkg/machinery/kernel/kernel.go#L18-L33

PR #7323, opened 2023-06-05
chore: deprectae `bootloader` installer option

frezbo:
Deprecate the `bootloader` installer option. This has not been used in a long while.


cmd/installer/pkg/install/verify.go
```diff
@@ -33,10 +33,6 @@ func VerifyEphemeralPartition(opts *Options) (err error) {
 
 // VerifyBootPartition verifies the supplied boot device options.
 func VerifyBootPartition(opts *Options) (err error) {
-	if opts.Bootloader {
```
smira, line 36: as it's true always, we should remove the whole method, as it was no-op

smira:
we also need a note in `hack/release.toml`

smira (approved):

Issue #7324, opened 2023-06-05
QEMU support for SecureBoot


frezbo:
## Feature Request

QEMU support for `talosctl cluster create` to test and develop on SecureBoot/TrustedBoot

PR #7325, opened 2023-06-05
feat: qemu secureboot

frezbo:
Add qemu support for secureboot testing via `talosctl cluster create`.

Can be tested via:

```bash
sudo -E _out/talosctl-linux-amd64 cluster create --provisioner=qemu $REGISTRY_MIRROR_FLAGS --controlplanes=1 --workers=1 --iso-path=_out/talos-uki-amd64.iso --with-secureboot=true --with-tpm2=true --skip-injecting-config --with-apply-config
```

This currently only supports just booting Talos in SecureBoot mode. Installation and Upgrade comes as extra PRs.

Fixes: #7324

### DIFF
--- a/pkg/provision/providers/qemu/preflight.go
+++ b/pkg/provision/providers/qemu/preflight.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -31,6 +32,8 @@ func (p *provisioner) preflightChecks(ctx context.Context, request provision.Clu
 		checkContext.checkKVM,
 		checkContext.qemuExecutable,
 		checkContext.checkFlashImages,
+		checkContext.ovmfctlExecutable,
+		checkContext.swtpmExecutable,
 		checkContext.cniDirectories,
 		checkContext.cniBundle,
 		checkContext.checkIptables,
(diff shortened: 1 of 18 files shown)


pkg/provision/providers/qemu/preflight.go
```diff
@@ -75,7 +75,7 @@ func (check *preflightCheckContext) qemuExecutable(ctx context.Context) error {
 }
 
 func (check *preflightCheckContext) checkFlashImages(ctx context.Context) error {
-	for _, flashImage := range check.arch.PFlash(check.options.UEFIEnabled, check.options.ExtraUEFISearchPaths) {
```
smira, line 78: we should add a check for `swtpm` in the pre-flight checks if secureboot enabled

smira: and same with `ovmfctl`

pkg/provision/providers/qemu/tpm2.go
```diff
@@ -0,0 +1,56 @@
...
+	}, nil
+}
+
+func (p *provisioner) destroyVirtualTPM2s(cluster provision.ClusterInfo) error {
+	errCh := make(chan error)
+
+	nodes := append([]provision.NodeInfo{}, cluster.Nodes...)
+
+	for _, node := range nodes {
+		tpm2PidPath := filepath.Join(node.TPM2StateDir, "swtpm.pid")
```
smira, line 37: what if secure boot is not enabled? should we check for `TPM2StateDir` being set?

frezbo: tpm2 is independent of secureboot, `TPM2StateDir` is set when `--with-tpm2` is passed

smira: I mean this code runs for `cluster destroy`, so if `--with-secureboot` is not set, this code will try to join empty `node.TPM2StateDir`

frezbo: not sure I get that, `node.TPM2StateDir` is set only when `--with-tpm2` is passed

frezbo: oh, I see that now, got it

smira (approved):

PR #7329, opened 2023-06-06
chore: use go-blockdevice for zeroing partitions

frezbo:
Use the `go-blockdevice` library to zero partitions.

Also added a test that writes `ones` to the partition and verifies its zeroes after zeroing it.


frezbo:
```bash
❯ go test -exec sudo github.com/siderolabs/talos/internal/pkg/partition
ok      github.com/siderolabs/talos/internal/pkg/partition      0.174s
```

PR #7330, opened 2023-06-06
chore: cleanup partition code

frezbo:
Cleanup partition code to be explicit about `Format` and `Partition` options.

### DIFF
--- a/internal/pkg/partition/partition.go
+++ b/internal/pkg/partition/partition.go
@@ -4,3 +4,120 @@
 
 // Package partition provides common utils for system partition format.
 package partition
+
+import (
+	"log"
+
+	"github.com/dustin/go-humanize"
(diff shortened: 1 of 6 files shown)


internal/pkg/partition/partition.go
```diff
@@ -4,3 +4,112 @@
...
+			Size:          EFISize,
+		}
+
+		if uki {
+			partitionOptions.Size = EFIUKISize
+		}
+
+		return partitionOptions
+	case constants.BIOSGrubPartitionLabel:
+		return &Options{
```
smira, line 88: if we pass `uki` here, does it make sense to also `panic` if something requests BIOSGrub, Boot partition in `uki` mode?

smira (approved):

internal/pkg/partition/partition.go
frezbo: should never happen, but good to be on the safe side.

PR #7334, opened 2023-06-07
chore: update Go to 1.20.5, bump dependencies

smira:
Go dependencies, new pkgs, extras, etc.


hack/ukify/go.mod
frezbo, line 1: oh I think this is missing in the `make go-mod-outdated` target

frezbo (approved):

PR #7335, opened 2023-06-07
chore: generic boootloader implementation

frezbo:
This changes the bootloader code to be generic enough to support multiple bootloader implementations.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -11,12 +11,10 @@ import (
 
 	"github.com/siderolabs/go-blockdevice/blockdevice"
 	"github.com/siderolabs/go-procfs/procfs"
-	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/board"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader"
-	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub"
 	"github.com/siderolabs/talos/internal/pkg/meta"
 	"github.com/siderolabs/talos/internal/pkg/mount"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
@@ -85,11 +83,6 @@ type Installer struct {
 	options    *Options
 	manifest   *Manifest
 	bootloader bootloader.Bootloader
-
-	bootPartitionFound bool
-
-	Current grub.BootLabel
-	Next    grub.BootLabel
 }
 
 // NewInstaller initializes and returns an Installer.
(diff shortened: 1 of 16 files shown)


cmd/installer/pkg/install/manifest.go
```diff
@@ -245,7 +245,9 @@ func (m *Manifest) executeOnDevice(device Device, targets []*Target) (err error)
 	}
 
 	if device.Zero {
-		if err = m.zeroDevice(device); err != nil {
+		if err = partition.Format(device.Device, &partition.FormatOptions{
```
frezbo, line 248: re-using same code from  `partition`

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -340,46 +338,21 @@ func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*ma
...
-			}
-		}()
-
-		disk := s.Controller.Runtime().State().Machine().Disk(disk.WithPartitionLabel(constants.BootPartitionLabel))
-		if disk == nil {
-			return fmt.Errorf("boot disk not found")
-		}
-
-		conf, err := grub.Read(grub.ConfigPath)
+		disk, err := s.Controller.Runtime().Config().Machine().Install().Disk()
```
frezbo, line 341: another option is to pass `s.Controller.Runtime()` to `grub.Probe`, but that means it's not generic

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -340,46 +338,21 @@ func (s *Server) Rollback(ctx context.Context, in *machine.RollbackRequest) (*ma
...
 		if err != nil {
 			return err
 		}
 
-		if conf == nil {
-			return fmt.Errorf("grub configuration not found, nothing to rollback")
-		}
-
-		next, err := grub.FlipBootLabel(conf.Default)
+		config, err := grub.Probe(disk)
```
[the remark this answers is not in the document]
frezbo: hmm I might need some context around the code here. (Seems it runs on a boot/upgrade process failed in betweem :thinking: ). I guess I would need to split the the revert code to support different cases?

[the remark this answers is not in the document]
frezbo: okay, used the generic code here also, but not sure how to test part of code in internal revert

cmd/installer/pkg/install/install.go
```diff
@@ -112,68 +111,27 @@ func NewInstaller(cmdline *procfs.Cmdline, seq runtime.Sequence, opts *Options)
 }
 
 // Verify existence of boot partition.
-func (i *Installer) probeBootPartition() error {
-	// there's no reason to discover boot partition if the disk is about to be wiped
-	if !i.options.Zero {
```
smira, line 117: not sure why this got dropped...

cmd/installer/pkg/install/install.go
```diff
@@ -112,68 +111,27 @@ func NewInstaller(cmdline *procfs.Cmdline, seq runtime.Sequence, opts *Options)
...
 	}
 
-	grubConf, err := grub.Read(grub.ConfigPath)
+	i.Current, i.Next, err = config.Flip()
 	if err != nil {
 		return err
 	}
 
-	next := grub.BootA
+	i.bootLoaderFound = true
```
DmitriyMV, line 125:
```suggestion
	i.bootLoaderFound = config != nil
```

Maybe? And you can remove `i.bootLoaderFound = false` below.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/probe.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+)
+
+// Probe probes a block device for GRUB bootloader.
+func Probe() (*Config, error) {
+	// attempt to probe BOOT partition directly
+	dev, err := probe.GetDevWithPartitionName(constants.BootPartitionLabel)
+	if os.IsNotExist(err) {
+		// no BOOT partition, nothing to do
+		return nil, nil
```
DmitriyMV, line 23: Maybe instead of `nil, nil` we should return an error which we can check at `Probe` callsite? Or at least document that this function can return both `nils` if no BOOT partition exists?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/probe.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !alreadyMounted {
+		if err = mp.Mount(); err != nil {
+			return nil, err
+		}
+
+		defer mp.Unmount() //nolint:errcheck
```
DmitriyMV, line 52: I'm a bit torn about defer blocks which ignore errors - we have a lot of them. We can do the named return error trick like:

```suggestion
		defer func() {
			unmountErr := mp.Unmount()
			if unmountErr != nil {
				err = unmountErr
			}
		}
```

Here and above, but I'm not sure if we can do anything meaningful on `Unmount` error. But at the same time, should we continue the execution as if nothing happened?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/probe.go
frezbo: I'm currently again re-structuring this differently after talking with Andrey

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/probe.go
smira: this way we trade a (probably) unimportant error for a real failure of a function.. the better approach might be to make `.Unmount` idempotent and call it explicitly on "success" path with `return grubConf, mp.Unmount()`

smira (approved):
lgtm, we can fix some issues as we proceed to real sd-boot

PR #7338, opened 2023-06-08
fix: don't load RDMA over Ethernet driver by default

smira:
Document the change for upgrade notes.


smira:
Reported by @nberlee

hack/release.toml
```diff
@@ -66,6 +66,13 @@ This change doesn't affect "cloud" platforms, like AWS, as Talos automatically a
         description="""\
 The `.machine.install.bootloader` option in the machine config is deprecated and will be removed in Talos 1.6.
 This was a no-op for a long time. The bootloader is always installed.
+"""
+
+    [notes.rdma-rxe]
+        title = "RDMA/RoCE support"
+        description="""\
+Talos no longer loads by default `rdma_rxe` Linux driver, which is required for RoCE support.
+If the driver is required, it can be enabled by specifying `rdma_rxe` in the `.machine.kernel.modules` field in the machine config.
```
frezbo, line 75: udev could still pick it up :thinking: ?

smira: I believe it doesn't do any PCI ids, so it shouldn't

Issue #7341, opened 2023-06-08
nvidia-nonfree installer does not contain mellanox drivers


cbc02009:
### Description
I've followed the steps to create a full Talos installer from https://www.talos.dev/v1.4/talos-guides/configuration/nvidia-gpu-proprietary but after upgrading my node with the image it can no longer connect to the network because it doesn't recognize the Mellanox CTX3 NIC anymore.

Frezbo on slack mentioned that maybe we can run depmod on the custom installer to resolve this issue but they would need to look into it.

The installer can be viewed here: https://github.com/users/cbc02009/packages/container/package/installer

- Talos version: 1.4.5
- Platform: bare metal

PR #7342, opened 2023-06-08
chore: use fixed module list

frezbo:
Use a fixed list of modules to copy into Talos initramfs.

This makes sure we can still enable thing in Talos kernel as modules but not ship it as default in Talos (extra modules could be extensions).

Also fixes: #7341


Dockerfile
```diff
@@ -804,7 +832,7 @@ ONBUILD RUN xz -d /usr/install/${TARGETARCH}/initramfs.xz \
     && rm /usr/install/${TARGETARCH}/initramfs \
     && rm rootfs.sqsh
 ONBUILD COPY --from=customization / /rootfs
-ONBUILD RUN find /rootfs \
+ONBUILD RUN depmod -b /rootfs $(ls /rootfs/lib/modules) \
```
frezbo, line 835: this ensures custom builds still have correct `modules.dep`

smira (approved):

website/content/v1.5/talos-guides/configuration/nvidia-gpu-proprietary.md
```diff
@@ -64,7 +64,7 @@ COPY --from=ghcr.io/talos-user/kernel:{{< release >}}-nvidia /boot/vmlinuz /usr/
 Now build the image and push it to the registry.
 
 ```bash
-DOCKER_BUILDKIT=0 docker build --squash --build-arg RM="/lib/modules" -t ghcr.io/talos-user/installer:{{< release >}}-nvidia .
+DOCKER_BUILDKIT=0 docker build --squash -t ghcr.io/talos-user/installer:{{< release >}}-nvidia .
```
frezbo, line 67: we should not remove modules shipped with the kernel built, should only use that when using a custom kernel only

PR #7348, opened 2023-06-09
feat: support API server load balancer

### LINKED ISSUE #7191: In-cluster internal Kubernetes controlplane endpoint

DmitriyMV:
This commit adds support for API load balancer. Quick way to enable it is during cluster creation using new `api-server-balancer-port` flag (0 by default - disabled). When enabled all API request will be routed across
cluster control plane endpoints.

Closes #7191


DmitriyMV:
One of the interesting things I noticed while working on it - you can't call `ctrl.v1alpha1Runtime.Config().Machine().Features()` in `v1alpha2.Controller` - it will throw nil pointer panic. Maybe we should restrict interface here somehow?

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -81,6 +81,7 @@ const (
 	bootloaderEnabledFlag         = "with-bootloader"
 	forceEndpointFlag             = "endpoint"
 	controlPlanePortFlag          = "control-plane-port"
+	apiServerBalancerPortFlag     = "api-server-balancer-port"
```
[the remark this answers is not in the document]
frezbo: thinking more, the existing cilium strict test can be updated to enable this option, so we test an actual use case

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2605,6 +2605,30 @@ type FeaturesConfig struct {
...
+	//   description: |
+	//     Enable API server load balancer support - will start local proxy.
+	ServerEnabled *bool `yaml:"enabled,omitempty"`
+	//   description: |
+	//     API server load balancer port.
+	ServerPort int `yaml:"port,omitempty"`
+}
+
+// Enabled implements config.APIServerBalancer.
+func (a *APIServerBalancer) Enabled() bool {
```
smira, line 2625: this should go to `v1alpha1_features.go`, this file is structs only

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2605,6 +2605,30 @@ type FeaturesConfig struct {
...
+	return pointer.SafeDeref(a.ServerEnabled)
+}
+
+// Port implements config.APIServerBalancer.
+func (a *APIServerBalancer) Port() int {
+	return a.ServerPort
```
smira, line 2631: we should have a sane default here, otherwise setting to enabled makes it pick up port 0??

pkg/machinery/resources/k8s/k8s.go
```diff
@@ -8,7 +8,7 @@ package k8s
...
+//go:generate deep-copy -type LoadBalancerStatusesSpec -type AdmissionControlConfigSpec -type APIServerEndpointsSpec -type APIServerConfigSpec -type AuditPolicyConfigSpec -type BootstrapManifestsConfigSpec -type ConfigStatusSpec -type ControllerManagerConfigSpec -type EndpointSpec -type ExtraManifestsConfigSpec -type KubeletLifecycleSpec -type KubeletSpecSpec -type ManifestSpec -type ManifestStatusSpec -type NodeLabelSpecSpec -type KubeletConfigSpec -type NodeIPSpec -type NodeIPConfigSpec -type NodenameSpec -type SchedulerConfigSpec -type SecretsStatusSpec -type StaticPodSpec -type StaticPodStatusSpec -type StaticPodServerStatusSpec -header-file ../../../../hack/boilerplate.txt -o deep_copy.generated.go .
```
smira, line 11: it's a good idea to add new resource to `k8s_test.go`

cmd/talosctl/cmd/mgmt/cluster/create.go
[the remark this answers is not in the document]
DmitriyMV: Done [here](https://github.com/siderolabs/talos/pull/7348/files#diff-5b485902c8e136421d712cffb891b255489198da8c13f27c3790db75c808039b) I think?

hack/test/e2e-qemu.sh
```diff
@@ -160,6 +160,7 @@ function create_cluster {
     --with-init-node=false \
     --cni-bundle-url=${ARTIFACTS}/talosctl-cni-bundle-'${ARCH}'.tar.gz \
     --crashdump \
+    --api-server-balancer-port 13336 \
```
smira, line 163: like Noel said, this should be structured as optional in one of the pipelines, we don't want to enable not-default settings by default

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -152,6 +153,7 @@ var (
 	extraBootKernelArgs        string
 	dockerDisableIPv6          bool
 	controlPlanePort           int
+	apiServerBalancerPort      int
```
smira, line 156: need to add flag name to the function `checkForDefinedGenFlag` below

smira (commented):
needs a note in `hack/release.toml`

pkg/machinery/config/types/v1alpha1/v1alpha1_features.go
```diff
@@ -38,3 +38,28 @@ func (f *FeaturesConfig) ApidCheckExtKeyUsageEnabled() bool {
...
+// APIServerBalancer implements config.Features interface.
+func (f *FeaturesConfig) APIServerBalancer() config.APIServerBalancer {
+	if f.APIServerBalancerSupport == nil {
+		return &APIServerBalancer{}
+	}
+
+	return f.APIServerBalancerSupport
+}
+
+const defaultAPIBalancerPort = 7445
```
frezbo, line 51: maybe a much higher port in the 50000 range since that's what talos usually uses

DmitriyMV: @smira asked for a lower than 10000 port. We agreed on 7445 during standup call.

frezbo (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
DmitriyMV: Done - now 7445 port will be used as the default if `ServerPort` is zero.

hack/release.toml
```diff
@@ -73,6 +73,13 @@ This was a no-op for a long time. The bootloader is always installed.
         description="""\
 Talos no longer loads by default `rdma_rxe` Linux driver, which is required for RoCE support.
 If the driver is required, it can be enabled by specifying `rdma_rxe` in the `.machine.kernel.modules` field in the machine config.
+"""
+
+    [notes.api-server-loadbalancer]
```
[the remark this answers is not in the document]
DmitriyMV: Yeah, fixed!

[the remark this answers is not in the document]
frezbo: maybe also a note that this requires discovery service to be enabled?

hack/release.toml
```diff
@@ -73,6 +73,20 @@ This was a no-op for a long time. The bootloader is always installed.
...
+        title = "API server load balancer support"
+        description="""\
+Talos now supports configuring the API Server load balancer using `--api-server-balancer-port` flag or machine config
+`features.apiServerBalancerSupport.port` and `features.apiServerBalancerSupport.enabled` fields. If API server load
+balancer is enabled, but port is undefined or zero, the port will be 7445.
+
+Load balancer will distribute traffic to all API servers in the cluster. For the control plane nodes, the load balancer
```
smira, line 85:
```
The loadbalancer will run on every machine in the cluster on the same port, binding to `localhost`. The in-cluster load-balancer endpoint will be used by the `kubelet`, `kube-scheduler`, `kube-controller-manager` and `kube-proxy` by default. The in-cluster loadbalancer endpoint (`https://localhost:7445` by default)  can be passed to the CNIs like Cilium and Calico.

The in-cluster loadbalancer will provide access to the Kubernetes API endpoint even if the external loadbalancer is not healthy provided that the worker nodes can reach to the controlplane machine addresses directly.
```

hack/test/e2e-qemu.sh
```diff
@@ -136,6 +136,15 @@ case "${WITH_SKIP_BOOT_PHASE_FINISHED_CHECK:-no}" in
       ;;
 esac
 
+case "${CUSTOM_CNI_NAME:-none}" in
```
smira, line 139: I would rather not bind together custom CNI and loadbalancer, let them be separate variables. 

```suggestion
case "${WITH_INCLUSTER_LOADBALANCER:-none}" in
```

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
```diff
@@ -0,0 +1,233 @@
...
+				return nil //nolint:nilerr
+			}
+		}
+	}
+
+	localEndpoint := fmt.Sprintf("localhost:%d", port)
+
+	logger.Info("api server load balancer is enabled", zap.String("endpoint", localEndpoint))
+
+	lb := loadbalancer.TCP{Logger: logger}
```
smira, line 90: I'm not sure why are we using here a raw loadbalancer, while we could use instead a high-level wrapper https://github.com/siderolabs/go-loadbalancer/blob/main/controlplane/controlplane.go

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
```diff
@@ -0,0 +1,233 @@
...
+	logger = logger.Named("api-endpoints-balancer")
+
+	port, err := ctrl.loadBalancerData(ctx, r, logger)
+	if err != nil {
+		return err
+	} else if ctx.Err() != nil {
+		return nil //nolint:nilerr
+	}
+
+	if port == 0 {
```
smira, line 76: this is not a way to really have reactive controller - the controller should actively listen for config changes, and bring the lb up/down if the config changes.

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
smira: and if we do, we should also override the default aggressive timeouts, so that the LB won't kick out the nodes too quickly.

sergelogvinov:
It awesome ❤️

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
DmitriyMV: Fixed, done!

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
```diff
@@ -0,0 +1,233 @@
...
+				return fmt.Errorf("error cleaning up specs: %w", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (ctrl *APILoadBalancerController) loadBalancerData(ctx context.Context, r controller.Runtime, _ *zap.Logger) (port int, _ error) {
+	// No way to get config during Controller variable creation, so we have to wait for it.
```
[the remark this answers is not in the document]
DmitriyMV: Done!

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer.go
DmitriyMV: Yeah - reworked. Done!

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/apiserver_loadbalancer_config.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/siderolabs/gen/channel"
+	"github.com/siderolabs/go-pointer"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/talos/pkg/machinery/resources/config"
+	"github.com/siderolabs/talos/pkg/machinery/resources/k8s"
+)
+
+// APILoadBalancerConfigController creates config for load balancer.
+type APILoadBalancerConfigController struct{}
```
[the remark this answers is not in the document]
DmitriyMV: I don't think so, because `transform.Controller` relies on `touched` map, and we explicitly do not want to preserve the resulting resource if we didn't have both parts.

Also I'm pretty sure `transform.Controller` is tailored towards collection of resources and disallowing deletion until child resource is removed.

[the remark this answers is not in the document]
Unix4ever: gotcha :heavy_check_mark:

Unix4ever (approved):

Issue #7233, opened 2023-05-16
Refactor `uncordonNode` & `labelControlPlane` tasks


smira:
### `labelControlPlane`

Only on controlplane machines.

1. Labels the `Node` as controlplane.
2. Taint the `Node` as unschedulable (if not disabled in the config).

### `uncordonNode`

1. Wait for the `Node` to be `Ready`.
2. Remove the Talos taint set on upgrade/etc. (when cordoning the node).

### Controlplane Label

Shouldn't ever block the boot sequence.

Can be implemented directly by injecting a proper label to the `NodeLabelsApplyController`.

### Tainting Controlplane Machines

Shouldn't ever block the boot sequence.

```mermaid
flowchart LR

  NodeName([NodeName ready]) --> WatchNode[/watch Node resource/]
  WatchNode --> Node_is_Ready([Node is Ready])

```

PR #7352, opened 2023-06-09
refactor: task `labelControlPlane` into controllers

smira:
See #7233

The controlplane label is simply injected into existing controller-based node label flow.

For controlplane taint default NoScheduleTaint, additional controller & resource was implemented to handle node taints.

This also fixes a problem with `allowSchedulingOnControlPlanes` not being reactive to config changes - now it is.


frezbo (approved):

smira:
I want to make a change to handle better `SkipNodeRegistration` config

smira:
> I want to make a change to handle better `SkipNodeRegistration` config

done

Issue #7313, opened 2023-06-02
allowSchedulingOnControlPlanes missing from gen config


si458:
### Description
when you generate a config with comments, `allowSchedulingOnControlPlanes` is missing completely from the comments
`.\talosctl-windows-amd64.exe gen config mycluster https://192.168.168.233:6443 --output-dir mycluster --kubernetes-version 1.26.4`

### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`] ANY
- Kubernetes version: [`kubectl version --short`] ANY
- Platform: ANY

smira:
Yes, in fact many options are missing, as we can't put each and every option, but this one probably deserves to be there.

si458:
@smira yes please add it! i was getting so confused as the online said its there but i couldnt find it anywhere in the config that was generated, i throught it might of been a hidden option maybe for advanced users

PR #7360, opened 2023-06-12
docs: include `allowSchedulingOnControlPlanes` on `talosctl gen config` output

utkuozdemir:
Include a description and a commented-out example for the `cluster.allowSchedulingOnControlPlanes` field on `talosctl gen config ...` output.

Closes siderolabs/talos#7313.


smira:
I guess this should also update docs...

smira (approved):

utkuozdemir:
> I guess this should also update docs...

right, just added

frezbo (approved):

PR #7362, opened 2023-06-12
chore: fix cilium e2e tests

DmitriyMV:
`WITH_CONFIG_PATCH_WORKER` check result was overriding any value set in `CONFIG_PATCH_FLAG` variable. Move it to the different variable.


frezbo (approved):

smira:
but tests are still broken though?

smira:
I see the issue, I will fix it

PR #7375, opened 2023-06-14
feat: implement secure boot from disk

smira:
This includes sd-boot handling, EFI variables, etc.

There are some TODOs which need to be addressed to make things smooth.

Install to disk, upgrades work.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -109,7 +109,7 @@ func NewInstaller(cmdline *procfs.Cmdline, seq runtime.Sequence, opts *Options) 
 		}
 	}
 
-	i.manifest, err = NewManifest(seq, bootLoaderPresent, i.options)
+	i.manifest, err = NewManifest(seq, i.bootloader.UEFIBoot(), bootLoaderPresent, i.options)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create installation manifest: %w", err)
 	}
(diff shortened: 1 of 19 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -109,7 +109,7 @@ func NewInstaller(cmdline *procfs.Cmdline, seq runtime.Sequence, opts *Options)
 		}
 	}
 
-	i.manifest, err = NewManifest(seq, bootLoaderPresent, i.options)
+	i.manifest, err = NewManifest(seq, i.bootloader.UEFIBoot(), bootLoaderPresent, i.options)
```
smira, line 112: TODO: above, on line 106, needs a way to force a specific bootloader based on `Options` (flags), so that one could create `sd-boot` metal image for example

hack/ukify/main.go
```diff
@@ -233,7 +233,7 @@ func run() error {
 	}{
 		Name:    version.Name,
 		ID:      strings.ToLower(version.Name),
-		Version: version.Tag,
+		Version: version.Tag, // TODO: this is abbrev tag, not the full tag, it should be fixed by building from sources + generate data
```
smira, line 236: ukify should be built after `generate` step, so that `version.Tag` is not abbreviated ?

internal/pkg/mount/pseudo.go
```diff
@@ -29,5 +33,10 @@ func PseudoSubMountPoints() (mountpoints *Points, err error) {
 	pseudo.Set("hugetlb", NewMountPoint("hugetlbfs", "/dev/hugepages", "hugetlbfs", 0, ""))
 	pseudo.Set("securityfs", NewMountPoint("securityfs", "/sys/kernel/security", "securityfs", unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV|unix.MS_RELATIME, ""))
 
+	if _, err := os.Stat(constants.EFIVarsMountPoint); err == nil {
```
smira, line 36: mount as `readonly`, remount in the code which writes the vars? better for security

smira:
@frezbo please help me to fix this when you're back:

```
the required flash image was not found in any of the expected paths for (["/usr/share/OVMF/OVMF_CODE_4M.secboot.fd"]), please install it with the package manager or specify --extra-uefi-search-paths
```

frezbo:
> /usr/share/OVMF/OVMF_CODE_4M.secboot.fd

Seems alpine has a different file: `/usr/share/qemu/edk2-x86_64-secure-code.fd`

hack/ukify/main.go
frezbo: fixed

internal/pkg/mount/pseudo.go
frezbo: TODO: for later

frezbo (approved):

PR #7384, opened 2023-06-16
feat: implement revert for sd-boot

frezbo:
Implement revert for sd-boot.


frezbo:
Also move the bootlabels back to grub, since it's only used there

internal/app/machined/pkg/runtime/v1alpha1/bootloader/mount/mount.go
```diff
@@ -0,0 +1,105 @@
...
+			if errors.Is(err, blockdevice.ErrMissingPartitionTable) || os.IsNotExist(err) {
+				return nil
+			}
+
+			return err
+		}
+
+		probedBlockDevice = dev
+	case disk == "":
+		// attempt to probe BOOT partition on any disk
```
Unix4ever, line 44:
```suggestion
		// attempt to probe partition with partitionLabel on any disk
```

PR #7392, opened 2023-06-21
feat: custom frontend for installer customization

frezbo:
Use a custom frontend for installer customization. This gets rid of all `ONBUILD` commands and can use buildx natively.

Example usage:

```Dockerfile
FROM ghcr.io/<custom user>/kernel:<custom tag> as kernel
FROM ghcr.io/<custom-user>/nonfree-kmod-nvidia:<custom tag> as customization

FROM ghcr.io/siderolabs/installer:<tag> as installer
```

This can be natively build with buildx.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -697,6 +697,34 @@ COPY --from=rootfs / /
 LABEL org.opencontainers.image.source https://github.com/siderolabs/talos
 ENTRYPOINT ["/sbin/init"]
 
+FROM build-go AS module-sig-verify-linux-amd64-build
+ARG GO_BUILDFLAGS
+ARG GO_LDFLAGS
+ARG GOAMD64
+WORKDIR /src/module-sig-verify
+COPY ./hack/module-sig-verify/go.mod ./hack/module-sig-verify/go.sum .
+RUN --mount=type=cache,target=/.cache go mod download
+COPY ./hack/module-sig-verify/main.go .
+RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=amd64 GOAMD64=${GOAMD64} go build -o module-sig-verify-linux-amd64 .
+
+FROM build-go AS module-sig-verify-linux-arm64-build
+ARG GO_BUILDFLAGS
+ARG GO_LDFLAGS
+ARG GOAMD64
+WORKDIR /src/module-sig-verify
+COPY ./hack/module-sig-verify/go.mod ./hack/module-sig-verify/go.sum .
+RUN --mount=type=cache,target=/.cache go mod download
+COPY ./hack/module-sig-verify/main.go .
+RUN --mount=type=cache,target=/.cache GOOS=linux GOARCH=arm64 GOAMD64=${GOAMD64} go build -o module-sig-verify-linux-arm64 .
+
+FROM scratch AS module-sig-verify-linux-amd64
+COPY --from=module-sig-verify-linux-amd64-build /src/module-sig-verify/module-sig-verify-linux-amd64 /module-sig-verify-linux-amd64
+
+FROM scratch AS module-sig-verify-linux-arm64
+COPY --from=module-sig-verify-linux-arm64-build /src/module-sig-verify/module-sig-verify-linux-arm64 /module-sig-verify-linux-arm64
+
+FROM module-sig-verify-linux-${TARGETARCH} AS module-sig-verify-targetarch
+
 FROM --platform=${BUILDPLATFORM} tools AS ukify-tools
 # base has the talos source with the non-abrev version of TAG
 COPY --from=base /src/pkg /go/src/github.com/pkg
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -825,30 +854,17 @@ LABEL org.opencontainers.image.source https://github.com/siderolabs/talos
...
-    && depmod -b /rootfs $(ls /rootfs/lib/modules) \
-    && set -o pipefail && mksquashfs /rootfs rootfs.sqsh -all-root -noappend -comp xz -Xdict-size 100% -no-progress \
-    && find . 2>/dev/null | cpio -H newc -o | xz -v -C crc32 -0 -e -T 0 -z > /usr/install/${TARGETARCH}/initramfs.xz \
-    && rm -rf /rootfs /initramfs /rootfs/kernel
-ONBUILD WORKDIR /
 
 FROM installer-image-squashed AS imager
 
+# installer frontend
+FROM erichripko/simple-frontend AS installer-frontend
```
frezbo, line 861: maybe we add code from this to `hack`? It's not too crazy, and we can have full control over it (the upstream uses an old go toolchain)

hack/Dockerfile.installer-frontend
```diff
@@ -0,0 +1,47 @@
...
+WORKDIR /initramfs
+ARG RM
+RUN <<EOF
+set -euo pipefail
+
+xz -d /usr/install/${TARGETARCH}/initramfs.xz
+cpio -idvm < /usr/install/${TARGETARCH}/initramfs
+unsquashfs -f -d /rootfs rootfs.sqsh
+
+for f in ${RM}; do
```
frezbo, line 17: do we need this, we can always assume to remove `/lib/modules` when using a custom installer?

hack/Dockerfile.installer-frontend
```diff
@@ -0,0 +1,47 @@
...
+find /rootfs/lib/modules/${KERNEL_VERSION} -type f -name '*.ko' | xargs -I {} /bin/module-sig-verify -cert /rootfs/kernel/signing_key.x509 -module {}
+
+mksquashfs /rootfs rootfs.sqsh -all-root -noappend -comp xz -Xdict-size 100% -no-progress
+find . 2>/dev/null | cpio -H newc -o | xz -v -C crc32 -0 -e -T 0 -z > /usr/install/${TARGETARCH}/initramfs.xz
+
+rm -rf /rootfs /initramfs /rootfs/kernel
+EOF
+
+WORKDIR /
+COPY --from=kernel /boot/vmlinuz /usr/install/${TARGETARCH}/vmlinuz
```
frezbo, line 47: should we implicitly assume this or ask the user to copy vmlinuz?

smira:
Can we use #7536 if we ask people to use Talos repo to build `initramfs.xz` out of new kernel?

frezbo:
> Can we use #7536 if we ask people to use Talos repo to build `initramfs.xz` out of new kernel?

another option is we move the copying of module list in go code, so imager can handle it

smira:
> > Can we use #7536 if we ask people to use Talos repo to build `initramfs.xz` out of new kernel?
> 
> another option is we move the copying of module list in go code, so imager can handle it

This means imager should handle pulling kernel pkgs... I'm not against it, it's still doable

frezbo:
> This means imager should handle pulling kernel pkgs... I'm not against it, it's still doable

does it really need? with something like this:

```Dockerfile
FROM ghcr.io/siderolabs/imager:<tag> as imager
COPY --from=ghcr.io/<custom user>/kernel:<custom tag> /vmlinuz /imager/vmlinuz/path
```

and the resulting imager can be used, since it has the custom kernel and imager has logic to copy required modules

smira:
> > This means imager should handle pulling kernel pkgs... I'm not against it, it's still doable
> 
> does it really need? with something like this:
> 
> ```dockerfile
> FROM ghcr.io/siderolabs/imager:<tag> as imager
> COPY --from=ghcr.io/<custom user>/kernel:<custom tag> /vmlinuz /imager/vmlinuz/path
> ```
> 
> and the resulting imager can be used, since it has the custom kernel and imager has logic to copy required modules

I mean `initramfs.xz` should be rebuilt with new modules...

frezbo:
> > > This means imager should handle pulling kernel pkgs... I'm not against it, it's still doable
> > 
> > 
> > does it really need? with something like this:
> > ```dockerfile
> > FROM ghcr.io/siderolabs/imager:<tag> as imager
> > COPY --from=ghcr.io/<custom user>/kernel:<custom tag> /vmlinuz /imager/vmlinuz/path
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and the resulting imager can be used, since it has the custom kernel and imager has logic to copy required modules
> 
> I mean `initramfs.xz` should be rebuilt with new modules...

ahh i see, maybe we need to handle it then, let's make it easier for users

smira:
Should be addressed in #7612

PR #7401, opened 2023-06-23
Compare only basename of `os.Args[0]`.

reitermarkus:
# Pull Request

## What? (description)

Only compare the basename of the 0th argument.

## Why? (reasoning)

Took me more than a day to debug why

```c
execl("/sbin/poweroff", "poweroff", (char *)NULL);
```

was not working in https://github.com/siderolabs/extensions/pull/173, only to discover that 

```c
execl("/sbin/poweroff", "/sbin/poweroff", (char *)NULL);
```

does work.

## Acceptance

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 	"time"
 
@@ -286,25 +287,25 @@ func run() error {
 }
 
 func main() {
-	switch os.Args[0] {
-	case "/apid":
+	switch filepath.Base(os.Args[0]) {
+	case "apid":
 		apid.Main()
 
 		return
-	case "/trustd":
+	case "trustd":
 		trustd.Main()
 
 		return
 	// Azure uses the hv_utils kernel module to shutdown the node in hyper-v by calling perform_shutdown which will call orderly_poweroff which will call /sbin/poweroff.
-	case "/sbin/poweroff":
+	case "poweroff":
 		poweroff.Main()
 
 		return
-	case "/sbin/wrapperd":
+	case "wrapperd":
 		wrapperd.Main()
 
 		return
-	case "/sbin/dashboard":
+	case "dashboard":
 		dashboard.Main()
 
 		return


frezbo:
`execl("/sbin/poweroff", "/sbin/poweroff", (char *)NULL);` this is the correct usage of `execve`, so seems the library has a bug?

reitermarkus:
> this is the correct usage

I'm pretty sure both variants are correct, so both should be supported.

In fact, `execl("/sbin/init", "poweroff", (char *)NULL);` would also be correct.

frezbo:
> > this is the correct usage
> 
> I'm pretty sure both variants are correct, so both should be supported.
> 
> In fact, `execl("/sbin/init", "poweroff", (char *)NULL);` would also be correct.

it works but the man page says it should be absolute:

```bash
  where  pathname  is  the absolute pathname of the file specified as the first argument of execve(), and arg...  is the series of words pointed to by the argv argument of execve(),
       starting at argv[1].  Note that there is no way to get the argv[0] that was passed to the execve() call.
```

frezbo:
the thing here is using `filepath.Base("/tmp/shutdown")` would also work

reitermarkus:
> it works but the man page says it should be absolute:

That part seems to be from the “Interpreter scripts” section of `execve`, which is specific to how files containing a shebang line are executed.

The man page of `execl` doesn't mention anything about `arg0` needing to be an absolute path or the same as the `path` argument.

frezbo:
true, it's just i've only most places using absolute paths only, I'm a little skeptical of dropping the full path, might need another set of eyes

reitermarkus:
> the thing here is using `filepath.Base("/tmp/shutdown")` would also work

I suspect you mean `/tmp/poweroff` here.

Yes, it would also work, as it should. I would very much expect it to behave like `poweroff` and not like `init` given the name.

frezbo:
> I suspect you mean /tmp/poweroff here.

an malicious actor could have a symlink to `/tmp/powerof` and execute it here (talos immutability should prevent that, but still it's an attack vector)

frezbo:
anyways I'll wait on others to also take a look

smira:
`argv[0]` is controlled by whoever is calling `exec*()`, so this field can contain any value (even without symlinks). the shell passes there the name as it was put on the command line, so I think it's fine to check only `basename` here as it doesn't change much.

smira (approved):

Issue #7373, opened 2023-06-14
research replacing sbsign & ovmfctl with native go library


smira:
See https://github.com/Foxboron/go-uefi

PR #7403, opened 2023-06-23
feat: auto-enroll secure boot keys

TimJones:
Uses the auto-enrollment feature of sd-boot to enroll required UEFI Secure Boot keys.

Fixes: #7373


cmd/installer/pkg/iso.go
```diff
@@ -98,6 +98,18 @@ func CreateUKIISO(iso, dir, arch string) error {
...
 
+	if _, err := cmd.Run("mmd", "-i", efiBootImg, "::loader"); err != nil {
+		return err
+	}
+
+	if _, err := cmd.Run("mmd", "-i", efiBootImg, "::loader/keys"); err != nil {
+		return err
+	}
+
+	if _, err := cmd.Run("mmd", "-i", efiBootImg, "::loader/keys/SECBOOT"); err != nil {
```
TimJones, line 109: `SECBOOT` here is arbitrary, just a name for the key group. Could be made user-configurable maybe?

frezbo: maybe just `auto`?

frezbo: ```NAME being the display name for the set of variables in the menu. If one of the sets is named **auto** then it might be enrolled automatically depending on whether "secure-boot-enroll" is set to force or not.```

i gues it then auto-enrolls

TimJones: ooooo nice

hack/ukify/main.go
```diff
@@ -47,18 +49,37 @@ var (
...
+		return "", err
+	}
+	cert, err := pem.GetCert()
+	if err != nil {
+		return "", err
+	}
+	key, err := pem.GetRSAKey()
+	if err != nil {
+		return "", err
+	}
```
frezbo, line 70: extract into function?

TimJones: just within the ukify `main.go` or into a module elsewhere?

frezbo: i think under certs is fine

hack/ukify/gen-certs/main.go
```diff
@@ -54,6 +59,65 @@ func generateSigningCerts(path, prefix, commonName string) error {
 	return nil
 }
 
+// ref: https://blog.hansenpartnership.com/the-meaning-of-all-the-uefi-keys/
+func generateSecureBootFiles(path, prefix string) error {
+	uuid, err := uuid.NewRandom()
```
frezbo, line 64: this I'm not sure, is it okay for the GUID to be random or should SideroLabs/customer has a unique one? :thinking:

hack/ukify/main.go
frezbo: I didn't extract since we don't import `gen-certs` into `ukify` for now.

hack/ukify/gen-certs/main.go
TimJones: This is only for generating the dev keys anyways, random is fine. Real keys would have a GUID fixed by the provider, but then they would also be providing the `PK.auth`, `KEK.auth` and `db.auth` files external to this process.

frezbo: makes sense, we need to now figure out some way for users to consume these (an easy way to generate these files if they only have certs and maybe have this be dynamic as part of installer)

TimJones: as of now, they have to manually put the files in `_out/uki-certs` before doing `make imager installer iso-uki` and it'll pull them in properly, but yeah, probably need a slightly nicer way.

frezbo (approved):

PR #7404, opened 2023-06-24
fix: refine OVMF search paths

twelho:
The static paths for the OVMF firmware are limited, and won't match, for example, any of the files installed by `edk2-ovmf` on a Fedora 38 system. This change separates the search paths and filenames, making sure all combinations are covered when looking for a suitable firmware. Similarly also cleans up the OVMF vars lookup.

## Acceptance

### DIFF
--- a/pkg/provision/providers/qemu/arch.go
+++ b/pkg/provision/providers/qemu/arch.go
@@ -5,7 +5,6 @@
 package qemu
 
 import (
-	"os"
 	"os/exec"
 	"path/filepath"
 )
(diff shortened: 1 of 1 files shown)


pkg/provision/providers/qemu/arch.go
```diff
@@ -94,25 +93,43 @@ func (arch Arch) PFlash(uefiEnabled, secureBootEnabled bool, extraUEFISearchPath
...
+		}
+
+		if !secureBootEnabled {
+			uefiSourceFiles = append(uefiSourceFiles, uefiSourceFilesInsecure...)
+		}
+
+		// Append extra search paths
 		for _, p := range extraUEFISearchPaths {
-			uefiSourcePaths = append(uefiSourcePaths, filepath.Join(p, "OVMF.fd"))
+			uefiSourcePathPrefixes = append(uefiSourcePathPrefixes, p)
```
frezbo, line 124:
```suggestion
			uefiSourcePathPrefixes = append(uefiSourcePathPrefixes, extraUEFISearchPaths...)
```

this would avoid the looping

pkg/provision/providers/qemu/arch.go
```diff
@@ -94,25 +93,43 @@ func (arch Arch) PFlash(uefiEnabled, secureBootEnabled bool, extraUEFISearchPath
...
 		}
 
-		if secureBootEnabled {
-			// picking exactly the last one, as it's the one having secure boot enabled
-			uefiSourcePaths = uefiSourcePaths[len(uefiSourcePaths)-1:]
+		var uefiSourcePaths []string
+
+		for _, p := range uefiSourcePathPrefixes {
+			for _, f := range uefiSourceFiles {
+				uefiSourcePaths = append(uefiSourcePaths, filepath.Join(p, f))
```
frezbo, line 131: this would mean `OVMF_CODE.secboot.fd` would get picked instead of `OVMF_CODE_4M.secboot.fd`, would prefer the 4M first, easy thing would be th re-arrange here https://github.com/siderolabs/talos/pull/7404/files#diff-0d7b72e333912658fa67ecfdfa944c0076a1f0643d6e1970d05c34546958738eR106

pkg/provision/providers/qemu/pflash.go
```diff
@@ -70,19 +70,41 @@ func (p *provisioner) createPFlashImages(state *vm.State, nodeName string, pflas
...
+
+		for _, p := range ovmfVarsPathPrefixes {
+			for _, f := range ovmfVarsFiles {
+				ovmfVarsPaths = append(ovmfVarsPaths, filepath.Join(p, f))
 			}
 		}
 
+		var ovmfVars string
+
+		for _, p := range ovmfVarsPaths {
```
frezbo, line 96: we can combine the loop above into this and breaking on the first existing file

frezbo (commented):
Thank you for this, this is so much cleaner, just small nits. Also might take a look at #7403 which removes some code kind of related to this

frezbo:
> 

if it's easier I can merge #7403 first and you can rebase on top of it

pkg/provision/providers/qemu/pflash.go
twelho: I've intentionally separated the two loops to allow for returning the considered paths in the error below, this is invaluable information when debugging a missing package, search path etc.

twelho:
Nits fixed, thanks for the quick review.

> if it's easier I can merge #7403 first and you can rebase on top of it

Based on a quick look https://github.com/siderolabs/talos/pull/7403 removes some of the duplication between `arch` and `pflash`, which would also be beneficial here. I can rebase on top of it if it gets merged first.

pkg/provision/providers/qemu/pflash.go
frezbo: i believe we can still do as the else condition for the `os.Stat` since we break on first file found, so end of loop means we can't find the file and directly return the error?

```go
var ovmfVarsPaths []string
var ovmfVars string

		for _, p := range ovmfVarsPathPrefixes {
                        found := false
			for _, f := range ovmfVarsFiles {
				ovmfVarsPaths = append(ovmfVarsPaths, filepath.Join(p, f))

                                if _, err := os.Stat(filepath.Join(p, f)); err == nil {
				ovmfVars = p
                                found = true
                      
				break
			}
                        if found {
                              break
                       }
			}
		}

if ovmfVars == "" {
			return nil, fmt.Errorf("failed to find OVMF vars in %q", ovmfVarsPaths)
		}
```

something around this (apologies for the bad formatting)

pkg/provision/providers/qemu/arch.go
```diff
@@ -94,31 +93,52 @@ func (arch Arch) PFlash(uefiEnabled, secureBootEnabled bool, extraUEFISearchPath
...
 		}
 
-		if secureBootEnabled {
-			// picking exactly the last one, as it's the one having secure boot enabled
-			uefiSourcePaths = uefiSourcePaths[len(uefiSourcePaths)-1:]
-			uefiVarsSourcePaths = uefiVarsSourcePaths[len(uefiVarsSourcePaths)-1:]
+		// Append extra search paths
+		uefiSourcePathPrefixes = append(uefiSourcePathPrefixes, extraUEFISearchPaths...)
+
+		var uefiSourcePaths []string
```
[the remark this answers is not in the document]
twelho: This is what gets returned as `SourcePaths` for the `PFlash` object below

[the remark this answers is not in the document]
twelho: (The diff doesn't show it since the name hasn't changed)

pkg/provision/providers/qemu/pflash.go
twelho: (This change is not needed now that https://github.com/siderolabs/talos/pull/7403 is merged)

PR #7407, opened 2023-06-26
chore: format and cleanup test scripts

frezbo:
This formats and cleanups the test scripts.

### DIFF
--- a/hack/test/e2e-qemu.sh
+++ b/hack/test/e2e-qemu.sh
@@ -4,28 +4,27 @@ set -eou pipefail
 
 export USER_DISKS_MOUNTS="/var/lib/extra,/var/lib/p1,/var/lib/p2"
 
+# shellcheck source=/dev/null
 source ./hack/test/e2e.sh
 
 PROVISIONER=qemu
-CLUSTER_NAME=e2e-${PROVISIONER}
+CLUSTER_NAME="e2e-${PROVISIONER}"
+
+QEMU_FLAGS=()
 
 case "${CI:-false}" in
-  true)
-    QEMU_FLAGS=""
-    INSTALLER_TAG="${TAG}"
+  false)
+    QEMU_FLAGS+=("--with-bootloader=false")
     ;;
   *)
-    QEMU_FLAGS="--with-bootloader=false"
-    INSTALLER_TAG="latest"
     ;;
 esac
 
 case "${CUSTOM_CNI_URL:-false}" in
   false)
-    CUSTOM_CNI_FLAG=
     ;;
   *)
-    CUSTOM_CNI_FLAG="--custom-cni-url=${CUSTOM_CNI_URL}"
+    QEMU_FLAGS+=("--custom-cni-url=${CUSTOM_CNI_URL}")
     ;;
 esac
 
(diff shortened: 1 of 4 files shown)


hack/test/e2e.sh
```diff
@@ -49,127 +51,126 @@ PROVISIONER=
...
 
 # Create a cluster via CAPI.
 function create_cluster_capi {
   trap cleanup_capi EXIT
 
-  ${KUBECTL} --kubeconfig /tmp/e2e/docker/kubeconfig apply -f ${TMP}/cluster.yaml
+  ${KUBECTL} --kubeconfig /tmp/e2e/docker/kubeconfig apply -f "${TMP}/cluster.yaml"
 
   # Wait for first controlplane machine to have a name
-  timeout=$(($(date +%s) + ${TIMEOUT}))
```
smira, line 62: this dropped `$` looks weird

frezbo: that's intentional, no need of interpolation when using the `$((` syntax

frezbo: https://www.shellcheck.net/wiki/SC2004

smira (approved):

hack/test/e2e-qemu.sh
```diff
@@ -173,19 +164,12 @@ function create_cluster {
     --cidr=172.20.1.0/24 \
     --user-disk=/var/lib/extra:100MB \
     --user-disk=/var/lib/p1:100MB:/var/lib/p2:100MB \
-    --install-image=${INSTALLER_IMAGE} \
+    --install-image="${INSTALLER_IMAGE}" \
     --with-init-node=false \
-    --cni-bundle-url=${ARTIFACTS}/talosctl-cni-bundle-'${ARCH}'.tar.gz \
+    --cni-bundle-url="${ARTIFACTS}/talosctl-cni-bundle-\${ARCH}.tar.gz" \
```
DmitriyMV, line 169: I'm not bash expert so is there any reason for `\` before `${ARCH}` here?

smira:
looks like `e2e` is broken

hack/test/e2e.sh
```diff
@@ -198,25 +199,25 @@ function get_kubeconfig {
...
 
 function build_registry_mirrors {
   if [[ "${CI:-false}" == "true" ]]; then
-    REGISTRY_MIRROR_FLAGS=
+    REGISTRY_MIRROR_FLAGS=()
 
     for registry in docker.io registry.k8s.io quay.io gcr.io ghcr.io registry.dev.talos-systems.io; do
       local service="registry-${registry//./-}.ci.svc"
-      local addr=`python3 -c "import socket; print(socket.gethostbyname('${service}'))"`
+      addr=$(python3 -c "import socket; print(socket.gethostbyname('${service}'))")
```
DmitriyMV, line 214: Why drop `local`?

frezbo:
> looks like `e2e` is broken

i think i fixed that now, let's see

DmitriyMV (approved):
🆒 usage of bash arrays.

hack/test/e2e-qemu.sh
frezbo: so that it does not expand, the cli expect `${ARCH}` as is and not expanded

hack/test/e2e.sh
frezbo: https://www.shellcheck.net/wiki/SC2155

hack/test/e2e-qemu.sh
DmitriyMV: Ah! Can't we use single quotes here tho?

hack/test/e2e.sh
DmitriyMV: Maybe declare it separately then?

frezbo: that didn't matter much here :sweat_smile:

hack/test/e2e-qemu.sh
frezbo: since the whole expression is inside a double quote, it would still expand

PR #7409, opened 2023-06-27
chore: add `/sbin/shutdown`

frezbo:
Some tools like qemu-guest-agent when ran as a extension service calls `/sbin/shutdown` instead of `/sbin/poweroff`. This adds handling for the same.

Ref: https://github.com/siderolabs/extensions/pull/173


smira (approved):

reitermarkus:
This should also support the different shutdown flags:

https://github.com/qemu/qemu/blob/master/qga/commands-posix.c#L92-L94

frezbo:
> This should also support the different shutdown flags:
> 
> https://github.com/qemu/qemu/blob/master/qga/commands-posix.c#L92-L94

Passing args in talos should have no effect. It'll be a no-op and talos would just shutdown.

reitermarkus:
Sure, but `shutdown -r` should reboot, not shut down.

PR #7411, opened 2023-06-27
fix: disable LVM backups/archive

### LINKED ISSUE #3129: Add dependencies for encrypted Ceph OSDs
### Description

Right now, Ceph/Rook run correctly on top of Talos clusters, but only if the OSDs aren't encrypted. Talos should provide support for encrypted OSDs.

https://docs.ceph.com/en/latest/ceph-volume/lvm/encryption/

smira:
Fixes #3129

Talos does not have a good location to keep LVM metadata backups.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -562,6 +562,7 @@ COPY --chmod=0644 hack/containerd.toml /rootfs/etc/containerd/config.toml
 COPY --chmod=0644 hack/cri-containerd.toml /rootfs/etc/cri/containerd.toml
 COPY --chmod=0644 hack/cri-plugin.part /rootfs/etc/cri/conf.d/00-base.part
 COPY --chmod=0644 hack/udevd/80-net-name-slot.rules /rootfs/usr/lib/udev/rules.d/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -562,6 +562,7 @@ COPY --chmod=0644 hack/containerd.toml /rootfs/etc/containerd/config.toml
 COPY --chmod=0644 hack/cri-containerd.toml /rootfs/etc/cri/containerd.toml
 COPY --chmod=0644 hack/cri-plugin.part /rootfs/etc/cri/conf.d/00-base.part
 COPY --chmod=0644 hack/udevd/80-net-name-slot.rules /rootfs/usr/lib/udev/rules.d/
+COPY --chmod=0644 hack/lvm.conf /rootfs/etc/lvm/lvm.conf
```
frezbo, line 565: `hack/lvm.conf` file missing :sweat_smile:

smira: yep, perfectly works fine locally :-D

frezbo (approved):

PR #7426, opened 2023-06-29
feat: implement machine config documents for event and log streaming

### LINKED ISSUE #7228: Machine configuration documents for SideroLink, kernel log and event streaming
Depends on #7230.

We need to find equivalent configuration documents for the following kernel args:

* `siderolink.api=`
* `talos.events.sink=` 
* `talos.logging.kernel=`

smira:
Fixes #7228

Add some changes to make Talos accept partial machine configuration without main v1alpha1 config.

With this change, it's possible to connect a machine already running with machine configuration (v1alpha1), the following patch will connect to a local SideroLink endpoint:

```yaml
apiVersion: v1alpha1
kind: SideroLinkConfig
apiUrl: grpc://172.20.0.1:4000/?jointoken=foo
---
apiVersion: v1alpha1
kind: KmsgLogConfig
name: apiSink
url: tcp://[fdae:41e4:649b:9303::1]:4001/
---
apiVersion: v1alpha1
kind: EventSinkConfig
endpoint: "[fdae:41e4:649b:9303::1]:8080"
```


go.sum
frezbo, line 1: not tidied?

pkg/machinery/config/types/runtime/testdata/kmsglog.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1alpha1
+kind: KmsgLogConfig
+name: apiSink
```
frezbo, line 3: why do we have a name here?

frezbo (approved):

website/content/v1.5/reference/configuration.md
```diff
@@ -583,14 +591,11 @@ scheduler:
 {{< /highlight >}}</details> | |
 |`discovery` |<a href="#clusterdiscoveryconfig">ClusterDiscoveryConfig</a> |Configures cluster member discovery. <details><summary>Show example(s)</summary>{{< highlight yaml >}}
 discovery:
-    enabled: true # Enable the cluster membership discovery feature.
```
frezbo, line 586: seems we lost all comments?

pkg/machinery/config/container/container.go
```diff
@@ -112,7 +112,7 @@ func (container *Container) Debug() bool {
 // Persist implements config.Config interface.
 func (container *Container) Persist() bool {
 	if container.v1alpha1Config == nil {
-		return false
+		return true
```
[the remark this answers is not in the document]
smira: yep, I think we should default to `Persist() == true` always, and probably deprecate `persist: false`, as this doesn't make much sense. it's a properly of v1alpha1 config, and you might not have it at all

website/content/v1.5/reference/configuration.md
smira: omg, I need to look into that, that was a change to use pointer receivers, but I can figure out some other way to fix that

pkg/machinery/config/types/runtime/testdata/kmsglog.yaml
smira: this is a `NamedConfiguration`, i.e. it has a `name:`, so there might be multiple config documents, so you can have kmsg log streaming to multiple destinations

go.sum
smira: I guess someone doing `go mod tidy` before me was using outdated version

pkg/machinery/config/types/runtime/kmsg_log.go
```diff
@@ -0,0 +1,97 @@
...
+	if s.MetaName == "" {
+		return nil, fmt.Errorf("name is required")
+	}
+
+	if s.KmsgLogURL.URL == nil {
+		return nil, fmt.Errorf("url is required")
+	}
+
+	switch s.KmsgLogURL.URL.Scheme {
+	case "tcp":
```
frezbo, line 91: didn't old one supported udp  too?

frezbo (approved):

pkg/machinery/config/types/runtime/kmsg_log.go
smira: you're right, good catch, I forgot about `udp`

Unix4ever (approved):

pkg/machinery/config/types/runtime/kmsg_log.go
smira: fixed

PR #7427, opened 2023-06-29
feat: add KMS assisted encryption key handler

### LINKED ISSUE #7374: network KMS disk encryption
Introduce new EncryptionKeyProvider which is configured with KMS API URL.

KMS implements the following API:

- `Seal(MachineID, data) -> sealed_data`
- `Unseal(MachineID, sealed_data) -> data`

When the disk encryption is initialized, Talos creates a random disk encryption key, seals it with the KMS and stores sealed value in `META` partition. On the next boot, Talos fetches sealed disk encryption key from `META`, Unseals it and unlocks the disk.

Unix4ever:
Talos now supports new type of encryption keys which rely on Sealing/Unsealing randomly generated bytes with a KMS server:
    
```yaml
systemDiskEncryption:
  ephemeral:
    keys:
      - kms:
          endpoint: https://1.2.3.4:443
        slot: 0
```
gRPC API definitions and a simple reference implementation of the KMS server can be found in this
[repository](https://github.com/siderolabs/kms-client/blob/main/cmd/kms-server/main.go).

FIxes #7374


hack/test/e2e-qemu.sh
```diff
@@ -100,7 +100,7 @@ case "${WITH_DISK_ENCRYPTION:-false}" in
   false)
     ;;
   *)
-    QEMU_FLAGS+=("--encrypt-ephemeral" "--encrypt-state")
+    QEMU_FLAGS+=("--encrypt-ephemeral" "--encrypt-state", "--disk-encryption-key-types", "kms")
```
frezbo, line 103: don;t we need a different test for this, one that tests with `uuid` and another that tests kms?

hack/release.toml
```diff
@@ -102,6 +102,23 @@ Talos now supports generating a custom iso that can be used with SecureBoot. Key
         description="""\
 Talos now supports setting `environmentFile` for an extension service container spec. Refer: https://www.talos.dev/v1.5/advanced/extension-services/#container
 The extension waits for the file to be present before starting the service.
+"""
+
+    [notes.kms-encryption-keys]
+        title = "KMS Encryption Keys"
```
smira, line 108:
```suggestion
        title = "Network KMS Disk Encryption"
```

hack/release.toml
```diff
@@ -102,6 +102,23 @@ Talos now supports generating a custom iso that can be used with SecureBoot. Key
...
+        description="""\
+Talos now supports new type of encryption keys which rely on Sealing/Unsealing randomly generated bytes with a KMS server:
```
smira, line 110:
```suggestion
Talos now supports new type of encryption keys which are sealed/unsealed with an external KMS server:
```

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	var resp *kms.Response
+
+	key := make([]byte, 32)
+	if _, err = io.ReadFull(rand.Reader, key); err != nil {
+		return nil, nil, err
+	}
+
+	err = retry.Exponential(time.Minute*5, retry.WithUnits(time.Second), retry.WithJitter(time.Second)).RetryWithContext(ctx, func(ctx context.Context) error {
```
frezbo, line 74: maybe a minute is enough? :thinking:

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// KMSKeyHandler seals token using KMS service.
+type KMSKeyHandler struct {
+	key
+	kmsEndpoint string
+}
+
+// NewKMSKeyHandler creates new KMSKeyHandler.
+func NewKMSKeyHandler(key key, kmsEndpoint string) (*KMSKeyHandler, error) {
```
smira, line 43: the public API with private type looks a bit weird, not sure if this should be moved into yet another subpackage or `Key` should be public instead?

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+	return &KMSKeyHandler{
+		key:         key,
+		kmsEndpoint: kmsEndpoint,
+	}, nil
+}
+
+// NewKey implements Handler interface.
+func (h *KMSKeyHandler) NewKey() (*encryption.Key, token.Token, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
smira, line 52: I think `ctx` should be passed in explicitly here, we should try to build a full context chain

hack/test/e2e-qemu.sh
Unix4ever: I've changed `apply-config.go` to add `nodeID` key type too, so it should be fine now.

internal/pkg/encryption/keys/kms.go
Unix4ever: actually maybe even 30 seconds is enough. updated

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+		resp, err = client.Seal(ctx, &kms.Request{
+			NodeUuid: s.SystemInformation.UUID,
+			Data:     key,
+		})
+		if err != nil {
+			if status.Code(err) == codes.Unavailable {
+				return retry.ExpectedError(err)
+			}
+
+			return fmt.Errorf("failed to seal KMS passphrase %w", err)
```
smira, line 84: I would add a key slot number and KMS endpoint here

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+	defer cancel()
+
+	conn, err := h.getConn(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error dialing KMS endpoint %q: %w", h.kmsEndpoint, err)
+	}
+
+	client := kms.NewKMSServiceClient(conn)
+
+	s, err := smbios.GetSMBIOSInfo()
```
frezbo, line 62: this is already populated in the systeminformation controller

internal/pkg/encryption/keys/kms.go
```diff
@@ -0,0 +1,165 @@
...
+		resp, err = client.Unseal(ctx, &kms.Request{
+			NodeUuid: s.SystemInformation.UUID,
+			Data:     token.UserData.SealedData,
+		})
+		if err != nil {
+			if status.Code(err) == codes.Unavailable {
+				return retry.ExpectedError(err)
+			}
+
+			return fmt.Errorf("failed to seal KMS passphrase %w", err)
```
smira, line 138: same as above, plus s/seal/unseal/

internal/pkg/encryption/keys/kms.go
Unix4ever: we don't have access to the controllers here.

internal/pkg/encryption/encryption.go
```diff
@@ -230,46 +261,129 @@ func (h *Handler) syncKeys(k *encryption.Key, path string) error {
...
+	//nolint:scopelint
+	sort.Slice(handlers, func(i, j int) bool { return handlers[i].Slot() < handlers[j].Slot() })
+
+	return handlers, nil
+}
+
+func (h *Handler) tryHandlers(handlers []keys.Handler, cb func(h keys.Handler) error) error {
+	var errs error
+
+	for _, h := range handlers {
```
smira, line 347: should we have a retry loop here: we try each slot, if the unlock fails for all them, we do another loop?

if we fail to unlock disk encryption, there's no much we can do to proceed with the boot, so I think we should try hard even e.g. if the KMS is not available, or TPM gives back an error.

internal/pkg/encryption/keys/kms.go
Unix4ever: that's going to be a _very_ long change and lots of places to change...

internal/pkg/encryption/encryption.go
Unix4ever: yeah, good idea. Let me fix that

cmd/talosctl/cmd/talos/upgrade.go
```diff
@@ -147,6 +147,8 @@ func upgradeGetActorID(ctx context.Context, c *client.Client, opts []client.Upgr
 
 func init() {
 	rebootModes := maps.KeysFunc(machine.UpgradeRequest_RebootMode_value, strings.ToLower)
+	sort.Strings(rebootModes)
```
Unix4ever, line 150: Not sure how it worked before.
But to me it seems to trigger dirty check 50% of time.

cmd/talosctl/cmd/talos/upgrade.go
```diff
@@ -148,7 +148,12 @@ func upgradeGetActorID(ctx context.Context, c *client.Client, opts []client.Upgr
 func init() {
 	rebootModes := maps.KeysFunc(machine.UpgradeRequest_RebootMode_value, strings.ToLower)
 	sort.Slice(rebootModes, func(i, j int) bool {
-		return machine.UpgradeRequest_RebootMode_value[rebootModes[i]] > machine.UpgradeRequest_RebootMode_value[rebootModes[j]]
+		var (
```
Unix4ever, line 151: I guess this function had a bug. Map entries are uppercase so it was always comparing nil and nil.

internal/pkg/encryption/keys/kms.go
smira: I don't think passing a context would be that of a nightmare, but it has to be done properly, starting one-off retry loops without cancellation doesn't sound cool to me

Unix4ever: Yeah, I don't mind to add it.
I think I managed to do it.

cmd/talosctl/cmd/talos/upgrade.go
smira: I have another fix for this issue, which is a bit shorter - just don't do `strings.ToLower` until we finish sorting, and do `stirngs.ToLower` as the last step

go.mod
```diff
@@ -92,7 +92,7 @@ require (
 	github.com/siderolabs/discovery-api v0.1.3
 	github.com/siderolabs/discovery-client v0.1.5
 	github.com/siderolabs/gen v0.4.5
-	github.com/siderolabs/go-blockdevice v0.4.5
+	github.com/siderolabs/go-blockdevice v0.4.6-0.20230704154634-fbb01f714bdc
```
smira, line 95: nit: it makes sense to release both go-blockdevice & kms-client before we merge this PR to establish base versions

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2081,6 +2084,13 @@ type EncryptionKeyStatic struct {
 	KeyData string `yaml:"passphrase,omitempty"`
 }
 
+// EncryptionKeyKMS represents a key that is generated and then sealed/unsealed by the KMS server.
+type EncryptionKeyKMS struct {
+	//   description: >
+	//     KMS endpoint to Seal/Unseal the key.
```
smira, line 2090: for documentation purposes, it makes sense to add an example here

internal/pkg/encryption/encryption.go
```diff
@@ -230,46 +264,131 @@ func (h *Handler) syncKeys(k *encryption.Key, path string) error {
...
+	}
+
+	//nolint:scopelint
+	sort.Slice(handlers, func(i, j int) bool { return handlers[i].Slot() < handlers[j].Slot() })
+
+	return handlers, nil
+}
+
+func (h *Handler) tryHandlers(ctx context.Context, handlers []keys.Handler, cb func(h keys.Handler) error) error {
+	return retry.Exponential(time.Minute, retry.WithUnits(time.Second), retry.WithJitter(time.Second)).RetryWithContext(ctx, func(ctx context.Context) error {
```
smira, line 348: I would probably bump the timeout here, as network issues might prevent the machine from booting.

second, I would limit timeout for each attempt (`retry` has this option), so that it never gets stuck "forever" on one of the key slots.

internal/pkg/encryption/encryption.go
```diff
@@ -230,46 +264,131 @@ func (h *Handler) syncKeys(k *encryption.Key, path string) error {
...
+		var errs error
+
+		for _, h := range handlers {
+			err := cb(h)
```
smira, line 352: the retry context should be passed to the `cb`, as the context will have appropriate deadline set

internal/pkg/encryption/keys/kms.go
smira: I think this is a really valid concern - if we want to make anything in Talos to fix "zero" UUIDs, we need to make sure we fetch UUID from a single place everywhere. we don't need to fetch UUID here, we could fetch it before hitting the encryption code and pass it in as a parameter

Unix4ever: Ok, I've made a change.

internal/pkg/mount/system.go
```diff
@@ -182,23 +185,40 @@ func SystemMountPointForLabel(device *blockdevice.BlockDevice, label string, opt
...
 //
 //nolint:gocyclo
-func SystemPartitionMount(r runtime.Runtime, logger *log.Logger, label string, opts ...Option) (err error) {
+func SystemPartitionMount(ctx context.Context, r runtime.Runtime, logger *log.Logger, label string, opts ...Option) (err error) {
 	device := r.State().Machine().Disk(disk.WithPartitionLabel(label))
 	if device == nil {
 		return fmt.Errorf("failed to find device with partition labeled %s", label)
 	}
 
+	systemInformation, err := safe.StateWatchFor[*hardware.SystemInformation](ctx, r.State().V1Alpha2().Resources(), hardware.NewSystemInformation(hardware.SystemInformationID).Metadata(),
```
Unix4ever, line 194: extract to function and limit wait time

internal/pkg/encryption/keys/keys.go
```diff
@@ -6,29 +6,57 @@
...
 		}
 
-		return NewStaticKeyHandler(k)
-	case key.NodeID() != nil:
-		return NewNodeIDKeyHandler()
+		return NewStaticKeyHandler(key, k), nil
+	case cfg.NodeID() != nil:
+		return NewNodeIDKeyHandler(key, opts.PartitionLabel, opts.NodeUUID), nil
+	case cfg.KMS() != nil:
+		return NewKMSKeyHandler(key, cfg.KMS().Endpoint(), opts.NodeUUID)
```
Unix4ever, line 38: if opts.NodeUUID is empty return error

internal/pkg/encryption/encryption.go
```diff
@@ -230,46 +268,132 @@ func (h *Handler) syncKeys(k *encryption.Key, path string) error {
...
+	return handlers, nil
+}
+
+func (h *Handler) tryHandlers(ctx context.Context, handlers []keys.Handler, cb func(ctx context.Context, h keys.Handler) error) error {
+	return retry.Exponential(keyFetchTimeout, retry.WithUnits(time.Second), retry.WithJitter(time.Second), retry.WithAttemptTimeout(time.Second*10)).RetryWithContext(ctx,
+		func(ctx context.Context) error {
+			var errs error
+
+			for _, h := range handlers {
+				err := cb(ctx, h)
```
Unix4ever, line 357: `ctx, cancel = context.WithTimeout(ctx, time.Second * 10)`

DmitriyMV (approved):

PR #7433, opened 2023-07-04
refactor: make maintenance service controller-based

### LINKED ISSUE #7430: Refactor maintenance service to be controller-driven
1. Have configuration for the maintenance service - when it needs to run, which interfaces to listen on, etc.
2. Have a controller to manage TLS certs for the maintenance service.
3. Run maintenance service using a controller, reconfigure it when maintenance config changes.

smira:
Fixes #7430

Introduce a set of resources which look similar to other API implementations: CA, certs, cert SANs, etc.

Introduce a controller which manages the service based on resource state.


pkg/machinery/config/container/container.go
```diff
@@ -226,16 +226,22 @@ func (container *Container) Validate(mode validation.RuntimeMode, opt ...validat
 		warnings, err = container.v1alpha1Config.Validate(mode, opt...)
 	}
 
+	var multiErr *multierror.Error
```
smira, line 229: this is a bug I found with multi-doc config validation (unrelated to this PR), regression tests added

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
```diff
@@ -0,0 +1,275 @@
...
+		Logger: logger.Sugar().Debugf,
+	}
+	tlsProvider := maintenance.NewTLSProvider()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case cfg := <-cfgCh:
+			configResource := config.NewMachineConfigWithID(cfg, config.MaintenanceID)
```
Unix4ever, line 122: Why can't we replace all that with `safe.WriterModify(...)`?

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
```diff
@@ -0,0 +1,275 @@
...
+				if err = configResource.Metadata().SetOwner(oldConfigResource.Metadata().Owner()); err != nil {
+					return fmt.Errorf("error setting owner: %w", err)
+				}
+
+				if err = r.Update(ctx, configResource); err != nil {
+					return fmt.Errorf("failed to update machine config: %w", err)
+				}
+			}
+
+			continue
```
Unix4ever, line 145: Can we replace all that with `safe.WriterModify(...)`?

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
```diff
@@ -0,0 +1,275 @@
...
+					grpc.Creds(
+						credentials.NewTLS(tlsConfig),
+					),
+				),
+
+				factory.WithUnaryInterceptor(injector.UnaryInterceptor()),
+				factory.WithStreamInterceptor(injector.StreamInterceptor()),
+			)
+
+			go func() {
```
Unix4ever, line 226: can we add that to errgoup and do `eg.Wait()` in `shutdownServer` to be on the save side?

Unix4ever (approved):

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
smira: I can look into that, need to make sure we can re-use the waitgroup, as the controller might start the server multiple times

internal/app/machined/pkg/controllers/secrets/maintenance_cert_sans.go
```diff
@@ -0,0 +1,110 @@
...
+
+			spec.AppendIPs(nodeAddresses.TypedSpec().IPs()...)
+			spec.AppendIPs(netip.MustParseAddr("127.0.0.1"))
+			spec.AppendIPs(netip.MustParseAddr("::1"))
+
+			if hostnameStatus != nil {
+				spec.AppendDNSNames(hostnameStatus.TypedSpec().DNSNames()...)
+			}
+
+			spec.FQDN = "maintenance.talos.dev"
```
frezbo, line 99: constants maybe? 
```suggestion
			spec.FQDN = "maintenanceservice.talos.dev"
```

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
smira: this resource is a bit special, it doesn't support changing the spec, and WriterModify requires a "null" value which is not really defined for machine config

frezbo (approved):

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
smira: fixed, added `sync.WaitGroup`

internal/app/machined/pkg/controllers/secrets/maintenance_cert_sans.go
smira: fixed, thanks!

PR #7443, opened 2023-07-07
feat: tpm2 encryption

frezbo:
Unseal doesn't work yet


frezbo:
> Unseal doesn't work yet

This is fixed

hack/tpm2/main.go
```diff
@@ -0,0 +1,809 @@
...
+
+type LuksHeader struct {
+	Type            string   `json:"type"`
+	Keyslots        []string `json:"keyslots"`
+	TPM2PrivateBlob []byte   `json:"tpm2_private_blob"`
+	TPM2PublicBlob  []byte   `json:"tpm2_public_blob"`
+	TPM2PCRS        []int    `json:"tpm2_pcrs"`
+	TPM2Alg         string   `json:"tpm2_alg"`
+	TPM2PolicyHash  []byte   `json:"tpm2_policy_hash"`
+	TPM2PublicKey   string   `json:"tpm2_public_key"`
```
frezbo, line 44: this is not used, not sure if we need to save that in luks header, the only thing we can do is validate the public key did not change for a new uki, but that would prevent from using a new signing key at all. So I guess we can drop it.

hack/tpm2/main.go
```diff
@@ -0,0 +1,809 @@
...
+const (
+	// The SRK handle is defined in the Provisioning Guidance document (see above) in the table "Reserved Handles
+	// for TPM Provisioning Fundamental Elements". The SRK is useful because it is "shared", meaning it has no
+	// authValue nor authPolicy set, and thus may be used by anyone on the system to generate derived keys or
+	// seal secrets. This is useful if the TPM has an auth (password) set for the 'owner hierarchy', which would
+	// prevent users from generating primary transient keys, unless they knew the owner hierarchy auth. See
+	// the Provisioning Guidance document for more details.
+	// https://trustedcomputinggroup.org/resource/tcg-tpm-v2-0-provisioning-guidance
+	// https://trustedcomputinggroup.org/resource/http-trustedcomputinggroup-org-wp-content-uploads-tcg-ek-credential-profile
+	TPM2_SRK_HANDLE = tpmutil.Handle(0x81000001)
```
frezbo, line 75: so this is not used in here, since we generate an ECC primary key everytime with the same seed, but as the comment says, if the tpm has auth set for owner hierarchy we might hit a snag, needs discussion

hack/tpm2/main.go
```diff
@@ -0,0 +1,809 @@
...
+	}
+
+	defer policyCloseFunc()
+
+	pubKey, err := parsePCRSigningPubKey()
+	if err != nil {
+		return nil, err
+	}
+
+	loadExternal := tpm2.LoadExternal{
```
frezbo, line 187: should we switch to ecc keys for signing expected pcr values? The rsa has a limitation that `LoadExternal` only supports rsa key size upto 2048 bits, but for ecc that is not an issue. 

If we change this in future we'll have to handle both cases which could be a mess

hack/tpm2/main.go
```diff
@@ -0,0 +1,809 @@
...
+			FlushHandle: loadExternalResponse.ObjectHandle,
+		}
+
+		_, flushErr := flush.Execute(t)
+		if flushErr != nil {
+			err = flushErr
+		}
+	}()
+
+	pcrSelector, err := createPCRSelection([]int{11})
```
frezbo, line 233: `createPCRSelection` was copied from the `go-tpm` tests, not sure if there's another way to create the byte representation form

hack/tpm2/main.go
```diff
@@ -0,0 +1,809 @@
...
+		return nil, fmt.Errorf("failed to execute policy get digest: %v", err)
+	}
+
+	var sigJSON PCRData
+
+	if err := json.Unmarshal(pcrDataJSON, &sigJSON); err != nil {
+		return nil, err
+	}
+
+	// TODO: verify the fingerprint matches the one in the JSON
```
frezbo, line 270: this is just a fail early situation

frezbo:
Superseded by #7464

PR #7448, opened 2023-07-07
Preserve bond and ip kernel params

jared-gs:
# Pull Request

## What? (description)

The bond and ip kernel params are typically used to configure networking on the host and should be preserved when Talos is installed

## Acceptance

### DIFF
--- a/api/resource/definitions/cluster/cluster.proto
+++ b/api/resource/definitions/cluster/cluster.proto
@@ -66,4 +66,3 @@ message MemberSpec {
   string operating_system = 5;
   ControlPlane control_plane = 6;
 }
-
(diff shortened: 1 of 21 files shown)


smira:
we would like to discuss this further, as it was supposed that kernel args only configure networking on the first boot (before machine config is acquired).

also there's a way to perform a special configuraiton of network params via `META` partition. in either case, I would love to see your flow and usecase to understand it better

zviratko:
This would actually break my flow if implemented badly - I use a provisioning VLAN/Subnet on first boot but then I switch to a completely different one and I don't expect Talos to use an interface I didn't provision.

PR #7449, opened 2023-07-07
feat: seed the kernel random pool from the TPM

smira:
Use the TPM2 feature to provide high-quality random bytes.

### DIFF
--- a/internal/pkg/rng/tpm.go
+++ b/internal/pkg/rng/tpm.go
@@ -1,0 +1,87 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package rng
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"time"
+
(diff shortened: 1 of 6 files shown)


internal/pkg/rng/tpm.go
```diff
@@ -0,0 +1,87 @@
...
+
+		if len(resp.RandomBytes.Buffer) == 0 {
+			return fmt.Errorf("received zero random bytes from the TPM: %w", err)
+		}
+
+		if err = WriteEntropy(resp.RandomBytes.Buffer); err != nil {
+			return fmt.Errorf("error writing random pool entropy: %w", err)
+		}
+
+		remaining -= len(resp.RandomBytes.Buffer)
```
frezbo, line 81: so what's about crediting versus just seeding?

frezbo (approved):

internal/pkg/rng/tpm.go
smira: systemd doesn't credit, so I thought we should do same

frezbo: yes, i didn't know the difference, finally read this: https://systemd.io/RANDOM_SEEDS/

PR #7450, opened 2023-07-10
feat: drop tpm simulator for ukify measure.

frezbo:
We do not need a tpm simulator for ukify measure. We can pre-calculate the values. This also means we can build ukify as a static binary.


hack/ukify/measure/extend/extend.go
```diff
@@ -0,0 +1,41 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
frezbo, line 1: we should move this to `internal/tpm2` or something when we move `ukify` into installer code

smira (approved):
:cool:

PR #7451, opened 2023-07-10
feat: update Kubernetes to 1.28.0-alpha.4

smira:
The Go modules were not tagged for alpha.4, so using alpha.3 tag.

Talos 1.5 will ship with Kubernetes 1.28.0.


internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -109,8 +109,7 @@ func (ctrl *KubeletSpecController) Run(ctx context.Context, r controller.Runtime
 		expectedNodename := nodenameSpec.Nodename
 
 		args := argsbuilder.Args{
-			"container-runtime-endpoint": "unix://" + constants.CRIContainerdAddress,
```
smira, line 112: migrated deprecated flag to kubelet config option (it's available for 16 months now)

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -271,18 +270,11 @@ func NewKubeletConfiguration(cfgSpec *k8s.KubeletConfigSpec) (*kubeletconfig.Kub
 	config.KubeletCgroups = constants.CgroupKubelet
 	config.RotateCertificates = true
 	config.ProtectKernelDefaults = true
+	config.ContainerRuntimeEndpoint = "unix://" + constants.CRIContainerdAddress
 
+	// SeccompDefault feature gate is enabled by default Kubernetes 1.25+, GA in 1.27
```
smira, line 275: no need to enable feature gate anymore (enabled in 1.25+), and kubelet complains about trying to enable GA flag

PR #7454, opened 2023-07-10
fix: rewrite encryption system information flow

Unix4ever:
Pass getter to the key handler instead of already fetched node uuid.

### DIFF
--- a/internal/pkg/mount/system.go
+++ b/internal/pkg/mount/system.go
@@ -110,7 +110,7 @@ func SystemMountPointForLabel(ctx context.Context, device *blockdevice.BlockDevi
 			device,
 			part,
 			o.Encryption,
-			o.NodeParams,
+			o.SystemInformationGetter,
 		)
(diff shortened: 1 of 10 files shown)


internal/pkg/mount/system.go
```diff
@@ -190,8 +190,12 @@ func SystemPartitionMount(ctx context.Context, r runtime.Runtime, logger *log.Lo
 	}
 
 	systemInformation, err := r.GetSystemInformation(ctx)
-	if err != nil && !errors.Is(err, context.Canceled) && state.IsNotFoundError(err) {
-		return err
+	if err != nil {
```
smira, line 193: I don't think this has anything to do with timeouts, as it fails after installing Talos, which is way late in the boot sequence

Unix4ever: yes, I found the actual bug.

smira (approved):

PR #7457, opened 2023-07-10
feat: implement CRI image management and pre-pull on K8s upgrade

### LINKED ISSUE #6391: pre-pulling Kubernetes images on upgrades
### Description

When using `talosctl upgrade-k8s --to 1.25.2` prior to the Sidero custom image for that version being available, upgrading puts the cluster into a state where I must manually revert image versions.  This is likely safe when upgrading a patch value, but I'm not entirely sure what would happen if this were (say) 1.26.0 as the target.

The proposal here is that, for the control nodes at least, we know what versions will be necessary.  Pre-pulling these images (like kubeadm can do) solves several issues and adds a significant safety net:

* Pre-pulling makes the following upgrade-downtime shorter, as the images are already present.
* Pre-pulling ensures the upgrade will succeed because everything needed is already present.
* Pre-pulling in a mixed location "kubespan" and an air-gapped setup could be very complex to "prime" to ensure the images are pullable, and the logs from "upgrade-k8s" doesn't indicate when an image is not available, which requires debugging.

When I used kubeadm, I always pre-pulled images.  Always.  I'd like the same safety net for talos.

### Environment

- Talos version: Any (1.2.3 used)
- Kubernetes version: Upgrade phase from 1.25.1 to 1.25.2 (prior to sidero images)
- Platform: N/A

smira:
Fixes #6391

Implement a set of APIs and commands to manage images in the CRI, and pre-pull images on Kubernetes upgrades.


frezbo (approved):

frezbo:
once we have the mirroring we could potentially even only pull once on a node and then share b/w nodes

maxpain:
Hello. Is it possible to pull images from private registries?

frezbo:
> Hello. Is it possible to pull images from private registries?

if you have the registry mirrors configured in machine config, it would pre-pull the k8s images

PR #7458, opened 2023-07-11
chore: bump sd-boot to v254-rc1

frezbo:
Bump sd-boot.
Fix parsing PE executable offsets.
Set the PE file alignment to be 512 bytes.


cmd/installer/pkg/iso.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	// UKIISOSize is the size of the UKI ISO.
-	UKIISOSize = 120 * 1024 * 1024
+	UKIISOSize = 80 * 1024 * 1024
```
frezbo, line 22: fixing the alignment means we can use a proper ISO size.

hack/ukify/main.go
```diff
@@ -107,9 +107,16 @@ func buildUKI(source, output string, sections []section) error {
 	// find the first VMA address
 	lastSection := peFile.Sections[len(peFile.Sections)-1]
 
-	const alignment = 0xfff
+	// align the VMA to 512 bytes
+	// https://github.com/saferwall/pe/blob/main/helper.go#L22-L26
+	const alignment = 0x200
```
smira, line 112: this should be `0x1ff` (alignment - 1) as we use it this way

smira (approved):

Issue #7423, opened 2023-06-29
upgrade to 1.4.6 breaks etcd


evanrich:
### Description

Have a 3 node cluster, 192.168.xxx.{10,11,12}
Talos 1.4.5
K8s: 1.27.2

Ran the following command to update first node:
```
talosctl upgrade -n 192.168.xxx.10 --image ghcr.io/siderolabs/installer:v1.4.6 
```

### Logs
```
◳ watching nodes: [192.168.xxx.10]
    * 192.168.xxx.10: 1 error(s) occurred:
    sequence error: sequence failed: error running phase 4 in upgrade sequence: task 1/1: failed, failed to leave cluster: 1 error(s) occurred:
    failed to remove member 5218892699632570834: etcdserver: server stopped
```

### Environment

- Talos version:1.4.5
- Kubernetes version: 1.27.2
- Platform: Dell Optiplex 7060 (bare metal)

ETCD just starts/stops/starts/stops on the node, cannot re-run upgrade command on it either.

evanrich:
FWIW, resetting the node and upgrading it a second time worked. But thought I'd post this here since an out of the box 1.4.5->1.4.6 had an issue.

smira:
This issue doesn't have enough information to proceed, if the member wasn't removed, after a reboot a node should come back healthy. The root cause is not clear, it might be that some other machine was not healthy/ready at that point.

cehoffman:
I encountered this exact same symptom. The only real difference is I had already done an upgrade to Kubernetes 1.27.3.

```
talosctl upgrade -n 192.168.1.82 --image ghcr.io/siderolabs/installer:v1.4.6                                                                                                                                         01:29 on Tue Jul 04
◱ watching nodes: [192.168.1.82]
    * 192.168.1.82: 1 error(s) occurred:
    sequence error: sequence failed: error running phase 4 in upgrade sequence: task 1/1: failed, failed to leave cluster: 1 error(s) occurred:
    failed to remove member 13637399532505415876: etcdserver: server stopped
```

This was from a 3 node talos cluster on 1.4.5. There were the expected 3 ETCD members at the start of this upgrade. Once the error stopped the upgrade there were only 2 ETCD members. I had already completed one node upgrade in the sequence and everything had returned to healthy before continuing.

smira:
@cehoffman can you please provide the full log if that happens again (`--debug` flag should be enough)?

or `talosctl dmesg` before the reboot, so that we can figure out what goes wrong.

[...]

PR #7462, opened 2023-07-11
fix: retry CRI pod removal, fix upgrade flow in the tests

smira:
It seems that CRI has a bit of eventual consistency, and it might fail to remove a stopped pod failing that it's still running.

Rewrite the upgrade API call in the upgrade test to actually wait for the upgrade to be successful, and fail immediately if it's not successfull. This should improve the test stability and it should make it easier to find issues immediately.


smira:
See also #7423 (this is not a fix, but related problem)

frezbo (approved):

Issue #7266, opened 2023-05-23
Encrypting STATE/EPHEMERAL with TPM2 and signed PCR values


frezbo:
## Feature Request

Research if cryptsetup supports TPM2 enrolling with signature

TPM2 primary key enrolling (this seems a good thing to read https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf)

PR #7464, opened 2023-07-11
feat: tpm2 based disk encryption

frezbo:
Support disk encryption using tpm2 and pre-calculated signed PCR values.

Fixes: #7266


internal/app/init/main.go
```diff
@@ -58,6 +59,11 @@ func run() (err error) {
 		log.Printf("failed to seed from the TPM: %s", err)
 	}
 
+	// extend PCR 11 with enter-initrd
+	if err = tpm2.PCRExtent(constants.UKIMeasuredPCR, []byte("enter-initrd")); err != nil {
```
smira, line 63: should we use a Phase constant here?

smira (approved):
:cool: I want to do a separate PR to bump Go to 1.20.6, but it's all good with some small nits

frezbo:
> cool I want to do a separate PR to bump Go to 1.20.6, but it's all good with some small nits

yup, go bump needs a separate PR, i just wanted to test in CI

internal/app/init/main.go
frezbo: fixed

PR #7469, opened 2023-07-12
refactor: move ukify into Talos code

### LINKED ISSUE #7470: move ukify to talos tree

smira:
This is intemediate step to move parts of the `ukify` down to the main Talos source tree, and call it from `talosctl` binary.

The next step will be to integrate it into the imager and move `.uki` build out of the Dockerfile.

Fixes #7470


pkg/splash/splash.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	_ "embed"
+)
+
+//go:embed sidero.bmp
+var sideroBMP []byte
+
+// GetBoot returns boot splash image.
+func GetBoot() []byte {
```
frezbo, line 16: Get
```suggestion
func GetSplash() []byte {
```

smira: `splash.GetSplash()` sounds weird, probably get `splash.GetEarlyBootBMP()` or something like that

cmd/talosctl/cmd/mgmt/gen/secureboot.go
```diff
@@ -43,22 +43,18 @@ var genSecurebootUKICmd = &cobra.Command{
...
 // genSecurebootPCRCmd represents the `gen secureboot pcr` command.
 var genSecurebootPCRCmd = &cobra.Command{
 	Use:   "pcr",
-	Short: "Generates a certificate which is used to sign TPM PCR values",
+	Short: "Generates a key which is used to sign TPM PCR values",
 	Long:  ``,
 	Args:  cobra.NoArgs,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		return generateSigningCerts(genSecurebootCmdFlags.outputDirectory, "pcr", genSecurebootPCRCmdFlags.commonName, 2048)
+		return generateSigningCerts(genSecurebootCmdFlags.outputDirectory, "pcr", "dummy", 2048, false)
```
frezbo, line 57: `dummy` intentional?

smira: yes, because we actually only need the key, there is no cert, but our library always goes through cert & key generation

internal/pkg/secureboot/measure/internal/pcr/bank_data_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/siderolabs/talos/internal/pkg/secureboot"
+	"github.com/siderolabs/talos/internal/pkg/secureboot/measure/internal/pcr"
+	talostpm2 "github.com/siderolabs/talos/internal/pkg/tpm2"
+)
+
+func TestCalculateBankData(t *testing.T) {
+	t.Parallel()
+
+	pemKey, err := os.ReadFile("../../testdata/pcr-signing-key.pem")
```
frezbo, line 24: embed maybe?

internal/pkg/secureboot/measure/internal/pcr/policy.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+
+	"github.com/google/go-tpm/tpm2"
+)
+
+// CalculatePolicy calculates the policy hash for a given PCR value and PCR selection.
+func CalculatePolicy(pcrValue []byte, pcrSelection tpm2.TPMLPCRSelection) []byte {
```
frezbo, line 15: there is a slightly duplicate version of this in `internal/tpm2`, we could combine them, but it's also fine to keep as is.

frezbo (approved):

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -68,7 +69,7 @@ func Switch(prefix string, mountpoints *mount.Points) (err error) {
 	}
 
 	// extend PCR 11 with leave-initrd
-	if err = tpm2.PCRExtent(constants.UKIMeasuredPCR, []byte(tpm2.LeaveInitrd)); err != nil {
+	if err = tpm2.PCRExtent(constants.UKIMeasuredPCR, []byte(secureboot.LeaveInitrd)); err != nil {
```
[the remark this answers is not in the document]
smira: I dropped the tmp2/constants.go file completely, or you mean something else?

internal/pkg/secureboot/measure/internal/pcr/policy.go
smira: I will take a look

internal/pkg/secureboot/measure/internal/pcr/bank_data_test.go
smira: I think in this case it probably makes less sense, as it's shared across multiple tests, wdyt?

we also use it in other places as a file path, so we to be consistent I'm not embedding it

frezbo: yup, makes sense

internal/pkg/secureboot/measure/internal/pcr/policy.go
smira: it feels like `tpm2` version is slightly different, so I don't feel confident unifying them, even though I see some overlap

frezbo: ahh yes, one passes in existing digest, other starts with zero

Issue #7432, opened 2023-07-04
rename 'in-cluster Kubernetes API load balancer' to KubePrism


smira:
Rename:

* machine config feature name
* release notes
* controllers/resources as appropriate

TimJones:
Or (Talos|Kube)Prism ...

sergelogvinov:
Is it possible to customize the listen IP address too?

smira:
> Is it possible to customize the listen IP address too?

no, it will always listen on localhost (never exposing it to the outside world)

PR #7479, opened 2023-07-14
chore: change missing renames of KubePrism

DmitriyMV:
For #7432

### DIFF
--- a/pkg/machinery/config/generate/options.go
+++ b/pkg/machinery/config/generate/options.go
@@ -34,10 +34,10 @@ func WithLocalAPIServerPort(port int) Option {
 	}
 }
 
-// WithAPIServerBalancerPort specifies the local API server load balancer port. If 0, lod balancer is disabled.
-func WithAPIServerBalancerPort(port int) Option {
+// WithKubePrismPort specifies the KubePrism port. If 0, lod balancer is disabled.
+func WithKubePrismPort(port int) Option {
 	return func(o *Options) error {
-		o.APIServerBalancerPort = port
+		o.KubePrismPort = port
 
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


smira (commented):
we had the flag in the e2e-qemu.sh, it needs to be updated

DmitriyMV:
@smira Its [already](https://github.com/siderolabs/talos/blob/main/hack/test/e2e-qemu.sh#L135) updated.

smira:
> @smira Its [already](https://github.com/siderolabs/talos/blob/main/hack/test/e2e-qemu.sh#L135) updated.

:shrug: how did it pass the tests then?

DmitriyMV:
> how did it pass the tests then?

Because the flag name was updated, but flag variable name wasn't.

smira:
> > how did it pass the tests then?
> 
> Because the flag name was updated, but the actual variable name wasn't.

got it, thanks

smira (approved):

pkg/machinery/config/generate/options.go
```diff
@@ -34,10 +34,10 @@ func WithLocalAPIServerPort(port int) Option {
 	}
 }
 
-// WithAPIServerBalancerPort specifies the local API server load balancer port. If 0, lod balancer is disabled.
-func WithAPIServerBalancerPort(port int) Option {
+// WithKubePrismPort specifies the KubePrism port. If 0, lod balancer is disabled.
```
frezbo, line 37:
```suggestion
// WithKubePrismPort specifies the KubePrism port. If 0, load balancer is disabled.
```

frezbo (approved):

pkg/machinery/config/generate/options.go
DmitriyMV: Lol, missed like by 0.1 second :D

DmitriyMV: I'll update it in the next PR.

PR #7480, opened 2023-07-14
fix: tpm2 encrypt/decrypt flow

frezbo:
The previous flow was using TPM PCR 11 values to bound the policy which means TPM cannot unseal when UKI changes. Now it's fixed to use PCR 7 which is bound to the SecureBoot state (SecureBoot status and Certificates). This provides a full chain of trust bound to SecureBoot state and signed PCR signature.

Also the code has been refactored to use PolicyCalculator from the TPM library.


smira (approved):
:heart:

PR #7493, opened 2023-07-19
chore: basic support for shutdown/poweroff flags

frezbo:
This adds basic support for shutdown/poweroff flags. it can distringuish between halt/shutdown/reboot.

In the case of Talos halt/shutdown is same op.


frezbo:
Should also address: https://github.com/siderolabs/talos/pull/7409#issuecomment-1610104060

and makes it easier to test the qemu-guest agent extension in CI

internal/app/poweroff/main.go
```diff
@@ -32,6 +35,62 @@ func Main() {
 		log.Fatalf(fmt.Errorf("error while creating machinery client: %w", err).Error())
 	}
 
+	// if we have args, try to parse them in a simple way
+	// combination of args is not supported, args are based on shutdown/poweroff man page
+	if len(args) > 1 {
```
smira, line 40: we could probably DRY it up a bit, by simply calculating first an action: `shutdown`/`reboot`, and executing it next.

and we could even make it a function: `action(args) -> shutdown/reboot`, and add some unit-tests for it

frezbo: fixed

smira (approved):

PR #7494, opened 2023-07-19
chore: support creating qemu guest socket

frezbo:
Support creating a qemu guest agent socket so we can test `qemu-guest-agent` extension in CI.

Ref: https://github.com/siderolabs/extensions/pull/173#issuecomment-1611911106

### DIFF
--- a/pkg/provision/providers/qemu/launch.go
+++ b/pkg/provision/providers/qemu/launch.go
@@ -255,6 +255,12 @@ func launchVM(config *LaunchConfig) error {
 		"-no-reboot",
 		"-boot", fmt.Sprintf("order=%s,reboot-timeout=5000", bootOrder),
 		"-smbios", fmt.Sprintf("type=1,uuid=%s", config.NodeUUID),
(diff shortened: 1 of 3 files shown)


pkg/provision/providers/qemu/launch.go
```diff
@@ -257,6 +258,18 @@ func launchVM(config *LaunchConfig) error {
 		"-smbios", fmt.Sprintf("type=1,uuid=%s", config.NodeUUID),
 	}
 
+	if config.QEMUGuestSocketEnabled {
+		args = append(
```
smira, line 262: should we just always create the guest agent socket? is there any harm in enabling it?

frezbo: no, it's fine, I'll update then

smira: I mean we could simplify by having it enabled always, one less flag to support :-D

smira (approved):

PR #7495, opened 2023-07-19
fix: log explicitly when kubelet has no nodeIP match

### LINKED ISSUE #7487: Kubelet starts with empty --node-ip flag
### Description
Kubelet starts with an empty --node-ip flag. Example of the wrong configuration:
```yaml
machine:
    kubelet:
        nodeIP:
            validSubnets:
            - '!10.0.0.0/8'
```
Here I forgot to add `- "0.0.0.0/0"` and kubelet was started with an empty `--node-ip` flag, but actually, it should crash.

### Logs

```bash
/usr/local/bin/kubelet --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubeconfig --cert-dir=/var/lib/kubelet/pki --config=/etc/kubernetes/kubelet.yaml --container-runtime-endpoint=unix:///run/containerd/containerd.sock --hostname-override=s1-ndstech-sip --kubeconfig=/etc/kubernetes/kubeconfig-kubelet --node-ip= --node-labels=server=game,relay=true
```

`talosctl get nodeaddresses -o yaml`:
```yaml
node: 123.123.123.123
metadata:
    namespace: k8s
    type: NodeIPs.kubernetes.talos.dev
    id: kubelet
    version: 1
    owner: k8s.NodeIPController
    phase: running
    created: 2023-07-14T13:42:50Z
    updated: 2023-07-14T13:42:50Z
spec:
    addresses: []
```

### Environment

- Talos version: v1.4.6
- Kubernetes version: v1.27.3
- Platform: Bare metal

smira:
Fixes #7487

When `.kubelet.nodeIP` filters yield no match, Talos should not start the kubelet, as using empty address list results in `--node-ip=` empty kubelet arg, which makes kubelet pick up "the first" address.

Instead, skip updating (creating) the nodeIP and log an explicit warning.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/address_filter_test.go
+++ b/internal/app/machined/pkg/controllers/k8s/address_filter_test.go
@@ -176,5 +176,7 @@ func (suite *K8sAddressFilterSuite) TearDownTest() {
 }
 
 func TestK8sAddressFilterSuite(t *testing.T) {
+	t.Parallel()
+
 	suite.Run(t, new(K8sAddressFilterSuite))
 }
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod_test.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod_test.go
@@ -773,5 +773,7 @@ func (suite *ControlPlaneStaticPodSuite) TearDownTest() {
 }
 
 func TestControlPlaneStaticPodSuite(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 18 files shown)


frezbo (approved):
I guess all the `t.Parallel` is from golangci-lint

smira:
> I guess all the `t.Parallel` is from golangci-lint

this was me trying to optimize time of the unit-tests, for the `k8s/` package it went down 5s -> 2s. I think we need to do that across all packages, and also migrate to `rtestutils`, but not in one step :)

PR #7496, opened 2023-07-20
refactor: rewrite the way Talos acquires the machine configuration

### LINKED ISSUE #7453: Refactor `loadConfig` task to be a controller
Acquire config in a loop until the `v1alpha1` full config is available (accepting multi-doc config documents on the way):

1. Load from `STATE`.
2. Load from platform.
3. Load from maintenance service.

smira:
Fixes #7453

The goal is to make it possible to load some multi-doc configuration from the platform source (or persisted in STATE) before machine acquires full configuraiton.


internal/app/machined/pkg/controllers/config/acquire.go
```diff
@@ -0,0 +1,437 @@
...
+			}
+
+			return fmt.Errorf("failed to get spec: %w", err)
+		}
+
+		// run the state machine
+		for {
+			newState, cfg, err := currentState(ctx, r, logger)
+			if err != nil {
+				ctrl.EventPublisher.Publish(ctx, &machineapi.ConfigLoadErrorEvent{
```
frezbo, line 139: what's different between EventPublisher and PlatformEvent?

internal/app/machined/pkg/controllers/config/acquire.go
```diff
@@ -0,0 +1,437 @@
...
+
+		return nil, fmt.Errorf("failed to stat %s: %w", ctrl.ConfigPath, err)
+	}
+
+	cfg, err := configloader.NewFromFile(ctrl.ConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load config from STATE: %w", err)
+	}
+
+	// this is legacy flow with persistence, it should be deprecated & removed
```
frezbo, line 219: also add a TODO maybe, so it's easier to search

internal/app/machined/pkg/controllers/config/acquire.go
```diff
@@ -0,0 +1,437 @@
...
+func (ctrl *AcquireController) statePlatform(ctx context.Context, r controller.Runtime, logger *zap.Logger) (stateMachineFunc, config.Provider, error) {
+	cfg, err := ctrl.loadFromPlatform(ctx, logger)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	switch {
+	case cfg == nil:
+		fallthrough
+	case !cfg.CompleteForBoot():
```
frezbo, line 253: so any state that's not CompleteForBoot is assumed to be done I guess, wonder if there's some state we'd need to handle :thinking:

internal/app/machined/pkg/controllers/config/acquire.go
smira: this is part of extracting from v1alpha1_sequencer:

* `EventPublisher` is Talos event stream
* `PlatformEvent` is an event sent via platform-specific API, only implemented for Equinix Metal

internal/app/machined/pkg/controllers/config/acquire_test.go
```diff
@@ -0,0 +1,455 @@
...
+		return nil
+	})
+
+	ev := suite.platformEvent.getEvents()[0]
+	suite.Assert().Equal(platform.EventTypeFailure, ev.Type)
+	suite.Assert().Equal("Error loading and validating Talos machine config.", ev.Message)
+	suite.Assert().Equal("failed to load config from STATE: unknown keys found during decoding:\naaaversion: v1alpha1 # Indicates the schema used to decode the contents.\n", ev.Error.Error())
+
+	suite.Assert().Equal(&machineapi.ConfigLoadErrorEvent{
+		Error: "failed to load config from STATE: unknown keys found during decoding:\naaaversion: v1alpha1 # Indicates the schema used to decode the contents.\n",
```
frezbo, line 250: error seems weird, we only wrote `aaa`, why does error says v1alpha1 :thinking:

smira: we prepend 'aaa' to the machine config, so that it was:

```yaml
version: v1alpha1
...
```

and becomes:

```yaml
aaaversion: v1alpha1
...
```

internal/app/machined/pkg/controllers/config/acquire_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+	cfg := suite.waitForConfig()
+	suite.Require().Equal(cfg.Cluster().Name(), suite.clusterName)
+
+	suite.Assert().Empty(suite.eventPublisher.getEvents())
+	suite.Assert().Equal(
+		[]platform.Event{
+			{
+				Type:    platform.EventTypeConfigLoaded,
+				Message: "Talos machine config loaded successfully.",
```
frezbo, line 311: maybe we update the even with the source too, then message becomes more clear from where the config was loaded from

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo, line 1: finally this got cleaned up, the new code is more easily readable :cool:

frezbo (approved):

internal/app/machined/pkg/controllers/config/acquire_test.go
frezbo: ahh I completely missed the prepending, now I see it

internal/app/machined/pkg/controllers/config/acquire_test.go
smira: I tried to keep things as compatible as it can be, idk if Equinix is using this event literally? it will be only visible on the EM console at the moment iirc

frezbo: i meant the general message that goes to dmesg, something like: `Talos machine config loaded successfully from disk` or `Talos machine config loaded successfully from platform`

smira: ah, this makes sense... it might be actually multiple sources, need to think about it...

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: yes, this was a mess to clean up, and also add new features :)

internal/app/machined/pkg/controllers/config/acquire.go
smira: done

internal/app/machined/pkg/controllers/config/acquire_test.go
smira: added sources

frezbo (approved):

PR #7500, opened 2023-07-20
feat: enforce kernel lockdown for UKI

smira:
UKI is meant to be for UEFI Secure Boot, so it's expected to enforce kernel lockdown. We might reconsider in the future to use a kernel patch instead: https://github.com/SUSE/kernel/commit/b1a0314b0827ea781c13a27abcc3c3b96fc0e946

### DIFF
--- a/cmd/ukify/main.go
+++ b/cmd/ukify/main.go
@@ -34,6 +34,10 @@ func run() error {
 		return err
 	}
 
+	if err := defaultCmdline.AppendAll(kernelpkg.SecureBootArgs); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/ukify/main.go
```diff
@@ -38,6 +38,10 @@ func run() error {
 		return err
 	}
 
+	if err := defaultCmdline.AppendAll([]string{"lockdown=confidentiality"}); err != nil {
```
frezbo, line 41: constants maybe?

smira: updated

frezbo (approved):

PR #7505, opened 2023-07-22
docs: format source code for new documentation framework

andrewrynhard:
This includes some minor formatting changes in order to be compatible with our internal documentation framework.


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1212,9 +1212,7 @@ type KubeletNodeIPConfig struct {
 
 // NetworkConfig represents the machine's networking config values.
 type NetworkConfig struct {
-	//   description: |
-	//     Used to statically set the hostname for the machine.
-	NetworkHostname string `yaml:"hostname,omitempty"`
+	NetworkHostname NetworkHostname `yaml:"hostname,omitempty" optional:"true"` // Used to statically set the hostname for the machine.
```
smira, line 1215: this is going to break a lot of things in config documentation/marshalling

andrewrynhard: Yep. This is a draft. It will make a lot more sense when I present the new documentation site. TLDR is that the documentation generation will be generated in an entirely new way.

PR #7522, opened 2023-07-25
docs: update go structs with updated format for docgen

andrewrynhard:


smira:
@andrewrynhard can you please use your fork for the PR?

andrewrynhard:
> @andrewrynhard can you please use your fork for the PR?

I feel that the Talos repo would be a better place for it. I am preparing for the strong possibility that this will not make it into 1.5. My idea is to make this branch the "alpha/beta" official docs as we work through things in 1.6. We will setup the new external docs to pull from here as we work through issues. This is a better space to collaborate with higher traffic and more eyes.

smira:
I don't get it, I'm saying about going from your own fork of `talos` repo, vs. pushing here, as it makes CI build things wrong. This has nothing to do with 1.5/1.6.

andrewrynhard:
> I don't get it, I'm saying about going from your own fork of `talos` repo, vs. pushing here, as it makes CI build things wrong. This has nothing to do with 1.5/1.6.

Maybe it will make more sense after tomorrow.

PR #7536, opened 2023-07-27
feat: implement image generation profiles

smira:
Support full configuration for image generation, including image outputs, support most features (where applicable) for all image output types, unify image generation process.


cmd/installer/pkg/imager/input.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// SecureBootAssets describes secureboot assets.
+type SecureBootAssets struct {
+	// SecureBoot signing key & cert.
+	SigningKeyPath  string `yaml:"signingKeyPath"`
+	SigningCertPath string `yaml:"signingCertPath"`
+	// PCR signing key & public key.
+	PCRSigningKeyPath string `yaml:"pcrSigningKeyPath"`
+	PCRPublicKeyPath  string `yaml:"pcrPublicKeyPath"`
```
frezbo, line 38: don;t we need the auto-enroll db file too?

smira: yeah, we need that, I'm still updating things, not sure where to put it yet (as it's ISO-specific).

frezbo: just a note, aws/cloud providers needs a different format though (we might need it eventually) https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/uefi-secure-boot-optionB.html

Makefile
```diff
@@ -164,8 +164,7 @@ docker buildx create --driver docker-container --name local --buildkitd-flags '-
 
 If you already have a compatible builder instance, you may use that instead.
 
-> Note: The security.insecure entitlement is only required, and used by the unit-tests target and targets which build container images
-for applications using `img` tool.
+> Note: The security.insecure entitlement is only required, and used by the unit-tests target.
```
frezbo, line 167: doesn;t generation of images still need that, or is this updated because it's now a `docker run` command?

pkg/imager/extensions/rebuild.go
```diff
@@ -0,0 +1,85 @@
...
+	pipeR, pipeW, err := os.Pipe()
+	if err != nil {
+		return err
+	}
+
+	defer pipeR.Close() //nolint:errcheck
+	defer pipeW.Close() //nolint:errcheck
+
+	// build cpio image which contains .sqsh images and extensions.yaml
+	cmd1 := exec.Command("cpio", "-H", "newc", "--create", "--reproducible")
```
frezbo, line 32: let's add a todo here to use the go equivalent, i think i used some when doing the extensions module support

pkg/machinery/constants/constants.go
```diff
@@ -517,8 +517,14 @@ const (
 	// UKIAssetPath is the path to the UKI in the installer.
 	UKIAssetPath = "/usr/install/%s/" + UKIAsset
 
+	// SDStubAsset defines a well known name for our systemd-stub filename.
+	SDStubAsset = "systemd-stub.efi"
+
+	// SDStubAssetPath is the path to the systemd-styff in the installer.
```
frezbo, line 523:
```suggestion
	// SDStubAssetPath is the path to the systemd-stub in the installer.
```

Makefile
smira: nope, it was actually never needed for a long time. we previously used `img` inside buildkit to build images of containers of services on talos rootfs, then it was refactored, then we dropped containers, so it works without it ;)

pkg/machinery/constants/constants.go
smira: lol... :-D

pkg/imager/extensions/rebuild.go
smira: I'm not sure if it supports all the stuff we use here, but certainly interesting to make it fully in Go, this way we could ship it as a binary

frezbo: we also need the option for auto-enroll if it's uki

smira: in the initramfs?

frezbo: the auto-enroll files are in the EFI partition

smira: yes, I'm far away from it right now, this code installs system extensions in the initramfs :)

frezbo: just brought up something I remembered

internal/pkg/secureboot/uki/assemble.go
```diff
@@ -69,7 +69,13 @@ func (builder *Builder) assemble() error {
 
 	args = append(args, builder.SdStubPath, builder.unsignedUKIPath)
 
-	cmd := exec.Command("objcopy", args...)
+	objcopy := "/usr/x86_64-alpine-linux-musl/bin/objcopy"
```
frezbo, line 72: i hate this, but i guess let's live with this until we completely remove dependency on objcopy

pkg/.importvet.yaml
frezbo, line 1: on purpose?

smira: :cry: yes... as `pkg/imager` imports from internals, but I want it to be in the public API, so one could build integrations with `imager`

frezbo (approved):

pkg/imager/imager.go
```diff
@@ -0,0 +1,287 @@
...
+		return err
+	}
+
+	if i.prof.SecureBootEnabled() {
+		if err = cmdline.AppendAll(kernel.SecureBootArgs); err != nil {
+			return err
+		}
+	}
+
+	// meta values can be written only to the "image" output
```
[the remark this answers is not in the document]
smira: yep, that's why we push them to the environment via kernel args to support meta values

`image` will have it written directly, while other ouputs will get injected via cmdline

pkg/imager/profile/default.go
```diff
@@ -0,0 +1,364 @@
...
+			Kind:      OutKindImage,
+			OutFormat: OutFormatXZ,
+			ImageOptions: &ImageOptions{
+				DiskSize:   DefaultRAWDiskSize,
+				DiskFormat: DiskFormatRaw,
+			},
+		},
+	},
+	// SBCs
+	"rpi_4": {
```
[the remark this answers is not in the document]
smira: sure, let me remove it then

frezbo (approved):

pkg/imager/profile/default.go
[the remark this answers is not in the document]
smira: removed it

pkg/imager/profile/profile.go
```diff
@@ -0,0 +1,155 @@
...
+			return fmt.Errorf("board is only supported for metal arm64")
+		}
+	}
+
+	if p.SecureBootEnabled() {
+		switch p.Output.Kind { //nolint:exhaustive
+		case OutKindISO, OutKindImage, OutKindInstaller, OutKindUKI:
+		default:
+			return fmt.Errorf("secureboot is not supported for %s output", p.Output.Kind)
+		}
```
DmitriyMV, line 79:
```suggestion
	if p.SecureBootEnabled() && !In(p.Output.Kind, OutKindISO, OutKindImage, OutKindInstaller, OutKindUKI) {
		return fmt.Errorf("secureboot is not supported for %s output", p.Output.Kind)
	}
```

DmitriyMV (approved):

Issue #7531, opened 2023-07-26
Talos upgrade doesn't use fast wipe


frezbo:
## Bug Report
Talos upgrade doesn't use fast wipe

### Description
This was probably caused from the bootloader code refactoring

### Logs

```text

172.16.56.129: user: warning: [2023-07-26T16:35:57.492593366Z]: 2023/07/26 16:35:57 zeroing out "/dev/nvme0n1p5"
172.16.56.129: user: warning: [2023-07-26T16:35:57.850023366Z]: 2023/07/26 16:35:57 zeroing out "/dev/nvme0n1p6"
172.16.56.129: user: warning: [2023-07-26T16:36:09.785623366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
172.16.56.129: user: warning: [2023-07-26T16:37:29.240040366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
172.16.56.129: user: warning: [2023-07-26T16:38:45.703981366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
172.16.56.129: user: warning: [2023-07-26T16:39:52.685989366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
172.16.56.129: user: warning: [2023-07-26T16:40:54.606820366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
172.16.56.129: user: warning: [2023-07-26T16:42:00.927927366Z]: [talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelModuleSpecController", "error": "error loading module \x5c"nvidia\x5c": module not found"}
[...]

PR #7538, opened 2023-07-27
fix: use fast wipe for upgrade

frezbo:
As part of bootloader refactoring `go-blockdevice` was used for wiping partitions in #7329, but used standard wipe which could be fast/slow depending on the blockdevice support. Switch to using fast-wipe for partitions. This should not affect `wipe` option in machineconfig.

Fixes: #7531

### DIFF
--- a/internal/pkg/partition/format.go
+++ b/internal/pkg/partition/format.go
@@ -57,9 +57,7 @@ func zeroPartition(devname string) (err error) {
 
 	defer part.Close() //nolint:errcheck
 
-	_, err = part.Wipe()
-
-	return err
+	return part.FastWipe()
 }
 
 func systemPartitionsFormatOptions(label string) *FormatOptions {


internal/pkg/partition/format.go
```diff
@@ -57,9 +57,7 @@ func zeroPartition(devname string) (err error) {
 
 	defer part.Close() //nolint:errcheck
 
-	_, err = part.Wipe()
-
-	return err
+	return part.FastWipe()
```
smira, line 60: Should we just use it if Size == 0?

smira: I mean how it was before we introduced the changes, or is Size now gone?

smira (approved):

internal/pkg/partition/format.go
frezbo: Size got moved to partition options, and we've only format options here now, I'm not sure if this is the same path used when wiping a disk with `wipe: true`

PR #7545, opened 2023-07-28
refactor: update AWS SDK and http-getter to v2 versions

smira:
Both are much modular and pull in much less dependendencies in to the Talos tree.

This solves the problem with allocations in AWS endpoints on import, and removes a bunch of dependencies.

Raw binary size: -10 MiB.

Memory usage (not scientific): around -5 MiB for all Talos services.


internal/app/machined/pkg/runtime/v1alpha1/platform/aws/metadata.go
```diff
@@ -7,9 +7,10 @@ package aws
 import (
 	"context"
 	"fmt"
-	"net/http"
+	"io"
 
-	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
+	"github.com/aws/smithy-go"
```
frezbo, line 13: at first thought this was something by Spencer :joy:

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/aws/metadata.go
smira: lol, that was my initial thought as well... why smithy? and it turns out to be an HTTP client on steroids

smira:
AWS metadata still works:


```yaml
node: 44.202.207.163
metadata:
    namespace: runtime
    type: PlatformMetadatas.talos.dev
    id: platformmetadata
    version: 1
    owner: network.PlatformConfigController
    phase: running
    created: 2023-07-28T12:09:12Z
    updated: 2023-07-28T12:09:12Z
spec:
    platform: aws
    hostname: ip-172-31-11-94.ec2.internal
    region: us-east-1
    zone: us-east-1a
    instanceType: t3.large
    instanceId: i-059f986c3bce6a3dd
    providerId: aws://us-east-1a/i-059f986c3bce6a3dd
```

Issue #7514, opened 2023-07-24
implement a resource to describe security state of Talos


smira:
Some resource to provide information about the "security" state of Talos:

* SecureBoot: true/false
* Fingerprint of SecureBoot key (can we?)
* Fingerprint of PCR signing key

PR #7550, opened 2023-07-28
feat: add security state resource

frezbo:
Add security state resource that describes the state of Talos SecureBoot and PCR signing key fingerprints.
    
The UKI fingerprint is currently not populated.

Fixes: #7514


internal/app/machined/pkg/controllers/runtime/security_state.go
```diff
@@ -0,0 +1,152 @@
...
+	_, data, err := attributes.ReadEfivarsWithGuid(sdboot.LoaderEntryDefaultName, *util.StringToGUID(sdboot.SystemdBootGUIDString))
+	if err != nil {
+		logger.Info("Failed to read systemd-boot loader entry", zap.Error(err))
+	}
+
+	loaderDefaultEntryName := data.String()
+
+	// lets mount the EFI partition and read the loader entry
+	if opErr := mount.PartitionOp(ctx, "", constants.EFIPartitionLabel, func() error {
+		if _, statErr := os.Stat(filepath.Join(constants.EFIMountPoint, "EFI", "Linux", loaderDefaultEntryName)); statErr == nil {
```
smira, line 78: I feel this is not really reliable, given the boot from PXE/ISO. we need some way to get that cert without going from the bootloader efi vars

frezbo: an easy fix is to put it into rootfs at a known path and we read from it (we could read the efi db variable, but that could contain not just the UKI signing keys, I couldn't think of a way to distinguish between them) :sweat_smile:

smira: if we can't do that right now, let's skip it

smira: what I'm trying to say is probably from an attestation point of view, I would rather see the keys which are authorized for secure boot, and the key which was used this time (if possible), but not the key the UKI is signed with... as we can be mislead to read the wrong UKI

frezbo: > and the key which was used this time (if possible)

not sure i follow

> but not the key the UKI is signed with... as we can be mislead to read the wrong UKI
there's only a single UKI booted and that the boot entry so this was just parsing the cert from the PE signature

I think i can add another field that lists all keys from UEFI db though (authorized keys for secureboot)

smira: I mean there was a secure boot process which verified the UKI, sd-boot, etc. there were some keys which were used to verify the signatures, which key it were?

> there's only a single UKI booted and that the boot entry so this was just parsing the cert from the PE signature

I don't think this holds true for ISO/PXE boot. For PXE boot, there's no UKI to read at all. 

So as long as we can't provide a reliable solution, we can skip it for now (and add it later)

frezbo: > I mean there was a secure boot process which verified the UKI, sd-boot, etc. there were some keys which were used to verify the signatures, which key it were?

that's the key used to sign the UKI, the public part is loaded into UEFI db, and firmware just verfies it's signed by the key in db

frezbo: > I don't think this holds true for ISO/PXE boot. For PXE boot, there's no UKI to read at all.

correct, but once talos is installed the data would get populated

smira: that's what I'm after - if we don't have a solution right now, let's keep it simple and skip this key. SecureBoot status is all we need right now

smira (approved):

PR #7557, opened 2023-07-31
chore: remove `cpu_manager_state` on `cpuManagerPolicy` change.

### LINKED ISSUE #7504: Activation of `cpuManagerPolicy=static` makes kubelet down
### Description

```yaml
machine:
  kubelet:
    extraConfig:
      cpuManagerPolicy: "static"
```

switching cpuManagerPolicy for kubelet requires to remove `/var/lib/kubelet/cpu_manager_state` file:

### Logs

```
192.168.88.209: {"ts":1690001513203.121,"caller":"kubelet/kubelet.go:1480","msg":"Failed to start ContainerManager","err":"start cpu manager error: could not restore state from checkpoint: configured policy \"static\" differs from state checkpoint policy \"none\", please drain this node and delete the CPU manager checkpoint file \"/var/lib/kubelet/cpu_manager_state\" before restarting Kubelet"}
```

### Environment

- Talos version: `v1.4.6`
- Kubernetes version: `v1.27.4`
- Platform: bare metal

### Workaround

Run before applying config 
```
kubectl debug node/talos-xjf-f8n -ti --image=alpine -n kube-system -- rm /host/var/lib/kubelet/cpu_manager_state
```

DmitriyMV:
After we closed `kubelet`, remove `/var/lib/kubelet/cpu_manager_state` if there are any changes in `cpuManagerPolicy`. We do not add any other safeguards, so it's user responsibility to cordon/drain the node in advance.

Also minor fixes in other files.

Closes #7504

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/base64"
+	stdjson "encoding/json"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -19,7 +20,10 @@ import (
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/channel"
+	"github.com/siderolabs/gen/optional"
 	"github.com/siderolabs/go-pointer"
 	"go.uber.org/zap"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -89,10 +93,8 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 	}
 
 	for {
-		select {
-		case <-ctx.Done():
-			return nil
-		case <-r.EventCh():
+		if _, ok := channel.RecvWithContext(ctx, r.EventCh()); !ok && ctx.Err() != nil {
+			return nil //nolint:nilerr
 		}
 
 		_, err := r.Get(ctx, resource.NewMetadata(files.NamespaceName, files.EtcFileStatusType, "machine-id", resource.VersionUndefined))
@@ -141,13 +143,11 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 	r.QueueReconcile()
 
 	for {
-		select {
-		case <-ctx.Done():
-			return nil
-		case <-r.EventCh():
+		if _, ok := channel.RecvWithContext(ctx, r.EventCh()); !ok && ctx.Err() != nil {
+			return nil //nolint:nilerr
 		}
 
-		cfg, err := r.Get(ctx, resource.NewMetadata(k8s.NamespaceName, k8s.KubeletSpecType, k8s.KubeletID, resource.VersionUndefined))
+		cfg, err := safe.ReaderGetByID[*k8s.KubeletSpec](ctx, r, k8s.KubeletID)
 		if err != nil {
 			if state.IsNotFoundError(err) {
 				continue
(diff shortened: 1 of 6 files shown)


DmitriyMV:
As for file itself, I don't really know what can we do with it aside from deleting it. It basically contains something like this
```json
{"policyName":"static","defaultCpuSet":"0-1","checksum":1636926438}
```

```json
{"policyName":"none","defaultCpuSet":"","checksum":1353318690}
```

```json
{"policyName":"static","defaultCpuSet":"0,2-3,6-7","entries":{"dae8d8fd-2321-413f-8775-201307c32252":{"myservice":"1,5","nginx":"4"}},"checksum":1986161931}
```

I still don't like the fact the we do not warn or prevent the user from changing `machineconfig` here before cordoning/draining node.

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -196,6 +197,22 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 			}
 		}
 
+		policy, err := getFromMap[string](cfgSpec.Config, "cpuManagerPolicy")
+		if err != nil {
+			return err
+		}
+
+		newPolicy := policy.ValueOrZero()
```
smira, line 205: this is a wrong way to detect the change, as it only works if the controller was up when the configuration change happened, while the controller might be down on config change (e.g. applying machine config with a reboot).

we should rather detect the policy name in the `state` file doesn't match the policy in the config.

DmitriyMV: Done here: https://github.com/siderolabs/talos/pull/7557/files#diff-fa70ddd6c68e2179d952fc464512dd777494c4fe1d53cce9940c2c5c20ed9183R234-R254

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -196,6 +196,25 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
...
+		if !equalPolicy(oldPolicy, newPolicy) {
+			logger.Info("cpuManagerPolicy changed", zap.String("old", oldPolicy), zap.String("new", newPolicy))
+
+			if removeErr := os.Remove("/var/lib/kubelet/cpu_manager_state"); removeErr != nil {
```
smira, line 213: this seems to be wrong, as the state file might not exist, while there's a policy change - e.g. starting up with a non-default policy

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -196,6 +196,25 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 			}
 		}
 
+		oldPolicy, err := loadPolicyFromFile("/var/lib/kubelet/cpu_manager_state")
```
smira, line 199: I think this new code should be moved out to function which manages policy change like other functions here

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
DmitriyMV: Yeah, I agree - if there is no file, there is nothing we can do.

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
DmitriyMV: Done

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
```diff
@@ -212,6 +216,92 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
...
+	}
+
+	newPolicy := policy.ValueOrZero()
+	if equalPolicy(oldPolicy, newPolicy) {
+		return nil
+	}
+
+	logger.Info("cpuManagerPolicy changed", zap.String("old", oldPolicy), zap.String("new", newPolicy))
+
+	_, err = os.Stat(managerFilename)
```
smira, line 244: do we need to stat once again, given that we return on line 226?

smira (approved):

internal/app/machined/pkg/controllers/k8s/kubelet_service.go
DmitriyMV: No, we do not, it's a leftover. Thanks for noticing!

PR #7568, opened 2023-08-03
fix: update providerid prefix for aws

rsmitty:
This PR updates the ProviderID format for aws resources. There seems to be a bug when using Talos CCM (which consumes this value from Talos) because the format is `aws://x/y` (two slashes) vs. the expected `aws:///x/y` (three slashes) that is set with the AWS CCM code [here](https://github.com/kubernetes/cloud-provider-aws/blob/d0551093673e8c355db17249b8f069767c014748/pkg/providers/v1/instances.go#L47-L53).

Setting only two slashes causes important software in the workload cluster to fail, specifically cluster-autoscaler. The regex they use for pulling providerID is [here](https://github.com/kubernetes/autoscaler/blob/702e9685d6c1d002f4a448f2d88007141dfa6d56/cluster-autoscaler/cloudprovider/aws/aws_cloud_provider.go#L195).


rsmitty:
@sergelogvinov please double check that this doesn't cause any unexpected issues for you with the Talos CCM

sergelogvinov:
Just check EKS.
```
providerID: aws:///us-east-1f/i-
```

3 slashes.

Thanks.

PR #7569, opened 2023-08-03
chore: unify semver under `github.com/blang/semver/v4`

DmitriyMV:
Currently, we use `github.com/coreos/go-semver/semver` and `github.com/hashicorp/go-version` for version parsing. As we use `github.com/blang/semver/v4` in our other projects, and it has more features, it makes sense to use it across the projects. It also doesn't allocate like crazy in `KubernetesVersion.SupportedWith`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,6 @@ require (
 	github.com/containernetworking/cni v1.1.2
 	github.com/containernetworking/plugins v1.3.0
 	github.com/coreos/go-iptables v0.7.0
-	github.com/coreos/go-semver v0.3.1
 	github.com/cosi-project/runtime v0.3.1
 	github.com/docker/distribution v2.8.2+incompatible
 	github.com/docker/docker v24.0.5+incompatible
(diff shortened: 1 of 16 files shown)


go.mod
```diff
@@ -315,3 +314,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.14.3-0.20230601165947-6ce0bf390ce3 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+require github.com/coreos/go-semver v0.3.1 // indirect
```
frezbo, line 318: i really hate `go mod tidy` doing this, new deps seems to be added to a new `require` block, not sure why

DmitriyMV: I think it's because we have more than two `require` blocks.

DmitriyMV: Fixed!

go.mod
```diff
@@ -66,7 +65,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.2
 	github.com/hashicorp/go-getter/v2 v2.2.1
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/hashicorp/go-version v1.6.0
```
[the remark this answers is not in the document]
DmitriyMV: Done!

PR #7582, opened 2023-08-04
chore: clean up the output of the `imager`

smira:
Use `Progress`, and options to pass around the way messages are written.

Fixed some tiny issues in the code, but otherwise no functional changes.

To make colored output work with `docker run`, switched back image generation to use volume mount for output (old mode is still functioning, but it's not the default, and it works when docker is not running on the same host).


smira:
[asciicast](https://asciinema.org/a/Tpo3tB1EzLSAODHb2t99X3tJj)

frezbo (approved):

PR #7585, opened 2023-08-07
chore: remove <-errCh where possible in grpc methods

DmitriyMV:
Simplify code by passing error directly into the pipe closer.

### DIFF
--- a/pkg/machinery/client/client.go
+++ b/pkg/machinery/client/client.go
@@ -178,10 +178,13 @@ func (c *Client) Close() error {
 }
 
 // KubeconfigRaw returns K8s client config (kubeconfig).
-func (c *Client) KubeconfigRaw(ctx context.Context) (io.ReadCloser, <-chan error, error) {
+//
+// This method doesn't support multiplexing of the result:
+// * either client.WithNodes is not used, or it contains a single node in the list.
+func (c *Client) KubeconfigRaw(ctx context.Context) (io.ReadCloser, error) {
 	stream, err := c.MachineClient.Kubeconfig(ctx, &emptypb.Empty{})
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	return ReadStream(stream)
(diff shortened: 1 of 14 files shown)


pkg/machinery/client/client.go
```diff
@@ -1007,16 +997,22 @@ func ReadStream(stream MachineStream) (io.ReadCloser, <-chan error, error) {
 			}
 
 			if data.Metadata != nil && data.Metadata.Error != "" {
+				var err error
+
 				if data.Metadata.Status != nil {
-					errCh <- status.FromProto(data.Metadata.Status).Err()
+					err = status.FromProto(data.Metadata.Status).Err()
```
smira, line 1003: I feel we might want to ignore here same errors as above, basically making it `CloseWithError(nil)` is it's Canceled/DeadlineExceeded

smira (approved):

pkg/machinery/client/client.go
DmitriyMV: Done.

pkg/machinery/client/client.go
```diff
@@ -532,12 +524,12 @@ func (c *Client) DiskUsage(ctx context.Context, req *machineapi.DiskUsageRequest
 }
 
 // Copy implements the proto.MachineServiceClient interface.
```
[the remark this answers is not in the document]
DmitriyMV: 🆗

DmitriyMV:
One cool thing from this, is new error output. For example when you try to get `kubeconfig` from worker node:

Was:
```
$ talosctl -n 172.20.0.5 kubeconfig
rpc error: code = Unimplemented desc = kubeconfig is only available on control plane nodes
error initializing gzip: EOF
```

Now:
$ talosctl -n 172.20.0.5 kubeconfig
error initializing gzip: rpc error: code = Unimplemented desc = kubeconfig is only available on control plane nodes

PR #7590, opened 2023-08-09
refactor: docgen and config examples

smira:
Short version is: move from global variables/`init()` function into explicit functions.

`docgen` was updated to skip creating any top-level global variables, now `Doc` information is generated on the fly when it is accessed. Talos itself doesn't marshal the configuration often, so in general it should never be accessed for Talos (but will be accessed e.g. for `talosctl`).

Machine config examples were changed manually from variables to functions returning a value and moved to a separate file.

There are no changes to the output of `talosctl gen config`.

There is a small change to the generated documentation, which I believe is a correct one, as previously due to value reuse it was clobbered with other data.

See #7578

### DIFF
--- a/hack/docgen/main.go
+++ b/hack/docgen/main.go
@@ -35,60 +35,68 @@ import (
 )
 
 {{ $tick := "` + "`" + `" -}}
-var (
-	{{ range $struct := .Structs -}}
-	{{ $struct.Name }}Doc encoder.Doc
-	{{ end -}}
-)
-
-func init() {
-	{{ range $struct := .Structs -}}
-	{{ $docVar := printf "%v%v" $struct.Name "Doc" }}
-	{{ $docVar }}.Type = "{{ $struct.Name }}"
-	{{ $docVar }}.Comments[encoder.LineComment] = "{{ $struct.Text.Comment }}"
-	{{ $docVar }}.Description = "{{ $struct.Text.Description }}"
-	{{ range $example := $struct.Text.Examples }}
-	{{ if $example.Value }}
-	{{ $docVar }}.AddExample("{{ $example.Name }}", {{ $example.Value }})
-	{{ end -}}
-	{{ end -}}
-	{{ if $struct.AppearsIn -}}
-	{{ $docVar }}.AppearsIn = []encoder.Appearance{
-	{{ range $value := $struct.AppearsIn -}}
-		{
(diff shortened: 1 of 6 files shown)


frezbo (approved):

hack/docgen/main.go
```diff
@@ -35,60 +35,68 @@ import (
...
-	{{ end -}}
-	{{ end -}}
-	{{ if $struct.AppearsIn -}}
-	{{ $docVar }}.AppearsIn = []encoder.Appearance{
-	{{ range $value := $struct.AppearsIn -}}
-		{
-			TypeName: "{{ $value.Struct.Name }}",
-			FieldName: "{{ $value.FieldName }}",
+{{ range $struct := .Structs -}}
+func (_ {{ $struct.Name }}) Doc() *encoder.Doc {
```
DmitriyMV, line 40: We don't need underscore here. 

```go
func (Config) Doc() *encoder.Doc {
```

works just as fine as

```go
func (_ Config) Doc() *encoder.Doc {
```

smira: sure, I will update

DmitriyMV (approved):

internal/pkg/tui/installer/state.go
```diff
@@ -268,27 +268,27 @@ func configureAdapter(installer *Installer, opts *machineapi.GenerateConfigurati
...
 					components.NewItem(
 						"MTU",
-						v1alpha1.DeviceDoc.Describe("mtu", true),
+						v1alpha1.Device{}.Doc().Describe("mtu", true),
 						&adapterSettings.Mtu,
 					),
 					components.NewItem(
 						"Route Metric",
-						v1alpha1.DeviceDoc.Describe("dhcpOptions", true),
+						v1alpha1.Device{}.Doc().Describe("dhcpOptions", true),
```
Unix4ever, line 291: Not sure if it helps, but maybe something like:

```go
v1alpha1.Describe[v1alpha1.Device]("dhcpOptions", true)

// or if we define it in the same module
describe[v1alpha1.Device]("dhcpOptions", true)
```
will be more readable here

Unix4ever (approved):

internal/pkg/tui/installer/state.go
smira: yep, good idea, I will update

smira: updated

PR #7565, opened 2023-08-03
docs: update cilium instructions

frezbo:
Update cilium instructions to skip mounting `bpffs`.

Also fix the TPM example in release notes.

PR #7593, opened 2023-08-09
docs: mount bpffs to the cilium container

nberlee:
# Pull Request

## What? (description)
This commit adds a hostPath to ensure cilium functions correctly after a container restart.

## Why? (reasoning)
In commit 76fa45a, bpf automount was turned off because it's mounted on the host. This led to the removal of the init container's mounting. Helm doesn't automatically include the hostPath mount. Additionally, neither /sys nor /sys/fs were added to the container. As a result, cilium couldn't determine if /sys/fs/bpf was already mounted and mounts it anyway (only in the container).

More details: https://github.com/siderolabs/talos/pull/7565#issuecomment-1671063014

## Acceptance

### DIFF
--- a/website/content/v1.5/kubernetes-guides/network/deploying-cilium.md
+++ b/website/content/v1.5/kubernetes-guides/network/deploying-cilium.md
@@ -79,6 +79,9 @@ cilium install \
     --helm-set=securityContext.capabilities.ciliumAgent="{CHOWN,KILL,NET_ADMIN,NET_RAW,IPC_LOCK,SYS_ADMIN,SYS_RESOURCE,DAC_OVERRIDE,FOWNER,SETGID,SETUID}" \
     --helm-set=securityContext.capabilities.cleanCiliumState="{NET_ADMIN,SYS_ADMIN,SYS_RESOURCE}" \
     --helm-set=cgroup.autoMount.enabled=false \
+    --helm-set=extraHostPathMounts[0].name=cilium-bpf \
+    --helm-set=extraHostPathMounts[0].mountPath=/sys/fs/bpf \
+    --helm-set=extraHostPathMounts[0].hostPath=/sys/fs/bpf \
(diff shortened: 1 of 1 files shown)


frezbo:
Seems I disabled the hostpaths here: https://github.com/cilium/cilium/pull/24161/files, maybe fix there, and for now we can fix the docs to not disable the initcontainer

frezbo:
> Seems I disabled the hostpaths here: https://github.com/cilium/cilium/pull/24161/files, maybe fix there, and for now we can fix the docs to not disable the initcontainer

hmm I forgot to update ci to test with this, I'll get it fixed tomorrow

nberlee:
> Seems I disabled the hostpaths here: https://github.com/cilium/cilium/pull/24161/files, maybe fix there, and for now we can fix the docs to not disable the initcontainer

I am not really sure its a cilium helm bug. I think it works as intended, and is inline with the description in helm about `bpf.autoMount`. This function was initially there for crio as crio has /sys/fs/bpf mounted by default.

nberlee:
> now we can fix the docs to not disable the initcontainer
Yes i am now convinced this is the best solution as /sys/fs/bpf is also not mounted to `clean-cilium-state` init container even with this PR

frezbo:
> > now we can fix the docs to not disable the initcontainer
> 
> Yes i am now convinced this is the best solution as /sys/fs/bpf is also not mounted to `clean-cilium-state` init container even with this PR

I think need to do another PR (but I remember testing it, maybe not with a reboot)

frezbo:
> > now we can fix the docs to not disable the initcontainer
> 
> Yes i am now convinced this is the best solution as /sys/fs/bpf is also not mounted to `clean-cilium-state` init container even with this PR

The docs still needs fix, right?

nberlee:
I thought i close it until I made your suggested changes and then reopen... But that does not seem possible anymore

nberlee:
Now it is ? The button was grayed out

PR #7605, opened 2023-08-11
feat: update Kubernetes to 1.28.0-rc.1

smira:
See https://github.com/kubernetes/kubernetes/releases/tag/v1.28.0-rc.1


hack/test/e2e.sh
```diff
@@ -325,7 +325,7 @@ function validate_rlimit_nofile {
 
 function validate_booted_secureboot {
   ${TALOSCTL} dmesg | grep "Secure boot enabled"
-  ${TALOSCTL} get securitystate -o json 
+  ${TALOSCTL} get securitystate -o json
```
frezbo, line 328: we can remove this now :sweat_smile:

frezbo (approved):

hack/test/e2e.sh
smira: I overall feel we should move this to the integration tests... I think we should cut our shell scripts

frezbo: ahh yes, and fix how we detect secureboot in integration test. I'll do in a separate PR

PR #7607, opened 2023-08-11
chore: update integration to use securitystate

frezbo:
Update integration test suite to use security state resource.

### DIFF
--- a/internal/integration/base/api.go
+++ b/internal/integration/base/api.go
@@ -17,12 +17,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/siderolabs/go-retry/retry"
 	"github.com/stretchr/testify/suite"
 	"google.golang.org/grpc/backoff"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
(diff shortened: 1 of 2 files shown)


internal/integration/base/api.go
```diff
@@ -167,11 +169,15 @@ func (apiSuite *APISuite) Capabilities() Capabilities {
 		}
 	}
 
-	if _, err = apiSuite.Client.LS(context.Background(), &machineapi.ListRequest{
-		Root:  sdboot.SystemdBootStubInfoPath,
-		Types: []machineapi.ListRequest_Type{machineapi.ListRequest_REGULAR},
-	}); err == nil {
-		caps.TrustedBoot = true
+	rList, err := apiSuite.Client.COSI.List(context.Background(), resource.NewMetadata(v1alpha1.NamespaceName, runtimeresources.SecurityStateType, "", resource.VersionUndefined))
```
smira, line 172: I wonder if we should also have the input to the test (test flag) which says it's expected to be secureboot

so that we can move the check from the shell script to the integration test - if expected to secureboot (or vice versa), the resource should appear

frezbo: yes, i see that makes things easier

PR #7610, opened 2023-08-13
chore: move bash tests to integration

frezbo:
move extensions and secureboot tests to integration. Makes it easier to test.


hack/test/extensions/extension-patch-filter.jq
```diff
@@ -4,7 +4,7 @@
         "path": "/machine/install/extensions",
         "value": [
             {
-                "image": map(select(. | contains("nvidia-container-toolkit") or contains("nvidia-fabricmanager") | not)) | .[]
+                "image": map(select(. | contains("nvidia") or contains("tailscale") | not)) | .[]
```
frezbo, line 7: tailscale keeps on restarting without auth and nvida will be tested with another e2e

Makefile
```diff
@@ -16,7 +16,7 @@ CLOUD_IMAGES_EXTRA_ARGS ?= ""
 
 ARTIFACTS := _out
 TOOLS ?= ghcr.io/siderolabs/tools:v1.5.0
-PKGS ?= v1.5.0
+PKGS ?= v1.5.0-1-g8a2227d
```
smira, line 19: `main` should be using `1.6.0-alpha.0` tags...

internal/integration/api/common.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for _, expectedModule := range expectedVirtIOModules {
+		suite.Require().Contains(loadedModules, expectedModule, "expected module %s to be loaded", expectedModule)
+	}
+}
+
+// TestCommonDefaults verifies that the default ulimits are set.
+func (suite *CommonSuite) TestCommonDefaults() {
+	expectedUlimit := `
+core file size (blocks)         (-c) 0
```
smira, line 98: are these values stable with different RAM config?

internal/integration/api/reset.go
```diff
@@ -218,9 +218,9 @@ func (suite *ResetSuite) TestResetWithSpecState() {
 		suite.T().Skip("cluster doesn't support reboot (and reset)")
 	}
 
-	if suite.Capabilities().TrustedBoot {
-		// this is because with trusted boot the talos.config is only applied and cannot be passed as kernel args
-		suite.T().Skip("cluster has trusted boot enabled, doesn't support reset")
+	if suite.TrustedBoot {
```
smira, line 221: I think the previous made more sense - if we detect secure boot, we skip the tests, vs. that we were asked to verify that trusted boot is enabled.

internal/integration/api/trustedboot.go
```diff
@@ -0,0 +1,99 @@
...
+func (suite *TrustedBootSuite) TearDownTest() {
+	if suite.ctxCancel != nil {
+		suite.ctxCancel()
+	}
+}
+
+// TestTrustedBootState verifies that the system is booted in secure boot mode
+// and that the disks are encrypted.
+func (suite *TrustedBootSuite) TestTrustedBootState() {
+	if testing.Short() {
```
smira, line 54: this should be a short test, so we could run it always?

internal/integration/api/trustedboot.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if !suite.TrustedBoot {
+		suite.T().Skip("skipping as talos is not explicitly trusted booted")
+	}
+
+	node := suite.RandomDiscoveredNodeInternalIP()
+	ctx := client.WithNode(suite.ctx, node)
+
+	securityResource, err := safe.StateGetByID[*runtimeres.SecurityState](ctx, suite.Client.COSI, runtimeres.SecurityStateID)
```
smira, line 65: in all these cases, better to use `safe.StateWatchFor` with types `Created, Updated`, as this doesn't depend on the resource being there at the moment of the test run

internal/integration/api/common.go
frezbo: yes, works fine locally and in ci

internal/integration/api/trustedboot.go
frezbo: hmm docker also uses short, so we'd then have to also check the provisioner used

internal/integration/api/reset.go
frezbo: didn't get that

smira: in one of the previous iterations of this PR, you had `Capabilities().TrustedBoot` driven by the `SecurityState` resource, I remember reviewing it.

what I mean that this test should be skipped if the cluster is running in trusted boot mode, no matter if the flag `-test.trustedboot` was specified or not.

internal/integration/api/reset.go
```diff
@@ -114,17 +114,9 @@ func (suite *ResetSuite) TestResetNodeByNode() {
...
-	}
-
-	if suite.Capabilities().TrustedBoot {
-		// this is because with trusted boot the talos.config is only applied and cannot be passed as kernel args
-		suite.T().Skip("cluster has trusted boot enabled, doesn't support reset")
-	}
-
-	if suite.Cluster == nil {
-		suite.T().Skip("without full cluster state reset test is not reliable (can't wait for cluster readiness in between resets)")
+	if !suite.Capabilities().SecureBooted {
```
smira, line 117: the condition is inverted here and other places

frezbo: ahh :face_with_head_bandage: , I'm dumb

PR #7628, opened 2023-08-17
fix: put correct capacity to the ovf

### LINKED ISSUE #7627: Talos v1.5.0 failing to deploy from ova template
### Description
Hello! When I try to deploy ova, the template wants to allocate 8192.0 TB on the datastore and then fails.

### Logs
`The operation failed due to File [nfs_default_1] talos-vmware-1.5.0/talos-vmware-1.5.0.vmdk is larger than the maximum size supported by datastore 'nfs_default_1' 
`
### Environment

- Talos version: 1.5.0
- Platform: VMware

smira:
With the refactoring to 1.5, the confusion between bytes and megabytes introduced this bug.

Fixes #7627

### DIFF
--- a/pkg/imager/ova/ova.go
+++ b/pkg/imager/ova/ova.go
@@ -161,9 +161,9 @@ func CreateOVAFromRAW(name, path, arch, scratchPath string, diskSize int64, prin
 		return err
 	}
 
-	size := f.Size()
+	imageSize := f.Size()
 
-	ovf, err := renderOVF(name, size, diskSize)
+	ovf, err := renderOVF(name, imageSize, diskSize)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


TimJones (approved):

pkg/imager/ova/ova.go
```diff
@@ -249,7 +249,7 @@ func renderOVF(name string, size, capacity int64) (string, error) {
 	}{
 		VMDK:     name + ".vmdk",
 		Size:     size,
-		Capacity: capacity,
+		Capacity: capacity / (1 << 20),
```
frezbo, line 252: should we set this in `capacity` variable then, or is only applicable to OVA,? I guess `Capacity` field expects a MB value?

frezbo (approved):

pkg/imager/ova/ova.go
smira: it's just the way we build the XML template... we say in the XML that it's in MiB. I could change the template, but I'm less sure it will work :-D

smira: I will probably rename the argument to `diskSize` to match other usecases

frezbo: yeh, just a comment saying it expects MiB is fine for future reference

smira: renamed to make things less confusing

frezbo (approved):

PR #7635, opened 2023-08-18
feat: add e2e-aws for nvidia extensions

frezbo:
Add e2e tests for nvidia

### DIFF
--- a/internal/integration/api/extensions_nvidia.go
+++ b/internal/integration/api/extensions_nvidia.go
@@ -1,0 +1,307 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_api
+
+package api
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/siderolabs/go-pointer"
+	"github.com/siderolabs/go-retry/retry"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	nodev1 "k8s.io/api/node/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/siderolabs/talos/internal/integration/base"
+)
+
+// ExtensionsSuiteNVIDIA verifies Talos is securebooted.
+type ExtensionsSuiteNVIDIA struct {
+	base.K8sSuite
+
+	ctx       context.Context //nolint:containedctx
+	ctxCancel context.CancelFunc
+}
+
+// SuiteName ...
+func (suite *ExtensionsSuiteNVIDIA) SuiteName() string {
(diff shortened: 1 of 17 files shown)


internal/integration/api/extensions_nvidia.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	nodes := suite.getNVIDIANodes("node.kubernetes.io/instance-type=g4dn.xlarge")
+	for _, node := range nodes {
+		suite.AssertServicesRunning(suite.ctx, node, map[string]string{
+			"ext-nvidia-persistenced": "Running",
+		})
+	}
+
+	// nodes = suite.getNVIDIANodes("node.kubernetes.io/instance-type=p4d.24xlarge")
```
smira, line 80: is it commented out for some reason?

internal/integration/base/api.go
```diff
@@ -480,6 +482,63 @@ func (apiSuite *APISuite) UserDisks(ctx context.Context, node string, sizeGreate
...
+	}
+}
+
+// AssertExpectedModules verifies that expected kernel modules are loaded on the node.
+func (apiSuite *APISuite) AssertExpectedModules(ctx context.Context, node string, expectedModules map[string]string) {
+	nodeCtx := client.WithNode(ctx, node)
+
+	fileReader, err := apiSuite.Client.Read(nodeCtx, "/proc/modules")
+	defer func() {
+		err = fileReader.Close()
```
smira, line 506:
```suggestion
		apiSuite.Require().NoError(fileReader.Close())
```

and it should be below line 509, as we shouldn't close if the `Read` call fails

internal/integration/base/api.go
```diff
@@ -480,6 +482,63 @@ func (apiSuite *APISuite) UserDisks(ctx context.Context, node string, sizeGreate
...
+	var loadedModules []string
+
+	for scanner.Scan() {
+		loadedModules = append(loadedModules, strings.Split(scanner.Text(), " ")[0])
+	}
+	apiSuite.Require().NoError(scanner.Err())
+
+	fileReader, err = apiSuite.Client.Read(nodeCtx, fmt.Sprintf("/lib/modules/%s/modules.dep", constants.DefaultKernelVersion))
+	defer func() {
+		err = fileReader.Close()
```
smira, line 522: same as above

internal/pkg/extensions/kernel_modules.go
```diff
@@ -132,7 +132,7 @@ func GenerateKernelModuleDependencyTreeExtension(extensionsPathWithKernelModules
 		return mount.Unmount(overlays)
 	})
 
-	log.Println("running depmod to generate kernel modules dependency tree")
+	printFunc("running depmod to generate kernel modules dependency tree")
```
smira, line 135: I was going to fix this, thanks for fixing it!

internal/integration/api/extensions_nvidia.go
frezbo: yes, the a100 gpu's are so costly compared to aws, so want to do that only in azure 30$/hr vs 3$/hr

smira (approved):

PR #7642, opened 2023-08-21
fix: use image digest when starting a container

### LINKED ISSUE #7640: Talos 1.5.0 doesn't support fully specified docker images for kubelet or etcd
### Description

After upgrading to Talos 1.5.0 it seems like you can't use a fully qualified image (i.e. with repository, tag and digest) for kubelet or etcd images any more.

If I drop the tag or digest, then it does work.

### Logs

Trying to use a fully qualified image gives me e.g.:

```
10.24.128.4: 2023/08/21 00:27:00.640656 [talos] service[etcd](Failed): Failed running service: error opening runner: image "gcr.io/etcd-development/etcd:v3.5.9@sha256:8c956d9b0d39745fa574bb4dbacd362ffdc1109479432f54094859d4cf984b17": not found
10.24.128.4: 2023/08/21 00:27:10.587930 [talos] service[kubelet](Failed): Failed running service: error opening runner: image "ghcr.io/siderolabs/kubelet:v1.28.0@sha256:50710f2cd3328c23f57dfc7fb00940d8cfd402315e33fc7cb8184fc660650a5c": not found
```

### Environment

- Talos version: 1.5.0
- Kubernetes version: 1.28.0
- Platform: aws

smira:
First of all, it seems to be "right way", as it makes sure the image is looked up by the digest.

Second, it fixes the case when image is specified with both tag and digest (which is not supposed to be the correct ref, but it is used frequently).

Talos since 1.5.0 stores images with the following aliases:

```
gcr.io/etcd-development/etcd:v3.5.9
gcr.io/etcd-development/etcd@sha256:8c956d9b0d39745fa574bb4dbacd362ffdc1109479432f54094859d4cf984b17
ghcr.io/siderolabs/kubelet:v1.28.0
ghcr.io/siderolabs/kubelet@sha256:50710f2cd3328c23f57dfc7fb00940d8cfd402315e33fc7cb8184fc660650a5c
sha256:50710f2cd3328c23f57dfc7fb00940d8cfd402315e33fc7cb8184fc660650a5c
sha256:8c956d9b0d39745fa574bb4dbacd362ffdc1109479432f54094859d4cf984b17
```

This change pulls the digest format (the last in this list) and uses it to start a container.

Fixes #7640

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -67,6 +67,8 @@ type Etcd struct {
 	args   []string
 	client *etcd.Client
 
+	imgRef string
+
 	// if the new member was added as a learner during the service start, its ID is kept here
 	learnerMemberID uint64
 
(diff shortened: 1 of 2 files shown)


frezbo:
> Second, it fixes the case when image is specified with both tag and digest (which is not supposed to be the correct ref, but it is used frequently).

I learned this recently and it makes sense, not to use a tag when specifying the digest

frezbo (approved):

smira:
> > Second, it fixes the case when image is specified with both tag and digest (which is not supposed to be the correct ref, but it is used frequently).
> 
> I learned this recently and it makes sense, not to use a tag when specifying the digest

It actually ignores the tag when a digest is specified

Unix4ever (approved):

PR #7647, opened 2023-08-21
fix: properly GC images supplied with both tag and digest

smira:
This is a follow-up fix for #7640

I noticed that image cleanup controller cleans up the images if specified with both tag and digest.

The problem was incorrectly building image references in the expected set of images, so they were incorrectly marked as unused.

Refactor the code to make the core part testable.

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
+++ b/internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
@@ -186,16 +186,7 @@ func (ctrl *CRIImageGCController) Run(ctx context.Context, r controller.Runtime,
 }
 
 //nolint:gocyclo
-func (ctrl *CRIImageGCController) cleanup(ctx context.Context, logger *zap.Logger, imageService images.Store, expectedImages []string) error {
-	logger.Debug("running image cleanup")
-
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/runtime/cri_image_gc.go
```diff
@@ -248,12 +236,36 @@ func (ctrl *CRIImageGCController) cleanup(ctx context.Context, logger *zap.Logge
 				if expectedDigested, ok := expectedRef.(docker.Digested); ok && ref.Digest() == expectedDigested.Digest() {
 					// this is expected image by digest, inject other forms of the ref
 					expectedImageNames[digest] = struct{}{}
-					expectedImageNames[expectedRef.String()] = struct{}{}
+					expectedImageNames[expectedRef.Name()+"@"+digest] = struct{}{}
```
frezbo, line 239: is there some upstream code in containerd/docker that handles this, rather than us doing this?

smira: we're already using the parser part, so it's not completely hand-written

the actual set of images is more the way CRI containerd manages images

PR #7660, opened 2023-08-23
docs: fix the installing system extensions doc

### LINKED ISSUE #7658: The Installing System Extensions documentation seems to be incomplete
The text at https://www.talos.dev/v1.5/talos-guides/configuration/system-extensions/#installing-system-extensions (which source is at https://github.com/siderolabs/talos/blob/v1.5.1/website/content/v1.5/talos-guides/configuration/system-extensions.md?plain=1#L14-L18) seems to be incomplete.

smira:
Fixes #7658


chrislwade:
Based on Slack conversations with @smira and @frezbo, I think adding a section to the documentation highlighting scenarios where only a custom installer is necessary would be useful.

smira:
> Based on Slack conversations with @smira and @frezbo, I think adding a section to the documentation highlighting scenarios where only a custom installer is necessary would be useful.

In this PR, an example with NVIDIA on bare-metal is exactly that, isn't it?

PR #7665, opened 2023-08-23
chore: update Go to 1.21

smira:
This fixes a problem in the `RouteSpecController` which is due to a subtle (but correct) change in the behavior in the `stdlib`.

Also some small (but should be safe) bumps.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/siderolabs/talos
 
-go 1.20
+go 1.21
```
frezbo, line 3: 1.21.0?

frezbo (approved):

go.mod
smira: I'm kinda torn on that, I don't see why we should do a minor version there? unless we depend on something to be exactly 1.21.0, but not 1.21.1 ? I understand this is a compatibility guarantee, or a constraint

internal/app/machined/pkg/controllers/network/route_spec.go
```diff
@@ -119,6 +120,17 @@ func (ctrl *RouteSpecController) Run(ctx context.Context, r controller.Runtime,
 	}
 }
 
+// netipPrefixBitsCorrected returns the number of bits in the prefix, corrected for zero value to have bits of 0.
+//
+// Go stdlib returns -1 for zero value, which is not what we want.
+func netipPrefixBitsCorrected(p netip.Prefix) int {
```
frezbo, line 126: can we fix our code to be same as expected spec, or am I missing something here?

smira: we use it to interface with Linux kernel, and there it expects DstLength to be 0 if the DstAddress is nil.

zero `netip.Prefix` gives us exactly `nil` address, but length (bits in the prefix) became wrong now, as Linux accepts `uint8`, and -1 becomes 255.

go.mod
smira: @DmitriyMV please help

smira: @DmitriyMV please help

PR #7671, opened 2023-08-24
chore: drop firmware from initramfs

frezbo:
Drop firmware from initramfs. Extra firmware can be added as system extensions enabled through imager service.

Before:

```bash
❯ du -sh _out/initramfs-amd64.xz
58M	_out/initramfs-amd64.xz
```

After:

```bash
❯ du -sh _out/initramfs-amd64.xz
56M	_out/initramfs-amd64.xz
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -680,12 +680,6 @@ FROM build AS initramfs-archive-arm64
 WORKDIR /initramfs
 COPY --from=squashfs-arm64 /rootfs.sqsh .
 COPY --from=init-build-arm64 /init .
-# copying over firmware binary blobs to initramfs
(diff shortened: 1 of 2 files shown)


frezbo:
> I think we need a note in `hack/release.toml` about it

Yeh, I was thinking once we have the imager service up and link to such docs.

smira:
we can do a note that it was dropped and we can add more stuff later, but so that we don't forget

frezbo:
> we can do a note that it was dropped and we can add more stuff later, but so that we don't forget

yup, added proper notes. We can link to imager service once it's available

Issue #7080, opened 2023-04-12
clock skew


uhthomas:
### Description

It looks like talos nodes aren't syncing their clocks properly. Ceph is complaining of clock skew and I can confirm this is the case on the hosts themselves. The difference between `2023-04-12 11:43:52.297503558 +0000 UTC` and `2023-04-12 11:43:52.430898682 +0000 UTC` is `133.395124ms`(!). ([playground](https://go.dev/play/p/Ba-CnUyot1r))

### Logs

```sh
❯ talosctl --talosconfig=unwind-talosconfig -n 10.0.0.101,10.0.0.102,10.0.0.103,10.0.0.104,10.0.0.105 time
NODE         NTP-SERVER     NODE-TIME                                 NTP-SERVER-TIME
10.0.0.101   pool.ntp.org   2023-04-12 11:43:52.324593578 +0000 UTC   2023-04-12 11:43:52.308599071 +0000 UTC
10.0.0.102   pool.ntp.org   2023-04-12 11:43:52.430898682 +0000 UTC   2023-04-12 11:43:52.303693959 +0000 UTC
10.0.0.103   pool.ntp.org   2023-04-12 11:43:52.317368811 +0000 UTC   2023-04-12 11:43:52.309190395 +0000 UTC
10.0.0.104   pool.ntp.org   2023-04-12 11:43:52.315811787 +0000 UTC   2023-04-12 11:43:52.307580784 +0000 UTC
10.0.0.105   pool.ntp.org   2023-04-12 11:43:52.297503558 +0000 UTC   2023-04-12 11:43:52.31071359 +0000 UTC
❯ talosctl --talosconfig=unwind-talosconfig -n 10.0.0.101,10.0.0.102,10.0.0.103,10.0.0.104,10.0.0.105 time -c time.cloudflare.com
NODE         NTP-SERVER            NODE-TIME                                 NTP-SERVER-TIME
10.0.0.101   time.cloudflare.com   2023-04-12 11:45:32.258062121 +0000 UTC   2023-04-12 11:45:32.249952965 +0000 UTC
10.0.0.102   time.cloudflare.com   2023-04-12 11:45:32.38238141 +0000 UTC    2023-04-12 11:45:32.245881097 +0000 UTC
10.0.0.103   time.cloudflare.com   2023-04-12 11:45:32.255538166 +0000 UTC   2023-04-12 11:45:32.249325211 +0000 UTC
[...]

PR #7683, opened 2023-08-29
refactor: update NTP spike detector

smira:
See https://github.com/siderolabs/talos/issues/7080#issuecomment-1696105986

The NTP spike detector code was refactored out of the main NTP code so that it can be unit-tested.

I dropped one check which I think is causing false-positives in the spike detector (when NTP offset is higher than the RTT of the best packet received so far).

The overall flow resembles the one in systemd-timesync, the current implementation has this check:

https://github.com/systemd/systemd/blob/6639ac474eb7a5325a72a3d7370492792dd00bc0/src/timesync/timesyncd-manager.c#L357-L360

This check was introduced in the initial release, after some refactoring:

https://github.com/systemd/systemd/commit/3dbc7620030b7eacfe19d55f1bcd8237dd3230ef#diff-4aa9995f07bb31b9884d40a7634f5f6d30245dfd26ac27b89cd5fd3bd4eef56aR429-R431

There is no equivalent of it in the RFC:

https://datatracker.ietf.org/doc/html/rfc5905#appendix-A.5.2

### DIFF
--- a/internal/pkg/ntp/internal/spike/spike.go
+++ b/internal/pkg/ntp/internal/spike/spike.go
@@ -1,0 +1,109 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package spike provides a spike detector for NTP responses.
+package spike
+
+import (
+	"math"
+
+	"github.com/beevik/ntp"
+)
+
+const defaultCapacity = 8
+
+// Sample is a single NTP response sample.
+type Sample struct {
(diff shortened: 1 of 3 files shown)


internal/pkg/ntp/internal/spike/spike.go
```diff
@@ -0,0 +1,109 @@
...
+		// always accept clock offset if that is larger than rtt
+		return false
+	}
+
+	if d.packetCount < 4 {
+		// need more samples to make a decision
+		return false
+	}
+
+	// This check was specifically removed (while it exists in systemd-timesync),
```
smira, line 92: the check I dropped

frezbo (approved):

Issue #7675, opened 2023-08-25
Show Secureboot/TrustedBoot status in Talos Dashboard


frezbo:

smira:
probably also encrypted status (?) for the disks

frezbo:
> probably also encrypted status (?) for the disks

yes, I was not sure if the dashboard shows secureboot and trustedboot explicitly

should we add trustedboot also as a field to securitystate?

smira:
I mean in general encrypted state might be nice... probably we need to show encryption method as well (?)

frezbo:
> I mean in general encrypted state might be nice... probably we need to show encryption method as well (?)

okay, this makes more sense, show encrypted state and method used

PR #7720, opened 2023-09-05
chore: show securtiystate on dashboard

frezbo:
Show Talos SecurityState and MountStatus on dashboard.

Fixes: #7675


frezbo:
Securebooted:

image

Normal:
image

internal/pkg/dashboard/components/talosinfo.go
```diff
@@ -93,6 +99,24 @@ func (widget *TalosInfo) updateNodeData(data resourcedata.Data) {
...
+	case *runtime.SecurityState:
+		if data.Deleted {
+			nodeData.secureBootState = notAvailable
+		} else {
+			nodeData.secureBootState = formatStatus(res.TypedSpec().SecureBoot)
+		}
+	case *runtime.MountStatus:
+		if data.Deleted {
+			nodeData.statePartitionMountStatus = notAvailable
+			nodeData.ephemeralPartitionMountStatus = notAvailable
```
utkuozdemir, line 111: shouldn't the switch below (between ephemeral and state partitions) apply to this as well?

internal/pkg/dashboard/components/talosinfo.go
```diff
@@ -158,8 +185,28 @@ func (widget *TalosInfo) redraw() {
...
 
 	widget.SetText(fields.String())
 }
+
+func mountStatus(encrypted bool, providers []string) string {
+	if !encrypted {
+		return "[green]OK[-]"
+	}
+
+	return fmt.Sprintf("[green]encrypted[-] (%s)", strings.Join(providers, ","))
```
utkuozdemir, line 211: Maybe still keep the `OK`, sth like
```suggestion
	return fmt.Sprintf("[green]OK - encrypted[-] (%s)", strings.Join(providers, ","))
```

But it's up to you

internal/pkg/dashboard/components/talosinfo.go
frezbo: yup, missed that

internal/pkg/mount/system.go
```diff
@@ -235,6 +237,25 @@ func SystemPartitionMount(ctx context.Context, r runtime.Runtime, logger *log.Lo
...
+			case cfg.NodeID() != nil:
+				encryptionProviders = append(encryptionProviders, cfg.NodeID().String())
+			case cfg.KMS() != nil:
+				encryptionProviders = append(encryptionProviders, cfg.KMS().String())
+			case cfg.TPM() != nil:
+				encryptionProviders = append(encryptionProviders, cfg.TPM().String())
+			}
+		}
+
+		mountStatus.TypedSpec().EncryptionProviders = maps.Keys(slices.ToSet(encryptionProviders))
```
utkuozdemir, line 256: We can directly gather them into a set (`map[string]struct{}{}`) instead of gathering them into a slice first and then doing `slices.ToSet(encryptionProviders)`.

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -1384,6 +1384,16 @@ func (e *EncryptionKey) TPM() config.EncryptionKeyTPM {
 	return e.KeyTPM
 }
 
+// String implements the config.Provider interface.
+func (e *EncryptionKeyNodeID) String() string {
+	return "tpm"
```
utkuozdemir, line 1389: is returning `tpm` here intentional (and not `node-id`)?

frezbo: co-pilot :joy:

utkuozdemir: happens to the best of us :D

internal/pkg/dashboard/components/talosinfo.go
frezbo: image
fixed

internal/pkg/dashboard/components/talosinfo.go
frezbo: fixed

internal/pkg/mount/system.go
frezbo: fixed, also verified by running reset on ephemeral and verified status changes

utkuozdemir (approved):
🆒

Issue #7612, opened 2023-08-14
imager: add handling of custom `pkgs`


smira:
Allow placing a custom `kernel` build, pulling the kernel image and recompiling `initramfs` with new kernel modules.

smira:
No progress, it's still possible to build `imager` from Talos sources with updated reference to the pkgs. Probably Talos 1.7 or later.

smira:
I don't quite see the question. Nothing really changed, as previously imager wasn't able to generate installer images. You can still generate installer directly from Talos sources.

yoctozepto:
Ok, I get it better now. It's just what was there already (have not used it before so thought it was a zero-to-hero development for 1.5...). So, there would be no problem in me adding a switch to consume the kernel and initramfs from the installer image itself, right? This way imager would be just a nice (and pretty straightforward) way to bake in extra components, leaving the basic kernel and initramfs customisation to the basic installer building workflow. It would be an application of [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), especially if applied by default. WDYT?

yoctozepto:
Ah, ok. Again, I get it yet better now. There are two thought issues I still have - 1) the imager does not really seem to be "building" the installer image as it consumes the "base installer" (as it calls it). 2) The imager is the only way to cleanly bake in extensions, the other process does not have this ability - this is also why I had this #7729 before.
(1) implies the imager is not a "single stop" solution for building assets. (2) implies it is required for extension consumption.
(1) and (2) imply that the overall process is more convoluted that it could (and should) probably be.

I hope you see my perspective better now.

Now, I agree it may be made to consume zero assets from the installer container image but this is not the case, at least in v1.5 and that's why I proposed the other way around.

yoctozepto:
I dug into it deeper and, to my surprise, imager image = installer image, except for the entrypoint... And it automagically sets reference-to-self as the default base installer image used.

smira:
Instead of implementing a custom path, we'd rather re-use the Talos build process for customization. The documentation is coming

PR #7727, opened 2023-09-07
chore: add PKG_KERNEL arg to customize used kernel

yoctozepto:
Related to #7612

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@
 
 ARG TOOLS
 ARG PKGS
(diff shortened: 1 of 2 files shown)


yoctozepto:
@smira @frezbo Could we get this in as it does not change the behaviour?

smira:
> @smira @frezbo Could we get this in as it does not change the behaviour?

I will take a look tomorrow

smira (approved):

PR #7729, opened 2023-09-07
PoC: feat: support early ucode loading

yoctozepto:
Related to #7612

Just a proof of concept (PoC).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -701,6 +701,16 @@ RUN set -o pipefail \
     | cpio --reproducible -H newc -o \
     | xz -v -C crc32 -0 -e -T 0 -z \
     > /initramfs.xz
+COPY --from=ghcr.io/siderolabs/intel-ucode:20230808 /rootfs/lib/firmware/intel-ucode /intel-ucode
+RUN mkdir /UCODE \
+    && cd /UCODE \
+    && mkdir -p kernel/x86/microcode \
+    && cat /intel-ucode/* > kernel/x86/microcode/GenuineIntel.bin \
+    && find . | cpio -o -H newc > /ucode.cpio \
+    && mv /initramfs.xz /initramfs.xz.orig \
+    && cat /ucode.cpio /initramfs.xz.orig > /initramfs.xz \
+    && cd / \
+    && rm -rf /UCODE /intel-ucode /ucode.cpio /initramfs.xz.orig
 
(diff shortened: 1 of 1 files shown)


smira:
Can you please add some details, do we have ucode in the "wrong" place right now?

yoctozepto:
> Can you please add some details, do we have ucode in the "wrong" place right now?

Sure, this is related also to the discussion we two had at slack's #support. Just adding the extension did nothing.

Following the https://www.kernel.org/doc/html/v6.1/x86/microcode.html, the recommended (and only one supported on my baremetal) method is the early loading. I reimplement their steps with talos extensions as shown in the diff. It can be done analogously for AMD but I don't have spare AMD servers to test so did not bother implementing.

smira:
I see, thanks for providing that link. I guess when we first implemented that, late loading was supposed to be okay.

This would be a significant amount of work to re-package them this way, and not sure how much it will increase the initramfs image.

frezbo:
> I see, thanks for providing that link. I guess when we first implemented that, late loading was supposed to be okay.
> 
> This would be a significant amount of work to re-package them this way, and not sure how much it will increase the initramfs image.

i guess we could still ship this as an extension? since we just need to have that path in the allowlist.

yoctozepto:
TBH, I would just bake both ucodes in their latest versions at the time into the default image ~> people need these even if they don't know about them. ;-) And provide an easy way to rebuild with newer ones inbetween talos releases ~> aaand this way we are back again at #7612.

frezbo:
> TBH, I would just bake both ucodes in their latest versions at the time into the default image ~> people need these even if they don't know about them. ;-) And provide an easy way to rebuild with newer ones inbetween talos releases ~> aaand this way we are back again at #7612.

we don;t want to unnecessarily increase the initramfs size, the image service is going to make this easier. There could be a flavour with both already available.

yoctozepto:
Fair point. I also thought about cloud images where ucode is simply ignored so it would be a waste. But for metal... maybe it just makes sense in this particular flavor (amd64 metal)?

smira:
It will be easier with image service, so you can get only Intel or AMD ucode, you never need both.

yoctozepto:
Agreed. Though, as it renders for me, AMD is under 100 KiB and Intel about 7 MiB so it is nowhere near a huge dependency to have them always around. :D That said, I still look forward to a better imaging experience and happy to collaborate on it.

PR #7747, opened 2023-09-12
docs: add notes about stable addressing

### LINKED ISSUE #7718: document etcd and kubelet subnet configuration
When a machine has multiple networks (e.g. private/public, IPv6/IPv4), or additional networks created dynamically (e.g. by the Kubernetes workloads), it's critical to make sure that

* kubelet IP is stable
* etcd IP is stable (this is very important)

Talos can handle changes to the addresses, but for `etcd` it might have completely undesired consequences.

Document that in the production notes: how to specify correct addresses/subnets.

yoctozepto:
etcd and kubelet may benefit from controlling the subnets used for establishing their addresses.

Closes #7718

### DIFF
--- a/website/content/v1.5/introduction/prodnotes.md
+++ b/website/content/v1.5/introduction/prodnotes.md
@@ -99,6 +99,48 @@ Then, your endpoint would be:
 https://kube.cluster1.mydomain.com:6443
 ```
 
+## Multihoming
+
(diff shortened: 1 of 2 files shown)


smira (approved):
thanks for writing this up, I will expand this a bit with examples

PR #7751, opened 2023-09-13
fix: handle correctly change of listen address for maintenance service

### LINKED ISSUE #7738: Kernel log sender gets stuck when the connection is aborted

smira:
Fixes #7738

If the SideroLink address changes, maintenance service should listen on new address. Previously it worked "sometimes", as there was a race on maintenance config either be removed/recreated or just updated. In case of an update the listen address was not updated properly, but recreate case worked correctly.


frezbo (approved):
maybe a test too?

smira:
> maybe a test too?

it'd be hard, as the test listens on localhost... but probably I could try listening on a different port :thinking:

Unix4ever (approved):

smira:
> > maybe a test too?
> 
> it'd be hard, as the test listens on localhost... but probably I could try listening on a different port 🤔

added a test with a change of port

PR #7763, opened 2023-09-15
refactor: reimplement the depmod extension rebuilder

smira:
Drop loop device/mounts completely, use userspace utilities to extract and lay over module trees in the tmpfs.

Discover kernel version automatically instead of hardcoding it to be current one (required for Image Service).

### DIFF
--- a/internal/pkg/extensions/discarder.go
+++ b/internal/pkg/extensions/discarder.go
@@ -1,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package extensions
+
+import (
+	"fmt"
+	"io"
+)
+
+// discarder is used to implement ReadAt from a Reader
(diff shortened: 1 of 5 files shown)


internal/pkg/extensions/discarder.go
```diff
@@ -0,0 +1,47 @@
...
+// is reached. It can only go forward. It is designed
+// for pipe-like files.
+type discarder struct {
+	r   io.Reader
+	pos int64
+}
+
+// ReadAt implements ReadAt for a discarder.
+// It is an error for the offset to be negative.
+func (r *discarder) ReadAt(p []byte, off int64) (int, error) {
```
smira, line 23: this allows to use `cpio` reader without fully uncompressing `initramfs`

frezbo, line 23: do we need a test for this? :thinking:

frezbo (approved):

internal/pkg/extensions/discarder.go
smira: I looked it up from the u-root internals, it's being used for non-seekable files, but there's no way I could use it directly

PR #7767, opened 2023-09-18
fix: build CPU ucode correctly for early loader

smira:
Closes #7729

This follows the steps described in
https://www.kernel.org/doc/html/v6.1/x86/microcode.html#early-load-microcode

### DIFF
--- a/internal/pkg/extensions/kernel_modules.go
+++ b/internal/pkg/extensions/kernel_modules.go
@@ -39,7 +39,7 @@ func (ext *Extension) KernelModuleDirectory() string {
 // GenerateKernelModuleDependencyTreeExtension generates a kernel module dependency tree extension.
 //
 //nolint:gocyclo
-func GenerateKernelModuleDependencyTreeExtension(extensionsPathWithKernelModules []string, initramfsPath string, printFunc func(format string, v ...any)) (*Extension, error) {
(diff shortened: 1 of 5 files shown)


smira:
(I'm looking to add some integration tests to the Image Factory/Service).

frezbo (approved):

yoctozepto:
Thanks for working on this!

smira:
> Thanks for working on this!

thank you for bringing this up!

internal/pkg/extensions/kernel_modules.go
```diff
@@ -105,14 +105,11 @@ func GenerateKernelModuleDependencyTreeExtension(extensionsPathWithKernelModules
 		return nil, fmt.Errorf("error running depmod: %w", err)
 	}
 
-	// we want this temp directory to be present until the extension is compressed later on, so not removing it here
-	kernelModulesDependencyTreeStagingDir, err := os.MkdirTemp("", "module-deps")
-	if err != nil {
-		return nil, err
-	}
+	// we want the extension to survive this function, so not storing in a temporary directory
```
smira, line 108: I fixed this to use an existing extension tree path instead of temp dir, otherwise it won't be cleaned up properly (leaking temporary directory)

PR #7769, opened 2023-09-18
docs: add control plane nodes as users of apid also for control plane nodes

yoctozepto:
Not only worker nodes tend to be accessed from control plane nodes. Other control plane nodes can also be accessed from another control plane node.


smira (approved):
thank you!

yoctozepto:
Also good to learn that trustd is really not used by other control plane nodes!

smira:
> Also good to learn that trustd is really not used by other control plane nodes!

This was the case some long time ago, but right now control plane nodes are self-sufficient - they can work in complete isolation.

yoctozepto:
> > Also good to learn that trustd is really not used by other control plane nodes!
> 
> This was the case some long time ago, but right now control plane nodes are self-sufficient - they can work in complete isolation.

Ack! This agrees with the observed behaviour and high-level code review by me but I was not sure yet so thanks for clarifying.

PR #7773, opened 2023-09-19
chore: add tests for chelsio drivers

frezbo:
Add tests for Chelsio drivers and firmware.

Ref: https://github.com/siderolabs/extensions/pull/232


smira (approved):
promote it?

frezbo:
> promote it?

well it's broken :sweat_smile: , since it uses old code to install extensions

frezbo:
i tested locally

smira:
> > promote it?
> 
> well it's broken 😅 , since it uses old code to install extensions

ah... yes... I need to get a fix in... my bad

PR #7776, opened 2023-09-20
chore: clear key after use.

frezbo:
There's no need to keep the random bytes around after it's used rather than wait for GC.


internal/pkg/encryption/keys/tpm2.go
```diff
@@ -53,6 +53,9 @@ func (h *TPMKeyHandler) NewKey(ctx context.Context) (*encryption.Key, token.Toke
 		return nil, nil, err
 	}
 
+	// explicitly clear key from memory since it's not needed anymore
```
smira, line 56: we seem to be using it below on line 72, so I don't think it's _that_ simple, we still need to pass it down to LUKS

frezbo: ohhh, I missed that completely :sweat_smile: . I saw it in the kms, but missed here

frezbo:
closing for now, since it needs more workarounds.

PR #7780, opened 2023-09-21
feat: support n-5 latest Kubernetes versions

smira:
For Talos 1.6 this means 1.24-1.29 Kubernetes.


smira:
(still need to catch up with tests)

smira:
> maybe move the `gen` changes to a different PR, I think I had one locally and lost it

this got mixed in as I had to update go-kubernetes, I can split it up tomorrow

.drone.jsonnet
```diff
@@ -505,7 +505,8 @@ local integration_bios_cgroupsv1 = Step('e2e-bios-cgroupsv1', target='e2e-qemu',
   SHORT_INTEGRATION_TEST: 'yes',
   WITH_UEFI: 'false',
   IMAGE_REGISTRY: local_registry,
-  WITH_CONFIG_PATCH: '[{"op": "add", "path": "/machine/install/extraKernelArgs/-", "value": "talos.unified_cgroup_hierarchy=0"}]',  // use cgroupsv1
+  // cgroupsv1 disabled, see: https://github.com/kubernetes/kubernetes/issues/120813
```
yoctozepto, line 508: The name of the step/test no longer makes sense now.

smira: Hopefully the upstream issue will get fixed, and we will re-enable

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -431,60 +432,86 @@ func syncManifests(ctx context.Context, objects []*unstructured.Unstructured, cl
 }
 
 //nolint:gocyclo
-func checkPodStatus(ctx context.Context, cluster UpgradeProvider, service, node, configVersion string) error {
+func checkPodStatus(ctx context.Context, cluster UpgradeProvider, options UpgradeOptions, service, node, configVersion string) error {
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
smira, line 436: rewrote this piece to use Kubernetes watch instead of polling

internal/integration/provision/provision.go
```diff
@@ -71,3 +103,533 @@ func trimVersion(version string) string {
 	// remove anything extra after semantic version core, `v0.3.2-1-abcd` -> `v0.3.2`
 	return regexp.MustCompile(`(-\d+-g[0-9a-f]+)$`).ReplaceAllString(version, "")
 }
+
+var defaultNameservers = []netip.Addr{netip.MustParseAddr("8.8.8.8"), netip.MustParseAddr("1.1.1.1")}
+
+// BaseSuite provides base features for provision tests.
+type BaseSuite struct {
```
smira, line 110: no real changes, pulled out common part of `UpgradeSuite` into `BaseSuite` to re-use with Kubernetes version test

internal/integration/provision/k8s_compatibility.go
```diff
@@ -5,3 +5,145 @@
...
+// K8sCompatibilitySuite ...
+type K8sCompatibilitySuite struct {
+	BaseSuite
+
+	track int
+
+	versionsSequence []string
+}
+
+const supportedPastKubernetesReleases = 4 // should be '5' once we upgrade to 1.29
```
smira, line 33: Kubernetes 1.29 is not released yet even for `alpha.1`, so we're still on 1.28, and limiting `n-5` to `n-4` to keep things correct

smira: I'm going to upgrade to 1.29.0-alpha.1 first and take it from there

.drone.jsonnet
frezbo: i guess we can re-enable with newer pkgs

hack/release.toml
```diff
@@ -51,6 +51,16 @@ Previously, [network device selectors](https://www.talos.dev/v1.6/talos-guides/n
         title = "KubePrism"
         description = """\
 [KubePrism](https://www.talos.dev/v1.6/kubernetes-guides/configuration/kubeprism/) is enabled by default on port 7445.
+"""
+
+    [notes.cgroupsv1]
```
frezbo, line 56: we can remove with newer pkgs

internal/integration/integration_test.go
```diff
@@ -165,7 +165,7 @@ func init() {
 	flag.Int64Var(&provision_test.DefaultSettings.CPUs, "talos.provision.cpu", provision_test.DefaultSettings.CPUs, "CPU count for each VM (provision tests only)")
 	flag.Int64Var(&provision_test.DefaultSettings.MemMB, "talos.provision.mem", provision_test.DefaultSettings.MemMB, "memory (in MiB) for each VM (provision tests only)")
 	flag.Uint64Var(&provision_test.DefaultSettings.DiskGB, "talos.provision.disk", provision_test.DefaultSettings.DiskGB, "disk size (in GiB) for each VM (provision tests only)")
-	flag.IntVar(&provision_test.DefaultSettings.MasterNodes, "talos.provision.masters", provision_test.DefaultSettings.MasterNodes, "master node count (provision tests only)")
+	flag.IntVar(&provision_test.DefaultSettings.ControlplaneNodes, "talos.provision.controlplanes", provision_test.DefaultSettings.ControlplaneNodes, "master node count (provision tests only)")
```
frezbo, line 168:
```suggestion
	flag.IntVar(&provision_test.DefaultSettings.ControlplaneNodes, "talos.provision.controlplanes", provision_test.DefaultSettings.ControlplaneNodes, "controlplane node count (provision tests only)")
```

internal/integration/provision/provision.go
```diff
@@ -71,3 +103,533 @@ func trimVersion(version string) string {
...
+	}
+}
+
+// waitForClusterHealth asserts cluster health after any change.
+func (suite *BaseSuite) waitForClusterHealth() {
+	runs := 1
+
+	singleNodeCluster := len(suite.Cluster.Info().Nodes) == 1
+	if singleNodeCluster {
+		// run health check several times for single node clusters,
```
frezbo, line 177: we don;t use self hosted anymore, right?

hack/release.toml
smira: yes, that's why I'm doing pkgs in the background :)

internal/integration/provision/provision.go
smira: I think it wasn't specifically because of self-hosted, but rather because a single API server might "lie" about the state, as it's all single kubelet/api-server and they might get out of sync when both are restarted

smira: and this is moved from another file, I haven't changed it actually.

go.mod
```diff
@@ -6,6 +6,9 @@ replace (
 	// Use nested module.
 	github.com/siderolabs/talos/pkg/machinery => ./pkg/machinery
 
+	// see https://github.com/vmware-tanzu/sonobuoy/pull/1933
+	github.com/vmware-tanzu/sonobuoy => github.com/smira/sonobuoy v0.0.0-20230925141431-e9307f0a884d
```
frezbo, line 10: i guess we get this in now and wait for upstream?

frezbo (approved):

go.mod
smira: Yes, otherwise it would waste 500% of the CPU by the end of the test

PR #7788, opened 2023-09-25
feat: enable KubePrism by default

### LINKED ISSUE #7787: enable KubePrism by default

smira:
Fixes #7787

### DIFF
--- a/pkg/machinery/config/generate/worker.go
+++ b/pkg/machinery/config/generate/worker.go
@@ -74,10 +74,17 @@ func (in *Input) worker() ([]config.Document, error) {
 		machine.MachineFeatures.DiskQuotaSupport = pointer.To(true)
 	}
 
-	if in.Options.KubePrismPort > 0 {
+	if kubePrismPort, optionSet := in.Options.KubePrismPort.Get(); optionSet { // default to enabled, but if set explicitly, allow it to be disabled
+		if kubePrismPort > 0 {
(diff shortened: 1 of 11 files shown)


frezbo (approved):

pkg/machinery/config/generate/worker.go
```diff
@@ -74,10 +74,17 @@ func (in *Input) worker() ([]config.Document, error) {
 		machine.MachineFeatures.DiskQuotaSupport = pointer.To(true)
 	}
 
-	if in.Options.KubePrismPort > 0 {
+	if in.Options.KubePrismPort.IsPresent() {
+		if kubePrismPort := in.Options.KubePrismPort.ValueOrZero(); kubePrismPort > 0 {
```
DmitriyMV, line 78: This broke my head. I just realized that we use `KubePrismPort.Value == 0` as a signal value, but `ValueOrZero` will also return `0` if there is nothing inside, and the thing protecting us from that is `in.Options.KubePrismPort.IsPresent()` expression above.

I actually think that you can do better here:
```suggestion
	if kubePrismPort, present := in.Options.KubePrismPort.Get(); present {
		if kubePrismPort > 0 {
```

This is both shorter and clearer IMHO.

DmitriyMV (approved):

pkg/machinery/config/generate/worker.go
smira: I thought of that, but it looks less explicit to me, but :shrug:  not a big deal

PR #7796, opened 2023-09-27
feat: update containerd to 1.7.x

smira:
Also update Linux and other pkgs.


pkg/machinery/proto/proto.go
```diff
@@ -254,7 +266,14 @@ func registerDefaultTypes() {
 
 	protoenc.RegisterEncoderDecoder(
 		func(v specs.Mount) ([]byte, error) {
-			mount := Mount(v)
+			mount := Mount{
```
smira, line 269: @DmitriyMV any idea how to make it better?

with `[]LinuxIDMapping` fields the struct types are no longer identical, so we can't type convert & assign.

But my biggest concern is that it's easy to miss if a new field is added to `specs.Mount`?

frezbo (approved):
probably update the `hack/release.toml` for the cgroup thing too

smira:
> probably update the `hack/release.toml` for the cgroup thing too

I don't need it (yet), as it hasn't been merged, I will remove it from the other PR

pkg/machinery/proto/proto.go
DmitriyMV: Yep. You can go with Go “type promotion” from typed<->untyped complex types.

That is you can define:
```go
func convertSpecMount(v struct {
	Destination string                 `json:"destination"`
	Type        string                 `json:"type,omitempty" platform:"linux,solaris,zos"`
	Source      string                 `json:"source,omitempty"`
	Options     []string               `json:"options,omitempty"`
	UIDMappings []specs.LinuxIDMapping `json:"uidMappings,omitempty" platform:"linux"`
	GIDMappings []specs.LinuxIDMapping `json:"gidMappings,omitempty" platform:"linux"`
}) Mount {
	return Mount{
		Destination: v.Destination,
		Type:        v.Type,
		Source:      v.Source,
		Options:     v.Options,
		UIDMappings: xslices.Map(v.UIDMappings, func(v specs.LinuxIDMapping) LinuxIDMapping { return LinuxIDMapping(v) }),
		GIDMappings: xslices.Map(v.GIDMappings, func(v specs.LinuxIDMapping) LinuxIDMapping { return LinuxIDMapping(v) }),
	}
}
```
And use it as

```go
		func(v specs.Mount) ([]byte, error) {
			mount := convertSpecMount(v)

			return protoenc.Marshal(&mount)
		},
```

DmitriyMV: https://go.dev/ref/spec#Assignability relevant.

smira: great idea, thanks!

Issue #7810, opened 2023-09-29
unsupported hash algorithm: 18


uhthomas:
### Description

I'm trying to install Talos fresh with system disk encryption and seeing the error `unsupported hash algorithm: 18`.

```yaml
  systemDiskEncryption:
    ephemeral:
      keys:
        - tpm: {}
          slot: 0
    state:
      keys:
        - tpm: {}
          slot: 0
```

### Environment

- Talos version: v1.5.3
- Kubernetes version: N/A
- Platform: bare metal

frezbo:
it seems the tpm doesn;t support sha256?

Could you provide the dmesg?

uhthomas:
![image](https://github.com/siderolabs/talos/assets/9749173/3745dc4a-220a-4d43-a5b4-3e2e32e3fde6)

uhthomas:
18 seems to be `BLAKE2b-384`?

https://go.dev/play/p/zXQHvXE22p2

frezbo:
it seems the tpm supports `BLAKE2b-384` as a has alg, and the tpm library used doesn't support that

frezbo:
i'll think of a fix for next week

uhthomas:
@frezbo I am confused as to where the value is coming from, as the TPM package does not seem to even recognise blake2b-384 at all and should have returned an error, which should have been handled correctly.

https://github.com/google/go-tpm/blob/5a514e64d1ed5d986ccfd38b958d2d5c038dcccd/tpm2/structures.go#L431

https://github.com/siderolabs/talos/blob/0bd1bdd744f68dc42ac64678972fede992a7189e/internal/pkg/secureboot/tpm2/pcr.go#L169

frezbo:
> @frezbo I am confused as to where the value is coming from, as the TPM package does not seem to even recognise blake2b-384 at all and should have returned an error, which should have been handled correctly.
> 
> https://github.com/google/go-tpm/blob/5a514e64d1ed5d986ccfd38b958d2d5c038dcccd/tpm2/structures.go#L431
> 
> https://github.com/siderolabs/talos/blob/0bd1bdd744f68dc42ac64678972fede992a7189e/internal/pkg/secureboot/tpm2/pcr.go#L169

yes, we reutrn the error as is since we don;t support BLAKE at all

uhthomas:
You're right, I misread the error. Thank you, it looks like that is the issue.

[...]

PR #7811, opened 2023-09-29
fix: use tpm2 hash algorithm constants and allow non-SHA-256 PCRs

uhthomas:
# Pull Request

## What? (description)

The conversion from TPM 2 hash algorithm to Go crypto algorithm will fail for uncommon algorithms like SM3256. This can be avoided by checking the constants directly, rather than converting them. It should also be fine to allow some non SHA-256 PCRs.

## Why? (reasoning)

Fixes: #7810

## Acceptance

### DIFF
--- a/internal/pkg/secureboot/tpm2/pcr.go
+++ b/internal/pkg/secureboot/tpm2/pcr.go
@@ -7,7 +7,6 @@ package tpm2
 
 import (
 	"bytes"
-	"crypto"
 	"crypto/sha256"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


frezbo:
i think i'd still fail if sha256 is not found, since it would fail elsewhere. I;ll clean that up

uhthomas:
@frezbo Is it possible to build a version of Talos with this change? I can test it on the affected system. I believe it should be fine as there will be sha256 banks available - the original code would fail if there were any non-sha256 banks present at all, which I believe was the main problem.

frezbo:
@uhthomas Could you try this image: `ghcr.io/frezbo/installer:v1.6.0-alpha.0-60-g733aa4df2-amd64-secureboot`?

This would work with a fresh boot only, an upgrade won't work

frezbo:
> ~i think i'd still fail if sha256 is not found, since it would fail elsewhere. I;ll clean that up~

this makes no sense, since it loops over all banks

frezbo (approved):

uhthomas:
@frezbo Do you also have an ISO I could use? I just tried with the installer you provided, but it had the same error (which indicates to me the code hasn't changed, and the ISO probably needs to be updated too).

frezbo:
> @frezbo Do you also have an ISO I could use? I just tried with the installer you provided, but it had the same error (which indicates to me the code hasn't changed, and the ISO probably needs to be updated too).

Ahh I see the problem, the first code runs from iso :sweat_smile: . I'll generate one and sent over

uhthomas:
@frezbo Thanks for building that ISO for me. I've been able to install Talos with secureboot and TPM encryption.

PR #7815, opened 2023-10-01
Docs: Fix typos in example talosctl commands

zmilonas:
## What? (description)
Fixed some typos i encountered when setting up


smira (approved):
thank you!

PR #7847, opened 2023-10-12
test: add config generation stability tests

smira:
These tests ensure that config generation for older versions of Talos is stable as we move forward and introduce new features in the machine configuration.


pkg/machinery/config/types/v1alpha1/v1alpha1_stability_test.go
```diff
@@ -0,0 +1,121 @@
...
+	const generateMode = false
+
+	secretsBundle, err := secrets.LoadBundle("testdata/stability/secrets.yaml")
+	require.NoError(t, err)
+
+	for _, versionContract := range []*config.VersionContract{
+		config.TalosVersion1_3,
+		config.TalosVersion1_4,
+		config.TalosVersion1_5,
+		config.TalosVersion1_6,
```
utkuozdemir, line 38: should we make this fail explicitly if there's a newer version untested? so we don't forget it in the future?

utkuozdemir (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_stability_test.go
smira: I can do that, good point

Unix4ever (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_stability_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/siderolabs/talos/pkg/machinery/gendata"
+)
+
+// TestConfigEncodingStability ensures that the encoding of a configuration is stable as we moved forward with the config format.
+func TestConfigEncodingStability(t *testing.T) {
+	t.Parallel()
+
+	// flip this to generate missing configs
+	const generateMode = false
```
utkuozdemir, line 32: cool trick

pkg/machinery/config/types/v1alpha1/v1alpha1_stability_test.go
smira: added a check

utkuozdemir (approved):

PR #7860, opened 2023-10-16
chore: support OCI layout as a source for profile input

smira:
This supports full Image interface.


smira:
Required for https://github.com/siderolabs/image-factory/issues/39

Unix4ever (approved):

Issue #7878, opened 2023-10-20
unable set sysctl with a dot in the path (vlan interface)


nberlee:
## Bug Report

dots in sysctls are always translated to a / whereas a vlan interface is in format `netiface.vlanid` which leads to an / between netiface and vlanid.

### Description
When adding to the machine config:
```yaml
machine:
   network:
      interfaces:
        -  interface: eth0
           vlans:
             - vlanId: 103
               dhcp: false
  sysctls:
    net.ipv6.conf.eth0.103.disable_ipv6: "1"
```
talosctl ls confirms the existence of `/proc/sys/net/ipv6/conf/eth0.103/disable_ipv6`

### Logs
I get the error in dmesg:
```
[talos] controller failed {"component": "controller-runtime", "controller": "runtime.KernelParamSpecController", "error": "1 error occurred:\x5cn\x5ct* open /proc/sys/net/ipv6/conf/eth0/103/disable_ipv6: no such file or directory\x5cn\x5cn"}
```
please note the slash between `eth0` and `103` this should be an `.` but there does not seems to be a way to escape the dot in machine config.  A dot in hex `\x2e`  also does not work as `\` gets escaped with `\x5c`
### Environment

- Talos version: 1.5.4
- Kubernetes version: [`kubectl version --short`]
- Platform: LInux

smira:
yep, that's a problem... but not sure what we can do here, need to invent something

PR #7879, opened 2023-10-20
fix: accept sysctl/sysfs paths with dots

nberlee:
fixes: #7878

# Pull Request

## Acceptance


smira (approved):
fixed it up to comply with `sysctl` utility handling, but same idea

nberlee:
> fixed it up to comply with `sysctl` utility handling, but same idea

Yes, thank you! This is safer because it mitigates path-traversal. Something only thought of later.

hack/release.toml
```diff
@@ -52,6 +52,20 @@ Previously, [network device selectors](https://www.talos.dev/v1.6/talos-guides/n
...
+        title = "Sysctl"
+        description = """\
+Talos now accepts '/' as a separator for sysctl keys, which allows to set sysctl keys which contain '.' in their name.
+
+Example:
+
+```yaml
+machine:
+  sysctls:
+    net/ipv6/conf/eth0.100/disable_ipv6: "1"
```
nberlee, line 67: this is backwards

`net.ipv6.conf.eth0/100.disable_ipv6`

pkg/machinery/config/types/v1alpha1/v1alpha1_examples.go
```diff
@@ -258,8 +258,9 @@ func machineTimeExample() *TimeConfig {
 
 func machineSysctlsExample() map[string]string {
 	return map[string]string{
-		"kernel.domainname":   "talos.dev",
-		"net.ipv4.ip_forward": "0",
+		"kernel.domainname":                   "talos.dev",
+		"net.ipv4.ip_forward":                 "0",
+		"net/ipv6/conf/eth0.100/disable_ipv6": "1",
```
nberlee, line 263: also here should be: `net.ipv6.conf.eth0/100.disable_ipv6`
This was my mistake

smira: I think it's correct according to the spec, but still my implementation was not correct:

```
Note that either "/" or "."  may be used as separators within
       sysctl variable names. If the first separator is a slash,
       remaining slashes and dots are left intact. If the first
       separator is a dot, dots and slashes are interchanged.
       "kernel.domainname=foo" and "kernel/domainname=foo" are
       equivalent and will cause "foo" to be written to
       /proc/sys/kernel/domainname. Either
       "net.ipv4.conf.enp3s0/200.forwarding" or
       "net/ipv4/conf/enp3s0.200/forwarding" may be used to refer to
       /proc/sys/net/ipv4/conf/enp3s0.200/forwarding
```

smira: we might provide either of as an example

smira:
> Yes, thank you! This is safer because it mitigates path-traversal. Something only thought of later.

actually path traversal was (and still is) a problem... not a security issue per se, as it's part of the machine config, but nice to avoid

pkg/machinery/config/types/v1alpha1/v1alpha1_examples.go
nberlee: Ok I see, but this test should pass right? It doesn't anymore....

```
		{
			name: "Test Sysctl Path with vlan interface other notation",
			param: &kernel.Param{
				Key: kernel.Sysctl + ".net.ipv6.conf.eth0/103.disable_ipv6",
			},
			want: "/proc/sys/net/ipv6/conf/eth0.103/disable_ipv6",
		},
```

smira: yes, working on that!

smira: should be better now

nberlee: Yes, this is great!

smira:
> > Yes, thank you! This is safer because it mitigates path-traversal. Something only thought of later.
> 
> actually path traversal was (and still is) a problem... not a security issue per se, as it's part of the machine config, but nice to avoid

I added `path.Clean()` to prevent "stupid" path traversal at least.

PR #7886, opened 2023-10-22
fix: make Talos work on Rockpi 4c boards again

Unix4ever:
Suppress `efivars` `ENODEV` errors: skip mount and proceed with boot sequence.

### DIFF
--- a/internal/pkg/mount/mount.go
+++ b/internal/pkg/mount/mount.go
@@ -7,6 +7,7 @@ package mount
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -67,6 +68,14 @@ func mountMountpoint(mountpoint *Point) (skipMount bool, err error) {
 
 	if !skipMount {
 		if err = mountpoint.Mount(); err != nil {
+			if mountpoint.MountFlags.Check(SkipIfNoDevice) && errors.Is(err, unix.ENODEV) {
+				if mountpoint.Logger != nil {
+					mountpoint.Logger.Printf("error mounting: %q: %s", mountpoint.Source(), err)
+				}
+
+				return true, nil
+			}
+
 			return false, fmt.Errorf("error mounting: %w", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/mount/pseudo.go
```diff
@@ -35,7 +35,9 @@ func PseudoSubMountPoints() (mountpoints *Points, err error) {
 
 	if _, err := os.Stat(constants.EFIVarsMountPoint); err == nil {
```
frezbo, line 36: i think the rockpi board lies about efivars being present to the kernel :sweat_smile:

Unix4ever: yeah, that might explain why it works on other boards :)

internal/pkg/mount/options.go
```diff
@@ -33,6 +33,8 @@ const (
 	SkipIfMounted
 	// SkipIfNoFilesystem is a flag for skipping formatting and mounting if the mountpoint has not filesystem.
 	SkipIfNoFilesystem
+	// SkipIfNoDevice is a flag for skipping errors when the device is not found.
+	SkipIfNoDevice
```
smira, line 37: I don't see this option being used at all

internal/pkg/mount/mount.go
```diff
@@ -67,6 +68,14 @@ func mountMountpoint(mountpoint *Point) (skipMount bool, err error) {
 
 	if !skipMount {
 		if err = mountpoint.Mount(); err != nil {
+			if errors.Is(err, unix.ENODEV) {
```
smira, line 71: this should probably use the new option?

Unix4ever: yeah...

internal/pkg/mount/options.go
```diff
@@ -49,6 +51,7 @@ type Options struct {
 	SystemInformationGetter helpers.SystemInformationGetter
 	Logger                  *log.Logger
 	ProjectQuota            bool
+	SkipFailure             bool
```
smira, line 54: what is that?

smira (approved):

Issue #7900, opened 2023-10-25
talosctl shutdown doesn't error if no talosconfig is passed.


iMartyn:
## Bug Report

I ended up in a terminal where I hadn't exported `TALOSCONFIG` and issued a `talosctl shutdown` command and nothing happened.  I assumed it was an async command and waited... and waited.

### Description

There should be an error message if you issue a `talosctl shutdown` command without either an explicit `--talosconfig` or a `TALOSCONFIG` env var.

### Logs

```
/home/martyn [martyn@i9worker] [22:16]
> unset TALOSCONFIG

/home/martyn [martyn@i9worker] [22:16]
> talosctl shutdown -n 192.168.1.206 -e 192.168.1.206 --force

/home/martyn [martyn@i9worker] [22:16]
```
(no output seen on shutdown command)

### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`] 1.5.4
- Kubernetes version: [`kubectl version --short`] v1.28.2
- Platform: metal-pine64

smira:
you probably have a default `talosconfig` which has some (unreachable?) nodes, which `talosctl is using`.

`talosctl config info` to help, not an issue in the CLI.

iMartyn:
No, I don't, and this is still an error that the CLI should handle, no matter what the cause!

```
/home/martyn [martyn@i9worker] [13:38]
> talosctl config info
no context is set

/home/martyn [martyn@i9worker] [13:38]
> talosctl shutdown -n 192.168.1.206 -e 192.168.1.206 --force

/home/martyn [martyn@i9worker] [13:39]
>
```
This is actually an issue in the CLI - no context should always return an error, not silently fail!

Please reopen.

smira:
yep, looks like error reporting is broken there

smira:
@utkuozdemir I think something is wrong here

https://github.com/siderolabs/talos/blob/ef5056122b38a168dd8ee429a6bc4cad0860177d/cmd/talosctl/pkg/talos/action/tracker.go#L210-L213

and many commands using this code path are broken same way

utkuozdemir:
@smira Yes, seems like it. I'll submit a fix.

PR #7906, opened 2023-10-27
fix: fix error output of cli action tracker

utkuozdemir:
Before we started a reboot/shutdown/reset/upgrade action with the action tracker (`--wait`), we were setting a flag to prevent cobra from printing the returned error from the command.

This was to prevent the error from being printed twice, as the reporter of the action tracker already prints any errors occurred during the action execution.

But if the error happens too early - i.e. before we even started the status printer goroutine, then that error wouldn't be printed at all, as we have suppressed the errors.

This PR moves the suppression flag to be set after the status printer is started - so we still do not double-print the errors, but neither do we suppress any early-stage error from being printed.

Closes siderolabs/talos#7900.

### DIFF
--- a/cmd/talosctl/cmd/talos/reboot.go
+++ b/cmd/talosctl/cmd/talos/reboot.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/siderolabs/talos/cmd/talosctl/cmd/common"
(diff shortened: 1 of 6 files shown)


frezbo (approved):
let's also add a test to the existing one

smira (approved):
yep, probably a simple CLI test where the command should fail early and print an error

utkuozdemir:
> yep, probably a simple CLI test where the command should fail early and print an error

done

PR #7907, opened 2023-10-27
fix: generate images for SBCs using imager

smira:
See https://github.com/siderolabs/image-factory/issues/43

Two fixes:

* pass path to the dtb, uboot and rpi-firmware explicitly
* include dtb, uboot and rpi-firmware into arm64 installer image when generated via imager (regular arm64 installer was fine)

(The generation of SBC images was not broken for Talos itself, but only when used via Image Factory).


frezbo (approved):

frezbo:
this is nice, later we could support the raspberry `config.txt` also as an option

PR #7926, opened 2023-11-03
feat: update Kubernetes to 1.29.0-alpha.3

smira:
See https://github.com/kubernetes/kubernetes/releases/v1.29.0-alpha.3


smira:
NB: need to pull in https://github.com/siderolabs/go-kubernetes/pull/12

PR #7930, opened 2023-11-06
feat: rework secureboot and PCR signing key

smira:
Support different providers, not only static file paths.

Drop `pcr-signing-key-public.pem` file, as we generate it on the fly now.

See https://github.com/siderolabs/image-factory/issues/19


smira:
This PR should be mostly no-op for Talos itself, but it provides a way to generate assets signed with Azure Key Vault, which will be consumed in the Image Factory.

go.mod
```diff
@@ -146,12 +149,20 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
```
frezbo, line 152: can we group this with others?

internal/pkg/secureboot/database/database.go
```diff
@@ -0,0 +1,62 @@
...
+	Name     string
+	Contents []byte
+}
+
+// Generate generates a UEFI database to enroll the signing certificate.
+//
+// ref: https://blog.hansenpartnership.com/the-meaning-of-all-the-uefi-keys/
+func Generate(enrolledCertificate []byte, signer pesign.CertificateSigner) ([]Entry, error) {
+	// derive UUID from enrolled certificate
+	uuid := uuid.NewHash(sha256.New(), uuid.NameSpaceX500, enrolledCertificate, 4)
```
frezbo, line 31: :cool:

pkg/imager/out.go
```diff
@@ -94,7 +98,56 @@ func (i *Imager) outISO(path string, report *reporter.Reporter) error {
 
 			ScratchDir: scratchSpace,
 			OutPath:    path,
-		})
+		}
+
+		if i.prof.Input.SecureBoot.PlatformKeyPath == "" {
```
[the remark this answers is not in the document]
smira: yes, if we were not given secure database keys explicitly, we generate them automatically on build

frezbo (approved):

PR #7931, opened 2023-11-07
refactor: update packet capture to use 'afpacket' interface

### LINKED ISSUE #7880: reimplement packet capture in machined

smira:
First of all, this interface is way more performant than `pcap` interface. It is Linux-specific, but we don't care in Talos Linux :)

Second, this drop dependency of `machined` on `gopacket/layers` package, which has huge issues with memory allocations and startup time.

This cuts around 20MiB of process RSS for all Talos processes. (`talosctl` still requires this `gopacket/layers` library for decoding packets).

Fixes #7880

### DIFF
--- a/internal/pkg/pcap/pcap.go
+++ b/internal/pkg/pcap/pcap.go
@@ -1,0 +1,109 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package pcap implements writing packet data to pcap files.
+package pcap
+
+import (
+	"encoding/binary"
+	"fmt"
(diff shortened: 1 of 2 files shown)


internal/pkg/pcap/pcap.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// Writer wraps an underlying io.Writer to write packet data in PCAP
+// format.  See http://wiki.wireshark.org/Development/LibpcapFileFormat
+// for information on the file format.
+//
+// For those that care, we currently write v2.4 files with nanosecond
+// or microsecond timestamp resolution and little-endian encoding.
+type Writer struct {
+	w   io.Writer
```
smira, line 24: that's stripped down version from `gopacket` which has no dependency on `gopacket/layers`

frezbo (approved):

PR #7932, opened 2023-11-07
fix: image build for nanopi_4s

smira:
Path was missing a slash.


frezbo (approved):
`filepath.Join` everywhere :sweat_smile:

DmitriyMV (approved):

DmitriyMV:
> filepath.Join everywhere 😅

No compile time function 😫

PR #7933, opened 2023-11-07
fix: proper logging in machined on startup

smira:
Move `setupLogging` inside the controller, so that logger is set up correctly before Talos starts printing first messages.

This fixes an inconsistency that first messages are printed using "default" logger, while after that the proper logger is set up, and format of the messages matches kernel log.

Also move `waitForUSBDelay` into the sequencer after `udevd` was started (this is when blockdevices including USB ones are discovered).

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -81,41 +79,45 @@ import (
...
+// WaitForUSB represents the WaitForUSB task.
+func WaitForUSB(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) (err error) {
-		machinedLog, err := r.Logging().ServiceLog("machined").Writer()
+		// Wait for USB storage in the case that the install disk is supplied over
+		// USB. If we don't wait, there is the chance that we will fail to detect the
+		// install disk.
+		file := "/sys/module/usb_storage/parameters/delay_use"
+
+		_, err = os.Stat(file)
```
DmitriyMV, line 90: Minor nit: if you do `:=` here and below linter probably will not complain since you do not have any inner `err`s.

DmitriyMV (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: yep, this is copy-pasted... I will drop named return args

frezbo (approved):

PR #7934, opened 2023-11-07
chore: drop sha1 from genereated pcr json

frezbo:
Drop `sha1` algorithm from expected PCR json calculation.


smira (approved):

smira:
```
internal/pkg/secureboot/measure/measure_test.go:24: line is 2623 characters (lll)
```

Issue #7694, opened 2023-09-01
allow overriding SMBIOS UUID via META


smira:
Quite often bare-metal machines comes with invalid UUIDs (all zeroes, or duplicate values).

Omni and Sidero Metal use UUIDs to identify machines, and duplicate UUIDs break this scheme.

Allow overriding the UUID by writing to an (optional) `META` field which will replace the UUID in all subsequent UUID-based operations.

Example flow:

* machine boots with UUID of all zeroes
* machine establishes SideroLink connection
* Omni (or other orchestrator) discovers invalid UUID, and using Talos API writes META value with new UUID
* machine reconnects with new UUID

Notes:

* if Talos is not installed yet, META is preserved in memory, and will be flushed on next install

Steps:

* allow overriding via META (in SystemInformation resource)
* check that all places use UUID coming from SystemInformation resource
* allow META APIs in maintenance mode (only via SideroLink)

DmitriyMV:
Closed in https://github.com/siderolabs/talos/pull/7935

PR #7935, opened 2023-11-07
chore: add ability to rewrite uuids and set unique tokens for Talos

DmitriyMV:
This PR does those things:
- It allows API calls `MetaWrite` and `MetaRead` in maintenance mode.
- SystemInformation resource now waits for available META
- SystemInformation resource now overwrites UUID from META if there is an override
- META now supports "UUID override" and "unique token" keys
- ProvisionRequest now includes unique token and Talos version

For #7694


internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -58,59 +74,83 @@ func (ctrl *SystemInfoController) Outputs() []controller.Output {
...
+			if state.IsNotFoundError(err) {
+				continue
+			}
+
+			return fmt.Errorf("error getting meta loaded resource: %w", err)
 		}
 
-		ctrl.SMBIOS = s
-	}
+		// smbios info is not available inside container, so skip the controller
```
smira, line 93: we could probably keep it at the top of the function, as in container mode we skip it completely...

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -90,6 +93,12 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 				ID:        optional.Some(hardware.SystemInformationID),
 				Kind:      controller.InputWeak,
 			},
+			{
```
smira, line 96: this one is a bit tricky, as the controller restarts the SideroLink provisioning on each resource change, it might happen that it will be triggered for `MetaKey` update for `UniqueToken` while it previously already tried to connect without it.

I think the only meaningful way to handle that unfortunately is to have another intermediate resource which is managed by a controller which waits for MetaLoaded, and once Loaded either creates an empty intermediate resource with unique token, or fills it with MetaKey value.

internal/pkg/meta/constants.go
```diff
@@ -23,4 +23,8 @@ const (
 	UserReserved2
 	// UserReserved3 is reserved for user-defined metadata.
 	UserReserved3
+	// UUIDRewrite stores the UUID that this machine will use instead of the one from the hardware.
+	UUIDRewrite
```
smira, line 27:
```suggestion
	UUIDOverride
```

might be a better name?

internal/pkg/meta/constants.go
```diff
@@ -23,4 +23,8 @@ const (
...
+	// UniqueToken store the unique token for this machine. It's useful because UUID may repeat or be filled with zeros.
+	UniqueToken
```
smira, line 29:
```suggestion
	UniqueMachineToken
```

might be a bit more clear what kind of token that is

smira (commented):
lgtm, with some smaller stuff

internal/app/machined/pkg/controllers/siderolink/manager.go
DmitriyMV: Can't we pass the state to the controller as we do with some of other controllers? That way we will not go for the `watch for input` behavior that we do not need here.

Another options are:
1. Add `Get` option that allows to bypass controller `Input`s.
2. Add special `Input` type that allows to read from namespace and type, but doesn't subscribe to it.

smira: neither of that helps here, I think. it's the logic: wait for META loaded, if machine token present, use it; if it changes, re-establish connection. It's pretty normal flow, there is no need for hacks, but we need to prepare proper Input for this controller in another controller, which will be better (separation of concerns)

DmitriyMV: Added new resource and controller.

internal/app/machined/pkg/controllers/runtime/unique_token.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/optional"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/talos/internal/pkg/meta"
+	"github.com/siderolabs/talos/pkg/machinery/resources/runtime"
+)
+
+// UniqueMachineTokenController provides a unique token the machine.
+type UniqueMachineTokenController = transform.Controller[*runtime.MetaLoaded, *runtime.UniqueMachineToken]
```
smira, line 22: clever trick!

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -198,10 +206,22 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 				}
 			}()
 
+			uniqTokenRes, rdrErr := safe.ReaderGetByID[*runtime.UniqueMachineToken](ctx, r, runtime.UniqueMachineTokenID)
+			if rdrErr != nil && !state.IsNotFoundError(rdrErr) {
+				return nil, fmt.Errorf("failed to get unique token: %w", rdrErr)
```
smira, line 211: we should skip here if the resource is not found, i.e. controller should only start when the resource is present?

internal/app/machined/pkg/controllers/runtime/unique_token.go
```diff
@@ -0,0 +1,54 @@
...
+	return transform.NewController(
+		transform.Settings[*runtime.MetaLoaded, *runtime.UniqueMachineToken]{
+			Name: "runtime.UniqueMachineTokenController",
+			MapMetadataFunc: func(in *runtime.MetaLoaded) *runtime.UniqueMachineToken {
+				return runtime.NewUniqueMachineToken()
+			},
+			TransformFunc: func(ctx context.Context, r controller.Reader, logger *zap.Logger, _ *runtime.MetaLoaded, out *runtime.UniqueMachineToken) error {
+				uniqueToken, err := safe.ReaderGetByID[*runtime.MetaKey](ctx, r, runtime.MetaKeyTagToID(meta.UniqueMachineToken))
+				if state.IsNotFoundError(err) {
+					return nil
```
smira, line 35: should clear the token here

```go
out.TypedSpec().Token = ""
```

smira (approved):

internal/app/machined/pkg/controllers/siderolink/manager.go
DmitriyMV: It's a leftover from the previous iteration. Thanks for noticing, I also removed the code below which worked with assumption that uniqueToken resource may not be present.

internal/app/machined/pkg/controllers/runtime/unique_token.go
DmitriyMV: Done, thanks!

PR #7940, opened 2023-11-08
feat: Linux 6.1.61, containerd 1.7.8, runc 1.1.10

smira:
Bump tools/pkgs/extras.

Update Go dependencies.


frezbo (approved):

smira:
There's something with new `pcap` and integration tests. I don't think it's related to this PR, but still need to understand it

Issue #7911, opened 2023-10-31
Supporting extra config files for k8s components


frezbo:
## Feature Request

Kubelet, API server, Kube scheduler etc require extra config files to be passed in as arguments for advanced tweaking. Currently Talos lacks an easy way to configure them. See:

* https://github.com/siderolabs/extensions/issues/241
* https://github.com/siderolabs/talos/issues/7905

### Proposal

Support something in the order of `extraConfigFiles` for the components. Eg:

```
machine:
  kubelet:
    extraConfigFiles:
      - name: credentialproviderconfig.yaml
        value:
          apiVersion: kubelet.config.k8s.io/v1
          kind: CredentialProviderConfig
```

The `value` can be any valid yaml and Talos will not validate the schema.

Talos can write down these files to `/etc/kubernetes/<component-name>.d/` directory and fully managed by the configuration in the machine config.

These folders would be mounted by default by Talos to the corresponding components. The user would still have to pass in `extraArgs/extraConfig` to use these files.

smira:
I think we should avoid `/etc/kubernetes` as it is backed by `/var` via overlay mount... much better place will be under `/system` with other config files we already generate. We can still mount it into the container as `/etc/kubernetes/<something>`

frezbo:
> I think we should avoid `/etc/kubernetes` as it is backed by `/var` via overlay mount... much better place will be under `/system` with other config files we already generate. We can still mount it into the container as `/etc/kubernetes/<something>`

`kubelet.conf` is already under `/var` right? and since we deterministically generate those, shouldn't it be okay?

smira:
I don't like the fact that we leave some files in `/var` for no reason. `kubelet.conf` is there for historical reasons, plus kubelet keeps its own state there. But we shouldn't put new stuff there if there's no reason to do so.

We don't need these files to persist reboots, we don't even need `/var` for them at all, so why put them there, if we could better handle them in a `tmpfs` location we own completely?

frezbo:
> We don't need these files to persist reboots, we don't even need /var for them at all, so why put them there, if we could better handle them in a tmpfs location we own completely?

that makes sense :+1:

PR #7942, opened 2023-11-09
feat: support kubelet credentialprovider config

frezbo:
Support configuring kubelet credential provider config.

Partially fixes: #7911


Dockerfile
```diff
@@ -568,7 +568,7 @@ COPY ./hack/cleanup.sh /toolchain/bin/cleanup.sh
 RUN <<END
     cleanup.sh /rootfs
     mkdir -pv /rootfs/{boot/EFI,etc/cri/conf.d/hosts,lib/firmware,usr/local/share,usr/share/zoneinfo/Etc,mnt,system,opt,.extra}
-    mkdir -pv /rootfs/{etc/kubernetes/manifests,etc/cni/net.d,usr/libexec/kubernetes}
+    mkdir -pv /rootfs/{etc/kubernetes/manifests,etc/cni/net.d,usr/libexec/kubernetes,/usr/local/lib/kubelet/credentialproviders}
```
smira, line 571: same for `arm64`?

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
```diff
@@ -166,6 +166,12 @@ func (ctrl *KubeletSpecController) Run(ctx context.Context, r controller.Runtime
 			return fmt.Errorf("error merging arguments: %w", err)
 		}
 
+		// these flags are present till v1.24
```
smira, line 169: from 1.24?

Dockerfile
frezbo: yes, missed it

internal/app/machined/pkg/controllers/k8s/kubelet_spec.go
frezbo: ah yes

smira (approved):

PR #7953, opened 2023-11-13
fix: allow rootfs propagation configuration for extension services

### LINKED ISSUE #7873: Mount Propogation and System Extensions
It feels like there's a bug somewhere with system extensions and mounts as far as I can tell.

The stargz extension has /var and a whole slew of other locations mounted up with rshared but propagation doesn't seem to work. Maybe there's some difference in the way we do mounts vs. how K8s does it or some difference in the two containerd configurations?

Andrew and I popped busybox into the stargz extension container, exec'ed into the container's namespace, and did some hacking around.

Here's how to reproduce it:

- Create a talos machine with these two extensions: `docker.io/rsmitty/fuse3:3.16.1-v1.6.0-alpha.0-19-g9b5c56f-dirty` and `docker.io/rsmitty/stargz-snapshotter:v0.14.3-v1.6.0-alpha.0-19-g9b5c56f-dirty`
- Once up and running, find the stargz extension pid with `talosctl containers`
- Create a debug pod. I used a hacked version of `https://github.com/siderolabs/operations/blob/master/apps/debug.yaml` that does a bidirectional mount of /var (posted below).
- Exec into the debug pod and issue `nsenter -t <stargz-extension-pid> -a /usr/local/bin/busybox sh`
- Make a `/var/hack` directory for mounting (`busybox mkdir /var/hack`)
- Created a mount from the Talos ISO with `busybox mount /dev/sr0 /var/hack`
- The mount shows in the container as `/dev/sr0 on /var/hack type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8)`
- The mount **does not** show on the host with `talosctl mounts` (or any of the /proc mount files, etc.)
- Exit busybox with a quick `exit`, this puts you back into the debug container.
- Create a second mount location with `mkdir /rootfs/var/hack-two` (path of var is different in debug mount).
- Issue `mount /dev/sr0 /rootfs/var/hack-two`.
- Mount shows up as expected in `taloctl mounts` or using mountinfo:

```
❯ talosctl -n 192.168.1.111 cat /proc/self/mountinfo | grep hack
763 49 11:0 / /var/hack-two ro,relatime shared:129 - iso9660 /dev/sr0 ro,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8
```

Here's the debug pod I used:

```
apiVersion: v1
kind: Pod
metadata:
  name: workspace
spec:
  hostPID: true
  hostIPC: true
  hostNetwork: true
  containers:
    - name: workspace
      image: ubuntu:latest
      command: ['/bin/sh', '-c', '--']
      args: ['trap : TERM INT; (while true; do sleep 1000; done) & wait']
      securityContext:
        privileged: true
      env:
        - name: CONTAINER_RUNTIME_ENDPOINT
          value: /rootfs/run/containerd/containerd.sock
      volumeMounts:
        - name: dev
          mountPath: /dev
          mountPropagation: Bidirectional
        - name: rootfsvar
          mountPath: /rootfs/var
          mountPropagation: Bidirectional
  volumes:
    - name: dev
      hostPath:
        path: /dev
        type: Directory
    - name: rootfsvar
      hostPath:
        path: /var
        type: Directory
```

smira:
Fixes #7873

Some services which perform mounts inside the container which require mounts to propagate back to the host (e.g. `stargz-snapshotter`) require this configuration setting.

### DIFF
--- a/internal/app/machined/pkg/system/services/extension.go
+++ b/internal/app/machined/pkg/system/services/extension.go
@@ -106,6 +106,7 @@ func (svc *Extension) DependsOn(r runtime.Runtime) []string {
 func (svc *Extension) getOCIOptions(envVars []string) []oci.SpecOpts {
 	ociOpts := []oci.SpecOpts{
 		oci.WithRootFSPath(filepath.Join(constants.ExtensionServicesRootfsPath, svc.Spec.Name)),
+		containerd.WithRootfsPropagation(svc.Spec.Container.Security.RootfsPropagation),
 		oci.WithCgroup(constants.CgroupExtensions),
 		oci.WithMounts(svc.Spec.Container.Mounts),
 		oci.WithHostNamespace(specs.NetworkNamespace),
--- a/internal/integration/api/extensions_qemu.go
+++ b/internal/integration/api/extensions_qemu.go
@@ -392,6 +392,30 @@ func (suite *ExtensionsSuiteQEMU) TestExtensionsGvisor() {
 	suite.Require().NoError(suite.WaitForPodToBeRunning(suite.ctx, 5*time.Minute, "default", "nginx-gvisor"))
 }
 
+// TestExtensionsStargz verifies stargz snapshotter.
+func (suite *ExtensionsSuiteQEMU) TestExtensionsStargz() {
+	_, err := suite.Clientset.CoreV1().Pods("default").Create(suite.ctx, &corev1.Pod{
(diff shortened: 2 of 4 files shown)


frezbo (approved):

smira:
I got a very simple test added here for integration-extensions

frezbo (approved):
that test should do

Issue #7083, opened 2023-04-13
Enable service extensions before machine config is applied


casdr:
## Feature Request

We're running Talos clusters within Xenserver Virtual Machines. Xenserver has the [xe-guest-utilities](https://github.com/xenserver/xe-guest-utilities) daemon which provides information about the OS and network interfaces and enables OS power control.

To setup our clusters and apply the machine config automatically, we need to grab the IP-addresses that have been assigned from the Xenserver API. For this information to be available, we need this daemon to be running.

I've created a system extension with a service container which runs this daemon and created a metal image including the extension. The only problem we're encountering is that the container starts after applying the machine config. However, to apply the machine config, we need the information provided by the daemon.

As you might notice, we've stumbled upon a circular dependency issue here :-)

It would be nice to be able to add a service extension that starts before the machine config is applied.

TheiLLeniumStudios:
@casdr is the extension for the daemon open source by any chance? :)

ScottKane:
We are experiencing the same issue while trying to provision a Talos cluster on Proxmox using the Proxmox Terraform provider which will feed into the Talos Terraform provider. Would be great if we can get some love on this issue to possibly enable a flag during the custom image build to allow extensions to start while in maintenance mode.

casdr:
Awesome, thank you @frezbo!

PR #7957, opened 2023-11-16
chore: start containerd early in boot

frezbo:
Start container early in the boot process so system extension services start in maintenance mode.

Fixes: #7083


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -105,6 +105,7 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 			"earlyServices",
 			StartUdevd,
 			StartMachined,
+			StartContainerd,
```
smira, line 108: also need to start it in contianerd mode above

Issue #4421, opened 2021-10-22
machine configuration to manage firewall


smira:
See https://github.com/sbezverk/nftableslib

First goal: define minimal but usable rule configuration we could put to the machine configuration. Format should be extensible in the future if we want to expand later.

- only ingress traffic
- if the port is defined in the rule, block everything which is not listed in `ingress:` section
- if the port is not defined in any rule, `NetworkDefaultAction` is applied

```yaml
kind: NetworkRuleConfig
portSelector:
  ports: [6443, 10250]
  protocol: tcp
ingress:
  - cidr: 10.0.0.0/8
  - cidr: 2000:fd00::/16
---
kind: NetworkRuleConfig
portSelector:
  ports: [80, 1000-2000]
  protocol: tcp
ingress:
  - cidr: 192.168.0.0/16
    except: 192.168.64.0/24
---
kind: NetworkDefaultActionConfig
ingress: allow/block
```

reyvonger:
I will be very happy to implement this feature

PR #7961, opened 2023-11-17
feat: implement nftables backend

smira:
Implement initial set of backend controllers/resources to handle nftables chains/rules etc.

Replace the KubeSpan nftables operations with controller-based.

See #4421


pkg/machinery/nethelpers/match_operator.go
```diff
@@ -0,0 +1,18 @@
...
+//go:generate enumer -type=MatchOperator -linecomment -text
+
+// MatchOperator is a netfilter match operator.
+type MatchOperator int
+
+// MatchOperator constants.
+//
+//structprotogen:gen_enum
+const (
+	OperatorEqual    MatchOperator = 0 // ==
```
frezbo, line 16: maybe add a comment that this matches the nftables library

pkg/machinery/nethelpers/nftables_chain_hook.go
```diff
@@ -0,0 +1,21 @@
...
+//go:generate enumer -type=NfTablesChainHook -linecomment -text
+
+// NfTablesChainHook wraps nftables.ChainHook for YAML marshaling.
+type NfTablesChainHook uint32
+
+// Constants copied from nftables to provide Stringer interface.
+//
+//structprotogen:gen_enum
+const (
+	ChainHookPrerouting  NfTablesChainHook = 0 // prerouting
```
frezbo, line 16: same here, comment helps

pkg/machinery/nethelpers/match_operator.go
smira: I'd love to, but this will break enumer

internal/app/machined/pkg/adapters/network/nftables_rule.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+
+	clampMSS := func(family nftables.TableFamily, mtu uint16) []expr.Any {
+		var mss uint16
+
+		switch family { //nolint:exhaustive
+		case nftables.TableFamilyIPv4:
+			mss = mtu - 40 // TCP + IPv4 overhead
+		case nftables.TableFamilyIPv6:
+			mss = mtu - 60 // TCP + IPv6 overhead
```
frezbo, line 376: would this change if wireguard is in use?

smira: so the MTU passed here via KubeSpan includes Wireguard overhead. In general this rule will clamp MSS to fit MTU

hack/structprotogen/consts/consts.go
```diff
@@ -237,7 +238,9 @@ func (b *ConstBlocks) FormatProtoFile(w io.Writer) error {
 
 		fmt.Fprintf(w, "enum %s {\n", block.ProtoMessageName())
 
-		if hasDuplicates(block.Consts, func(c Constant) string { return c.Value }) {
+		hasZeroNotFirstConstValue := slices.IndexFunc(block.Consts, func(c Constant) bool { return c.Value == "0" }) > 0
```
smira, line 241: this fixes a case when enum is like `{-100, 0, 100}`, so it has zero, but not as the first value

internal/app/machined/pkg/adapters/network/nftables_rule.go
```diff
@@ -0,0 +1,551 @@
...
+			Register: 1,
+			Data:     []byte{byte(nftables.TableFamilyIPv6)},
+		},
+	}
+)
+
+// Compile translates the rule into the set of nftables instructions.
+//
+//nolint:gocyclo,cyclop
+func (a nftablesRule) Compile() (*NfTablesCompiled, error) {
```
smira, line 153: this is the core compiler of the resource spec for ntftables rule down to nftables instructions

internal/app/machined/pkg/controllers/kubespan/manager.go
```diff
@@ -154,16 +152,6 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 		}
 	}()
 
-	var nfTablesMgr NfTablesManager
```
smira, line 157: not using specialized nftables backend here, but using generic nftables manager

internal/app/machined/pkg/controllers/network/nftables_chain.go
```diff
@@ -0,0 +1,174 @@
...
+		if err != nil {
+			return fmt.Errorf("error listing existing nftables chains: %w", err)
+		}
+
+		for _, chain := range existingChains {
+			if chain.Table.Name != ctrl.TableName { // not our chain
+				continue
+			}
+
+			conn.DelChain(chain)
```
frezbo, line 104: this returns no error?

frezbo (approved):
:cool:

internal/app/machined/pkg/controllers/network/nftables_chain.go
smira: yep, it just queues the message

Unix4ever (approved):

Issue #7613, opened 2023-08-14
Ability to remove image-default kernel args


vladimirfx:
## Feature Request

Please add 'removeKernelArgs' setting to install config which removes specified arg from image default kernel args. Args to remove should match exactly. 

### Description

Sometimes default kernel args can break boot. See https://github.com/siderolabs/talos/discussions/7465

smira:
Probably going to be replaced by dropping console args in Talos 1.6 for `metal` platform

vladimirfx:
> Probably going to be replaced by dropping console args in Talos 1.6 for `metal` platform

Yes, it is fixed our case. But this issue is more generic - it balances kernel args modification functionality. It up to you how to proceed 👍

smira:
One idea: allow `-argument` and `-argument=value` notations, the first removes all `argument` values, and the second one removes just one value. Support it throughout the kernel arg handling/merging code.

vladimirfx:
Thanks

PR #7981, opened 2023-11-23
feat: support sanitized kernel args

frezbo:
Support dropping kernel args that start with `-`.

Fixes: #7613

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -101,6 +101,7 @@ func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options) 
 		opts.ExtraKernelArgs,
 		procfs.WithOverwriteArgs("console"),
 		procfs.WithOverwriteArgs(constants.KernelParamPlatform),
+		procfs.WithDeleteNegatedArgs(),
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


cmd/installer/pkg/install/errata.go
```diff
@@ -101,3 +103,28 @@ func upgradeFromPreIfnamesTalos() (bool, error) {
...
+
+// SanitizeKernelArgs removes kernel arguments that are not needed for the installer.
+func SanitizeKernelArgs(cmdline []string) []string {
+	kernelArgsToDrop := xslices.Filter(cmdline, func(s string) bool {
+		// it's valid to have a kernel arg just as `-`.
+		// Ref: https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html
+		return len(s) > 1 && s[0] == '-'
+	})
+
+	return xslices.Filter(cmdline, func(s string) bool {
```
smira, line 115: would it make more sense to put this into the `cmdline.AppendAll` code?

frezbo: yeh, that's fine, I was not sure if we should do that there

smira (approved):
Can you please add an example to the Boot Assets documentation?

And probably machine config extra kernel args?

PR #7987, opened 2023-11-24
chore: add custom node taints

### LINKED ISSUE #7581: Support for Specifying Custom Node Taints via Machine Configuration
**Description**:

In the current implementation of Talos, there is no native support for specifying custom node taints directly through the machine configuration. While control plane nodes are automatically tainted, other nodes lack this configurability.

**Problem Statement:**

Node taints are a critical part of Kubernetes cluster management, enabling operators to control which nodes specific pods can be scheduled on. The lack of support for custom taints within Talos's machine configuration makes it difficult to manage taints at scale and can lead to inefficiencies and inconsistencies across the cluster.

**Current Workaround:**

At present, custom taints can be applied using the kubectl command after the node is installed. While this method does provide a way to apply taints, it introduces extra manual steps and can lead to inconsistencies in the configuration of nodes within a cluster. Automating this process within Talos would greatly streamline cluster configuration, ensuring that taints are applied uniformly and without requiring individual manual intervention.

**Proposed Solution:**

Introduce native support in Talos for specifying custom node taints within the machine configuration. This would allow taints to be defined as part of the machine's initial setup, ensuring a consistent configuration across all nodes in the cluster. The solution would provide cluster administrators with a more efficient and robust way to manage node taints, aligning with common Kubernetes best practices.

DmitriyMV:
This PR adds support for custom node taints. Refer to `nodeTaints` in the `configuration` for more information.

Closes #7581

### DIFF
--- a/pkg/machinery/labels/validate.go
+++ b/pkg/machinery/labels/validate.go
@@ -10,11 +10,13 @@ package labels
 import (
 	"fmt"
 	"regexp"
-	"sort"
+	"slices"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/siderolabs/gen/maps"
+
+	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
 // Validate validates that a set of labels are correctly defined.
@@ -22,7 +24,7 @@ func Validate(labels map[string]string) error {
 	var multiErr *multierror.Error
 
 	keys := maps.Keys(labels)
-	sort.Strings(keys)
+	slices.Sort(keys)
 
 	for _, k := range keys {
 		if err := ValidateQualifiedName(k); err != nil {
@@ -129,4 +131,42 @@ func ValidateLabelValue(value string) error {
 	}
 
 	return nil
+}
+
+// ValidateTaints validates that a set of taints is correctly defined.
+func ValidateTaints(taints map[string]string) error {
+	var multiErr *multierror.Error
+
+	keys := maps.Keys(taints)
+	slices.Sort(keys)
+
+	for _, k := range keys {
+		if err := ValidateQualifiedName(k); err != nil {
+			multiErr = multierror.Append(multiErr, err)
+
+			continue
+		}
+
+		val, effect, found := strings.Cut(taints[k], ":")
+		if !found {
+			effect = val
+		}
+
+		// validate that the taint has a valid effect, which is required to add the taint
+		if !slices.Contains(constants.ValidEffects, effect) {
+			multiErr = multierror.Append(multiErr, fmt.Errorf("invalid taint effect: %q", effect))
+
+			continue
+		}
+
+		if found {
+			if err := ValidateLabelValue(val); err != nil {
+				multiErr = multierror.Append(multiErr, err)
+
+				continue
+			}
+		}
+	}
+
+	return multiErr.ErrorOrNil()
 }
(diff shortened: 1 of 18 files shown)


pkg/machinery/labels/validate.go
```diff
@@ -130,3 +130,28 @@ func ValidateLabelValue(value string) error {
 
 	return nil
 }
+
+// ValidateTaints validates that a set of taints is correctly defined.
+func ValidateTaints(taints map[string]string) error {
```
DmitriyMV, line 135: Is this correct way of validating taints?

DmitriyMV:
There is also https://github.com/siderolabs/talos/blob/dbaf5c69978fd1d22737385ddd096798d408254c/internal/integration/api/node-labels.go#L85-L90 integration test. Should I update it to handle the taints too?

pkg/machinery/labels/validate.go
```diff
@@ -130,3 +130,28 @@ func ValidateLabelValue(value string) error {
...
+			multiErr = multierror.Append(multiErr, fmt.Errorf("invalid taint effect: %s", effect))
+		}
+
+		if err := ValidateLabelValue(val); err != nil {
+			multiErr = multierror.Append(multiErr, err)
+		}
+	}
+
+	return multiErr.ErrorOrNil()
+}
```
DmitriyMV, line 157: Is this correct way of validating taints?

smira:
> integration test. Should I update it to handle the taints too?

yes, absolutely, probably another test which will be very similar

pkg/machinery/labels/validate.go
smira: you can check with either k8s docs or their code (we can't use it in machinery, but you can get the idea there)

smira (commented):
lgtm, just need the integration test!

pkg/machinery/labels/validate.go
```diff
@@ -130,3 +130,28 @@ func ValidateLabelValue(value string) error {
...
+	keys := maps.Keys(taints)
+	slices.Sort(keys)
+
+	for _, k := range keys {
+		if err := ValidateQualifiedName(k); err != nil {
+			multiErr = multierror.Append(multiErr, err)
+		}
+
+		val, effect, found := strings.Cut(taints[k], ":")
+		if found && effect != "NoSchedule" && effect != "PreferNoSchedule" && effect != "NoExecute" {
```
frezbo, line 147: doesn't k8s supports taints other than this :thinking: ?

frezbo: cool, maybe we keep this in `constants`? and then filter on a slice?

frezbo (approved):

pkg/machinery/labels/validate.go
DmitriyMV: I dunno, both works for me.

DmitriyMV: Updated the code.

pkg/machinery/labels/validate.go
DmitriyMV: Yup, it uses the same logic https://github.com/kubernetes/kubectl/blob/b35935138f5330d299e89b1278b5738487e4f015/pkg/cmd/taint/utils.go#L39

frezbo (approved):

PR #7989, opened 2023-11-24
chore: support getting multiple endpoints from the `Provision` rpc call

### LINKED ISSUE #7973: allow multiple SideroLink endpoints
Allow multiple Wireguard endpoints in response to `Provision()` request.

DmitriyMV:
Closes #7973


internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -246,6 +246,13 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 		linkSpec := network.NewLinkSpec(network.ConfigNamespaceName, network.LayeredID(network.ConfigOperator, network.LinkID(constants.SideroLinkName)))
 		addressSpec := network.NewAddressSpec(network.ConfigNamespaceName, network.LayeredID(network.ConfigOperator, network.AddressID(constants.SideroLinkName, nodeAddress)))
 
+		endpoints := resp.GetEndpoints()
```
smira, line 249: this is not that easy... we should support cycling through the endpoints if the connection is down (today we just do full reconnect, but we should try all endpoints first)

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -146,106 +148,129 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
...
+		if len(endpoints) == 0 {
+			return errors.New("host returned no endpoints")
+		}
+
+		// Rotate through the endpoints.
+		ep := endpoints[ctrl.endpointIdx]
+		ctrl.endpointIdx = (ctrl.endpointIdx + 1) % len(endpoints)
+
+		logger.Info(
+			"configurating siderolink connection",
```
smira, line 268:
```suggestion
			"configuring siderolink connection",
```

internal/app/machined/pkg/controllers/siderolink/manager.go
DmitriyMV: Hmm, I tried approaching this by storing the endpoints response in the controller and rotating through them until we reach the 0 idx, in which case we try the provisioning again.

DmitriyMV:
The code is now readable :)

smira (approved):

PR #7999, opened 2023-11-28
feat: implement ingress firewall rules

### LINKED ISSUE #4421: machine configuration to manage firewall

smira:
Fixes #4421

See documentation for details on how to use the feature.

With `talosctl cluster create`, firewall can be easily test with
`--with-firewall=accept|block` (default mode).


Dockerfile
```diff
@@ -292,7 +292,7 @@ COPY --from=generate-build /api/inspect/*.pb.go /pkg/machinery/api/inspect/
 COPY --from=go-generate /src/pkg/flannel/ /pkg/flannel/
 COPY --from=go-generate /src/pkg/imager/profile/ /pkg/imager/profile/
 COPY --from=go-generate /src/pkg/machinery/resources/ /pkg/machinery/resources/
-COPY --from=go-generate /src/pkg/machinery/config/types/v1alpha1/ /pkg/machinery/config/types/v1alpha1/
+COPY --from=go-generate /src/pkg/machinery/config/types/ /pkg/machinery/config/types/
```
frezbo, line 295: ahh I was going to say our generate was broken, since it didn't generate for the extension config ones I was working on

internal/app/machined/pkg/adapters/network/nftables_rule.go
```diff
@@ -91,7 +91,7 @@ func (set NfTablesSet) SetElements() []nftables.SetElement {
 
 		for _, p := range set.Ports {
 			from := binaryutil.BigEndian.PutUint16(p[0])
-			to := binaryutil.BigEndian.PutUint16(p[1])
+			to := binaryutil.BigEndian.PutUint16(p[1] + 1)
```
frezbo, line 94: was this a bug?

pkg/machinery/config/types/network/port_range_test.go
```diff
@@ -0,0 +1,124 @@
...
+				name: "inversion",
+				prs:  network.PortRanges{{Lo: 8081, Hi: 8080}},
+
+				expectedError: "invalid port range: 8081-8080",
+			},
+			{
+				name: "overlap",
+				prs:  network.PortRanges{{Lo: 1000, Hi: 2000}, {Lo: 80, Hi: 80}, {Lo: 1500, Hi: 2500}},
+
+				expectedError: "invalid port range: 1500-2500",
```
frezbo, line 105: maybe a better error message for overlap and duplicate would help

Dockerfile
smira: No idea how I did previous configs ))

internal/app/machined/pkg/adapters/network/nftables_rule.go
smira: yes, I will separate it out from the machine config changes when I finalize it.

It was even visible in the test lol

pkg/machinery/config/types/network/port_range_test.go
smira: yes, good point

internal/app/machined/pkg/controllers/network/nftables_chain_config.go
```diff
@@ -0,0 +1,201 @@
...
+	"go.uber.org/zap"
+
+	v1alpha1runtime "github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+	"github.com/siderolabs/talos/pkg/machinery/nethelpers"
+	"github.com/siderolabs/talos/pkg/machinery/resources/config"
+	"github.com/siderolabs/talos/pkg/machinery/resources/network"
+)
+
+// NfTablesChainConfigController generates configuration for kmsg log delivery.
```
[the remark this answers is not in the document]
smira: it was me being dumb!

frezbo (approved):
:cool: :dancers:

smira:
> don't we need some tests to verify the access is blocked?

hmm... good point... I can do that

smira:
> don't we need some tests to verify the access is blocked?

added a test for kubelet API availability

PR #8003, opened 2023-11-29
feat: Kubernetes 1.29.0-rc.1 and other bumps

smira:
Bump Go modules, final tools and semi-final pkgs.


go.mod
```diff
@@ -7,7 +7,7 @@ replace (
 	github.com/siderolabs/talos/pkg/machinery => ./pkg/machinery
 
 	// github.com/cilium/ebpf@v0.11.0/internal/unix/types_linux.go:41:36: undefined: linux.BPF_F_KPROBE_MULTI_RETURN
-	golang.org/x/sys => golang.org/x/sys v0.13.0
+	// golang.org/x/sys => golang.org/x/sys v0.13.0
```
frezbo, line 10: remove?

frezbo (approved):

Unix4ever (approved):

go.mod
smira: oh yeah, I found out that it's not needed, but forgot to remove it completely, thanks!

PR #8006, opened 2023-11-29
fix: support slices of enums and fix NfTablesConntrackStateMatch

DmitriyMV:
We already have the code which supports custom enums, so let's extend it to support custom enums in slices and fix the NfTablesConntrackStateMatch proto definition.


smira (approved):
❤️

frezbo (approved):

PR #8024, opened 2023-12-04
fix: trim leading and trailing spaces\newlines in inline manifest contents

### LINKED ISSUE #7993: initialManifests not working as expected
i get mapping errors trying to install manifests on bootstrap

patch file:
```yaml
```

```yaml
```

DmitriyMV:
In route `LoadPatches` -> `configpatcher.Apply` -> `configloader.NewFromBytes` any leading newlines will be transformed  into `|4` yaml. We want to prevent that.

Closes #7993


smira:
Is there any simple regression test we could add?

PR #8026, opened 2023-12-04
chore: enable kubespan+firewall for cilium tests

frezbo:
Enable kubespan and default block firewall with cilium tests.

### DIFF
--- a/.drone.jsonnet
+++ b/.drone.jsonnet
@@ -477,6 +477,7 @@ local integration_cilium = Step('e2e-cilium', target='e2e-qemu', privileged=true
   SHORT_INTEGRATION_TEST: 'yes',
   WITH_SKIP_BOOT_PHASE_FINISHED_CHECK: 'yes',
   WITH_CUSTOM_CNI: 'cilium',
+  WITH_FIREWALL: 'accept',
   QEMU_WORKERS: '2',
   WITH_CONFIG_PATCH: '[{"op": "add", "path": "/cluster/network", "value": {"cni": {"name": "none"}}}]',
   IMAGE_REGISTRY: local_registry,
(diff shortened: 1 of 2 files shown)


.drone.jsonnet
```diff
@@ -477,6 +477,8 @@ local integration_cilium = Step('e2e-cilium', target='e2e-qemu', privileged=true
   SHORT_INTEGRATION_TEST: 'yes',
   WITH_SKIP_BOOT_PHASE_FINISHED_CHECK: 'yes',
   WITH_CUSTOM_CNI: 'cilium',
+  WITH_KUBESPAN: 'true',
```
smira, line 480: I'm not sure if it'd be fair to run with KubeSpan always, as it might affect Cilium and we might miss a bug without KubeSpan?

Also should we enabled for `e2e-kubespan`? We could do another Cilium test with KubeSpan as the third entry?

smira: I'm pretty sure I tested it locally :crossed_fingers:

frezbo:
> So I tried to change things a bit, and I think I hit the wall, but this expected. Disabling node-to-node encryption makes sense (as KubeSpan already does that), and connectivity test expects to observe the traffic between nodes, but as it goes via KubeSpan, I guess it can't see it, so we have the final set of failures:
> 
> ```
>   [-] Scenario [node-to-node-encryption/node-to-node-encryption]
>   [.] Action [node-to-node-encryption/node-to-node-encryption/ping-ipv4: cilium-test/client-59bddc8c5-g2t8g (10.244.4.6) -> cilium-test/host-netns-xt8tm (172.20.1.6:0)]
>   ❌ Expected to see unencrypted packets, but none found. This check might be broken
>   📄 No flows recorded for peer cilium-test/client-59bddc8c5-g2t8g during action ping-ipv4
>   📄 No flows recorded for peer cilium-test/host-netns-xt8tm during action ping-ipv4
>   [.] Action [node-to-node-encryption/node-to-node-encryption/ping-ipv4: cilium-test/host-netns-d2n4n (172.20.1.5) -> cilium-test/host-netns-xt8tm (172.20.1.6:0)]
>   ❌ Expected to see unencrypted packets, but none found. This check might be broken
>   ❌ Expected to see unencrypted packets, but none found. This check might be broken
>   📄 No flows recorded for peer cilium-test/host-netns-d2n4n during action ping-ipv4
>   📄 No flows recorded for peer cilium-test/host-netns-xt8tm during action ping-ipv4
> ```
> 
> So idk what is the proper way here? I don't see a way to skip some tests with `cilium connectivity`

We can disable those tests with `--test "!<test name>"

smira:
> We can disable those tests with `--test "!"

Please do that, so that we can merge this one!

.drone.jsonnet
```diff
@@ -490,6 +491,16 @@ local integration_cilium_strict = Step('e2e-cilium-strict', target='e2e-qemu', p
   WITH_CONFIG_PATCH: '[{"op": "add", "path": "/cluster/network", "value": {"cni": {"name": "none"}}}, {"op": "add", "path": "/cluster/proxy", "value": {"disabled": true}}]',
```
smira, line 491: I think I somehow removed `WITH_FIREWALL: 'accept'` from this block

smira (approved):

.drone.jsonnet
frezbo: yes, or could be my rebase, fixed, anyways it should pass, worked locally, running again

PR #8032, opened 2023-12-05
chore: optimize pcap dump

### LINKED ISSUE #7994: :bug: fix crash/slowness in talosctl pcap at high packet rate (when decoding packets)

DmitriyMV:
Reimplement `gopacket.PacketSource.PacketsCtx` as `forEachPacket`.

- Use `ZeroCopyPacketDataSource` instead of `PacketDataSource`. I didn't find any specific reason why `PacketDataSource` exists at all, since `NewPacket` is doing copy inside if you don't explicitly tell it not to.
- Use `WithPool` to pool packet buffers. It doesn't fully remove allocations, but it's a safe start. Send packets back into the pool after we are done with them.
- Pass `Packet` directly to the closure instead of waiting for it on the channel. We don't store this packet anywhere so there is no reason to async this part.
- Drop `time.Sleep` code in `forEachPacket` body.

Closes #7994


DmitriyMV:
@smira I changed code a bit - by passing error explicitly to the callback. On `talosctl -n 172.20.0.2 pcap -i enx5a6230c7d08d` where `talosctl -n 172.20.0.2 get links enx5a6230c7d08d -o yaml` is

```
node: 172.20.0.2
metadata:
    namespace: network
    type: LinkStatuses.net.talos.dev
    id: enx5a6230c7d08d
    version: 3
    owner: network.LinkStatusController
    phase: running
    created: 2023-12-05T12:09:31Z
    updated: 2023-12-05T12:09:32Z
spec:
    index: 8
    type: ether
    linkIndex: 0
    flags: UP,BROADCAST,RUNNING,MULTICAST,LOWER_UP
    hardwareAddr: 5a:62:30:c7:d0:8d
    permanentAddr: 5a:62:30:c7:d0:8d
    broadcastAddr: ff:ff:ff:ff:ff:ff
    mtu: 1500
    queueDisc: pfifo_fast
    operationalState: up
    kind: ""
    slaveKind: ""
    busPath: "0000:00:02.0"
    driver: virtio_net
    driverVersion: 1.0.0
    productID: "0x1000"
    vendorID: "0x1af4"
    product: Virtio network device
    vendor: Red Hat, Inc.
    linkState: true
    speedMbit: 4294967295
    port: Other
    duplex: Unknown
```

Even without printing packets it's always behind but it looks like it doesn't crash or slowdown. The results are like this:

```
packet capture error after 28 good packets
packet capture error: capture length exceeds snap length: 4592 > 4096
packet capture error: capture length exceeds snap length: 2379269936 > 4096
packet capture error: capture length exceeds snap length: 346851346 > 4096
packet capture error: capture length exceeds snap length: 1153569802 > 4096
packet capture error: capture length exceeds snap length: 679152136 > 4096
packet capture error: capture length exceeds snap length: 1154075890 > 4096
packet capture error: capture length exceeds snap length: 3190633179 > 4096
packet capture error: capture length exceeds snap length: 3105433045 > 4096
packet capture error: capture length exceeds snap length: 1417025788 > 4096
packet capture error: capture length exceeds snap length: 1669822213 > 4096
packet capture error: capture length exceeds snap length: 3251886405 > 4096
packet capture error after 1 good packets
packet capture error: capture length exceeds snap length: 4899 > 4096
packet capture error: capture length exceeds snap length: 2379269936 > 4096
packet capture error: capture length exceeds snap length: 346837265 > 4096
packet capture error: capture length exceeds snap length: 1153614363 > 4096
packet capture error: capture length exceeds snap length: 679152136 > 4096
packet capture error: capture length exceeds snap length: 3085534033 > 4096
packet capture error: capture length exceeds snap length: 1773334212 > 4096
packet capture error: capture length exceeds snap length: 1997150536 > 4096
packet capture error: capture length exceeds snap length: 1047492300 > 4096
packet capture error: capture length exceeds snap length: 2063852047 > 4096
packet capture error: capture length exceeds snap length: 3240910719 > 4096
packet capture error after 3 good packets
packet capture error: capture length exceeds snap length: 6714 > 4096
packet capture error: capture length exceeds snap length: 4521992 > 4096
packet capture error: capture length exceeds snap length: 346817024 > 4096
packet capture error: capture length exceeds snap length: 411086516 > 4096
packet capture error: capture length exceeds snap length: 2344592585 > 4096
packet capture error: capture length exceeds snap length: 2044819425 > 4096
packet capture error: capture length exceeds snap length: 2142457517 > 4096
packet capture error: capture length exceeds snap length: 4072632877 > 4096
packet capture error: capture length exceeds snap length: 2790351858 > 4096
packet capture error: capture length exceeds snap length: 2646865750 > 4096
packet capture error: capture length exceeds snap length: 773497481 > 4096
packet capture error after 28 good packets
packet capture error: capture length exceeds snap length: 7306 > 4096
packet capture error: capture length exceeds snap length: 2379269936 > 4096
packet capture error: capture length exceeds snap length: 346872071 > 4096
packet capture error: capture length exceeds snap length: 1153627761 > 4096
packet capture error: capture length exceeds snap length: 679152136 > 4096
packet capture error: capture length exceeds snap length: 2197096417 > 4096
packet capture error: capture length exceeds snap length: 682724056 > 4096
packet capture error: capture length exceeds snap length: 3253611061 > 4096
packet capture error: capture length exceeds snap length: 1281535514 > 4096
packet capture error: capture length exceeds snap length: 3020549361 > 4096
packet capture error: capture length exceeds snap length: 1176808472 > 4096
```

I dunno if anything else can be done here, but dropping `Sleep` certainly helped.

smira:
yep, client-side was buggy... I still wonder if there's an issue also on server-side under high load

smira:
This PR even if it doesn't solve the whole thing, should be good anyways

DmitriyMV:
Okay, after reading kernel docs (man [didn't](https://man7.org/linux/man-pages/man7/packet.7.html) help at all) I'm convinced that the code responsible for `snap_len` never worked properly at all. 

Per [documentation](https://www.kernel.org/doc/html/latest/networking/packet_mmap.html):
> A frame can be of any size with the only condition it can fit in a block.

And TPACKET_V3 (which is used by default if `afpacket` module is not told otherwise) is also

> Non static frame size to capture entire packet payload

Google Project Zero research agrees with that:

> Here’s what each field means in the tpacket_req3 struct:
>```
>tp_block_size - the size of each block.
>tp_block_nr - the number of blocks.
>tp_frame_size - the size of each frame, ignored for TPACKET_V3.
>tp_frame_nr - the number of frames, ignored for TPACKET_V3.
>tp_retire_blk_tov - timeout after which a block is retired, even if it’s not fully filled with data (see below).
>tp_sizeof_priv - the size of per-block private area. This area can be used by a user to store arbitrary information associated with each block.
>tp_feature_req_word - a set of flags (actually just one at the moment), which allows to enable some additional functionality.
>```

So it appears that it's either unused or used for something entirely different. The code in afpacket also works in [blocks](https://github.com/gopacket/gopacket/blob/master/afpacket/afpacket.go#L462) in V3. I checked the source and didn't find anything related to the ignoring of `tp_frame_size` in the [kernel](https://github.com/torvalds/linux/blob/v6.6/net/packet/af_packet.c#L4444) sources. However, I did find that `tp_frame_size` and others should [respect](https://elixir.bootlin.com/linux/latest/source/net/packet/af_packet.c#L4446) certain rules (which we never checked in our code):

> tp_block_size must be a multiple of PAGE_SIZE (1)
tp_frame_size must be greater than TPACKET_HDRLEN (obvious)
tp_frame_size must be a multiple of TPACKET_ALIGNMENT
tp_frame_nr must be exactly frames_per_block*tp_block_nr

So even if we set `tp_block_size`, we should get the page size before doing that. 

So for now I, again, propose to just drop this [line](https://github.com/siderolabs/talos/blob/e22ab440d7794a9c46edf1357124571057b6b19d/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go#L2216-L2215) and the support of `SnapLen` in client.

smira (approved):

smira:
:cool:  research!

PR #8033, opened 2023-12-05
docs: rework machine config documentation generation

### LINKED ISSUE #7995: :blue_book:  provide documentation for multi-doc machine config

### LINKED ISSUE #8023: :bug: talos 1.6.0-beta.0 cannot boot in qemu/kvm
### Description

While trying to upgrade to talos 1.6.0-beta.0 at https://github.com/rgl/terraform-libvirt-talos/tree/upgrade-to-talos-1.6 (in qem/kvm VMs), talos cannot boot in qemu/kvm due to:

image

digging a little bit at the grub prompt with `ls (hd0,gpt3)/grub/x86_64-efi/`, it seems there's a bug somewhere trying to find files with the `error: invalid XFS directory entry` error:

image

it seems this kind of bug is being reported elsewhere, I found these might help understand/fix the problem:

* https://github.com/siderolabs/pkgs/commit/70919d8
* https://savannah.gnu.org/bugs/?64376
* https://github.com/calamares/calamares/issues/2166
* https://git.savannah.gnu.org/cgit/grub.git/commit/?id=aa7c1322671eef48ba72d3b733da37e63eb37328
* https://git.savannah.gnu.org/cgit/grub.git/commit/?id=ad7fb8e2e02bb1dd0475ead9919c1c82514d2ef8

### Environment

- Talos version: talos 1.6.0-beta.0
- Kubernetes version: 1.27.8
- Platform: qemu/kvm

smira:
Generate a structured table of contents following the structure of the config.

Make high-level examples follow the full structure of the config.

Document new multi-doc machine config.

Fixes #7995


pkg/machinery/config/types/v1alpha1/schemas/v1alpha1_config.schema.json
```diff
@@ -2395,19 +2389,6 @@
       "additionalProperties": false,
       "type": "object"
     },
-    "PodCheckpointer": {
```
smira, line 2398: should not be documented

pkg/machinery/config/types/network/rule_config.go
```diff
@@ -41,27 +41,78 @@ var (
...
 	Subnet netip.Prefix `yaml:"subnet"`
-	Except netip.Prefix `yaml:"except,omitempty"`
+	//   description: |
+	//     Except defines a source subnet to exclude from the rule, it gets excluded from the `subnet`.
+	Except Prefix `yaml:"except,omitempty"`
+}
+
+// Prefix is a wrapper for netip.Prefix.
+//
+// It implements IsZero() so that yaml.Marshal correctly skips empty values.
```
smira, line 106: this is where our encoder and `yaml.Marshal` act differently on `omitempty` - ours correctly skips zero `netip.Prefix`, while `yaml.Marshal` checks public struct fields which are not there for `netip.Prefix`

hack/docgen/main.go
```diff
@@ -346,6 +348,8 @@ func collectFields(s *structType, aliases map[string]aliasType) (fields []*Field
 	for _, f := range s.node.Fields.List {
 		if f.Names == nil {
 			// This is an embedded struct.
+			fields = append(fields, &Field{Type: "unknown"})
```
Unix4ever, line 351: How will it look like in the docs?
Or that should never be rendered?

pkg/machinery/config/encoder/markdown.go
```diff
@@ -60,86 +28,119 @@ type FileDoc struct {
...
+
+		slices.Reverse(parts)
+
+		for _, part := range parts {
+			switch part {
+			case "":
+				in = []any{in}
+			case "-":
+				in = map[string]any{
+					"example.com": in,
```
Unix4ever, line 137: How is this rendered? Is it ignored?  Should we use some constant here?

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -63,7 +63,9 @@ type Config struct {
 	//     - false
 	//     - no
 	ConfigDebug *bool `yaml:"debug,omitempty"`
-	//docgen:nodoc
+	// docgen:nodoc
+	//
+	// Not supported anymore.
```
Unix4ever, line 68: Should we mark as deprecated too?

Unix4ever (approved):

hack/docgen/main.go
smira: it is like below in the same block, it gets skipped in the docs

the problem was that with the embedded struct the field offset gets off by one, so the comment blocks are attached to the wrong fields when marshalling

all new documents look like that:

```go
struct SomeConfig{
   meta.Meta

   SomeValue string
}```

So without it, the index of the fields is off by one, but with this fix it is correct. the `meta.Meta` is not documented either way

pkg/machinery/config/encoder/markdown.go
smira: an example is here: https://deploy-preview-8033--talos-dev.netlify.app/v1.6/reference/configuration/v1alpha1/config/#Config.machine.registries.config.-.tls

the only time we have it in v1alpha1 is the registry config, where the key is the hostname of some sort.

when we render inner example, to make it an example rooted at the full config, we use this fake key

hack/docgen/main.go
Unix4ever: ah, I see

PR #8060, opened 2023-12-13
fix: leave discovery service later in the reset sequence

### LINKED ISSUE #8057: 🐛 graceful reset (etcd leave) doesn't work with KubeSpan
The machine leaves Discovery Service earlier than leaving `etcd`, so by the time etcd leave happens, links to the controlplane nodes are already broken, and etcd member can't communicate with other peers.

The fix is to defer leaving Discovery Service on reset to some later phase, when network connectivity via KubeSpan is no longer needed.

smira:
Fixes #8057

I went back and forth on the way to fix it exactly, and ended up with a pretty simple version of a fix.

The problem was that discovery service was removing the member at the initial phase of reset, which actually still requires KubeSpan to be up:

* leaving `etcd` (need to talk to other members)
* stopping pods (might need to talk to Kubernetes API with some CNIs)

Now leaving discovery service happens way later, when network interactions are no longer required.


.drone.jsonnet
```diff
@@ -547,7 +547,6 @@ local integration_kubespan = Step('e2e-kubespan', target='e2e-qemu', privileged=
   WITH_CLUSTER_DISCOVERY: 'true',
   WITH_KUBESPAN: 'true',
   IMAGE_REGISTRY: local_registry,
-  WITH_CONFIG_PATCH: '[{"op": "replace", "path": "/cluster/discovery/registries/kubernetes/disabled", "value": false}]',  // use Kubernetes discovery backend
```
frezbo, line 550: don't need this here anymore?

smira: yep.. I think it's disabled by default for a long time now

PR #8061, opened 2023-12-13
chore: fix the gvisor test

frezbo:
The gvisor test was not using the correct runtimeclass and would have always passed the regardless.


smira:
The change in this PR is good (tested via `release-1.6`)

PR #8076, opened 2023-12-15
fix: use correct prefix when installing SBC files

smira:
When creating an image under non-default mount prefix, it should be explictly used when copying SBC files.

See https://github.com/siderolabs/image-factory/issues/65

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/board/bananapi_m64/bananapi_m64.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/board/bananapi_m64/bananapi_m64.go
@@ -73,7 +73,7 @@ func (b *BananaPiM64) Install(options runtime.BoardInstallOptions) (err error) {
 	}
 
 	src := filepath.Join(options.DTBPath, dtb)
-	dst := filepath.Join("/boot/EFI/dtb", dtb)
+	dst := filepath.Join(options.MountPrefix, "/boot/EFI/dtb", dtb)
 
 	err = os.MkdirAll(filepath.Dir(dst), 0o600)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


internal/app/machined/pkg/runtime/v1alpha1/board/bananapi_m64/bananapi_m64.go
```diff
@@ -73,7 +73,7 @@ func (b *BananaPiM64) Install(options runtime.BoardInstallOptions) (err error) {
 	}
 
 	src := filepath.Join(options.DTBPath, dtb)
-	dst := filepath.Join("/boot/EFI/dtb", dtb)
+	dst := filepath.Join(filepath.Join(options.MountPrefix, "/boot/EFI/dtb"), dtb)
```
frezbo, line 76:
```suggestion
	dst := filepath.Join(options.MountPrefix, "/boot/EFI/dtb", dtb)
```

cleaner I guess?

frezbo: well other places too

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/board/bananapi_m64/bananapi_m64.go
smira: oh yeah, I somehow missed that, thanks

smira: yes, fixed them all

frezbo (approved):

smira:
Verified the fix with Image Factory.

PR #8088, opened 2023-12-18
docs: announce that SecureBoot is available

smira:
Restructure the docs a bit to start with the easiest option (via Image Factory).


frezbo (approved):
:heart:

PR #8090, opened 2023-12-19
Improve catch-all network interface `deviceSelector`.

reitermarkus:
# Pull Request

## What? (description)

Change example for catch-all network interface `deviceSelector`.

## Why? (reasoning)

The existing example does not work on Raspberry Pi, I assume since the network interface is not a PCIe device but connected to the USB controller.

I am using Raspberry Pis (Hardware PHY) and Proxmox VMs (VirtIO PHY). The common attributes of the ethernet interfaces seem to be `type: ether` and `kind: ''`.

## Acceptance

### DIFF
--- a/website/content/v1.6/talos-guides/network/predictable-interface-names.md
+++ b/website/content/v1.6/talos-guides/network/predictable-interface-names.md
@@ -32,7 +32,9 @@ There are two ways to solve this:
     network:
       interfaces:
         - deviceSelector:
-            busPath: "0*" # should select any hardware network device, if you have just one, it will be selected
-          # any configuration can follow, e.g:
+            # This should select any hardware ethernet interface. If you have just one, it will be selected.
+            type: ether
+            kind: ''
+          # Any configuration can follow, e.g.:
           addresses: [10.3.4.5/24]
   ```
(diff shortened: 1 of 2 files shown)


reitermarkus:
Nevermind, it seems these fields are not supported in the `deviceSelector` config.

What is the recommended way to match the network interface on Raspberry Pis? It seems the only options are to match the driver (`bcmgenet`) or by a specific `hardwareAddr`, but then this won't match on the Proxmox VMs.

smira:
> Nevermind, it seems these fields are not supported in the `deviceSelector` config.
> 
> What is the recommended way to match the network interface on Raspberry Pis? It seems the only options are to match the driver (`bcmgenet`) or by a specific `hardwareAddr`, but then this won't match on the Proxmox VMs.

Are you looking for a `deviceSelector` that would match across different hardware types? I guess this might be tricky, but still might be a nice feature request. We could add a match for physical interfaces (that's what you described with `type: ether` and `kind: `)

reitermarkus:
> Are you looking for a `deviceSelector` that would match across different hardware types?

For my simple setup with only a single hardware interface per device, ideally I would not have to specify any selector and Talos should just use the only available interface.

smira:
you can also use `net.ifnames=0` and have `eth0` if you have a single interface, but the request for a simple device selector makes sense

smira:
#8173 implements `physical: true` device selector.

PR #8091, opened 2023-12-19
feat: support iPXE direct booting in `talosctl cluster create`

smira:
This embeds a tiny TFTP server which serves UEFI iPXE which embeds a script that chainloads a given iPXE script.

> Note: only supports UEFI boot (BIOS boot will need a bit different flow).


smira:
I'm thinking about adding a set of integration tests for Image Factory, but I want to implement downloading of other assets as well.

smira:
Now something like that works:


```shell
$ talosctl cluster create ... --ipxe-boot-script=https://pxe.factory.talos.dev/pxe/cf9b7aab9ed7c365d5384509b4d31c02fdaa06d2b3ac6cc0bc806f28130eff1f/v1.6.0/metal-amd64
```

Unix4ever (approved):

rsmitty:
:cool:

PR #8092, opened 2023-12-19
docs: secureboot sd-boot menu clarification

TimJones:
Add note to try spamming Esc to bring up the sd-boot menu option if keys don't automatically enroll in UEFI firmware.


frezbo (approved):

smira (approved):
should we have higher timeout when on ISO?

PR #8093, opened 2023-12-19
fix: imager should support different Talos versions

smira:
Add some quirks to make images generated with newer Talos compatible with images generated by older Talos.

Specifically, reset options were adding in Talos 1.4, so we shouldn't add them for older versions.

See https://github.com/siderolabs/image-factory/issues/20


smira:
Draft, I want to rewrite this.

frezbo (approved):

PR #8101, opened 2023-12-21
chore: bump dependencies

smira:
Go modules, tools, etc.


frezbo (approved):

smira:
I will pull in https://github.com/siderolabs/pkgs/pull/858 here as it builds

PR #8103, opened 2023-12-22
release(v1.6.1): prepare release

smira:
This is the official v1.6.1 release.


frezbo (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_stability_test.go
```diff
@@ -42,6 +42,7 @@ func TestConfigEncodingStability(t *testing.T) {
 	}
 
 	currentVersion := ensure.Value(semver.ParseTolerant(gendata.VersionTag))
+	currentVersion.Patch = 0
```
smira, line 45: last minute fix, the check was wrong, as version contract is like `1.6.0`, but last version is `1.6.1`, so we should make patch version 0

I will take it back to `main` as well

PR #8118, opened 2023-12-28
fix: default priority for ipv6

sergelogvinov:
# Pull Request

## Why? (reasoning)

We will use the default IPv6 gateway priority as 2048. The RA default is 1024, which leads to verbose messages such as 'error adding route: netlink receive: file exists.'

Azure uses DHCPv6 and RA for configuring IPv6 on the node. The platform sets the default gateway as a fallback in case 'accept_ra' is not set to 2.

PS. @smira can we backport it to 1.6.x?  Thanks.

## Acceptance


smira (approved):
thank you!

PR #8120, opened 2023-12-29
feat: implement local caching dns server

### LINKED ISSUE #7693: implement host-level DNS cache
## Feature Request

Talos doesn't do any kind of DNS response caching on the host.

The `CoreDNS` configured by default does caching for Kubernetes workloads, but this doesn't apply to host networking activities.

It might be good to have a DNS cache as a separate component running on the host.

DmitriyMV:
This PR adds a new controller - `DNSServerController` that starts tcp and udp dns servers locally. Just like `EtcFileController` it monitors `ResolverStatusType` and updates the list of destinations from there.

Most of the caching logic is in our "lobotomized" "`CoreDNS` fork. We need this fork because default `CoreDNS` carries full Caddy server and various other modules that we don't need in Talos. On our side we implement random selection of the actual dns and request forwarding.

Closes #7693


internal/app/machined/pkg/controllers/network/dns_server.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil
+	})
+}
+
+func (ctrl *DNSServerController) addr() string {
+	if addr := ctrl.Addr; addr != "" {
+		return addr
+	}
+
+	return ":53"
```
DmitriyMV, line 162: Maybe it should be `127.0.0.1:53`?

internal/pkg/dns/dns.go
```diff
@@ -0,0 +1,251 @@
...
+
+	var (
+		resp *dns.Msg
+		err  error
+	)
+
+	h.logger.Info("dns request", zap.Stringer("data", msg))
+
+	for {
+		prx, ok := randomGet(h.dests)
```
DmitriyMV, line 133: We should probably try all upstreams before we give up (CoreDNS actually does this).

Dockerfile
```diff
@@ -791,7 +791,7 @@ COPY --from=pkg-grub-arm64 /usr/lib/grub /usr/lib/grub
 COPY --from=pkg-grub-amd64 /usr/lib/grub /usr/lib/grub
 COPY --from=unicode-pf2 /usr/share/grub/unicode.pf2 /usr/share/grub/unicode.pf2
 RUN ln /bin/installer /bin/imager
-RUN find /bin /etc /lib /usr /sbin | grep -Ev '/etc/hosts|/etc/resolv.conf' \
+RUN find /bin /etc /lib /usr /sbin | grep -Ev '/etc/hosts|/etc/resolv.conf|/etcd/resolv-global.conf' \
```
frezbo, line 794:
```suggestion
RUN find /bin /etc /lib /usr /sbin | grep -Ev '/etc/hosts|/etc/resolv.conf|/etc/resolv-global.conf' \
```

internal/app/machined/pkg/controllers/network/etcfile_test.go
frezbo, line 1: don't we need a test for global-resolv conf file content?

internal/pkg/dns/dns.go
```diff
@@ -0,0 +1,251 @@
...
+		return dns.RcodeServerFailure, err
+	}
+
+	if !req.Match(resp) {
+		resp = new(dns.Msg).SetRcode(req.Req, dns.RcodeFormatError)
+
+		err = wrt.WriteMsg(resp)
+		if err != nil {
+			// We can't do much here, but at least log the error.
+			h.logger.Warn("error writing dns formerr", zap.Error(err))
```
frezbo, line 164: `formerr`? (typo or something valid?)

internal/pkg/utils/random.go
```diff
@@ -0,0 +1,32 @@
...
+package utils
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+)
+
+// MakeRandomGetter returns a function that can be used to randomly access the given store.
+func MakeRandomGetter[V, T any](ln func(V) int, getFn func(V, int) T) func(v V) (T, bool) {
```
frezbo, line 14: do we need tests for this?

internal/pkg/dns/dns.go
DmitriyMV: Fixed thanks!

internal/pkg/utils/random.go
DmitriyMV: On further thinking about this, I decided to remove this thing entirely. It's a leftover from the time before I moved to using slices for upstreams.

internal/pkg/dns/dns.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		// we should probably call newProxy.Healthcheck() if there are too many errors
+		c.logger.Warn("error serving dns request", zap.Error(err))
+	}
+}
+
+// Handler is a dns proxy selector.
+type Handler struct {
+	mx     sync.RWMutex
+	rnd    *rand.Rand
```
DmitriyMV, line 103: This should be removed.

internal/app/machined/pkg/controllers/network/etcfile_test.go
DmitriyMV: Done!

go.mod
```diff
@@ -3,12 +3,14 @@ module github.com/siderolabs/talos
 go 1.21.5
 
 replace (
+	github.com/coredns/coredns => github.com/DmitriyMV/coredns v0.0.0-20240110133452-9aa56cf61808
```
smira, line 6: we should have this under `siderolabs/` org

I created repo for you: https://github.com/siderolabs/coredns/, and you're admin there

Please also tag it properly so we can track updates (with upstream version?)

We also need automation to import upstream changes there

go.mod
```diff
@@ -3,12 +3,14 @@ module github.com/siderolabs/talos
...
 	// Use nested module.
 	github.com/siderolabs/talos/pkg/machinery => ./pkg/machinery
 
 	// forked go-yaml that introduces RawYAML interface, which can be used to populate YAML fields using bytes
 	// which are then encoded as a valid YAML blocks with proper indentiation
 	gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20220527175918-f17b0f05cf2c
+
```
smira, line 13: empty line?

smira:
hack/release.toml should have a line describing the changes

Dockerfile
```diff
@@ -791,7 +791,7 @@ COPY --from=pkg-grub-arm64 /usr/lib/grub /usr/lib/grub
 COPY --from=pkg-grub-amd64 /usr/lib/grub /usr/lib/grub
 COPY --from=unicode-pf2 /usr/share/grub/unicode.pf2 /usr/share/grub/unicode.pf2
 RUN ln /bin/installer /bin/imager
-RUN find /bin /etc /lib /usr /sbin | grep -Ev '/etc/hosts|/etc/resolv.conf' \
+RUN find /bin /etc /lib /usr /sbin | grep -Ev '/etc/hosts|/etc/resolv.conf|/etc/resolv-global.conf' \
```
smira, line 794: it doesn't have to be there (these are not Talos files)

Dockerfile
```diff
@@ -652,7 +652,7 @@ COPY --chmod=0644 hack/udevd/80-net-name-slot.rules /rootfs/usr/lib/udev/rules.d
 COPY --chmod=0644 hack/lvm.conf /rootfs/etc/lvm/lvm.conf
 RUN <<END
     ln -s /usr/share/zoneinfo/Etc/UTC /rootfs/etc/localtime
-    touch /rootfs/etc/{extensions.yaml,resolv.conf,hosts,os-release,machine-id,cri/conf.d/cri.toml,cri/conf.d/01-registries.part,cri/conf.d/20-customization.part}
+    touch /rootfs/etc/{extensions.yaml,resolv.conf,resolv-global.conf,hosts,os-release,machine-id,cri/conf.d/cri.toml,cri/conf.d/01-registries.part,cri/conf.d/20-customization.part}
```
smira, line 655: I wonder if this is a good name, it doesn't have to be in `/etc` at all, we could perfectly write it to `tmpfs` under `/system`. This way we don't have to manage it with bind-mount tricks.

The `/etc/resolv.conf` is there as it's a well known path, while the path to the `resolv.conf` we pass to the `kubelet` doesn't have to be anything fixed.

internal/app/machined/pkg/controllers/network/dns_server.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctxutil.Cause(ctx)
+		case <-r.EventCh():
+		}
+
+		logger.Info("listing resolver status resources")
```
smira, line 106: this is crazy verbose and low information value

internal/app/machined/pkg/controllers/network/status.go
```diff
@@ -142,7 +142,7 @@ func (ctrl *StatusController) Run(ctx context.Context, r controller.Runtime, log
 		// etc files
 		result.EtcFilesReady = true
 
-		for _, requiredFile := range []string{"hosts", "resolv.conf"} {
+		for _, requiredFile := range []string{"hosts", "resolv.conf", "resolv-global.conf"} {
```
smira, line 145: I think it's less relevant for the "network up" status.

internal/pkg/dns/dns.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return &Cache{cache: c, logger: l}
+}
+
+// ServeDNS implements [dns.Handler].
+func (c *Cache) ServeDNS(wr dns.ResponseWriter, msg *dns.Msg) {
+	_, err := c.cache.ServeDNS(context.Background(), wr, msg)
+	if err != nil {
+		// we should probably call newProxy.Healthcheck() if there are too many errors
+		c.logger.Warn("error serving dns request", zap.Error(err))
```
smira, line 96: if the logger here is coming from the controller, it might spam a lot `controller-runtime` logs, probably we'd rather have a separate log for the dns server?

pkg/machinery/resources/network/dns_server.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// DNSServerType is type of DNSServer resource.
+const DNSServerType = resource.Type("DNSServers.net.talos.dev")
+
+// DNSServer resource holds DNS resolver info.
+type DNSServer = typed.Resource[DNSServerSpec, DNSServerExtension]
```
smira, line 22: I think this should be `ResolverStatus`, but we have this already occupied.

`DNSServer` is not really an accurate name, as we don't run an authoritative DNS server, but rather a resolver which forwads the requests. Probably something more about `resolved` ? And same for the controller name possibly. Need some clever name here.

Dockerfile
DmitriyMV: That is possible, but we will have to rework `EtcFileController` or/and `EtcFileController`.

smira: There's totally no need to do that, just write down contents to e.g. `/system/resolved/resolv.conf` for example. It's tmpfs, always mounted and writeable, you don't need any kind of `EtcFileController` there

DmitriyMV: There is `renderResolvConf` method in `EtcFileController` and it also contains dns resolve logic other than just simple write of `nameserver 127.0.0.1`. We will have to copy/modify it.

internal/app/machined/pkg/controllers/network/etcfile.go
```diff
@@ -158,13 +172,17 @@ func (ctrl *EtcFileController) Run(ctx context.Context, r controller.Runtime, lo
...
-		if i >= 3 {
-			// only use firt 3 nameservers, see MAXNS in https://linux.die.net/man/5/resolv.conf
-			break
-		}
+	if resolverStatus == nil {
+		fmt.Fprintln(&buf, "nameserver 127.0.0.1")
+	} else {
+		for i, resolver := range resolverStatus.DNSServers {
+			if i >= 3 {
+				// only use firt 3 nameservers, see MAXNS in https://linux.die.net/man/5/resolv.conf
```
utkuozdemir, line 180:
```suggestion
				// only use first 3 nameservers, see MAXNS in https://linux.die.net/man/5/resolv.conf
```

go.mod
DmitriyMV: Done. I used the vX.Y.ZZ+50 version scheme. So that `v1.11.1` is `v1.11.51` in our fork.

go.mod
DmitriyMV: Fixed.

Dockerfile
DmitriyMV: Removed.

Dockerfile
DmitriyMV: Oh, okay. Yeah now `github.com/siderolabs/talos/internal/app/machined/pkg/controllers/network.EtcFileController` writes directly to `/system/resolved/resolv.conf`.

> manage it with bind-mount tricks.

It's still is in `github.com/siderolabs/talos/internal/app/machined/pkg/system/services.Kubelet.Runner` otherwise it doesn't work.

internal/app/machined/pkg/controllers/network/dns_server.go
DmitriyMV: Removed.

internal/app/machined/pkg/controllers/network/status.go
DmitriyMV: Removed.

internal/pkg/dns/dns.go
DmitriyMV: Yeah, I thought about this too. I redid initialization in `github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha2.Controller` so it now have a helper function which is used for both controller logs and logs for dns cache. Im was a bit surprised that `logger.With` method doesn't overwrite existing keys, instead it appends to them.

pkg/machinery/resources/network/dns_server.go
DmitriyMV: It's `DNSResolveCache` now. I think it sounds pretty nice.

internal/app/machined/pkg/controllers/network/etcfile.go
DmitriyMV: Done, thanks!

Dockerfile
smira: yes, I mean bind-mount of different sort when we write to read-only `/etc` ;)

DmitriyMV:
I updated the PR. Relevant stuff:
- We can disable/enable dns caching on the fly now. This requires us to atomically update `/etcd/resolv.conf` file. 
- Runner is now stable to context cancelations and properly handles situations where `Stop` was called before we started listening. This is was actually quite subtle to implement correctly.
- We enable local dns only in Talos >= 1.7

internal/pkg/utils/utils.go
```diff
@@ -0,0 +1,123 @@
...
+var (
+	// ErrAlreadyRunning is the error that is returned when runner is already running/closing/closed.
+	ErrAlreadyRunning = errors.New("runner is already running/closing/closed")
+	// ErrNotRunning is the error that is returned when runner is not running/closing/closed.
+	ErrNotRunning = errors.New("runner is not running/closing/closed")
+)
+
+// Stop stops runner. It's safe to call even if runner is already stopped or in process of being stopped.
+func (r *Runner) Stop() error {
+	if r.status.CompareAndSwap(notRunning, closing) || !r.status.CompareAndSwap(running, closing) {
```
DmitriyMV, line 107: This is important! Otherwise `fn` can happen after `stop` was called and dns library doesn't handle this situation.

pkg/machinery/resources/network/dns_resolve_cache.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// DNSResolveCacheType is type of DNSResolveCache resource.
+const DNSResolveCacheType = resource.Type("DNSServers.net.talos.dev")
```
utkuozdemir, line 19: Should the type name stay unchanged (`DNSServers`)?

smira: it should be updated to match the constant

hack/release.toml
```diff
@@ -31,6 +31,12 @@ Talos is built with Go 1.21.6.
         title = "Device Selectors"
         description = """\
 Talos Linux now supports `physical: true` qualifier for device selectors, it selects non-virtual network interfaces (i.e. `en0` is selected, while `bond0` is not).
+"""
+
+    [notes.dns-resolve-cache]
+        title = "DNS Caching"
+        description = """\
+Talos now caches DNS lookups for host and host-networked pods. This brings us in line with the behavior of other Kubernetes distributions.
```
smira, line 39:
```suggestion
Talos Linux now  provides a caching DNS resolver for host workloads (including host networking pods). It can be disabled with:

```yaml
machine:
   features:
       blah: blah
```
```

File: internal/app/machined/pkg/controllers/network/etcfile.go [hunk 19]
```diff
@@ -214,3 +246,47 @@ func (ctrl *EtcFileController) renderHosts(hostnameStatus *network.HostnameStatu
 
 	return buf.Bytes(), nil
 }
+
+func atomicFileWrite(dst string, b []byte, perm os.FileMode) error {
```
smira, line 250: we seem to be creating too many functions doing same thing. in this case, as we mount this file into the container, I think renaming won't work with bind mounting properly.

it's better to use here a write which doesn't rewrite the file if it doesn't change. 

See https://github.com/siderolabs/talos/pull/8126/files#diff-025f2ce8630bc16d48cc048303aa0fe8e6b42d5f5ee59056c59163fbcc0ff2deR69-R78

smira (approved):

PR #8126, opened 2024-01-02
feat: extensions services config

frezbo:
Support config files for extension services.

Fixes: #7791


internal/app/machined/pkg/controllers/cri/seccomp_profile_test.go
```diff
@@ -85,29 +85,6 @@ func (suite *CRISeccompProfileSuite) TestReconcileSeccompProfile() {
 		})
 	}
 
-	suite.AssertWithin(1*time.Second, 100*time.Millisecond, func() error {
```
frezbo, line 88: this was already asserted in the table test above

internal/app/machined/pkg/controllers/runtime/extension_services_config_files.go
```diff
@@ -0,0 +1,153 @@
...
+						}
+					}
+				}
+
+				touchedIDs[fileName] = struct{}{}
+			}
+		}
+
+		// cleanup
+		if err := filepath.WalkDir(ctrl.ExtensionsConfigBaseDir, func(path string, d fs.DirEntry, err error) error {
```
frezbo, line 112: the ext service is not restarted on config file removal, that still needs to be fixed

internal/app/machined/pkg/system/services/extension.go
```diff
@@ -156,6 +169,24 @@ func (svc *Extension) Runner(r runtime.Runtime) (runner.Runner, error) {
 		}
 	}
 
+	extServicesConfig := resource.NewMetadata(runtimeres.NamespaceName, runtimeres.ExtensionServicesConfigType, svc.Spec.Name, resource.VersionUndefined)
+
+	specResource, err := r.State().V1Alpha2().Resources().Get(context.Background(), extServicesConfig)
```
frezbo, line 174: I don't like this and ignoring the `state.IsNotFoundError`, any better ideas?

pkg/machinery/config/configloader/internal/decoder/decoder.go
```diff
@@ -124,6 +128,8 @@ func decode(manifest *yaml.Node) (target config.Document, err error) {
 	}
 
 	switch {
+	case version == "":
```
frezbo, line 131: this fixes a bug when `apiVersion` is not defined properly in the config document and the error was not clear before

pkg/machinery/extensions/services/services.go
```diff
@@ -76,6 +76,8 @@ type Dependency struct {
 	Network []nethelpers.Status `yaml:"network,omitempty"`
 	// Time sync check.
 	Time bool `yaml:"time,omitempty"`
+	// Depends on configfiles being present.
+	ConfigFiles []string `yaml:"configFiles,omitempty"`
```
frezbo, line 80: I'm  not sure if this should be part of `depends` or something more explicit

internal/app/machined/pkg/controllers/runtime/extension_services_config_files.go
frezbo: resolved

internal/app/machined/pkg/controllers/runtime/extension_services_config.go
```diff
@@ -0,0 +1,93 @@
...
+		cfg, err := safe.ReaderGetByID[*config.MachineConfig](ctx, r, config.V1Alpha1ID)
+		if err != nil && !state.IsNotFoundError(err) {
+			return fmt.Errorf("error getting machine config: %w", err)
+		}
+
+		r.StartTrackingOutputs()
+
+		if cfg != nil && cfg.Config() != nil && cfg.Config().ExtensionServicesConfig() != nil {
+			for _, ext := range cfg.Config().ExtensionServicesConfig().ConfigData() {
+				extensionsCfgFiles := make([]runtime.ExtensionServicesConfigFile, 0, len(ext.ConfigFiles()))
```
smira, line 70: could be replaced with `xslices.Map`

smira (commented):
I'd guess we need some documentation updates?

internal/app/machined/pkg/controllers/runtime/extension_services_config.go
```diff
@@ -0,0 +1,93 @@
...
+					spec.TypedSpec().Files = extensionsCfgFiles
+
+					return nil
+				}); err != nil {
+					return err
+				}
+			}
+		}
+
+		if err = safe.CleanupOutputs[*runtime.ExtensionServicesConfig](ctx, r); err != nil {
```
smira, line 89: this method doesn't support clean up of resources which have finalizers set on them :(

smira, line 89: this method doesn't support clean up of resources which have finalizers set on them :(

internal/app/machined/pkg/controllers/runtime/extension_services_config.go
```diff
@@ -0,0 +1,93 @@
...
+
+// Inputs implements controller.Controller interface.
+func (ctrl *ExtensionServicesConfigController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineConfigType,
+			ID:        optional.Some(config.V1Alpha1ID),
+		},
+	}
```
smira, line 37: if we want to track and properly destroy ouputs with finalizers, we need to add outputs as an input with `Kind: InputDestroyReady`

internal/app/machined/pkg/controllers/runtime/extension_services_config.go
```diff
@@ -0,0 +1,93 @@
...
+func (ctrl *ExtensionServicesConfigController) Name() string {
+	return "runtime.ExtensionServicesConfigController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *ExtensionServicesConfigController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: config.NamespaceName,
+			Type:      config.MachineConfigType,
```
smira, line 34: `Kind` is missing here

internal/app/machined/pkg/controllers/runtime/extension_services_config_files.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+		}
+
+		touchedDirs := map[string]struct{}{}
+
+		for iter := list.Iterator(); iter.Next(); {
+			extensionConfigPath := filepath.Join(ctrl.ExtensionsConfigBaseDir, iter.Value().Metadata().ID())
+
+			switch iter.Value().Metadata().Phase() {
+			case resource.PhaseTearingDown:
```
smira, line 90: this feels really complicated, and hard to test properly

I haven't read the whole change, but I assume we say that Talos owns all files for the extensions configuration, so in that case we could simply remove any file which shouldn't be there as part of any reconciliation (and that allows to skip finalizers)

pkg/machinery/config/configloader/internal/decoder/decoder.go
smira: I wonder if that should be moved down, as it might break loading config patches with v1alpha1 strategic merge (???) not sure

but if the version is not defined, it should actually return an error from `registry.New()` below?

internal/app/machined/pkg/controllers/runtime/extension_services_config_files.go
frezbo: hmm that might be easier, i went down the path of checking the teraingdownphase since even a file removal needs a restart of the extension service

smira: makes sense... can we detect which services were affected from the paths? as we update/remove things?

frezbo: kind of, from the parent folder, but it just seemed brittle and if someone manually creates a folder, the code would try to restart a non-existing service,

frezbo: but i guess maybe we can still remove files we didn't touch and clean up part of code

smira: as it's `tmpfs`, we still know which parts we touched, as it can't be leftover from previous runs, as the controller runs in can keep the mapping of paths and which extensions they are related to

hack/release.toml
```diff
@@ -31,6 +31,24 @@ Talos is built with Go 1.21.6.
...
+        title = "Extension Services Config"
+        description = """\
+Talos now supports supplying configuration files for extension services that can be mounted into the extension service container.
+The extension service configuration is a separate config document. An example is shown below:
+
+```yaml
+---
+apiVersion: v1alpha1
+kind: ExtensionServicesConfig
+config:
```
andrewrynhard, line 46: I feel like we have two standards here now. We use `spec` (and so does Kubernetes). We also don't use it in other places I can't remember at the moment. I feel like there should be some consistency across the formats.

frezbo: it's just a top level key and can be anything, for firewall rules it's something like this:

```yaml
apiVersion: v1alpha1
kind: NetworkDefaultActionConfig
ingress: block
---
apiVersion: v1alpha1
kind: NetworkRuleConfig
name: kubelet-ingress
portSelector:
  ports:
    - 10250
  protocol: tcp
ingress:
  - subnet: 172.20.0.0/24
    except: 172.20.0.1/32
```

we never used `spec` anywhere

andrewrynhard: We do use `spec` ... `talosctl get` a resource and you can see. It feels inconsistent to me (and I feel like I represent an average user at this point :D)

andrewrynhard: One of the goals this year is to clean up these types of things... resources names, resource layouts, etc.

andrewrynhard: I don't want to block on this but I am going to work on making us all aware of this. I think this is a larger issue that needs to be tracked.

hack/release.toml
```diff
@@ -6,7 +6,7 @@ github_repo = "siderolabs/talos"
 match_deps = "^github.com/((talos-systems|siderolabs)/[a-zA-Z0-9-]+)$"
 
 # previous release
-previous = "v1.6.0"
+previous = "v1.6.3"
```
smira, line 9: this shouldn't be updated, 1.7.0 will always consider 1.6.0 as the previous

pkg/machinery/config/configloader/internal/decoder/decoder.go
smira: I think this is still important

smira (commented):
I think we also need to update extensions documentation for v1.7

pkg/machinery/config/configloader/internal/decoder/decoder.go
frezbo: moved to #8270

website/content/v1.6/reference/api.md
```diff
@@ -3814,6 +3816,37 @@ EventSinkConfigSpec describes configuration of Talos event log streaming.
 
 
 
+<a name="talos.resource.definitions.runtime.ExtensionServicesConfigFile"></a>
```
smira, line 3819: this should be reverted back, v1.6 doesn't have it

smira (approved):

PR #8130, opened 2024-01-04
fix: fix nodes on dashboard footer when node names are used in `--nodes`

utkuozdemir:
When the dashboard is used via the CLI through a proxy, e.g., through Omni, node names or IDs can be used in the `--nodes` flag instead of the IPs.

This caused rendering inconsistencies in the dashboard, as some parts of it used the IPs and some used the names passed in the context.

Fix this by collecting all node IPs on dashboard start, and map these IPs to the respective nodes passed as the `--nodes` flag.

On the dashboard footer, we always display the node names as they are passed in the `--nodes` flag.

As part of it, remove the node list change reactivity from the dashboard, so it will always take the passed nodes as the truth.

The IP to node mapping collection at dashboard startup also solves another issue where the first API call by the dashboard triggered the interactive API authentication (e.g., the OIDC flow). Previously, because the terminal was already switched to the raw mode, it was not possible to authenticate properly.


frezbo (approved):

smira (commented):
If I understand the problem here correctly, it's that there's an extra layer (Omni) which converts the `node(s)` to some address, so the response field in the metadata contains some address we don't expect (as it doesn't match what we sent).

If we are to solve that, shouldn't we instead of trying to look into node's addresses (which is fragile to other changes or the way things are resolved), call some other "easy" API, e.g. `Version()` and grab the response from the metadata as the expected response from Talos API router? So this way we can map whatever we have in `node(s)` to what Talos API will enhance the response with?

utkuozdemir:
> If I understand the problem here correctly, it's that there's an extra layer (Omni) which converts the `node(s)` to some address, so the response field in the metadata contains some address we don't expect (as it doesn't match what we sent).
> 
> If we are to solve that, shouldn't we instead of trying to look into node's addresses (which is fragile to other changes or the way things are resolved), call some other "easy" API, e.g. `Version()` and grab the response from the metadata as the expected response from Talos API router? So this way we can map whatever we have in `node(s)` to what Talos API will enhance the response with?

Makes sense, I replaced the IP collection logic with a `Version()` call here: https://github.com/siderolabs/talos/pull/8130/files#diff-34c63597702e7f2ab227072fa7a1932d65b8ae823a38b3648f93c20befebaefeR511-R536

smira (approved):

PR #8145, opened 2024-01-12
feat: generate a single JSON schema for multidoc config

utkuozdemir:
Rework docgen to scan a whole directory for multidoc config types recursively and generate a single schema for all of them.

Annotate the files which need to be scanned by docgen while generating a schema by `//docgen:jsonschema`.

Move and rename the schema.

Bring back schema tests.


frezbo (approved):

pkg/machinery/config/schemas/config.schema.json
```diff
@@ -0,0 +1,3212 @@
+{
```
utkuozdemir, line 1: This is the new schema. Main changes are:
- Types in "defs" are now prefixed with their package: `AdminKubeconfigConfig` -> `v1alpha1.AdminKubeconfigConfig`, to ensure they stay unique in the multidoc world.
- As the root property, use `oneOf` instead of `$ref` (This schema will validate a single config doc at a time).

There are also some minor fixes in the schema since we last visited it.

pkg/machinery/config/config_schema_test.go
```diff
@@ -2,11 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package v1alpha1_test
+package config_test
 
-/* Commented out to workaround an issue with go1.20.5.
```
utkuozdemir, line 7: This file was totally commented out - brought it back and added tests for other document types (only `NetworkRuleConfig` for now)

hack/docgen/schema.go
```diff
@@ -9,18 +9,19 @@ import (
 	"fmt"
```
utkuozdemir, line 9: Main changes in docgen:
- Make it capable of generating regular docs (`*_doc.go`) and JSON schemas independently - you can now request one or the other, or both.
- Make it able to scan go package directories recursively to generate a single JSON schema from multiple packages with config structs inside them.

pkg/machinery/config/config.go
```diff
@@ -5,6 +5,8 @@
 // Package config provides methods to generate and consume Talos configuration.
 package config
 
+//go:generate docgen -generate-schema-from-dir types/ -json-schema-output schemas/config.schema.json -version-tag-file ../gendata/data/tag
```
utkuozdemir, line 8: The single, large image generation now happens here.

pkg/machinery/config/types/network/default_action_config.go
```diff
@@ -4,6 +4,8 @@
 
 package network
 
+//docgen:jsonschema
```
utkuozdemir, line 7: This is a marker to hint docgen to scan this file's structs for the JSON schema (files without this are ignored). Now that we scan all config dirs at once, listing each and every file as command line arg felt tedious, so I went with this way. Open for better ideas though.

pkg/machinery/config/types/network/default_action_config.go
```diff
@@ -36,6 +38,8 @@ var (
 //	examples:
 //	  - value: exampleDefaultActionConfigV1Alpha1()
 //	alias: NetworkDefaultActionConfig
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/NetworkDefaultActionConfig
```
utkuozdemir, line 42: Added these two new fields.
`schemaMeta` - hints docgen that this document has a required "meta" section which should match the apiVersion/Type in the value.
`schemaRoot` - it will end up in the `oneOf` section at the root of the JSON schema. e.g.:
```json
{
  "oneOf": [
    {
      "$ref": "#/$defs/network.DefaultActionConfigV1Alpha1"
    },
    {
      "$ref": "#/$defs/network.RuleConfigV1Alpha1"
    },
    {
      "$ref": "#/$defs/runtime.EventSinkV1Alpha1"
    },
    {
      "$ref": "#/$defs/runtime.KmsgLogV1Alpha1"
    },
    {
      "$ref": "#/$defs/siderolink.ConfigV1Alpha1"
    },
    {
      "$ref": "#/$defs/v1alpha1.Config"
    }
  ]
}

File: pkg/machinery/config/types/network/rule_config.go [hunk 7]
```diff
@@ -45,10 +47,13 @@ var (
 //	examples:
 //	  - value: exampleRuleConfigV1Alpha1()
 //	alias: NetworkRuleConfig
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/NetworkRuleConfig
 type RuleConfigV1Alpha1 struct {
 	meta.Meta `yaml:",inline"`
 	//   description: |
 	//     Name of the config document.
+	//   schemaRequired: true
```
utkuozdemir, line 56: This property makes docgen mark the field as required (for JSON schemas).

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1605,12 +1608,16 @@ type ResourcesConfig struct {
 	//   examples:
 	//     - name: resources requests.
 	//       value: resourcesConfigRequestsExample()
+	//   schema:
+	//     type: object
```
utkuozdemir, line 1612: small fix

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1605,12 +1608,16 @@ type ResourcesConfig struct {
...
 	Requests Unstructured `yaml:"requests,omitempty"`
 	//   description: |
 	//     Limits configures the maximum cpu/memory resources a container can use.
 	//   examples:
 	//     - name: resources requests.
 	//       value: resourcesConfigLimitsExample()
+	//   schema:
+	//     type: object
```
utkuozdemir, line 1620: small fix

pkg/machinery/go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/jsimonetti/rtnetlink v1.4.0
 	github.com/mdlayher/ethtool v0.1.0
 	github.com/opencontainers/runtime-spec v1.1.0-rc.1
+	github.com/santhosh-tekuri/jsonschema/v5 v5.3.1
```
smira, line 20: is this supposed to be here?

utkuozdemir: this is because I uncommented the json schema tests

smira: ah, it's probably part of the test...

smira: I'm always looking to keep the machinery deps to the absolute minimum, but I believe Go should not pull it in as it's only in the test

utkuozdemir: My IDE complained about it, let me remove and test it via our build mechanism.

utkuozdemir: ah ok then, that's my understanding as well.

Issue #8112, opened 2023-12-22
KubePrism goes away under load


alexandrem:
### Description

We've noticed that certain K8s worker nodes lose their connectivity towards the kube-apiserver through KubePrism when running under heavy load (thousands of pods during load tests).

It seems that the kubelet is not able to talk to KubePrism anymore.

### Logs

```
# NOTE only kept the bare metal machines in this output
$ k get nodes                                          
NAME                                 STATUS     ROLES           AGE   VERSION
talos-0qj-rku                        Ready      <none>          21h   v1.26.11
talos-0uz-2wc                        Ready      <none>          21h   v1.26.11
talos-282-nnw                        Ready      <none>          21h   v1.26.11
talos-316-05w                        Ready      <none>          21h   v1.26.11
talos-3dl-buq                        Ready      <none>          21h   v1.26.11
talos-41w-n9m                        Ready      <none>          21h   v1.26.11
talos-5b9-zy0                        NotReady   <none>          20h   v1.26.11
talos-5is-v65                        Ready      <none>          21h   v1.26.11
talos-5nb-8jg                        NotReady   <none>          21h   v1.26.11
talos-7x6-yx4                        NotReady   <none>          21h   v1.26.11
talos-8fd-pea                        Ready      <none>          21h   v1.26.11
talos-9sq-e1t                        Ready      <none>          21h   v1.26.11
talos-a61-2e4                        Ready      <none>          20h   v1.26.11
talos-ae4-qig                        Ready      <none>          20h   v1.26.11
talos-bq7-4vg                        Ready      <none>          21h   v1.26.11
talos-cgj-3lk                        Ready      <none>          21h   v1.26.11
talos-es0-6rn                        Ready      <none>          20h   v1.26.11
talos-gq0-voj                        Ready      <none>          20h   v1.26.11
talos-k2s-v5t                        Ready      <none>          21h   v1.26.11
talos-kr3-413                        Ready      <none>          21h   v1.26.11
talos-pre-wy5                        Ready      <none>          21h   v1.26.11
talos-qnh-9n4                        Ready      <none>          21h   v1.26.11
talos-r70-odn                        Ready      <none>          20h   v1.26.11
talos-tw0-rk6                        Ready      <none>          20h   v1.26.11
talos-v38-xvz                        Ready      <none>          21h   v1.26.11
talos-waa-xp7                        Ready      <none>          18h   v1.26.11
talos-wui-eek                        Ready      <none>          21h   v1.26.11
[...]

PR #8151, opened 2024-01-15
fix: force KubePrism to connect using IPv4

DmitriyMV:
Before this change KubePrism used hardcoded "localhost" as destination which Go could resolve to IPv6 destination and then fail to connect to. This change forces KubePrism to connect using IPv4 and uses hardcoded "127.0.0.1" destination so it will always use IPv4.

For #8112


smira (approved):

frezbo:
shouldn't the local ipv6 also work?

smira:
> shouldn't the local ipv6 also work?

KubePrism LB doesn't bind to both at the moment, as IPv6 is optional, the easiest fix is to force IPv4

frezbo (approved):

frezbo:
> > shouldn't the local ipv6 also work?
> 
> KubePrism LB doesn't bind to both at the moment, as IPv6 is optional, the easiest fix is to force IPv4

ahh that makes sense

PR #8156, opened 2024-01-16
fix: efi fat32 compatibility with nvme 4KBn sector size

aarnaud:
# Pull Request

## What? (description)
Fix the compatibility For Advanced Format 4K Native drives (4-KB-per-sector) drives, the minimum size is 260 MB, due to a limitation of the FAT32 file format.
## Why? (reasoning)
FAT32 spec :
- https://github.com/dosfstools/dosfstools/blob/master/src/mkfs.fat.c#L125
- https://www.cs.fsu.edu/~cop4610t/assignments/project3/spec/fatspec.pdf
- https://superuser.com/questions/1702331/what-is-the-minimum-size-of-a-4k-native-partition-when-formatted-with-fat32
- https://bbs.archlinux.org/viewtopic.php?id=168014

## Acceptance

### DIFF
--- a/internal/pkg/partition/constants.go
+++ b/internal/pkg/partition/constants.go
@@ -30,7 +30,9 @@ const (
 const (
 	MiB = 1024 * 1024
 
-	EFISize      = 100 * MiB
+	// The actual minimum size is 65595 × 4 KB(NVME sector size)
+	// = 268677120 bytes = 256.23046875 MB.
+	EFISize      = 260 * MiB
 	BIOSGrubSize = 1 * MiB
 	BootSize     = 1000 * MiB
 	// EFIUKISize is the size of the EFI partition when UKI is enabled.


smira:
Thanks for your PR, it does make a lot of sense. We need to consider on our side the effects of changing size of the partition for existing installs (when upgrading). It makes perfect sense for upgrades of course.

frezbo:
> Thanks for your PR, it does make a lot of sense. We need to consider on our side the effects of changing size of the partition for existing installs (when upgrading). It makes perfect sense for upgrades of course.

while we move to sd-boot for 1.7 we can maybe make it  bigger to accomodate UKI too? :thinking:

smira:
If we move to `sd-boot`, we have 1 GiB EFI partition there. This change is about GRUB-style partitioning scheme.

aarnaud:
I don't know if it's still require with 1.7 and sd-boot ?

smira:
Talos 1.10 switches to systemd-boot by default and 1GiB by default EFI partition, so I think we don't need this anymore.

PR #8158, opened 2024-01-17
fix: fix .der output in `talosctl gen secureboot`

utkuozdemir:
PEM was converted to DER incorrectly when the output was a X509 certificate and not a public key.

Skip unnecessary parsing of it to an RSA public key before writing it in DER format as output.

Simplify the code as we do not generate `*-signing-public-key.pem` anymore.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/gen/secureboot.go
+++ b/cmd/talosctl/cmd/mgmt/gen/secureboot.go
@@ -6,7 +6,6 @@ package gen
 
 import (
 	"context"
-	stdlibx509 "crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"io/fs"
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/mgmt/gen/secureboot.go
```diff
@@ -120,10 +119,8 @@ func generateSigningCerts(path, prefix, commonName string, rsaBits int, outputCe
 			return err
 		}
 
-		if outputDER {
```
smira, line 123: is it unused now?

utkuozdemir: Yes, it is unnecessary. It kinda made sense when we were outputting both the cert and the signing public key, but not anymore. It was never bound to a flag.

smira: ah, makes sense... I thought it's a flag.

smira (approved):

smira:
please `/promote integration-secureboot` once it's built

PR #8160, opened 2024-01-17
fix: watch bufer overrun for RouteStatus

### LINKED ISSUE #8157: 🐛 buffer overrun in RouteStatuses
This happens e.g. on very high rate of Cilium pod restarts.

smira:
Fixes #8157

This PR contains two fixes, both related to the same problem.

Several routes for different links but  same IPv6 destination might exist at the same time, so route resource ID should handle that. The problem was that these routes were mis-reported causing internally updates for the same resources multiple times (equal to the number of the links).

Don't trigger controllers more often than 10 times/seconds (with burst of 5) for kernel notifications. This ensures Talos doesn't try to reflect current state of the network subsystem too often as resources, which causes excessive CPU usage and might potentially lead to the buffer overrun under high rate of changes.


pkg/machinery/resources/network/network.go
```diff
@@ -36,17 +36,21 @@ func LinkID(linkName string) string {
...
 
 	if table != nethelpers.TableMain {
-		tablePrefix = fmt.Sprintf("%s/", table)
+		prefix = fmt.Sprintf("%s/", table)
 	}
 
-	return fmt.Sprintf("%s%s/%s/%s/%d", tablePrefix, family, string(gw), string(dst), priority)
+	if family == nethelpers.FamilyInet6 {
+		prefix += fmt.Sprintf("%s/", outLinkName)
+	}
```
utkuozdemir, line 51: is the issue only with ipv6? cannot the same problem happen with ipv4?

utkuozdemir (approved):

pkg/machinery/resources/network/network.go
smira: nope, kernel doesn't allow that (that's why I made the route ID this way initially), but IPv6 is special :)

pkg/machinery/resources/network/network.go
```diff
@@ -36,17 +36,21 @@ func LinkID(linkName string) string {
...
+	prefix := ""
 
 	if table != nethelpers.TableMain {
-		tablePrefix = fmt.Sprintf("%s/", table)
+		prefix = fmt.Sprintf("%s/", table)
 	}
 
-	return fmt.Sprintf("%s%s/%s/%s/%d", tablePrefix, family, string(gw), string(dst), priority)
+	if family == nethelpers.FamilyInet6 {
+		prefix += fmt.Sprintf("%s/", outLinkName)
```
frezbo, line 50: why do we add the link name now, i guess it was missed before?

smira: IPv6 allows onlink routes to the same address on different interfaces, so Talos incorrectly was trying to put all such route into a single `RouteStatus` resource causing an avalanche of useless updates, and mis-reporting routes as well.

PR #8164, opened 2024-01-17
chore: allow custom registry to build installer/imager

aarnaud:
# Pull Request

## What? (description)

use registry and username inside Dockerfile to allow build with custom registry

## Why? (reasoning)

impossible to `make imager` with specific kernel built and pkgs version

## Acceptance

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,90 +7,91 @@ ARG PKGS
 ARG PKG_KERNEL
 ARG EXTRAS
 ARG INSTALLER_ARCH
+ARG PKGS_PREFIX
 
 # Resolve package images using ${PKGS} to be used later in COPY --from=.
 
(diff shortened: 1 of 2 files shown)


smira (commented):
this won't work this way, as REGISTRY and USERNAME are specified for output images, not input images

the `ghcr.io/siderolabs` prefix can be customized, but it should be some other way

aarnaud:
Can we define a new variable ? like `INPUT_REGISTRY` do we need to split the registry and username ? 

```
INPUT_REGISTRY=ghcr.io/siderolabs
```

smira:
> Can we define a new variable ? like `INPUT_REGISTRY` do we need to split the registry and username ?
> 
> ```
> INPUT_REGISTRY=ghcr.io/siderolabs
> ```

I would probably call this `PKGS_PREFIX` specifically, so a package will be something like `${PKGS_PREFIX}/kernel:${PKGS}`

PR #8166, opened 2024-01-18
fix: strategic patch merging for audit policy

smira:
The audit policy is marked as `merge: replace`, but there's no check for zero value. So the problem is that any patch which has `cluster:` section zeroes out previously set `cluster.apiServer.auditPolicy`.

Add regression tests.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/testdata/strategic/005/expected.yaml
+++ b/pkg/machinery/config/types/v1alpha1/testdata/strategic/005/expected.yaml
@@ -1,0 +1,93 @@
+version: v1alpha1
+debug: false
+persist: true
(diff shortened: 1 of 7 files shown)


utkuozdemir (approved):

pkg/machinery/config/types/v1alpha1/testdata/strategic/005/expected.yaml
frezbo, line 1: these tests gets picked up automatically?>

smira: yep, it lists the directory to find all testcases

PR #8168, opened 2024-01-18
fix: use correct TTL for talosconfig in `talosctl config new`

smira:
See #8152


smira:
TBD: merge this once 1.7 docs are split out

frezbo (approved):

PR #8173, opened 2024-01-18
feat: implement device selector for 'physical'

smira:
Closes #8090


hack/release.toml
```diff
@@ -25,6 +25,12 @@ runc: 1.1.11
 Flannel: 0.24.1
 
 Talos is built with Go 1.21.6.
+"""
+
+    [notes.device_selectors]
+        title = "Device Selectors"
+        description = """\
+Talos now supports `physical: true` qualifier for device selectors.
```
frezbo, line 33: probably needs to be more clear what physcial is, like physically attached NIC's

frezbo (approved):
I'm not sure if the word `physical` makes sense exactly, but I don't have any better suggestions too

hack/release.toml
smira: yep, I will clarify this, thanks

PR #8188, opened 2024-01-22
fix: allow META encoded values to be compressed

### LINKED ISSUE #8186: Support bigger META values via `INSTALLER_META` environment variable
Talos support for compression of `INSTALLER_META` values to handle big META values which overflow kernel cmdline.

smira:
Fixes #8186

This is planned to be backported to Talos 1.6.3.

This allows to pass large META values (YAML for platform network configuration) which might otherwise exceed the limit for kernel command line params.

### DIFF
--- a/pkg/machinery/meta/meta.go
+++ b/pkg/machinery/meta/meta.go
@@ -6,8 +6,11 @@
 package meta
 
 import (
+	"bytes"
+	"compress/gzip"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


smira:
(needs some more docs updates and testing)

pkg/machinery/meta/meta.go
```diff
@@ -82,3 +123,11 @@ func DecodeValues(s string) (Values, error) {
 
 	return result, nil
 }
+
+func hasGzipMagic(b []byte) bool {
+	if len(b) < 2 {
+		return false
+	}
+
+	return b[0] == 0x1f && b[1] == 0x8b
```
frezbo, line 132: maybe a comment here from the spec?

smira: what kind of comment are you looking for?

smira: it's gzip magic number as the function name suggests!

https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/compress/gzip/gunzip.go;l=20-21

smira: but I see your point

pkg/machinery/meta/meta.go
```diff
@@ -66,6 +90,23 @@ func DecodeValues(s string) (Values, error) {
...
+		}
+
+		b, err = io.ReadAll(gzR)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := gzR.Close(); err != nil {
+			return nil, err
+		}
```
utkuozdemir, line 107: shouldn't we do this using `defer` - to make sure it closes unconditionally even if `ReadAll` fails?

utkuozdemir (approved):

pkg/machinery/meta/meta.go
smira: I think it'd be more clear, but I think it's not relevant, `.Close()` verifies the checksum, it actually doesn't close (like deallocate) anything

smira: updated with deferred Close

PR #8198, opened 2024-01-24
docs: update go structs with updated format for docgen

andrewrynhard:


smira (commented):
this would break machine config comments, JSON schema generation, etc. 

it doesn't cover new machine config documents either

andrewrynhard:
> this would break machine config comments, JSON schema generation, etc.
> 
> it doesn't cover new machine config documents either

- Config comments: They tend to be more confusing for users than helpful. I think we can solve our intention there was something else.
- JSON schema generation: this is still generated with https://github.com/siderolabs/documentation/tree/main/gen for the code and https://github.com/siderolabs/documentation/blob/main/ui/src/pages/docs/schema.json is the output. I can get it it published in the docs like it is now. Does it need to be in the Talos repo?
- Hmm I could have sworn I grabbed them in the merge. What is missing?

smira:
I think this need to be planned properly with exact ideas on what is going to be dropped, moved, removed, etc.  

Can we do it with more transitional changes and less breaking changes? 

How it's going to look in the end and so on. So far it won't pass the CI even.

andrewrynhard:
Picking this back up finally. Not ready for review yet but I want to use CI for testing. I plan on basing other PRs off of this one and going through a series of PRs that we can get aligned and push through.

andrewrynhard:
Not ready for review. Using CI to test.

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -43,300 +40,143 @@ func init() {
...
-	//    Note: In the default Kubernetes configuration, worker nodes are not allowed to
-	//    modify the taints (see [NodeRestriction](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#noderestriction) admission plugin).
-	//  examples:
-	//    - name: node taints example.
-	//      value: 'map[string]string{"exampleTaint": "exampleTaintValue:NoSchedule"}'
-	MachineNodeTaints map[string]string `yaml:"nodeTaints,omitempty"`
+	// Specifies the machine's role within the cluster. The roles can be either "controlplane" or "worker".
+	// The "controlplane" role hosts etcd and the Kubernetes control plane components such as API Server, Controller Manager, Scheduler.
+	// The "worker" role is available for scheduling workloads.
+	MachineType string `yaml:"type" docgen:"{'values':['controlplane','worker'],'in':'1.5'}"`
```
smira, line 73: not sure what `in` means, but this field was there forever, not since Talos 1.5

PR #8208, opened 2024-01-26
docs: document local path provisioner install

smira:
Use kustomize (as the official supported way for Local Path Provisioner).


website/content/v1.6/kubernetes-guides/configuration/local-storage.md
```diff
@@ -33,3 +33,51 @@ Both `EPHEMERAL` partition and user disks can be used for `hostPath` mounts.
...
+    metadata:
+      name: local-path-config
+      namespace: local-path-storage
+    data:
+      config.json: |-
+        {
+                "nodePathMap":[
+                {
+                        "node":"DEFAULT_PATH_FOR_NON_LISTED_NODES",
+                        "paths":["/var/local-path-provisioner"]
```
utkuozdemir, line 58: We don't need to mount this into kubelet as well, right? It was required for hostPaths as I remember.

smira: it seems to work without it just fine, I haven't tested everything of course, but at least this gives a better start option for those poor souls like myself ;)

utkuozdemir (approved):

PR #8209, opened 2024-01-26
docs: copy generated JSON schema to host

utkuozdemir:
After the JSON schema is generated in a build container, copy it over to the host, so it becomes a part of the codebase.

This is required as the location of the schema changed recently from being under `pkg/machinery/config/types/` to be under `pkg/machinery/config/schemas/`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -300,6 +300,7 @@ COPY --from=generate-build /api/inspect/*.pb.go /pkg/machinery/api/inspect/
 COPY --from=go-generate /src/pkg/flannel/ /pkg/flannel/
 COPY --from=go-generate /src/pkg/imager/profile/ /pkg/imager/profile/
 COPY --from=go-generate /src/pkg/machinery/resources/ /pkg/machinery/resources/
+COPY --from=go-generate /src/pkg/machinery/config/schemas/ /pkg/machinery/config/schemas/
 COPY --from=go-generate /src/pkg/machinery/config/types/ /pkg/machinery/config/types/
 COPY --from=go-generate /src/pkg/machinery/nethelpers/ /pkg/machinery/nethelpers/
 COPY --from=go-generate /src/pkg/machinery/extensions/ /pkg/machinery/extensions/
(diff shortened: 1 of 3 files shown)


andrewrynhard (approved):

andrewrynhard:
@utkuozdemir I could use some help in thinking through how we might handle the schema once the generation of it is outside of the code base. Or even before that, if it even makes sense to have it outside the codebase. I would like to opensource the docgen stuff as it's own project so pulling it out would be ideal but not crucial I suppose.

smira (approved):
should the `website/` version be updated as well?

utkuozdemir:
> @utkuozdemir I could use some help in thinking through how we might handle the schema once the generation of it is outside of the code base. Or even before that, if it even makes sense to have it outside the codebase. I would like to opensource the docgen stuff as it's own project so pulling it out would be ideal but not crucial I suppose.

I think it would make sense to be in a separate codebase, but only if we are going to make the tool much more generic and "proper". In its current form it's a bit of a mess. How much value moving it out would bring to us is indeed questionable though imo.

Integrating it to the current codebase shouldn't be difficult, we could treat it similarly to the other 3rd part tools in our Dockerfile (DEEPCOPY_GEN, VTPROTOBUF etc).

smira:
Moving to a different project makes development super hard, but it also depends on where the "source of truth" is - right now the Go structs for the configuration documents are the source of truth, so the tool is next to them, and it produces the result here. 

The result (JSON schema) can be easily copied to a different repo for consumption - that's what Omni does on build for example.

Separating build tools from repositories they process puts a huge toll on making any changes, so it should be justified by something big.

I don't think we can have any other source of truth than these Go structs, which live in this project. So whatever tools processes them and generates whatever output (whether it is documentation, JSON schema, or something else) should live in this repo as well. The generated version can be kept here as well, so it can be easily consumed for a specific tag/release/branch/whatever.

Whether the concern is the CI build time, we could optimize with filtering on changes which need to run the full stack or not.

smira:
And getting back to Utku's point - as this tooling is not generic at all, and depends heavily on the specific structure in this repo, making it a separate repo makes little sense, as we can't re-use it.

andrewrynhard:
> And getting back to Utku's point - as this tooling is not generic at all, and depends heavily on the specific structure in this repo, making it a separate repo makes little sense, as we can't re-use it.

There has been work done to standardize what `docgen` does. At this point I would consider it "generic" albeit for golang specific projects but still generic.

PR #8213, opened 2024-01-29
docs: provide sane defaults for each release series in vmware script

rsmitty:
This PR sets proper defaults based on the series of talos. Defaults to last release in each series.


frezbo (approved):
I wonder if we can use some templating or generate this smilar to our docs

smira (approved):

PR #8227, opened 2024-01-31
fix: retry blockdevice open in the installer

smira:
We had these retries in other places, but not here.

This seems to happen more frequently with Linux 6.6 update, the tl;dr is same: `udevd` tries to rescan the partition table at the wrong moment, preventing Talos installer to open the partition which was just created.

It's a race, so workaround it by retrying the call.


utkuozdemir (approved):

frezbo:
i really wish if there was a way to check with guarantee that udevd has settled, i guess this is just an async operation

PR #8231, opened 2024-02-01
chore: fix channel blocking on qemu launch

fazledyn-or:
# Pull Request
## What? (description)
This PR fixes a channel blocking bug in your code.

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/pkg/provision/providers/qemu/launch.go
+++ b/pkg/provision/providers/qemu/launch.go
@@ -394,6 +394,8 @@ func launchVM(config *LaunchConfig) error {
 				return fmt.Errorf("failed to kill process %w", err)
 			}
 
+			<-done
+
 			return fmt.Errorf("process stopped")
 		case err := <-done:
 			if err != nil {


smira:
Thanks for the fix, for anyone wondering it's a minor cosmetic issue, as the process is terminating, so the blocked goroutine is going to die anyway with the process.

smira (approved):
thank you!

smira:
## Why? (reasoning)
While triaging your project, our bug fixing tool generated the following warning-
```
// Possible blocking channel operation!
//
// In goroutine #1:
//	File: launch.go, Line: 385
//		done <- cmd.Wait()
//		In function launchVM$2, there is a send operation.
//		But no matching receive operation is found on that channel from any other goroutine.
```

After closer inspection, we found a `select` statement in file [launch.go]()-
```go
select {
case sig := <-config.c:
	fmt.Fprintf(os.Stderr, "exiting VM as signal %s was received\n", sig)

	if err := cmd.Process.Kill(); err != nil {
		return fmt.Errorf("failed to kill process %w", err)
	}

	return fmt.Errorf("process stopped")
case err := <-done:
	if err != nil {
		return fmt.Errorf("process exited with error %s", err)
	}

	// graceful exit
	return nil
case command := <-config.controller.CommandsCh():
	if command == VMCommandStop {
		fmt.Fprintf(os.Stderr, "exiting VM as stop command via API was received\n")

		if err := cmd.Process.Kill(); err != nil {
			return fmt.Errorf("failed to kill process %w", err)
		}

		<-done

		return nil
	}
}
```
This statement is run inside an infinite `for` loop. Inside the `select` statement, for case `sig := <-config.c`, if the process gets killed, then before returning `fmt.Errorf("process stopped")`, there can be contents inside the channel `done`. According to the [documentation](https://go.dev/ref/spec#Select_statements)- 
> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.

So, if the first and second case inside the `select` statement execute at the same time, and the first case is selected, then the `done` channel will halt unless a receive is performed. Which is why, we need to clear the contents of the channel by performing a receive operation. This will free the goroutine on the sending end.

PR #8236, opened 2024-02-01
feat: update runc 1.1.12, containerd 1.7.13

smira:
Also:

* Linux 6.6.14 + XDP enablement
* etcd 3.5.12

Various other bumps for the tools, utilities, and Go modules.


smira:
This should be the base to cut Talos 1.7.0-alpha.0

Unix4ever (approved):

PR #8256, opened 2024-02-04
fix: cancel all operations if `c.V1Alpha2().Run` fails.

DmitriyMV:
The current code contains a deadlock if `c.V1Alpha2().Run` fails with an error because `EnforceKSPPRequirements` will indefinitely wait for a condition to happen, but the controller runtime is not running. This commit addresses this by replacing the error channel with `context.WithCancelCause` and canceling the context if the controller runner fails.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/siderolabs/talos/internal/app/poweroff"
 	"github.com/siderolabs/talos/internal/app/trustd"
 	"github.com/siderolabs/talos/internal/app/wrapperd"
+	"github.com/siderolabs/talos/internal/pkg/ctxutil"
 	"github.com/siderolabs/talos/internal/pkg/mount"
 	"github.com/siderolabs/talos/pkg/httpdefaults"
 	"github.com/siderolabs/talos/pkg/machinery/api/common"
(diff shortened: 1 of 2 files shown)


internal/app/machined/main.go
```diff
@@ -195,7 +194,13 @@ func run() error {
 	go runDebugServer(ctx)
 
 	// Schedule service shutdown on any return.
-	defer system.Services(c.Runtime()).Shutdown(ctx)
+	s := system.Services(c.Runtime())
+	defer func() {
+		shutdownCtx, shutdownCtxCancel := context.WithTimeout(context.Background(), time.Minute)
```
DmitriyMV, line 199: This is required, because when we cancel the context, `Shutdown(context)` sequence below will never run properly.

smira:
we should rather fix the indefinite wait if we want to, but this place requires too much testing to get it right in all cases

DmitriyMV:
> we should rather fix the indefinite wait if we want to, but this place requires too much testing to get it right in all cases

@smira are there any valid scenarios where running of sequencer should continue if the runtime failed?

smira:
> > we should rather fix the indefinite wait if we want to, but this place requires too much testing to get it right in all cases
> 
> @smira are there any valid scenarios where running of sequencer should continue if the runtime failed?

probably not, but runtime failing correctly stops Talos except for the initialize sequence

DmitriyMV:
Closed, there should be better solution.

PR #8257, opened 2024-02-04
Update nocloud.md Proxmox information

b-:
# Pull Request

## What? (description)

This PR updates the documentation that suggested reading and editing the VM configuration by hand, and replaces that with CLI commands to do the same.

I also updated the UUID to be a real UUID, and then I tested these commands on my home Proxmox server to ensure they make the changes they're expected to.

## Why? (reasoning)

Proxmox _does_ support manually editing the configuration files, but a safer option is to use the CLI or API for the sake of option validation.

The `qm` command needs to be run from a root shell, but you need to be `root` to edit (or even read!) the configuration via something like SFTP, anyway, so I think this ultimately is a safer option than suggesting manual editing of the config file.

## Acceptance


smira (approved):
thank you!

PR #8262, opened 2024-02-05
chore: fix markdown-lint call

smira:
Don't ask me why this weird syntax for flags.

Don't ask me why it fails with exit code zero (success) on invalid flags.


utkuozdemir (approved):

smira:
> isn't `--` a bash thing

no, it's not... it's supposed to separate args from flags, but this is command itself

frezbo:
> > isn't `--` a bash thing
> 
> no, it's not... it's supposed to separate args from flags, but this is command itself

lol, omg, that's :hankey: then

PR #8265, opened 2024-02-05
chore: use "handle-like" resource in `DNSResolveCacheController`

DmitriyMV:
Rework (and simplify) `DNSResolveCacheController` to use `DNSUpstream` "handle-like" resources.

Depends on https://github.com/cosi-project/runtime/pull/400


internal/app/machined/pkg/controllers/network/dns_upstream.go
```diff
@@ -0,0 +1,162 @@
...
+// DNSUpstreamController is a controller that manages DNS upstreams.
+type DNSUpstreamController struct{}
+
+// Name implements controller.Controller interface.
+func (ctrl *DNSUpstreamController) Name() string {
+	return "network.DNSUpstreamController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *DNSUpstreamController) Inputs() []controller.Input {
```
Unix4ever, line 33: Maybe we case use `generic.Transform` controller here?

internal/app/machined/internal/server/v1alpha1/v1alpha1_inspect.go
```diff
@@ -44,6 +46,9 @@ func (s *InspectServer) ControllerRuntimeDependencies(ctx context.Context, in *e
 			edgeType = inspectapi.DependencyEdgeType_INPUT_WEAK
 		case controller.EdgeInputDestroyReady:
 			edgeType = inspectapi.DependencyEdgeType_INPUT_DESTROY_READY
+		case controller.EdgeInputQPrimary:
```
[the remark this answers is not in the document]
smira: we could at least return an error here, as Talos doesn't have QControllers (yet), we will hit this error if it does in the future

pkg/machinery/go.mod
```diff
@@ -9,7 +9,8 @@ replace gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20220527175918-f17b
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/containerd/go-cni v1.1.9
-	github.com/cosi-project/runtime v0.3.20
+	github.com/coredns/coredns v1.11.1
```
smira, line 12: that looks bad - we should never import it from machinery

internal/app/machined/pkg/controllers/network/dns_upstream.go
DmitriyMV: There is 1->M relationship here. I actually asked @smira the same question, and he suggested that I do in plain and simple.

internal/app/machined/internal/server/v1alpha1/v1alpha1_inspect.go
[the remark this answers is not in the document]
DmitriyMV: Linter complained about that switch, and I forgot to update it back to `nolint`. I actually don't know what should be here.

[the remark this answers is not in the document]
smira: return an error

pkg/machinery/go.mod
smira: so if the logic of the PR enforces this resource to reference a "handle" from coredns package, we should rather not to it at all :(

DmitriyMV: So I dropped the concrete type in "machinery" for the interface. It requires a small type-asssertion in the controllers, but otherwise it's pretty much the same. Except it machinery no longer needs "coredns".

We could use generics there, but in the absence of generic aliases the usage will look ugly.

internal/app/machined/internal/server/v1alpha1/v1alpha1_inspect.go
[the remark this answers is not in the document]
DmitriyMV: Done.

pkg/machinery/resources/network/dns_upstream.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Proxy is essentially a proxy.Proxy interface. It's here because we don't want machinery to depend on coredns.
+// The good thing we don't need any additional methods from coredns, so we can use a simple interface.
+type Proxy interface {
+	Addr() string
+	Fails() uint32
+	Healthcheck()
+	Stop()
+	Start(duration time.Duration)
+}
```
DmitriyMV, line 82: @smira `DNSUpstream` resource doesn't need anything concrete from `coredns` so that's good.

smira (approved):

Issue #8271, opened 2024-02-06
Support extra env vars as part of `ExtensionServicesConfig`


frezbo:
## Feature Request

Support extra env vars as part of `ExtensionServicesConfig`

PR #8273, opened 2024-02-06
feat: support environment in `ExtensionServiceConfig`

frezbo:
Support setting extension services environment variables in `ExtensionServiceConfig` document.

Fixes: #8271


Unix4ever (approved):

hack/release.toml
```diff
@@ -79,6 +79,23 @@ config:
         - content: MONITOR ${upsmonHost} 1 remote pass password
           mountPath: /usr/local/etc/nut/upsmon.conf
 ```
+"""
+
+ [notes.extensions-environment]
```
smira, line 84: it's a nit, but it might be easier to merge these two notes in a single item, as they both talk about same are - configuring extension services

pkg/machinery/config/types/extensionservicesconfig/extension_services_config.go
```diff
@@ -49,6 +49,7 @@ type V1Alpha1 struct {
 type ExtensionServiceConfig struct {
 	ExtensionName               string                       `yaml:"name"`
 	ExtensionServiceConfigFiles []ExtensionServiceConfigFile `yaml:"configFiles"`
+	ExtensionServiceEnvironment []string                     `yaml:"environment"`
```
smira, line 52: something is not wired properly, as I don't see the config reference documentation for this new config type

smira (approved):

pkg/machinery/config/types/runtime/extensions/service_config.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Name implements config.ExtensionServiceConfig interface.
+func (e *ServiceConfigV1Alpha1) Name() string {
+	return e.ServiceName
+}
+
+// ConfigFiles implements config.ExtensionServiceConfig interface.
+func (e *ServiceConfigV1Alpha1) ConfigFiles() []config.ExtensionServiceConfigFile {
+	return xslices.Map(e.ServiceConfigFiles, func(c ConfigFile) config.ExtensionServiceConfigFile {
+		return &ConfigFile{
```
smira, line 124: that's a nit, but looks like we could either do `return &c` here, or we could define methods `Content()` and `MountPath()` on non-pointer `ConfigFile` so we can avoid this copying

we still need `.Map()` to convert to the interface type though

PR #8282, opened 2024-02-07
feat: support systemd-boot ISO enroll keys option

### LINKED ISSUE #8196: Allow auto-enrollment of SecureBoot Keys into UEFI
## Feature Request

Currrently, the systemd-boot configuration is setup to only automatically enroll secure boot keys if it is booted in a VM:
https://github.com/siderolabs/talos/blob/main/pkg/imager/iso/loader.conf

In our environment we would like to set that option to `force` to always auto-enroll keys as long as the UEFI firmware is in setup mode, even when running bare-metal.

smira:
Fixes #8196

Example (profile excerpt):

```yaml
output:
  kind: iso
  isoOptions:
    sdBootEnrollKeys: force
  outFormat: raw
```

Defaults are still same (`if-safe` unless explicitly overridden).


pkg/imager/iso/uefi.go
```diff
@@ -54,6 +58,8 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 		return err
 	}
 
+	printf("preparing raw image")
```
frezbo, line 61: debug?

smira: nope, just some progress reporting ;)

pkg/imager/iso/uefi.go
```diff
@@ -75,6 +81,18 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
...
 	}
 
+	printf("preparing loader.conf")
+
+	var loaderConfigOut bytes.Buffer
+
+	if err := template.Must(template.New("loader.conf").Parse(loaderConfigTemplate)).Execute(&loaderConfigOut, struct {
+		SecureBootEnroll string
+	}{
+		SecureBootEnroll: options.SDBootSecureBootEnrollKeys,
```
frezbo, line 91: should we do some validation of input? I don;t know how sd-boot behaves for an invalid key

smira: yep, I could add that, good point

frezbo (approved):

frezbo:
maybe also something in the release notes?

pkg/imager/iso/uefi.go
smira: I will actually rework it to use constants, this will be way better

frezbo (approved):

Issue #8275, opened 2024-02-06
[upgrade-k8s] images paths overwrite by const while using private registries 


mstrohl:
## Bug Report
Unable to kee private registry image path on upgrade-k8s command
### Description
The command line "talosctl upgrade-k8s --to xxxx" replaced images repository to the default public registry instead of changing tag only
  
### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`]
1.6.3
- Kubernetes version: [`kubectl version --short`]
1.28.2 to 1.29.0
- Platform:
VM on OpenStack

smira:
Yes, this is not supported as of right now, but you have two ways to get around it:

* update k8s manually according to the [documentation](https://www.talos.dev/v1.6/kubernetes-guides/upgrading-kubernetes/#manual-kubernetes-upgrade)
* use [registry mirrors](https://www.talos.dev/v1.6/talos-guides/configuration/pull-through-cache/) so that the image references are "canonical", but the actual registry used if your private one

mstrohl:
Hello @smira! Thank you to answer. 
Registry mirrors have been setup but it seems to doesn't work like that:
```
    registries:
        mirrors:
            docker.io:
                endpoints:
                    - https://private-registry.example.com/harbor_folder/
                overridePath: true 
            gcr.io:
                endpoints:
                    - https://private-registry.example.com/harbor_folder/
                overridePath: true 
            ghcr.io:
                endpoints:
                    - https://private-registry.example.com/harbor_folder/
                overridePath: true 
            registry.k8s.io:
                endpoints:
                    - https://private-registry.example.com/harbor_folder/
                overridePath: true 
        config:
            private-registry.example.com:
                auth:
                    username: xxxxxx 
                    password: yyyyyyy 
                    auth: zzzzzz 
```
That is why we have forked and tested a change ton the code. The PR will come in a couple of minutes.

smira:
We have even docs for Harbor, I'm pretty sure it works: https://www.talos.dev/v1.6/talos-guides/configuration/pull-through-cache/#using-harbor-as-a-caching-registry

PR #8283, opened 2024-02-08
feat: custom image settings for k8s upgrade

mstrohl:
# Pull Request

Fixes: #8275 
## What? (description)
* Adding management of private registries for k8s upgrades
* set image before condition
* format

Due to the usage of authentication for image pullin adding this in the pre-pull feature was to big that is why we enable the feature only on --pre-pull=false.
It could be another feature. 

## Why? (reasoning)
The cli patch MachineConfig using public registries. This break automatic upgrade because of public registries overwriting.
 
## Acceptance

Co-author : @g3offrey

### DIFF
--- a/cmd/talosctl/cmd/talos/upgrade-k8s.go
+++ b/cmd/talosctl/cmd/talos/upgrade-k8s.go
@@ -49,6 +49,13 @@ func init() {
 	upgradeK8sCmd.Flags().BoolVar(&upgradeOptions.DryRun, "dry-run", false, "skip the actual upgrade and show the upgrade plan instead")
 	upgradeK8sCmd.Flags().BoolVar(&upgradeOptions.PrePullImages, "pre-pull-images", true, "pre-pull images before upgrade")
 	upgradeK8sCmd.Flags().BoolVar(&upgradeOptions.UpgradeKubelet, "upgrade-kubelet", true, "upgrade kubelet service")
+
+	upgradeK8sCmd.Flags().StringVar(&upgradeOptions.KubeletImage, "kubelet-image", constants.KubeletImage, "kubelet image to use")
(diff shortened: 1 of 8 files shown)


smira (commented):
this PR sounds good, but it has a subtle issue - previously `talosctl` would always set the proper image ref, but right now it will try to update the previous one.

upstream k8s images, as well as Talos-provided kubelet images were migrated a couple of times, and this change will break that migration

> k8s.gcr.io -> registry.k8s.io
> ghcr.io/talos-systems -> ghcr.io/siderolabs

smira:
the reason I brought this up is that _probably_ a more proper way is to use flags which give you base image references for each component, and you can override them if you want, but the tool still always doesn't patch the tag, but sets a complete reference.

also there's another ticket about pinning the images using digests (SHAs), which this PR would not support (but currently upgrade-k8s doesn't support them either)

mstrohl:
Another commit will be pushed to add a flag for upgrade options

mstrohl:
@smira I'm trying to manage proxy image too but I'm not sure where is managed .ProxyImage in DaemonSet template and if the manifest is managed differently between init and upgrade be cause as far as I know the version of the proxy image is base on talosctl default kubernetes version

mstrohl:
Hello, I just have tested new flag and proxy kubelet images management that works well. Any change to see the PR moving forward ? (FYI gpg has been generated in a wsl, and i'm can't understand what is appening on number of commits.

Regards

mstrohl:
Ok thank you, I wanted to wait this on to be merged before make a proposal on the digest capability base on that code.

smira:
@mstrohl please take a look, I reworked it to be a bit different, and more flexible

mstrohl:
@smira look good to me

utkuozdemir (approved):

frezbo (approved):
Maybe a release notes too?

smira:
> Maybe a release notes too?

probably, and also some words in the docs

PR #8294, opened 2024-02-09
docs: rpi talosctl install

### LINKED ISSUE #8129: Add arm64 version of talosctl installation on SBC pages.
## Feature Request
Add or Replace the amd64 talosctl command install version with the arm64 version.

### Description
On the installation pages of the SBC (example https://www.talos.dev/v1.6/talos-guides/install/single-board-computers/rpi_generic/) the amd64 version is given as an example of the to be installed version. Using an arm64 based install host (single architecture env.), i always end up with a wrong architecture version installed.

rothgar:
Replaces default architecture in install instructions

fixes #8129

# Pull Request

## Acceptance

### DIFF
--- a/website/content/v1.7/talos-guides/install/single-board-computers/rpi_generic.md
+++ b/website/content/v1.7/talos-guides/install/single-board-computers/rpi_generic.md
@@ -12,6 +12,7 @@ If you have tested this on other Raspberry Pi boards, please let us know.
 ## Video Walkthrough
 
 To see a live demo of this writeup, see the video below:
+
 <iframe width="560" height="315" src="https://www.youtube.com/embed/aHu1lFir7UU" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
 ## Prerequisites
@@ -24,8 +25,7 @@ You will need
 Download the latest `talosctl`.
 
 ```bash
-curl -Lo /usr/local/bin/talosctl https://github.com/siderolabs/talos/releases/download/{{< release >}}/talosctl-$(uname -s | tr "[:upper:]" "[:lower:]")-amd64
-chmod +x /usr/local/bin/talosctl
+curl -sL 'https://www.talos.dev/install' | bash
 ```
 
 ## Updating the EEPROM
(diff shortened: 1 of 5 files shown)


frezbo:
@rothgar can you add your keys here: https://github.com/siderolabs/signing-keys

website/content/v1.7/talos-guides/install/single-board-computers/rpi_generic.md
```diff
@@ -24,7 +25,7 @@ You will need
 Download the latest `talosctl`.
 
 ```bash
-curl -Lo /usr/local/bin/talosctl https://github.com/siderolabs/talos/releases/download/{{< release >}}/talosctl-$(uname -s | tr "[:upper:]" "[:lower:]")-amd64
+curl -Lo /usr/local/bin/talosctl https://github.com/siderolabs/talos/releases/download/{{< release >}}/talosctl-$(uname -s | tr "[:upper:]" "[:lower:]")-arm64
```
smira, line 28: I don't think this is right way to do it though, that is your client, so it might run different arch than your board. 

we also have https://www.talos.dev/v1.6/introduction/quickstart/#prerequisites

smira (commented):
the commit title should be `docs:` rather than `fix:`, as there is no user-visible fix to Talos itself

website/content/v1.7/talos-guides/install/single-board-computers/rpi_generic.md
rothgar: That's a good point. It appears the people who run into this problem are using a different pi to manage their pi install. Maybe another good reason to encourage people to use homebrew (automatic architecture parsing).

The only other option would be to add something like this at the end (which is ugly and prone to failure)
```bash
if [ $(uname -m) == "x86_64" ]; then echo "amd64"; else echo "arm64"; fi
```

rothgar: For now I'll just add a 2nd section that clearly states which architecture they're downloading and give both commands.
Because the code box in the site cuts off the back part of the command it's impossible to know which which you're downloading without scrolling.

smira: https://talos.dev/install this script should handle it all?

rothgar: TIL that was an option. I'll update to use that script.

smira (approved):

smira:
@rothgar also a small thing - even though you can crate branches at `origin` (this repo), we create PRs from the forks. 

PRs via origin branches work, but they trigger CI twice, and potentially confuse `go` tools

PR #8313, opened 2024-02-13
feat: support hardware watchdog timers

### LINKED ISSUE #8284: Watchdog timer (WDT) support
## Feature Request
Support hardware/firmware watchdog. This could increase resilience and help fulfill requirements some users might add to their systems.

### Description
Just as systemd is able to do it Talos init process should be able to arm and feed the watchdog if one is detected and configured by the user. If the system hangs (for example due to a driver crashing the kernel or hardware instability) WDT trips after not being fed for a timeout and resets the system, recovering it from a freeze. As hardware or firmware might be quirky I wouldn't risk enabling it in default config.

I want to try myself at implementing this feature in case it's approved.

dsseng:
# Pull Request

## What? (description)
Add support for Linux hardware watchdog API and corresponding configuration values. Watchdog timer is armed and fed by a controller (is it reliable to be sure a ticker will tick fine as long as no hang occurs?).

Watchdog timer can help automatically restoring a bare metal/supported VM node after a hang caused by a software or hardware error.

## Why? (reasoning)
Fixes #8284, approved by developers in the comments

## Acceptance

- [ ] you included tests (not applicable due to small footprint and mainly being ioctl-centered)
- [ ] you ran conformance (`make conformance` - not done due to certainly being unrelated)


internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			if wd == nil {
+				wd, err = os.OpenFile(dev, syscall.O_RDWR, 0o600)
+				if err != nil {
+					return errors.Errorf("failed to open watchdog device: %s", err)
+				}
+
+				logger.Info("Opened hardware watchdog", zap.String("filename", dev))
+				// TODO: support reboot/shutdown watchdogs
```
dsseng, line 117: We currently disable watchdog when quitting. systemd however allows for shutdown, reboot and kexec watchdogs to be set up to ensure power operations don't hang the system in inaccessible state. I could also do so if we clarify requirements for this and how are controllers being shut down during a restart or poweroff to be sure what timeout would be sane

smira: Controllers are terminated right before the reboot/shutdown (from the controller point of view, context will be cancelled).

dsseng: Okay, then I might add a prop for that after the first review. However shutdown/reboot watchdog won't be turned on by default as I'm not sure what are the safe values and they depend on target config.

dsseng:
Why are the CI checks failing? Does `conform/commit/gpg-identity` require me to add a GPG key and re-sign commits using it?

frezbo:
> Why are the CI checks failing? Does `conform/commit/gpg-identity` require me to add a GPG key and re-sign commits using it?

You can ignore the gpg check, it needs to be signed by one of our devs

dsseng:
No problem, I'll do needed improvements as soon as you request something. Thanks for reviewing!

dsseng:
Note: kernel changes are required to load watchdog drivers, I haven't committed those yet

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
```diff
@@ -0,0 +1,135 @@
...
+				defer wdClose()
+			}
+
+			if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, wd.Fd(), uintptr(unix.WDIOC_SETTIMEOUT), uintptr(unsafe.Pointer(&timeout))); err != 0 {
+				return fmt.Errorf("failed to set watchdog timeout: %w", err)
+			}
+
+			// 3 pings per timeout should suffice in any case
+			ticker = time.NewTicker(time.Duration(timeout/3) * time.Second)
```
smira, line 126: we don't stop old ticker here... do we even need a ticker?

we could do something like https://github.com/cosi-project/runtime/blob/main/pkg/controller/runtime/internal/qruntime/internal/timer/resettable.go, but for Tickers

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
```diff
@@ -0,0 +1,135 @@
...
+
+			if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, wd.Fd(), uintptr(unix.WDIOC_KEEPALIVE), uintptr(0)); err != 0 {
+				return fmt.Errorf("failed to feed watchdog: %w", err)
+			}
+
+			logger.Info("Set hardware watchdog timeout", zap.Int("timeout", timeout), zap.Int("feed_interval", timeout/3))
```
smira, line 132: log messages should be lowercase 
```suggestion
			logger.Info("set hardware watchdog timeout", zap.Int("timeout", timeout), zap.Int("feed_interval", timeout/3))
```

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
```diff
@@ -0,0 +1,135 @@
...
+				if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, wd.Fd(), unix.WDIOC_KEEPALIVE, 0); err != 0 {
+					return fmt.Errorf("failed to feed watchdog: %w", err)
+				}
+			}
+
+			continue
+		case <-r.EventCh():
+		}
+
+		r.ResetRestartBackoff()
```
smira, line 88: this should be at the bottom of the loop (once everything is good)

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
dsseng: Doesn't a ticker stop when GC'ed? Actually we can make use of resettable timer and reset it at each trigger, or develop an in-tree wrapper (or one contributed to the Cosi runtime) for ticker yes.

dsseng: Added ticker.Stop

dsseng:
Will update the kernel once more and rebase/squash stuff

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
smira: I meant to have a resettable Ticker here copying the approach from the above (to some extent).

The reason is that we don't want to have a ticker if there's no watchdog.

Or it could be implemented easier (there are plenty of examples in the codebase) by using a channel variable and setting it to nil when ticket is inactive.

internal/app/machined/pkg/controllers/runtime/watchdog_timer.go
```diff
@@ -0,0 +1,139 @@
...
+
+			if wd == nil {
+				wd, err = os.OpenFile(dev, syscall.O_RDWR, 0o600)
+				if err != nil {
+					return errors.Errorf("failed to open watchdog device: %s", err)
+				}
+
+				logger.Info("Opened hardware watchdog", zap.String("filename", dev))
+				// TODO: support reboot/shutdown watchdogs
+				defer wdClose()
```
smira, line 116: this is wrong, as if we open/close the wd multiple times, we'll have a chain of `defer`s attached;

we should have a single `defer wdClose()` top-level, and make it do nothing if `wd` is nil

smira:
@dsseng refactored controller a bit, other changes are config-related:

* moved configuration to a separate machine config document (we're slowly moving towards multi-doc)
* validation for the config, use proper time.Duration type
* introduced intermediate resource/controller to simplify the Watchdog controller

I will ask you to do integration test as a separate PR.

smira:
@dsseng added `WatchdogTimerStatus`, we can even try to read this:

```
All watchdog drivers are required return more information about the system,
some do temperature, fan and power level monitoring, some can tell you
the reason for the last reboot of the system.  The GETSUPPORT ioctl is
available to ask what the device can do:

	struct watchdog_info ident;
	ioctl(fd, WDIOC_GETSUPPORT, &ident);

the fields returned in the ident struct are:

        identity		a string identifying the watchdog driver
	firmware_version	the firmware version of the card if available
	options			a flags describing what the device supports
````

File: hack/release.toml [hunk 6]
```diff
@@ -155,6 +155,22 @@ machine:
     servers:
       - /dev/ptp0
 ```
+"""
+
+    [notes.watchdog]
+        title = "Hardware Watchdog Timers"
```
frezbo, line 161: i guess we also mention this is  only available for amd64?

pkg/machinery/config/types/runtime/event_sink.go
```diff
@@ -93,6 +93,11 @@ func (s *EventSinkV1Alpha1) KmsgLogURLs() []*url.URL {
 	return nil
 }
 
+// WatchdogTimer implements config.RuntimeConfig interface.
+func (s *EventSinkV1Alpha1) WatchdogTimer() config.WatchdogTimerConfig {
```
frezbo, line 97: hmm why does eventsink need watchdogtimer method implemented?

frezbo (approved):

dsseng:
> @dsseng added `WatchdogTimerStatus`, we can even try to read this:
> 
> ```
> All watchdog drivers are required return more information about the system,
> some do temperature, fan and power level monitoring, some can tell you
> the reason for the last reboot of the system.  The GETSUPPORT ioctl is
> available to ask what the device can do:
> 
> 	struct watchdog_info ident;
> 	ioctl(fd, WDIOC_GETSUPPORT, &ident);
> 
> the fields returned in the ident struct are:
> 
>         identity		a string identifying the watchdog driver
> 	firmware_version	the firmware version of the card if available
> 	options			a flags describing what the device supports
> ```

We should probably report that using sysfs to also know timeout and timeleft

hack/release.toml
dsseng: It is available on aarch64. We actually had all drivers built-in previously, so I didn't have to modify aarch64 config except for sysfs

frezbo: oh, were they built in? I guess we need to keep them also as modules if not, just to be in parity with amd64 :thinking: . I'm not so sure, @smira wdyt?

dsseng: https://github.com/siderolabs/pkgs/blob/6364d991db023e5f614cbbc450b4afd471e16203/kernel/build/config-arm64#L4441

Maybe some of those mustn't be modules due to the way they need to be probed, so that should be checked

frezbo: it supports PCI id database https://cateee.net/lkddb/web-lkddb/WATCHDOG.html, but we might hit other issue like we try to read before udev picks them up, I was more talking about these: 

```text
❯ grep IPMI_WA
kernel/build/config-arm64
3586:# CONFIG_IPMI_WATCHDOG is not set

kernel/build/config-amd64
3157:CONFIG_IPMI_WATCHDOG=m

❯ grep MFD_CORE
kernel/build/config-arm64
4514:CONFIG_MFD_CORE=y

kernel/build/config-amd64
3630:CONFIG_MFD_CORE=m

❯ grep SP5100
kernel/build/config-amd64
3581:CONFIG_SP5100_TCO=m
```

which are not consistent across amd64 and arm64

pkg/machinery/config/types/runtime/event_sink.go
smira: that's the way I implemented `Runtime()` config interface - all documents implement all methods, but return `nil` if it's not their thing. it might be good or bad idea, idk, but it is this way right now ;)

smira:
> We should probably report that using sysfs to also know timeout and timeleft

Timeout we already report (as we set it), while `timeleft` makes sense only in the moment, and the controller wakes up on `feedInterval`, so basically from controller point of view it's always `timeleft` == `timeout`, as we ping the watchdog.

hack/release.toml
smira: we can fix it up, but it shouldn't be related to to this PR specifically. 

I think arm64 in general needs a module cleanup the way I did for amd64 - try to make everything which can be a module to be a module.

arm64 kernel is not compressed, so moving to initramfs.xz is a good thing imho

smira (approved):

hack/release.toml
frezbo: yeh, arm64 needs some cleanup

dsseng: SP5100 is AMD WDT, MFD is for Intel one, but IPMI might work for ARM. Regarding PCI ID DB: aarch64 WDTs are typically part of the SoC

PR #8321, opened 2024-02-15
chore: update extensions test

frezbo:
Update extensions test to use strategic patches and use `ExtensionServiceConfig` document.

Depends on: https://github.com/siderolabs/extensions/pull/320


smira (approved):
:cool:

PR #8326, opened 2024-02-16
feat: allow access to all resources over siderolink in maintenance mode

utkuozdemir:
SideroLink is a secure channel, so we can allow read access to the resources. This will give us more control of the node via Omni and/or other systems using SideroLink.


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -142,7 +142,7 @@ func (s *Server) Register(obj *grpc.Server) {
 
 	// wrap resources with access filter
 	resourceState := s.Controller.Runtime().State().V1Alpha2().Resources()
-	resourceState = state.WrapCore(state.Filter(resourceState, resources.AccessPolicy(resourceState)))
+	resourceState = state.WrapCore(state.Filter(resourceState, resources.AccessPolicy(resourceState, false)))
```
smira, line 145: this works, but feels a bit like twisted flow.

as a proposal to simplify and streamline things, should we add a middleware which checks whether the peer is SideroLink, and injects `role.Admin` into the context?

this way in resource service things will "just work", while in methods which are only available over SideroLink we can assert on the role.Admin, and return the error otherwise?

utkuozdemir: changed it, pls have another look.

internal/app/maintenance/server.go
```diff
@@ -63,15 +64,25 @@ func (s *Server) Register(obj *grpc.Server) {
 
 	// wrap resources with access filter
 	resourceState := s.controller.Runtime().State().V1Alpha2().Resources()
-	resourceState = state.WrapCore(state.Filter(resourceState, resources.AccessPolicy(resourceState)))
+
+	// If the peer is a siderolink peer, allow access to all resources.
+	accessPolicyFilter := func(ctx context.Context, access state.Access) error {
```
smira, line 69: do we even need that given that the the siderolink user will have admin role injected?

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
```diff
@@ -117,11 +116,7 @@ func (ctrl *MaintenanceServiceController) Run(ctx context.Context, r controller.
 	srv := maintenance.New(cfgCh)
 
 	injector := &authz.Injector{
-		Mode: authz.ReadOnly,
-	}
-
-	if debug.Enabled {
```
smira, line 123: why was this removed?

internal/app/maintenance/server.go
```diff
@@ -130,16 +141,12 @@ func (s *Server) GenerateConfiguration(ctx context.Context, in *machine.Generate
...
 
 // GenerateClientConfiguration implements the [machine.MachineServiceServer] interface.
-func (s *Server) GenerateClientConfiguration(ctx context.Context, in *machine.GenerateClientConfigurationRequest) (*machine.GenerateClientConfigurationResponse, error) {
+func (s *Server) GenerateClientConfiguration(context.Context, *machine.GenerateClientConfigurationRequest) (*machine.GenerateClientConfigurationResponse, error) {
 	return nil, status.Error(codes.Unimplemented, "client configuration (talosconfig) can't be generated in the maintenance mode")
 }
 
 // Version implements the machine.MachineServer interface.
-func (s *Server) Version(ctx context.Context, in *emptypb.Empty) (*machine.VersionResponse, error) {
-	if err := assertPeerSideroLink(ctx); err != nil {
```
smira, line 139: what is the equivalent replacement for it?

smira (commented):
we need some unit-tests at least for it

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
utkuozdemir: changed it to be that way.

internal/app/machined/pkg/controllers/runtime/maintenance_service.go
utkuozdemir: was accidental, brought it back

internal/app/maintenance/server.go
utkuozdemir: I went a bit wrong direction on this - now everything should be in order, including the replacement for this.

internal/app/maintenance/server.go
utkuozdemir: nope, I reverted this part, as I changed the approach.

utkuozdemir:
> we need some unit-tests at least for it



> we need some unit-tests at least for it

Just added to the `maintenance_service_test.go`

PR #8333, opened 2024-02-19
feat: add custom config support to `talosctl cluster create`

utkuozdemir:
Allow passing a static, custom config to every machine created by `talosctl cluster create`. This is useful for testing, e.g., passing partial machine config via cloud user data.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -706,6 +706,17 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
 		return err
 	}
 
+	bundleTalosconfig := configBundle.TalosConfig()
+	if bundleTalosconfig == nil {
+		if clusterWait {
+			return errors.New("no talosconfig in the config bundle: cannot wait for cluster")
+		}
+
+		if applyConfigEnabled {
+			return errors.New("no talosconfig in the config bundle: cannot apply config")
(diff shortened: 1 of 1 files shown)


frezbo (approved):

frezbo:
We should probably plan later to support a config file like kind, we have too many options :sweat_smile:

smira (commented):
(thinking about having less flags)

would it work if we re-use `inputDir` flag, and pass same `controlplane.yaml`/`worker.yaml` without `talosconfig` ? (and add handling for empty talosconfig?)

utkuozdemir:
> (thinking about having less flags)
> 
> would it work if we re-use `inputDir` flag, and pass same `controlplane.yaml`/`worker.yaml` without `talosconfig` ? (and add handling for empty talosconfig?)

Yep, I thought about it as well. The main reason to go this way was to be able to pass small configs directly as flags instead of as files. Also they have a less change of being different across control plane and worker. Creating a directory structure to meet this need felt a bit too much for this use case.

I think this is more convenient to use, but I can change, no problem.

utkuozdemir:
> (thinking about having less flags)
> 
> would it work if we re-use `inputDir` flag, and pass same `controlplane.yaml`/`worker.yaml` without `talosconfig` ? (and add handling for empty talosconfig?)

Changed it to be that way, hence got significantly simpler.

smira (approved):
:cool:

PR #8348, opened 2024-02-23
chore: allow non preallocated disks for QEMU cluster

dsseng:
# Pull Request

## What? (description)
Added an option to disable `fallocate` calls on QEMU images

## Why? (reasoning)
Preallocation is still done by default for correct max usage estimates, but in development environment it could be beneficial not to use up that space, so I added a flag to disable preallocation and provision the image thinly.

## Acceptance


smira (approved):
:sunglasses:

smira:
One thing I forgot about is that the pre-allocate should be still enabled by default for the provisioning library. I will update

Issue #8131, opened 2024-01-04
Support Secure Boot in OVMF Environment


la7eralus:
## Feature Request
Support Secure Boot in OVMF Environment
### Description
I'm running Talos in Proxmox (QEMU), and want to enable Secure Boot and TPM Encryption. Unfortunately the Keys cant be enrolled automatically: 
![image](https://github.com/siderolabs/talos/assets/113297102/91a94018-104f-4992-9ba0-a6b3fd265107)

So I tried to import the keys manually (loader/keys/auto/*.auth):
![image](https://github.com/siderolabs/talos/assets/113297102/b33034ac-45d9-489e-873e-493f646dfc37)
![image](https://github.com/siderolabs/talos/assets/113297102/8fc10e1a-189d-4ab0-bcef-1f1e1f6219e8)

It seems like OVMF only supports Keys in der format (and not pem).

frezbo:
from the error it seems it failed to write the EFI variables, seems the firmware is passed in as readonly as opposed to read/write.

Talosctl supports generating keys in DER format `talosctl gen secureboot`

5olu7ion:
I created the UEFI disk without pre enrolled Keys, so I think it should be in setup mode?

Are you sure, that the default keys (the one that ship with the talks secure boot iso) are in DER format?

frezbo:
> Are you sure, that the default keys (the one that ship with the talks secure boot iso) are in DER format?

probably not the ones in the iso, but you can always generate your own keys

5olu7ion:
Yes, this seems doable, but it's ultimately forcing me to operate or provide a private registry for the custom built talos image. Is there a chance to provide the secure iso with keys, that work with OVMF?

frezbo:
Yeh, i think we can fix that

5olu7ion:
Awesome, thank you 😁

5olu7ion:
@frezbo This should be fixed with 1.6.2 right? -> https://github.com/siderolabs/talos/commit/3576d113c

frezbo:
yeh, the fix was not correct, will get it fixed

frezbo:
We still need to inlcude those files in the iso, I'll take care of that

frezbo:
@5olu7ion I took a look at the issue again, from the image it seems it reported a security violation
![image](https://github.com/siderolabs/talos/assets/18496730/8019f393-c419-4046-994e-13647532f767)

[...]

PR #8359, opened 2024-02-23
chore: uki der certs in iso

frezbo:
Add the uki signing cert into iso.

Fixes: #8131

### DIFF
--- a/pkg/imager/iso/uefi.go
+++ b/pkg/imager/iso/uefi.go
@@ -28,6 +28,9 @@ type UEFIOptions struct {
 	// A value in loader.conf secure-boot-enroll: off, manual, if-safe, force.
 	SDBootSecureBootEnrollKeys string
 
+	// UKISigningCertDer is the DER encoded UKI signing certificate.
+	UKISigningCertDerPath string
+
 	// optional, for auto-enrolling secureboot keys
 	PlatformKeyPath    string
 	KeyExchangeKeyPath string
@@ -116,6 +119,10 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 		return err
 	}
 
+	if _, err := cmd.Run("mmd", "-i", efiBootImg, "::EFI/keys"); err != nil {
+		return err
+	}
+
 	if _, err := cmd.Run("mmd", "-i", efiBootImg, "::loader"); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/imager/iso/uefi.go
```diff
@@ -167,6 +170,10 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 		}
 	}
 
+	if _, err := cmd.Run("mcopy", "-i", efiBootImg, options.UKISigningCertDerPath, "::EFI/Linux/uki-signing-cert.der"); err != nil {
```
smira, line 173: is it some standard path? or just dumping something?

frezbo: no standard path, i initially put it under `keys/secureboot` but it seems OVMF doesn;t allow selecting from that path, only `EFI` and `loader` were allowed

frezbo: maybe i can do `loader/keys`

smira: loader is I guess for sd-boot loader

frezbo: yeh, but it's `loader/auto`

frezbo: does `EFI/Linux/keys/uki-signing-key.der` makes sense?

pkg/imager/iso/uefi.go
```diff
@@ -149,6 +152,10 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 		return err
 	}
 
+	if _, err := cmd.Run("mcopy", "-i", efiBootImg, options.UKISigningCertDerPath, "::loader/keys/uki-signing-cert.der"); err != nil {
```
smira, line 155: my only concern is - should we move it to `::EFI/` so that we don't create something which would confuse systemd-boot

smira (approved):

pkg/imager/iso/uefi.go
frezbo: yeh, that works, so `::EFI/enrollment/uki-signing-key.der` or `::EFI/keys/uki-signing-key.der`?

smira: lgtm!

PR #8368, opened 2024-02-26
chore: listen for dns requests on 127.0.0.53

### LINKED ISSUE #8328: local dns resolver to listen on 127.0.0.53

DmitriyMV:
Turns out there is actually no black magic in systemd, they simply listen on 127.0.0.53 and forward dns requests there in resolv.conf.
Reason is the same as ours — to preserve compatibility with other applications. So we do the same in our code.

This PR also does two things:
- Adds `::1` into resolv.conf for IPv6 only resolvers.
- Drops `SO_REUSEPORT` from control options (it works without them).

Closes #8328

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
+++ b/internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
@@ -189,7 +189,7 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
 		&network.AddressStatusController{},
 		&network.DeviceConfigController{},
 		&network.DNSResolveCacheController{
-			Addr:   "127.0.0.1:53",
+			Addr:   "127.0.0.53:53",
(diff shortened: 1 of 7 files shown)


internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -189,7 +189,7 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
 		&network.AddressStatusController{},
 		&network.DeviceConfigController{},
 		&network.DNSResolveCacheController{
-			Addr:   "127.0.0.1:53",
+			Addr:   "127.0.0.53:53",
 			AddrV6: "[::1]:53",
```
frezbo, line 193: systemd uses 5353 for ipv6 i believe

DmitriyMV:
It's merged but github doesn't pick up this fact for some reason.

PR #8363, opened 2024-02-26
feat: enables imager to use private registry for pulling extensions

salkin:
# Pull Request

## What? (description)
Adds registry-credentials option to imager.

## Why? (reasoning)
When building Talos image with private extensions and access to registry is protected by authentication there is need for providing imager with credentials

## Acceptance

Without providing credentials:
```
docker run -t ghcr.io/nwik/imager:v1.6.0-alpha.1-195-g67ac6933d-dirty iso --system-extension-image MASKED/hardware-watchdog:1.27
profile ready:
arch: amd64
platform: metal
secureboot: false
version: v1.6.0-alpha.1-195-g67ac6933d-dirty
input:
  kernel:
    path: /usr/install/amd64/vmlinuz
  initramfs:
    path: /usr/install/amd64/initramfs.xz
  baseInstaller:
    imageRef: ghcr.io/nwik/installer:v1.6.0-alpha.1-195-g67ac6933d-dirty
    authconfig: null
  systemExtensions:
    - imageRef: MASKED/hardware-watchdog:1.27
      authconfig:
        username: ""
        password: ""
        auth: ""
        identitytoken: ""
        registrytoken: ""
output:
  kind: iso
  outFormat: raw
◱ error pulling image MASKED/hardware-watchdog:1.27: GET https://MASED/hardware-watchdog/manifests/1.27: UNAUTHORIZED: unauthorized to access repository: hardware-watchdog, action: pull: unauthorized to access repository: hardware-watchdog, action: pull
Error: error pulling image MASKED/hardware-watchdog:1.27: GET https://MASKED/hardware-watchdog/manifests/1.27: UNAUTHORIZED: unauthorized to access repository: hardware-watchdog, action: pull: unauthorized to access repository: hardware-watchdog, action: pull
```

When using credentials:
```
docker run -t ghcr.io/nwik/imager:v1.6.0-alpha.1-195-g67ac6933d-dirty iso --registry-credentials USER:PASS --system-extension-image MASKED/hardware-watchdog:1.27
profile ready:
arch: amd64
platform: metal
secureboot: false
version: v1.6.0-alpha.1-195-g67ac6933d-dirty
input:
  kernel:
    path: /usr/install/amd64/vmlinuz
  initramfs:
    path: /usr/install/amd64/initramfs.xz
  baseInstaller:
    imageRef: ghcr.io/nwik/installer:v1.6.0-alpha.1-195-g67ac6933d-dirty
    authconfig: null
  systemExtensions:
[...]

PR #8369, opened 2024-02-26
fix: provide auth when pulling images in the imager

smira:
Use standard Docker/Podman auth methods plus `GITHUB_TOKEN`.

See #8363


smira:
@salkin does this look good to you?

frezbo (approved):

salkin:
> @salkin does this look good to you?

looks good will try to use this is in our CICD

PR #8370, opened 2024-02-26
fix: ignore 'no such device' in addition to 'no such file'

smira:
This errors pops up when `udevd` rescans the partition table with Talos trying to mount a device concurrently.

This feels to be something new with Linux 6.6 probably.


smira:
> I wonder if we can rely on kobject later on and know exactly when to try/retry rather than relying on udevd

We don't rely on udevd, we do it right ourselves. The thing is that udevd does partition rescan ioctl once it sees Talos closing the file after writing the partition table (we can't avoid it). So if there was no `udevd`, the code would still work, it's `udevd` which messes up our workflow.

Issue #8350, opened 2024-02-23
Imager Overlay specification


smira:
Imager Overlay is a container image with the following structure:

```
.
├── artifacts
│   └── arm64
│       ├── dtb
│       │   └── rpi_4
│       ├── firmware
│       │   └── rpi4
│       │       └── firmware_20240302.bin
│       └── u-boot
│           └── rpi4.uboot.img
├── installers
│   └── rpi_generic
└── profiles
    ├── rpi_4.yaml
    └── rpi_5.yaml

12 directories, 3 files
```

An overlay can be passed to the `imager` an argument (part of the imager `Profile`), it is the first thing that imager will pull, unpack and use.

Imager Profile changes:

```yaml
...
overlay:
   image:
      imageRef: ghcr.io/siderolabs/overlay-rpi-generic:v24
      # ociPath: /overlays/rpi.oci
   name: rpi_generic
   options: # map[string]any
       bootconfig: |
           hdmi=2
```

When the overlay is specified, the `imager` will read extra profiles from `profiles/` directory and append them to the list of default profiles available.
       
After that `imager` works almost same way, but it has two hooks into the `installers` binary picked up by `overlay.name`:

```go
package overlay

// Provided by overlay installer
// Used by the imager/Talos installer
type Installer interface{
   GetOptions(extra map[string]any) (Options, error)
   Install(options InstallOptions) error
}

// Returned by the overlay installer
type Options struct{
   Name string
   KernelArgs []string
   PartitionOptions struct {
       Offset uint64
   }
}

// Passed to the overlay installer
type InstallOptions struct {
   InstallDisk string // device name
   MountPrefix string // mount path of new Talos root (e.g. /boot)
   ArtifactsPath string // path to `artifacts/` directory from an overlay
   ExtraOptions map[string]any // overlay.options in the Profile
}
```

The interface between `imager` and a binary will be command-line stdin/stdout based, described in more details below.

We will implement Go wrappers to implement this `overlay.Installer` interface via command-line for both `imager` and overlay installer. 

Draft of the command-line interface:

```shell
$ overlay-installer get-options
# STDIN << YAML
extraOptions: // map[string]any
   bootconfig: foo
# STDOUT >> YAML
name: rpi_4
kernelArgs:
  - console=ttyS1
...
# STDERR: error message (if failed)
invalid board
# EXIT CODE: 0 for no error, != 0 for error
```

```shell
$ overlay-installer install
# STDIN << YAML
[...]

PR #8377, opened 2024-02-28
feat: imager overlay

frezbo:
Support overlays for imager.
The `Install` interface is not wired yet, it will be done as a different PR.

This should be a no-op for existing imager.

Part of: #8350

### DIFF
--- a/cmd/installer/cmd/imager/root.go
+++ b/cmd/installer/cmd/imager/root.go
@@ -37,6 +37,8 @@ var cmdFlags struct {
 	OutputPath            string
 	OutputKind            string
 	TarToStdout           bool
+	OverlayName           string
+	OverlayImage          string
 }
 
 // rootCmd represents the base command when called without any subcommands.
@@ -72,6 +74,15 @@ var rootCmd = &cobra.Command{
 						ExtraKernelArgs: cmdFlags.ExtraKernelArgs,
 						MetaContents:    cmdFlags.MetaValues.GetMetaValues(),
 					},
+				}
+
+				if cmdFlags.OverlayName != "" || cmdFlags.OverlayImage != "" {
+					prof.Overlay = &profile.OverlayOptions{
+						Name: cmdFlags.OverlayName,
+						Image: profile.ContainerAsset{
+							ImageRef: cmdFlags.OverlayImage,
+						},
+					}
 				}
 
 				prof.Input.SystemExtensions = xslices.Map(
@@ -163,4 +174,8 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&cmdFlags.OutputPath, "output", "/out", "The output directory path")
 	rootCmd.PersistentFlags().StringVar(&cmdFlags.OutputKind, "output-kind", "", "Override output kind")
 	rootCmd.PersistentFlags().BoolVar(&cmdFlags.TarToStdout, "tar-to-stdout", false, "Tar output and send to stdout")
+	rootCmd.PersistentFlags().StringVar(&cmdFlags.OverlayName, "overlay-name", "", "The name of the overlay to use")
+	rootCmd.PersistentFlags().StringVar(&cmdFlags.OverlayImage, "overlay-image", "", "The image reference to the overlay")
+	rootCmd.MarkFlagsMutuallyExclusive("board", "overlay-name")
+	rootCmd.MarkFlagsMutuallyExclusive("board", "overlay-image")
 }
--- a/pkg/imager/imager.go
+++ b/pkg/imager/imager.go
@@ -10,21 +10,26 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/siderolabs/go-procfs/procfs"
+	"gopkg.in/yaml.v3"
 
-	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	talosruntime "github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/board"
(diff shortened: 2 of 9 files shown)


cmd/installer/cmd/imager/root.go
```diff
@@ -163,4 +174,8 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&cmdFlags.OutputPath, "output", "/out", "The output directory path")
 	rootCmd.PersistentFlags().StringVar(&cmdFlags.OutputKind, "output-kind", "", "Override output kind")
 	rootCmd.PersistentFlags().BoolVar(&cmdFlags.TarToStdout, "tar-to-stdout", false, "Tar output and send to stdout")
+	rootCmd.PersistentFlags().StringVar(&cmdFlags.OverlayName, "overlay-name", "", "The name of the overlay to use")
+	rootCmd.PersistentFlags().StringVar(&cmdFlags.OverlayImage, "overlay-image", "", "The image reference to the overlay")
+	rootCmd.MarkFlagsMutuallyExclusive([]string{"board", "overlay-name"}...)
+	rootCmd.MarkFlagsMutuallyExclusive([]string{"board", "overlay-image"}...)
```
smira, line 180:
```suggestion
	rootCmd.MarkFlagsMutuallyExclusive("board", "overlay-image")
```

cmd/installer/cmd/imager/root.go
```diff
@@ -74,6 +76,15 @@ var rootCmd = &cobra.Command{
 					},
 				}
 
+				if cmdFlags.OverlayName != "" || cmdFlags.OverlayImage != "" {
```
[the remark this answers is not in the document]
smira: okay, I saw below it defaults to `default` ;)

pkg/imager/imager.go
```diff
@@ -252,7 +329,7 @@ func (i *Imager) buildCmdline() error {
 
 	// board kernel args
 	if i.prof.Board != "" {
-		var b runtime.Board
+		var b talosruntime.Board
```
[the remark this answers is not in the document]
smira: or make it `// TODO:` if we don't want to break anything for now

pkg/imager/imager.go
```diff
@@ -342,3 +429,62 @@ func (i *Imager) buildUKI(ctx context.Context, report *reporter.Reporter) error
 
 	return nil
 }
+
+type overlayInstallerExecutor struct {
```
smira, line 433: let's move it out to a separate file, it could be an `overlay` subpackage in the `imager`, and add a type assertion on the interface to make it more clear what it does

pkg/imager/profile/profile.go
```diff
@@ -37,10 +37,22 @@ type Profile struct {
...
 	Input Input `yaml:"input"`
+	// Overlay describes overlay options for image generation.
+	Overlay *OverlayOptions `yaml:"overlay,omitempty"`
 	// Output describes image generation result.
 	Output Output `yaml:"output"`
 }
 
+// OverlayOptions describes overlay options for image generation.
+type OverlayOptions struct {
+	// Name of the overlay.
```
smira, line 48:
```suggestion
	// Name of the overlay installer, defaults to `default` if not set.
```

pkg/imager/profile/profile.go
```diff
@@ -67,7 +79,7 @@ func (p *Profile) Validate() error {
 	}
 
 	if p.Board != "" {
-		if !(p.Arch == arm64 && p.Platform == "metal") {
+		if p.Arch != arm64 || p.Platform != "metal" {
```
smira, line 82: we could also validate here that `Board` should not be set if we support overlays, or via TODO

pkg/machinery/overlay/adapter/adapter.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/siderolabs/talos/pkg/machinery/overlay"
+)
+
+// Execute executes the overlay installer.
+func Execute(installer overlay.Installer) {
+	if len(os.Args) < 2 {
```
smira, line 19: we could tidy up things a bit by calling from this function a function which returns an error:

```go
if err := execute(installer); err != nil {
   fmt.Fprint(os.Stderr, err.Error())
   os.Exit(1)
}
// success
```

this way we don't need to repeat this everywhere, and just return an error

pkg/imager/imager.go
```diff
@@ -342,3 +429,62 @@ func (i *Imager) buildUKI(ctx context.Context, report *reporter.Reporter) error
...
+	// parse extra as yaml
+	extraYAML, err := yaml.Marshal(extra)
+	if err != nil {
+		return overlay.Options{}, fmt.Errorf("failed to marshal extra: %w", err)
+	}
+
+	cmd := exec.Command(o.commandPath, cmdArgs...)
+	cmd.Stdin = bytes.NewReader(extraYAML)
+
+	cmd.Stderr = os.Stderr
```
smira, line 451: we should capture `stderr` I guess and report it as an error, but also we can tidy up things a bit by having a more generic command executor which accepts an arg, stdin, and optionally unmarshals stdout, handles the error, etc.

pkg/imager/imager.go
```diff
@@ -251,8 +327,8 @@ func (i *Imager) buildCmdline() error {
 	cmdline.SetAll(p.KernelArgs().Strings())
 
 	// board kernel args
-	if i.prof.Board != "" {
-		var b runtime.Board
+	if i.prof.Board != "" && !quirks.New(i.prof.Version).SupportsOverlay() {
```
smira, line 330: I guess this would break `/promote integration-images` as `make sbcs` would no longer work? 

not a big deal for me, but we should probably remove it from the `Makefile`?

smira (approved):

pkg/imager/imager.go
frezbo: yeh, i'll keep it commented

Issue #8351, opened 2024-02-23
Talos Imager: support overlays


smira:
Support new `Profile` options, pulling overlay images, running overlay installers.

Goal: to produce `rpi_4` image without built-in Talos support, only using overlay.

PR #8383, opened 2024-02-29
feat: implement `Install` for imager overlays

frezbo:
Implement `Install` for imager overlays.
Also add support for generating installers.

Depends on: #8377

Fixes: #8350
Fixes: #8351
Fixes: #8350

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -5,17 +5,20 @@
 package install
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 	"syscall"
 	"time"
 
 	"github.com/siderolabs/go-blockdevice/blockdevice"
 	"github.com/siderolabs/go-procfs/procfs"
 	"github.com/siderolabs/go-retry/retry"
+	"gopkg.in/yaml.v3"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/board"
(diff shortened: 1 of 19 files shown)


cmd/installer/pkg/install/install.go
```diff
@@ -68,7 +76,15 @@ func (m Mode) IsImage() bool {
 }
 
 // Install installs Talos.
-func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options) (err error) {
+//
+//nolint:gocyclo
+func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options) error {
+	if c := procfs.ProcCmdline().Get(constants.KernelParamBoard).First(); c != nil {
+		if *c != constants.BoardNone && quirks.New(version.Tag).SupportsOverlay() {
+			return fmt.Errorf("using standard installer image is not supported for board: %s, use an installer with overlay", *c)
```
smira, line 84: when upgrading from Talos 1.6 to Talos 1.7 with overlay, the board parameter will be present (from 1.6 Talos), so if we fail hard here, we'll block the upgrade path

cmd/installer/pkg/install/install.go
```diff
@@ -79,35 +95,36 @@ func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options)
 	cmdline.SetAll(p.KernelArgs().Strings())
 
 	// first defaults, then extra kernel args to allow extra kernel args to override defaults
-	if err = cmdline.AppendAll(kernel.DefaultArgs); err != nil {
+	if err := cmdline.AppendAll(kernel.DefaultArgs); err != nil {
 		return err
 	}
 
-	if opts.Board != constants.BoardNone {
```
smira, line 86: if we drop this, how would disk image creation work for Talos < 1.7?

smira (approved):

cmd/installer/pkg/install/install.go
frezbo: ahh I forgot `version.Tag` comes from `installer` and not from the one on disk, i think we still need to fail early rather than make the board unbootable

cmd/installer/pkg/install/install.go
frezbo: hmmm, i'll re-check

frezbo: Doesn't it use the installer from that specific version?

smira: Disk image creation works using `main` installer code.

smira:
As a result of these changes, I guess `make sbcs` should be broken, so we should remove it everywhere (?)

And we should drop SBC assets from the release step in Drone

smira (approved):
I would assume documentation update will be another PR?

frezbo:
> probably we need at least a note in `hack/release.toml` for an alpha release that SBC images are not supported at the moment

yeh, I can add that, docs can be another PR

PR #8390, opened 2024-03-05
chore: move support bundle generation code to machinery

Unix4ever:
This should allow using it in 3rd party projects without the need to pull the main Talos repo.


frezbo (approved):

Unix4ever:
will extract it to a separate repo as it pull lots of K8s dependencies

PR #8391, opened 2024-03-05
feat: implement Siderolink wireguard over GRPC.

### LINKED ISSUE #8064: HTTP tunnel for SideroLink
## Rationale

In some environments, UDP is not allowed (only TCP to specific ports), or even no outbound traffic is allowed, so traffic can only go via an HTTP(s) proxy.

SideroLink is a very valuable tool to perform remote management of Talos machines.

## Proposal

Use SideroLink API (a gRPC API used to provision Wireguard connection) as a channel to transport packets between Talos and SideroLink management endpoint.

gRPC API: use bi-directional stream gRPC API to send packet data.

### Using kernel Wireguard implementation

In the network stack, create a `tun` interface on both ends. 

Wireguard still works same way, as if there's no tunnel, so it provides encryption and authentication, and gRPC tunnel API doesn't have to provide that.

Wireguard packets as they leave the interface are routed to the `tun` device, where they get picked up by the userspace, sent over gRPC API to to the other end, injected into `tun` device, sent to the Wireguard for decryption.

### Using userspace Wireguard implementation

Force to send packets over gRPC API instead of injecting into the Linux network stack.

## Prior Art

* wstunnel
* https://www.reddit.com/r/WireGuard/comments/p6123n/wireguard_over_tcp_reviewing_existing_options_to/?utm_source=share&utm_medium=web2x&context=3

DmitriyMV:
Fixes #8064


go.mod
```diff
@@ -162,6 +164,11 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require (
+	golang.zx2c4.com/wireguard v0.0.0-20231211153847-12269c276173
```
frezbo, line 168: probably needs to go to main require block

go.mod
```diff
@@ -162,6 +164,11 @@ require (
...
+	gopkg.in/yaml.v2 v2.4.0
```
frezbo, line 169: yaml v3

internal/app/machined/pkg/controllers/siderolink/config.go
```diff
@@ -70,9 +71,17 @@ func (ctrl *ConfigController) Run(ctx context.Context, r controller.Runtime, _ *
...
+
+			parsed, err = endpoint.Parse(ep)
+			if err != nil {
+				return fmt.Errorf("failed to parse siderolink API endpoint: %w", err)
+			}
+
 			if err = safe.WriterModify(ctx, r, siderolink.NewConfig(config.NamespaceName, siderolink.ConfigID), func(c *siderolink.Config) error {
-				c.TypedSpec().APIEndpoint = endpoint
+				c.TypedSpec().APIEndpoint = ep
+				c.TypedSpec().Tunnel = parsed.GetParam("grpc_tunnel") != ""
```
frezbo, line 84: should we check the value? as in `?grpc_tunnel=false`? i don;t think we should ever hit that

internal/app/machined/pkg/controllers/siderolink/manager.go
```diff
@@ -235,6 +239,32 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
...
+			ourAddr, err = netip.ParseAddrPort(ctrl.pd.grpcPeerAddrPort)
+			if err != nil {
+				return err
+			}
+
+			if err = safe.WriterModify(ctx, r, siderolink.NewTunnel(),
+				func(tunnel *siderolink.Tunnel) error {
+					tunnel.TypedSpec().NodeAddress = ourAddr
+					tunnel.TypedSpec().APIEndpoint = ctrl.pd.apiEndpont
+					tunnel.TypedSpec().IFaceName = constants.SideroLinkName
```
smira, line 254:
```suggestion
					tunnel.TypedSpec().LinkName = constants.SideroLinkName
```

we use links everywhere, not interfaces

internal/app/machined/pkg/controllers/siderolink/config.go
```diff
@@ -70,9 +73,17 @@ func (ctrl *ConfigController) Run(ctx context.Context, r controller.Runtime, _ *
...
+			var parsed endpoint.Endpoint
+
+			parsed, err = endpoint.Parse(ep)
+			if err != nil {
+				return fmt.Errorf("failed to parse siderolink API endpoint: %w", err)
+			}
+
 			if err = safe.WriterModify(ctx, r, siderolink.NewConfig(config.NamespaceName, siderolink.ConfigID), func(c *siderolink.Config) error {
-				c.TypedSpec().APIEndpoint = endpoint
+				c.TypedSpec().APIEndpoint = ep
```
smira, line 85: there's a bit of lack of symmetry now here.

the API endpoint might contain a join token and a grpc_tunnel flag, we parse the tunnel flag, but not the API token

we should either parse both here or none

go.mod
DmitriyMV: Done!

go.mod
DmitriyMV: Fixed

internal/app/machined/pkg/controllers/siderolink/config.go
DmitriyMV: It now uses this check `parsed.GetParam("grpc_tunnel") == "true" || parsed.GetParam("grpc_tunnel") == "y"`

internal/app/machined/pkg/controllers/siderolink/userspace.go
```diff
@@ -0,0 +1,200 @@
...
+		tunnelDevice atomic.Pointer[wgtunnel.TunnelDevice]
+		tunnelRelay  atomic.Pointer[tunnelProps]
+	)
+
+	defer func() {
+		tunnelRelay.Load().Relay().Close()
+		tunnelDevice.Load().Close()
+	}()
+
+	qp := wgbind.NewQueuePair(100, 100)
```
smira, line 73: magic numbers

internal/app/machined/pkg/controllers/siderolink/userspace.go
```diff
@@ -0,0 +1,200 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create tunnel device: %w", err)
+			}
+
+			tunnelDevice.Store(td)
+
+			logger.Info("wg over grpc tunnel device created", zap.String("interface", res.TypedSpec().IFaceName))
+
+			eg.Go(func() error {
+				logger.Info("running tunnel device")
```
smira, line 107: should be `.Debug` ? why should user care about it?

internal/app/machined/pkg/controllers/siderolink/userspace.go
```diff
@@ -0,0 +1,200 @@
...
+
+			logger.Info(
+				"updating tunnel relay",
+				zap.String("old_endpoint", tn.DstHost()),
+				zap.Stringer("old_node_address", tn.OurAddrPort()),
+				zap.String("new_endpoint", dstHost),
+				zap.Stringer("new_node_address", ourAddrPort),
+			)
+
+			relay, err := wggrpc.NewRelayToHost(dstHost, 10*time.Second, qp, ourAddrPort, withTransportCredentials(ep.Insecure))
```
smira, line 133: magic numbers

internal/app/machined/pkg/controllers/siderolink/userspace.go
```diff
@@ -0,0 +1,200 @@
...
+
+	logger.WithOptions()
+
+	var (
+		tunnelDevice atomic.Pointer[wgtunnel.TunnelDevice]
+		tunnelRelay  atomic.Pointer[tunnelProps]
+	)
+
+	defer func() {
+		tunnelRelay.Load().Relay().Close()
```
smira, line 69: how does it work if the pointer is nil?

DmitriyMV: All the types in the chain support being called on a nil pointer. That was the whole idea.

internal/app/machined/pkg/controllers/siderolink/userspace.go
DmitriyMV: Thanks, fixed!

internal/app/machined/pkg/controllers/siderolink/userspace.go
```diff
@@ -0,0 +1,200 @@
...
+	return []controller.Output{}
+}
+
+// Run implements controller.Controller interface.
+//
+//nolint:gocyclo
+func (ctrl *UserspaceWireguardController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	eg, ctx := errgroup.WithContext(ctx)
+
+	logger.WithOptions()
```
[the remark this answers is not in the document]
DmitriyMV: Brain dump. Sorry, removed.

internal/app/machined/pkg/controllers/siderolink/userspace.go
DmitriyMV: Fixed!

internal/app/machined/pkg/controllers/siderolink/userspace.go
DmitriyMV: Thanks, added comments and constants.

internal/app/machined/pkg/controllers/siderolink/config.go
DmitriyMV: Fixed!

internal/app/machined/pkg/controllers/siderolink/manager.go
DmitriyMV: Done!

smira (approved):
add a constant for new ULA prefix

smira (changes requested):
too early

PR #8393, opened 2024-03-05
chore: move cli and version modules to machinery

Unix4ever:
To be used in the `go-talos-support` module without importing the whole Talos repo.


go.mod
```diff
@@ -162,6 +162,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require gopkg.in/yaml.v2 v2.4.0
```
smira, line 165: this should be replaced with .v3 (some wrong import)

go.mod
```diff
@@ -56,7 +56,7 @@ require (
 	github.com/docker/go-connections v0.5.0
 	github.com/dustin/go-humanize v1.0.1
 	github.com/ecks/uefi v0.0.0-20221116212947-caef65d070eb
-	github.com/emicklei/dot v1.6.1
+	github.com/emicklei/dot v1.6.1 // indirect
```
smira, line 59: should be moved to the indirect section below

smira (commented):
do we really need all of the `cli/` ? it was a pretty weird package with pretty different utilities...

Unix4ever:
> do we really need all of the cli/ ? it was a pretty weird package with pretty different utilities...

Maybe not. Can extract `formatters.go` only

go.mod
Unix4ever: I think it was there before me, it just got moved here somehow :)
But I'll fix that for sure

smira (approved):

PR #8399, opened 2024-03-06
fix: merge PodSecurity admission configs without duplicate exemptions

utkuozdemir:
When we strategically merge admission plugin configs of type "PodSecurity", we were concatenating the fields `namespaces`, `runtimeClasses` and `usernames`, which are string slices.

This did not pass the validation in the kube-apiserver and put it into a crash loop with errors like:
```
run.go:74] "command failed" err="failed to apply admission: failed to initialize admission plugin \"PodSecurity\": PodSecurity invalid: exemptions.namespaces[1]: Duplicate value: \"kube-system\"
```

Fix it by treating `PodSecurity` specially and not duplicating the elements in these fields while merging.

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/testdata/strategic/004/expected.yaml
+++ b/pkg/machinery/config/types/v1alpha1/testdata/strategic/004/expected.yaml
@@ -69,8 +69,15 @@ cluster:
                     namespaces:
                         - kube-system
                         - rook-system
-                    runtimeClasses: []
-                    usernames: []
+                    runtimeClasses:
+                        - rc-1
+                        - rc-2
+                        - rc-3
+                    usernames:
+                        - user-1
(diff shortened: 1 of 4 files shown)


utkuozdemir:
> hm... why would a patch contain a duplicate then?

Good question, I noticed this when I ended up applying a duplicate patch - seeing that this was causing the apiserver crash but everything else looked ok, I assumed that we should be handling this on merge.

But now I notice that this was not the only field that got duplicated - there were many more. So now it's clear to me that this is not the right solution - the patch is "broken" and shouldn't be applied at all. Apologies for the misunderstanding.

smira:
I think our patches should work with the "least surprise" model, so we could implement something via annotations and common merge to say "don't allow duplicate values in slices", but we'd need to document it I think

PR #8406, opened 2024-03-06
chore: use go-copy

frezbo:
Use go-copy and drop `pkg/copy`.


smira:
I guess it needs to wait before I bump tools and stuff... still working on that, I'll finish it tomorrow

PR #8412, opened 2024-03-08
feat: multiple docker clusters

### LINKED ISSUE #8141: Suport multiple clusters in single machine when using talos in docker
## Feature Request
I would like to be able to create multiple talos clusters in docker in the same machine. Something like: 
```
#creation of first cluster
talosctl cluster create --name cluster-1 --cidr 10.0.0.0/24
#creation of second cluster
talosctl cluster create --name cluster-2 --cidr 10.1.0.0/24
```

### Description
With the current state of talos, when I try to create multiple cluster specifying different CIDRs for each of the clusters:
```
#creation of first cluster
talosctl cluster create --name cluster-1 --cidr 10.0.0.0/24
#creation of second cluster
talosctl cluster create --name cluster-2 --cidr 10.1.0.0/24
```

I get the following error: 
```
creating controlplane nodes
1 error occurred:
	* Error response from daemon: driver failed programming external connectivity on endpoint cluster-2-controlplane-1 (2d9f0d9b4b4fdc138024dbe66457737e63b5470c6cc90a4cab934857be120163): Bind for 0.0.0.0:50000 failed: port is already allocated
```

The current problem is that the container is configuring the following bindings:
```
5000:5000
6443:6443
```
It is not configurable, so when the second cluster is created, it fails as it tries to use the same ports. 

As a user I would like to be able to have multiple clusters, without taking care of the bindings. For example when creating multiple kind clusters in the same machine, the binding is done automatically in different ports:
image

dsseng:
# Pull Request

## What? (description)
Making talosctl capable of creating multiple Docker clusters in case port forwarding is being used (Docker Desktop or WSL). Currently WIP

## Why? (reasoning)
Fixes #8141

## Acceptance

### DIFF
--- a/pkg/cluster/kubernetes.go
+++ b/pkg/cluster/kubernetes.go
@@ -15,7 +15,7 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	k8s "github.com/siderolabs/talos/pkg/kubernetes"
-	// "github.com/siderolabs/talos/pkg/machinery/constants"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
 // KubernetesClient provides Kubernetes client built via Talos API Kubeconfig.
@@ -25,6 +25,8 @@ type KubernetesClient struct {
 
 	// ForceEndpoint overrides default Kubernetes API endpoint.
 	ForceEndpoint string
+	// ForcePort overrides default Kubernetes API port
+	ForcePort int
 
 	KubeHelper *k8s.Client
 
@@ -68,7 +70,11 @@ func (k *KubernetesClient) K8sRestConfig(ctx context.Context) (*rest.Config, err
 	config.Timeout = time.Minute
 
 	if k.ForceEndpoint != "" {
-		config.Host = fmt.Sprintf("%s:%d", k.ForceEndpoint, 7444 /*constants.DefaultControlPlanePort*/)
+		port := k.ForcePort
+		if port == 0 {
+			port = constants.DefaultControlPlanePort
+		}
+		config.Host = fmt.Sprintf("%s:%d", k.ForceEndpoint, port)
 	}
 
 	return config, nil
--- a/pkg/cluster/kubernetes.go
+++ b/pkg/cluster/kubernetes.go
@@ -15,7 +15,7 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	k8s "github.com/siderolabs/talos/pkg/kubernetes"
-	"github.com/siderolabs/talos/pkg/machinery/constants"
+	// "github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
 // KubernetesClient provides Kubernetes client built via Talos API Kubeconfig.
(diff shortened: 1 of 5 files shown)


pkg/cluster/kubernetes.go
```diff
@@ -68,7 +68,7 @@ func (k *KubernetesClient) K8sRestConfig(ctx context.Context) (*rest.Config, err
 	config.Timeout = time.Minute
 
 	if k.ForceEndpoint != "" {
-		config.Host = fmt.Sprintf("%s:%d", k.ForceEndpoint, constants.DefaultControlPlanePort)
+		config.Host = fmt.Sprintf("%s:%d", k.ForceEndpoint, 7444 /*constants.DefaultControlPlanePort*/)
```
dsseng, line 71: Unsure how to properly plumb `controlPlanePort` from cluster create args into here

smira:
Nit: for this repo, please create a PR from your fork (vs. pushing a branch to this repo).

pkg/provision/providers/docker/docker.go
```diff
@@ -85,11 +85,11 @@ func (p *provisioner) GetLoadBalancers(networkReq provision.NetworkRequest) (int
 	case "darwin", "windows":
 		return "", "127.0.0.1"
```
smira, line 86: I guess we'd want all of the platforms to return this actually?

pkg/cluster/kubernetes.go
smira: I'd think we need to add a flag here which tells the control plane port, or even a full hostname:port pair

dsseng: `--control-plane-port` does override that, but port is not plumbed in here thus waiting for k8s pods fails due to failure to dial port 6443

smira: hold on, these are different

* `--control-plane-port` is the port API server listens on (in this case, _inside_ the container) (defaults to 6443)
* there's also internal LB (KubePrism) which is not accessible outside of the cluster (defaults to 7445)
* there's a port we mapped the `--control-plane-port` to outside of the cluster -- this should go into the `kubeconfig`

What should happen is:  `talosctl cluster create` in Docker mode should create a Docker port mapping from a random available host port (e.g. 34321) to `--control-plane-port` (e.g. 6443) in the container. Inside the container nothing changes, but outside the container `kubeconfig` points to `127.0.0.1:34321` now. So the `pkg/cluster` code needs a hint to know which one if the external endpoint, it could be a full hostname:port pair as well.

dsseng: `--control-plane-port` also affects to where the port is mapped, check https://github.com/siderolabs/talos/pull/8412/files#diff-6ed6be17858f4d4a230f75c46a3ade864e225e78d87760e13788db4bb58fe534R778

smira: this is something we need to change and map to a random available localhost port

the whole goal of the issue is to find two available ports on localhost, and then map Talos API and Kubernetes API to those ports.

dsseng: yes, but it doesn't work in the current condition as of main due to this preference not being forwarded to the status checkers

smira: see above - we need to implement a proper way to pass the new endpoint down the stack, one of the options I suggested is to accept full `hostname:port` pair as on override

pkg/provision/providers/docker/docker.go
dsseng: Unsure, why pollute local ports (also exposing them on 0.0.0.0) without a reason when we can just access those by IP. However a CLI option could by good (if it doesn't pollute opts)

smira: We should probably expose only on the localhost, and making things work same way across platforms makes it easier to test it. We don't have tests running on !Linux, so we won't be testing this code if we don't make it act same way.

smira:
> Nit: for this repo, please create a PR from your fork (vs. pushing a branch to this repo).

please re-create it from the fork, thanks!

dsseng:
> please re-create it from the fork, thanks!

Alright, will do now. Currently docker cluster startup fails for some reason, troubleshooting

PR #8420, opened 2024-03-11
chore: print all available logs containers in `logs` command completions

DmitriyMV:
This is a small quality of life improvement that allows `logs` subcommand to suggest all available logs.


DmitriyMV:
Looks like this 
image

internal/app/machined/pkg/runtime/logging/circular.go
```diff
@@ -41,6 +42,8 @@ type CircularBufferLoggingManager struct {
 	sendersRW      sync.RWMutex
 	senders        []runtime.LogSender
 	sendersChanged chan struct{}
+
+	registeredLogs containers.ConcurrentMap[string, struct{}]
```
smira, line 46: why do we need another map given that we already have `buffers` ?

api/api.descriptors
smira, line 1: we shouldn't be regenerating this every time (otherwise it makes zero sense), but only if we do a _breaking_ change that we know about

DmitriyMV: Reverted.

internal/app/machined/pkg/runtime/logging/circular.go
DmitriyMV: Somehow I missed that we can range over buffers even tho I saw that that was a `sync.Map`. Thanks, we now use `buffers` instead.

PR #8445, opened 2024-03-15
chore: re-enable DRBD extension

smira:
See https://github.com/siderolabs/extensions/pull/343


frezbo (approved):
pkgs needs to be bumped?

smira:
> pkgs needs to be bumped?

ah... yeah, because the build will pull it

frezbo (approved):

PR #8448, opened 2024-03-15
feat: add kernel version overriding

si0ls:
# Add Kernel version overriding

## What?

Add a Makefile argument `KERNEL_VERSION` thats override the DefaultKernelVersion in Talos codebase.

This PR is a draft as I'm not sure this is the best way to do, maybe some of you have better ideas.

## Why?

To allow easy upstream kernel version testing (with newest releases for example) or other taints.

## Acceptance


smira:
let's not go too crazy here I think, some changes require changes to the Talos source code, otherwise we'd push every option to the Makefile. 

if any, I would rather make Talos auto-detect Talos version in runtime where needed

PR #8449, opened 2024-03-15
fix: patch correctly config in `talosctl upgrade-k8s`

smira:
The current code was stipping non-`v1alpha1.Config` documents. Provide a proper method in the config provider, and update places using it.

### DIFF
--- a/pkg/machinery/config/provider.go
+++ b/pkg/machinery/config/provider.go
@@ -42,6 +42,9 @@ type Provider interface {
 	// Clone returns a copy of the Provider.
 	Clone() Provider
 
+	// PatchV1Alpha1 patches the container's v1alpha1.Config while preserving other config documents.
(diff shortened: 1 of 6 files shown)


pkg/machinery/config/provider.go
```diff
@@ -42,6 +42,9 @@ type Provider interface {
 	// Clone returns a copy of the Provider.
 	Clone() Provider
 
+	// PatchV1Alpha1 the container's v1alpha1.Config while preserving other config documents.
```
utkuozdemir, line 45:
```suggestion
	// PatchV1Alpha1 patches the container's v1alpha1.Config while preserving other config documents.
```

utkuozdemir (approved):

pkg/machinery/config/provider.go
smira: thanks, fixed

frezbo (approved):

Unix4ever (approved):

PR #8461, opened 2024-03-19
fix: service lifecycle issues

smira:
The core change is moving the context out of the `ServiceRunner` struct to be a local variable, and using a channel to notify about shutdown events.

Add more synchronization between Run and the moment service started to avoid mis-identifying not running (yet) service as succefully finished.


internal/app/machined/pkg/system/events/events.go
```diff
@@ -67,7 +70,7 @@ type ServiceEvent struct {
 func (event *ServiceEvent) AsProto(service string) *machineapi.ServiceStateEvent {
 	return &machineapi.ServiceStateEvent{
 		Service: service,
-		Action:  machineapi.ServiceStateEvent_Action(event.State),
+		Action:  machineapi.ServiceStateEvent_Action(event.State), // [TODO]: fix protobuf
```
frezbo, line 73: issue for merging?

smira: I actually updated, need to remove TODO

frezbo (approved):

PR #8476, opened 2024-03-20
feat: support multiple Docker cluster in talosctl cluster create

### LINKED ISSUE #8141: Suport multiple clusters in single machine when using talos in docker

dsseng:
Dynamically map Kubernetes and Talos API ports to an available port on
the host, so every cluster gets its own unique set of parts.

As part of the changes, refactor the provision library and interfaces,
dropping old weird interfaces replacing with (hopefully) much more
descriprive names.

Fixes #8141

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -617,38 +617,43 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
 			)
 		}
 
-		defaultInternalLB, defaultEndpoint := provisioner.GetLoadBalancers(request.Network)
-
-		if defaultInternalLB == "" {
-			// provisioner doesn't provide internal LB, so use first controlplane node
-			defaultInternalLB = ips[0][0].String()
-		}
+		externalKubernetesEndpoint := provisioner.GetExternalKubernetesControlPlaneEndpoint(request.Network, controlPlanePort)
 
 		if useVIP {
-			defaultInternalLB = vip.String()
+			externalKubernetesEndpoint = "https://" + nethelpers.JoinHostPort(vip.String(), controlPlanePort)
 		}
 
-		var endpointList []string
+		provisionOptions = append(provisionOptions, provision.WithKubernetesEndpoint(externalKubernetesEndpoint))
+
+		endpointList := provisioner.GetTalosAPIEndpoints(request.Network)
 
 		switch {
-		case defaultEndpoint != "":
-			if forceEndpoint == "" {
-				forceEndpoint = defaultEndpoint
-			}
-
-			fallthrough
 		case forceEndpoint != "":
-			endpointList = []string{forceEndpoint}
 			// using non-default endpoints, provision additional cert SANs and fix endpoint list
-			provisionOptions = append(provisionOptions, provision.WithEndpoint(forceEndpoint))
+			endpointList = []string{forceEndpoint}
 			genOptions = append(genOptions, generate.WithAdditionalSubjectAltNames(endpointList))
 		case forceInitNodeAsEndpoint:
 			endpointList = []string{ips[0][0].String()}
-		default:
+		case len(endpointList) > 0:
+			for _, endpointHostPort := range endpointList {
+				endpointHost, _, err := net.SplitHostPort(endpointHostPort)
+				if err != nil {
+					endpointHost = endpointHostPort
+				}
+
+				genOptions = append(genOptions, generate.WithAdditionalSubjectAltNames([]string{endpointHost}))
+			}
+		case endpointList == nil:
 			// use control plane nodes as endpoints, client-side load-balancing
 			for i := range controlplanes {
 				endpointList = append(endpointList, ips[0][i].String())
 			}
+		}
+
+		inClusterEndpoint := provisioner.GetInClusterKubernetesControlPlaneEndpoint(request.Network, controlPlanePort)
+
+		if useVIP {
+			inClusterEndpoint = "https://" + nethelpers.JoinHostPort(vip.String(), controlPlanePort)
 		}
 
 		genOptions = append(genOptions, generate.WithEndpointList(endpointList))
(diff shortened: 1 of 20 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -646,12 +684,13 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
...
 		}
 
+		endpointList = []string{fmt.Sprintf("%s:%d", forceEndpoint, apiPort)}
 		genOptions = append(genOptions, generate.WithEndpointList(endpointList))
 		configBundleOpts = append(configBundleOpts,
 			bundle.WithInputOptions(
 				&bundle.InputOptions{
 					ClusterName: clusterName,
-					Endpoint:    fmt.Sprintf("https://%s", nethelpers.JoinHostPort(defaultInternalLB, controlPlanePort)),
+					Endpoint:    fmt.Sprintf("https://%s", nethelpers.JoinHostPort(defaultEndpoint, externalControlPlanePort)),
```
dsseng, line 693: Unsure why `defaultInternalLB` was here and what could possibly break

smira: this is in-cluster Kubernetes endpoint (not outside of cluster): 

* so in case of docker it is `10.5.0.2:6433` (first controlplane container:6443)
* in case of QEMU it's `10.5.0.1:6443` loadbalancer listening on the bridge

we shouldn't be changing it

smira: the way this thing works is that inside the cluster the endpoint is "normal", but `ForceEndpoint` overrides the IP address of the endpoint when we pull `kubeconfig` out of the user.

So in case of Docker, with plain defaults it looked like that (before your changes):

* inside cluster: `https://10.5.0.2:6443`
* outside (for the user): `https://127.0.0.1:6443`

With your changes, "inside" should stay the same, while outside will be `https://127.0.0.1:<dynamic_allocated_port>`

pkg/provision/providers/docker/docker.go
```diff
@@ -85,11 +85,11 @@ func (p *provisioner) GetLoadBalancers(networkReq provision.NetworkRequest) (int
 	case "darwin", "windows":
 		return "", "127.0.0.1"
```
smira, line 86: this function should always return `"", "127.0.0.1"` no matter which runtime.GOOS to make it consistent

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -774,7 +813,7 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
 		}
 
 		if i == 0 {
-			nodeReq.Ports = []string{"50000:50000/tcp", fmt.Sprintf("%d:%d/tcp", controlPlanePort, controlPlanePort)}
+			nodeReq.Ports = []string{fmt.Sprintf("%d:50000/tcp", apiPort), fmt.Sprintf("%d:%d/tcp", externalControlPlanePort, controlPlanePort)}
```
smira, line 816: we can make this `if Docker` to make it more clear that it's not used by QEMU

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -440,6 +466,16 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
 		return err
 	}
 
+	apiPort, err := getDynamicTCPPort()
```
smira, line 469: we need to allocate only if the provisioner is docker...

smira: in fact, we could probably hide it more under the `provisioner` interface, and make only Docker one do that

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -646,12 +684,13 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
 			}
 		}
 
+		endpointList = []string{fmt.Sprintf("%s:%d", forceEndpoint, apiPort)}
```
smira, line 687: this is certainly a bug, this should be merged via the `case defaultEndpoint != ""` branch on line 667

the Docker provisioner should return "127.0.0.1:dynamic_port" as the `defaultEndpoint`

pkg/provision/providers/docker/docker.go
dsseng: And also perhaps we should make Docker only listen on 127.0.0.1?

smira: I didn't get that. Inside the container it will listen on a container port, outside the container should be only exposed to the localhost.

dsseng: Yes, but currently it forwards to 0.0.0.0 using `50000:50000/tcp` and so on

smira:
Moved to #8545

PR #8478, opened 2024-03-20
feat: report Docker node resources in cluster show

yadutaf:
# Pull Request

## What? (description)

For consistency with `talosctl cluster create`, this commit fetches the missing information through an inspect call for each node so that this information are available from a `Reflect` initialization.

## Why? (reasoning)

`talosctl cluster show` currently reports '-' in the CPU and RAM columns because this information are not available directly from the container list API call.

## Open questions

➡️ Placing a call to Docker inspect in this location does not feel right from an architecture point of view. It should probably be in `listNodes` instead, but it would change the signature from `[]types.Container` to `[]types.Container`. While this type is mostly a superset of the former, it will require adaptations to the existing callers. Let me know if you want me to do this.

## Acceptance

- [x] you ran conformance (`make conformance`)... But it complains with "openpgp: signature made by unknown entity"
- [x] you formatted your code (`make fmt`) --> Fails on 'api/*'

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	github.com/coreos/go-iptables v0.7.0
 	github.com/cosi-project/runtime v0.4.0-alpha.9
 	github.com/distribution/reference v0.5.0
-	github.com/docker/docker v25.0.4+incompatible
+	github.com/docker/docker v25.0.5+incompatible
 	github.com/docker/go-connections v0.5.0
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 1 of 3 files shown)


smira:
> ➡️ Placing a call to Docker inspect in this location does not feel right from an architecture point of view. It should probably be in listNodes instead, but it would change the signature from []types.Container to []types.Container. While this type is mostly a superset of the former, it will require adaptations to the existing callers. Let me know if you want me to do this.

I think this way is fine, as we only need details in `cluster show`

smira (approved):
thank you!

PR #8482, opened 2024-03-21
chore: import siderolink as `siderolink-launch` subcommand.

DmitriyMV:
This PR ensures that we can test our siderolink communication using embedded siderolink-agent. If `--with-siderolink` provided during `talos cluster create` `talosctl` will embed proper kernel string and setup `siderolink-agent` as a separate process. It should be used with combination of `--skip-injecting-config` and `--with-apply-config` (the latter will use newly generated IPv6 siderolink addresses which talosctl passes to the agent as a "pre-bind").


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -746,9 +750,38 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
...
+
+		request.SiderolinkRequest.WireguardEndpoint = net.JoinHostPort(wgHost, strconv.Itoa(wgPort))
+		request.SiderolinkRequest.APIEndpoint = ":" + apiPort
+		request.SiderolinkRequest.SinkEndpoint = ":" + sinkPort
+		request.SiderolinkRequest.LogEndpoint = ":" + logPort
+
+		apiLink := "grpc://" + net.JoinHostPort(wgHost, apiPort) + "?jointoken=foo"
+
+		extraKernelArgs.Append("siderolink.api", apiLink)
+		extraKernelArgs.Append("talos.events.sink", "[fdae:41e4:649b:9303::1]:"+sinkPort)
```
frezbo, line 776: probably makes sense to define this IP as a constant

pkg/provision/options.go
```diff
@@ -133,6 +133,15 @@ func WithKMS(endpoint string) Option {
 	}
 }
 
+// WithSideorlinkAgent enables or disables siderolink agent.
+func WithSideorlinkAgent(v bool) Option {
```
frezbo, line 137:
```suggestion
func WithSiderolinkAgent(v bool) Option {
```

.drone.jsonnet
```diff
@@ -644,6 +644,13 @@ local integration_cloud_images = Step('cloud-images', depends_on=[integration_im
 
 local integration_reproducibility_test = Step('reproducibility-test', target='reproducibility-test', depends_on=[load_artifacts], environment={ IMAGE_REGISTRY: local_registry });
 
+local integration_siderolink = Step('e2e-siderolink', target='e2e-qemu', privileged=true, depends_on=[load_artifacts], environment={
```
smira, line 647: it should `depends_on` on the previous set (default-hostname in this case)

cmd/talosctl/cmd/mgmt/cluster/create.go
smira, line 776: let's use proper `net.JoinHostPort` instead of this, and the IP we should know from the SideroLink library

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -746,9 +750,38 @@ func create(ctx context.Context, flags *pflag.FlagSet) error {
...
+	if withSiderolinkAgent {
+		if extraKernelArgs == nil {
+			extraKernelArgs = procfs.NewCmdline("")
+		}
+
+		if extraKernelArgs.Get("siderolink.api") != nil || extraKernelArgs.Get("talos.events.sink") != nil || extraKernelArgs.Get("talos.logging.kernel") != nil {
+			return errors.New("siderolink kernel arguments are already set, cannot run with --with-siderolink")
+		}
+
+		wgPort := rand.IntN(20000) + 40000
```
smira, line 762: instead of this, do something like https://github.com/siderolabs/talos/pull/8476/files#diff-6ed6be17858f4d4a230f75c46a3ade864e225e78d87760e13788db4bb58fe534R261-R284

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -1153,6 +1208,8 @@ func init() {
 	createCmd.Flags().IntVar(&bandwidth, "with-network-bandwidth", 0, "specify bandwidth restriction (in kbps) on the bridge interface when creating a qemu cluster")
 	createCmd.Flags().StringVar(&withFirewall, firewallFlag, "", "inject firewall rules into the cluster, value is default policy - accept/block (QEMU only)")
 	createCmd.Flags().BoolVar(&withUUIDHostnames, "with-uuid-hostnames", false, "use machine UUIDs as default hostnames (QEMU only)")
+	createCmd.Flags().BoolVar(&withSiderolinkAgent, "with-siderolink", false, "enables the use of siderolink agent as configuration apply mechanism")
+	createCmd.Flags().MarkHidden("with-siderolink") //nolint:errcheck
```
smira, line 1212: why hidden? we should have it

cmd/talosctl/cmd/siderolink/agent.go
```diff
@@ -0,0 +1,61 @@
...
+	joinToken         string
+	wireguardEndpoint string
+	sinkEndpoint      string
+	apiEndpoint       string
+	logEndpoint       string
+	predefinedPairs   []string
+}
+
+var siderolinkCmd = &cobra.Command{
+	Use:    "siderolink-launch",
```
smira, line 25: why do we need another folder? can we put it next to other `*-launch` commands, and we can make the whole file `//go:build linux` to have less moving parts

.drone.jsonnet
DmitriyMV: Done.

cmd/talosctl/cmd/mgmt/cluster/create.go
DmitriyMV: Done. We now use `GetAgentNodeAddr` which method which uses generated prefix from`wireguard.NetworkPrefix("")`.

cmd/talosctl/cmd/mgmt/cluster/create.go
DmitriyMV: Done. We now generate ports in `getDynamicPorts` which uses `getDynamicPort` and checks that generated ports do not overlap.

cmd/talosctl/cmd/mgmt/cluster/create.go
DmitriyMV: Removed.

cmd/talosctl/cmd/siderolink/agent.go
DmitriyMV: Done. Moved to `cmd/talosctl/cmd/mgmt/siderolink_launch_linux.go`

cmd/talosctl/cmd/mgmt/cluster/create.go
DmitriyMV: Done - we now use `GetAgentNodeAddr()` which method which uses generated prefix from `wireguard.NetworkPrefix("")`.

smira (approved):
:cool:

PR #8520, opened 2024-04-01
chore: add `ExtraInfo` field for extensions

frezbo:
Add an extra field to extensions to store arbitrary info.


smira:
`make generate docs` :)

smira (approved):

PR #8521, opened 2024-04-01
fix: don't set default endpoints on gen config

### LINKED ISSUE #8500: Do not set a default endpoint value in talosctl gen config
## Feature Request
Do not supply a default value for endpoints when creating talosconfig via `talosctl gen config`
### Description
``` 
talosctl gen config talos-test https://10.23.40.1:6443 
 more talosconfig 
context: talos-test
contexts:
    talos-test:
        endpoints:
            - 127.0.0.1
```
But connecting to loopback will never work, and results in an unclear error:
`rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:50000: connect: connection refused"`

It is better to not set the endpoint.
The current error is better:
`error constructing client: failed to determine endpoints`
A better error would be, if endpoints are not set, would be:
`endpoints are not set. Please pass in the endpoints to talosctl via the --endpoints flag, or set them in talosconfig`

smira:
Fixes #8500

### DIFF
--- a/cmd/talosctl/cmd/mgmt/gen/config.go
+++ b/cmd/talosctl/cmd/mgmt/gen/config.go
@@ -170,9 +170,6 @@ func GenerateConfigBundle(genOptions []generate.Option,
 		return nil, fmt.Errorf("failed to generate config bundle: %w", err)
 	}
 
-	// We set the default endpoint to localhost for configs generated, with expectation user will tweak later
-	configBundle.TalosConfig().Contexts[clusterName].Endpoints = []string{"127.0.0.1"}
-
 	return configBundle, nil
 }
 


cmd/talosctl/cmd/mgmt/gen/config.go
```diff
@@ -170,9 +170,6 @@ func GenerateConfigBundle(genOptions []generate.Option,
 		return nil, fmt.Errorf("failed to generate config bundle: %w", err)
```
frezbo, line 170: I guess we also need a error message if no endpoints is set, unless it's  already there?

frezbo (approved):

cmd/talosctl/cmd/mgmt/gen/config.go
smira: we don't need it, it's fine to have no endpoints (`talosctl` will complain when such `talosconfig` is used). 

from the point of view of generation, it's fine to have no endpoints. it just drops unfortunate default

DmitriyMV (approved):

PR #8524, opened 2024-04-01
feat: add support for static extra fields for JSON logs

### LINKED ISSUE #7356: Feature request: Include tags when sending JSON logs
### Description

Add a `talos-node` or `hostname` field to the service and kernel logs when sending them over TCP or UDP using json_lines.

The [Talos docs](https://www.talos.dev/v1.4/talos-guides/configuration/logging/) describe how to send logs e.g. to Loki using Fluent-bit. When logs from many nodes are sent to a centralised endpoint, there's no way to tell which node kernel and service logs came from. Adding a field with the node's name would solve this.

smira:
Fixes #7356

### DIFF
--- a/internal/app/machined/pkg/runtime/logging/sender_jsonlines.go
+++ b/internal/app/machined/pkg/runtime/logging/sender_jsonlines.go
@@ -13,10 +13,12 @@ import (
 	"time"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/pkg/machinery/config/config"
 )
 
 type jsonLinesSender struct {
-	endpoint *url.URL
+	endpoint  *url.URL
+	extraTags map[string]string
 
 	sema chan struct{}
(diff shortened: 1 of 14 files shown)


internal/app/machined/pkg/runtime/logging/sender_jsonlines.go
```diff
@@ -55,6 +59,10 @@ func (j *jsonLinesSender) marshalJSON(e *runtime.LogEvent) ([]byte, error) {
 	m["talos-time"] = e.Time.Format(time.RFC3339Nano)
 	m["talos-level"] = e.Level.String()
 
+	for k, v := range j.extraTags {
```
frezbo, line 62: should  we validate the standard ones are not overriden?

smira: I don't think so... if someone wants it... they get it?

frezbo (approved):

PR #8530, opened 2024-04-02
fix: allow platform cmdline args to be platform-specific

### LINKED ISSUE #8529: Talos platforms should support KernelCmdline per arch
## Feature Request

Talos platforms should support KernelCmdline per arch

### Description

For eg: with `equinixMetal` we have different console arguments for amd64 and arm64, so when requesting an arm64 asset we should generate then accordingly

smira:
Fix Equnix Metal (where proper arm64 args are known) and metal platform (using generic arm64 console arg).

Other platforms might need to be updated, but correct settings are not known at the moment.

Fixes #8529

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/equinixmetal/equinix.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/equinixmetal/equinix.go
@@ -110,9 +110,18 @@ func (p *EquinixMetal) Mode() runtime.Mode {
 }
 
 // KernelArgs implements the runtime.Platform interface.
-func (p *EquinixMetal) KernelArgs() procfs.Parameters {
-	return []*procfs.Parameter{
-		procfs.NewParameter("console").Append("ttyS1,115200n8"),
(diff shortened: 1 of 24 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/equinixmetal/equinix.go
```diff
@@ -110,9 +110,18 @@ func (p *EquinixMetal) Mode() runtime.Mode {
...
-		procfs.NewParameter("console").Append("ttyS1,115200n8"),
+func (p *EquinixMetal) KernelArgs(arch string) procfs.Parameters {
+	switch arch {
+	case "amd64":
+		return []*procfs.Parameter{
+			procfs.NewParameter("console").Append("ttyS1,115200n8"),
+		}
+	case "arm64":
+		return []*procfs.Parameter{
+			procfs.NewParameter("console").Append("ttyAMA0,115200"),
```
frezbo, line 121: this is documented here: https://deploy.equinix.com/developers/docs/metal/operating-systems/custom-ipxe/

frezbo (approved):

PR #8532, opened 2024-04-02
fix: force Flannel CNI to use KubePrism Kubernetes API endpoint

### LINKED ISSUE #8501: Default Flannel installation does not account for KubePrism endpoint
### Description

Talos 1.6.6 deploys Flannel and KubePrism by default, but does not update the Flannel arguments to redirect it to the KubePrism endpoint. The impact is that the Flannel pod will `CrashLoopBackoff` on any node that doesn't have connectivity to the control plane node on port `:6443`.

Patching the`DaemonSet` like this is one workaround:

```yaml
     containers:
      - image: ghcr.io/siderolabs/flannel:v0.23.0
        args:
        - --ip-masq
        - --kube-subnet-mgr
        command:
        - /opt/bin/flanneld
        env:
        - name: KUBERNETES_SERVICE_HOST       # kubeprism
          value: 127.0.0.1                    # kubeprism
        - name: KUBERNETES_SERVICE_PORT       # kubeprism
          value: "7445"                       # kubeprism
```

Passing an extra `args` would also do the trick:

```
  -kube-api-url string
    	Kubernetes API server URL. Does not need to be specified if flannel is running in a pod.
```

### Environment

- Talos version: `v1.6.6`
- Kubernetes version: `v1.29.3`

smira:
Fixes #8501


pkg/flannel/gen.go
```diff
@@ -91,6 +91,7 @@ func marshal(out io.Writer, obj runtime.Object) {
 	m = regexp.MustCompile(`status:\n(  .+\n)+`).ReplaceAll(m, nil)
 
 	m = regexp.MustCompile(`( +)- EXTRA_ARGS_PLACEHOLDER`).ReplaceAll(m, []byte("$1{{- range $$arg := .FlannelExtraArgs }}\n$1- {{ $$arg | json }}\n$1{{- end }}"))
+	m = regexp.MustCompile(`( +)- name: ETXRA_ENV_PLACEHOLDER`).ReplaceAll(m, []byte("$1{{- if .FlannelKubeServiceHost }}\n$1- name: KUBERNETES_SERVICE_HOST\n$1  value: {{ .FlannelKubeServiceHost | json }}\n$1{{- end }}\n$1{{- if .FlannelKubeServicePort }}\n$1- name: KUBERNETES_SERVICE_PORT\n$1  value: {{ .FlannelKubeServicePort | json }}\n$1{{- end }}"))
```
bgagnon, line 94:
```suggestion
	m = regexp.MustCompile(`( +)- name: EXTRA_ENV_PLACEHOLDER`).ReplaceAll(m, []byte("$1{{- if .FlannelKubeServiceHost }}\n$1- name: KUBERNETES_SERVICE_HOST\n$1  value: {{ .FlannelKubeServiceHost | json }}\n$1{{- end }}\n$1{{- if .FlannelKubeServicePort }}\n$1- name: KUBERNETES_SERVICE_PORT\n$1  value: {{ .FlannelKubeServicePort | json }}\n$1{{- end }}"))
```

pkg/flannel/gen.go
```diff
@@ -188,6 +189,9 @@ var Template = []byte(`+"`", url)
 			ds.Spec.Template.Spec.Containers[0].Args = append(ds.Spec.Template.Spec.Containers[0].Args,
 				"EXTRA_ARGS_PLACEHOLDER")
 
+			ds.Spec.Template.Spec.Containers[0].Env = append(ds.Spec.Template.Spec.Containers[0].Env,
+				corev1.EnvVar{Name: "ETXRA_ENV_PLACEHOLDER"})
```
bgagnon, line 193:
```suggestion
				corev1.EnvVar{Name: "EXTRA_ENV_PLACEHOLDER"})
```

frezbo (approved):

pkg/flannel/gen.go
smira: thanks, fixed

PR #8533, opened 2024-04-02
feat: add support for CoreDNS forwarding to host DNS

### LINKED ISSUE #8329: forward local DNS to the CoreDNS pod as upstream
Forward local DNS to the CoreDNS by listening on K8s service IP (need to research which exactly), but e.g. `10.96.0.9:53` (once the config is available) and use that in CoreDNS as an upstream.

smira:
This PR adds the support for CoreDNS forwarding to host DNS. We try to bind on 9th address on the first element from
`serviceSubnets` and create a simple service so k8s will not attempt to rebind it. Closes #8329 

Signed-off-by: Andrey Smirnov <andrey.smirnov@siderolabs.com>
Co-authored-by: Dmitriy Matrenichev <dmitry.matrenichev@siderolabs.com>


smira:
@DmitriyMV sample manifest we need to inject:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: host-dns
  namespace: kube-system
spec:
  clusterIP: 10.96.0.9
  ports:
  - name: dns
    port: 53
    protocol: UDP
    targetPort: 53
  - name: dns-tcp
    port: 53
    protocol: TCP
    targetPort: 53
  type: ClusterIP
```

internal/app/machined/pkg/controllers/network/hostdns_config.go
```diff
@@ -0,0 +1,202 @@
...
+		}); err != nil {
+			return fmt.Errorf("error writing host dns config: %w", err)
+		}
+
+		if oldServiceAddr == newServiceAddr {
+			continue
+		}
+
+		if !value.IsZero(oldServiceAddr) {
+			// new and old service addresses are different, so we need to destroy the old address spec
```
smira, line 130: this is a bit harder to manage than it should be, we can simplify it to dropping all addresspec resources we haven't touched:  https://github.com/siderolabs/talos/blob/5c0f74b377757ffee8f455ac2bcd9d81e3bf7717/internal/app/machined/pkg/controllers/siderolink/manager.go#L456-L482

DmitriyMV: Fixed!

smira (commented):
:+1:

DmitriyMV (approved):

PR #8537, opened 2024-04-03
fix: report unsupported x86_64 microarchitecture level

### LINKED ISSUE #8361: Avoid boot loop on old CPUs (and cputype-misconfigured VMs)
## Bug Report

Figuring out the boot loop and then ending at already existing bug [#7091](https://github.com/siderolabs/talos/issues/7091#issuecomment-1963107798) was needlessly difficult.
The issue appears even when running on recent-enough CPU, with misconfigured VM `cputype`.

### Description

Immediate reboot owerwrites the error message, and actually getting the log requires logging serial console to a file (e.g. running `qm terminal 123` in `screen -L`).

Any chance of patching [this](https://go.googlesource.com/go/+/refs/tags/go1.19.4/src/runtime/asm_amd64.s#366) in the `/init` process to just hang? Terminating the init process causes immediate reboot.
Or likely better way, doing the x86-64-v2 check in kernel and printing the error message and stopping there?

### Logs

needlessly difficult to obtain

### Environment

- Platform:

smira:
Fixes #8361

Talos requires v2 (circa 2008), but VMs are often configured to limit the exposed features to the baseline (v1).

```
[    0.779218] [talos] [initramfs] booting Talos v1.7.0-alpha.1-35-gef5bbe728-dirty
[    0.779806] [talos] [initramfs] CPU: QEMU Virtual CPU version 2.5+, 4 core(s), 1 thread(s) per core
[    0.780529] [talos] [initramfs] x86_64 microarchitecture level: 1
[    0.781018] [talos] [initramfs] it might be that the VM is configured with an older CPU model, please check the VM configuration
[    0.782346] [talos] [initramfs] x86_64 microarchitecture level 2 or higher is required, halting
```

### DIFF
--- a/internal/app/init/main.go
+++ b/internal/app/init/main.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/freddierice/go-losetup/v2"
+	"github.com/klauspost/cpuid/v2"
 	"github.com/siderolabs/go-kmsg"
 	"github.com/siderolabs/go-procfs/procfs"
 	"golang.org/x/sys/unix"
@@ -66,6 +67,8 @@ func run() (err error) {
 	}
 
 	log.Printf("booting Talos %s", version.Tag)
+
+	cpuInfo()
 
 	// Mount the rootfs.
 	if err = mountRootFS(); err != nil {
@@ -219,6 +222,24 @@ func bindMountExtra() error {
 	log.Printf("bind mounting %s", constants.SDStubDynamicInitrdPath)
 
 	return unix.Mount(constants.SDStubDynamicInitrdPath, filepath.Join(constants.NewRoot, constants.SDStubDynamicInitrdPath), "", unix.MS_BIND|unix.MS_RDONLY, "")
+}
+
+func cpuInfo() {
+	log.Printf("CPU: %s, %d core(s), %d thread(s) per core", cpuid.CPU.BrandName, cpuid.CPU.PhysicalCores, cpuid.CPU.ThreadsPerCore)
+
+	if runtime.GOARCH == "amd64" {
+		log.Printf("x86_64 microarchitecture level: %d", cpuid.CPU.X64Level())
+
+		if cpuid.CPU.X64Level() < constants.MinimumGOAMD64Level {
+			if cpuid.CPU.VM() {
+				log.Printf("it might be that the VM is configured with an older CPU model, please check the VM configuration")
+			}
+
+			log.Printf("x86_64 microarchitecture level %d or higher is required, halting", constants.MinimumGOAMD64Level)
+
+			time.Sleep(365 * 24 * time.Hour)
+		}
+	}
 }
 
 func main() {
--- a/internal/app/machined/pkg/controllers/block/internal/kobject/kobject.go
+++ b/internal/app/machined/pkg/controllers/block/internal/kobject/kobject.go
@@ -77,7 +77,7 @@ func (w *Watcher) Run(logger *zap.Logger) <-chan *Event {
 		for {
 			ev, err := w.cli.Receive()
 			if err != nil {
-				if err.Error() == "use of closed file" { // unfortunately not an exported error, just errors.New()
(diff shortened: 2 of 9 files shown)


smira:
Docs: https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

internal/app/machined/pkg/controllers/block/internal/kobject/kobject.go
```diff
@@ -77,7 +77,7 @@ func (w *Watcher) Run(logger *zap.Logger) <-chan *Event {
 		for {
 			ev, err := w.cli.Receive()
 			if err != nil {
-				if err.Error() == "use of closed file" { // unfortunately not an exported error, just errors.New()
+				if err.Error() != "use of closed file" { // unfortunately not an exported error, just errors.New()
```
frezbo, line 80: bug fix?

utkuozdemir (approved):

frezbo (approved):

internal/app/init/main.go
```diff
@@ -221,6 +224,24 @@ func bindMountExtra() error {
...
+		log.Printf("x86_64 microarchitecture level: %d", cpuid.CPU.X64Level())
+
+		if cpuid.CPU.X64Level() < constants.MinimumGOAMD64Level {
+			if cpuid.CPU.VM() {
+				log.Printf("it might be that the VM is configured with an older CPU model, please check the VM configuration")
+			}
+
+			log.Printf("x86_64 microarchitecture level %d or higher is required, halting", constants.MinimumGOAMD64Level)
+
+			time.Sleep(365 * 24 * time.Hour)
```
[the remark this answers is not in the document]
smira: now Go panics on `select {}` - it became smarter

internal/app/machined/pkg/controllers/block/internal/kobject/kobject.go
smira: yes, I fixed the error mapping, but the condition was reversed :facepalm:

pkg/machinery/constants/constants.go
```diff
@@ -992,6 +992,9 @@ const (
 
 	// SyslogListenSocketPath is the path to the syslog socket.
 	SyslogListenSocketPath = "/dev/log"
+
+	// MinimumGOAMD64Level is the minimum x86_64 microarchitecture level required by Talos.
+	MinimumGOAMD64Level = 2
```
giautm, line 997: Hello @smira, the doc said, I can use `GOAMD64=v1` to build talos for the old CPU arch. But look like this const never set to `1` base on the `GOAMD64` value.

https://github.com/siderolabs/talos/blob/bc8bf9e8a865473011dcf5ad2df08c8ff1af5110/website/content/v1.7/advanced/building-images.md?plain=1#L47-L52

smira: yes, you would need to change this field as well.

giautm: Thank you,

`GOAMD64=v1 GO_LDFLAGS="-X github.com/siderolabs/talos/pkg/machinery/constants.MinimumGOAMD64Level=1"` should work, right?

smira: I don't have a good way to test this, but I would assume so. We almost never had a user who actually had amd64-v1 device, but it was always misconfiguration of the VM assuming lower level of amd64.

giautm: I just want to use an old laptop to run Talos, its CPU was old, but still able to handle the workload for home lab. :)

giautm: I can confirm it works. However, I have to change the sourcecode to update `MinimumGOAMD64Level` to 1. It can't be change via `-X` (only work for variable, not constants).

Is there any change you can change it to a variable instead, I don't want to modify the source code. Just want build new version using build args.

giautm: @smira : Please help me review there: https://github.com/siderolabs/talos/pull/11790

PR #8538, opened 2024-04-03
fix: always update firewall rules (kubespan)

### LINKED ISSUE #8498: NfTables chain does not get updated with latest KubeSpan peer when `allowDownPeerBypass: true`
### Description

When KubeSpan is enabled with `allowDownPeerBypass: true`, `kubespan.ManagerController` does not update the `NfTablesChains` resources to include the IPs of the newly joined peer.

It seems to be caused by the combination of:
- The logic here: https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/kubespan/manager.go#L328
- And this micro-optimization: https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/kubespan/manager.go#L357-L362

### Environment

- Talos version: 1.6.4
- Kubernetes version: 1.28.6
- Platform: Metal

smira:
Fixes #8498

Before KubeSpan was reimplemented to use resources for firewall rules, the update was happening always, but it got moved to a wrong section of the controller which gets executed on resource updates, but ignores updates of the peer statuses.

### DIFF
--- a/internal/app/machined/pkg/controllers/kubespan/manager.go
+++ b/internal/app/machined/pkg/controllers/kubespan/manager.go
@@ -354,113 +354,9 @@ func (ctrl *ManagerController) Run(ctx context.Context, r controller.Runtime, lo
 			}
 		}
 
-		if !updateSpecs {
-			// micro-optimization: skip updating specs if there are no changes to the incoming resources and no endpoint changes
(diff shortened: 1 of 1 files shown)


utkuozdemir (approved):
🆒

PR #8545, opened 2024-04-04
feat: support multiple Docker cluster in talosctl cluster create

smira:
Dynamically map Kubernetes and Talos API ports to an available port on the host, so every cluster gets its own unique set of parts.

As part of the changes, refactor the provision library and interfaces, dropping old weird interfaces replacing with (hopefully) much more descriptive names.

Fixes https://github.com/siderolabs/talos/issues/8141


smira:
forked from #8476 as conform is picky about branch names :disappointed:

andrewrynhard (approved):

PR #8546, opened 2024-04-04
fix: make safeReset truly safe to call multiple times

DmitriyMV:
Reading documentation is important, because `timer.Stop()` explicitly says that it will return false if it already expired *OR* it has been already stopped. Previous version of code would block forever and because of that code tunnel relay never started.

Take that into account with new version.


smira:
Probably use this https://github.com/cosi-project/runtime/blob/main/pkg%2Fcontroller%2Fruntime%2Finternal%2Fqruntime%2Finternal%2Ftimer%2Fresettable.go?

smira (approved):
Let's make sure it passes integration-misc

PR #8553, opened 2024-04-05
docs: change localDNS to hostDNS in release notes yaml section

DmitriyMV:
Also add a note about how-to enable dns caching for k8s pods.


smira:
nit: should be `docs`, not `fix`

PR #8556, opened 2024-04-07
feat: machined, talosctl: enable listing xattrs

### LINKED ISSUE #1542: Add extended attributes support in ls
Something like `osctl ls --xattr` would be nice: https://github.com/pkg/xattr

dsseng:
# Pull Request

## What? (description)
Extended attributes are useful for SELinux labels and more

Tested in Docker on a SELinux-enforcing host (openSUSE)

## Why? (reasoning)
Fixes #1542 

## Acceptance


dsseng:
Will have to fix the CI...

smira (commented):
is there anything meaningful in xattrs except for SELinux labels for us?

In other words, should we limit it to SELinux labels only?

dsseng:
IMA/EVM and AppArmor use them as well. IMA/EVM has been requested afaik, and we should consider it someday

smira:
> IMA/EVM and AppArmor use them as well. IMA/EVM has been requested afaik, and we should consider it someday

So my 2 cents:

* it'd be easier to just have `list -l` show SELinux labels as a column vs. a separate output for xattrs
* how resource intensive is reading xattrs for every file even if we never use them?

dsseng:
> it'd be easier to just have list -l show SELinux labels as a column vs. a separate output for xattrs

Well, if we only consider SELinux, yes. We could also just avoid listing those and directly get `security.selinux`

> how resource intensive is reading xattrs for every file even if we never use them?

According to [source](https://github.com/pkg/xattr/blob/master/xattr_linux.go) there should be 1 syscall to list and 1 syscall per xattr to read. Anyway these API calls are not expected to be ran frequently.

smira:
it also allocates a buffer for each xattr read. 

what I mean is that make xattr reading optional, enable only if we list with `--long`, and for now in talosctl show just selinux label to make it less of a mess

dsseng:
Alright, will do

dsseng:
Will reword after review when merging

api/machine/machine.proto
```diff
@@ -491,6 +493,8 @@ message FileInfo {
   uint32 uid = 10;
   // Owner gid
   uint32 gid = 11;
+  // Label is filled with the value of `security.selinux` xattr if present and requested
+  string label = 12;
```
smira, line 497: I think on API level (as it's hard to change later), we should return all labels.

The CLI might report a single label for now

go.mod
```diff
@@ -177,6 +177,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require github.com/pkg/xattr v0.4.9
```
smira, line 180: should be cleaned up so that it goes into `require` block above

api/machine/machine.proto
```diff
@@ -451,6 +451,8 @@ message ListRequest {
   // Types indicates what file type should be returned. If not indicated,
   // all files will be returned.
   repeated Type types = 4;
+  // Report SELinux labels
+  bool labels = 5;
```
smira, line 455: probably this will be better? or `report_xattrs` ? and returns `xattrs` ?
```suggestion
  bool report_labels = 5;
```

api/machine/machine.proto
dsseng: > all labels

You mean all xattrs?

smira: yes, all xattrs

smira (approved):

cmd/talosctl/cmd/talos/list.go
```diff
@@ -148,13 +149,23 @@ var lsCmd = &cobra.Command{
 					}
 				}
 
-				fmt.Fprintf(w, "%s\t%s\t%d\t%d\t%s\t%s\t%s\n",
+				label := ""
+				if info.Xattrs != nil && len(info.Xattrs) > 0 {
```
stasfilin, line 153: Like for me - no need to add `len(info.Xattrs) > 0`

The check `len(info.Xattrs) > 0` is redundant because of the preceding condition `info.Xattrs != nil` already ensures that `info.Xattrs` is not nil. If `info.Xattrs` is not nil, then it is safe to iterate over it, even if it is an empty slice.

stasfilin: So `if info.Xattrs != nil {` will be enough

cmd/talosctl/cmd/talos/list.go
```diff
@@ -148,13 +149,23 @@ var lsCmd = &cobra.Command{
...
+					for _, l := range info.Xattrs {
+						if l.Name == "security.selinux" {
+							label = string(l.Data)
```
stasfilin, line 156: after this, maybe better to use `break`?
```go
        if l.Name == "security.selinux" {
            label = string(l.Data)
            break
        }
```
exit the loop once the desired Name is found, preventing unnecessary iterations.

cmd/talosctl/cmd/talos/list.go
dsseng: Thank you, fixed!

smira (approved):
lgtm, you probably need Noel's PR with CSI tests to merge first and then rebase

PR #8567, opened 2024-04-09
chore: replace []byte with string and use go:embed for templates

DmitriyMV:
Optimize code a bit.


internal/app/machined/pkg/controllers/k8s/templates.go
```diff
@@ -203,390 +56,34 @@ subjects:
...
-        health {
-            lameduck 5s
-        }
-        ready
-        log . {
-            class error
-        }
-        prometheus :9153
+//
+//go:embed templates/kube-config-in-cluster.yaml
```
Unix4ever, line 60: Maybe to keep up with the naming scheme?

```suggestion
//go:embed templates/kube-config-in-cluster-template.yaml
```

Unix4ever (approved):

PR #8575, opened 2024-04-10
test: add a test for 'spin' container runtime

smira:
See https://github.com/siderolabs/extensions/pull/355


utkuozdemir (approved):
first time heard about spin 🤔

smira:
> first time heard about spin 🤔

myself as well... but it works :-D

PR #8583, opened 2024-04-11
feat: use container DNS when in container mode

### LINKED ISSUE #8303: CoreDNS fails when using Talos in Docker with docker default nameserver
### Description

We were just debugging some DNS issues when running Talos locally as a docker container. Since `8.8.8.8` is not accessible inside our network, we implemented a fix to use the locally available DNS in `/etc/resolv`. Unfortunately, this does not work on Linux machines since there is a DNS resolver hosted on `127.0.0.1`. When [trying to use the docker dns `127.0.0.11`](https://medium.com/@prajwal.chin/understanding-docker-dns-2ed4b070a0#:~:text=By%20default%2C%20Docker%20uses%20the,0.11), we ended up with an infinite dns loop. It looks like [kind is aware of that and tries to mitigate it](https://hwchiu.medium.com/fun-dns-facts-learned-from-the-kind-environment-241e0ea8c6d4), could talos do the same?

Thank you!
### Logs
```
❯ kubectl -n kube-system logs coredns-565ffcd8f4-hnmq6
.:53
[INFO] plugin/reload: Running configuration SHA512 = f43368fe881b6cd37b121f37ba0b71c065df5bfc99b5c5c05d7f95bf82289d7ab7e78d5b98c1f02172d8004a8a8f34027cef04e86c780d40a7c5d1301559f5b3
CoreDNS-1.11.1
linux/amd64, go1.20.7, ae2bbc2
[FATAL] plugin/loop: Loop (127.0.0.1:35656 -> :53) detected for zone ".", see https://coredns.io/plugins/loop#troubleshooting. Query: "HINFO 1776841706389146686.271906425746229838."
```
### Environment

- Talos version: v1.6.4
- Platform: Docker

smira:
More specifically, pick up `/etc/resolv.conf` contents by default when in container mode, and use that as a base resolver for the host DNS.

Fixes #8303

### DIFF
--- a/internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
+++ b/internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	"github.com/siderolabs/gen/xslices"
+	"github.com/siderolabs/go-pointer"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 
@@ -169,7 +170,7 @@ func (suite *LocalAffiliateSuite) TestCPGeneration() {
 		asrt.Equal("Talos ("+version.Tag+")", spec.OperatingSystem)
 		asrt.Equal(cluster.KubeSpanAffiliateSpec{}, spec.KubeSpan)
 		asrt.NotNil(spec.ControlPlane)
-		asrt.Equal(6445, spec.ControlPlane.APIServerPort)
+		asrt.Equal(6445, pointer.SafeDeref(spec.ControlPlane).APIServerPort)
 	})
 
 	discoveryConfig.TypedSpec().DiscoveryEnabled = false
--- a/internal/app/machined/pkg/controllers/config/acquire.go
+++ b/internal/app/machined/pkg/controllers/config/acquire.go
@@ -214,6 +214,12 @@ func (validationModeDiskConfig) RequiresInstall() bool {
 	return false
 }
 
+// InContainer implements validation.RuntimeMode interface.
+func (validationModeDiskConfig) InContainer() bool {
+	// containers don't persist config to disk
+	return false
+}
+
 // String implements validation.RuntimeMode interface.
 func (validationModeDiskConfig) String() string {
 	return "diskConfig"
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/container/container.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/container/container.go
@@ -6,7 +6,6 @@
 package container
 
 import (
-	"bytes"
 	"context"
 	"encoding/base64"
 	"log"
@@ -16,8 +15,8 @@ import (
 	"github.com/siderolabs/go-procfs/procfs"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/container/internal/files"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
-	"github.com/siderolabs/talos/pkg/machinery/resources/network"
 	runtimeres "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
 )
 
@@ -60,26 +59,25 @@ func (c *Container) KernelArgs(string) procfs.Parameters {
 func (c *Container) NetworkConfiguration(ctx context.Context, _ state.State, ch chan<- *runtime.PlatformNetworkConfig) error {
 	networkConfig := &runtime.PlatformNetworkConfig{}
 
-	hostname, err := os.ReadFile("/etc/hostname")
+	hostnameSpec, err := files.ReadHostname("/etc/hostname")
 	if err != nil {
-		return err
-	}
-
-	hostname = bytes.TrimSpace(hostname)
-
-	hostnameSpec := network.HostnameSpecSpec{
-		ConfigLayer: network.ConfigPlatform,
-	}
-
-	if err := hostnameSpec.ParseFQDN(string(hostname)); err != nil {
 		return err
 	}
 
 	networkConfig.Hostnames = append(networkConfig.Hostnames, hostnameSpec)
 
+	resolverSpec, err := files.ReadResolvConf("/etc/resolv.conf")
+	if err != nil {
+		return err
+	}
+
+	if len(resolverSpec.DNSServers) > 0 {
+		networkConfig.Resolvers = append(networkConfig.Resolvers, resolverSpec)
+	}
+
 	networkConfig.Metadata = &runtimeres.PlatformMetadataSpec{
 		Platform:     c.Name(),
-		Hostname:     string(hostname),
+		Hostname:     hostnameSpec.FQDN(),
 		InstanceType: os.Getenv("TALOSSKU"),
 	}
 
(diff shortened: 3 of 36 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/container/container.go
```diff
@@ -77,6 +78,34 @@ func (c *Container) NetworkConfiguration(ctx context.Context, _ state.State, ch
...
+	resolverSpec := network.ResolverSpecSpec{
+		ConfigLayer: network.ConfigPlatform,
+	}
+
+	resolvers, err := os.ReadFile("/etc/resolv.conf")
+	if err != nil {
+		return err
+	}
+
+	for _, line := range bytes.Split(resolvers, []byte("\n")) {
```
DmitriyMV, line 90: Maybe use `bufio.NewScanner(f)` and `f, err := os.Open("/etc/resolv.conf")`? 

That way there is no need for `[]byte(...)` conversions.

smira: let's call it PoC for now, I think this way is actually easier for a small file.

internal/app/machined/pkg/controllers/cluster/local_affiliate_test.go
```diff
@@ -169,7 +169,10 @@ func (suite *LocalAffiliateSuite) TestCPGeneration() {
 		asrt.Equal("Talos ("+version.Tag+")", spec.OperatingSystem)
 		asrt.Equal(cluster.KubeSpanAffiliateSpec{}, spec.KubeSpan)
 		asrt.NotNil(spec.ControlPlane)
-		asrt.Equal(6445, spec.ControlPlane.APIServerPort)
+
+		if spec.ControlPlane != nil {
+			asrt.Equal(6445, spec.ControlPlane.APIServerPort)
+		}
```
DmitriyMV, line 175: IMHO this looks weird given the check above. I think we can actually do:

```suggestion
		asrt.Equal(6445, pointer.SafeDeref(spec.ControlPlane).APIServerPort)
```

pkg/machinery/config/generate/worker.go
```diff
@@ -96,10 +96,14 @@ func (in *Input) worker() ([]config.Document, error) {
...
 
-	if in.Options.VersionContract.LocalDNSEnabled() {
+	if in.Options.VersionContract.HostDNSEnabled() || in.Options.HostDNSForwardKubeDNSToHost.ValueOrZero() {
 		machine.MachineFeatures.HostDNSSupport = &v1alpha1.HostDNSConfig{
 			HostDNSEnabled: pointer.To(true),
 		}
+
+		if in.Options.HostDNSForwardKubeDNSToHost.ValueOrZero() {
+			machine.MachineFeatures.HostDNSSupport.HostDNSForwardKubeDNSToHost = pointer.To(true)
+		}
```
DmitriyMV, line 106: I think we can set the value directly in `HostDNSSupport`:

```go
		machine.MachineFeatures.HostDNSSupport = &v1alpha1.HostDNSConfig{
			HostDNSEnabled: pointer.To(true),
			HostDNSForwardKubeDNSToHost: in.Options.HostDNSForwardKubeDNSToHost.Ptr(),
		}
```

pkg/machinery/config/generate/init.go
```diff
@@ -95,10 +95,14 @@ func (in *Input) init() ([]config.Document, error) {
...
 
-	if in.Options.VersionContract.LocalDNSEnabled() {
+	if in.Options.VersionContract.HostDNSEnabled() || in.Options.HostDNSForwardKubeDNSToHost.ValueOrZero() {
 		machine.MachineFeatures.HostDNSSupport = &v1alpha1.HostDNSConfig{
 			HostDNSEnabled: pointer.To(true),
 		}
+
+		if in.Options.HostDNSForwardKubeDNSToHost.ValueOrZero() {
+			machine.MachineFeatures.HostDNSSupport.HostDNSForwardKubeDNSToHost = pointer.To(true)
+		}
```
DmitriyMV, line 105: I think we can set the value directly in `HostDNSSupport`:

```go
		machine.MachineFeatures.HostDNSSupport = &v1alpha1.HostDNSConfig{
			HostDNSEnabled: pointer.To(true),
			HostDNSForwardKubeDNSToHost: in.Options.HostDNSForwardKubeDNSToHost.Ptr(),
		}
```

DmitriyMV (commented):
I'm trying to understand and failing. 

So if we run Talos in Docker container we use container passed `resolv.conf`. **But** we require user to set `forwardKubeDNSToHost: true` which means we pass our own DNS to pods and use it for networking. How does this works together? Sorry if missing something obvious.

pkg/machinery/config/generate/worker.go
smira: yep, great idea!

smira:
> So if we run Talos in Docker container we use container passed `resolv.conf`. **But** we require user to set `forwardKubeDNSToHost: true` which means we pass our own DNS to pods and use it for networking. How does this works together? Sorry if missing something obvious.

so the idea is the following:

1. `/etc/resolv.conf` on container boot contains container runtime provided resolvers, which we want to read and use 
2. we could rewrite `/etc/resolv.conf` with HostDNS, but that would require synchronization between reading and re-writing, and might be more confusing for someone doing `read /etc/resolv.conf`, so on the host we use just container-runtime provided resolvers; we could rewrite `/etc/resolv.conf` though, it would work same way
3. we enforce `forwardKubeDNSToHost` so that CoreDNS in Kubernetes uses host DNS to passthrough requests to container runtime resolver. if we don't do it, e.g. Docker default '127.0.0.11' won't work from CoreDNS, but '10.96.0.9' (host DNS) would perfectly work

Hope I explained it.

internal/app/machined/pkg/controllers/config/acquire.go
```diff
@@ -214,6 +214,12 @@ func (validationModeDiskConfig) RequiresInstall() bool {
 	return false
 }
 
+// RequiresInstall implements validation.RuntimeMode interface.
+func (validationModeDiskConfig) InContainer() bool {
```
Unix4ever, line 218:
```suggestion
// InContainer implements validation.RuntimeMode interface.
func (validationModeDiskConfig) InContainer() bool {
```

Unix4ever (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/container/internal/files/resolv.go
```diff
@@ -0,0 +1,47 @@
...
+		ConfigLayer: network.ConfigPlatform,
+	}
+
+	resolvers, err := os.ReadFile(path)
+	if err != nil {
+		return resolverSpec, err
+	}
+
+	for _, line := range bytes.Split(resolvers, []byte("\n")) {
+		log.Printf("line orig = %s", string(line))
```
frezbo, line 28: needed?

frezbo (approved):

internal/app/machined/pkg/runtime/v1alpha1/platform/container/internal/files/resolv.go
smira: thanks, dropped!

internal/app/machined/pkg/controllers/config/acquire.go
smira: thanks, fixed!

DmitriyMV (approved):

PR #8584, opened 2024-04-11
fix: do not ignore IPv6 errors in DNSResolveCacheController

DmitriyMV:
Currently, if we try to raise dns listener on IPv6 address and listening fails for any reason, we just silently ignore this for and move on. The reason for this is that raising listener "by default" on IPv6 `::1` sometimes simply doesn't work. However, if the user decides to use IPv6 service subnet CIDR and `forwardKubeDNSToHost` AND starting dns listening on this IPv6 network fails for any reason, we will just silently ignore the error, which is not what we want.

After Andrey's [PR](https://github.com/siderolabs/talos/pull/8583) that will remove `::1` from default `resolv.conf` we no longer have the first problematic part, so it makes sense to no longer ignore IPv6 dns errors. This commit does this.

Depends on #8583

### DIFF
--- a/internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
+++ b/internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
@@ -127,10 +127,6 @@ func (ctrl *DNSResolveCacheController) Run(ctx context.Context, r controller.Run
 						return fmt.Errorf("error creating dns runner: %w", rErr)
 					}
 
-					if runner == nil {
-						continue
-					}
-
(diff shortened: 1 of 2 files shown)


smira:
I can probably pull it in #8583, as I'd need to update the tests and do fixes like that.

I plan to work on some documentation updates today, tomorrow is going to be beta.1 release, and we can try to fit it in before GA

DmitriyMV:
@smira works for me.

smira:
I merged this into https://github.com/siderolabs/talos/pull/8583, so I'll close this one, thanks @DmitriyMV !

PR #8610, opened 2024-04-17
feat: add host dns support for resolving member addrs

### LINKED ISSUE #8330: resolve machine hostnames (from discovery) via local DNS resolver

DmitriyMV:
Closes #8330


DmitriyMV:
Okay, I went the this way:
```shell
> sudo -E talosctl cluster create \ 
--provisioner=qemu \
--cidr=172.20.0.0/24 \
...
--controlplanes 1 \
--workers 1 \
--with-bootloader=false \
--skip-injecting-config \
--with-apply-config \
--config-patch=@./patch.yaml
```

With `patch.yaml` being

```yaml
version: v1alpha1
machine:
    features:
        hostDNS:
            enabled: true
            forwardKubeDNSToHost: true
            resolveMemberNames: true
```

Then

```shell
> kubectl run alpine --image alpine -it -- ash
```

Then 

```shell
> ping talos-default-controlplane-1
```

And result is:
image

🎉

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
```diff
@@ -288,3 +295,29 @@ func newDNSRunner(cfg runnerConfig, cache *dns.Cache, logger *zap.Logger) (*dns.
...
+
+func (s *stateMapper) ResolveAddr(ctx context.Context, name string) (netip.Addr, bool) {
+	name = strings.TrimRight(name, ".")
+
+	list, err := safe.ReaderListAll[*cluster.Member](ctx, s.state)
+	if err != nil {
+		return netip.Addr{}, false
+	}
+
+	elem, ok := list.Find(func(res *cluster.Member) bool { return res.TypedSpec().Hostname == name })
```
DmitriyMV, line 311: I'm not sure that `id == .Hostname` is always true so I went with this. If it is always true then this can be dropped and we can get spec by id above.

smira: as we discussed, `Hostname` might be a FQDN as well, so the match should be updated. 

If we match on the first part of the FQDN, it's a match (`foo` -> `foo.example.com`).
If we match on the full name, it's a match (`foo.example.com` -> `foo.example.com`).
We should also probably match on `Nodename` same way as well.

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
```diff
@@ -288,3 +295,29 @@ func newDNSRunner(cfg runnerConfig, cache *dns.Cache, logger *zap.Logger) (*dns.
...
+		return netip.Addr{}, false
+	}
+
+	result := elem.TypedSpec().Addresses
+
+	if len(result) == 0 {
+		return netip.Addr{}, false
+	}
+
+	return result[0], true
```
smira, line 322: why one address? we should resolve to all addresses

internal/pkg/dns/dns.go
```diff
@@ -158,6 +160,83 @@ func (h *Handler) SetProxy(prxs []*proxy.Proxy) bool {
...
+	// Check if the request is for a node.
+	result, ok := h.mapper.ResolveAddr(ctx, req.Name())
+	if !ok {
+		return h.next.ServeDNS(ctx, wrt, msg)
+	}
+
+	var rtype uint16
+
+	if result.Is4() {
+		rtype = dns.TypeA
```
smira, line 209: we should answer by query, not by result I guess

we should filter the result by v4/v6 based on query

internal/pkg/dns/dns.go
```diff
@@ -158,6 +160,83 @@ func (h *Handler) SetProxy(prxs []*proxy.Proxy) bool {
...
+	}
+
+	resp := new(dns.Msg).SetReply(req.Req)
+	resp.Authoritative = true
+	resp.Answer = append(resp.Answer, &dns.A{
+		Hdr: dns.RR_Header{
+			Name:   req.Name(),
+			Rrtype: rtype,
+			Class:  dns.ClassINET,
+			Ttl:    10,
```
smira, line 221: should be a constant I guess

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
DmitriyMV: Done. We now return multiple DNS answers.

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
DmitriyMV: Done. We are also testing this matching in controller unit-tests.

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
```diff
@@ -288,3 +297,49 @@ func newDNSRunner(cfg runnerConfig, cache *dns.Cache, logger *zap.Logger) (*dns.
...
+func (s *stateMapper) ResolveAddr(ctx context.Context, qType uint16, name string) []netip.Addr {
+	name = strings.TrimRight(name, ".")
+
+	list, err := safe.ReaderListAll[*cluster.Member](ctx, s.state)
+	if err != nil {
+		return nil
+	}
+
+	elem, ok := list.Find(func(res *cluster.Member) bool {
+		return fqdnMatch(name, res.TypedSpec().Hostname) || fqdnMatch(name, res.TypedSpec().NodeID)
```
smira, line 314: `Nodename`, not `NopeID`

smira (approved):

internal/app/machined/pkg/controllers/network/dns_resolve_cache.go
DmitriyMV: There is no `Nodename` in the member spec, but `AffiliateSpec` has it. But it is used to populate the id of the member resource [here](https://github.com/siderolabs/talos/blob/fa3b93370501009283e110b74876b18ce6bad4f9/internal/app/machined/pkg/controllers/cluster/member.go#L75) so I think we can use that.

smira: ah yeah, right, it's the Member

PR #8614, opened 2024-04-18
fix: close pcap file

testwill:
# Pull Request

## What? (description)
close pcap  output file
## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/pcap.go
+++ b/cmd/talosctl/cmd/talos/pcap.go
@@ -108,6 +108,11 @@ e.g. by excluding packets with the port 50000.
 				if err != nil {
 					return err
 				}
+				defer func() {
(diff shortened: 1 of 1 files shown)


cmd/talosctl/cmd/talos/pcap.go
```diff
@@ -108,6 +108,11 @@ e.g. by excluding packets with the port 50000.
 				if err != nil {
 					return err
 				}
+				defer func() {
```
smira, line 111: this is totally minor, as the file will be closed anyway as the program terminates

but assignment to `err` won't work here

so better `defer out.Close() //nolint:errcheck` simply

DmitriyMV: Another way is to print the error to `os.Stderr` which should be available here.

PR #8627, opened 2024-04-22
fix: return proper value from Bridge.STP instead of plain nil

### LINKED ISSUE #8626: Go Panic in LinkConfigController when attempting to add bridge
### Description
When attempting to add a bridge using the following config fragment, it fails to add a bridge and logs a go panic message to console.
```yaml
machine:
  nodeLabels:
    role: worker
  network:
    interfaces:
      - interface: eth0
        dhcp: true
      - interface: eth1
        dhcp: false
      - interface: br0
        dhcp: false
        bridge:
          interfaces:
            - eth1
```
### Logs
```
 user: warning: [2024-04-22T06:23:22.338896757Z]: [talos] controller failed {"component": "controller-runtime", "controller": "network.LinkConfigController", "error": "controller \"network.LinkConfigController\" panicked: runtime error: invalid memory address or nil pointer
 dereference\n\ngoroutine 502 [running]:\nruntime/debug.Stack()\n\t/toolchain/go/src/runtime/debug/stack.go:24 +0x5e\ngithub.com/cosi-project/runtime/pkg/controller/runtime/internal/rruntime.(*Adapter).runOnce.func2()\n\t/.cache/mod/github.com/cosi-project/runtime@v0.4.1/pkg/
 controller/runtime/internal/rruntime/run.go:67 +0x58\npanic({0x2bc4b60?, 0x526cd90?})\n\t/toolchain/go/src/runtime/panic.go:770 +0x132\ngithub.com/siderolabs/talos/internal/app/machined/pkg/controllers/network.SetBridgeMaster(0xc001b3b500, {0x367f010?, 0xc000813820?})\n\t/src/
 internal/app/machined/pkg/controllers/network/network.go:133 +0x58\ngithub.com/siderolabs/talos/internal/app/machined/pkg/controllers/network.(*LinkConfigController).processDevicesConfiguration(0..
```
### Environment

- Talos version:
```
Client:                                                                                                                                 Tag:         v1.7.0                                                                                                             SHA:         70fb41ff                                                                                                   
        Built:
        Go version:  go1.22.2
        OS/Arch:     windows/amd64
Server:
        NODE:        pathweb-worker-1.servers.home.yrf.me
        Tag:         v1.7.0
        SHA:         70fb41ff
        Built:
        Go version:  go1.22.2
        OS/Arch:     linux/amd64
        Enabled:     RBAC
```
- Kubernetes version:
```
Client Version: v1.29.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.29.3
```
- Platform: metal-amd64 (Proxmox)

DmitriyMV:
Since we are returning interfaces in config, we have to return something typed for method chaining to work. Otherwise, it simply doesn't know what method to call because there is no type information. We also don't want to change the default config behavior, so we don't try to check for `nil` after calling `.STP()`.

Closes #8626

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
@@ -1047,7 +1047,7 @@ func (b *Bridge) Interfaces() []string {
 // STP implements the config.Bridge interface.
 func (b *Bridge) STP() config.STP {
 	if b.BridgeSTP == nil {
-		return nil
+		return (*STP)(nil)
 	}
 
 	return b.BridgeSTP


utkuozdemir (approved):

frezbo (approved):
any other places that might need the fix too?

DmitriyMV:
@frezbo 

> any other places that might need the fix too?

Looks like in other places either don't use method chaining before checking for `nil` or we return proper objects from the method.

Unix4ever (approved):

smira (approved):

PR #8632, opened 2024-04-22
test: add a test for watchdog timers

dsseng:
# Pull Request

## What? (description)
Add a test for watchdog

## Why? (reasoning)
Ensure integration tests cover all code

## Acceptance

### DIFF
--- a/internal/integration/api/watchdog.go
+++ b/internal/integration/api/watchdog.go
@@ -1,0 +1,134 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_api
+
+package api
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"path/filepath"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+
+	"github.com/siderolabs/talos/internal/integration/base"
+	"github.com/siderolabs/talos/pkg/machinery/client"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/runtime"
+	runtimeres "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
+)
+
+// WatchdogSuite ...
+type WatchdogSuite struct {
+	base.APISuite
+
+	ctx       context.Context //nolint:containedctx
+	ctxCancel context.CancelFunc
+}
+
+// SuiteName ...
+func (suite *WatchdogSuite) SuiteName() string {
+	return "api.WatchdogSuite"
+}
+
+// SetupTest ...
+func (suite *WatchdogSuite) SetupTest() {
+	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 1*time.Minute)
+
+	if suite.Cluster == nil || suite.Cluster.Provisioner() != "qemu" {
+		suite.T().Skip("skipping watchdog test since provisioner is not qemu")
+	}
+}
+
+// TearDownTest ...
+func (suite *WatchdogSuite) TearDownTest() {
+	if suite.ctxCancel != nil {
+		suite.ctxCancel()
+	}
+}
+
+func (suite *WatchdogSuite) readWatchdogSysfs(nodeCtx context.Context, watchdog, property string) string { //nolint:unparam
+	r, err := suite.Client.Read(nodeCtx, filepath.Join("/sys/class/watchdog", watchdog, property))
+	suite.Require().NoError(err)
+
+	value, err := io.ReadAll(r)
+	suite.Require().NoError(err)
+
+	suite.Require().NoError(r.Close())
+
+	return string(bytes.TrimSpace(value))
+}
+
+// TestWatchdogSysfs sets up the watchdog and validates its parameters from the /sys/class/watchdog.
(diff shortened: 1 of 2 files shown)


smira:
I'm going to take a look into this to add more generic helpers to help with patching machine configuration.

smira:
Feedback/notes:

* we should use `test:` for changes around tests 
* I added some base helpers to help with patching - the easiest is to use strategic merge config patches to update/add multi-doc machine config documents.
* it's always better to use native types than to build YAML as bytes manually
* after applying machine config we can't expect immediate change being applied, as controllers work async - so we use controller output resources as a signal that the word had been done
* extract common helper to read watchdog properties

smira:
```shell
$ _out/integration-test-linux-amd64 -test.v -talos.crashdump=false -talos.provisioner=qemu -talos.talosctlpath=$PWD/_out/talosctl-linux-amd64 -test.run TestIntegration/api.Watch
=== RUN   TestIntegration
=== RUN   TestIntegration/api.WatchdogSuite
=== RUN   TestIntegration/api.WatchdogSuite/TestWatchdogSysfs
    watchdog.go:69: testing watchdog on node 172.20.0.2
    api.go:593: patched machine config: Applied configuration without a reboot
    api.go:593: patched machine config: Applied configuration without a reboot
--- PASS: TestIntegration (0.30s)
    --- PASS: TestIntegration/api.WatchdogSuite (0.29s)
        --- PASS: TestIntegration/api.WatchdogSuite/TestWatchdogSysfs (0.14s)
PASS
```

internal/integration/api/watchdog.go
```diff
@@ -0,0 +1,108 @@
...
+	suite.T().Logf("testing watchdog on node %s", node)
+
+	// build a Talos API context which is tied to the node
+	nodeCtx := client.WithNode(suite.ctx, node)
+
+	// pick a watchdog
+	const watchdog = "watchdog0"
+
+	wdState := suite.readWatchdogSysfs(nodeCtx, watchdog, "state")
+	suite.Require().Equal("inactive", wdState)
```
dsseng, line 79: What if the test node has WDT already enabled? Maybe we should first remove the document and assert inactive state?

dsseng:
Thanks a lot for your help, will see and do anything else to be done if anything tomorrow

internal/integration/api/watchdog.go
smira: yes, we can add a remove on top, but as the test removes the watchdog, we might assume that a "clean" state is that watchdog is not active

we assume that tests run in a clean environment

smira: either way is fine

internal/integration/api/watchdog.go
```diff
@@ -0,0 +1,123 @@
...
+	wdState = suite.readWatchdogSysfs(nodeCtx, watchdog, "state")
+	suite.Require().Equal("active", wdState)
+
+	wdTimeout := suite.readWatchdogSysfs(nodeCtx, watchdog, "timeout")
+	suite.Require().Equal("120", wdTimeout)
+
+	cfgDocument.WatchdogTimeout = 60 * time.Second
+	suite.PatchMachineConfig(nodeCtx, cfgDocument)
+
+	_, err = suite.Client.COSI.WatchFor(nodeCtx, runtimeres.NewWatchdogTimerStatus(runtimeres.WatchdogTimerConfigID).Metadata(), state.WithEventTypes(state.Created, state.Updated))
```
smira, line 102: It won't work this way, as the resource is already present, so you want to watch for a resource to have expected state, that is timeout of 60 seconds

dsseng: isn't state.Updated expected to catch this?

smira: first of all, you have both there; second it's always a race - did you start watching before it got updated or after, you never know; does any other controller update the resource (e.g. put a finalizer on it)

what you need here is: https://pkg.go.dev/github.com/cosi-project/runtime@v0.4.1/pkg/state#WithCondition

```go
state.WithCondition(func (r resource.Resource) bool {
   return r.(*runtime.WatchdogTimeStatus).TypedSpec().Timeout == 2 * time.Minute
})
```

smira: (wrong timeout, but I hope my point is clear)

internal/integration/api/watchdog.go
```diff
@@ -0,0 +1,136 @@
...
+	suite.Require().NoError(err)
+
+	wdState = suite.readWatchdogSysfs(nodeCtx, watchdog, "state")
+	suite.Require().Equal("active", wdState)
+
+	state.WithCondition(func(r resource.Resource) (bool, error) {
```
smira, line 109: this statement has no effect, it should instead be moved into line 103 as the argument to the `WatchFor` function

internal/integration/api/watchdog.go
```diff
@@ -0,0 +1,136 @@
...
+		return (r.(*runtimeres.WatchdogTimerStatus).TypedSpec().Timeout == 1*time.Minute), nil
+	})
+	wdTimeout = suite.readWatchdogSysfs(nodeCtx, watchdog, "timeout")
+	suite.Require().Equal("60", wdTimeout)
+
+	cfgDocument.WatchdogTimeout = 90 * time.Second
+	suite.PatchMachineConfig(nodeCtx, cfgDocument)
+
+	state.WithCondition(func(r resource.Resource) (bool, error) {
```
smira, line 118: same here

smira (approved):
fixed up small issues to get this PR merged

PR #8636, opened 2024-04-23
feat: add `--non-masquerade-cidrs` flag to `talosctl cluster create`

utkuozdemir:
Allow skipping NAT for the given destinations from a cluster network. This option makes it possible to form an etcd cluster from clusters in different networks created by running `talosctl cluster create` command multiple times using different CIDRs: they simply should have the CIDR of the other clusters passed with `--non-masquerade-cidrs`.


smira (approved):
lgtm, I wonder if the option and setting `NoMasquerade` is more natural from English point of view (`no` vs. `non`)

utkuozdemir:
> lgtm, I wonder if the option and setting `NoMasquerade` is more natural from English point of view (`no` vs. `non`)

Done 👍

PR #8642, opened 2024-04-23
feat: kresify part (1/x)

frezbo:
Kresify, start with Makefile and keep it compatible with drone/dockerfile.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -1,0 +1,634 @@
+---
+kind: golang.Toolchain
+spec:
+  # image: ghcr.io/siderolabs/tools:v1.8.0-alpha.0
+  buildTags:
+    - tcell_minimal
+    - grpcnotrace
+  makefile:
+    extraVariables:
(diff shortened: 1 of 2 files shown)


.kres.yaml
```diff
@@ -0,0 +1,634 @@
+---
+kind: golang.Toolchain
+spec:
+  # image: ghcr.io/siderolabs/tools:v1.8.0-alpha.0
+  buildTags:
+    - tcell_minimal
+    - grpcnotrace
+  makefile:
+    extraVariables:
+      - name: NAME
```
smira, line 10: I think this is a bit of a disaster, we're basically rewriting `Makefile` as `.kres.yaml`, 

Can we take a step back and try to analyze what can be done better if we can do that? Or at least analyze some options?

smira:
@frezbo should we close this?

frezbo:
> @frezbo should we close this?

Yes, closing

PR #8650, opened 2024-04-24
test: improve the reset integration tests

smira:
Provide a trace for each step of the reset sequence taken, so if one of those fails, integration test produces a meaningful message instead of proceeding and failing somewhere else.

More cleanup/refactor, should be functionally equivalent.


frezbo (approved):

smira:
There are test failures with qemu-encrypted-vip which I want to take a look separately.

PR #8664, opened 2024-04-26
feat: use zstd compression in place of xz

smira:
Initramfs and kernel are compressed with zstd.

Extensions are compressed with zstd for Talos 1.8+.


pkg/imager/extensions/rebuild.go
```diff
@@ -10,14 +10,16 @@ import (
...
 	"os"
 	"os/exec"
+
+	"github.com/siderolabs/talos/pkg/machinery/imager/quirks"
 )
 
 // rebuildInitramfs rebuilds finalized initramfs with extensions.
 //
 // If uncompressedListing is not empty, contents will be prepended to the initramfs uncompressed.
 // Contents from compressedListing will be appended to the initramfs compressed (xz) as a second block.
```
frezbo, line 20: i guess the comment needs update?

frezbo (approved):
another place to update is the overlays to use zst

pkg/imager/extensions/rebuild.go
smira: thanks, updated

smira:
> another place to update is the overlays to use zst

where do we use xz there?

frezbo:
> > another place to update is the overlays to use zst
> 
> where do we use xz there?

part of profile definition: https://github.com/siderolabs/sbc-raspberrypi/blob/main/profiles/rpi_generic/rpi_generic.yaml#L6

smira:
> > > another place to update is the overlays to use zst
> > 
> > 
> > where do we use xz there?
> 
> part of profile definition: [siderolabs/sbc-raspberrypi@`main`/profiles/rpi_generic/rpi_generic.yaml#L6](https://github.com/siderolabs/sbc-raspberrypi/blob/main/profiles/rpi_generic/rpi_generic.yaml?rgh-link-date=2024-04-26T13%3A41%3A31Z#L6)

ah, makes sense... this one is tricky, we only need to update for 1.8 probably, as 1.7 imager won't handle it

PR #8669, opened 2024-04-28
docs: remove examples from generated files

andrewrynhard:
docs: remove examples from generated files

This removes examples from configuration generation.

Should be merged after #8198.


andrewrynhard:
Not ready for review. Using CI to test.

PR #8697, opened 2024-05-03
docs: fix grpc_tunnel value to true

Simon-Boyer:
# Pull Request

## What? (description)

Changing documentation about siderolink so that the value to enable grpc_tunnel is `true`, so it will be `https://siderolink.api/?jointoken=token&grpc_tunnel=true`, which is what the code is actually looking for.

## Why? (reasoning)

grpc_tunnel is described as being enabled by using the value yes in the docs, but it should be true.

## Acceptance


smira (approved):
thank you!

PR #8717, opened 2024-05-07
chore: enable forwardKubeDNSToHost by default

DmitriyMV:
This PR enables by default the usage of our host DNS resolver as upstream for Kubernetes CoreDNS pods.


smira:
This should be a `feat:` as it's user-visible change, it should be listed in `hack/release.toml` notes, and let's clear it out of the other PR bits. So this PR will stay in v1.8, and the fix will be backported to v1.7

DmitriyMV:
@smira 

> This should be a `feat:`

Done.

> it should be listed in hack/release.toml notes

Done.

> and let's clear it out of the other PR bits. So this PR will stay in v1.8, and the fix will be backported to v1.7

Also done.

smira (approved):

PR #8724, opened 2024-05-08
fix: do not try to bootstrap cluster if `wait` flag is false

Unix4ever:
Otherwise to make `--wait=false` work you have to also pass `--with-init-node` flag, which looks weird.

I guess it should be a nice change, unless it breaks some backward compability.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -922,14 +922,14 @@ func nodeName(clusterName, role string, index int, uuid uuid.UUID) string {
 }
 
 func postCreate(ctx context.Context, clusterAccess *access.Adapter) error {
+	if !clusterWait {
+		return nil
+	}
+
 	if !withInitNode {
 		if err := clusterAccess.Bootstrap(ctx, os.Stdout); err != nil {
 			return fmt.Errorf("bootstrap error: %w", err)
 		}
-	}
-
-	if !clusterWait {
-		return nil
 	}
 
 	// Run cluster readiness checks


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -922,16 +922,16 @@ func nodeName(clusterName, role string, index int, uuid uuid.UUID) string {
 }
 
 func postCreate(ctx context.Context, clusterAccess *access.Adapter) error {
+	if !clusterWait {
```
smira, line 925: this feels a bit wrong, as if we do `--wait=false`, then it will never bootstrap a cluster at all (?)

smira: I agree that flags are confusing though

Unix4ever: maybe something like no bootstrap flag will work here.

Unix4ever: On other hand, if you don't want to wait for cluster to be ready you might not want to wait for bootstrap too, no?

smira: well... I agree it should be something like `--skip-bootstrap` but it's overall a mess :face_exhaling:

Unix4ever: Another thing I can suggest is `--bootstrap` flag which is true by default.

Unix4ever: But I don't like the fact that you have to remember that you should do both `--wait=false` and `--bootstrap=false`.
To me it looks logical to skip bootstrap if `--wait=false` :)

smira: `--wait=false` means you don't want to run health checks, e.g. they won't pass as I don't have CNI ready, but I still want to bootstrap.

`--with-init-node` is a legacy thing, but it allows to test init machine type, and also disables bootstrap (as you won't need it)

in Omni, we used both as we don't supply the machine config at all.

I would say we should have a better way to communicate these intentions, and rework the whole thing. 

What is the problem you're trying to fix with this PR? Should we work backwards from it?

Unix4ever: I want `--wait=false` not to demand `--with-init-node`.

Maybe we need another command that can just spin the nodes and that's it.
Going from that end we might need to extract machine and network creation bits into a separate chunk, then create a cmd that can do that.

smira: You can do that today, `pkg/provision` is a separate library, e.g. Sfyra uses it this way.

PR #8742, opened 2024-05-15
fix: update github.com/siderolabs/siderolink to v0.3.7

DmitriyMV:
Version 0.3.6 contains incorrect server implementation which breaks our integration tests.


smira:
1.7 has 0.3.5 siderolink, so I guess it should be fine

smira (approved):

frezbo (approved):

PR #8766, opened 2024-05-20
chore: move to containerd 2.0 API

smira:
Lots of module moves/renames.


internal/pkg/containers/image/image.go
```diff
@@ -95,11 +92,6 @@ func Pull(ctx context.Context, reg config.Registries, client *containerd.Client,
 			containerd.WithPullUnpack,
 			containerd.WithResolver(resolver),
 			containerd.WithChildLabelMap(images.ChildGCLabelsFilterLayers),
-			containerd.WithUnpackOpts(
-				[]containerd.UnpackOpt{
-					containerd.WithUnpackDuplicationSuppressor(unpackDuplicationSuppressor),
```
frezbo, line 100: not needed anymore?

frezbo (approved):

internal/pkg/containers/image/image.go
smira: we can't use it anymore (as the required parts became private), but I don't think we need it, as this is only needed when multiple requests for the same image to be pulled arrive.

smira: and it's only an optimization - it makes such pull operations run faster

PR #8788, opened 2024-05-24
fix(ci): fix crons by setting up buildx always

frezbo:
Fix crons by setting up buildx always, also make sure `images-essential` has `uki-certs` as dependency.

Also use platform as `linux/amd64` in CI integration tests and cron jobs, since we don't run tests with arm64 binaries.


smira (approved):

frezbo:
I hope this is the last fix :sweat_smile:

PR #8797, opened 2024-05-24
feat: implement Talos diagnostics

### LINKED ISSUE #8682: 💉  Talos/Omni Doctor
Proposal: have a somewhat simple rule system that evaluates current state against common pitfalls and provides solutions.

Two problems:

* enabling kubelet server certificate rotation and not approving CSRs (gate: check number of not approved certificates for the kubelet)
* overlapping Talos host address and K8s pod/service CIDRs

Report such issues e.g. to the resource (in Omni? Talos?) and show it using Talos dashboard/Omni dashboard

smira:
Talos doctor analyzes current system state and comes up with detailed warnings on the system misconfiguration which might be tricky to figure out other way.

Fixes #8682


smira:
Linux console:

image


Local dashboard:

image

smira:
Health: 

image

internal/app/machined/pkg/controllers/runtime/internal/doctor/kubelet_csr_not_approved.go
```diff
@@ -0,0 +1,131 @@
...
+
+		return true
+	})
+
+	if len(csrs.Items) == 0 {
+		return nil, nil
+	}
+
+	return &runtime.DoctorWarningSpec{
+		Message: "kubelet server certificate rotation is enabled, but CSR is not approved",
```
frezbo, line 120: should we actually check if the certificate rotation flag is enabled in kubelet, might be easier to shortcut there

frezbo (approved):

internal/app/machined/pkg/controllers/runtime/internal/doctor/kubelet_csr_not_approved.go
smira: it's a bit harder imho, as it can be enabled via the flags or via the config file

andrewrynhard (commented):
I don’t think we should call this “doctor”. 


Can we hold off on merging until I can settle in after my flight and we can discuss what to call it?

andrewrynhard:
> I don’t think we should call this “doctor”.
> 
> Can we hold off on merging until I can settle in after my flight and we can discuss what to call it?

The proper word here would be “diagnostics”

> I don’t think we should call this “doctor”.
> 
> Can we hold off on merging until I can settle in after my flight and we can discuss what to call it?

The proper term would be “diagnostics”.

rothgar:
There's some precedence for "doctor" in naming (e.g. [`brew doctor`](https://docs.brew.sh/Troubleshooting)) which could be helpful. I'm not sure it's the correct term to be used, but it may be a familiar term for a subset of users.

andrewrynhard:
> There's some precedence for "doctor" in naming (e.g. [`brew doctor`](https://docs.brew.sh/Troubleshooting)) which could be helpful. I'm not sure it's the correct term to be used, but it may be a familiar term for a subset of users.

I think that’s fair. Talos Doctor sounds a bit cheesy IMHO. I want Talos to feel like a premium Apple/Mercedes experience and “Doctor” brings it more towards playful. At least to me.

andrewrynhard:
> > There's some precedence for "doctor" in naming (e.g. [`brew doctor`](https://docs.brew.sh/Troubleshooting)) which could be helpful. I'm not sure it's the correct term to be used, but it may be a familiar term for a subset of users.
> 
> I think that’s fair. Talos Doctor sounds a bit cheesy IMHO. I want Talos to feel like a premium Apple/Mercedes experience and “Doctor” brings it more towards playful. At least to me.

I also think it should feel like Talos is diagnosing itself. It has some intelligence. Talos showing “diagnostics” elicits that for ke better than Talos Doctor. It feels like Doctor as a separate entity. I know this sounds really silly but the subtle things like this add up when we start marketing Talos properly.

smira:
`Diagnostics` it is

smira:
image

PR #8816, opened 2024-05-28
fix: replace `nslookup` with `dig` in integration tests

DmitriyMV:
This should be more reliable on `integration-aws-*` and others.

### DIFF
--- a/internal/integration/api/common.go
+++ b/internal/integration/api/common.go
@@ -167,12 +167,16 @@ func (suite *CommonSuite) TestDNSResolver() {
 	suite.Require().Equal("", stdout)
 	suite.Require().Contains(stderr, "'index.html' saved")
 
-	stdout, stderr, err = suite.ExecuteCommandInPod(suite.ctx, namespace, pod, "nslookup really-long-record.dev.siderolabs.io")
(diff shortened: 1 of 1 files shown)


internal/integration/api/common.go
```diff
@@ -167,12 +167,12 @@ func (suite *CommonSuite) TestDNSResolver() {
 	suite.Require().Equal("", stdout)
 	suite.Require().Contains(stderr, "'index.html' saved")
 
-	stdout, stderr, err = suite.ExecuteCommandInPod(suite.ctx, namespace, pod, "nslookup really-long-record.dev.siderolabs.io")
+	stdout, stderr, err = suite.ExecuteCommandInPod(suite.ctx, namespace, pod, "dig really-long-record.dev.siderolabs.io")
```
smira, line 170: I thought it's not available by default in `alpine` container?

DmitriyMV: @smira maybe can we install it using `apk add --update bind-tools` the final command will look like `apk add --update bind-tools && dig really-long-record.dev.siderolabs.io`

DmitriyMV: Lets install it with `apk add --update bind-tools`.

smira (approved):

PR #8827, opened 2024-05-29
386 build

attilaolah:
# Pull Request

## What? (description)

Adds files with conditional compilation for handling unix.Time* values on GOARCH=386 systems.

## Why? (reasoning)

To support building `machined` for i686. (I realise that Talos doesn't support i386, but I have an experimental Nix flake that attempts to generate a container image for running worker nodes on i686 hardware, and decided to upstream some changes for building `machined` for 32-bit targets.)

Feel free to reject if this is not something you plan on supporting.

## Acceptance

`make fmt` modifies lots of unrelated proto files. `make conformance` complains about commit-case, but the first line should already conform, as far as I can tell.

### DIFF
--- a/internal/pkg/ntp/ntp.go
+++ b/internal/pkg/ntp/ntp.go
@@ -357,11 +357,11 @@ func (syncer *Syncer) queryPTP(server string) (*Measurement, error) {
 		return nil, err
 	}
 
-	offset := time.Until(time.Unix(ts.Sec, ts.Nsec))
+	offset := time.Until(time.Unix(int64(ts.Sec), int64(ts.Nsec)))
 	syncer.logger.Debug("PTP clock",
 		zap.Duration("clock_offset", offset),
-		zap.Int64("sec", ts.Sec),
-		zap.Int64("nsec", ts.Nsec),
+		zap.Int64("sec", int64(ts.Sec)),
+		zap.Int64("nsec", int64(ts.Nsec)),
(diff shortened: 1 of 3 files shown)


smira:
The changes might be refactored a bit to be more readable, but my concern is that we never build & test 32-bit Talos, so even if we take the fixes, we won't be able to confirm that they still work.

attilaolah:
> The changes might be refactored a bit to be more readable, but my concern is that we never build & test 32-bit Talos, so even if we take the fixes, we won't be able to confirm that they still work.

Yeah that was my concern too, but adding 32-bit runs to the CI seems like a bigger undertaking. And even if `machined` builds, testing that it actually functions is quite some work.

Let me close the change then, I'll use the diff from this PR for local builds.

internal/pkg/ntp/ntp.go
```diff
@@ -357,11 +357,11 @@ func (syncer *Syncer) queryPTP(server string) (*Measurement, error) {
 		return nil, err
 	}
 
-	offset := time.Until(time.Unix(ts.Sec, ts.Nsec))
+	offset := time.Until(time.Unix(ts.Unix()))
```
attilaolah, line 360: However this one line here might still be a good idea, for readability.

smira: yep, that's a good one for sure

PR #8860, opened 2024-06-04
Export Cobra CMDs

Ornias1993:
# Pull Request

## What? (description)

This PR exports all Cobra CMD variables, and just that.

## Why? (reasoning)

A lot of functional code for talos command line functions, is "hidden away" within the cobra.command variables.
These are currently not exported.

However, when one, for example, would want to write a custom tool based on TalosCTL, it's nice to at least have theoretical access to this code. 

Even more so because Cobra does allow execution of commands through go code, as long as the cobra.command variable is exported and initialised in the current project.

---

To clearify: 
This does not guarantee that importing/using said variables will work in all cases. 
It will work in most talctl commands, but might not play nice with others.

But at least it prevents downstream users from using hacky solutions like partly-copied code or, worse, `linkname` references

## Acceptance

- [ ] you linked an issue (if applicable) -> not applicable or at least not super-relevant
- [ ] you included tests (if applicable) -> not applicable
- [ ] you linted your code (`make lint`) -> If the lint tests demand this to be unexported, the tests needs changing
- [ ] you generated documentation (`make docs`) -> Shouldn't be relevant as no docs are changed


Ornias1993:
@smira I looked at your PRs and you're passing GPG conformity and i'm not.
While my commit obviously is GPG signed with github [vigilant mode](https://docs.github.com/github/authenticating-to-github/displaying-verification-statuses-for-all-of-your-commits) enabled + confirming the signature...

Any tips on that?

smira:
> While both our commits obviously are GPG signed with github [vigilant mode](https://docs.github.com/github/authenticating-to-github/displaying-verification-statuses-for-all-of-your-commits) enabled + confirming the signature...

don't worry about it, they should be signed with our key (we will re-sign if we decide to merge).

---

I don't feel comfortable making commands an exported API of Talos, as it gives some sense of compatibility guarantee, which we don't want to give (at least right now).

I'm up for trying to move some of the CLI code into its own structured subpackages under `pkg/` so that they can be re-used if necessary, but this should be more case-by-case and with proper structure

Ornias1993:
> I'm up for trying to move some of the CLI code into its own structured subpackages under pkg/ so that they can be re-used if necessary, but this should be more case-by-case and with proper structure

I totally agree that having things moved to PKG is prefered...

Primarily I'm concerned about the fact that a significant portion of the following, basic, cmds, is hidden behind non-exported cobra run sections:
- apply
- bootstrap
- Upgrade
- Reset

Those are basically the core/basics of deploying talosctl config to nodes, adding nodes, updating nodes etc.
I'm okey with having those moved to pkg instead, because that is, in fact, cleaner.

smira:
I think all of them are mostly trivial, and with most of the code already in external packages. But certainly it might be good to have a package which "extends" Client to support easier flows for some of these commands. But if we do that, let's try to do it step by step in small increments when it's easier to review.

These packages shouldn't have any CLI handling themselves (printing messages, parsing args, etc.). They should start with having a reference to the client and context.

PR #8869, opened 2024-06-05
fix: clean up VM runners on cluster destroy

smira:
This never worked properly, as `Wait()` doesn't work for child processes, and `talosctl cluster destroy` is not a child of processes created by `talosctl cluster create`.


smira:
This bug was leading to incomplete CNI cleanup on `cluster destroy`.

Unix4ever (approved):

PR #8874, opened 2024-06-07
chore: output more info when `ExecuteCommandInPod` fails

DmitriyMV:
This should make investigating things like [this](https://github.com/siderolabs/talos/actions/runs/9411253542/job/25924192027) easier.


smira (approved):

frezbo (approved):

frezbo:
I think we can merge this and it's the ci is flaky at times

Issue #8367, opened 2024-02-26
Volume Management


smira:
Closely related: #8016 

## Problem Statement

Talos Linux is not flexible in the way it manages volumes, it occupies the whole system disk, creating an EPHEMERAL partition covering 99% of the disk space. User disk management is fragile, requires extra steps to get it to work properly (mounting into the kubelet), doesn’t support wiping disks, etc. Talos does not detect properly various partition types which leads to wiping user data (e.g. Ceph bluestore).

There were following requests from the users/customers which can’t be addressed in the current design:

- running Talos network booted (i.e. with `STATE` / `EPHEMERAL` on `tmpfs`)
- running Talos installed e.g. to the SBC’s SD card, but mounting `/var` from an NVMe/SSD
- Azure VMs with directly attached NVMes will always install Talos to the network volume, so locally attached NVMe can’t be used for e.g. `containerd` state
- splitting `etcd` data directory to a separate disk
- performing wipe operations on the contents of the `/var`:
    - wiping system directories (e.g. `containerd` state)
    - wiping user data (e.g. `/var/mnt/foo`)
- disk encryption of the user disks (volumes)
- read-only user disks mounts (e.g. mounting my precious photo archive to the Talos machine, making sure that Talos never touches the contents)
- volume/disk management operations without reboot:
    - wiping user disks/volumes
    - wiping system areas
- creating additional user-managed partitions on the system disk:
    - `/data`
    - swap space
- container image cache storage
- some persistent across reboots logs storage (e.g. storing installation logs during staged upgrades)

The proposed design provides an option to solve the issues mentioned above.

## Groundwork

Before we move into volume management operations, there is some amount of work that needs to be done to improve the blockdevice management operations:

- Talos should quickly and easily detect various filesystem/partition types, including the most common ones, the ones that can be used in Kubernetes clusters. The minimum detection of the filesystem/partition prevents the disk from being considered empty and eligible for allocation
- The blockdevices/partitions should be presented as resources in a way that it allows to render them as a tree presenting a user with a view of available block devices, their types, partitions, filesystem types, available space, etc.
- Talos should detect and reliably show information which matches various standard Linux tools, e.g. `blkid`, to allow easier identification of storage objects.

## Installation Process

Talos installation should do the bare minimum to make sure that Talos can be booted from the disk, without touching the pieces which are not strictly required to boot Talos. This might include installing Talos without having machine configuration.

So the install should only touch following partitions/objects:

[...]

PR #8876, opened 2024-06-07
feat: provide disk detection based on new blockdevices

smira:
Uses go-siderolabs/go-blockdevice/v2 for all the hard parts, provides new resource `Disk` which describes all disks in the system.

Additional resource `SystemDisk` always point to the system disk (based on the location of `META` partition).

The `Disks` API (and `talosctl disks`) provides a view now into the `talosctl get disks` to keep backwards compatibility.

QEMU provisioner can now create extra disks of various types: IDE, AHCI, SCSI, NVME, this allows to test detection properly.

The new resource will be the foundation for volume provisioning (to pick up the disk to provision the volume on).

Part of #8367 

Example:

```
talosctl -n 172.20.0.5 get disks
NODE         NAMESPACE   TYPE   ID        VERSION   SIZE          READ ONLY   TRANSPORT   ROTATIONAL   WWID                                                               MODEL            SERIAL
172.20.0.5   runtime     Disk   loop0     1         65568768      true
172.20.0.5   runtime     Disk   nvme0n1   1         10485760000   false       nvme                     nvme.1b36-6465616462656566-51454d55204e564d65204374726c-00000001   QEMU NVMe Ctrl   deadbeef
172.20.0.5   runtime     Disk   sda       1         10485760000   false       virtio      true                                                                            QEMU HARDDISK
172.20.0.5   runtime     Disk   sdb       1         10485760000   false       sata        true         t10.ATA     QEMU HARDDISK                           QM00013        QEMU HARDDISK
172.20.0.5   runtime     Disk   sdc       1         10485760000   false       sata        true         t10.ATA     QEMU HARDDISK                           QM00001        QEMU HARDDISK
172.20.0.5   runtime     Disk   vda       1         12884901888   false       virtio      true
```


frezbo:
does it make sense to show a human readable size field?

smira:
> does it make sense to show a human readable size field?

yeah, I thought about it, but it probably needs some changes in COSI to add additional formatting options, so I left it out for now.

frezbo (approved):

PR #8879, opened 2024-06-08
chore: fix our dns server implementation

DmitriyMV:
This PR does those things:
* No longer shuffles dns servers for each request.
* Sets a context timeout of 4.5 seconds.
* Correctly returns a proper error from the root layer.


frezbo (approved):

frezbo:
do we need any tests for this? Since it's marked as fix

PR #8887, opened 2024-06-10
fix: detect CD devices, fix user disks wipe test

smira:
Detect CD devices, and set size to 0 for CD without media.

In user disk wipe tests, skip device mapper devices and CD-ROM.


DmitriyMV (approved):

smira:
> I think the disk finding logic for extensions tests might also need an update

let me dig that as well

Issue #8643, opened 2024-04-23
Add SideroLink endpoint to Talos dashboard


rothgar:
## Feature Request
It would be great to have the SideroLink domain available in the console dashboard (when used). Because the API is locked down to Omni/SideroLink we need to know where that configuration can come from.

### Description
An example config that might work in the dashboard
![image](https://github.com/siderolabs/talos/assets/371796/56a356e9-99b7-4dde-87c4-816e0ef96de7)

We could show status of the link (if there's room), but it's probably easier just to show the status with a font/color eg (red and italicized for disconnected and green and bold for connected). I mention using a font because I am red green color blind and if we only rely on colors it can cause problems. Another option for status is a unicode check ✓ (U+2713) and ✗ (U+2717) if those characters are available.

smira:
Unicode characters are fun with Linux console, as it usually doesn't have full Unicode font, but rather maps some Unicode glyphs to the font entries. (But Unicode is fine when `talosctl dashboard` runs in the terminal).

PR #8909, opened 2024-06-14
feat: show siderolink status on dashboard

utkuozdemir:
Add a new resource, `SiderolinkStatus`, which combines the following info:
- The Siderolink API endpoint without the query parameters or fragments (potentially sensitive info due to the join token)
- The status of the Siderolink connection

This resource is not set as sensitive, so it can be retrieved by the users with `os:operator` role (e.g., using `talosctl dashboard` through Omni).

Make use of this resource in the dashboard to display the status of the Siderolink connection.

Additionally, rework the status columns in the dashboard to:
- Display a Linux terminal compatible "tick" or a "cross" prefix for statuses in addition to the red/green color coding.
- Move and combine some statuses to save rows and make them more even.

Closes siderolabs/talos#8643.

## Screenshots
qemu-1
qemu-2
cli-1
cli-2


frezbo:
could you add a screenshot?

internal/app/machined/pkg/controllers/siderolink/status_test.go
```diff
@@ -0,0 +1,132 @@
...
+	rtestutils.AssertResources(suite.Ctx(), suite.T(), suite.State(), []resource.ID{siderolink.StatusID},
+		func(c *siderolink.Status, assert *assert.Assertions) {
+			assert.Equal(endpoint, c.TypedSpec().APIEndpoint)
+			assert.Equal(connected, c.TypedSpec().Connected)
+		})
+}
+
+type mockWgClient struct {
+	device *wgtypes.Device
+	lock   sync.Mutex
```
DmitriyMV, line 109: Minor nit — usually its `mx` and it's before protected fields. I know `vet` will complain, so just silence it.

internal/pkg/dashboard/components/components.go
```diff
@@ -75,14 +75,26 @@ func formatStatus(status any) string {
...
+		return formatText(statusStr, true)
 	case "stopped", "unhealthy", "false":
-		return fmt.Sprintf("[red]%s[-]", statusStr)
+		return formatText(statusStr, false)
 	default:
 		return statusStr
 	}
 }
 
+func formatText(text string, status bool) string {
```
DmitriyMV, line 86: Maybe `status` to `ok`?

DmitriyMV (approved):
🆒

utkuozdemir:
> could you add a screenshot?

sure, will do a bit later

smira:
> could you add a screenshot?

and to make it perfect, screenshot from QEMU - if you need help with that, I can try it out myself

internal/app/machined/pkg/controllers/siderolink/status_test.go
smira: I think it's `mu`, not `mx`, and if just two fields, and a mock who cares :)

internal/pkg/dashboard/components/components.go
```diff
@@ -75,14 +75,26 @@ func formatStatus(status any) string {
...
+	if text == "" {
+		return ""
+	}
+
+	if status {
+		return fmt.Sprintf("[green]\u221A %s[-]", text)
```
smira, line 92: we can use the actual character here, it's Go, source is UTF-8!

this will make it a bit more readable

smira (approved):
waiting for a screenshot, cool :+1: 

let me know if you need help with QEMU

utkuozdemir:
> waiting for a screenshot, cool :+1: 
> 
> 
> 
> let me know if you need help with QEMU

Nope, all good, just AFK for a while. I'll provide both screenshots.

utkuozdemir:
@smira @frezbo added the screenshots

frezbo:
> @smira @frezbo added the screenshots
> 
> You guys have any idea on getting rid of `STATE` on left column, so we can save 1 more row?

don't see it yet :sweat_smile:

utkuozdemir:
> > @smira @frezbo added the screenshots
> > You guys have any idea on getting rid of `STATE` on left column, so we can save 1 more row?
> 
> don't see it yet 😅

here :) 
image

utkuozdemir:
> what does that even mean by `STATE: OK`? I think we can drop it, its confusing

It reports if the state partition is mounted. Also, if it's encrypted, it looks like `OK - encrypted`.

It's been there since a long time. I agree that it my itself doesn't make too much sense, but the encryption info is good.

frezbo:
> > what does that even mean by `STATE: OK`? I think we can drop it, its confusing
> 
> It reports if the state partition is mounted. Also, if it's encrypted, it looks like `OK - encrypted`.
> 
> It's been there since a long time. I agree that it my itself doesn't make too much sense, but the encryption info is good.

ahh now I remeber, the `encrypted` info i useful, i think added when we did the TPM thing, so now not so sure

smira:
One small nit: let's drop the protocol/port from the SideroLink, just the hostname. It helps to identify Omn instance, but it's not meant to provide the full URL

smira:
> ahh now I remeber, the `encrypted` info i useful, i think added when we did the TPM thing, so now not so sure

I'd say drop it... probably we could later merge Secureboot/disk encryption into something more helpful, like "Security: Secureboot/Disk encryption"

utkuozdemir:
Updated preview after changes:
preview

frezbo (approved):

PR #8950, opened 2024-06-30
feat: add support for configuring vlan filtering on the bridge

### LINKED ISSUE #8941: Bridge vlan_filtering support
## Feature Request
Add ability to configure VLAN filtering on a host bridge interface.

### Description
I'm trying to use KubeVirt on my Talos cluster, in a setup that has certain workload types separated by VLAN. Prior to Talos, this would be accomplished by setting the `vlan_filtering` field on a host bridge, and then using the `vlan` field with the `bridge` CNI in a multus `NetworkAttachmentDefinition` to allow the bridge to handle the VLAN filtering. However, Talos does not currently allow me to affect this configuration.

There are two unergonomic workarounds to the situation (outside of using an external CNI plugin):
* create a privileged pod with host networking to create and configure the bridge on each node
* create and maintain VLAN interfaces off the bridge for each VLAN, then have multus create its own bridges off of these vlan interfaces

Ideally this would be a parameter in `MachineConfig.network`.

Thanks!

smira:
Fixes #8941


frezbo (approved):

camrossi:
@smira  I am testing this and seems to be working fine however I am lost on how I actually trunk a VLAN up my link 

I have this config:

```
            - interface: br0
              bridge:
                interfaces:
                    - bond0
                stp:
                    enabled: false
                vlan:
                    vlanFiltering: true
              dhcp: false
```

And when I deploy a NAD the bridge is configure correctly on the K8s side of things:
```
port              vlan-id
bond0             1 PVID Egress Untagged
br0               1 PVID Egress Untagged
vethf3bcfd0f      1 Egress Untagged
                  3712 PVID Egress Untagged

```

but VLAN 3712 in this example is not added to bond0 so nothing works. I can't figure out how I add a vlan to the machine config. 

For now I can simply do:
`bridge vlan add vid 3712 dev bond0` from a priviledge container and all works fine but this should be done somehow from the machine config I think. 

What am I missing ?

smira:
> What am I missing ?

I guess you might need to have VLAN configured as well?

camrossi:
I did try to do that as well. But where should I put the config ?

I tried adding a vlan statement under the bridge and under the bond, but that doesn't work either as it create bond0.vlanID or BR0.vlanID but is not adding the vlan on the bond0 itself as a tagged vlan. 
I can't find any doc so I am going a bit blind in this 😅

smira:
I don't know what is the path exactly, but I guess you need to create a VLAN under the bond and add it to the bridge? 

See [docs](https://www.talos.dev/v1.7/reference/configuration/v1alpha1/config/#Config.machine.network.interfaces..vlans.)

```yaml
machine:
  network:
    interaces:
       - interface: bond0
         bond: 
             ...
         vlans:
             - vlanId: 25
       - interface: br0
         bridge:
            interfaces:
                - bond0.25
```

camrossi:
The config you show me is what I would do if I do not have VLAN Filtering and this would work but requires me to create as many bond and bridge interfaces as vlans I need for my KubeVirt VMs.

This article does a great work explaining this and makes a comparison with and without VLAN Filtering:
https://developers.redhat.com/blog/2017/09/14/vlan-filter-support-on-bridge

If I read #8941 correctly what we want is indeed, as explained in the article above, a way to configure a single bridge and bond interface and no VLAN interfaces.

The config I think we are looking for is something like this: (This is a copy paste from my OpenShift Cluster running NMState + KubeVirt)
```
# oc get nncp bond1-br0 -o yaml
apiVersion: nmstate.io/v1
kind: NodeNetworkConfigurationPolicy
metadata:
  name: bond1-br0
spec:
  desiredState:
    interfaces:
    - bridge:
        options:
          stp:
            enabled: false
        ports:
        - name: bond1
          vlan:
            mode: trunk
            trunk-tags:
            - id: 315
            - id: 316
            - id-range:
                max: 3799
                min: 3700
      name: br0
      state: up
      type: linux-bridge
```

If I look at my talos config I have:

```
    network:
        interfaces:
            - interface: bond0
              bond:
                deviceSelectors:
                    - driver: bcmgenet
                    - driver: r8152
                mode: 802.3ad
                lacpRate: fast
              dhcp: false
            - interface: br0
              bridge:
                interfaces:
                    - bond0
                stp:
                    enabled: false
                vlan:
                    vlanFiltering: true
              dhcp: false
```

What I am missing here is a way to tell what VLANs to trunk on Bond0 i.e. this section from my OpenShift Cluster:

```
        - name: bond1
          vlan:
            mode: trunk
            trunk-tags:
            - id: 315
            - id: 316
            - id-range:
                max: 3799
                min: 3700
```

If there is a way to do so I am really unable to find it in the doc. All I can do is creating VLAN sub-interfaces on the bond or the bridge but this is not what we need. 

Happy to open a new git issue if needed be but vlan filtering implemented as is, is not really solving the issue.

smira:
Yes, I guess what you're looking for is different from what was implemented in this PR.

The easiest for us would be to see the equivalent setup using e.g. `ip` CLI to understand what you're looking for.

camrossi:
Seems to me as well but I am not sure what is the use of this feature without trunks... Would be interesting to know how @e3b0c442 meant to use this. 

I will open a new issue then with the full set of commands I need using the ip commands, assuming is possible I will have to check as I always used the `bridge` commands

e3b0c442:
I didn't get back around to close the loop here and apologize for the incompleteness in my original issue; @camrossi is correct, we also need the ability to set the VLANs the bridge carries.

I think this article https://developers.redhat.com/blog/2017/09/14/vlan-filter-support-on-bridge#with_vlan_filtering explains it clearly.

So in essence, we need to be able to:
```bash
# 1. create the bridge and attach the upstream interfaces (implemented)
ip link add br0 type bridge
ip link set br0 up
ip link set eth0 master br0

# 2. Allow VLAN filtering on the bridge (implemented)
ip link set br0 type bridge vlan_filtering 1

# 3. Add the VLANs to the upstream interfaces (missing)
bridge vlan add dev eth0 vid 2 master
bridge vlan add dev eth0 vid 3 master
```

I don't believe that this can be done with just the `ip` command, it also requires the `bridge` command which is part of the `iproute` distribution.

smira:
thank you, if you as experts could create an issue on what's missing, we would be able to look into missing pieces and implement them, thank you!

PR #8962, opened 2024-07-03
chore: support version with and without `v` prefix

frezbo:
Support passing in version with and without `v` prefix to Talos machine config version contract parser.

### DIFF
--- a/pkg/machinery/config/contract.go
+++ b/pkg/machinery/config/contract.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 )
 
 // VersionContract describes Talos version to generate config for.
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/contract.go
```diff
@@ -39,6 +40,8 @@ var versionRegexp = regexp.MustCompile(`^v(\d+)\.(\d+)($|\.)`)
 
 // ParseContractFromVersion parses Talos version into VersionContract.
 func ParseContractFromVersion(version string) (*VersionContract, error) {
+	version = "v" + strings.TrimPrefix(version, "v")
```
smira, line 43: can we update the unit-test please?

smira (approved):

PR #8979, opened 2024-07-05
test: fix the integrtion tests for apply-config

smira:
They got broken after refactoring.

Also use this PR to test things before the release.


frezbo (approved):

pkg/machinery/config/configdiff/configdiff.go
```diff
@@ -0,0 +1,129 @@
...
+// Package configdiff provides a way to compare two config trees.
+package configdiff
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/hexops/gotextdiff"
```
DmitriyMV, line 14: > This repository has been archived by the owner on Feb 9, 2024. It is now read-only.

Can't we use https://github.com/sergi/go-diff?

pkg/machinery/config/configdiff/configdiff.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// One of the resources might be nil.
+func Diff(w io.Writer, oldCfg, newCfg config.Encoder) error {
+	var (
+		oldYaml, newYaml []byte
+		err              error
+	)
+
+	if oldCfg != nil {
+		oldYaml, err = oldCfg.EncodeBytes(encoder.WithComments(encoder.CommentsDisabled))
```
DmitriyMV, line 32: There is EncodeToString method, but it's rather ineffective. 

FWIW I think we can also do better here:
```diff
diff --git a/pkg/machinery/config/container/container.go b/pkg/machinery/config/container/container.go
index 1a75e9b02..b326c414b 100644
--- a/pkg/machinery/config/container/container.go
+++ b/pkg/machinery/config/container/container.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"strings"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/siderolabs/gen/xslices"
@@ -203,41 +204,58 @@ func (container *Container) Bytes() ([]byte, error) {
 
 // EncodeString configuration to YAML using the provided options.
 func (container *Container) EncodeString(encoderOptions ...encoder.Option) (string, error) {
-	b, err := container.EncodeBytes(encoderOptions...)
+	var buf strings.Builder
+
+	err := container.encodeBytes(&buf, encoderOptions...)
 	if err != nil {
 		return "", err
 	}
 
-	return string(b), nil
+	return buf.String(), nil
 }
 
 // EncodeBytes configuration to YAML using the provided options.
 func (container *Container) EncodeBytes(encoderOptions ...encoder.Option) ([]byte, error) {
 	var buf bytes.Buffer
 
+	err := container.encodeBytes(&buf, encoderOptions...)
+	if err != nil {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
+
+type buffer interface {
+	Len() int
+	Write(p []byte) (n int, err error)
+	WriteString(s string) (n int, err error)
+}
+
+func (container *Container) encodeBytes(buf buffer, encoderOptions ...encoder.Option) error {
 	if container.v1alpha1Config != nil {
 		b, err := encoder.NewEncoder(container.v1alpha1Config, encoderOptions...).Encode()
 		if err != nil {
-			return nil, err
+			return err
 		}
 
-		buf.Write(b)
+		buf.Write(b) //nolint:errcheck
 	}
 
 	for _, doc := range container.documents {
 		if buf.Len() > 0 {
-			buf.WriteString("---\n")
+			buf.WriteString("---\n") //nolint:errcheck
 		}
 
 		b, err := encoder.NewEncoder(doc, encoderOptions...).Encode()
 		if err != nil {
-			return nil, err
+			return err
 		}
 
-		buf.Write(b)
+		buf.Write(b) //nolint:errcheck
 	}
 
-	return buf.Bytes(), nil
+	return nil
 }
 
 // Validate checks configuration and returns warnings and fatal errors (as multierror).
```

And replace above and below calls with `EncodeString` dropping `string(...)` conversation below.

pkg/machinery/config/configdiff/configdiff.go
smira: I would rather use this one, we already use it in Omni without issues.

pkg/machinery/config/configdiff/configdiff.go
smira: let's keep it for another PR, it's a good thing, but diff is not a hot path, so it's not any better/worse than it was before

pkg/machinery/config/configdiff/configdiff.go
smira: And it doesn't support unified diffs, so it's useless here

pkg/machinery/config/configdiff/configdiff.go
smira: my goal is to make tests pass and make a release, everything else can wait for another PR after the release

DmitriyMV (approved):

PR #8980, opened 2024-07-05
fix: properly output multi-doc machine config in `get mc`

### LINKED ISSUE #8687: YAML output of machine config incorrectly formatted for multidoc configurations
## Bug Report
Multidoc YAML configurations after the first document are incorrectly indented 4 spaces
### Description
if you edit the machine config and append a new document for the ingress firewall rules, it displays correctly with `talosctl edit mc`, but the output of `talosctl get mc` will show the documents after the first YAML doc incorrectly indented, which breaks processing with `yq`.

The multi-docs are applied correctly by Talos - it is just in display they are mal-formatted.

### Logs

e.g. append a YAML document End  via editing the machine config thus:
```
  #       apiVersion: v1
  #       kind: Namespace
  #       metadata:
  #             name: ci
    
... 
```
and it will show thus via `get mc -o yaml`
```
      #       apiVersion: v1
      #       kind: Namespace
      #       metadata:
      #       	name: ci
    
    ...
```
### Environment

- Talos version: 1.7
- Platform:
- This happens with `talosctl` and also `omnictl get redactedclustermachineconfig`

DmitriyMV:
For `get mc -o json|yaml` we pretend that `spec` field is string and not an actual yaml map. That way you can see the full spec in unformatted view using `talosctl -n <node> get mc -o yaml` or formatted using `talosctl -n <node> get mc -o yaml | yq .spec`.

`edit mc` command is unaffected.

Fixes #8687

### DIFF
--- a/cmd/talosctl/cmd/talos/edit.go
+++ b/cmd/talosctl/cmd/talos/edit.go
@@ -58,8 +58,7 @@ func editFn(c *client.Client) func(context.Context, string, resource.Resource, e
 			return errors.New("only the machineconfig resource can be edited")
 		}
 
-		metadata := mc.Metadata()
-		id := metadata.ID()
+		id := mc.Metadata().ID()
 
 		body, err := yaml.Marshal(mc.Spec())
 		if err != nil {
--- a/cmd/talosctl/cmd/talos/output/json.go
+++ b/cmd/talosctl/cmd/talos/output/json.go
@@ -13,6 +13,8 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource/meta"
 	"github.com/cosi-project/runtime/pkg/state"
 	yaml "gopkg.in/yaml.v3"
+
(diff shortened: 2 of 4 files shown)


DmitriyMV:
This looks and feels like hack, but I haven't found a better way :/

DmitriyMV:
Output example:
image

and with `... | yq .spec`
image

smira (approved):

DmitriyMV:
@frezbo

> actually I don't like this change, previous output of get mc -o yaml was really pretty, now have to use another tool to view that

Yeah, we don't either, but the problem is how `yaml` module formats strings which contains `\n<space><space>\n`. We could fix those strings, but we would rather do not edit user typed machine configs. And because `edit mc` already provides you a nice formatted text I don't think many people will notice. And there is also a problem about the fact that current multi-doc yaml being inside `spec` field does not work well.

frezbo:
got it, Andrey explained why this change, all good

PR #9001, opened 2024-07-10
fix: remove host bind mount for `/tmp` for trustd

smira:
Not sure why this mount was needed, but it was added long time ago, and I believe it's no longer needed.


Unix4ever (approved):

frezbo (approved):
found from the tmpfs filling up issue?

smira:
> found from the tmpfs filling up issue?

I found this by accident looking into that, but not related, as tmpfs is filling up on workers, and this is a controlplane only thing.

PR #9002, opened 2024-07-10
feat: provide machine config document to update trusted CA roots

### LINKED ISSUE #8867: Allow configuration of custom CAs via the machine configuration
## Feature Request
A mean to add a custom Certification Authority to the trusted CA certificates used by talos images to make them able to connect to a Omni server on prem that as been certified by that custom CA.
 
### Description
I am using Omni On Prem on a VM and inside our network, our tls certificates are delivered by a private custom CA. So to be able to connect to Omni through https, the talos kernel which is using a specific parameter (siderolinks) needs to trust the certificate annonced by Omni. For that, the CA certificate must be trusted inside the talos image at boot time.

I know that it is possible to build a custom image for talos but I think it is not really a convenient way to manage images and upgrades on a large scale (in a "industrial" way).

That's why I would like very much to be able to add the certificates of a custom CA in the trusted certificates list at boot time or through image factory for example ?

It would be great. 

Thanks, 
Regards

smira:
Fixes #8867


smira:
Need some documentation/tests still.

internal/app/machined/pkg/controllers/secrets/trusted_roots.go
```diff
@@ -0,0 +1,95 @@
...
+// TrustedRootsController manages CA trusted roots based on configuration.
+type TrustedRootsController struct{}
+
+// Name implements controller.Controller interface.
+func (ctrl *TrustedRootsController) Name() string {
+	return "secrets.TrustedRootsController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *TrustedRootsController) Inputs() []controller.Input {
```
frezbo, line 38: hmm i guess this needs to happen way early, for eg an extension that might call out to a https server and depends on the host ca as a mount

internal/app/machined/pkg/controllers/secrets/trusted_roots.go
```diff
@@ -0,0 +1,95 @@
...
+
+		cfg, err := safe.ReaderGetByID[*config.MachineConfig](ctx, r, config.V1Alpha1ID)
+		if err != nil && !state.IsNotFoundError(err) {
+			return fmt.Errorf("failed to get machine config: %w", err)
+		}
+
+		contents := slices.Clone(defaultCACertificates)
+
+		if cfg != nil {
+			contents = slices.Concat(contents, []byte(strings.Join(cfg.Config().TrustedRoots().ExtraTrustedRootCertificates(), "\n\n")))
```
frezbo, line 76: should we validate user input, would a random string appended to trusted CA break any tools?

internal/app/machined/pkg/controllers/secrets/trusted_roots.go
smira: I guess we should at least wait for the initial CAs to be populated before starting various services

frezbo: maybe we just write the one from ca-certificates if no config is available?

smira: That's what this PR does, controller always writes the default one, and appends extra CAs as configured.

frezbo: ahhh I missed that

> I guess we should at least wait for the initial CAs to be populated before starting various services

i guess we can add a dependency on etcfilespecstatus?

smira: yep, that's the plan!

Unix4ever (approved):
:cool:

PR #9004, opened 2024-07-11
chore: make using action tracker easier

frezbo:
Refactor so that action tracker accepts an interface.


frezbo:
This is so that the TF provider can use action tracker with an existing client

cmd/talosctl/pkg/talos/action/tracker.go
```diff
@@ -121,19 +119,18 @@ func WithDebug(debug bool) TrackerOption {
...
 	opts ...TrackerOption,
 ) *Tracker {
 	tracker := Tracker{
 		expectedEventFn:          expectedEventFn,
 		actionFn:                 actionFn,
-		nodeToLatestStatusUpdate: make(map[string]reporter.Update, len(cliContext.Nodes)),
+		nodeToLatestStatusUpdate: make(map[string]reporter.Update, len(clientExecutor.NodeList())),
 		reporter:                 reporter.New(),
 		reportCh:                 make(chan nodeUpdate),
-		isTerminal:               isatty.IsTerminal(os.Stderr.Fd()),
```
smira, line 135: this shouldn't be dropped? or we should allow to override it via an option?

frezbo: we could, this has no effect on the provider, so i ignored, I could make it an option

smira: but it shouldn't be removed either way

frezbo: oh, that was my mistake, forgot to add it back

PR #9005, opened 2024-07-11
feat: provide an option to enforce SecureBoot for TPM enrollment

### LINKED ISSUE #8995: TPM Encryption does not fail if secureboot is not enabled
### Description
If the host machine has secureboot disabled, but the TPM available, Talos proceeds to install. There is no way to restrict the encryption. An attacker could boot an alternate OS and retrieve the encryption keys.

It would be great to have an additional config parameter in the tpm encryption requiring secure boot to be enabled, else disk encryption fails.

smira:
Fixes #8995

There is no security impact, as the actual SecureBoot state/configuration is measured into the PCR 7 and the disk encryption key unsealing is tied to this value.

This is more to provide a way to avoid accidentally encrypting to the TPM while SecureBoot is not enabled.

### DIFF
--- a/internal/pkg/encryption/keys/tpm2.go
+++ b/internal/pkg/encryption/keys/tpm2.go
@@ -8,8 +8,10 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"fmt"
 	"io"
 
+	"github.com/foxboron/go-uefi/efi"
 	"github.com/siderolabs/go-blockdevice/blockdevice/encryption"
 	"github.com/siderolabs/go-blockdevice/blockdevice/encryption/luks"
 	"github.com/siderolabs/go-blockdevice/blockdevice/encryption/token"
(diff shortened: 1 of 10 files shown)


internal/pkg/encryption/keys/tpm2.go
```diff
@@ -32,17 +34,30 @@ type TPMToken struct {
...
 
 // NewKey implements Handler interface.
 func (h *TPMKeyHandler) NewKey(ctx context.Context) (*encryption.Key, token.Token, error) {
+	if h.checkSecurebootOnEnroll {
+		if !efi.GetSecureBoot() {
+			return nil, nil, fmt.Errorf("failed to enroll the TPM2 key, as SecureBoot is disabled")
+		}
+
+		if efi.GetSetupMode() {
+			return nil, nil, fmt.Errorf("failed to enroll the TPM2 key, as the system is in setup mode")
```
frezbo, line 57: i'm confused on the error message

frezbo: ahh now makes sense, since this has nothing to do with tpm2 enrollment, we could say not continuing due to secureboot enforcement check

smira: yes, thanks, I will update the message as well

smira: updated now

frezbo (approved):

PR #9021, opened 2024-07-16
chore: ensure tls required on s3 buckets

TimJones:
Ensure that non-TLS connections are denied to S3 buckets via policy. Required for compliance.

### DIFF
--- a/hack/cloud-image-uploader/aws.go
+++ b/hack/cloud-image-uploader/aws.go
@@ -22,6 +22,28 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var denyInsecurePolicyTemplate = `{
+  "Id": "ExamplePolicy",
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "AllowSSLRequestsOnly",
+      "Action": "s3:*",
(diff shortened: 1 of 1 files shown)


hack/cloud-image-uploader/aws.go
```diff
@@ -139,6 +140,42 @@ func (au *AWSUploader) registerAMI(ctx context.Context, region string, svc *ec2.
 		}
 	}()
 
+	denyInsecurePolicy := map[string]interface{}{
```
frezbo, line 143: let's just define this as a var and just do `fmt.Sprintf` and use that as string

TimJones: Done, but imo that looks way worse :sweat_smile:

frezbo: oh  I meant define as a `var ()` block at the top, but nothing critical here

frezbo (approved):

hack/cloud-image-uploader/aws.go
TimJones: OK, I redid that then :)

frezbo (approved):

PR #9022, opened 2024-07-16
feat: provide a simple way to delete keys in config merge

smira:
With strategic merge patches, we had no way to delete values by patching.

It might be tricky to implement it fully, but one simple case is easy - deleting keys from the map (e.g. `nodeLabels`) on merge.


TimJones (approved):

smira:
Marking this as draft for now:

- can we do a more complete solution?
- needs documentation updates and release notes either way

PR #9026, opened 2024-07-17
feat: implement specifying filesystem for user disks

samip5:
# Pull Request

## What? (description)

Implement specifying filesystem for user disks and fallback to xfs as before if omitted.

## Why? (reasoning)

Not all disks will always be formatted as xfs, and thus it couldn't mount other partitions even if kernel has the required support like ext4.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -952,8 +952,13 @@ func mountDisks(logger *log.Logger, r runtime.Runtime) (err error) {
 				}
 			}
 
+			filesystemType := disk.FileSystemType()
+			if filesystemType == "" {
+				filesystemType = "xfs"
+			}
+
 			mountpoints.Set(partname,
-				mount.NewMountPoint(partname, part.MountPoint(), "xfs", unix.MS_NOATIME, "",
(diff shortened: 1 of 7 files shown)


samip5:
It's probably not as simple as I had thought.

```
192.168.2.129: user: warning: [2024-07-17T04:57:56.597248761Z]: error mounting "/dev/sdc1": error mounting: 1 error(s) occurred:
192.168.2.129: user: warning: [2024-07-17T04:57:56.681674761Z]:  error repairing: xfs_repair: unsupported filesystem type: ext4
```

samip5:
> We're going with #8367 which should make this obsolete by the time 1.8 is released (hopefully).

Will it support not creating new volumes, but instead mount existing ones?

smira:
Probably, but either way we don't want to make changes to the existing code.

samip5:
> Probably, but either way we don't want to make changes to the existing code.

So currently to do what I need, aka mount the partitions and run mergerfs on top of it, would be to use a privileged container that does all that?

smira:
I don't have any answer today for `mergerfs`, it's not our priority at the moment. Once volume management lands, we can start talking about all these extra filesystems.

samip5:
This PR is in a state where it would work if the container had all the extra binaries but it doesn't...

samip5:
> I don't have any answer today for `mergerfs`, it's not our priority at the moment. Once volume management lands, we can start talking about all these extra filesystems.

Can we yet start talking about these?

smira:
Talking - absolutely. Once we land confirmed features for 1.9, we'll look once again into user volumes.

PR #9027, opened 2024-07-17
feat(talosctl): append microsoft secure boot certs

jfroy:
# Pull Request

feat(talosctl): append microsoft secure boot certs

## What? (description)

This is an initial proposal to support adding the well-known Microsoft UEFI certificates (DB and KEK) to secure boot ESLs generated by talosctl. The DB certificates allow OptionROMs and other UEFI programs like IPXE to be trusted and loaded. The KEK certificates allow the public revocation list to be trusted and loaded.

## Why? (reasoning)

OptionROMs are important pieces of firmware for many use cases. An operator should be able to make the decision to trust the Microsoft UEFI DB certificates to use them. There is practically no chance for an operator to re-sign OptionROMs with their own keys. The default remains off.

The UEFI revocation database is an important part of maintaining the secure boot chain of trust by disallowing UEFI programs with known vulnerabilities. It would be burdensome for operators to maintain a separate list, at minimum requiring re-signing with their own KEK. Operations should make the choice to trust the Microsoft UEFI KEK certificates instead. The default remains off.

## Acceptance


jfroy:
This is POC-ish.

I am pretty sure the certs should not be downloaded on demand, but maybe they should not be loaded in `init` or cause a panic on parse fail. But then again, loading the embedded data should never fail -- if it does fail, the binary is corrupt.

Maybe this should not be so Microsoft-centric and be more generic -- "here's a bunch of other KEK/DB certs, please include them". But then again, the Microsoft certs are special in the real-world ecosystem -- all OptionsROMs that are signed will be signed by these certificates.

jfroy:
Also, of course, please verify the signature of the included certificates before signing the PR!

smira:
Thank you, that makes sense. I can take a look and see if we should probably restructure this a bit, and probably support additional certs while "microsoft" being one of the default choices.

frezbo:
I think we can put the certs in pkgs, so that we verify the checksum and just pull in that layer here, less binary data in talos code

smira:
> I think we can put the certs in pkgs, so that we verify the checksum and just pull in that layer here, less binary data in talos code

I think it's fine to keep it in Talos as well, but we need to document how we downloaded them.

smira:
I did some mostly cosmetic changes, and changed the way MS certificates are acquired.

Unix4ever (approved):

Dockerfile
```diff
@@ -309,6 +309,16 @@ FROM scratch AS ipxe-generate
 COPY --from=pkg-ipxe-amd64 /usr/libexec/snp.efi /amd64/snp.efi
 COPY --from=pkg-ipxe-arm64 /usr/libexec/snp.efi /arm64/snp.efi
 
+FROM scratch AS microsoft-secureboot-database
+ADD https://github.com/microsoft/secureboot_objects.git /
```
frezbo, line 313: hmm so once cached we don't get any updates?

smira: yes, unless the upstream repository changes, and we'll see a change here in `git diff` (but it shouldn't change of course)

frezbo (approved):

PR #9034, opened 2024-07-18
 docs: update deploying cilium docs for v1.8

ChickenIQ:
# Pull Request

## What? (description)
Replace quay.io/cilium/cilium-cli-ci:latest with alpine/helm:latest and modify the command accordingly in "Method 5: Using a job".

## Why? (reasoning)
The current method does not support arm64 as there are no available images for this architecture. This solves that problem and is a drop-in replacement.

## Acceptance
Most of the tasks in this checklist have generated unnecessary commits by removing newlines from existing code, so I decided not to include them.


website/content/v1.8/kubernetes-guides/network/deploying-cilium.md
```diff
@@ -357,8 +357,11 @@ We can leverage this to our advantage to install Cilium by using an inline manif
                 - name: KUBERNETES_SERVICE_PORT
                   value: "6443"
                 command:
-                  - cilium
+                  - helm
                   - install
+                  - cilium
```
smira, line 362: is it a typo or should it be two times?

I would want some input from @leppeK on this change, as he brought this change initially

ChickenIQ: I believe it is correct.
helm install [NAME] [CHART] [flags]

leppeK: For helm it is correct, unfortunately helm leaves traces of the release which you might not want.

I'll see if I can convince Cilium to provide an official container image for all platforms

leppeK: I think the comments already indicate that you can use this method to do anything, if you want to use this to install something else using helm or whatever it works as well.


A use case could be installing your CNI and/or ArgoCD at bootstrap time

ChickenIQ: Isn't it the same as manually installing using helm after bootstrap?

leppeK: Doing stuff at bootstrap means you can do it completely declaratively and in one go and at the right time... Great for reproducibilty

ChickenIQ:
Added the namespace. Previously it was trying to install to the "default" namespace instead of "kube-system". I somehow managed to not include it in the commit.
Any progress on getting the cilium-cli-ci image built for arm64?

leppeK:
I've asked over at Cilium to provide an official image so we'd have to wait for that

leppeK:
https://github.com/cilium/cilium-cli/issues/2755

leppeK:
> Added the namespace. Previously it was trying to install to the "default" namespace instead of "kube-system". I somehow managed to not include it in the commit. Any progress on getting the cilium-cli-ci image built for arm64?

can you test docker pull leppek/cilium-cli:latest if it works for you?

PR #9039, opened 2024-07-20
docs: use ConfigMap for Cilium job based installation example

mattwillsher:
# Pull Request

## What? (description)

Passing the command args to cilium CLI results in an error where the list elements are interpreted as a single argument, producing an error. Attempts to split them out lead to issues parsing the capabilities list.

Using a ConfigMap avoids these issues.

#9034 changes the same example.

## Why? (reasoning)

I don't understand why the example in the docs didn't work for me, but this new approach worked both when applied directly and when run via the talos Terraform provider.

## Acceptance


mattwillsher:
I'm happy that this is closed if it is not appropriate.

Issue #9007, opened 2024-07-11
provide memfd-bind for runc


smira:
More information: https://github.com/opencontainers/runc/tree/main/contrib/cmd/memfd-bind

(We probably just need to more or less copy the code and make it part of core Talos to have that memfd-bind always available).

The reason: if you take all the space in host `/tmp` filesystem (e.g. with simple `dd` from a privileged container), the `kubectl exec` and friends start failing as `runc` can't copy itself: 

```
failed to exec in container: failed to start exec "9d300e09a4fd9ff6e586c548087a39658fffc3300b0562ff0367a54c8a671721": OCI runtime exec failed: write /tmp/runc-process3947626735: no space left on device: unknown
```

PR #9069, opened 2024-07-26
feat: runc memfd-bind service

frezbo:
Add a `runc-memfd-bind` service so that runc binary is not copied for every `runc` invocation.

Fixes: #9007.


internal/app/runc_memfd_bind/runc_memfd_bind.go
```diff
@@ -0,0 +1,135 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
frezbo, line 3: should we also copy over the license text from source?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -632,6 +632,15 @@ func StartMachined(_ runtime.Sequence, _ any) (runtime.TaskExecutionFunc, string
 	}, "startMachined"
 }
 
+// StartRuncMemfdBind represents the task to start runc-memfd-bind.
+func StartRuncMemfdBind(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
```
smira, line 636: I almost feel like it would be easier to run this is a simple controller without any inputs/outputs.

Containerd doesn't depend on this (it needs for the first invocation of runc only, and even if not available, runc would still work).

If we want to make a dependency, we could have used an output resource.

I prefer controllers over anything else, as it's the way forward from the code point of view.

internal/app/runc_memfd_bind/runc_memfd_bind.go
smira: if we copied a significant amount of code, let's add the original license and link in the commands

smira (approved):
main question - if we exhaust `/tmp` space, can we still do `kubectl exec`?

PR #9078, opened 2024-07-29
fix: update containerd configuration and settings

smira:
Provide `XDG_RUNTIME_DIR` environment variable, this specifically fixes the `kubectl exec` action when `/tmp` is filled up.

Update containerd configuration to version 3 and fix it up.

### DIFF
--- a/hack/containerd.toml
+++ b/hack/containerd.toml
@@ -1,7 +1,9 @@
-version = 2
+version = 3
 
 disabled_plugins = [
     "io.containerd.grpc.v1.cri",
+    "io.containerd.cri.v1.images",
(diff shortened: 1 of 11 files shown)


frezbo (approved):

frezbo:
added the label for the nvidia tests, since those runtimes use `XDG` vars, want to see if we broke something

smira:
@frezbo do you I think I broke nvidia?

frezbo:
> @frezbo do you I think I broke nvidia?

let me rerun and check

smira:
Reworked `XDG_RUNTIME_DIR` injection.

frezbo (approved):

PR #9085, opened 2024-07-30
chore: add test for apparmor

frezbo:
Add a test that verifies pods can be scheduled with `RuntimeDefault` apparmor profile.

### DIFF
--- a/internal/integration/k8s/apparmor.go
+++ b/internal/integration/k8s/apparmor.go
@@ -1,0 +1,74 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_k8s
+
+package k8s
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"strings"
+	"time"
+
+	"github.com/siderolabs/talos/internal/integration/base"
+)
+
+// ApparmorSuite verifies that a pod with apparmor security context with `RuntimeDefault` works.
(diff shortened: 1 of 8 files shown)


internal/integration/k8s/apparmor.go
```diff
@@ -0,0 +1,55 @@
...
+var apparmorPodSpec []byte
+
+// SuiteName returns the name of the suite.
+func (suite *ApparmorSuite) SuiteName() string {
+	return "k8s.ApparmorSuite"
+}
+
+// TestApparmor verifies that a pod with apparmor security context with `RuntimeDefault` works.
+func (suite *ApparmorSuite) TestApparmor() {
+	if suite.Cluster == nil {
```
smira, line 32: can we detect (e.g. by reading something) that AppArmor is enabled and skip the test if not?

I wonder if we should run by default without AppArmor enabled, and enable it only for some integration-misc pipeline with AppArmor enabled?

frezbo: > can we detect (e.g. by reading something) that AppArmor is enabled and skip the test if not?

yeh, that's easy

> I wonder if we should run by default without AppArmor enabled, and enable it only for some integration-misc pipeline with AppArmor enabled?

I was trying skip adding a new pipeline since it means a new cluster and takes more time

smira: > I was trying skip adding a new pipeline since it means a new cluster and takes more time

I want to make sure that if we keep the default to be AppArmor to be disabled, we run most of the tests with AppArmor disabled

so the cost of having another pipeline makes sense imho

smira (approved):

PR #9093, opened 2024-07-31
feat: publish installed extensions as node labels/annotations

### LINKED ISSUE #8971: Add .machine.nodeAnnotations
## Feature Request

I'd like to set a node annotation in the same way that I can currently set a node label using the machine configuration.

### Description

There may be other uses for this feature, but see the [Longhorn documentation for customizing default disks for new nodes](https://longhorn.io/docs/1.6.2/nodes-and-volumes/nodes/default-disk-and-node-config/#customizing-default-disks-for-new-nodes) for the one I'm thinking of. When my cluster comes up, I want Longhorn to be configured to use some dedicated disks on some of the nodes without manual intervention. This way of declaratively configuring Longhorn uses a node label and annotation for the node specific (non-global) parameters.

I can currently configure a single, Longhorn-dedicated disk per node if I mount it to the same place on each node (location configured with Longhorn `default-data-path` global option). If I had a way to add annotations to the node before Longhorn comes up, I could automatically add multiple disks per node to Longhorn when I create the cluster. Then I could set my machine configuration to something like this:

```yaml
machine:
  disks:
    - device: "/dev/sdb"
      partitions:
        - mountpoint: "/var/mnt/disk1"
    - device: "/dev/sdc"
      partitions:
        - mountpoint: "/var/mnt/disk2"
  nodeLabels:
    node.longhorn.io/create-default-disk: config
  nodeAnnotations:  # doesn't exist yet!
    node.longhorn.io/default-disks-config: |
      [
          { 
              "name":"disk1",
              "path":"/var/mnt/disk1"
          },
          { 
              "name":"disk2",
              "path":"/var/mnt/disk2"
          }
      ]
```

Alternatives for my use case:

* `kubectl annotate node ...` before deploying Longhorn works, but I'm trying to deploy the entire cluster and workload declaratively.
* I can just add the disks manually using the Longhorn web UI.

### LINKED ISSUE #9089: Put the schematic ID in a node annotation
## Feature Request
Put the schematic ID in a node annotation

### Description
It would be helpful for automations like [`system-upgrade-controller`](https://github.com/buroa/system-upgrade-controller/blob/master/examples/talos.yaml) to have the Talos schematic ID in a node annotations (ie `cluster.talos.dev/schematic-id`) so it can be used to pull the correct [installer image](https://github.com/buroa/system-upgrade-controller/blob/master/examples/talos.yaml#L48).

Currently my cluster needs two different schematic IDs so I cannot set a single value in my `system-upgrade-controller` plan.

Thanks!

smira:
Extensions are posted the following way:

`extensions.talos.dev/<name>=<version>`

The name should be valid as a label (annotation) key.

If the value is valid as a label value, use labels, otherwise use annotations.

Also implements node annotations in the machine config as a side-effect.

Fixes #9089

Fixes #8971

See #9070


frezbo (approved):

hack/release.toml
```diff
@@ -108,6 +108,25 @@ Talos Linux on config generation now adds a label `node.kubernetes.io/exclude-fr
...
+"""
+
+    [notes.kubelet]
+        title = "Extensions in Kubernetes Nodes"
+        description = """\
+Talos Linux now publishes list of installed extensions as Kubernetes node labels/annotations.
+
+The key format is `extensions.talos.dev/<name>` and the value is the extension version.
+If the extension name is not valid as a label key, it will be skipped.
+If the version is valid is a label value, it will be put to the label; otherwise it will be put to the annotation.
```
frezbo, line 126:
```suggestion
If the version is a valid label value, it will be put to the label; otherwise it will be put to the annotation.
```

internal/app/machined/pkg/controllers/k8s/node_label_spec.go
```diff
@@ -97,3 +111,27 @@ func (ctrl *NodeLabelSpecController) Run(ctx context.Context, r controller.Runti
...
+	}
+
+	for iter := extensionStatuses.Iterator(); iter.Next(); {
+		extensionStatus := iter.Value()
+
+		if extensionStatus.TypedSpec().Metadata.Name == "" {
+			continue
+		}
+
+		name := "extensions.talos.dev/" + extensionStatus.TypedSpec().Metadata.Name
```
frezbo, line 128: constant not used here

smira: nice catch, thank you!

frezbo (approved):

frezbo (approved):

PR #9095, opened 2024-08-01
docs: AWS getting started re-write

rothgar:
Updated with mulit-AZ subnets for control plane group and better copy/paste ability

### DIFF
--- a/website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
+++ b/website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
@@ -7,51 +7,104 @@ aliases:
 
 ## Creating a Cluster via the AWS CLI
 
-In this guide we will create an HA Kubernetes cluster with 3 worker nodes.
-We assume an existing VPC, and some familiarity with AWS.
+In this guide we will create an HA Kubernetes cluster with 3 control plane nodes across 3 availability zones.
+You should have an existing AWS account and have the AWS CLI installed and configured.
 If you need more information on AWS specifics, please see the [official AWS documentation](https://docs.aws.amazon.com).
 
-### Set the needed info
-
-Change to your desired region:
+To install the dependencies for this tutorial you can use homebrew on macOS or Linux:
 
 ```bash
-REGION="us-west-2"
-aws ec2 describe-vpcs --region $REGION
-
-VPC="(the VpcId from the above command)"
+brew install siderolabs/tap/talosctl kubectl jq curl xz
 ```
 
-### Create the Subnet
+If you would like to create infrastructure via `terraform` or `opentofu` please see the exmaple in the [contrib repository](https://github.com/siderolabs/contrib/tree/main/examples/terraform/aws).
 
-Use a CIDR block that is present on the VPC specified above.
+> Note: this guide is not a production set up and steps were tested in `bash` and `zsh` shells.
+
+### Create AWS Resources
+
+We will be creating a control plane with 3 Ec2 instances spread across 3 availability zones.
+It is recommended to not use the default VPC so we will create a new one for this tutorial.
+
+Change to your desired region and CIDR block and create a VPC:
+
+> Make sure your subnet does not overlap with `10.244.0.0/16` or `10.96.0.0/12` the [default pod and services subnets in Kubernetes]({{% relref "../../../introduction/troubleshooting.md#conflict-on-kubernetes-and-host-subnets" %}}).
 
 ```bash
-aws ec2 create-subnet \
-    --region $REGION \
-    --vpc-id $VPC \
-    --cidr-block ${CIDR_BLOCK}
+AWS_REGION="us-west-2"
+IPV4_CIDR="10.1.0.0/18"
+VPC_ID=$(aws ec2 create-vpc \
+    --cidr-block $IPV4_CIDR \
+    --output text --query 'Vpc.VpcId')
 ```
 
-Note the subnet ID that was returned, and assign it to a variable for ease of later use:
+### Create the Subnets
+
+Create 3 smaller CIDRs to use for each subnet in different availability zones.
+Make sure to adjust these CIDRs if you changed the default value from the last command.
 
 ```bash
-SUBNET="(the subnet ID of the created subnet)"
+IPV4_CIDRS=( "10.1.0.0/22" "10.1.4.0/22" "10.1.8.0/22" )
+```
+
+Next create a subnet in each availability zones.
+
+> Note: If you're using zsh you need to run `setopt KSH_ARRAYS` to have arrays referenced properly.
+
+```bash
+CIDR=0
+declare -a SUBNETS
+AZS=($(aws ec2 describe-availability-zones \
+    --query 'AvailabilityZones[].ZoneName' \
+    --filter "Name=state,Values=available" \
+    --output text | tr -s '\t' '\n' | head -n3))
+
+for AZ in ${AZS[@]}; do
+        SUBNETS[$CIDR]=$(aws ec2 create-subnet \
+            --vpc-id $VPC_ID \
+            --availability-zone $AZ \
+            --cidr-block ${IPV4_CIDRS[$CIDR]} \
+            --query 'Subnet.SubnetId' \
+            --output text)
+        aws ec2 modify-subnet-attribute \
+            --subnet-id ${SUBNETS[$CIDR]} \
+            --private-dns-hostname-type-on-launch resource-name
+        echo ${SUBNETS[$CIDR]}
+        ((CIDR++))
+done
+```
+
+Create an internet gateway and attach it to the VPC:
+
+```bash
+IGW_ID=$(aws ec2 create-internet-gateway \
+    --query 'InternetGateway.InternetGatewayId' \
+    --output text)
+
+aws ec2 attach-internet-gateway \
+    --vpc-id $VPC_ID \
+    --internet-gateway-id $IGW_ID
+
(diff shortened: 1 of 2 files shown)


website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -7,52 +7,102 @@ aliases:
...
+
+### Create AWS Resources
 
-VPC="(the VpcId from the above command)"
+We will be creating a control plane with 3 Ec2 instances spread across 3 availability zones.
+It is recommended to not use the default VPC so we will create a new one for this tutorial.
+
+Change to your desired region and CIDR block and create a VPC:
+
+> Make sure your subnet does not overlap with `10.244.0.0/16` or `10.96.0.0/12` the [default pod and services subnets in Kubernetes]({{% ref "/v1.7/introduction/troubleshooting.md#conflict-on-kubernetes-and-host-subnets" %}}).
```
smira, line 31: we should use `relref`, not `ref`, and without the version, to keep docs same across versions

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -7,52 +7,102 @@ aliases:
 
 ## Creating a Cluster via the AWS CLI
 
-In this guide we will create an HA Kubernetes cluster with 3 worker nodes.
-We assume an existing VPC, and some familiarity with AWS.
+In this guide we will create an HA Kubernetes cluster with 3 control plane nodes across 3 availability zones.
```
smira, line 10: I was always confused by spreading the etcd across AZes:

* increased latency
* a chance of network partitioning

As long as you don't have 3x workers spread across AZes same way, what kind of benefit does it give to the availability of the cluster?

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -106,191 +156,196 @@ Once the `SnapshotTaskDetail.Status` indicates `completed`, we can register the
...
-
-aws ec2 authorize-security-group-ingress \
-    --region $REGION \
-    --group-name talos-aws-tutorial-sg \
-    --protocol tcp \
-    --port 50000-50001 \
-    --cidr 0.0.0.0/0
+    --group-id $SECURITY_GROUP_ID \
+    --ip-permissions \
+        IpProtocol=tcp,FromPort=50000,ToPort=50001,IpRanges="[{CidrIp=0.0.0.0/0}]" \
```
smira, line 199: I'm confused with this rule. We should expose 50000 on the controlplanes to the world, but we should never expose 50001 to the world.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -106,191 +156,196 @@ Once the `SnapshotTaskDetail.Status` indicates `completed`, we can register the
...
+    --default-actions Type=forward,TargetGroupArn=$TARGET_GROUP_ARN \
+    --query 'Listeners[].ListenerArn' \
+    --output text)
 ```
 
 ### Create the Machine Configuration Files
 
-Using the DNS name of the loadbalancer created earlier, generate the base configuration files for the Talos machines.
-> Note that the `port` used here is the externally accessible port configured on the load balancer - 443 - not the internal port of 6443:
+We will create a [machine config patch]({{% ref "/v1.7/talos-guides/configuration/patching.md#rfc6902-json-patches" %}}) to use the AWS time servers.
```
smira, line 260: same about `relref`

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
rothgar: AZ spread is recommended for all AWS HA architecture. Single AZs have problems more often than people realize. All AZs are supposed to be in "single digit" ms latency between locations (Azure and GCP don't guarantee that). I know this is a getting started guide, but we should still recommend some HA configurations for the components that matter.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
rothgar: I realized 6443 should only be exposed on the VPC too. I'll update it.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
smira: So it's not that I'm trying to say that spreading CPs across AZs is a bad idea, but I want to make sure that users understand what gets actually protected and what does not by spreading CPs across AZs.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
rothgar: FWIW the existing guide exposes 6443, 50000, and 50001 to 0.0.0.0/0

smira: 50001 should be never exposed outside of the cluster.

50000 - we recommend only controlplanes
6443 - same only controlplanes (unless you use AWS LB), in that case you don't need to expose it?

smira: https://www.talos.dev/v1.7/learn-more/talos-network-connectivity/

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
rothgar: This version of the doc didn't have the autoscaling group. I just updated it with the correct version. Workers are created in an ASG and if an AZ goes away the ASG will re-balance nodes or create new nodes in AZs that are available.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -7,52 +7,102 @@ aliases:
...
 
-Use a CIDR block that is present on the VPC specified above.
+Create 3 smaller CIDRs to use for each subnet in different availability zones.
 
 ```bash
-aws ec2 create-subnet \
-    --region $REGION \
-    --vpc-id $VPC \
-    --cidr-block ${CIDR_BLOCK}
+IPV4_CIDRS=( $(ipcalc -S 22 --no-decorate $IPV4_CIDR | head -n 3) )
```
elreydetoda, line 46: the `ipcalc` didn't work for me at all on linux (AWS cloudshell), but I just did a space separated list of IP addresses (i.e. `'10.1.0.0/24' '10.1.1.0/24' '10.1.2.0/24'` ). I didn't dig into it much though, so it might just require another package.

image


IDK what kind of CIDRs this generates, but you could offer something like this as a substitution:
```shell
IPV4_CIDRS=( $(printf '10.1.0.0/24\n10.1.1.0/24\n10.1.2.0/24\n') )
```

image

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -106,228 +156,329 @@ Once the `SnapshotTaskDetail.Status` indicates `completed`, we can register the
...
 ```bash
-talosctl --talosconfig talosconfig config endpoint <control plane 1 PUBLIC IP>
-talosctl --talosconfig talosconfig config node <control plane 1 PUBLIC IP>
+talosctl config endpoints $(aws ec2 describe-instances \
+    --instance-ids $(echo $CP_INSTANCES[@]) \
+    --query 'Reservations[].Instances[].PublicIpAddress' \
+    --output text)
+
+talosctl config nodes $(aws ec2 describe-instances \
+    --instance-ids $(echo $CP_INSTANCES[1]) \
```
elreydetoda, line 400: I didn't actually execute anything past here, since I knew how to grab this content, but again with the `{}` for your arrays:
```suggestion
    --instance-ids $(echo ${CP_INSTANCES[1]}) \
```

elreydetoda (commented):
Thanks for linking me to these updated instructions @rothgar, they were awesome!
When walking through them on the AWS cloudshell I ran into a few issues, and I've commented on the PR to let you know what I ran into. Hope this helps :grin:
(P.S. I only looked at the `v1.7` docs)

One of the other things to consider is possible mentioning or integrating (I haven't tested it yet) is the AWS Cloud Controller Manager (CCM) that the talos blog re-posted here: https://www.siderolabs.com/blog/deploying-talos-on-aws-with-cdk/

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
```diff
@@ -7,52 +7,102 @@ aliases:
...
-    --region $REGION \
-    --vpc-id $VPC \
-    --cidr-block ${CIDR_BLOCK}
+IPV4_CIDRS=( $(ipcalc -S 22 --no-decorate $IPV4_CIDR | head -n 3) )
+```
+
+Next create a subnet in each availability zones.
+
+```bash
+CIDR=1
```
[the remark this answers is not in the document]
rothgar: Thanks for calling that out. I ran through an early version of the doc in bash and didn't test it with the final version. I'll figure something out.

website/content/v1.7/talos-guides/install/cloud-platforms/aws.md
rothgar: I think this is a bug with ipcalc packaged with brew. I'll change it to a manual step and make a note that people need to adjust it for their CIDRs

smira (approved):

PR #9099, opened 2024-08-02
fix: fix graph diffs in dashboard when node aliases are used

utkuozdemir:
When `talosctl dashboard` is used with node "aliases" (e.g., node names or machine IDs in Omni) passed via `-n` flag, the graphs in the monitor tab were not rendered correctly: The matching of the old and current data were done incorrectly.

Fix this by doing the comparison over the non-resolved node names, not over the resolved ones.

### DIFF
--- a/internal/pkg/dashboard/apidata/source.go
+++ b/internal/pkg/dashboard/apidata/source.go
@@ -12,12 +12,16 @@ import (
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/types/known/emptypb"
 
+	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
+	"github.com/siderolabs/talos/internal/pkg/dashboard/resolver"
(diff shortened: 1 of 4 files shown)


internal/pkg/dashboard/apidata/data.go
```diff
@@ -18,6 +18,11 @@ type Data struct {
 	// Data per each node.
 	Nodes map[string]*Node
 
+	// UnresolvedNodes contain the same Node data as Nodes
```
smira, line 21: is this some kind of a race that we resolve after the graphs start working?

utkuozdemir: Nope, it wasn't a race, the comparison was done wrongly in a consistent way. So when a node alias (name etc.) was used, the graphs were never updating. it went unnoticed.

Also, I now solved the issue in a different way, so this code is no more there.

smira (approved):

PR #9131, opened 2024-08-08
fix talos version in vmware.sh for 1.7.6

saintdle:
## What? (description)
fix talos version in vmware.sh for 1.7.6

## Why? (reasoning)

using the script from the v1.7 repo/website version, it uploads an OVA of talos 1.1, which won't work


smira:
I think to prevent this in the future and fix it once and for everything, we should better do something like (in the `.md`):

```
curl ..../vmware.sh | sed s/TALOS_VERSION/{{< release >}}
```

so that `vmware.sh` stays the same, but version gets automagically updated

rsmitty (approved):

Issue #8907, opened 2024-06-13
implement kexec support for EFI zBoot kernel


smira:
See https://git.kernel.org/pub/scm/utils/kernel/kexec/kexec-tools.git/tree/kexec/kexec-pe-zboot.c

PR #9162, opened 2024-08-13
chore: support arm64 kexec from zboot kernel images

frezbo:
When using kernel images that are using ZBOOT for arm64 we need to extract the vmlinux from the vmlinuz EFI file and pass it on the the kexec call.

Ref: https://git.kernel.org/pub/scm/utils/kernel/kexec/kexec-tools.git/tree/kexec/kexec-pe-zboot.c

Fixes: #8907

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -15,6 +15,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	goruntime "runtime"
 	"slices"
 	"strconv"
 	"strings"
@@ -64,6 +65,7 @@ import (
 	"github.com/siderolabs/talos/internal/pkg/partition"
 	"github.com/siderolabs/talos/internal/pkg/secureboot"
 	"github.com/siderolabs/talos/internal/pkg/secureboot/tpm2"
+	"github.com/siderolabs/talos/internal/zboot"
 	"github.com/siderolabs/talos/pkg/conditions"
 	"github.com/siderolabs/talos/pkg/images"
 	"github.com/siderolabs/talos/pkg/kernel/kspp"
@@ -2141,6 +2143,24 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
 
 		defer kernel.Close() //nolint:errcheck
 
+		fd := int(kernel.Fd())
+
+		// on arm64 we need to extract the kernel from the zboot image if it's compressed
+		if goruntime.GOARCH == "arm64" {
+			var fileCloser io.Closer
+
+			fd, fileCloser, err = zboot.Extract(kernel)
+			if err != nil {
+				return err
+			}
+
+			defer func() {
+				if fileCloser != nil {
+					fileCloser.Close() //nolint:errcheck
+				}
+			}()
+		}
+
 		initrd, err := os.Open(initrdPath)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -2141,6 +2144,51 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
...
+
+			if _, err := kernel.Read(peZbootheaderData[:]); err != nil {
+				return fmt.Errorf("failed to read PE Zboot header: %w", err)
+			}
+
+			if !bytes.Equal(peZbootheaderData[:4], []byte("MZ\x00\x00")) {
+				return fmt.Errorf("invalid PE Zboot header")
+			}
+
+			if !bytes.Equal(peZbootheaderData[4:8], []byte("zimg")) {
```
frezbo, line 2161: i guess we should skip this if it's not a `zimg`? for upgrades from 1.7

smira: that would be a downgrade from 1.8 to 1.7 I guess, as the host Talos is 1.8, but the kernel is not zboot. I guess in this case we could say just skip to plain `kexec`

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -2177,6 +2225,32 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
...
+	if err != nil {
+		return 0, fmt.Errorf("memfdCreate: %v", err)
+	}
+
+	err = unix.Ftruncate(fd, int64(len(b)))
+	if err != nil {
+		return 0, fmt.Errorf("ftruncate: %v", err)
+	}
+
+	data, err := unix.Mmap(fd, 0, len(b), unix.PROT_READ|unix.PROT_WRITE, unix.MAP_SHARED)
```
smira, line 2239: can't we just "write" to the memfd file here? as it seems we first decompress into memory, then mmap the memfd file, and copy there? so that we could decompress directly to memfd file?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -2141,6 +2144,51 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
 
 		defer kernel.Close() //nolint:errcheck
 
+		fd := int(kernel.Fd())
+
+		if goruntime.GOARCH == "arm64" {
```
smira, line 2149: I would probably extract this to a separate file

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -2141,6 +2144,51 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
...
+			// https://git.kernel.org/pub/scm/utils/kernel/kexec/kexec-tools.git/tree/include/kexec-pe-zboot.h
+			var peZbootheaderData [24]byte
+
+			if _, err := kernel.Read(peZbootheaderData[:]); err != nil {
```
smira, line 2153: we should use `io.ReadFull` (https://pkg.go.dev/io#ReadFull) here, as the read is not guaranteed to fill the whole buffer

internal/zboot/zboot.go
```diff
@@ -0,0 +1,75 @@
...
+
+	fd, err := unix.MemfdCreate("vmlinux", 0)
+	if err != nil {
+		return 0, nil, fmt.Errorf("memfdCreate: %v", err)
+	}
+
+	kernelMemfd := os.NewFile(uintptr(fd), "vmlinux")
+
+	if _, err := io.Copy(kernelMemfd, z); err != nil {
+		return 0, nil, fmt.Errorf("failed to copy zstd data to memfd: %w", err)
```
smira, line 67: we should close the kernelMemFd on the error paths here and below

smira (approved):

PR #9200, opened 2024-08-19
fix: bind HostDNS to 169.254.x link-local address

### LINKED ISSUE #9196: DNS resolution seems broken with `nftables` and `forwardKubeDNSToHost` in development version
### Description
It seems that introduction of `nftables` being used by default in the current `1.8` development branch (tried `v1.8.0-alpha.1-81-ge193e7db9`) - breaks the `forwardKubeDNSToHost`. Here's what I did:
 - bootstrapped a single-node control plane node on AWS, from the latest published image, with pretty much all default parameters (all components are pointing to v1.31 of k8s already)
 - started a simple `alpine:3.20` pod with `hostNetwork: true` solely to run `nslookup`:
 ```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
  namespace: default
spec:
  restartPolicy: Never
  automountServiceAccountToken: false
  hostNetwork: true
  containers:
    - name: test
      image: alpine:3.20
      command: [ "/bin/sh" ]
      args:
        - "-c"
        - "trap : TERM INT; sleep infinity & wait"
  tolerations:
    - key: node-role.kubernetes.io/control-plane
      effect: NoSchedule
```
- attached with shell and tried to lookup an ec2 endpoint within the same region:
```
> nslookup ec2.us-east-2.amazonaws.com
nslookup: write to '10.96.0.9': Operation not permitted
;; connection timed out; no servers could be reached
```
As can be seen - the address of the `host-dns` service is correct (the 9th ip addr), but no resolution happens. Though, network connectivity seems to be working fine, as simply pointing to a different nameserver works, e.g here is an answer from the Cloudflare NS:
```
> nslookup ec2.us-east-2.amazonaws.com 1.1.1.1
Server:		1.1.1.1
Address:	1.1.1.1:53

Non-authoritative answer:
Name:	ec2.us-east-2.amazonaws.com
Address: 99.78.178.238
```

What's interesting here, is that disabling `forwardKubeDNSToHost` brings the DNS resolution into functioning state, but that's doesn't seem like a solution.

So, I've started to play with different things, an accidentally tried to run `kube-proxy` with `iptables` backend instead, as before. I've reprovisioned the control-plane node with a single change in it's config file:
```yaml
cluster:
  proxy:
    extraArgs:
      proxy-mode: iptables
```

After this change - DNS resolution worked:
```
> nslookup ec2.us-east-2.amazonaws.com
Server:		10.96.0.9
Address:	10.96.0.9:53

Non-authoritative answer:
Name:	ec2.us-east-2.amazonaws.com
Address: 99.78.176.220
```
Thus, I suspect that something isn't fully configured with the latest `nftables` rules or maybe it's kube-proxy mangling something.

### Logs
Logs don't contain any suspicious errors or warnings. Cluster boostraps just fine, within 30 seconds, and node is marked ready.

### Environment

- Talos version:
```
Client:
	Tag:         v1.8.0-alpha.1-81-ge193e7db9
	SHA:         e193e7db
	Built:       
	Go version:  go1.22.6
	OS/Arch:     linux/amd64
Server:
	NODE:        talos.flab.dev
	Tag:         v1.8.0-alpha.1-81-ge193e7db9
	SHA:         e193e7db
	Built:       
	Go version:  go1.22.6
	OS/Arch:     linux/arm64
	Enabled:     RBAC
```
- Kubernetes version:
```
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.31.0
```
- Platform: AWS, ARM64, AMI_ID: `ami-0e29f054809ce5025 (talos-v1.8.0-alpha.1-81-ge193e7db9-us-east-2-arm64)`

smira:
This is an attempt to fix many issues related with trying to use Service IP for host DNS.

Fixes #9196


pkg/machinery/constants/constants.go
```diff
@@ -1024,6 +1024,11 @@ const (
 
 	// ConsoleLogErrorSuppressThreshold is the threshold for suppressing console log errors.
 	ConsoleLogErrorSuppressThreshold = 4
+
+	// HostDNSAddress is the address of the host DNS server.
+	//
+	// Note: 116 = 't' and 108 = 'l' in ASCII.
```
utkuozdemir, line 1030: 😎

utkuozdemir (approved):

hack/release.toml
```diff
@@ -77,6 +77,10 @@ machine:
 ```
 
 Please note that on running cluster you will have to kill CoreDNS pods for this change to apply.
+
+The IP address used to forward DNS queries changed to be fixed `169.254.116.108`.
```
frezbo, line 81:
```suggestion
The IP address used to forward DNS queries has been changed to  fixed address `169.254.116.108`.
```

or probably better wording from Tim

smira: thanks, fixed

maxpain:
Does this work with Cilium?

smira:
> Does this work with Cilium?

you can tell us! previous approach worked with Cilium in default config, failed with non-default settings on Cilium side

maxpain:
@smira `169.254.116.108` isn't pingable from the pod's network when BPF masquerading is enabled.

```yaml
bpf:
  masquerade: true
```

maxpain:
I even tried to disable masquerading for link-local addresses using following configuration: 

```yaml
bpf:
  masquerade: true

ipMasqAgent:
  enabled: true
  config:
    masqLinkLocal: false
```

```
root@w1:/home/cilium# cilium-dbg bpf ipmasq list
IP PREFIX/ADDRESS   
169.254.0.0/16   
```

But it didn't help.

smira:
You should ask Cilium guys, it works with Cilium with/without kube-proxy, but there are too many Cilium configuration options.

dhess:
After upgrading my Cilium v1.7.x cluster to v1.8.1, the default DNS configuration is completely broken, presumably due to this issue.

It would be helpful to add a warning to the Talos docs for users running Cilium.

Nomsplease:
@smira can you add the "defaults" you are referring to while testing this deployment of Cilium? Per your docs I have replicated your documented defaults on my Talos 1.8.2 cluster with Cilium 1.16.3, and I get i/o timeouts from the CoreDNS pods. So something was clearly broken when making this change.

These are the helm values the pod was deployed with, they do not work.
`---
cgroup:
  autoMount:
    enabled: false
  hostRoot: /sys/fs/cgroup
cluster:
  id: 1
  name: main
cni:
  exclusive: false
devices: br+
ipam:
  mode: "kubernetes"
ipv4NativeRoutingCIDR: 10.244.0.0/16
k8sServiceHost: 127.0.0.1
k8sServicePort: 7445
kubeProxyReplacement: true
securityContext:
  capabilities:
    ciliumAgent:
      - CHOWN
      - KILL
      - NET_ADMIN
      - NET_RAW
      - IPC_LOCK
      - SYS_ADMIN
      - SYS_RESOURCE
      - DAC_OVERRIDE
      - FOWNER
      - SETGID
      - SETUID
    cleanCiliumState:
      - NET_ADMIN
      - SYS_ADMIN
      - SYS_RESOURCE`

maxpain:
@Nomsplease have you restarted the `coredns` pods after deploying Cilium?

Nomsplease:
> @Nomsplease have you restarted the `coredns` pods after deploying Cilium?

Multiple times, even uninstalled my CoreDNS I had seperate and moved back to the Talos resources. Reapplied everything with k8s upgrade command. This clearly broke with 1.8.X, and I have been pulling my hair out trying to fix it.

maxpain:
@Nomsplease Are you sure you are not using `bpf.masquerade=true` in the cilium helm values?
I faced the same [issue](https://github.com/cilium/cilium/issues/35153) and decided to disable `forwardKubeDNSToHost`. Reusing the DNS cache between the core pods and the host does not offer much benefit.

Nomsplease:
> @Nomsplease Are you sure you are not using `bpf.masquerade=true` in the cilium helm values? I faced the same [issue](https://github.com/cilium/cilium/issues/35153) and decided to disable `forwardKubeDNSToHost`. Reusing the DNS cache between the core pods and the host does not offer much benefit.

I saw your previous issue, hence what led me here. My helm values are above. My previous values included bpf.masquerade, but I was trying to rule out variables hence strip everything to the "defaults". Maybe ill just turn off the DNS to Host since clearly something wasnt tested, or was missed. Cilium guys dont seem to want to look into it.

smira:
> @smira can you add the "defaults" you are referring to while testing this deployment of Cilium? Per your docs I have replicated your documented defaults on my Talos 1.8.2 cluster with Cilium 1.16.3, and I get i/o timeouts from the CoreDNS pods. So something was clearly broken when making this change.

You can see yourself in the CI: https://github.com/siderolabs/talos/actions/runs/11566835784/job/32196143544

rkerno:
I've just stumbled across this issue.  The configuration below works until I switch to bpf.hostLegacyRouting=false.  I'm restoring Talos from backups taken before the CNI is installed to test all Cilium configuration options...it's the only way I can be sure that the configuration is applied correctly without side-effects from a previous configuration.  I'm going to test the cilium ipMasqAgent settings now that @maxpain mentioned above, and will report back with findings.

```
helm template cilium cilium/cilium \
	--version 1.16.3 \
	--namespace kube-system \
	--set securityContext.capabilities.ciliumAgent="{CHOWN,KILL,NET_ADMIN,NET_RAW,IPC_LOCK,SYS_ADMIN,SYS_RESOURCE,DAC_OVERRIDE,FOWNER,SETGID,SETUID}" \
	--set securityContext.capabilities.cleanCiliumState="{NET_ADMIN,SYS_ADMIN,SYS_RESOURCE}" \
	--set cgroup.autoMount.enabled=false \
	--set cgroup.hostRoot=/sys/fs/cgroup \
	--set kubeProxyReplacement=true \
	--set k8sServiceHost=127.0.0.1 \
	--set k8sServicePort=7445 \
	--set ipv4.enabled=true \
	--set ipv4NativeRoutingCIDR="10.244.0.0/16" \
	--set ipam.operator.clusterPoolIPv4PodCIDRList="10.244.0.0/16" \
	--set devices="eth0 eth1" \
	--set routingMode=native \
	--set autoDirectNodeRoutes=true \
	--set bpf.masquerade=true \
	--set bpf.hostLegacyRouting=true \
	--set bpf.datapathMode=netkit \
	--set enableIPv4Masquerade=true \
	--set hubble.enabled=true \
	--set hubble.relay.enabled=true \
	--set hubble.ui.enabled=true \
	> cilium.yaml
```

rkerno:
I think this is related to this cilium issue: https://github.com/cilium/cilium/issues/29413

The recommendation is to move the host's DNS from the loopback device to the dummy device when using bpf routing.  I've had a quick look and didn't see any Talos configuration options to enable achieve this, so I guess if I want bpf routing I need to disable Talos' `forwardKubeDNSToHost`.

kinolaev:
In https://github.com/cilium/cilium/issues/29413 @jbmolle found a solution: setting global scope for a link-local address (in our case 169.254.116.108).
This patch have the same effect on talos:
```yaml
machine:
  network:
    interfaces:
      - interface: lo
        addresses:
          - 169.254.116.108/32
```
The patch allows to use `forwardKubeDNSToHost=true` with cilium's `bpf.masquerade=true`.
To be honest I don't understand the difference between address scopes (host/global). I would appreciate if somebody explains whether such change is safe for production environment or it is better to disable `forwardKubeDNSToHost`

e3b0c442:
It should be noted that this isn't just restricted to bpf masquerading. I had masquerade turned completely off in my Cilium installation since everything is routable in my network and hit this; using bpf routing also triggers it. Might be worth a quick mention in the docs as I had originally overlooked this given I didn't have masquerading enabled at all.

PR #9202, opened 2024-08-19
fix: always handle `PermissionDenied` in dashboard resource watches

utkuozdemir:
A single resource not being there (i.e., the type does not exist on an older version of Talos) or not allowed to be read for whatever reason should not interrupt the refresh cycle of the other resources' status.

### DIFF
--- a/internal/pkg/dashboard/resourcedata/resourcedata.go
+++ b/internal/pkg/dashboard/resourcedata/resourcedata.go
@@ -99,84 +99,47 @@ func (source *Source) runResourceWatchWithRetries(ctx context.Context, node stri
 	}
 }
 
-//nolint:gocyclo,cyclop
+//nolint:gocyclo
 func (source *Source) runResourceWatch(ctx context.Context, node string) error {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
 	eventCh := make(chan state.Event)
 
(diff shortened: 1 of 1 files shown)


internal/pkg/dashboard/resourcedata/resourcedata.go
```diff
@@ -99,84 +100,53 @@ func (source *Source) runResourceWatchWithRetries(ctx context.Context, node stri
...
-	if err := source.COSI.WatchKind(ctx, network.NewRouteStatus(network.NamespaceName, "").Metadata(), eventCh, state.WithBootstrapContents(true)); err != nil {
-		return err
-	}
-
-	if err := source.COSI.WatchKind(ctx, network.NewLinkStatus(network.NamespaceName, "").Metadata(), eventCh, state.WithBootstrapContents(true)); err != nil {
-		return err
-	}
+	defer func() {
+		if tarik != nil {
+			log.Printf("TARIK: %v", tarik)
```
frezbo, line 110: what's TARIK? :thinking:

frezbo (approved):

internal/pkg/dashboard/resourcedata/resourcedata.go
utkuozdemir: oops, debug code :D it's something like foobar basically :)

smira (approved):

PR #9213, opened 2024-08-22
fix: drop /opt mount for containers/tink

smira:
The `/opt/cni/bin` in the rootfs contains CNI binaries, which get overwritten by the volume mount.


frezbo (approved):

frezbo:
i guess docs needs updating?

smira:
> i guess docs needs updating?

yes, I updated them

PR #9218, opened 2024-08-23
chore: add java package to protos

eddiezane:
# Pull Request

## What? (description)

This PR adds `java_package` options to all protos.

## Why? (reasoning)

I am generating a Java/Kotlin library to interact with clusters from an Android app. Java packages, much like Go, must match their folder path on disk. You cannot provide `java_package` as a build time option to the Java protoc compiler. So for the package to be defined as anything other than `cluster`, the option must be defined at the proto level. I set it to what I thought was reasonable but totally open to changing it.

## Acceptance


api/resource/definitions/cluster/cluster.proto
```diff
@@ -3,6 +3,7 @@ syntax = "proto3";
 package talos.resource.definitions.cluster;
 
 option go_package = "github.com/siderolabs/talos/pkg/machinery/api/resource/definitions/cluster";
+option java_package = "dev.talos.api.resource.definitions.cluster";
```
smira, line 6: the files under `api/resource/definitions` are automatically generated (you can check that by running `make generate`), so you'd need to update the generator (should be hack/structprotogen iirc)

eddiezane: Added and generated!

smira (approved):
thank you!

Issue #8695, opened 2024-05-02
Probelm with HP dl360 g7,g8,g9


cybercoder:
## Bug Report
Talos metal amd64 iso, keyboard freezes after boot (No F2, F3), and the DHCP client doesn't work.

### Description
We have mounted the ```metal-amd-64``` iso file, both 1.6,1.7, and also 1.7.1 on 3 HP servers using iLO.
But we have no keyboard and also Nics won't get IP address.
### Environment
- Platform: HP proliant dl360

smira:
Not exactly clear, but you can try removing `console=` args before the boot.

cybercoder:
> Not exactly clear, but you can try removing `console=` args before the boot.

had no success. Also, I've configured:
```
console=ttyS0,115200n8 console=tty0
```
And doesn't work. 
Talos 1.5.3 will get IP via DHCP but Talos 1.7.0 not.

cybercoder:
I have tried to upgrade from 1.5.3 to 1.7.5 too. ( 1.5.3 get IP address from DHCP but 1.6 and upper not )
After upgrade there's no IP address:

![Screen Shot 1403-04-14 at 17 50 13](https://github.com/siderolabs/talos/assets/437872/b2c8a84e-1f43-4328-80b4-963a3bf372d1)
It seems it doesn't detect the NIC.

Issue #8657, opened 2024-04-25
Stuck at efi stub: loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path


muvaf:
### Description

I have tried two methods to install Talos OS to a bare metal instance in Hetzner (not Hetzner Cloud):
- Log into rescue mode with SSH and write content of "metal-amd64.raw.xz" to the boot disk /dev/nvme0n1 (tried writing to other disk as well /dev/nvme1n1
- Have Hetzner technician write the metal-amd64.iso file to a USB and plug it in, then choose the USB from boot menu.

In both cases, when it boots, it's stuck at black screen with the following message:
`EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path`

All talosctl commands time out.

### Logs
```
EFI stub: Loaded initrd from LINUX_EFI_INITRD_MEDIA_GUID device path
```

One thing I noticed, when I go to rescue mode and open the disk I `dd`'ed Talos raw image, I see the following:
```
GPT PMBR size mismatch (181931 != 1000215215) will be corrected by write.
The backup GPT table is not on the end of the device. This problem will be corrected by write.
The device contains 'iso9660' signature and it will be removed by a write command. See fdisk(8) man page and --wipe option for more details.
```

### Environment

- Talos version: v1.7.0
- Platform: Hetzner Robot dedicated root server with Intel i5 13500 ([product page](https://www.hetzner.com/dedicated-rootserver/ex44/))

smira:
You can try removing all `console=` args at the GRUB boot prompt by editing the command line.

muvaf:
@smira I don't have access to GRUB boot prompt unless I get Hetzner to attach a KVM device to the server. In my second try, having just `console=tty0` worked as well. I'd like to update [this doc](https://www.talos.dev/v1.7/talos-guides/install/cloud-platforms/hetzner/#rescue-mode) with this information but the doc is tailored for Hetzner Cloud which is completely separate from Hetzner Robot dedicated server service where there isn't even a snapshotting machinery. You just write the `metal-amd64.iso` to the whole disk.

aarnaud:
Same issue with `agent-amd64`  booting on a qemu/kvm VM over PXE with UEFI with sidero 0.6.4.
![Screenshot from 2024-05-06 11-02-08](https://github.com/siderolabs/talos/assets/4802105/acff626f-5b35-41b0-ade3-b279725fe6cb)

[...]

Issue #8127, opened 2024-01-02
No video output on metal after grub boot


imle:
## Bug Report
Stuck in boot after `kern/efi/sb.c:111: UEFI Secure Boot state: Disabled` is printed.

### Description
Trying to boot into 1.6.1 or 1.4.4 version of Talos on a `HP Elitedesk 705 G4 DM`. Without `set debug=all`, the machine hangs with a black screen and an underscore. After enabling the debug option, the boot process hangs immediately after printing the above log line.

I have attempted the boot process with all drives removed and a single stick of ram to eliminate as many variables as possible and this happens with all 3 of the machines I have.

### Logs
![PXL_20240102_153129451.jpg](https://github.com/siderolabs/talos/assets/4809109/946c5171-c264-4ff1-ade5-e664d38e6662)

imle:
Only mention I can find of a similar issue anywhere:
https://forums.gentoo.org/viewtopic-p-8755201.html?sid=7b3162f6c0ec892d909661710c21fb5f

imle:
Also want to note that trying to set `set debug=all,-mmap` does not seem to be working.

https://www.gnu.org/software/grub/manual/grub/html_node/debug.html

imle:
Ok, upon further work, it does seem that it is booting, but just nothing is coming out of either display port.

smira:
Usually it's related to the `console=` settings, you can try tweaking it, removing `console=ttyS0`, changing to `console=tty1`.

Unfortunately it's hardware dependent.

runningman84:
I have the exact same issue with my n100 board from asrock. I can boot any other Linux distribution like Ubuntu just fine.

runningman84:
![image](https://github.com/siderolabs/talos/assets/1699128/bdaf507f-55a5-4a15-81bb-40c162085da2)

with talos 1.7.0 beta 0 that is the output.

smira:
Try removing `console=ttyS0` boot argument.

runningman84:
Removing all console= parameters it boots just fine with version 1.6.7 and 1.7.0.beta.0.  Are these settings persistent after the install? Why do all linux distros work out of the box but talos has issues like this?

runningman84:
These settings get lost after the initial install so I have to modify the grub parameters for each boot. How can I persist that in talos?

smira:
[...]

PR #9222, opened 2024-08-25
chore: drop console `ttyS0` argument.

frezbo:
Drop `console=ttyS0` argument for metal images/installer.

`console=ttyS0` causes lot of issues with bare metal hardware when trying to use a physical serial port.

Ref:

* https://bugzilla.redhat.com/show_bug.cgi?id=1839923
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=763601;msg=17
* https://www.kernel.org/doc/html/latest/admin-guide/serial-console.html
* https://github.com/coreos/fedora-coreos-tracker/issues/567

Fixes: #8695
Fixes: #8657
Fixes: #8127

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"slices"
 	"syscall"
 	"time"
 
@@ -146,6 +147,15 @@ func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options) 
 		opts.OverlayName = overlayOpts.Name
 
 		cmdline.SetAll(overlayOpts.KernelArgs)
+	}
+
+	// preserve console=ttyS0 if it was already present in cmdline for metal platform
+	existingCmdline := procfs.ProcCmdline()
+
+	if *existingCmdline.Get(constants.KernelParamPlatform).First() == constants.PlatformMetal && existingCmdline.Get("console").Contains("ttyS0") {
+		if !slices.Contains(opts.ExtraKernelArgs, "console=ttyS0") {
+			cmdline.Append("console", "ttyS0")
+		}
 	}
 
 	if err := cmdline.AppendAll(
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -162,7 +162,7 @@ func (m *Metal) KernelArgs(arch string) procfs.Parameters {
 	switch arch {
 	case "amd64":
 		return []*procfs.Parameter{
-			procfs.NewParameter("console").Append("ttyS0").Append("tty0"),
+			procfs.NewParameter("console").Append("tty0"),
 		}
 	case "arm64":
 		return []*procfs.Parameter{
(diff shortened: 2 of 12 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
```diff
@@ -162,7 +162,7 @@ func (m *Metal) KernelArgs(arch string) procfs.Parameters {
 	switch arch {
 	case "amd64":
 		return []*procfs.Parameter{
-			procfs.NewParameter("console").Append("ttyS0").Append("tty0"),
```
smira, line 165: we need to make it dependent on Talos version (`quirks`) - we want to make images for Talos < 1.8 still append `ttyS0`.

smira: we also need to figure out compatibility on upgrades - if the installed Talos runs with `console=ttyS0`, we should preserve that.

internal/pkg/install/install.go
```diff
@@ -190,6 +191,13 @@ func RunInstallerContainer(disk, platform, ref string, cfg configcore.Config, cf
 		}
 	}
 
+	if !slices.Contains(options.ExtraKernelArgs, "console=ttyS0") {
+		// preserve console=ttyS0 if it was already present in cmdline
+		if procfs.ProcCmdline().Get("console").Contains("ttyS0") {
+			args = append(args, "--extra-kernel-arg", "console=ttyS0")
```
smira, line 197: I think we should better move this into the installer itself, this code runs on the host

pkg/machinery/imager/quirks/quirks.go
```diff
@@ -98,3 +98,16 @@ func (q Quirks) SupportsMultidoc() bool {
...
 }
+
+// maxVersionPhyscicalSerialConsoleDropped is the version that dropped console=ttyS0 for metal image.
+var maxVersionPhyscicalSerialConsoleDropped = semver.MustParse("1.8.0")
+
+// SupportsPhysicalSerialConsole returns true if the Talos version supports already has console=ttyS0 kernel argument.
+func (q Quirks) SupportsPhysicalSerialConsole() bool {
+	// if the version doesn't parse, we assume it's latest Talos
+	if q.v == nil {
+		return true
```
smira, line 109: should be `false` ?

cmd/installer/pkg/install/install.go
```diff
@@ -148,6 +148,13 @@ func Install(ctx context.Context, p runtime.Platform, mode Mode, opts *Options)
 		cmdline.SetAll(overlayOpts.KernelArgs)
 	}
 
+	// preserve console=ttyS0 if it was already present in cmdline for metal platform
+	existingCmdline := procfs.ProcCmdline()
+
+	if *existingCmdline.Get(constants.KernelParamPlatform).First() == constants.PlatformMetal && existingCmdline.Get("console").Contains("ttyS0") {
+		cmdline.Append("console", "ttyS0")
```
frezbo, line 155: if `extraKernelArgs` already has `ttyS0` this would add a duplicate, should I check there also?

smira: yeah, let's not do duplicates

smira (approved):

PR #8901, opened 2024-06-12
feat: support volume configuration, provisioning, etc.

smira:
This implements the first round of changes, replacing the volume backend
with the new implementation, while keeping most of the external
interfaces intact.

See #8367

PR #9233, opened 2024-08-27
fix: host DNS access with firewall enabled

smira:
Explicitly enable access to host DNS from pod/service IPs.

Also fix the Kubernetes health checks to assert number of ready pods to match expectation, otherwise the check might skip a pod (e.g. `kube-proxy` one) which is not ready, allowing the test to proceed too early.

Update DNS test to print more logs on error.


smira:
Split from #8901

frezbo (approved):

PR #9239, opened 2024-08-28
chore: aarch64 qemu local secureboot support

frezbo:
Support booting with SecureBoot on aarch64 with `talosctl cluster create` with QEMU provisioner.

### DIFF
--- a/pkg/provision/providers/qemu/launch.go
+++ b/pkg/provision/providers/qemu/launch.go
@@ -43,14 +43,11 @@ type LaunchConfig struct {
 	DiskDrivers       []string
 	VCPUCount         int64
 	MemSize           int64
-	QemuExecutable    string
-	Architecture      string
 	KernelImagePath   string
 	InitrdPath        string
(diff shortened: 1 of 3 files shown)


pkg/provision/providers/qemu/launch.go
```diff
@@ -439,13 +439,19 @@ func launchVM(config *LaunchConfig) error {
 			return err
 		}
 
+		tpmDevice := "tpm-tis,tpmdev=tpm0"
+
+		if Arch(config.Architecture) == ArchArm64 {
```
smira, line 444: nit: we can make this a method on `Arch`: 

```go
func (a Arch) TMPDeviceArg() string
```

smira (approved):

pkg/provision/providers/qemu/launch.go
frezbo: I restructured more code to be defined on `Arch`, so we pass around less stuff

PR #9246, opened 2024-08-29
docs: kubeProxyReplacement from "disabled" to "false"

a-mader:
# Pull Request

## What? 
Fix incorrect helm install command for cilium

## Why? 
User have a better experience installing cilium

## Acceptance


smira (approved):
thank you

frezbo (commented):
note to self, check if we are using same values in our ci, i think I updated there, but not docs

frezbo (approved):

frezbo:
> note to self, check if we are using same values in our ci, i think I updated there, but not docs

This is already done

PR #9247, opened 2024-08-29
chore: bump dependencies

smira:
Update tools, pkgs, extras, Go dependencies, Go tools, etc.

Linux 6.6.47 and containerd 2.0.0-rc.4.


Unix4ever (approved):

go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	k8s.io/apiserver v0.31.0
 	k8s.io/client-go v0.31.0
 	k8s.io/component-base v0.31.0
-	k8s.io/cri-api v0.31.0
+	k8s.io/cri-api v0.32.0-alpha.0
```
frezbo, line 36: was this intentional?

internal/app/machined/pkg/system/system.go
```diff
@@ -193,7 +193,7 @@ func (s *singleton) Start(serviceIDs ...string) error {
 					msg = strings.ToUpper(msg[:1]) + msg[1:]
 				}
 
-				svcrunner.UpdateState(context.Background(), events.StateFailed, msg)
+				svcrunner.UpdateState(context.Background(), events.StateFailed, "%s", msg)
```
frezbo, line 196: what's the formatter doing here?

frezbo (approved):

go.mod
smira: no

internal/app/machined/pkg/system/system.go
smira: it is the linter which found an issue - `UpdateState` is `printf`-like, so passing `msg` as the first argument is wrong - if it contains `%` instructions, it might lead to invalid result

so we pass `fmt.Printf("%s", v)`

go.mod
smira: but it comes from containerd v2.0.0, so I guess it's intentional :)

PR #9267, opened 2024-09-03
chore: remove matrix links from docs

rsmitty:
This PR removes the matrix links since those rooms are no longer in use.

### DIFF
--- a/website/content/v0.10/_index.md
+++ b/website/content/v0.10/_index.md
@@ -21,9 +21,6 @@ If you are just getting familiar with Talos, we recommend starting here:
 
 - GitHub: [repo](https://github.com/talos-systems/talos)
 - Slack: Join our [slack channel](https://slack.dev.talos-systems.io)
-- Matrix: Join our Matrix channels:
-  - Community: [#talos:matrix.org](https://matrix.to/#/#talos:matrix.org)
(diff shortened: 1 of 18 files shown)


smira (approved):

frezbo (approved):

davralin:
Why was these rooms removed? I used those to access the chat, so they were obviously in use by at least one... :/

smira:
It was announced in the rooms as well:

> :rotating_light: Posting this to each channel that has a matrix bridge:
> Hey folks, heads up that our team has decided we'll be getting rid of the Matrix bridge into our slack. We've only got so much time and focus, and it's certainly better spent making Talos + Omni better and not community moderation. The Matrix rooms themselves will also no longer be monitored by our team. If someone would like to volunteer to become admins to these rooms, feel free to ping me. If no one is interested, I'll also be deleting the Matrix rooms completely next Friday (8/30). This will involve booting everyone from these rooms and then leaving the room myself, triggering a 7 day wait until full deletion.
Other ways to get in touch with us are through our community slack (https://slack.dev.talos-systems.io/), as well as through Github Discussions (https://github.com/siderolabs/talos/discussions + https://github.com/siderolabs/omni/discussions)

PR #9270, opened 2024-09-04
chore: support gcp in cloud-image-uploader

frezbo:
Add support for uploading images to GCP in cloud image uploader.

GCP is not enabled by default since it's going to be used for e2e-tests for now.


smira (approved):
:cool:

PR #9275, opened 2024-09-04
feat: implement "$patch: delete" logic

DmitriyMV:
This PR implements "delete patches", same as in k8s.


smira (approved):
looks very :cool: 

we need to add docs:

* to the config patching document to explain `$patch: delete`
* and add to `hack/release.toml` a note about new feature

smira (approved):

PR #9313, opened 2024-09-12
fix: ignore invalid NTP responses

smira:
Due to the bug introduced when refactoring for PTP devices, invalid NTP responses (including for example NTP kiss of death), were incorrectly handled when only a single NTP server was used.

The error was logged, but the response was used to adjust the time which leads to unexpected time jumps.

Properly ignore any invalid NTP response.

Without this fix, the adjustment from the invalid response is applied.

### DIFF
--- a/internal/pkg/ntp/ntp.go
+++ b/internal/pkg/ntp/ntp.go
@@ -392,18 +392,15 @@ func (syncer *Syncer) queryNTP(server string) (*Measurement, error) {
 	)
 
 	validationError := resp.Validate()
+	if validationError != nil {
+		return nil, validationError
+	}
 
-	measurement := &Measurement{
+	return &Measurement{
 		ClockOffset: resp.ClockOffset,
 		Leap:        resp.Leap,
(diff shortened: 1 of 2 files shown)


internal/pkg/ntp/ntp.go
```diff
@@ -392,18 +392,15 @@ func (syncer *Syncer) queryNTP(server string) (*Measurement, error) {
...
-		Spike:       false,
-	}
-
-	if validationError == nil {
-		measurement.Spike = syncer.isSpike(resp)
-	}
-
-	return measurement, validationError
+		Spike:       syncer.isSpike(resp),
+	}, validationError
```
frezbo, line 403: i guess return nil here, since `validationError` would be nil in this case?

frezbo (approved):

internal/pkg/ntp/ntp.go
smira: yes, I can make it more clear with `nil`, thanks

PR #9315, opened 2024-09-12
test: add a test for inline machine config trusted roots

smira:
Run SideroLink API server via TLS with self-signed certificate, inject that certificate into Talos via `talos.config.inline=`.

Fix a couple of place where our special TLS root CA provider supporting reloading on the fly was not used.


Unix4ever (approved):

frezbo (approved):
:cool: 

Nice that the tests caught missing stuff

PR #9329, opened 2024-09-16
fix: strategic merge patch delete for map keys

### LINKED ISSUE #9325: Can't seem to remove `node.kubernetes.io/exclude-from-external-load-balancers` from `machine.nodeLabels`
## Bug Report

In spite of the following patch contents:

```
machine:
  nodeLabels:
    node.kubernetes.io/exclude-from-external-load-balancers:
      $patch: delete
```

apparently being cleanly applied, the node label is not removed from the config and from the Kubernetes node.

### Logs

None

### Environment

- Talos version: v1.8.0-beta.0, terraform provider 0.6.0-beta.0
- Kubernetes version: v1.31.0
- Platform: metal

smira:
When a map key is deleted, it should be deleted as a whole. Before the fix it was zeroing out map value by key.

Fixes #9325

### DIFF
--- a/pkg/machinery/config/configloader/internal/decoder/selector.go
+++ b/pkg/machinery/config/configloader/internal/decoder/selector.go
@@ -187,7 +187,7 @@ func deleteForPath(val reflect.Value, path []string, key, value string) error {
 
 			if idx := val.MapIndex(searchForVal); idx.IsValid() {
 				if len(path) == 0 {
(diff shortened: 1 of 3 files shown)


frezbo (approved):

DmitriyMV:
> If elem is the zero Value, SetMapIndex deletes the key from the map.

I missed that in docs :|

DmitriyMV:
👍

DmitriyMV (approved):

Issue #9300, opened 2024-09-11
Boot failed on hosts with LVM thin/raid logical volumes


eshikhov:
### Discussed in https://github.com/siderolabs/talos/discussions/9140

<div type='discussions-op-text'>

<sup>Originally posted by **eshikhov** August  9, 2024</sup>
Hi,

When using the lvm mirror and lvm thin pool functionality, the OS is loaded randomly. 

When setting up for the first time, everything is fine, the kernel modules (dm_raid and dm_thin_pool) are loaded with the config. Then the lvm mirror and lvm thin pool volumes are created and everything works well until the server is rebooted (and the reboot can be caused by various reasons, be it talosctl upgrade with the --reboot-mode powercycle option or just talosctl reboot --mode powercycle). After a restart, talos does not start, but constantly goes into cyclic restarts and at some time of restart it can boot successfully.

Screenshots of unsuccessful launches that I managed to take:

![screenshot_1](https://github.com/user-attachments/assets/d3e049d1-f81e-40ad-9a34-54138d76f6d4)

![screenshot_2](https://github.com/user-attachments/assets/24fb60eb-63b2-429d-8882-b6a6e929f80a)

![screenshot_3](https://github.com/user-attachments/assets/4fa99fec-5ed5-4e73-aa0d-495b0c9a89f0)

![screenshot_4](https://github.com/user-attachments/assets/72adf05c-49ff-4473-8c70-ae86289d5698)

![screenshot_5](https://github.com/user-attachments/assets/64c96f52-72cd-41b7-9b16-8a46ee64678a)

Dump debug information about the cluster:

[support.zip](https://github.com/user-attachments/files/16562249/support.zip)

Talos v1.7.6</div>

vladimirfx:
We reproduced the issue on VM. Logs and VMs can be found here: https://drive.google.com/drive/folders/1aj4GUOjO_EqSFdz4GgAZXhpCJ75ggZOm

In production, we are in a situation where we cannot reboot or update servers that have LV with thin or raid1 type. After reboot, the server can't boot from 30 minutes to several hours.

Of course, all servers have the correct modules configured and LVs created in the past year (using LVM CSI).

It looks like some race in kernel module initialization.

As a workaround, can you provide a machine config option to not activate LVM at boot time? (because all known LVM CSI implementations are active LVs themselves)

vladimirfx:
@mst1711 FYI

smira:
[...]

PR #9333, opened 2024-09-18
feat: support lvm auto activation

frezbo:
Support lvm auto-activation as per
https://man7.org/linux/man-pages/man7/lvmautoactivation.7.html.

This changes from how Talos previously used to unconditionally tried to activate all volume groups to based on udev events.

Fixes: #9300

### DIFF
--- a/internal/app/machined/pkg/controllers/block/lvm.go
+++ b/internal/app/machined/pkg/controllers/block/lvm.go
@@ -1,0 +1,176 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package block
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/hashicorp/go-multierror"
+	"github.com/siderolabs/gen/optional"
+	"github.com/siderolabs/go-cmd/pkg/cmd"
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+	"github.com/siderolabs/talos/pkg/machinery/resources/block"
+	runtimeres "github.com/siderolabs/talos/pkg/machinery/resources/runtime"
+	"github.com/siderolabs/talos/pkg/machinery/resources/v1alpha1"
+)
+
+// LVMActivationController activates LVM volumes when they are discovered by the block.DiscoveryController.
+type LVMActivationController struct {
+	seenVolumes  map[string]struct{}
+	activatedVGs map[string]struct{}
+}
+
+// Name implements controller.Controller interface.
+func (ctrl *LVMActivationController) Name() string {
+	return "block.LVMActivationController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *LVMActivationController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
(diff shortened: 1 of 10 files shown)


internal/app/machined/pkg/controllers/block/lvm.go
```diff
@@ -0,0 +1,132 @@
...
+					// wait for the mount status to be available
+					continue
+				}
+
+				return fmt.Errorf("failed to get mount status: %w", err)
+			}
+		}
+
+		discoveredVolumes, err := safe.ReaderListAll[*block.DiscoveredVolume](ctx, r)
+		if err != nil && !state.IsNotFoundError(err) {
```
smira, line 77: `List` can't return `NotFound` error

internal/app/machined/pkg/controllers/block/lvm.go
```diff
@@ -0,0 +1,132 @@
...
+//nolint:gocyclo
+func (ctrl *LVMActivationController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	touchedIDs := make(map[string]struct{})
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+			if _, err := safe.ReaderGetByID[*runtimeres.MountStatus](ctx, r, constants.EphemeralPartitionLabel); err != nil {
```
smira, line 66: nit: we can move it out of `select`

internal/app/machined/pkg/controllers/block/lvm.go
```diff
@@ -0,0 +1,132 @@
...
+				"vgchange",
+				"-aay",
+				"--autoactivation",
+				"event",
+				vgName,
+			); err != nil {
+				return fmt.Errorf("failed to activate LVM volume %s: %w", vgName, err)
+			}
+
+			touchedIDs[vgName] = struct{}{}
```
smira, line 129: we could move `touchedIDs` out to the `type LVMActivationController struct{}`, so that it survives controller restarts

internal/app/machined/pkg/controllers/block/lvm.go
```diff
@@ -0,0 +1,132 @@
...
+				"--listvg",
+				"--checkcomplete",
+				"--vgonline",
+				"--autoactivation",
+				"event",
+				"--udevoutput",
+				iterator.Value().TypedSpec().DevPath,
+			)
+			if err != nil {
+				return fmt.Errorf("failed to check LVM volume availability: %w", err)
```
smira, line 102: should we log the error and continue looking for other volumes here?

smira: or grab all errors and report the error in the end, as this will block all LVMs if one of them fails

smira (commented):
looks :cool:

internal/app/machined/pkg/controllers/block/lvm.go
frezbo: the udev rule was always getting triggered, but the command output is empty and it skips that way, i wonder if we need `touchedIDs` at all :thinking:

smira: I think it had something like "first time seen" in the list of steps as well.

internal/app/machined/pkg/controllers/block/lvm.go
```diff
@@ -0,0 +1,172 @@
...
+				"vgchange",
+				"-aay",
+				"--autoactivation",
+				"event",
+				vgName,
+			); err != nil {
+				return fmt.Errorf("failed to activate LVM volume %s: %w", vgName, err)
+			}
+
+			ctrl.activatedVGs[vgName] = struct{}{}
```
smira, line 134: it looks like we never use this value, do we need to skip activation if already activated?

smira (approved):

PR #9339, opened 2024-09-18
Allow /sbin/ldconfig (support for glibc extension for nvidia gpu-operator)

jfroy:
# Pull Request

## What? (description)

This PR allows `/sbin/ldconfig` in extensions.

## Why? (reasoning)

Allowing `/sbin/ldconfig` in extensions is a simple change that should have basically no bad side effects. It allows the NVIDIA container toolkit to work on Talos without modification.

## Original "Why?" that allowed containers with SYS_MODULE

The NVIDIA gpu-operator includes machinery to manage the driver on nodes via driver containers. These containers, usually provided by NVIDIA, include basically a giant shell script that will fetch, build, and load the kernel modules for the node's kernel. It will also unload these modules when the container terminates, and provide the userspace components by copying them to `/run/nvidia/driver` on the node.

More recently, driver containers have gained the ability to provide [precompiled kernel modules](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/precompiled-drivers.html) on supported Linux distributions (Ubuntu today).

The intention here is for SideroLabs to build their own driver containers and let the gpu-operator manage them. Is it not a clear-cut better approach than the existing extensions, but it better aligns with NVIDIA's general direction for supporting GPUs in Kubernetes and provides more flexibility for cluster operators by decoupling NVIDIA driver releases from Talos releases (even with the recent addition of LTS and production extension flavors).

Of course, producing driver containers will either need to be coupled with kernel builds to use the ephemeral module signing key, or SideroLabs may change its approach and e.g. retain the signing key for each Talos release to decouple NVIDIA driver builds from kernel builds. Another approach would be to use `SYSTEM_EXTRA_CERTIFICATE` (see https://github.com/torvalds/linux/commit/c4c36105958576fee87d2c75f4b69b6e5bbde772) to allow Image Factory to insert a signing certificate provided by the user, which could then build their own driver containers.

## Acceptance

### DIFF
--- a/pkg/machinery/extensions/extensions.go
+++ b/pkg/machinery/extensions/extensions.go
@@ -12,7 +12,10 @@ var AllowedPaths = []string{
 	"/etc/cri/conf.d",
 	"/lib/firmware",
 	"/lib/modules",
+	// The glibc loader is required by glibc dynamic binaries.
 	"/lib64/ld-linux-x86-64.so.2",
+	// /sbin/ldconfig is required by the nvidia container toolkit.
+	"/sbin/ldconfig",
 	"/usr/etc/udev/rules.d",
 	"/usr/local",
 	// glvnd, egl and vulkan are needed for OpenGL/Vulkan.


frezbo:
adding `/sbin/ldconfig` is all right

But we don't want to change `sys_module` capability, does the gpu operator has support for ignoring driver loading and let talos still load the drivers? or the drivers could still be provided as container images (pre-compiled kernel modules), but talos loads it to a place where the operator expects it? Would be a much better than reducing default security

jfroy:
> adding `/sbin/ldconfig` is all right
> 
> But we don't want to change `sys_module` capability, does the gpu operator has support for ignoring driver loading and let talos still load the drivers? or the drivers could still be provided as container images (pre-compiled kernel modules), but talos loads it to a place where the operator expects it? Would be a much better than reducing default security

I think both of those options are possible with caveats. Driver containers really are glorified shell scripts. You can see my prototype here: https://github.com/jfroy/gpu-driver-container/tree/main/talos. So presumably a driver container for Talos could simply check for the nvidia modules and not support loading or unloading.

The caveat is that depending on the user configuration, driver containers can load the modules with parameters. That would break / not work on Talos if driver containers do not manage the kernel modules.

But maybe that is an acceptable first step. One possible second step from there would be to add a machined API to load/unload kernel modules, and then Talos driver containers could be granted the rights to invoke that API.

frezbo:
> But maybe that is an acceptable first step. One possible second step from there would be to add a machined API to load/unload kernel modules, and then Talos driver containers could be granted the rights to invoke that API.

so talos already support loading modules with kernel parameters, with the WIP SELinux work, maybe we could at some point somehow label the driver containers so they could only load modules, we can start with the `ldconfig` change only at first, and see how we proceed

jfroy:
> > But maybe that is an acceptable first step. One possible second step from there would be to add a machined API to load/unload kernel modules, and then Talos driver containers could be granted the rights to invoke that API.
> 
> so talos already support loading modules with kernel parameters, with the WIP SELinux work, maybe we could at some point somehow label the driver containers so they could only load modules, we can start with the `ldconfig` change only at first, and see how we proceed

Let me test/prototype this first and I'll update the PRs.

andrewrynhard:
@jfroy since this would be a pretty big change for us philosophically I would it if we could hop on a call to chat through this some more. Would you be open to that?

jfroy:
> @jfroy since this would be a pretty big change for us philosophically I would it if we could hop on a call to chat through this some more. Would you be open to that?

Sure, reach out by email.

frezbo (approved):

jfroy:
As a note (which I guess you probably know), the validator needs a module update to import the new package from talos.

frezbo:
> As a note (which I guess you probably know), the validator needs a module update to import the new package from talos.

yeh, next PR should be to validator

PR #9343, opened 2024-09-20
chore: refactor tests

frezbo:
Refactor tests to avoid code duplication.


smira:
should we promote it to some pipelines to verify?

smira (approved):

frezbo:
> should we promote it to some pipelines to verify?

shouldn't need to, anyways I re-triggered just to be safe

PR #9346, opened 2024-09-20
chore: better lvm2 tests

frezbo:
Use LVM2 tests that relies on module loading by lvm.

Fixes: #9300


smira:
why do we have gitbook here?

frezbo:
> why do we have gitbook here?

I've no idea

smira:
we should pull back in [`ec32f44` (#9348)](https://github.com/siderolabs/talos/pull/9348/commits/ec32f44c31ef2f3750a62b3fc39ac0a7ae01bde5)

smira (approved):

PR #9362, opened 2024-09-23
docs: Fix Kernel Self Protection Project (KSPP) references

danrue:
Update KSPP related URLs and fix acronym typo from "KSSP" to "KSPP".


smira (approved):
thank you!

Issue #9412, opened 2024-10-01
Talos leaks file descriptors to child processes


smira:
tl;dr is missing `O_CLOEXEC` in some libraries we're using
```[tasklist]
### Tasks
- [ ] https://github.com/insomniacslk/dhcp/pull/550
- [ ] https://github.com/safchain/ethtool/pull/88
```

smira:
> Should we maybe consider adding a rule to golangci-lint to check for syscalls without cloexec to avoid problematic code being merged later?

I would rather prefer a test, but not sure how to put #9414 into a useful thing, as we need to fork directly from machined to verify this. System extensions don't work, I don't want to put test binary into any Talos build. I could use some debug feature as well to enable this test controller

dsseng:
Will take a look a bit later, maybe I have some ideas about this

smira:
Update: the problem itself is fixed, but the issue is left open to track upstream PRs.

smira:
https://github.com/vishvananda/netlink/pull/1023 can go away if #9566 is merged

PR #9413, opened 2024-10-01
fix: prevent file descriptors leaks to child processes

smira:
See #9412

I'll keep the issue open to track upstream PR status and remove replace directives.


smira:
I used #9414 to verify fixes

go.mod
```diff
@@ -26,6 +26,18 @@ replace (
 	gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20220527175918-f17b0f05cf2c
 )
 
+// fd-leak related replacements: https://github.com/siderolabs/talos/issues/9412
```
dsseng, line 29: Maybe leave a TODO/FIXME to not forget to replace these versions with upstream ones once the fixes are upstreamed. Maybe we can just leave the issue open instead

dsseng (approved):

go.mod
smira: that's what I'm doing: 

> I'll keep the issue open to track upstream PR status and remove replace directives.

Unix4ever (approved):

dsseng (approved):
reviewed after force-push

PR #9414, opened 2024-10-01
chore: test code to verify fd leaks

smira:
See #9413

Not sure how to turn this into a useful test.


dsseng:
One idea would be to build such a thing only in case of building for tests, so adding a special build tag for test

smira:
We already have `sidero.debug` build tag which we can use. There should be also some more stuff, like how to expose the status of the test (I did very dumb version which did printing to the console).

PR #9430, opened 2024-10-03
doc: Update Nvidia docs with Proprietary/OSS hardware reqs

SlothCroissant:
# Pull Request

## What? (description)

Added details about the difference in hardware support between Proprietary and OSS drivers for Nvidia. 

## Why? (reasoning)

I tripped myself up by choosing the OSS driver (we love OSS after all :) ), without realizing that my Pascal-era Tesla P4s aren't supported by the OSS driver.

## Acceptance

### DIFF
--- a/website/content/v1.8/talos-guides/configuration/nvidia-gpu-proprietary.md
+++ b/website/content/v1.8/talos-guides/configuration/nvidia-gpu-proprietary.md
@@ -21,6 +21,17 @@ Create the [boot assets]({{< relref "../install/boot-assets" >}}) which includes
 > Make sure the driver version matches for both the `nonfree-kmod-nvidia` and `nvidia-container-toolkit` extensions.
 > The `nonfree-kmod-nvidia` extension is versioned as `<nvidia-driver-version>-<talos-release-version>` and the `nvidia-container-toolkit` extension is versioned as `<nvidia-driver-version>-<nvidia-container-toolkit-version>`.
 
+## Proprietary vs OSS Nvidia Driver Support
+
+The NVIDIA Linux GPU Driver contains several kernel modules: `nvidia.ko`, `nvidia-modeset.ko`, `nvidia-uvm.ko`, `nvidia-drm.ko`, and `nvidia-peermem.ko`.
+Two "flavors" of these kernel modules are provided, and both are available for use within Talos:
+
+- Proprietary, This is the flavor that NVIDIA has historically shipped.
+- Open, i.e. source-published/OSS, kernel modules that are dual licensed MIT/GPLv2.
+  With every driver release, the source code to the open kernel modules is published on https://github.com/NVIDIA/open-gpu-kernel-modules and a tarball is provided on https://download.nvidia.com/XFree86/.
+
+The choice between Proprietary/OSS may be decided after referencing the Official [NVIDIA announcement](https://developer.nvidia.com/blog/nvidia-transitions-fully-towards-open-source-gpu-kernel-modules/).
+
 ## Enabling the NVIDIA modules and the system extension
 
 Patch Talos machine configuration using the patch `gpu-worker-patch.yaml`:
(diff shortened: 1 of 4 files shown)


website/content/v1.8/talos-guides/configuration/nvidia-gpu-proprietary.md
```diff
@@ -21,6 +21,21 @@ Create the [boot assets]({{< relref "../install/boot-assets" >}}) which includes
...
+## Proprietary vs OSS Nvidia Driver Support
+
+The NVIDIA Linux GPU Driver contains several kernel modules: `nvidia.ko`, `nvidia-modeset.ko`, `nvidia-uvm.ko`, `nvidia-drm.ko`, and `nvidia-peermem.ko`. Two "flavors" of these kernel modules are provided, and both are available for use within Talos:
+
+* Proprietary. This is the flavor that NVIDIA has historically shipped.
+* Open, i.e. source-published/OSS, kernel modules that are dual licensed MIT/GPLv2. With every driver release, the source code to the open kernel modules is published on https://github.com/NVIDIA/open-gpu-kernel-modules and a tarball is provided on https://download.nvidia.com/XFree86/.
+
+The choice between Proprietary/OSS may depend on your hardware:
+
+* The proprietary flavor supports the GPU architectures [Maxwell](https://en.wikipedia.org/wiki/Maxwell_(microarchitecture)), [Pascal](https://en.wikipedia.org/wiki/Pascal_(microarchitecture)), [Volta](https://en.wikipedia.org/wiki/Volta_(microarchitecture)), [Turing](https://en.wikipedia.org/wiki/Turing_(microarchitecture)), and later GPUs.
```
smira, line 33: Thanks for the update, but I would rather prefer it to be less specific, as these details would change over time.

Talos already ships with two versions of NVIDIA drivers (535/555), and they might have different support for GPUs. 

So if we can link towards the NVIDIA documentation on that, that would be perfect.

SlothCroissant: Hey there @smira , not sure the best approach here then - Nvidia simply says "OSS drivers are for Turing and newer", without being much more specific (the reliance on GSP is the reason). You're right in that this may change as time goes on, but not sure how you'd want to document it.

NVIDIA's docs aren't super obvious, you need to look at each driver release's README or their announcement blogs to get the details, but once you do, it's always the same - Turning+ = OSS. Before that, closed-source. This means it's difficult to just redirect users to the NVIDIA docs as a one-size-fits-all reference.

I personally think it's safe to go with the above (it was pulled straight from their docs) or a variation of it, and barring any fundamental changes NVIDIA won't be backporting the OSS drivers to pre-Turing (due to GSP, which doesn't exist at a hardware level before Turing). 

Up to you on how you'd like to implement this, I'm good with whatever (or nothing), just wanted to pass this along to help others if they stumble on the same issue :)

frezbo: @jfroy have any suggestions?

jfroy: Speaking as an NVIDIA employee, I would recommend you link to [this post](https://developer.nvidia.com/blog/nvidia-transitions-fully-towards-open-source-gpu-kernel-modules/) on our technical blog, and summarize our guidance directly if you wish.

Quoting:

> ### Supported GPUs
> Not every GPU is compatible with the open-source GPU kernel modules.
> 
> For cutting-edge platforms such as NVIDIA Grace Hopper or NVIDIA Blackwell, you must use the open-source GPU kernel modules. The proprietary drivers are unsupported on these platforms.
> 
> For newer GPUs from the Turing, Ampere, Ada Lovelace, or Hopper architectures, NVIDIA recommends switching to the open-source GPU kernel modules.
> 
> For older GPUs from the Maxwell, Pascal, or Volta architectures, the open-source GPU kernel modules are not compatible with your platform. Continue to use the NVIDIA proprietary driver.
> 
> For mixed deployments with older and newer GPUs in the same system, continue to use the proprietary driver.

To summarize, "use the open-source GPU kernel modules whenever possible".

frezbo (approved):

PR #9453, opened 2024-10-05
fix: do not stop udevd before unmounting volumes

dsseng:
As udevd is required by cryptsetup which will timeout if udevd is not working, do not stop it in StopServicesEphemeral, but let StopAllServices handle udev shutdown after cryptsetup close is called

Ref: https://bbs.archlinux.org/viewtopic.php?id=162415


dsseng:
e2e-qemu-short failed due to `TestIntegration/cli.SupportSuite/TestSupport`

dsseng:
Unsure about LVM2 test failures though

smira:
I want to fix `encrypted-vip` tests in https://github.com/siderolabs/talos/pull/9442

smira (approved):

dsseng:
Can this be merged with this CI failure?

smira:
> Can this be merged with this CI failure?

yes, unlabel the encrypted-vip, rebase, push and you should be gtg

PR #9455, opened 2024-10-06
fix: improve container detection

dsseng:
Instead of relying on cmdline (which will not work in case it's TinK on Talos, for example), add a file to container rootfs to signal the platform to machined.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -760,6 +760,7 @@ RUN <<END
 END
 
 FROM rootfs-base-${TARGETARCH} AS rootfs-base
+RUN echo "true" > /rootfs/usr/etc/in-container
 RUN find /rootfs -print0 \
     | xargs -0r touch --no-dereference --date="@${SOURCE_DATE_EPOCH}"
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -760,6 +760,7 @@ RUN <<END
 END
 
 FROM rootfs-base-${TARGETARCH} AS rootfs-base
+RUN echo "true" > /rootfs/usr/etc/is-container
```
smira, line 763: I think `in-container` might sound better.

internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
```diff
@@ -67,6 +67,12 @@ const (
 
 // CurrentPlatform is a helper func for discovering the current platform.
 func CurrentPlatform() (p runtime.Platform, err error) {
+	if _, err := os.Stat("/usr/etc/is-container"); err == nil {
+		fmt.Println("Found /usr/etc/is-container: platform is container")
```
smira, line 71: I don't think we should leave this message here, as it runs super early in the boot, probably before the log is set up.

smira (approved):

PR #9464, opened 2024-10-07
fix: update grpc-go the latest patch release

smira:
See https://github.com/grpc/grpc-go/releases/tag/v1.66.3

Specifically stream failures, I wonder if that is causing flaky support script.


dsseng (approved):
We'll see whether this fixes flaky test

Issue #8805, opened 2024-05-26
[Feature] [Hetzner cloud] Please accept server user-data in base64+gzip format


conrity:
## Feature Request

Please accept server user-data in base64+gzip format. Hetzner allow only 32Kb user-data size, which makes impossible to deploy talos generated configs without gzipping (gzip works only with base64).

I use terraform for deploying talos and endcode data with hashicorp/cloudinit provider:

```
data "cloudinit_config" "cpn" {
  gzip          = true
  base64_encode = true

  # Main cloud-config configuration file.
  part {
    content_type = "text/cloud-config"
    content      = "${data.talos_machine_configuration.cpn.machine_configuration}"
  }

}
```

Hetzner puts decoded file here:

![image](https://github.com/siderolabs/talos/assets/82505321/46144232-91bc-4ff9-b4f1-67682324c75d)

steverfrancis:
You can generate your Taos configs with
--with-examples=false --with-docs=false

Which should make them fit fine as is

PR #9467, opened 2024-10-07
feat: optionally decode hcloud userdata as base64 when fetching machine configuration

axiphi:
# Pull Request

## What? (description)

When fetching the machine configuration in the hcloud platform implementation, try to decode the data returned from the 'userdata' endpoint as a base64 string. If the data is not in base64 format, decoding does not succeed and the unmodified data is used (the previous behavior).
By allowing this base64 based input method, users can provide binary data instead of just UTF-8 as the machine configuration input and therefore make use of the built in Gzip compression mechanism. 

## Why? (reasoning)

The userdata that can be provided when creating a hcloud server is limited to 32kB.
When the machine configuration is provided in plain text format, this limit can be reached quickly, especially when using inline manifests. 

[This issue](https://github.com/siderolabs/talos/issues/8805) is caused by this limitation, though this PR provides a different solution from what is suggested in the issue.
(This PR tries to decode the userdata as base64 string directly instead of using the cloud-init mechanism.)

## Acceptance


smira:
Sorry, I don't think I understand this - base64 data is bigger than plain text, so how does it help?

axiphi:
Sorry, could have been more clear about this initially: With the base64-based input method, users can provide binary data instead of just UTF-8 as the machine configuration input and therefore make use of the built in Gzip compression mechanism. This is the part where the data actually becomes smaller.

sergelogvinov:
It make sense for all platforms.. Most of them have size limitations...

Thanks.

smira (approved):
I did some changes, thank you for your PR!

axiphi:
> It make sense for all platforms.. Most of them have size limitations...
> 
> Thanks.

Hey @sergelogvinov,
sorry for the very late reply, I didn't see your message.
The reason I put the base64 decoding step into the hcloud-specific module instead of the generic initialization code is that other platforms already have built-in support for base64-encoded user-data at the platform level (e.g. see [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html)).

I guess you could make an argument that Talos could support this generically for all platforms anyway though.

PR #9475, opened 2024-10-08
docs: add egress documentation

rsmitty:
This PR adds a list of the domains I had to allow for a Talos cluster to pull all assets needed to install and bootstrap. I've added these docs back to 1.6 of Talos, as I'm not certain they would apply to anything earlier.

### DIFF
--- a/website/content/v1.6/advanced/egress-domains.md
+++ b/website/content/v1.6/advanced/egress-domains.md
@@ -1,0 +1,44 @@
+---
+title: "Egress Domains"
+description: "Allowing outbound access for installing Talos"
+aliases:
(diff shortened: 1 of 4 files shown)


smira:
Should we move it closer to the airgapped docs in the advanced section?

rsmitty:
Yeah, maybe. I waffled on where to put this, as it's kind of required to get an install working in these type of environments. But advanced makes just as much sense. I'll move it there as I fix these lint errors.

rsmitty:
Now moved into advanced and updated to fix lint errors.

smira (approved):

PR #9484, opened 2024-10-09
feat: add recommended app.kubernetes.io labels to control-plane Pods

### LINKED ISSUE #9450: Add newer well-known `app.kubernetes.io/` labels to control plane Pods
## Feature Request

In addition to `tier` and `k8s-app` the control plane pods should probably use labels such as `app.kubernetes.io/part-of` and `app.kubernetes.io/name`. Those would be my suggestions for the analog labels at least. It might be worth checking what labels/values other distributions/tools set.

### Description

I suspect `tier` and `k8s-app` are older well-known labels but the `app.kubernetes.io` labels are by now pretty well established and understood by other tools, for example Loki.

Happy to make a PR one of these days, just adding these and leaving the others for now, if it would be approved.

michaelbeaumont:
# Pull Request

## What? (description)

Closes #9450. This PR adds most of the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). These are just my own suggestions though, feel free to suggest different ones or fewer. In particular, not sure about /component vs /part-of so maybe better to leave them out? I also left out `/instance`, not quite sure what to set that to or how, probably something with the node name though.

## Why? (reasoning)

It's useful to have a set of uniform labels across Pods when looking at metrics and logs or even `kubectl` output, for example.

## Acceptance

  - not really applicable
  - seemed to have unrelated failures

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -462,8 +462,13 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 					constants.AnnotationStaticPodConfigVersion:     configResource.Metadata().Version().String(),
 				},
 				Labels: map[string]string{
-					"tier":    "control-plane",
-					"k8s-app": k8s.APIServerID,
+					"tier":                         "control-plane",
+					"k8s-app":                      k8s.APIServerID,
+					"component":                    k8s.APIServerID,
+					"app.kubernetes.io/name":       k8s.APIServerID,
+					"app.kubernetes.io/version":    compatibility.VersionFromImageRef(cfg.Image).String(),
+					"app.kubernetes.io/component":  "control-plane",
+					"app.kubernetes.io/managed-by": "Talos",
 				},
 			},
 			Spec: v1.PodSpec{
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
```diff
@@ -462,8 +462,13 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
...
 				},
 				Labels: map[string]string{
-					"tier":    "control-plane",
-					"k8s-app": k8s.APIServerID,
+					"tier":                         "control-plane",
+					"k8s-app":                      k8s.APIServerID,
+					"app.kubernetes.io/name":       k8s.APIServerID,
+					"app.kubernetes.io/version":    compatibility.VersionFromImageRef(cfg.Image).String(),
+					"app.kubernetes.io/component":  "control-plane",
+					"app.kubernetes.io/part-of":    "Kubernetes",
```
smira, line 470: I wonder if it should be lower-case?

smira: (I don't have kubeadm cluster to check this out, but probably should be same)

smira: kubeadm does something different: https://github.com/kubernetes/kubernetes/blob/2fda3e4a650d8e729c509609aa865f0b6d843b4e/cmd/kubeadm/app/util/staticpod/utils.go#L75-L76

smira: I'm totally fine for Talos to have more labels, but I guess it makes sense to be somewhat compatible

smira: I did some small changes, thank you for your PR

smira (approved):

PR #3441, opened 2021-04-06
fix: document HDMI problem on RPi 4

AlekSi:
Closes #3414.

- [x] you linked an issue (if applicable)
- [ ] you included tests (if applicable)
- [x] you ran conformance (`make conformance`)
- [x] you formatted your code (`make fmt`)
- [x] you linted your code (`make lint`)
- [x] you generated documentation (`make docs`)
- [x] you ran unit-tests (`make unit-tests`)

PR #9486, opened 2024-10-10
chore: bump grpc library

### LINKED ISSUE #9463: Find a way to use `grpc-go >= 1.67.0` with `ALPN` disabled
Starting with [v1.67.0](https://github.com/grpc/grpc-go/releases/tag/v1.67.0) Go gRPC implementation no longer work with TLS connections that don't support ALPN. There is a special header `GRPC_ENFORCE_ALPN_ENABLED` which, when set to `false`, should disable this behavior. But setting it after the process started (even in `init` block) has no effect, because [relevant variable](https://github.com/grpc/grpc-go/blob/859602c14c6c8bcc4f3b7e6c129577229c9f9294/internal/envconfig/envconfig.go#L48) initialization happens before any init blocks.

DmitriyMV:
Bump grpc library to 1.67.1 and ensure that we it still works with HTTP/2 ALPN value changes.


.golangci.yml
```diff
@@ -109,6 +109,8 @@ linters-settings:
       - github.com/insomniacslk/dhcp
       - github.com/safchain/ethtool
       - github.com/vishvananda/netlink
+      - github.com/siderolabs/crypto
```
smira, line 112: I guess we don't need that?

go.mod
```diff
@@ -139,7 +139,7 @@ require (
 	github.com/ryanuber/go-glob v1.0.0
 	github.com/safchain/ethtool v0.4.1
 	github.com/scaleway/scaleway-sdk-go v1.0.0-beta.30
-	github.com/siderolabs/crypto v0.4.4
+	github.com/siderolabs/crypto v0.5.0
```
smira, line 142: can we split up this PR into the piece which only bumps crypto and fixes ALPN (which we can backport to 1.8), and another PR which does the rest of updates?

cmd/talosctl/acompat/acompat.go
```diff
@@ -0,0 +1,14 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package acompat provides compatibility with gRPC 1.67.0 and later.
+package acompat
+
+import "os"
+
+func init() {
+	if err := os.Setenv("GRPC_ENFORCE_ALPN_ENABLED", "false"); err != nil {
```
smira, line 11: this will unfortunately be something we need more of: as the Talos cluster is upgraded, worker nodes talk to `trustd` on the controlplane, and also `apid`s talk to each other. So if 1.9 machine tries to talk to 1.8 machine, and ALPN issue will pop up

.golangci.yml
DmitriyMV: Yeah. thanks.

go.mod
DmitriyMV: Will do.

cmd/talosctl/acompat/acompat.go
DmitriyMV: Done in https://github.com/siderolabs/talos/pull/9504

DmitriyMV:
Superseded by #3441

PR #9489, opened 2024-10-11
feat: remove wrapperd and launch processes directly

### LINKED ISSUE #9472: refactor: remove `wrapperd` in favor launcher
Experiment, remove `wrapperd` completely replacing it with https://pkg.go.dev/kernel.org/pub/linux/libs/security/libcap/cap#NewLauncher

The idea is that we should be able to set most (if not all) of wrapperd stuff by using Launcher.

This should cut our latency to fork/exec child programs.

dsseng:
Fixes #9472 

TODO:
- [x] tests
- [x] finalize


internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,49 +86,97 @@ func (p *processRunner) Close() error {
...
 func (p *processRunner) build() (commandWrapper, error) {
-	args := []string{
-		fmt.Sprintf("-name=%s", p.args.ID),
-		fmt.Sprintf("-dropped-caps=%s", strings.Join(p.opts.DroppedCapabilities, ",")),
-		fmt.Sprintf("-cgroup-path=%s", cgroup.Path(p.opts.CgroupPath)),
-		fmt.Sprintf("-oom-score=%d", p.opts.OOMScoreAdj),
-		fmt.Sprintf("-uid=%d", p.opts.UID),
+	wrapper := commandWrapper{}
+
+	env := []string{fmt.Sprintf("PATH=%s", constants.PATH)}
```
DmitriyMV, line 102: Simple string + should do the trick here I think

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,49 +86,97 @@ func (p *processRunner) Close() error {
...
+	env = append(env, p.opts.Env...)
```
DmitriyMV, line 103: There is `slices.Concat` exactly for this case.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,49 +86,97 @@ func (p *processRunner) Close() error {
...
+	if p.opts.UID > 0 {
+		launcher.SetUID(int(p.opts.UID))
 	}
 
-	args = append(args, p.args.ProcessArgs...)
+	droppedCaps := strings.Join(p.opts.DroppedCapabilities, ",")
+
+	prop, err := krnl.ReadParam(&kernel.Param{Key: "proc.sys.kernel.kexec_load_disabled"})
+	if v := strings.TrimSpace(string(prop)); err == nil && v != "0" {
+		log.Printf("kernel.kexec_load_disabled is %v, skipping dropping capabilities", v)
```
DmitriyMV, line 115: %v should be %q or %s

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +88,75 @@ func (p *processRunner) Close() error {
...
+				fmt.Printf("failed to parse capability: %s", capErr)
+			}
+
+			return capability
+		})
+
+		// reduce capabilities and assign them to launcher
+		iab := cap.IABGetProc()
+		if err = iab.SetVector(cap.Bound, true, dropCaps...); err != nil {
+			return commandWrapper{}, fmt.Errorf("failed to set capabilities: %s", err)
```
DmitriyMV, line 130: Always use `%w` for variables that implement or themselves are `error` type in `fmt.Errorf`. [Details](https://pkg.go.dev/fmt#Errorf).

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +282,53 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
 	}
 
-	eventSink(events.StateRunning, "Process %s started with PID %d", p, cmdWrapper.cmd.Process.Pid)
+	process, err := os.FindProcess(pid)
+	if err != nil {
+		return fmt.Errorf("could not find process: %w", err)
+	}
+
+	if err := sys.AdjustOOMScore(pid, p.opts.OOMScoreAdj); err != nil {
+		return fmt.Errorf("failed to change OOMScoreAdj of process %s to %d", p, p.opts.OOMScoreAdj)
```
DmitriyMV, line 296: I think it's better that we return `AdjustOOMScore` error here too.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +282,53 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
+	}
+
+	cgroupPath := cgroup.Path(p.opts.CgroupPath)
+
+	if cgroups.Mode() == cgroups.Unified {
+		cgv2, err := cgroup2.Load(cgroupPath)
+		if err != nil {
+			return fmt.Errorf("failed to load cgroup %s: %s", cgroupPath, err)
```
DmitriyMV, line 304: Always use `%w` for variables that implement or themselves are `error` type in `fmt.Errorf`. [Details](https://pkg.go.dev/fmt#Errorf). You can specify it several times.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +282,53 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
+		}
+
+		if err := cgv2.AddProc(uint64(pid)); err != nil {
+			return fmt.Errorf("failed to move process %s to cgroup: %s", p, err)
```
DmitriyMV, line 308: Always use `%w` for variables that implement or themselves are `error` type in `fmt.Errorf`. [Details](https://pkg.go.dev/fmt#Errorf).

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +282,53 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
+		}
+	} else {
+		cgv1, err := cgroup1.Load(cgroup1.StaticPath(cgroupPath))
+		if err != nil {
+			return fmt.Errorf("failed to load cgroup %s: %s", cgroupPath, err)
```
DmitriyMV, line 313: Always use `%w` for variables that implement or themselves are `error` type in `fmt.Errorf`. [Details](https://pkg.go.dev/fmt#Errorf).

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +282,53 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
+		}
+
+		if err := cgv1.Add(cgroup1.Process{
+			Pid: pid,
+		}); err != nil {
+			return fmt.Errorf("failed to move process %s to cgroup: %s", p, err)
```
DmitriyMV, line 319: Always use `%w` for variables that implement or themselves are `error` type in `fmt.Errorf`. [Details](https://pkg.go.dev/fmt#Errorf).

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +88,90 @@ func (p *processRunner) Close() error {
...
+		pa.Sys.Setctty = true
+	}
+
+	pa.Files = []uintptr{
+		wrapper.stdin,
+		wrapper.stdout,
+		wrapper.stderr,
+	}
+
+	// TODO: use pa.Sys.CgroupFD here when we can be sure clone3 is available
```
smira, line 152: let's absolutely do this

dsseng: So implement 2 different pathways? We use this code in container mode as well, and we cannot really expect containers to be ran on kernel 5.5+ which has clone3

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +88,90 @@ func (p *processRunner) Close() error {
...
+	stdout           uintptr
+	stderr           uintptr
 	afterStart       func()
 	afterTermination func() error
 }
 
+func dropCaps(droppedCapabilities []string, launcher *cap.Launcher) error {
+	droppedCaps := strings.Join(droppedCapabilities, ",")
+
+	prop, err := krnl.ReadParam(&kernel.Param{Key: "proc.sys.kernel.kexec_load_disabled"})
```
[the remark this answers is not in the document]
dsseng: Yes, because kexec is not the single bad thing (and also it's typically protected by secureboot validation). Will remove this check and drop anyway

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +85,86 @@ func (p *processRunner) Close() error {
...
 
-	args = append(args, p.args.ProcessArgs...)
+	env := slices.Concat([]string{"PATH=" + constants.PATH}, p.opts.Env, os.Environ())
+	launcher := cap.NewLauncher(p.args.ProcessArgs[0], p.args.ProcessArgs, env)
 
-	cmd := exec.Command("/sbin/wrapperd", args...)
+	// forkExec fails if we do setuid
+	// if p.opts.UID > 0 {
+	// 	launcher.SetUID(int(p.opts.UID))
+	// }
```
dsseng, line 160: And if we move this to system process args, then Launcher's validation triggers. Going to take a look later.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -107,18 +173,46 @@ func (p *processRunner) build() (commandWrapper, error) {
 	}
 
 	var writer io.Writer
-	if p.debug { // TODO: wrap it into LoggingManager
+	if p.debug || true { // TODO: wrap it into LoggingManager
```
dsseng, line 176: todo remove

.golangci.yml
```diff
@@ -105,6 +105,7 @@ linters-settings:
       - golang.zx2c4.com/wireguard
       - golang.zx2c4.com/wireguard/wgctrl
       - cloud.google.com/go
+      - kernel.org/pub/linux/libs/security/libcap/cap
```
dsseng, line 108: TODO: remove as this was only debug

go.mod
```diff
@@ -24,6 +24,9 @@ replace (
 	// forked go-yaml that introduces RawYAML interface, which can be used to populate YAML fields using bytes
 	// which are then encoded as a valid YAML blocks with proper indentiation
 	gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20220527175918-f17b0f05cf2c
+
+	// improved error logging
+	kernel.org/pub/linux/libs/security/libcap/cap => github.com/dsseng/go-libcap/cap v0.0.0-20241015195416-c3ab072bd718
```
dsseng, line 29: todo remove

internal/app/machined/pkg/system/runner/process/process_test.go
```diff
@@ -224,6 +224,7 @@ func TestProcessSuite(t *testing.T) {
 		t.Skip("wrapperd not found")
 	}
 
+	// What's the purpose of this test? Should it be replaced for new subprocess start method?
```
dsseng, line 227: Needs review

internal/app/machined/pkg/system/services/dashboard.go
```diff
@@ -66,7 +66,7 @@ func (d *Dashboard) Runner(r runtime.Runtime) (runner.Runner, error) {
 		}),
 		runner.WithStdinFile(tty),
 		runner.WithStdoutFile(tty),
-		runner.WithCtty(1),
+		runner.WithCtty(0),
```
dsseng, line 69: ctty(1) failed as stdout is opened as write-only. We might want to change that, but perhaps using stdin as ctty is perfectly fine as we pass both for TTY to work

dsseng:
Maybe the test should be a separate PR going in before refactoring. If so, will do later today

internal/app/machined/pkg/system/runner/process/process_test.go
smira: process reaper is to the fact that Talos machine is PID=1 and it should reap all zombies, so it runs a `waitpid()` loop which breaks Go's wait for the process (as it reaps before Go might have a chance), so unrelated to wrapperd

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -158,37 +251,60 @@ func (p *processRunner) build() (commandWrapper, error) {
...
+func applyProperties(p *processRunner, pid int) error {
+	path := cgroup.Path(p.opts.CgroupPath)
+
+	if cgroups.Mode() == cgroups.Unified {
+		cgv2, err := cgroup2.Load(path)
+		if err != nil {
+			return fmt.Errorf("failed to load cgroup %s: %w", path, err)
+		}
+
+		if err := cgv2.AddProc(uint64(pid)); err != nil {
```
smira, line 287: can we move to cgroupFD?

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +322,29 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
+		return fmt.Errorf("could not find process: %w", err)
 	}
 
-	eventSink(events.StateRunning, "Process %s started with PID %d", p, cmdWrapper.cmd.Process.Pid)
-
 	waitCh := make(chan error)
 
 	go func() {
-		waitCh <- reaper.WaitWrapper(usingReaper, notifyCh, cmdWrapper.cmd)
+		_, err := process.Wait()
```
smira, line 346: should use reaper, don't change this part please

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Don't we also use this launch method when running in a container (thus not sure we have clone3 available)? Cgroup in clone3 is actually only available since kernel 5.7

smira: good point, let's skip it for now

frezbo: can we optionally have two paths? like auto-detect?

dsseng: Unsure whether that's going to be beneficial, but it's possible for sure

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Okay, refactoring here to accept raw pid and so: https://github.com/siderolabs/go-cmd/blob/main/pkg/cmd/proc/reaper/wait.go#L13-L38

internal/app/machined/pkg/system/runner/process/process.go
dsseng: clone3 into cgroup is only possible with cgv2 ([ref](https://github.com/systemd/systemd/blob/e8fb0643c1bea626d5f5e880c3338f32705fd46d/src/basic/process-util.c#L2043)). Will implement clone3 for the case we are running native with cgv2.

frezbo: we'll always be running with cgroupv2, cgroupv1 is also not fully removed from systemd, so that's the general directions from all distros

smira: We still support cgroupsv1 both for native launch, and we'll have to support in container mode.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -158,37 +264,75 @@ func (p *processRunner) build() (commandWrapper, error) {
...
-			return w.Close()
-		},
-	}, nil
+	platform, err := platform.CurrentPlatform()
+	if err == nil {
+		cgroupFdSupported = platform.Mode() != runtime.ModeContainer
+	}
+
+	// cgroupfd is more reliable, use it when possible
+	if cgroups.Mode() == cgroups.Unified && cgroupFdSupported && p.opts.UID == 0 {
```
dsseng, line 295: Instead of using `p.opts.UID == 0` we could chown the dashboard cgroup to be owned by dashboard's UID. Else it'll give EACCES on clone3 call

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +88,95 @@ func (p *processRunner) Close() error {
...
-		fmt.Sprintf("-uid=%d", p.opts.UID),
+func dropCaps(droppedCapabilities []string, launcher *cap.Launcher) error {
+	droppedCaps := strings.Join(droppedCapabilities, ",")
+
+	if droppedCaps != "" {
+		caps := strings.Split(droppedCaps, ",")
+		dropCaps := xslices.Map(caps, func(c string) cap.Value {
+			capability, capErr := cap.FromName(c)
+			if capErr != nil {
+				fmt.Printf("failed to parse capability: %s", capErr)
```
smira, line 109: my understanding is that should never happen, as `cap.FromName` is statically defined, so we could even `panic` here

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +88,95 @@ func (p *processRunner) Close() error {
...
+		}
+
+		launcher.SetIAB(iab)
+	}
+
+	return nil
+}
+
+// This callback is run in the thread before executing child process.
+func beforeExecCallback(pa *syscall.ProcAttr, data interface{}) error {
```
smira, line 127: nit, better syntax:

```suggestion
func beforeExecCallback(pa *syscall.ProcAttr, data any) error {
```

smira (approved):
great work! :cool:

dsseng:
https://github.com/siderolabs/talos/actions/runs/11515928408/job/32057560561?pr=9489#step:18:6799

there's a chance this process just completes before cgroup code kicks in. What can we do? Qualify `no such process` as a warning?

smira:
> [siderolabs/talos/actions/runs/11515928408/job/32057560561?pr=9489#step:18:6799](https://github.com/siderolabs/talos/actions/runs/11515928408/job/32057560561?pr=9489#step:18:6799)
> 
> there's a chance this process just completes before cgroup code kicks in. What can we do? Qualify `no such process` as a warning?

I guess so. As we have a test which verifies that processes are in a proper cgroup, we should be good

internal/app/machined/pkg/system/runner/process/process.go
smira: I think either way is fine, we can certainly chown the cgroups if we know the UIDs of the owner (e.g. the dashboard).

dsseng: Will then chown, since we decided our preferred way is using cgroupfd when possible

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -158,37 +264,84 @@ func (p *processRunner) build() (commandWrapper, error) {
...
+
+	// cgroupfd is more reliable, use it when possible
+	if cgroups.Mode() == cgroups.Unified && cgroupFdSupported {
+		cg, err := os.Open(cgroupPath)
+		if err == nil {
+			wrapper.cgroupFd = int(cg.Fd())
+		}
+	}
+
+	return wrapper, nil
```
frezbo, line 310: @smira since we return here do we need to tell the go runtime to keep the file open  :thinking:

smira: good question, let's pull in our expert @DmitriyMV

internal/app/machined/pkg/system/runner/process/process.go
dsseng: chown is not really sufficient

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -113,12 +192,40 @@ func (p *processRunner) build() (commandWrapper, error) {
...
+		if err != nil {
+			e = err
+		}
+
+		err = pw.Close()
+		if err != nil {
+			e = err
+		}
+
+		return e
```
DmitriyMV, line 223:
```suggestion
		for _, closer := range []io.Closer{w, pr, pw} {
			err := closer.Close()
			if err != nil {
				e = err
			}
		}
		
		return e
```

internal/app/machined/pkg/system/runner/process/process.go
DmitriyMV: From the docs for [os.File.Fd](https://pkg.go.dev/os#File.Fd):

>// If f is closed, the file descriptor becomes invalid.
// If f is garbage collected, a **finalizer may close** the file descriptor,
// **making it invalid**; see [runtime.SetFinalizer] for more information on when
// a finalizer might be run. On Unix systems this will cause the [File.SetDeadline]
// methods to stop working.

Emphasis mine. So yeah, GC is free to close this handle (and handles above) and code will break. I'm actually curious why you store `uintptr` instead of `*os.File` or `interface {io.Closer; Fd() uintptr}`? This would be much safer.

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Really nice, thank you!

internal/app/machined/pkg/system/runner/process/process.go
dsseng: okay, we can probably store them in commandWrapper. Is that going to be a sufficient lifetime? Perhaps so. Anyway, do we have to close those, or they're going to be cleaned up by kernel when child process exits? Going to take a look

DmitriyMV: Just use `afterTermination` function field call to close them anyway.

dsseng: Will also store cgroupfd there as we must dispose of it after forking

dsseng: @DmitriyMV PTAL at the recent fix.

I found no issue with stdio, but replaced storing raw fds with *os.File, yes. Maybe I should add more non-nil checks to all the places they're used in, but likely not.

Added cgroupFile in the same way to close it

DmitriyMV: Are you sure you need to close `stdin`/`stdout`/`stderr` right after forking ends and not after forked process completes?

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -206,20 +359,29 @@ func (p *processRunner) run(eventSink events.Recorder) error {
...
 	}
 
-	eventSink(events.StateRunning, "Process %s started with PID %d", p, cmdWrapper.cmd.Process.Pid)
-
 	waitCh := make(chan error)
 
 	go func() {
-		waitCh <- reaper.WaitWrapper(usingReaper, notifyCh, cmdWrapper.cmd)
+		err := reaper.ProcessWaitWrapper(usingReaper, notifyCh, process)
+		waitCh <- err
```
DmitriyMV, line 384:
```suggestion
		waitCh <- reaper.ProcessWaitWrapper(usingReaper, notifyCh, process)
```

Perhaps?

dsseng: :100: applied locally together with lint fix, thank you!

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Well, it was done by the old code (the part invoking wrapperd) and tests are okay. I didn't reconsider this and am not sure how are such handed-off fd's handled after fork, but I believe it'd fail reliably in case our approach to it is wrong.

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -78,27 +90,95 @@ func (p *processRunner) Close() error {
...
+			pa.Sys = &syscall.SysProcAttr{}
+		}
+
+		pa.Sys.Ctty = ctty
+		pa.Sys.Setsid = true
+		pa.Sys.Setctty = true
+	}
+
+	pa.Files = []uintptr{
+		wrapper.stdin.Fd(),
```
dsseng, line 147: In case the caller doesn't specify the stdin file to use, we end up with nil -> fd 0 here, so the child will be given Talos' own stdin. Should we override that with /dev/null?

frezbo: i think we need nil checks then, otherwise calling methods on `os.File` would panic :thinking:

dsseng: We only add calls to .Close in case we opened the file successfully, and .Fd doesn't fail but returns 0 (stdin) in case of being called on nil

frezbo (approved):

PR #9501, opened 2024-10-14
feat: introduce metal agent mode

utkuozdemir:
Introduce a new operating mode called the metal agent mode.

The mode is activated by the presence of a `/usr/local/etc/is-metal-agent` file under the root FS.

In this mode, Talos will:
- Only run the Initialize sequence, won't follow it up with the install/boot sequences
- Mark STATE partitions as `missing`, so Talos will always be in "not installed" state.
- Block applying configuration via API while in maintenance mode.

This mode can be used e.g., to collect hardware information from bare-metal servers.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
@@ -113,7 +113,11 @@ func newPlatform(platform string) (p runtime.Platform, err error) {
 	case "hcloud":
 		p = &hcloud.Hcloud{}
 	case constants.PlatformMetal:
-		p = &metal.Metal{}
+		_, metalAgentCheckErr := os.Stat(constants.MetalAgentModeFlagPath)
+
+		p = &metal.Metal{
+			IsAgent: metalAgentCheckErr == nil,
+		}
 	case "opennebula":
 		p = &opennebula.OpenNebula{}
 	case "openstack":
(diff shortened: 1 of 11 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
```diff
@@ -113,7 +113,11 @@ func newPlatform(platform string) (p runtime.Platform, err error) {
 	case "hcloud":
 		p = &hcloud.Hcloud{}
 	case constants.PlatformMetal:
-		p = &metal.Metal{}
+		_, metalAgentCheckErr := os.Stat("/usr/local/etc/is-metal-agent")
```
smira, line 116: I know it's used only once, but it'd be nice to move it out to `pkg/machinery/constants` for consistency

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -168,7 +175,7 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
 	phases := PhaseList{}
 
 	switch r.State().Platform().Mode() { //nolint:exhaustive
-	case runtime.ModeContainer:
+	case runtime.ModeContainer, runtime.ModeMetalAgent:
```
smira, line 178: should never get here?

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
utkuozdemir: yep, was a defensive check, but now removed.

internal/app/machined/pkg/runtime/v1alpha1/platform/platform.go
utkuozdemir: done

PR #9514, opened 2024-10-16
feat: update Kubernetes to v1.32.0-alpha.1

smira:
Talos 1.9 is going to be shipped with Kubernetes v1.32 by default.


frezbo (approved):

DmitriyMV (approved):

smira:
I can only do that _after_ I release Talos 1.9.0-alpha.0, as the test thinks it is on 1.8, and fails to upgrade with K8s 1.32.

PR #9518, opened 2024-10-17
feat: update dependencies

smira:
Bring in new tools, pkgs, update Go dependencies and others.

In preparation for Talos 1.9.0-alpha.0.


internal/app/machined/pkg/controllers/cluster/discovery_service.go
```diff
@@ -239,6 +241,9 @@ func (ctrl *DiscoveryServiceController) Run(ctx context.Context, r controller.Ru
 				TTL:           defaultDiscoveryTTL,
 				Insecure:      discoveryConfig.TypedSpec().ServiceEndpointInsecure,
 				ClientVersion: version.Tag,
+				TLSConfig: &tls.Config{
```
frezbo, line 244: is this something we missed earlier?

smira: yes, I did the update in the client, this will support live reload of TLS certs, but afaik no one waits for it

frezbo (approved):

dsseng (approved):

PR #9524, opened 2024-10-18
test: add an integration test for verifying process parameters

dsseng:
Validate capabilities are dropped and cgroup, UID, environment and OOM adjustments are set


smira (approved):

internal/integration/api/process.go
```diff
@@ -0,0 +1,171 @@
...
+		r, err := suite.Client.Processes(nodeCtx)
+		suite.Require().NoError(err)
+
+		found := 0
+		for _, msg := range r.Messages {
+			var procs []*machineapi.ProcessInfo
+			procs = msg.Processes
+
+			for _, p := range procs {
+				if p.Command == "systemd-udevd" {
```
frezbo, line 82: probably use `switch` with fallthrough

dsseng: Oops, forgot that cherry-pick. Thanks!

frezbo (approved):

PR #9555, opened 2024-10-23
fix: rework the 'metal-iso' config acquisition

### LINKED ISSUE #9538: Impossible to use "metal-iso" from any partition
## Bug Report

In docs it is stated, that any correctly labeled partition on block device can be used to provide configuration to Talos - https://www.talos.dev/v1.8/reference/kernel/#metal-iso. But is it not. Only labeled filesystem that occupies whole block device can be used.

### Description

i digged the code and placing config file in partition doesn’t work. at least in version 1.7.7, that i checked. in [readConfigFromISO](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go#L122) definition probe.GetDevWithFileSystemLabel(constants.MetalConfigISOLabel) return correct device (partition) name (for example, /dev/loop0p7), but later in filesystem.Probe(dev.Device().Name()) .Name() return block device name without partition in it (/dev/loop0) and .Probe() can’t parse superblock because of it.

List of partitions:
```
NAME                     MAJ:MIN RM   SIZE RO TYPE  MOUNTPOINTS NAME               FSTYPE            FSVER    LABEL     UUID
loop0                      7:0    0    10G  0 loop              loop0
├─loop0p1                259:0    0   100M  0 part              loop0p1            vfat              FAT32    EFI       A9C5-03C6
├─loop0p2                259:1    0     1M  0 part              loop0p2
├─loop0p3                259:2    0  1000M  0 part              loop0p3            xfs                        BOOT      8310ba69-b39b-482b-80b6-79ebbdf1ffc9
├─loop0p4                259:3    0     1M  0 part              loop0p4
└─loop0p5                259:4    0     9M  0 part              loop0p5            ext4              1.0      metal-iso 2a1e612c-a81d-49e2-80e9-d6c27679874d
```

### Logs

[metal-iso-logs.txt](https://github.com/user-attachments/files/17463060/metal-iso-logs.txt)

### Environment

- Talos version: 1.7.7, 1.8.1
- Kubernetes version: doesn't apply
- Platform: libvirt/qemu/kvm

This patch should fix issue:
```
diff --git a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
index ccbf858f4..dfc456c24 100644
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/metal/metal.go
@@ -132,7 +132,7 @@ func readConfigFromISO(ctx context.Context, r state.State) ([]byte, error) {
        //nolint:errcheck
        defer dev.Close()

-       sb, err := filesystem.Probe(dev.Device().Name())
+       sb, err := filesystem.Probe(dev.Path)
        if err != nil {
                return nil, err
        }
@@ -141,7 +141,7 @@ func readConfigFromISO(ctx context.Context, r state.State) ([]byte, error) {
                return nil, stderrors.New("error while substituting filesystem type")
        }

-       if err = unix.Mount(dev.Device().Name(), mnt, sb.Type(), unix.MS_RDONLY, ""); err != nil {
+       if err = unix.Mount(dev.Path, mnt, sb.Type(), unix.MS_RDONLY, ""); err != nil {
                return nil, fmt.Errorf("failed to mount iso: %w", err)
        }
```

smira:
Fixes #9538

Re-do the implementation by using the volume management primitives, so that we can avoid/skip old code. This should fix all issues related to the partition/whole disk.

Fix issues in the volume management (exposed, as we haven't used it this way before).

Build a test case in `talosctl cluster create` to inject machine config via `metal-iso`.


pkg/provision/providers/qemu/node.go
```diff
@@ -309,3 +325,25 @@ func (p *provisioner) populateSystemDisk(disks []string, clusterReq provision.Cl
...
+		return "", err
+	}
+
+	defer os.RemoveAll(tmpDir) //nolint:errcheck
+
+	if err = os.WriteFile(filepath.Join(tmpDir, "config.yaml"), []byte(config), 0o644); err != nil {
+		return "", err
+	}
+
+	_, err = cmd.Run("mkisofs", "-joliet", "-rock", "-volid", "metal-iso", "-output", isoPath, tmpDir)
```
frezbo, line 343: I guess we need a check for the binary similar to how we check if other binaries exist

smira: oh... it might fail in the CI because of that.

I feel we should split it up and keep it for power devs only (not for users)

DmitriyMV (approved):

PR #9559, opened 2024-10-23
chore: simplify `DNSUpstreamController` and `DNSUpstream` resource

DmitriyMV:
This PR does those things:
- Fixes race condition where controller could potentially modify upstream, while other controller is copying its internals to the slice.
- Simplifies `run` function in `DNSUpstreamController` by removing all `Idx` handling.
- Removes `Idx` field from `DNSUpstream`. Upstreams are now sorted by their id with №X prefix.
- `Proxy` Stop is now called from the finalizer. In combination with iterators, this ensures that we only stop upstream when it's fully unreachable.


internal/app/machined/pkg/controllers/network/dns_upstream.go
```diff
@@ -103,36 +103,22 @@ func (ctrl *DNSUpstreamController) run(ctx context.Context, r controller.Runtime
...
+	}
+
+	for i, srv := range rs.TypedSpec().DNSServers {
+		remoteHost := srv.String()
 
 		if err = safe.WriterModify[*network.DNSUpstream](
 			ctx,
 			r,
-			network.NewDNSUpstream(remoteAddr),
+			network.NewDNSUpstream("№"+strconv.Itoa(i)+" "+remoteHost),
```
frezbo, line 117: "№" wha's this :thinking:

DmitriyMV: It's [numero sign](https://en.wikipedia.org/wiki/Numero_sign). I always forget that it's mostly replaced by # now.

internal/app/machined/pkg/controllers/network/dns_upstream.go
```diff
@@ -103,36 +103,22 @@ func (ctrl *DNSUpstreamController) run(ctx context.Context, r controller.Runtime
...
+	}
+
+	for i, srv := range rs.TypedSpec().DNSServers {
+		remoteHost := srv.String()
 
 		if err = safe.WriterModify[*network.DNSUpstream](
 			ctx,
 			r,
-			network.NewDNSUpstream(remoteAddr),
+			network.NewDNSUpstream(fmt.Sprintf("#%d %s", i, remoteHost)),
```
smira, line 117: if the goal is to get them sorted, let's use `%03d`

internal/app/machined/pkg/controllers/network/dns_upstream.go
```diff
@@ -144,7 +130,43 @@ func (ctrl *DNSUpstreamController) run(ctx context.Context, r controller.Runtime
...
+	existingConn := make(map[string]*network.DNSConn, upstream.Len())
+
+	for u := range upstream.All() {
+		existingConn[u.TypedSpec().Value.Conn.Addr()] = u.TypedSpec().Value.Conn
+	}
+
+	return func(spec *network.DNSUpstreamSpecSpec, remoteHost string, l *zap.Logger) {
+		remoteAddr := net.JoinHostPort(remoteHost, "53")
+		if spec.Conn != nil && spec.Conn.Addr() == remoteAddr {
+			l.Info("reusing existing upstream spec", zap.String("addr", remoteAddr))
```
smira, line 148: this feels like all should be `Debug`, nothing for the user to see here

pkg/machinery/go.mod
```diff
@@ -39,6 +39,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require go.uber.org/zap v1.27.0
```
smira, line 42: re-sort it back

DmitriyMV: 👍

internal/app/machined/pkg/controllers/network/dns_upstream.go
DmitriyMV: 👍

smira (approved):

PR #9564, opened 2024-10-24
fix: talosctl process null character

frezbo:
Fixes using grep on `talosctl ps` complaining about being a binary output.

### DIFF
--- a/internal/pkg/miniprocfs/processes.go
+++ b/internal/pkg/miniprocfs/processes.go
@@ -164,7 +164,7 @@ func (procs *Processes) readProc(pidString string) (*machine.ProcessInfo, error)
 	var label string
 
 	if err = procs.readFileIntoBuf(path + "attr/current"); err == nil {
-		label = string(bytes.TrimSpace(procs.buf))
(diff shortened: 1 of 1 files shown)


internal/pkg/miniprocfs/processes.go
```diff
@@ -164,7 +164,7 @@ func (procs *Processes) readProc(pidString string) (*machine.ProcessInfo, error)
 	var label string
 
 	if err = procs.readFileIntoBuf(path + "attr/current"); err == nil {
-		label = string(bytes.TrimSpace(procs.buf))
+		label = string(bytes.Trim(procs.buf, "\x000\n"))
```
dsseng, line 167: Shouldn't it still trim spaces as well?

frezbo: it does, `bytes.Trim` acceps cutstrings `"\x000\n`

smira: good point, we can add ` ` as well (idk if it can be there?)

smira (approved):

PR #9567, opened 2024-10-24
Add Mynewsdesk to ADOPTERS.md

dbackeus:
Ok I guess I managed to fail all CI with this tiny doc change.

Do I have to spend the time to figure this out or can we just push it through? 😅


DmitriyMV:
Greetings! Thanks for the PR. I'll update this PR and sign it, so CI should not complain about it.

smira (approved):

Issue #9565, opened 2024-10-24
Talos Secureboot ISO doesn't adhere to spec


frezbo:
## Bug Report

Talos Secureboot ISO doesn't adhere to spec, ref: #9397

See the difference below:

```bash
❯ iso-info -l _out/metal-amd64.iso 
iso-info version 2.1.0 x86_64-redhat-linux-gnu
Copyright (c) 2003-2005, 2007-2008, 2011-2015, 2017 R. Bernstein
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
__________________________________
ISO 9660 image: _out/metal-amd64.iso
Preparer    : XORRISO-1.5.6 2023.06.07.180001, LIBISOBURN-1.5.6, LIBISOFS-1.5.6, LIBBURN-1.5.6
Volume      : TALOS_V1_8_0_ALPHA_2_18_GD739D40
Volume Set  : Talos-v1.8.0-alpha.2-18-gd739d40db
No Joliet extensions
__________________________________
ISO-9660 Information
/:
  dr-xr-xr-x   1 0 0 [LSN     19]      2048 Sep 11 2024 17:08:04  .
  dr-xr-xr-x   1 0 0 [LSN     19]      2048 Sep 11 2024 17:08:04  ..
  dr-xr-xr-x   1 0 0 [LSN     21]      2048 Sep 11 2024 17:08:04  boot
  -r--r--r--   1 0 0 [LSN   1665]      2048 Sep 11 2024 17:08:04  boot.catalog
  dr-xr-xr-x   1 0 0 [LSN     76]      2048 Sep 11 2024 17:08:04  efi
  -r--r--r--   1 0 0 [LSN     84]   2949120 Sep 11 2024 17:08:04  efi.img
  -r--r--r--   1 0 0 [LSN   1664]         0 Sep 11 2024 17:08:04  mach_kernel
  dr-xr-xr-x   1 0 0 [LSN     78]      2048 Sep 11 2024 17:08:04  System
  dr-xr-xr-x   1 0 0 [LSN     81]      2048 Sep 11 2024 17:08:04  .disk

/boot/:
  dr-xr-xr-x   1 0 0 [LSN     21]      2048 Sep 11 2024 17:08:04  .
  dr-xr-xr-x   1 0 0 [LSN     19]      2048 Sep 11 2024 17:08:04  ..
[...]

PR #9580, opened 2024-10-27
fix: generation of SecureBoot iso

frezbo:
The Secureboot ISO previously generated has just an EFI parition.

Now generate iso with both the EFI partition and also having ISO filesystem having the content so *File System transposition* also works.

Fixes: #9565

### DIFF
--- a/pkg/imager/iso/uefi.go
+++ b/pkg/imager/iso/uefi.go
@@ -6,15 +6,14 @@ package iso
 
 import (
 	"bytes"
-	"context"
 	_ "embed"
 	"fmt"
 	"os"
 	"path/filepath"
 	"text/template"
-	"time"
 
 	"github.com/siderolabs/go-cmd/pkg/cmd"
+	"github.com/siderolabs/go-copy/copy"
 
 	"github.com/siderolabs/talos/pkg/imager/utils"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
@@ -67,6 +66,8 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 
 	efiBootImg := filepath.Join(options.ScratchDir, "efiboot.img")
 
+	isoRoot := filepath.Join(options.ScratchDir, "isoroot")
+
 	// initial size
 	isoSize := int64(10 * mib)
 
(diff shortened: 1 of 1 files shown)


pkg/imager/iso/uefi.go
```diff
@@ -188,33 +187,30 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
...
+		"all",
+		"-partition_offset", "16",
+		"-iso_mbr_part_type", "0x83",
 		"-no-emul-boot",
 		"-o", options.OutPath,
 		options.ScratchDir,
-		"--",
-	}
-
-	if epoch, ok, err := utils.SourceDateEpoch(); err != nil {
```
frezbo, line 199: `xorriso` tools support `SOURCE_DATE_EPOCH` from env vars and we run imager with it set, so this was redundant. From the man page:

```text
 SOURCE_DATE_EPOCH  belongs  to the specs of reproducible-builds.org.  It is supposed to be either undefined or to contain a decimal number which tells the seconds since january 1st 1970.
       If it contains a number, then it is used as time value to set the default of --modification-date=.  --gpt_disk_guid defaults to "modification-date".  The default of  --set_all_file_dates
       is then "set_to_mtime".  Further the "now" time for ISO nodes without disk source is then set to the SOURCE_DATE_EPOCH value.
       Startup files and program options can override the effect of SOURCE_DATE_EPOCH.
```

I've tested locally setting this produces a reproducible iso (tested with a simple directory since UKI is not reproducible)

frezbo:
iso-info:

```text
❯ iso-info -l _out/metal-amd64-secureboot.iso
iso-info version 2.1.0 x86_64-redhat-linux-gnu
Copyright (c) 2003-2005, 2007-2008, 2011-2015, 2017 R. Bernstein
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
__________________________________
ISO 9660 image: _out/metal-amd64-secureboot.iso
Preparer    : XORRISO-1.5.6 2023.06.07.180001, LIBISOBURN-1.5.6, LIBISOFS-1.5.6, LIBBURN-1.5.6
Volume      : TALOS_SB_V1_9_0_ALPHA_0_20_GA13C
Volume Set  : Talos-SB-v1.9.0-alpha.0-20-ga13cf76a3-dirty
Joliet Level: 3
__________________________________
ISO-9660 Information
/:
  d [LSN     57]      2048 Oct 24 2024 21:59:21  .
  d [LSN     57]      2048 Oct 24 2024 21:59:21  ..
  d [LSN     58]      2048 Oct 24 2024 21:59:21  EFI
  - [LSN     78]      2048 Oct 24 2024 21:59:21  boot.catalog
  d [LSN     62]      2048 Oct 24 2024 21:59:21  loader

/EFI/:
  d [LSN     58]      2048 Oct 24 2024 21:59:21  .
  d [LSN     58]      2048 Oct 24 2024 21:59:21  ..
  d [LSN     59]      2048 Oct 24 2024 21:59:21  BOOT
  d [LSN     60]      2048 Oct 24 2024 21:59:21  Linux
  d [LSN     61]      2048 Oct 24 2024 21:59:21  keys

/EFI/BOOT/:
  d [LSN     59]      2048 Oct 24 2024 21:59:21  .
  d [LSN     59]      2048 Oct 24 2024 21:59:21  ..
  - [LSN     79]     90800 Oct 24 2024 21:59:21  BOOTX64.EFI

/EFI/Linux/:
  d [LSN     60]      2048 Oct 24 2024 21:59:21  .
  d [LSN     60]      2048 Oct 24 2024 21:59:21  ..
  - [LSN    125] 101790896 Oct 24 2024 21:59:21  Talos-v1.9.0-alpha.0-20-ga13cf76a3-dirty.efi

/EFI/keys/:
  d [LSN     61]      2048 Oct 24 2024 21:59:21  .
  d [LSN     61]      2048 Oct 24 2024 21:59:21  ..
  - [LSN    124]      1325 Oct 24 2024 21:59:21  uki-signing-cert.der

/loader/:
  d [LSN     62]      2048 Oct 24 2024 21:59:21  .
  d [LSN     62]      2048 Oct 24 2024 21:59:21  ..
  d [LSN     63]      2048 Oct 24 2024 21:59:21  keys
  - [LSN  49834]        69 Oct 24 2024 21:59:21  loader.conf

/loader/keys/:
  d [LSN     63]      2048 Oct 24 2024 21:59:21  .
  d [LSN     63]      2048 Oct 24 2024 21:59:21  ..
  d [LSN     64]      2048 Oct 24 2024 21:59:21  auto

/loader/keys/auto/:
  d [LSN     64]      2048 Oct 24 2024 21:59:21  .
  d [LSN     64]      2048 Oct 24 2024 21:59:21  ..
  - [LSN  49830]      3492 Oct 24 2024 21:59:21  KEK.auth
  - [LSN  49832]      3492 Oct 24 2024 21:59:21  PK.auth
  - [LSN  49828]      3492 Oct 24 2024 21:59:21  db.auth
```

```text
❯ fdisk -l _out/metal-amd64-secureboot.iso
Disk _out/metal-amd64-secureboot.iso: 207.29 MiB, 217362432 bytes, 424536 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x00000000

Device                           Boot  Start    End Sectors  Size Id Type
_out/metal-amd64-secureboot.iso1          64 200703  200640   98M 83 Linux
_out/metal-amd64-secureboot.iso2      200704 423935  223232  109M ef EFI (FAT-12/16/32)
```

The first one is the standard iso directory to support File System Transposition and the second block is the EFI partition that UEFI can boot from.

The downside is this doubles the iso size, but it's a fair downside to have it work across multiple tools.

pkg/imager/iso/uefi.go
```diff
@@ -65,7 +64,7 @@ func CreateUEFI(printf func(string, ...any), options UEFIOptions) error {
 
 	printf("preparing raw image")
 
-	efiBootImg := filepath.Join(options.ScratchDir, "efiboot.img")
```
smira, line 68: this looks wrong, `os.TempDir()` is basically `/tmp`

frezbo: yeh, I just didn't want this file to go in iso

smira (commented):
so the ISO contains all files twice?

smira:
We need to think how we're going to take it to the Image Factory? Invalidate all previous SecureBoot ISOs?

frezbo:
> so the ISO contains all files twice?

yes, one as a EFI partition and another as just files in ISO (file system transposition). I've confirmed the same with an arch iso

pkg/imager/iso/uefi.go
frezbo: i'll use `os.MkdirTemp` to use a temp dir and cleanup afterwards,

frezbo:
> We need to think how we're going to take it to the Image Factory? Invalidate all previous SecureBoot ISOs?

I guess so, I just by luck it booted at most UEFI firmwares

pkg/imager/iso/uefi.go
smira: we could probably just create a directory structure under scratch dir

```
scratch_dir
  efiroot
     EFI
     loader
  efiboot.img
```

I think that would be more clean?

frezbo: :+1: makes more sense

smira (approved):

dsseng (approved):

PR #9583, opened 2024-10-28
feat: add parsing of vlanNNNN:ethX style VLAN cmdline args

### LINKED ISSUE #9552: Talos fails to parse `vlan` cmdline parameter if the interface uses a `VLAN_PLUS_VID ` form
### Description

According to the [dracut man page](https://man7.org/linux/man-pages/man7/dracut.cmdline.7.html), there are four different forms of VLAN names:
* VLAN_PLUS_VID (vlan0005)
* VLAN_PLUS_VID_NO_PAD (vlan5)
* DEV_PLUS_VID (eth0.0005)
* DEV_PLUS_VID_NO_PAD (eth0.5)

When parsing the parameter, [the parsing function](https://github.com/siderolabs/talos/blob/49bbadc4bf1e79e48c057d473ae21426b273c588/internal/app/machined/pkg/controllers/network/cmdline.go#L338) assumes that the field uses either the `DEV_PLUS_VID` form or the `DEV_PLUS_VID_NO_PAD` form.

Instead, the function should do the following:

```go
var vlanNumberString string
if strings.HasPrefix(vlanName, "vlan") {
    vlanNumberString = strings.TrimPrefix(vlanName, "vlan")
} else {
    _, vlanNumberString, ok := strings.Cut(vlanName, ".")
    if !ok {
        return settings, fmt.Errorf("malformed vlan commandline argument: %s", *vlanSettings)
    }
}
```

### Environment

- Talos version: v1.8.1
- Kubernetes version: N/A
- Platform: metal/amd64

jnohlgard:
## What? (description)

Fixes #9552 

Recognize the alternative form `vlan=vlan1:eth1` for specifying VLANs on the kernel command line. Additionally adds a check for the vlan ID range.
The created links will still use the format ethX.NNNN for the interface names, but at least this patch allows parsing both forms on the command line.

## Why? (reasoning)

The Talos docs at https://www.talos.dev/v1.8/reference/kernel/#vlan refers to the dracut command line docs which describe two alternative formats for `vlan=`.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/network/cmdline.go
+++ b/internal/app/machined/pkg/controllers/network/cmdline.go
@@ -337,10 +337,17 @@ func ParseCmdlineNetwork(cmdline *procfs.Cmdline) (CmdlineNetworking, error) {
 
 		_, vlanNumberString, ok := strings.Cut(vlanName, ".")
 		if !ok {
-			return settings, fmt.Errorf("malformed vlan commandline argument: %s", *vlanSettings)
+			vlanNumberString, ok = strings.CutPrefix(vlanName, "vlan")
+			if !ok {
+				return settings, fmt.Errorf("malformed vlan commandline argument: %s", *vlanSettings)
+			}
 		}
 
 		vlanID, err := strconv.Atoi(vlanNumberString)
+
+		if vlanID < 1 || 4095 < vlanID {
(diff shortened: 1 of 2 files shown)


smira (approved):

smira:
thank you!

PR #9591, opened 2024-10-28
fix: do not trim 0 from process SELinux label

dsseng:
Seems like a typo, zero is a valid character in SELinux labels

### DIFF
--- a/internal/pkg/miniprocfs/processes.go
+++ b/internal/pkg/miniprocfs/processes.go
@@ -164,7 +164,7 @@ func (procs *Processes) readProc(pidString string) (*machine.ProcessInfo, error)
 	var label string
 
 	if err = procs.readFileIntoBuf(path + "attr/current"); err == nil {
-		label = string(bytes.Trim(procs.buf, "\x000\n"))
+		label = string(bytes.Trim(procs.buf, "\x00\n"))
 	}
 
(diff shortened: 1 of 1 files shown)


internal/pkg/miniprocfs/processes.go
```diff
@@ -164,7 +164,7 @@ func (procs *Processes) readProc(pidString string) (*machine.ProcessInfo, error)
 	var label string
 
 	if err = procs.readFileIntoBuf(path + "attr/current"); err == nil {
-		label = string(bytes.Trim(procs.buf, "\x000\n"))
+		label = string(bytes.Trim(procs.buf, "\x00\n"))
```
smira, line 167: we need to strip it in the CLI then, as `\0` is not a valid output

smira: see #9564

dsseng: I strip `\x00` == `\0` in C convention. `\x000` is like `\x00` and `0`

frezbo: yeh, my mistake should have been just `\x00`

frezbo (approved):

smira (approved):

PR #9596, opened 2024-10-29
feat: support conditional start of IPv6 dns servers

DmitriyMV:
This PR does those things:
- [x] Refactored `DNSResolveCacheController`. Most of the logic moved to `dns` package types. Simplify and streamline logic.
- [x] Replace most of the goroutine orchestration with suture package.
- [x] Support per-item reaction to the dns listeners/servers failing to start. This allows us to ignore IPv6 errors if it's disabled.
- [x] Support per-item reaction to the dns listeners/servers failing to stop.
- [ ] Raise IPv6 listener on link-local address for dns (both TCP and UDP).
- [ ] Update kubelet's `resolv.conf` IPv4/IPv6 endpoints.


smira (approved):

binary-zero-one:
The IPv4 address belongs to the link-local 169.254.0.0/16, so why is ULA used for IPv6 instead of link-local fe80::/10 too?

Issue #9543, opened 2024-10-22
PoC runc overlayfs instead of memfd


frezbo:
## Feature Request

PoC on Talos to try and see if runc overlayfs would work with `/tmp` being full instead of `memfd-bind`.

See: https://github.com/opencontainers/runc/pull/4448

PR #9608, opened 2024-10-31
chore: drop runc memfd bind added in #9069

frezbo:
As of runc 1.2.0, it internally uses an `ovverlay` mount instead of cloning `/proc/self/exe` and
https://github.com/opencontainers/runc/pull/4482/files drops and the `DMZ` code. This also reduces code that Talos has to maintain.

Fixes: #9543


frezbo:
putting on hold since I've doubts, added a comment here: https://github.com/opencontainers/runc/pull/4482/files#r1824376587

smira:
@frezbo I wonder if we should disable it by default, but leave the controller around just in case if they change their mind :)

frezbo:
> @frezbo I wonder if we should disable it by default, but leave the controller around just in case if they change their mind :)

okay, I'll keep the controller around and not load it

PR #9127, opened 2024-08-07
feat: add necessary SELinux allow rules

dsseng:
# Pull Request

## What? (description)
Make everything work with SELinux

## Why? (reasoning)
Improve security, #8016 

## Acceptance

Please use the following checklist:

- [x] you linked an issue (if applicable)
- [x] you included tests (if applicable)
- [ ] you ran conformance (`make conformance`)
- [x] you formatted your code (`make fmt`)
- [x] you linted your code (`make lint`)
- [x] you generated documentation (`make docs`)
- [ ] you ran unit-tests (`make unit-tests`)

> See `make help` for a description of the available targets.

Build SELinux policy without full generate step (might produce different results due to reliance on host tooling, but is faster):
```bash
secilc -o ./internal/pkg/selinux/policy/policy.33 -f ./internal/pkg/selinux/policy/file_contexts -c 33 ./internal/pkg/selinux/policy/selinux/**/*.cil -vvvvv -O
```

Read and analyze SELinux denials (`make generate` or alternative build required to match against the current policy):
```bash
talosctl --nodes 172.20.0.2 logs auditd > /tmp/audit.log && talosctl --nodes 172.20.0.3 logs auditd >> /tmp/audit.log && PYTHONPATH=/tmp/selinux/python/sepolgen/src python3 /tmp/selinux/python/audit2allow/audit2allow -C -p ./internal/pkg/selinux/policy/policy.33 -i /tmp/audit.log > /tmp/audit2allow.cil && PYTHONPATH=/tmp/selinux/python/sepolgen/src python3 /tmp/selinux/python/audit2allow/audit2allow -C -p ./internal/pkg/selinux/policy/policy.33 -e -i /tmp/audit.log > /tmp/audit2allow-explain.cil
```

~~Tip: to debug denials in QEMU:~~
```bash
[...]

PR #9617, opened 2024-10-31
feat: machined: initial SELinux bring-up

dsseng:
Part of: #9127

Label executables and processes, build, load and manage SELinux policy, enable audit support.

Labeling filesystems, devices and runtime files will be done in further changes, see the full PR.

TODO: label static pods


dsseng (commented):
There is some potential to further reduce the amount of CIL code for the initial bring-up. I might try to do so. Also please take notes if anything from here needs to be moved into further commits from #9127

dsseng:
```
; audit(1730388431.799:429):
;  scontext="system_u:system_r:sys_containerd_t:s0" tcontext="system_u:object_r:init_exec_t:s0"
;  class="file" perms="execute"
;  comm="runc:[2:INIT]" exe="" path=""
;  message="[    4.376219] audit: type=1400 audit(1730388431.799:429): avc:
;   denied  { execute } for  pid=1948 comm="runc:[2:INIT]" name="dashboard"
;   dev="loop0" ino=503 scontext=system_u:system_r:sys_containerd_t:s0
;   tcontext=system_u:object_r:init_exec_t:s0 tclass=file permissive=1 "
```

selinux/services/udev.cil
```diff
@@ -0,0 +1,33 @@
...
+
+; execute is standard for systemd-udevd, execute_no_trans for udevadm called in init scope
+(allow init_t udev_exec_t (file (execute_no_trans)))
+
+; cap-launcher
+(allow udev_t udev_exec_t (file (entrypoint execute_no_trans)))
+
+; TODO: label /run
+
+; modprobe called by kernel or udev
```
frezbo, line 28: `vgchange` can also modprobe

Dockerfile
```diff
@@ -716,6 +722,7 @@ RUN <<END
     ln -s /etc/ssl /rootfs/usr/local/share/ca-certificates
     ln -s /etc/ssl /rootfs/etc/ca-certificates
 END
+RUN mkdir -p /rootfs/selinux
```
smira, line 725: let's unify this with other mkdirs, see line 703-707

Dockerfile
```diff
@@ -375,6 +375,10 @@ COPY --from=embed-abbrev-generate /src/pkg/machinery/gendata/data /pkg/machinery
 COPY --from=embed-abbrev-generate /src/_out/talos-metadata /_out/talos-metadata
 COPY --from=embed-abbrev-generate /src/_out/signing_key.x509 /_out/signing_key.x509
 
+FROM tools as selinux
```
smira, line 378: Docker complains about mixed casing:

```suggestion
FROM tools AS selinux
```

internal/pkg/install/install.go
```diff
@@ -204,7 +205,7 @@ func RunInstallerContainer(disk, platform, ref string, cfg configcore.Config, cf
 		oci.WithReadonlyPaths(nil),
 		oci.WithWriteableSysfs,
 		oci.WithWriteableCgroupfs,
-		oci.WithSelinuxLabel(""),
+		oci.WithSelinuxLabel("system_u:system_r:installer_t:s0"),
```
smira, line 208: should it be a constant?

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -104,6 +119,40 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
...
+	}
+
+	log.Println("SELinux policy loaded")
+
+	if err = os.WriteFile("/proc/thread-self/attr/exec", []byte("system_u:system_r:init_t:s0"), 0o777); err != nil {
+		return err
+	}
+
+	if selinuxMode == "enforcing" {
+		log.Println("Setting SELinux mode to enforcing")
```
smira, line 149: nit: use lowercase logs
```suggestion
		log.Println("setting SELinux mode to enforcing")
```

internal/pkg/mount/v2/pseudo.go
```diff
@@ -21,6 +21,9 @@ func Pseudo() Points {
 		NewPoint("tmpfs", "/run", "tmpfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_RELATIME), WithData("mode=0755")),
 		NewPoint("tmpfs", "/system", "tmpfs", WithData("mode=0755")),
 		NewPoint("tmpfs", "/tmp", "tmpfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV), WithData("size=64M"), WithData("mode=0755")),
+		// TODO: mount this under /sys/fs/selinux as in most distros?
+		// That'd be harder to move when we switch root
```
smira, line 25: good question, I wasn't clear if `Move` moves correctly sub-mounts, so if we move `/sys`, would it move `/sys/fs/selinux` as well?

If it does, we can probably simply move selinux mount to a different set of points which we mount, but don't move

internal/pkg/selinux/selinux.go
```diff
@@ -0,0 +1,28 @@
...
+// otherwise it returns false. It also ensures we're not in a container.
+func IsEnabled() bool {
+	// TODO: resolve circular dependency with platform
+	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
+		return false
+	}
+
+	val := procfs.ProcCmdline().Get(constants.KernelParamSELinux).First()
+
+	return val != nil
```
smira, line 27: do we accept any value? (just the arg?

selinux/services/udev.cil
dsseng: Unsure, maybe I will do policies for it later. Or we just let it `execute_no_trans` and run in init's domain.

Dockerfile
```diff
@@ -375,6 +375,10 @@ COPY --from=embed-abbrev-generate /src/pkg/machinery/gendata/data /pkg/machinery
...
+COPY ./selinux /selinux
```
frezbo, line 379: what path does standard distros use?

dsseng: In etc

internal/pkg/mount/v2/pseudo.go
```diff
@@ -21,6 +21,9 @@ func Pseudo() Points {
...
+		NewPoint("selinuxfs", "/selinux", "selinuxfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_RELATIME)),
```
frezbo, line 26: yes, let's use standard paths as tools like cilium/csi's might try to mount this whem selinux is enabled to gather info

.gitignore
```diff
@@ -39,3 +39,7 @@ node_modules
 website/resources
 website/public
 website/.hugo_build.lock
+
+# Files built for SELinux troubleshooting
+policy.33
+file_contexts
```
smira, line 45: how do they end up in working directory?

dsseng: I build policy locally for `audit2allow` purposes. They aren't created by build system normally, but I use those for development.

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -28,6 +30,9 @@ var preservedPaths = map[string]struct{}{
 	constants.SDStubDynamicInitrdPath: {},
 }
 
+//go:embed policy.33
```
smira, line 33: I think with this file we need to follow common `generate` approach - that is we should have it generated, put back to the source and committed to the repository.

Otherwise Talos won't be buildable with `go` tools outside of buildkit

.gitignore
frezbo: let's put those under `hack` then if we ever want to debug

hack/labeled-squashfs.sh
```diff
@@ -0,0 +1,6 @@
+#!/toolchain/bin/bash
+
+set -e
+# set SELinux labels for files according to file_contexts supplied
+/toolchain/sbin/setfiles -r $1 -F -vv $3 $1
```
smira, line 5: would this adjust the modified time that we want to preserve? do even need this to be a shell script? can we instead use `RUN <<END` ?

pkg/machinery/kernel/kernel.go
```diff
@@ -31,6 +31,14 @@ var DefaultArgs = []string{
 	"ima_template=ima-ng",
 	"ima_appraise=fix",
 	"ima_hash=sha512",
+	// TODO: configure it properly
+	"talos.selinux=permissive",
```
smira, line 35: this is a default set of args which applies to any version of Talos, we should not probably modify it that aggressively.

we can modify per-test if we want to, but probably we should split it per-version

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -104,6 +119,40 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
...
+	log.Println("SELinux policy loaded")
+
+	if err = os.WriteFile("/proc/thread-self/attr/exec", []byte("system_u:system_r:init_t:s0"), 0o777); err != nil {
+		return err
+	}
+
+	if selinuxMode == "enforcing" {
+		log.Println("Setting SELinux mode to enforcing")
+
+		if err = os.WriteFile("/selinux/enforce", []byte("1"), 0o777); err != nil {
```
dsseng, line 151: TODO: test this (should crash right after enforce)

internal/pkg/mount/switchroot/switchroot.go
smira: "selinux/policy.33"

smira: move to `internal/pkg/selinux`, embed as a variable, have a function which returns the value (policy)

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -104,6 +119,40 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
...
+
+	var err error
+
+	if err = os.WriteFile("/selinux/load", selinuxPolicy, 0o777); err != nil {
+		return err
+	}
+
+	log.Println("SELinux policy loaded")
+
+	if err = os.WriteFile("/proc/thread-self/attr/exec", []byte("system_u:system_r:init_t:s0"), 0o777); err != nil {
```
smira, line 144: label should be a constant?

internal/pkg/mount/v2/pseudo.go
smira: I will send a PR to fix this

pkg/machinery/kernel/kernel.go
smira, line 35: should be only for 1.9 (see `cmdline` in the imager)

pkg/machinery/kernel/kernel.go
```diff
@@ -31,6 +31,14 @@ var DefaultArgs = []string{
...
+	"audit=1",
```
smira, line 36: drop all these audit/debugging args from the defaults

pkg/machinery/kernel/kernel.go
frezbo, line 35: audit can probably use these libraries:

* https://github.com/elastic/go-libaudit
* https://github.com/slackhq/go-audit

smira:
https://github.com/elastic/go-libaudit

.gitignore
smira: we generate & commit those files

pkg/machinery/kernel/kernel.go
frezbo: we can probably use this https://github.com/elastic/go-libaudit/blob/main/audit.go#L119 by passing in netlink from https://pkg.go.dev/github.com/mdlayher/netlink#example-Conn-ListenMulticast

dsseng: We could do something like `security=selinux selinux=1 enforcing=1` to match common distros (this is cmdline fragment from SUSE). Maybe some tools care about this

frezbo: https://www.kernel.org/doc/html/v6.11/admin-guide/kernel-parameters.html seems it's all standard kernel parameters

Dockerfile
dsseng: Will remove as we mount nothing here now.

pkg/machinery/kernel/kernel.go
dsseng: Yes, and also collecting support bundles in CI would be much better than https://github.com/siderolabs/talos/pull/9127/commits/0ac02dc9b6ea2ead19d4ef86648ce323e40f9464

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -625,6 +625,7 @@ func WriteUdevRules(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
 			content.WriteByte('\n')
 		}
 
+		// TODO: SELinux label?
```
dsseng, line 628: We should perhaps label our custom rules as well. I thought labels for rules will come later, but will do now

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -104,6 +119,40 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
 		return fmt.Errorf("error executing /sbin/init: %w", err)
 	}
 
+	runtime.UnlockOSThread()
+
+	return nil
+}
+
+func initSelinux(selinuxMode string) error {
```
dsseng, line 127: Move to the selinux package, together with policy.33

selinux/services/udev.cil
```diff
@@ -0,0 +1,33 @@
...
+(context udev_exec_t (system_u object_r udev_exec_t (systemLow systemLow)))
+(filecon "/sbin/systemd-udevd" file udev_exec_t)
+(filecon "/usr/bin/udevadm" file udev_exec_t)
+(filecon "/sbin/udevadm" symlink udev_exec_t)
+
+; Do not reorder: label non-executable rules files as executable
+(type udev_rules_t)
+(call system_f (udev_rules_t))
+(filecon "/usr/lib/udev/rules.d(/.*)?" any (system_u object_r udev_rules_t (systemLow systemLow)))
+; TODO: separate label to be called by udev only?
```
dsseng, line 12: Or maybe better just allow udev executing itself and that's fine. No actual improvement in case we add more contexts.

selinux/common/typeattributes.cil
```diff
@@ -0,0 +1,106 @@
...
+(typeattributeset any_p system_service_p)
+(typeattributeset any_p system_container_p)
+(typeattributeset any_p client_p)
+(typeattributeset any_p pod_p)
+
+(typeattribute system_p)
+(typeattributeset system_p kernel_t)
+(typeattributeset system_p init_t)
+(typeattributeset system_p system_service_p)
+(typeattributeset system_p client_p)
```
dsseng, line 102: Some stuff like `client_p` doesn't make much sense to be separate from `system_service_p`, so perhaps the number of typeattributes could be reduced to ease understanding, visualization and possibly reduce policy size and overhead

selinux/common/typeattributes.cil
```diff
@@ -0,0 +1,106 @@
...
+(typeattributeset any_f common_f)
+(typeattributeset any_f protected_f)
+(typeattributeset any_f system_f)
+(typeattributeset any_f system_socket_f)
+(typeattributeset any_f device_f)
+
+(typeattribute unconfined_f)
+(typeattributeset unconfined_f tmpfs_t)
+(typeattributeset unconfined_f common_f)
+(typeattributeset unconfined_f common_device_f)
```
dsseng, line 55: File attributes could also benefit from less aggregation

internal/pkg/mount/switchroot/switchroot.go
dsseng: tested

pkg/machinery/kernel/kernel.go
dsseng: Even without audit options I get audit logs in dmesg. Should I explicitly disable audit for now then, until I redirect it?

dsseng: `[    0.014439] Unknown kernel command line parameters "pti=on selinux=1", will be passed to user space.`

selinux/services/udev.cil
dsseng: Implemented it in the way misc sbin things can be ran by machined, and run in init context. We should also take care of the cases when udev runs some binaries not stored in its rules directory

selinux/services/machined.cil
```diff
@@ -0,0 +1,33 @@
...
+; Should not occur unless misconfigured by machined
+(type unconfined_service_t)
+(call system_service_p (unconfined_service_t))
+
+(type sbin_exec_t)
+(call system_f (sbin_exec_t))
+(filecon "/sbin(/.*)?" any (system_u object_r sbin_exec_t (systemLow systemLow)))
+; Typically machined executes LVM, cryptsetup and similar utilities
+; They are short-running, come from the rootfs and do not accept user input, so can be started in init_t domain
+(allow init_t sbin_exec_t (file (execute execute_no_trans)))
```
dsseng, line 33: Does machined execute anything apart from sbin things? Does udev execute anything other than modprobe and stuff like `scsi_id`?

internal/pkg/selinux/selinux.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	log.Println("selinux: policy loaded")
+
+	if IsEnforcing() {
+		log.Println("selinux: setting mode to enforcing")
+
+		if err := os.WriteFile("/sys/fs/selinux/enforce", []byte("1"), 0o777); err != nil {
+			return err
+		}
```
dsseng, line 88: From kernel docs, we could probably avoid doing this manually. But also I should check what happens if kernel enables enforce before we load policy. Looks weird.

```
        enforcing=      [SELINUX] Set initial enforcing status.
                        Format: {"0" | "1"}
                        See security/selinux/Kconfig help text.
                        0 -- permissive (log only, no denials).
                        1 -- enforcing (deny and log).
                        Default value is 0.
                        Value can be changed at runtime via
                        /sys/fs/selinux/enforce.
```

pkg/machinery/constants/constants.go
```diff
@@ -672,7 +672,7 @@ const (
 	// SystemOverlaysPath is the path to the system overlay directory.
 	SystemOverlaysPath = SystemPath + "/overlays"
 
-	// RuncMemfdLabel is the SELinux label for runc executable memfd
+	// RuncMemfdLabel is the SELinux label for runc executable memfd.
```
frezbo, line 675: we can remove this after #9608

selinux/services/udev.cil
frezbo: yeh, udev rules can execute binaries on host and if we block that it might break tools

pkg/machinery/kernel/kernel.go
frezbo:
```
audit=          [KNL] Enable the audit sub-system
                        Format: { "0" | "1" | "off" | "on" }
                        0 | off - kernel audit is disabled and can not be
                            enabled until the next reboot
                        unset - kernel audit is initialized but disabled and
                            will be fully enabled by the userspace auditd.
                        1 | on - kernel audit is initialized and partially
                            enabled, storing at most audit_backlog_limit
                            messages in RAM until it is fully enabled by the
                            userspace auditd.
                        Default: unset
```

I think dmesg will always log that, this is probably to control enabling the netlink socket maybe?

selinux/services/machined.cil
frezbo: > Does machined execute anything apart from sbin things?

I think no, also does `kernel_t` has permissions to execute binaries under `/sbin`, I know in azure, kernel actually calls `/sbin/shutdown`

> Does udev execute anything other than modprobe and stuff like scsi_id?

this depends on what's in the udev rules

internal/pkg/selinux/selinux.go
frezbo: we should check what happens if `enforcing=1` is set and booted

pkg/machinery/constants/constants.go
dsseng: Yes, once rebased will drop

selinux/services/udev.cil
dsseng: We need to figure out which binaries

selinux/services/machined.cil
dsseng: > kernel actually calls /sbin/shutdown

Perhaps I already allowed kernel_t to execute init binary

internal/pkg/selinux/selinux.go
dsseng: Yes, with other commits from #9127 rebased on top, since now it fails to read anything

smira (approved):

selinux/services/machined.cil
frezbo: yeh, but this is after switchroot happens :thinking:

internal/pkg/selinux/selinux.go
frezbo: ahh probably it's read only if the cmdline is set

selinux/services/machined.cil
dsseng: Will take a look at kernel source

internal/pkg/selinux/selinux.go
dsseng: Well, we should probably deny setenforce, so that enforcement could not be lifted after enforcing in either way

internal/pkg/selinux/selinux.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Ensure we run in the same task to transition the init process into proper context
+	runtime.LockOSThread()
+
+	if err := os.WriteFile("/proc/thread-self/attr/exec", []byte(constants.SelinuxLabelMachined), 0o777); err != nil {
+		return err
+	}
+
+	return nil
```
frezbo, line 98: should we unlock the os thread here, or it's okay to do in init?

frezbo (approved):

internal/pkg/selinux/selinux.go
dsseng: It's certainly not ok here, since init then needs to exec from the same thread. Maybe locking should be better moved into switchroot.

pkg/machinery/kernel/kernel.go
dsseng: fixed by auditd PR

selinux/services/machined.cil
dsseng: `/sbin/poweroff` and `/sbin/reboot` can be called by kernel, yes

dsseng: That might need a typetransition: actually we could avoid transitioning in switchroot but rather make kernel do transition for us

internal/pkg/selinux/selinux.go
dsseng: Dropped as I moved to typetransition'ing kernel_t -> init_t when executing init_exec_t

internal/pkg/install/install.go
```diff
@@ -184,6 +184,7 @@ func RunInstallerContainer(disk, platform, ref string, cfg configcore.Config, cf
 		constants.KernelParamEquinixMetalEvents,
 		constants.KernelParamDashboardDisabled,
 		constants.KernelParamNetIfnames,
+		constants.KernelParamSELinux,
```
dsseng, line 187: need to preserve enforce as well

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -72,6 +73,10 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
 		return fmt.Errorf("error deleting initramfs: %w", err)
 	}
 
+	if err := selinux.Init(); err != nil {
+		return err
+	}
```
dsseng, line 78: We could potentially initialize it in first stages of machined as well. Let's consider pros and cons

internal/pkg/selinux/selinux.go
```diff
@@ -0,0 +1,91 @@
...
+
+//go:embed policy/policy.33
+var policy []byte
+
+// IsEnabled checks if SELinux is enabled on the system by reading
+// the kernel command line. It returns true if SELinux is enabled,
+// otherwise it returns false. It also ensures we're not in a container.
+// By default SELinux is disabled.
+func IsEnabled() bool {
+	// TODO: resolve circular dependency with platform
```
dsseng, line 27: Remove this todo, it's enough to ensure we're not in container

pkg/machinery/constants/constants.go
```diff
@@ -696,6 +702,21 @@ const (
 	// CgroupSystemRuntimeMillicores is the CPU weight for the system containerd process.
 	CgroupSystemRuntimeMillicores = 500
 
+	// SelinuxLabelMachined is the SELinux label for machined.
+	SelinuxLabelMachined = "system_u:system_r:init_t:s0"
```
dsseng, line 706: Only used in integration test now

pkg/provision/providers/qemu/node.go
```diff
@@ -84,6 +84,10 @@ func (p *provisioner) createNode(state *vm.State, clusterReq provision.ClusterRe
 	// Talos config
 	cmdline.Append("talos.platform", constants.PlatformMetal)
 
+	// SELinux
+	cmdline.Append("selinux", "1")
+	cmdline.Append("enforcing", "0")
```
dsseng, line 89: Maybe we should add options to talosctl for this?

selinux/common/network.cil
```diff
@@ -0,0 +1,73 @@
...
+(allow any_p any_p (smc_socket (accept append bind connect create getattr getopt ioctl listen lock map name_bind read recvfrom sendto setattr setopt shutdown write)))
+(allow any_p any_p (xdp_socket (accept append bind connect create getattr getopt ioctl listen lock map name_bind read recvfrom sendto setattr setopt shutdown write)))
+
+; UDS within any domain is always allowed
+(allow any_p self (unix_stream_socket (accept append bind connect create getattr getopt ioctl listen lock map name_bind read recvfrom sendto setattr setopt shutdown write)))
+(allow any_p self (unix_dgram_socket (accept append bind connect create getattr getopt ioctl listen lock map name_bind read recvfrom sendto setattr setopt shutdown write)))
+
+; Netlink
+(allow any_p self (netlink_classes (full)))
+(allow system_p any_p (netlink_classes (full)))
```
dsseng, line 65: Need to see how other distros handle selinux and audit netlink sockets, for example. Maybe it's not OK to make those available to any process

selinux/services/dashboard.cil
```diff
@@ -0,0 +1,4 @@
+(type dashboard_t)
+(call system_service_p (dashboard_t))
+(allow init_t init_exec_t (file (execute)))
```
dsseng, line 3: We should probably make `system_service_p` and `client_p` accept the executable label as well, so that we allow init to execute that directly in a macro.

selinux/services/machined.cil
```diff
@@ -0,0 +1,32 @@
...
+
+(type init_t)
+(roletype system_r init_t)
+(typetransition kernel_t init_exec_t process init_t)
+(allow init_t init_exec_t (file (execute entrypoint)))
+
+; we create a sealed runc memfd to be shared
+; by all containerd instances and containers
+(type runc_memfd_t)
+(call common_f (runc_memfd_t))
```
dsseng, line 17: do sth like `tmp_f`?

selinux/services/selinux.cil
```diff
@@ -0,0 +1,14 @@
+(allow kernel_t security_t (security (load_policy setbool setenforce setsecparam)))
+(allow init_t security_t (security (load_policy setbool setenforce setsecparam)))
```
dsseng, line 2: if we end up using `enforcing` kernel parameter, we're not going to use setenforce from init process, so it must be always denied for all so that enforce could never be removed in runtime.

selinux/services/system-containerd.cil
```diff
@@ -0,0 +1,30 @@
...
+(call system_f (containerd_exec_t))
+(context containerd_exec_t (system_u object_r containerd_exec_t (systemLow systemLow)))
+(filecon "/bin/containerd" any containerd_exec_t)
+(filecon "/bin/containerd-shim-runc-v2" any containerd_exec_t)
+(filecon "/bin/runc" any containerd_exec_t)
+
+; System containerd
+(type sys_containerd_t)
+(call system_service_p (sys_containerd_t))
+(allow init_t containerd_exec_t (file (execute)))
```
dsseng, line 11: Also needs to by in macro for `system_service_p`

selinux/services/system-containerd.cil
```diff
@@ -0,0 +1,30 @@
...
+(allow sys_containerd_t containerd_exec_t (file (entrypoint execute execute_no_trans)))
+
+; TODO: label /system
+
+; sealed runc memfd created by machined
+(allow sys_containerd_t runc_memfd_t (file (execute execute_no_trans)))
```
dsseng, line 17: remove memfd and replace?

selinux/services/cri.cil
```diff
@@ -0,0 +1,23 @@
...
+(type pod_containerd_t)
+(call client_p (pod_containerd_t))
+(allow init_t containerd_exec_t (file (execute)))
+(allow pod_containerd_t containerd_exec_t (file (entrypoint execute execute_no_trans)))
+(allow init_t pod_containerd_t (process (transition)))
+
+; TODO: label /run
+
+; sealed runc memfd created by machined
+(allow pod_containerd_t runc_memfd_t (file (execute execute_no_trans)))
```
dsseng, line 11: remove memfd and replace?

selinux/services/system-containerd.cil
frezbo: yes, don't use anymore

selinux/services/cri.cil
frezbo: same

selinux/services/machined.cil
frezbo: this can be removed i guess

selinux/services/selinux.cil
dsseng: yes, `enforcing=1` will enforce even before init tries to setenforce. Thus I'll deny setenforce and load_policy for everyone, as we let kernel enforce it itself

internal/pkg/mount/switchroot/switchroot.go
dsseng: As we use `typetransition` now, then not, because we need machined to already be in the correct context.

pkg/imager/imager.go
```diff
@@ -349,6 +349,12 @@ func (i *Imager) buildCmdline() error {
 		cmdline.Append("console", "tty0")
 	}
 
+	if quirks.New(i.prof.Version).SupportsSELinux() {
+		// Talos 1.9 introduces SELinux in permissive mode
+		cmdline.Append("selinux", "1")
+		cmdline.Append("enforcing", "0")
```
dsseng, line 355: How do we handle the next update where we're going to set `enforcing=1`? Will we be able to ask installer to override `enforcing`? Or maybe we should omit enforcing option for now at all, as kernel default is permissive

pkg/provision/providers/qemu/node.go
smira: we already have `--extra-boot-kernel-args`, but for now let's keep it this way, we can re-consider later

pkg/imager/imager.go
smira: if we can omit, let's omit for now.

PR #9618, opened 2024-10-31
fix: mount pseudo sub-mountpoints in init

smira:
Mount them early, they will be moved to the new root in switch root process as they're under.


dsseng (approved):

frezbo (approved):

dsseng (approved):
Works perfectly for selinuxfs

PR #9640, opened 2024-11-05
feat: label system socket and runtime files

dsseng:
Set SELinux labels so that services could gain access permissions.


smira (approved):

internal/app/apid/main.go
```diff
@@ -157,6 +158,10 @@ func apidMain() error {
 		return fmt.Errorf("error creating listner: %w", err)
 	}
 
+	if err = selinux.SetLabel(constants.APISocketPath, "system_u:object_r:apid_socket_t:s0"); err != nil {
```
frezbo, line 161: constant maybe?

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -96,6 +97,10 @@ func (o *APID) PreFunc(ctx context.Context, r runtime.Runtime) error {
 		return err
 	}
 
+	if err := selinux.SetLabel(constants.APIRuntimeSocketPath, "system_u:object_r:apid_runtime_socket_t:s0"); err != nil {
```
dsseng, line 100: Constant

internal/pkg/logind/broker.go
```diff
@@ -48,11 +50,19 @@ func NewBroker(serviceSocketPath, clientSocketPath string) (*DBusBroker, error)
 		return nil, err
 	}
 
+	if err = selinux.SetLabel(serviceSocketPath, "system_u:object_r:dbus_service_socket_t:s0"); err != nil {
```
frezbo, line 53: should we do similar like standard distro, `system_dbus_var_run_t` :thinking:

internal/pkg/selinux/policy/selinux/services/cri.cil
```diff
@@ -15,3 +15,30 @@
 (call pod_p (etcd_t))
 ; FIXME: insecure as anyone with access to the pod containerd may obtain this domain (executable in ephemeral)
```
frezbo, line 16: is the comment still valid?

dsseng: Yes. Ephemeral label actually comes later in the series, and even after then we need to think how to label kubelet/etcd executables and assign labels to static pods

internal/pkg/logind/broker.go
dsseng: Unsure. No reason to either do as in reference policy nor do it differently. We can do as in ref though

internal/pkg/selinux/policy/selinux/services/cri.cil
frezbo: okay, let's keep what we have then, unless we figure we need to relabel it later

internal/pkg/logind/broker.go
frezbo: let keep what you have then.

internal/integration/api/selinux.go
```diff
@@ -64,6 +67,92 @@ func (suite *SELinuxSuite) getLabel(nodeCtx context.Context, pid int32) string {
...
+	suite.checkFileLabels(workers, expectedLabelsWorker)
+	suite.checkFileLabels(controlplanes, expectedLabelsControlPlane)
+}
+
+func (suite *SELinuxSuite) checkFileLabels(nodes []string, expectedLabels map[string]string) {
+	for _, node := range nodes {
+		nodeCtx := client.WithNode(suite.ctx, node)
+		cmdline := suite.ReadCmdline(nodeCtx)
+
+		seLinuxEnabled := pointer.SafeDeref(procfs.NewCmdline(cmdline).Get(constants.KernelParamSELinux).First()) != ""
```
frezbo, line 111: this can be moved in setuptest so we can skip earlier

frezbo (approved):

internal/integration/api/selinux.go
dsseng: I do it here to check per-node, as we might have different cmdline on each node

PR #9643, opened 2024-11-05
chore: generate support zip for crashdump

frezbo:
Generate support zip on crashdump.

### DIFF
--- a/internal/integration/provision/provision.go
+++ b/internal/integration/provision/provision.go
@@ -142,10 +142,6 @@ func (suite *BaseSuite) TearDownSuite() {
 		// for failed tests, produce crash dump for easier debugging,
 		// as cluster is going to be torn down below
 		suite.provisioner.CrashDump(suite.ctx, suite.Cluster, os.Stderr)
-
-		if suite.clusterAccess != nil {
-			suite.clusterAccess.CrashDump(suite.ctx, os.Stderr)
-		}
 	}
 
 	if suite.clusterAccess != nil {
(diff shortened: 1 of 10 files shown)


internal/integration/provision/provision.go
```diff
@@ -142,10 +142,6 @@ func (suite *BaseSuite) TearDownSuite() {
 		// for failed tests, produce crash dump for easier debugging,
 		// as cluster is going to be torn down below
 		suite.provisioner.CrashDump(suite.ctx, suite.Cluster, os.Stderr)
-
```
frezbo, line 145: now sure if this needed to be dropped, this just seemed to gather the service logs and print to stdout, support zip  seemed a better option

internal/integration/provision/provision.go
```diff
@@ -142,10 +142,6 @@ func (suite *BaseSuite) TearDownSuite() {
 		// for failed tests, produce crash dump for easier debugging,
 		// as cluster is going to be torn down below
 		suite.provisioner.CrashDump(suite.ctx, suite.Cluster, os.Stderr)
```
frezbo, line 144: can update this also to save to a file

smira (approved):
lgtm, I think Dmitry's goal is to have this even for non-failed builds

frezbo:
> lgtm, I think Dmitry's goal is to have this even for non-failed builds

we could have another flag for intergration and cluster create that does this unconditionally

dsseng (approved):
Yes, my goal is to save it anyway, as for now we're testing in permissive mode and it'd be useful to be able to download audit logs anyway

PR #9650, opened 2024-11-06
fix: selinux handling and apparmor tests

frezbo:
Conditionally mount selinuxfs only if it's present.

Fix AppArmor tests, `apparmor` and other minor LSM's and set
`apparmor=1`.

### DIFF
--- a/internal/pkg/mount/v2/pseudo.go
+++ b/internal/pkg/mount/v2/pseudo.go
@@ -33,13 +33,19 @@ func PseudoSubMountPoints() Points {
 		NewPoint("bpf", "/sys/fs/bpf", "bpf"),
 		NewPoint("securityfs", "/sys/kernel/security", "securityfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV|unix.MS_RELATIME)),
 		NewPoint("tracefs", "/sys/kernel/tracing", "tracefs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV)),
-		NewPoint("selinuxfs", "/sys/fs/selinux", "selinuxfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_RELATIME)),
 	}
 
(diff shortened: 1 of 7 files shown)


internal/pkg/mount/v2/pseudo.go
```diff
@@ -43,5 +42,12 @@ func PseudoSubMountPoints() Points {
 		)
 	}
 
+	if _, err := os.Stat("/sys/fs/selinux"); err == nil {
+		// mount selinuxfs if it exists
```
dsseng, line 46: In which cases it's not present? When using AppArmor as the major LSM?

pkg/provision/providers/qemu/node.go
```diff
@@ -84,9 +84,6 @@ func (p *provisioner) createNode(state *vm.State, clusterReq provision.ClusterRe
 	// Talos config
 	cmdline.Append("talos.platform", constants.PlatformMetal)
 
-	// SELinux
-	cmdline.Append("selinux", "1")
-
 	// add overrides
 	if nodeReq.ExtraKernelArgs != nil {
 		if err = cmdline.AppendAll(nodeReq.ExtraKernelArgs.Strings()); err != nil {
```
smira, line 89: can we instead use negative merging here and pass `-selinux` in case of apparmor test?

smira: I would rather leave `selinux=1` enabled by default for now

internal/pkg/mount/v2/pseudo.go
frezbo: yes, when lsm  is apparmor

pkg/provision/providers/qemu/node.go
frezbo: so check if we pass `apparmor=1`, then add `-selinux`?

frezbo: actually I got it, better way

smira: I mean in the `e2e-qemu.sh` directly if apparmor was requrest

frezbo: yeh, did the same

dsseng (approved):

PR #9658, opened 2024-11-07
feat: check bridged interfaces should not have addresses

jnohlgard:
## What? (description)

Add a similar check for bridged interfaces to ensure that they do not have addresses configured. The implementation combines the checks for bonded and bridged interfaces to avoid code duplication. Error messages have been slightly rephrased to match the changed scope.

Additionally fixes the test case implementation for addresses on bonded interfaces to include eth2 which was configured but not actually used in the test.

## Why? (reasoning)

Bridged interfaces should not have IP addresses, the addresses should be configured on the parent bridge instead.

## Acceptance


smira (approved):
thank you!

PR #9659, opened 2024-11-07
docs: add apparmor enablement release notes

frezbo:
Add AppArmor release notes.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -27,6 +27,22 @@ runc: 1.2.1
 Talos is built with Go 1.23.2.
 """
 
+    [notes.apparmor]
+        title = "AppArmor"
+        description = """\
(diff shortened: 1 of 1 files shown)


hack/release.toml
```diff
@@ -25,6 +25,21 @@ Kubernetes: 1.32.0-beta.0
...
+        title = "AppArmor"
+        description = """\
+Talos Linux starting with v1.9 will ship with SELinux LSM enabled by default.
+If you need to use AppArmor LSM add the following to the machine configuration:
+
+```yaml
+machine:
+  install:
+     extraKernelArgs:
+      - lsm=lockdown,capability,yama,apparmor,bpf
```
smira, line 40: do we need `-selinux` here?

smira (approved):

hack/release.toml
frezbo: probably yes

Issue #8554, opened 2024-04-05
Support for pod user namespaces


Piccirello:
This issue is to track Talos's support for user namespaces[^0] in Kubernetes pods. User namespaces allow for strict separation between the root user in pods and the root user on the host. From the docs: "A process running as root in a container can run as a different (non-root) user in the host; in other words, the process has full privileges for operations inside the user namespace, but is unprivileged for operations outside the namespace."

User namespaces requires at least Linux 6.3, which it appears Talos v1.7.0 will support. The Kubernetes docs also state that "containerd v1.7 is not compatible with the userns support in Kubernetes v1.27 to v1.29." That _may_ mean waiting for containerd 2.0[^1], though this is unclear to me.

When user namespaces are eventually supported, it would be worth mentioning as a feature in the Talos release's changelog.

[^0]: https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/
[^1]: https://github.com/containerd/containerd/releases/tag/v2.0.0-rc.0

sanmai-NL:
How will this affect Talos Linux nodes running inside containers? And potentially, in user-namespaced/rootless containers?

Piccirello:
Based on #8766, #8777, and #8484, it appears Talos 1.8 will use containerd 2.0. That _may_ mean that pod user namespaces will be supported in Talos 1.8.

smira:
Yes, it should be ready for that, I believe there's nothing to be done on Talos OS side itself to support that. If you have a good testcase for user namespaces (e.g. something you can `kubectl apply`), we'd be happy to get it into the integration tests. Thanks!

Piccirello:
It appears that there are two feature gates that need to be enabled: `UserNamespacesSupport` and `UserNamespacesPodSecurityStandards`. Both of these [currently](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) default to `false`, with the latter's effect described [here](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/127-user-namespaces/README.md#pod-security-standards-pss-integration).

[...]

PR #9662, opened 2024-11-07
docs: add docs on usernamespace support in k8s

frezbo:
Add docs and test for usernamespaces support in Kubernetes.

Fixes: #8554

### DIFF
--- a/internal/integration/k8s/usernamespace.go
+++ b/internal/integration/k8s/usernamespace.go
@@ -1,0 +1,128 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_k8s
+
+package k8s
+
+import (
+	"bufio"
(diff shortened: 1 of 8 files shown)


internal/integration/k8s/usernamespace.go
```diff
@@ -0,0 +1,109 @@
...
+
+	reader, err = suite.Client.Read(nodeCtx, fmt.Sprintf("/proc/%d/status", sleepProcessPID))
+	suite.Require().NoError(err)
+
+	var processStatus bytes.Buffer
+
+	_, err = processStatus.ReadFrom(reader)
+	suite.Require().NoError(err)
+
+	suite.Require().NotContains(processStatus.String(), "Uid:	0	0	0	0", "sleep process should not have root UID in host namespace")
```
smira, line 104: this might be a bit weak check, I think we should better find a line with `Uid:`, split it up and look that first field is not 0 ?

Otherwise if the kernel format slightly changes, the check would pass, while it might be still uid 0

dsseng (approved):

smira (approved):

PR #9667, opened 2024-11-07
chore(ci): save support zip always after tests

frezbo:
Save `support.zip` always, also use a different folder for saving logs, so we can save artifacts of multi cluster tests.


smira (approved):

smira (approved):

dsseng (approved):
:100: Perfect, thanks!

PR #9676, opened 2024-11-08
test: fix unpriviliged process runner test

smira:
Don't try cgroups/OOM scoe if not specified in the runner spec.


Unix4ever (approved):

frezbo (approved):

smira:
Now `go test -v ./...` works once again (without sudo).

dsseng (approved):

PR #9648, opened 2024-11-06
feat: label mounts and data directories

dsseng:
- **feat: add SELinux labels to mounts**
- **test: check labels for mounts and files in there**

Issue #8897, opened 2024-06-12
Refactoring initialize sequence tasks


smira:
Goal: reduce number of tasks in the `initialize` sequence, eventually get rid of the sequence.

```[tasklist]
### Tasks
- [ ] https://github.com/siderolabs/talos/issues/8898
- [ ] https://github.com/siderolabs/talos/issues/8899
- [ ] https://github.com/siderolabs/talos/issues/8900
```

PR #9687, opened 2024-11-09
refactor: mount /run, /system and /tmp via tasks, not in initramfs

dsseng:
These are not used in initramfs, and require extra actions to relabel /system and /run in case they are mounted before loading the SELinux policy. Move them farther thus

Part of #9648

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
@@ -91,6 +91,9 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 		)
 	default:
 		phases = phases.Append(
+			"mountTmpfs",
+			MountTmpFilesystems,
+		).Append(
 			"systemRequirements",
 			EnforceKSPPRequirements,
 			SetupSystemDirectory,
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
@@ -1949,6 +1949,22 @@ func haltIfInstalled(seq runtime.Sequence, _ any) (runtime.TaskExecutionFunc, st
 	}, "haltIfInstalled"
 }
 
+// MountTmpFilesystems represents the MountTmpFilesystems task.
+func MountTmpFilesystems(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
+	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
+		tmpPoints := mountv2.Points{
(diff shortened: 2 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1949,6 +1949,22 @@ func haltIfInstalled(seq runtime.Sequence, _ any) (runtime.TaskExecutionFunc, st
 	}, "haltIfInstalled"
 }
 
+// MountTmpFilesystems represents the MountTmpFilesystems task.
+func MountTmpFilesystems(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
+	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
+		tmpPoints := mountv2.Points{
+			mountv2.NewPoint("tmpfs", "/run", "tmpfs", mountv2.WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_RELATIME), mountv2.WithData("mode=0755")),
+			mountv2.NewPoint("tmpfs", "/system", "tmpfs", mountv2.WithData("mode=0755")),
+			mountv2.NewPoint("tmpfs", "/tmp", "tmpfs", mountv2.WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV), mountv2.WithData("size=64M"), mountv2.WithData("mode=0755")),
```
dsseng, line 1958: We do not actually need to move /tmp, as it's not labeled (uses tmpfs_t gained by its type). Please share your thoughts whether or not we should mount it here or in initramfs

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -91,6 +91,9 @@ func (*Sequencer) Initialize(r runtime.Runtime) []runtime.Phase {
 		)
 	default:
 		phases = phases.Append(
+			"mountTmpfs",
+			MountTmpFilesystems,
```
smira, line 95: I know you did that coming from a perspective of doing things consistently, but I think we should drop this from the sequencer.

The problem is that controllers and sequencer start at the same time, and it's very inconvenient not to have `/tmp` or `/system` at this point.

So following #8897, I would rather move it out to be a proper function which wraps the initialization as a whole.

This addresses also your concern about unmounting... so something like:

```go
func initEarly() err {
   unmounter, err := mountTmpFilesystems()
   if err != nil { ... }
   
  defer unmounter()
  
  doOtherStuff()
}
```

(and we need to skip this for containers, so place this accordingly)

dsseng: Well, my primary concern for review here is whether or not it's okay and not too late time to mount these? It worked for me, but I'm not quite sure initramfs init won't need either of these.

smira:
Reworked into #9709

Issue #9694, opened 2024-11-11
HTTP 500 when attempting download of image


Hexoplon:
### Description

When attempting to download a metal image with multiple extensions of the new v1.9.0-alpha.2 release, I get a HTTP 500 internal server error response. However, it works for the plain image, as well as with a single extension.

See `856b061c615e1d15460c46bd87a3021d353c0a3f1fae5f2c68f22dd3d3084e14`, [metal-amd64.iso](https://factory.talos.dev/image/856b061c615e1d15460c46bd87a3021d353c0a3f1fae5f2c68f22dd3d3084e14/v1.9.0-alpha.2/metal-amd64.iso) as an example.

### Environment

- Talos version: v1.9.0-alpha.2

smira:
```
{"level":"info","ts":1731321404.9692018,"caller":"http/http.go:164","msg":"request","frontend":"http","method":"GET","path":"/image/856b061c615e1d15460c46bd87a3021d353c0a3f1fae5f2c68f22dd3d3084e14/v1.9.0-alpha.2/metal-amd64.iso","error":"error generating asset: error validating extension \"glibc\": version constraint >= v1.9.0 can't be satisfied with Talos version 1.8.0-alpha.1"}
```

(This is the error, and it's wrong).

Hexoplon:
The issue is resolved now! Thanks :D

PR #9696, opened 2024-11-11
fix: use imager incoming version for extension validation

frezbo:
Use the version coming from imager to validate extension constraints.

Part of : #9694

### DIFF
--- a/pkg/imager/extensions/extensions.go
+++ b/pkg/imager/extensions/extensions.go
@@ -24,6 +24,8 @@ type Builder struct {
 	Arch string
 	// ExtensionTreePath is a path to the extracted extension tree.
 	ExtensionTreePath string
+	// ExtensionValidateContents enables validation of the extension contents.
+	ExtensionValidateContents bool
 	// Printf is used for logging.
 	Printf func(format string, v ...any)
 	// Quirks for the Talos version being used.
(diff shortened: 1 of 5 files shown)


pkg/machinery/extensions/validate.go
```diff
@@ -12,28 +12,21 @@ import (
...
 
 // Validate the extension: compatibility, contents, etc.
-func (ext *Extension) Validate() error {
-	if err := ext.validateConstraints(); err != nil {
+func (ext *Extension) Validate(talosVersion semver.Version) error {
+	if err := ext.validateConstraints(talosVersion); err != nil {
 		return err
 	}
 
 	return ext.validateContents()
```
smira, line 23: how about `validateContents()` - as it's different for different versions of Talos, it might still break?

smira (approved):

pkg/machinery/extensions/validate.go
frezbo: I guess then  we need have something similar to quirks for extension allowed paths

smira: yep, either that, or a way to disable it in `Validate()` - e.g. Image Factory works on actually approved/shipped versions, so contents should be fine?

pkg/imager/extensions/extensions.go
```diff
@@ -24,6 +24,8 @@ type Builder struct {
 	Arch string
 	// ExtensionTreePath is a path to the extracted extension tree.
 	ExtensionTreePath string
+	// ExtensionValidateContents enables validation of the extension contents.
+	ExtensionValidateContents bool
```
frezbo, line 28: I can drop this since we only use it for the deprecated `.machine.install.extensions` https://github.com/siderolabs/talos/pull/9696/files#diff-9adbdc7691682c62b382f9040db20e4e0d54e5be12c74378bef14a586a4adf20R22

pkg/machinery/extensions/validate.go
```diff
@@ -16,24 +16,83 @@ import (
...
 
-func (ext *Extension) validateConstraints() error {
-	constraint := ext.Manifest.Metadata.Compatibility.Talos.Version
+// ValidationOption is a function that configures the validation options.
+type ValidationOption func(*ValidationOptions) error
 
-	if constraint != "" {
-		talosVersion, err := semver.ParseTolerant(version.Tag)
+// Validate the extension: compatibility, contents, etc.
+func (ext *Extension) Validate(opts ...ValidationOption) error {
```
frezbo, line 61: did this, so `Validate` is backword compatible, I could also negate the options, so `Validate` works as is

smira (approved):

PR #9705, opened 2024-11-12
Fix a typo

sophia-coldren:
The documentation for running talos on vultr misspells "bootstrap" as "boostrap".  Just want to fix that so the next person to do it can copy-paste, and not have to fix it themself.


smira (approved):
thank you!

PR #9709, opened 2024-11-12
refactor: move early initialization functions to pre-initialize phase

### LINKED ISSUE #8900: Extract common tasks as simple function calls
Code in question: https://github.com/siderolabs/talos/blob/1fb8453c2db1659dd6c1670e4174125b26e777c5/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go#L89-L103

Here we execute a sequence of steps (a bit different for containers/non-containers), which might look better as a simple sequence of Go function calls with `defer` steps to undo the action, as they should be called strictly before anything else starts running, and undone before the machine shuts down operations.

smira:
Fixes #8900

Closes #9687

(contains splitting of late mounts)

The benefits:

* tasks run _before_ controllers are started
* tasks can register `defer` to undo actions

This decomposes sequencer tasks a bit.

### DIFF
--- a/internal/app/machined/main.go
+++ b/internal/app/machined/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/siderolabs/go-cmd/pkg/cmd/proc/reaper"
 	debug "github.com/siderolabs/go-debug"
 	"github.com/siderolabs/go-procfs/procfs"
+	"go.uber.org/zap"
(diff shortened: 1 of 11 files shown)


Unix4ever (approved):

dsseng (approved):
Tested with mount labeling branch, works great. Thanks!

PR #9715, opened 2024-11-13
feat: allow customizing PXE boot order in `talosctl cluster create`

utkuozdemir:
Expose more PXE boot handles to the `talosctl cluster create` command. Allows overriding boot order to prioritize booting from network.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -164,6 +164,8 @@ var (
 	crashdumpOnFailure        bool
 	skipKubeconfig            bool
 	skipInjectingConfig       bool
+	defaultBootOrder          string
+	pxeBooted                 bool
 	talosVersion              string
 	encryptStatePartition     bool
 	encryptEphemeralPartition bool
@@ -904,6 +906,8 @@ func create(ctx context.Context) error {
 			Disks:                 disks,
 			Mounts:                mountOpts.Value(),
 			SkipInjectingConfig:   skipInjectingConfig,
+			DefaultBootOrder:      defaultBootOrder,
+			PXEBooted:             pxeBooted,
 			ConfigInjectionMethod: configInjectionMethod,
 			BadRTC:                badRTC,
 			ExtraKernelArgs:       extraKernelArgs,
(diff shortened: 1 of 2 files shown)


smira (commented):
idk how I feel about using `talosctl cluster create` for PXE booted VMs, otherwise we end up creating every possible option. 

It might be way easier to directly use provisioning library in the integration test (like Sfyra does) to have full control over all options? Why can't we use the library to create VMs?

utkuozdemir:
> idk how I feel about using `talosctl cluster create` for PXE booted VMs, otherwise we end up creating every possible option.
> 
> It might be way easier to directly use provisioning library in the integration test (like Sfyra does) to have full control over all options? Why can't we use the library to create VMs?

We can, these options are just handy for local testing/development. If we don't add these, I will have to replicate them using the library, at least for my local development. There are cases where I don't want to run an integration test, but simply run a few commands to test some flow fully manually.

smira:
> We can, these options are just handy for local testing/development. If we don't add these, I will have to replicate them using the library, at least for my local development. There are cases where I don't want to run an integration test, but simply run a few commands to test some flow fully manually.

We have the code here: https://github.com/siderolabs/sidero/blob/main/sfyra/pkg/bootstrap/cluster.go

It looks way more complicated than you actually need, as you only need PXE VMs (and it creates both management nodes and VMs), and you can wrap it as a `cmd` in your project tailored to your specific needs.

My concern is that you'd need more features in the end - you want them to be PXE-booted or rely on proxy-DHCP to test both flows, and etc. And as the scope grows, it will be less and less feasible to add 101st flag to `talosctl cluster create`, while you can have something which works great exactly for your scenario.

smira:
you don't need 99% of `talosctl cluster create`, but rather you need just PXE-booted VMs, that's it; in fact, you don't need a loadbalancer, you don't need any config injection, etc.

utkuozdemir:
> > We can, these options are just handy for local testing/development. If we don't add these, I will have to replicate them using the library, at least for my local development. There are cases where I don't want to run an integration test, but simply run a few commands to test some flow fully manually.
> 
> We have the code here: [siderolabs/sidero@`main`/sfyra/pkg/bootstrap/cluster.go](https://github.com/siderolabs/sidero/blob/main/sfyra/pkg/bootstrap/cluster.go?rgh-link-date=2024-11-13T13%3A34%3A54Z)
> 
> It looks way more complicated than you actually need, as you only need PXE VMs (and it creates both management nodes and VMs), and you can wrap it as a `cmd` in your project tailored to your specific needs.
> 
> My concern is that you'd need more features in the end - you want them to be PXE-booted or rely on proxy-DHCP to test both flows, and etc. And as the scope grows, it will be less and less feasible to add 101st flag to `talosctl cluster create`, while you can have something which works great exactly for your scenario.

Yep, I actually took the idea from exactly this code. But wrapping in a cmd sound ok, I'll do that instead.

PR #9716, opened 2024-11-13
feat: implement tracking of blockdevice secondaries

smira:
This is going to be used to detect disks that are safe to wipe.

For blockdevices, track secondaries as direct references, e.g. encrypted `STATE` partition might have secondary `vda5`.

For disks, re-map secondaries to be whole devices names, e.g. `vda`.


frezbo:
need tests?

smira:
> need tests?

yes... good idea, I will do

smira:
> > need tests?
> 
> yes... good idea, I will do

added some tests for secondary disks detection

frezbo (approved):

PR #9722, opened 2024-11-14
feat: registry proxy

DmitriyMV:
Implement container registry proxy.

### DIFF
--- a/internal/app/machined/pkg/system/services/registry/app/main.go
+++ b/internal/app/machined/pkg/system/services/registry/app/main.go
@@ -1,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"io/fs"
+	"os"
+	"os/signal"
+	"path/filepath"
+
+	"go.uber.org/zap"
+
+	"github.com/siderolabs/talos/internal/app/machined/pkg/system/services/registry"
+)
+
+func main() {
+	if err := app(); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func app() error {
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
+
+	development, err := zap.NewDevelopment()
+	if err != nil {
+		return fmt.Errorf("failed to create development logger: %w", err)
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to get user home directory: %w", err)
+	}
+
+	it := func(yield func(fs.StatFS) bool) {
+		for _, root := range []string{"registry-cache-2", "registry-cache"} {
+			if !yield(os.DirFS(filepath.Join(homeDir, root)).(fs.StatFS)) {
+				return
+			}
+		}
+	}
+
+	return registry.NewService(registry.NewMultiPathFS(it), development).Run(ctx)
+}
--- a/internal/app/machined/pkg/system/services/registry/registry.go
+++ b/internal/app/machined/pkg/system/services/registry/registry.go
@@ -1,0 +1,271 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package registry provides a simple container registry service.
+package registry
+
+import (
+	"bytes"
+	"cmp"
+	"context"
+	"crypto/sha256"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/containerd/containerd/v2/core/content"
+	"github.com/distribution/reference"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/siderolabs/gen/xerrors"
+	"go.uber.org/zap"
+)
+
+// NewService creates a new instance of the registry service.
+func NewService(root fs.StatFS, logger *zap.Logger) *Service {
+	return &Service{root: root, logger: logger}
+}
+
+// Service is a container registry service.
+type Service struct {
+	logger *zap.Logger
+	root   fs.StatFS
+}
+
+// Run is an entrypoint to the API service.
+func (svc *Service) Run(ctx context.Context) error {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("GET /v2/{args...}", svc.serveHTTP)
+
+	giveOk := func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }
+	for _, p := range []string{"v2", "healthz"} {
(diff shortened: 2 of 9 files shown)


hack/cri-plugin.part
```diff
@@ -1,4 +1,2 @@
 version = 3
 
-[plugins."io.containerd.cri.v1.images"]
-  discard_unpacked_layers = true
```
smira, line 4: we don't need this and matching changing `image.go`

internal/app/machined/pkg/system/services/registry/registry.go
```diff
@@ -0,0 +1,196 @@
...
+
+	mux.HandleFunc("GET /v2/{args...}", s.serveHTTP)
+
+	giveOk := func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }
+	for _, p := range []string{"v2", "healthz"} {
+		mux.HandleFunc("GET /"+p, giveOk)
+		mux.HandleFunc("GET /"+p+"/{$}", giveOk)
+	}
+
+	server := http.Server{Addr: ":3172", Handler: mux}
```
smira, line 51:
```suggestion
	server := http.Server{Addr: "localhost:3172", Handler: mux}
```

internal/app/machined/pkg/system/services/registry/registry.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Run is an entrypoint to the API service.
+func (s *Service) Run(ctx context.Context) error {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("GET /v2/{args...}", s.serveHTTP)
+
+	giveOk := func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }
+	for _, p := range []string{"v2", "healthz"} {
+		mux.HandleFunc("GET /"+p, giveOk)
```
smira, line 47: I'm not sure if `mux` handles `HEAD` transparently, but I believe they do `HEAD /v2` requests as well

internal/app/machined/pkg/system/services/registry/registry.go
```diff
@@ -0,0 +1,248 @@
...
+	return manifest.MediaType, manifestBlob, nil
+}
+
+type canonical struct {
+	reference.NamedTagged
+	digest digest.Digest
+}
+
+func (c *canonical) String() string        { return c.NamedTagged.String() + "@" + c.digest.Encoded() }
+func (c *canonical) Digest() digest.Digest { return c.digest }
```
dsseng, line 248: Weird formatting, won't classic one with newlines look less out of place?

internal/app/machined/pkg/system/services/registry/store_darwin.go
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
dsseng, line 1: Does this module get pulled into talosctl build? Then perhaps needs a Windows counterpart, else why does it need a Darwin implementation?

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -53,7 +53,7 @@ func (o *APID) ID(r runtime.Runtime) string {
 }
 
 // apidResourceFilter filters access to COSI state for apid.
-func apidResourceFilter(ctx context.Context, access state.Access) error {
+func apidResourceFilter(_ context.Context, access state.Access) error {
```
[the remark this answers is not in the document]
DmitriyMV: Moved to the https://github.com/siderolabs/talos/pull/9760

internal/app/machined/pkg/system/services/registry/registry.go
DmitriyMV: Usually gofmt allows functions with 105 characters or less to be single line.

dsseng (approved):

internal/app/machined/pkg/system/services/registry/app/main.go
```diff
@@ -0,0 +1,41 @@
...
+func app() error {
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer cancel()
+
+	development, err := zap.NewDevelopment()
+	if err != nil {
+		return fmt.Errorf("failed to create development logger: %w", err)
+	}
+
+	homeDir, err := os.UserHomeDir()
```
frezbo, line 35: why do we need the homedir here? i guess it's `/root` in talos :thinking:

DmitriyMV: It's just an standalone example that you can run and test against.

internal/app/machined/pkg/system/services/registryd.go
```diff
@@ -0,0 +1,80 @@
...
+				logging.WithColoredLevels(),
+			)),
+		).Run(ctx)
+	}, runner.WithLoggingManager(r.Logging())), nil
+}
+
+// HealthFunc implements the HealthcheckedService interface.
+func (m *Registryd) HealthFunc(runtime.Runtime) health.Check {
+	return func(ctx context.Context) error {
+		// TODO: implement me
```
smira, line 72: we should just do TCP/HTTP check here, should be trivial, e.g. like `kubelet` health check

internal/app/machined/pkg/system/services/registry/registry.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/containerd/containerd/v2/core/content"
+	"github.com/distribution/reference"
+	"github.com/opencontainers/go-digest"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/siderolabs/gen/xerrors"
+	"go.uber.org/zap"
+)
+
+// NewService creates a new instance of the registry service.
+func NewService(root string, logger *zap.Logger) *Service {
```
smira, line 32: let's add multiple roots (probably best if we can inject them on the fly? or the service itself can read that from the resource, but that can come later)

smira (approved):

internal/app/machined/pkg/system/services/registry/registry.go
smira, line 32:
```suggestion
func NewService(rootsFunc func() (paths []string, err error), logger *zap.Logger) *Service {
```

smira (approved):

frezbo (approved):

internal/app/machined/pkg/system/services/registry/registry.go
frezbo: this probably needs to be added

smira: it was added, but via MultiFS iterator (see the example in the service start function)

frezbo: oh i thought I commented on the healthcheck part, sorry

internal/app/machined/pkg/system/services/registryd.go
```diff
@@ -0,0 +1,61 @@
...
+// NewRegistryD returns a new docker mirror registry service.
+func NewRegistryD() system.Service                                       { return &registryD{} }
+func (r *registryD) ID(runtime.Runtime) string                           { return "registryd" }
+func (r *registryD) HealthSettings(runtime.Runtime) *health.Settings     { return &health.DefaultSettings }
+func (r *registryD) PreFunc(context.Context, runtime.Runtime) error      { return nil }
+func (r *registryD) PostFunc(runtime.Runtime, events.ServiceState) error { return nil }
+func (r *registryD) Condition(runtime.Runtime) conditions.Condition      { return nil }
+func (r *registryD) DependsOn(runtime.Runtime) []string                  { return nil }
+
+func (r *registryD) HealthFunc(runtime.Runtime) health.Check {
```
DmitriyMV, line 37: @smira healthcheck ^^^

PR #9723, opened 2024-11-14
fix: registry mirror fallback handling

### LINKED ISSUE #9613: adjust Talos resolver to use upstream as a fallback
There's a mismatch between Talos docker resolver and containerd one: containerd defaults to having original endpoint (e.g. ghcr.io) as a fallback, while Talos doesn't.

Talos resolver is used to pull installer, etcd, kubelet images, while containerd pulls all other images.

> Note: this statement might need additional verification.

smira:
Fixes #9613

This has two changes:

* adjust Talos registry resolver to match containerd (CRI) resolver: use by default upstream as a fallback
* add a machine config option to skip upstream as a fallback, and adjust CRI configuration accordingly

See https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-configuration---examples for details on CRI's `hosts.toml`.

### DIFF
--- a/internal/pkg/containers/image/resolver.go
+++ b/internal/pkg/containers/image/resolver.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containerd/containerd/v2/core/remotes"
 	"github.com/containerd/containerd/v2/core/remotes/docker"
 	"github.com/hashicorp/go-cleanhttp"
+	"github.com/siderolabs/gen/xslices"
 
 	"github.com/siderolabs/talos/pkg/httpdefaults"
 	"github.com/siderolabs/talos/pkg/machinery/config/config"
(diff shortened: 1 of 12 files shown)


internal/pkg/containers/image/resolver.go
```diff
@@ -97,25 +98,54 @@ func RegistryHosts(reg config.Registries) docker.RegistryHosts {
...
+	OverridePath bool
+}
+
+// RegistryEndpointEntriesFromConfig returns registry endpoints per host.
+func RegistryEndpointEntriesFromConfig(host string, reg config.RegistryMirrorConfig) ([]EndpointEntry, error) {
+	entries := xslices.Map(reg.Endpoints(), func(endpoint string) EndpointEntry {
+		return EndpointEntry{Endpoint: endpoint, OverridePath: reg.OverridePath()}
+	})
+
+	if !reg.SkipFallback() {
```
dsseng, line 113: nit: just return when skipping to reduce code nesting?

dsseng (approved):

internal/pkg/containers/image/resolver.go
smira: yeah good idea, thanks, I will update!

Issue #9611, opened 2024-10-31
generating raw image cache


smira:
Implement a Go module/command (might be part of `talosctl`? or separate) which given an input of a list of image references generate the image cache as a file structure:

```
+
| - blob/ # blob storage, not namespaced
|   | - sha256:c0b4755c9c53ddd4455e73fd52155ebff41e27c36f75d0bfd0bf0d8ee925e086
|   | - sha256:e5c17e4c7610e97c3f9db67d76b9e1404acd0ea0ebb2aeefe6319125c047f382
|   \ - ...
\ - manifests/
    | - ghcr.io/ # registry name
    |   \ - siderolabs/ # image name, split by /
    |       \ - kubelet/
    |           | - reference/
    |           |   \ - v1.31.0 # manifest by reference, ghcr.io/siderolabs/kubelet:v1.31.0
    |           \ - digest/
    |               \ - sha256:e5c17e4c7610e97c3f9db67d76b9e1404acd0ea0ebb2aeefe6319125c047f382 # manifest by digest
    \ - registry.k8s.io/
        \ - kube-apiserver/
            | - reference/
            \ - digest/ 
```

PR #9733, opened 2024-11-15
feat: image cache gen

frezbo:
Fixes: #9611

### DIFF
--- a/pkg/imager/cache/cache.go
+++ b/pkg/imager/cache/cache.go
@@ -1,0 +1,189 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package cache provides facilities for generating a cache tarball from images.
+package cache
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/authn/github"
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+
+	"github.com/siderolabs/talos/pkg/imager/filemap"
+)
+
+const (
+	blobsDir     = "blob"
+	manifestsDir = "manifests"
+)
+
+// Generate generates a cache tarball from the given images.
+//
+//nolint:gocyclo,cyclop
+func Generate(images []string, platform string, insecure bool, dest string) error {
+	v1Platform, err := v1.ParsePlatform(platform)
+	if err != nil {
+		return fmt.Errorf("parsing platform: %w", err)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "talos-image-cache-gen")
+	if err != nil {
+		return fmt.Errorf("creating temporary directory: %w", err)
+	}
+
+	defer os.RemoveAll(tmpDir) //nolint:errcheck
+
+	if err := os.MkdirAll(filepath.Join(tmpDir, blobsDir), 0o755); err != nil {
+		return err
+	}
+
+	nameOptions := []name.Option{
+		name.StrictValidation,
+	}
+
+	craneOpts := []crane.Option{
+		crane.WithAuthFromKeychain(
+			authn.NewMultiKeychain(
+				authn.DefaultKeychain,
+				github.Keychain,
+			),
+		),
+	}
+
+	remoteOpts := []remote.Option{
+		remote.WithAuthFromKeychain(authn.NewMultiKeychain(
+			authn.DefaultKeychain,
+			github.Keychain,
+		)),
+		remote.WithPlatform(*v1Platform),
+	}
+
+	if insecure {
+		craneOpts = append(craneOpts, crane.Insecure)
+		nameOptions = append(nameOptions, name.Insecure)
+	}
+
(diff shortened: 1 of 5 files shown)


frezbo:
```bash
❯ _out/talosctl-linux-amd64 image cache-create --images ghcr.io/siderolabs/extensions:sha256-1ec1e7d0c561ae2466292e3308792c5ab9dc7e3c8fea3025744c4537be6a2f69.sig --images ghcr.io/siderolabs/extensions:v1.8.3 --cache-path /tmp/cache.tar

❯ crane push /tmp/cache.tar 127.0.0.1:5010/frezbo/cache:latest
2024/11/15 19:58:10 existing blob: sha256:a348e59f06de8f5212048da8615a70e3ee9bd1f075c3896e3ad1db7659440615
2024/11/15 19:58:10 pushed blob: sha256:fa6ca17b2c3e49fa512f43313f09b8ba5adfa238ad853ea794601db4dd8a99f9
2024/11/15 19:58:10 127.0.0.1:5010/frezbo/cache:latest: digest: sha256:036d51c6f7e4190debfd7d2bc3bf20566689dbb6c486df89c4a6a0ed408e9275 size: 425
127.0.0.1:5010/frezbo/cache@sha256:036d51c6f7e4190debfd7d2bc3bf20566689dbb6c486df89c4a6a0ed408e9275

❯ crane export 127.0.0.1:5010/frezbo/cache:latest - | tar tf - | pbcopy
tar: Removing leading `/' from member names
↕️  3 talos on  feat/image-cache-gen is 📦 v0.0.1 via  v23.2.0 
❯ crane export 127.0.0.1:5010/frezbo/cache:latest - | tar tf -
tar: Removing leading `/' from member names
/blob/sha256:bd2fb5c432256b78a3b635fe6101fc07cd03468a6500a194279bf22122a4e8aa
/blob/sha256:d1e1a3f48a04df2cd2351d92a00777641ff422eec753ae3db1ba6bd6f80da164
/blob/sha256:e36ac478125ae264890cc4ac584e893957f00c242b21d2606c2b735b8ddbd132
/manifests/ghcr.io/siderolabs/extensions/digest/sha256:1ec1e7d0c561ae2466292e3308792c5ab9dc7e3c8fea3025744c4537be6a2f69
/manifests/ghcr.io/siderolabs/extensions/digest/sha256:64972d0a5e2f41ba37212e10392f6ebc8a6422b3db3ae0f6cd7cefffb3541515
/manifests/ghcr.io/siderolabs/extensions/reference/sha256-1ec1e7d0c561ae2466292e3308792c5ab9dc7e3c8fea3025744c4537be6a2f69.sig
/manifests/ghcr.io/siderolabs/extensions/reference/v1.8.3
```

pkg/imager/cache/cache.go
```diff
@@ -0,0 +1,183 @@
...
+		),
+	}
+
+	remoteOpts := []remote.Option{
+		remote.WithAuthFromKeychain(authn.NewMultiKeychain(
+			authn.DefaultKeychain,
+			github.Keychain,
+		)),
+		remote.WithPlatform(*v1Platform),
+	}
```
dsseng, line 74: Does this auth get used? If so, does it allow pulling of private images if there are any to be pulled?

dsseng (approved):

pkg/imager/cache/cache.go
```diff
@@ -0,0 +1,183 @@
...
+			if err != nil {
+				return fmt.Errorf("getting layer reader: %w", err)
+			}
+
+			file, err := os.Create(filepath.Join(tmpDir, blobsDir, digest.String()))
+			if err != nil {
+				return err
+			}
+
+			defer file.Close() //nolint:errcheck
```
DmitriyMV, line 147: `defer`s are function scoped, not lexical, so they'll get executed after all loops. I don't think you wont this here.

dsseng: Won't this be like an automatic closure here? Because file variable is also scoped in this iteration of the loop, it would probably be replicated for each iteration? Or not, and it produces a single deferred call after the last iteration with the last value of the variable?

pkg/imager/cache/cache.go
frezbo: yeh, the library handles github_token and docker auth config

pkg/imager/cache/cache.go
DmitriyMV: It will produce defers for each file variable individually, but it will not execute them until function finishes.

dsseng: Yes, this is what I also supposed would happen. Thanks for clarification

pkg/imager/cache/cache.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return fmt.Errorf("creating artifacts layer: %w", err)
+	}
+
+	newImg, err = mutate.AppendLayers(newImg, artifactsLayer)
+	if err != nil {
+		return fmt.Errorf("appending artifacts layer: %w", err)
+	}
+
+	return tarball.WriteToFile(dest, nil, newImg)
```
smira, line 188: not super important, but could be OCI output instead of a tarball?

pkg/imager/cache/cache.go
```diff
@@ -0,0 +1,189 @@
...
+
+		rmt, err := remote.Get(
+			ref,
+			remoteOpts...,
+		)
+		if err != nil {
+			return fmt.Errorf("fetching image %q: %w", src, err)
+		}
+
+		if !strings.HasPrefix(ref.Identifier(), "sha256:") {
```
smira, line 111: I wonder if we should handle the reference as 

`some/image:v1.3.5@sha256:abcdef` (pinned ref), and write both tag & sha versions of the digest

smira (approved):
should we add a test which does `talosctl images default | talosctl images cache-gen` ?

pkg/imager/cache/cache.go
frezbo: so we enforce it has a digest as input always?

smira: no, not enforce, but handle that correctly.

smira:
> ```shell
> ❯ crane export 127.0.0.1:5010/frezbo/cache:latest - | tar tf -
> tar: Removing leading `/' from member names
> /blob/sha256:bd2fb5c432256b78a3b635fe6101fc07cd03468a6500a194279bf22122a4e8aa
> /manifests/ghcr.io/siderolabs/extensions/reference/v1.8.3
> ```


I guess we should drop the leading `/`

PR #9737, opened 2024-11-16
refactor: remove SELinux client_u and client_r

dsseng:
I added those in first days of the current policy development, yet there was no use for them. This change simplifies the policy and handling of labels.


internal/pkg/selinux/policy/policy.33
frezbo, line 1: seems this reduced the policy size a lot :sweat_smile:

frezbo (approved):

internal/pkg/selinux/policy/policy.33
dsseng: Huh, interesting :)

It's 24.6 KB total in here, idk where is the 99% reduction https://github.com/siderolabs/talos/blob/71ff71ccbd407719a90b4f1f007bf90c8f3e383d/internal/pkg/selinux/policy/policy.33

PR #9738, opened 2024-11-17
feat: add with-kvm flag to the create cluster command

Orzelius:
Allow enabling or disabling kvm via the cli when running the create cluster command.
This is useful for testing in/for environments where kvm is not available.
Context: I'm working on getting the qemu provider to work on mac, but also want to test that my changes don't break stuff on linux side of things. I only have access to linux via docker on a mac host, which doesn't support KVM.

this is my first time contributing to talos so please let me know if something's missing :)

### DIFF
--- a/pkg/provision/providers/qemu/arch.go
+++ b/pkg/provision/providers/qemu/arch.go
@@ -200,11 +200,14 @@ func generateUEFIPFlashList(uefiSourcePathPrefixes, uefiSourceFiles, uefiVarsFil
 }
 
 // QemuExecutable returns name of qemu executable for the arch.
-func (arch Arch) QemuExecutable() string {
+func (arch Arch) QemuExecutable(kvmEnabled bool) string {
 	binaries := []string{
 		"qemu-system-" + arch.QemuArch(),
-		"qemu-kvm",
-		"/usr/libexec/qemu-kvm",
+	}
+	if kvmEnabled {
+		binaries = append(binaries,
+			"qemu-kvm",
+			"/usr/libexec/qemu-kvm")
 	}
 
 	for _, binary := range binaries {
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -124,6 +124,7 @@ var (
 	applyConfigEnabled        bool
 	bootloaderEnabled         bool
 	uefiEnabled               bool
+	kvmEnabled                bool
 	tpm2Enabled               bool
 	extraUEFISearchPaths      []string
 	configDebug               bool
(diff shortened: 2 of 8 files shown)


pkg/provision/providers/qemu/arch.go
```diff
@@ -200,11 +200,14 @@ func generateUEFIPFlashList(uefiSourcePathPrefixes, uefiSourceFiles, uefiVarsFil
...
 // QemuExecutable returns name of qemu executable for the arch.
-func (arch Arch) QemuExecutable() string {
+func (arch Arch) QemuExecutable(kvmEnabled bool) string {
 	binaries := []string{
 		"qemu-system-" + arch.QemuArch(),
-		"qemu-kvm",
-		"/usr/libexec/qemu-kvm",
+	}
+	if kvmEnabled {
+		binaries = append(binaries,
```
smira, line 208: I'm confused a bit, this just changes the search path by removing `qemu-kvm` ? What does it do in terms of features?

dsseng: I believe the correct way to handle this parameter would be the function at the bottom of this file: https://github.com/siderolabs/talos/blob/27d771e0fe355f54079bde5566c56143b027431a/pkg/provision/providers/qemu/arch.go#L243

This is because we should first exclude the argument demanding the use of KVM rather than finding a binary that does not include the KVM acceleration support. However `qemu-kvm` command might force KVM enabled on some distros: not really sure about that

smira: I haven't tested that myself, but I guess `kvm` is enabled by default, so we can skip forcing it (which would fix the case when KVM is not available), but without KVM emulation would be super slow

Orzelius: > I'm confused a bit, this just changes the search path by removing qemu-kvm

Yes, this line just removes `qemu-kvm`binary  from the search list if KVM is disabled. Maybe `qemu-kvm` can be used even without kvm, but I think it'd be confusing and I haven't looked into it.

> What does it do in terms of features?

Do you mean the PR in general? It allows to disable KVM usage, which was mandatory before and would error if KVM could not be used. Now a user can optionally set `--with-kvm` to false and thus run talos via qemu even on a system which doesn't have kvm (e.q. wsl, docker on a non linux host or another vm).

smira: I think `qemu-kvm` is just yet another alias present on some systems, and removing it from the search path doesn't make any difference, or make it more confusing in the end. 

`talosctl cluster create` doesn't work on non-Linux, as CNI binaries used to set up networking won't work on non-Linux hosts, so the only case is Linux with KVM disabled, which probably exists, but rare to find.

Orzelius: Would it then make more sense to just log a warning and proceed with KVM off when it's not available?
I think it still makes sense as it's not a big change and would aid in development and testing of the create command. At least I would benefit from it (see pr description)

Orzelius:
closing in favor to an upcoming pr

Issue #9725, opened 2024-11-14
Kubelet/etcd uses wrong IPv6 Address


trevex:
### Description

When Talos is run in an IPv6 Single-Stack environment and is assigned multiple IPs by DHCP and RA (although this will most likely apply to Dual-Stack as well) the Kubelet will use the wrong Address.

In our case Talos is running in KubeVirt with the Passt network binding plugin and gets an IP via RA followed by an /128 IP from DHCPv6. Only the latter has full bi-directional connectivity.

The preferred  /128 address has the flag `permanent` while the RA address has the flag `mngmtmpaddr`.

The `permanent` address should be preferred.

### Logs

Relevant excerpts from `omnictl support`:

AddressStatuses
```yaml
# cat fd01:cafe::5054:ff:fe1f:c7bd/resources/addressstatuses.net.talos.dev.yaml
metadata:
    namespace: network
    type: AddressStatuses.net.talos.dev
    id: eth0/fd01:cafe::5054:ff:fe1f:c7bd/64
    version: 1
    owner: network.AddressStatusController
    phase: running
    created: 2024-11-14T15:52:53Z
    updated: 2024-11-14T15:52:53Z
spec:
    address: fd01:cafe::5054:ff:fe1f:c7bd/64
    linkIndex: 8
    linkName: eth0
    family: inet6
    scope: global
    flags: mngmtmpaddr
---
metadata:
    namespace: network
    type: AddressStatuses.net.talos.dev
    id: eth0/fd01:cafe::f14c:9fa1:8496:557f/128
    version: 1
    owner: network.AddressStatusController
    phase: running
    created: 2024-11-14T15:52:55Z
    updated: 2024-11-14T15:52:55Z
spec:
    address: fd01:cafe::f14c:9fa1:8496:557f/128
    linkIndex: 8
    linkName: eth0
    family: inet6
    scope: global
    flags: permanent
---
metadata:
    namespace: network
    type: AddressStatuses.net.talos.dev
    id: eth0/fe80::5054:ff:fe1f:c7bd/64
    version: 2
    owner: network.AddressStatusController
    phase: running
    created: 2024-11-14T15:52:50Z
    updated: 2024-11-14T15:52:52Z
spec:
    address: fe80::5054:ff:fe1f:c7bd/64
    linkIndex: 8
    linkName: eth0
    family: inet6
    scope: link
    flags: permanent
---
metadata:
    namespace: network
    type: AddressStatuses.net.talos.dev
[...]

PR #9749, opened 2024-11-18
feat: prefer longer prefixes in IPv6 only environments as default Nod…

trevex:
# Pull Request

## What? (description)

Prefers longer prefixes in IPv6 only environments as default address, but keeps current behaviour everywhere else.

## Why? (reasoning)

Talos IP preference strategy does not well in IPv6 Single-Stack environments as RA SLAAC addresses will be preferred over DHCPv6. In environments such as KubeVirt with Passt the SLAAC address is not routable by other Pods. If a DHCPv6 address is available it should be preferred.

The current filtering methods do not work as SLAAC and DHCPv6 address can be from the same prefix or in the case of KubeVirt a non-predictable pod IP.

This change only changes the preference in IPv6 only environments by adapting the default NodeAddress accordingly.
In the future it might be desirable to do a larger rewrite or even introduce CEL or expr to let the user define filtering and expression functions. As this is my first PR to Talos I kept the changes as minimal as possible.

Fixes https://github.com/siderolabs/talos/issues/9725.

## Acceptance

Unfortunately `make fmt` changed lots of `*.proto` files, so I did not include the changes.
`make unit-tests` did not run (might be a NixOS issue), but running relevant tests with `go test` passed.
`make docs` did not include relevant updates to the documentation.


trevex:
This is my first PR for Talos and I am not overly familiar with the codebase, but my assumption is changing the default NodeAddress should fix #9725.

internal/app/machined/pkg/controllers/network/node_address.go
```diff
@@ -127,9 +127,17 @@ func (ctrl *NodeAddressController) Run(ctx context.Context, r controller.Runtime
 				continue
 			}
 
+			// in IPv6 only environments we want to prefer longer prefixes, in other environments we keep old behavior
```
smira, line 130: I wonder if the proper fix is to re-sort the `addresses` list in the first place. Today the order comes from the alphabetical order of address resource IDs which look like `<link>/address/bits` so that the IPv6 addresses for the same link are sorted with longest prefix first?

smira: This way it will not only affect `defaultAddress`, but all other list of addresses which are used to pick up etcd, and other listening addresses

trevex: I am not overly familiar with the Talos codebase, is the `defaultAddress` not used by etcd as well?

Sorting the addresses was my initial thought, but I shied away from that approach as I was uncertain how to handle Dual-Stack environments properly. As a first step I could however change the sorting for IPv6 Single-Stack environments.

Another third option, but larger change might be to introduce CEL or [expr](https://github.com/expr-lang/expr) to allow users to customize filtering and IP selection.

smira: Let me think about this more, but my gut feeling is that sorting by string value (as it is today) is not quite good

trevex: I will have another look this evening and/or tomorrow as I also believe sorting might be the better approach. Grepping the codebase once more I am not certain I even understand what the purpose of defaultAddress is as it is only used in the hostname and etcfile controller (today, might be a relic of the past).

trevex:
Finally found some time to work on resorting IPs. I am planning to improve the test coverage, but before doing that, it would be great to some feedback, whether I am on the right track.

(Also feel free to suggest alternative names for the feature.)

smira:
> Finally found some time to work on resorting IPs. I am planning to improve the test coverage, but before doing that, it would be great to some feedback, whether I am on the right track.

I think you're on the right track, I just feel we need to change sorting algorithm a bit more dramatically, and disable it by default.

I will try to get to it today and get it before we cut 1.9 beta.

trevex:
Should I close this PR in favor of yours? Let me know if I can help in any way.

smira:
> Should I close this PR in favor of yours? Let me know if I can help in any way.

I did a bit of refactoring right now to make it easier to plug in more pieces, but I will probably steal something from your PR. Let's keep it open and see if I can converge the two. Thank you!

smira:
> @trevex can you please take a look at #9847 to ensure I haven't messed up the sorting (I made it somewhat more complicated to make it stable), specifically `compare.go`: [#9847 (files)](https://github.com/siderolabs/talos/pull/9847/files#diff-377996f3612a63b53ec9fe9092b1f68188d07c6d60c7d7fdb94db41e88798767)
> 
> I don't have the new algorithm wired in any way, but it's fully tested now.

If there are any tricky cases you think makes sense to add as a test - how should the addresses be ordered in general, that also makes perfect sense.

trevex:
@smira Neat, looks good and would solve all issues related to IPv6 address on our end. The test cases of the compare functions cover all cases that I had in mind as well.

Closing this PR in favor of yours now.

PR #9755, opened 2024-11-19
feat: implement block device wipe

### LINKED ISSUE #9731: support wiping non-volume blockdevices
Required for Omni bare-metal infra provider. cc @utkuozdemir

smira:
Fixes #9731

The wipe doesn't require a reboot, but it requires the blockdevice not to be used as a volume.


frezbo (approved):

dsseng (approved):

cmd/talosctl/cmd/talos/wipe.go
```diff
@@ -0,0 +1,78 @@
...
+		})
+	},
+}
+
+func wipeMethodValues() []string {
+	var method storage.BlockDeviceWipeDescriptor_Method
+
+	values := make([]string, method.Descriptor().Values().Len())
+
+	for idx := range method.Descriptor().Values().Len() {
```
DmitriyMV, line 63: Why not `for idx := range len(storage.BlockDeviceWipeDescriptor_Method_name)`?

internal/app/storaged/server.go
```diff
@@ -7,23 +7,35 @@ package internal
...
 )
 
 // Server implements storage.StorageService.
 //
 // It is only kept here for compatibility purposes, proper API is to query `block.Disk` resources.
 type Server struct {
 	storage.UnimplementedStorageServiceServer
-	Controller runtime.Controller
+	Controller      runtime.Controller
+	MaintenanceMode bool
```
DmitriyMV, line 38: It's always `true` any reason for this variable?

DmitriyMV (approved):

cmd/talosctl/cmd/talos/wipe.go
smira: in theory it might have gaps (we don't have it, but that was my understanding)

PR #9758, opened 2024-11-19
fix: don't activate LVM volumes in agent mode

smira:
In agent mode, we don't need to activate LVMs so that we can safely wipe them.


utkuozdemir (approved):
👍

PR #9760, opened 2024-11-19
chore: small fixes

DmitriyMV:
QoL PR


internal/app/machined/pkg/system/services/registryd.go
```diff
@@ -0,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
dsseng, line 1: Perhaps this went to wrong branch: this is missing from #9722

DmitriyMV: Thanks! Missed.

PR #9762, opened 2024-11-19
refactor: optimize flags and SetLabel

dsseng:
Do not do string lookups in repetitive calls. We do not support changing SELinux status during runtime, so once we read this we can assume status does not change.

Also avoid unneeded FS writes when appropriate label is already set on file.

### DIFF
--- a/internal/pkg/selinux/selinux.go
+++ b/internal/pkg/selinux/selinux.go
@@ -6,12 +6,14 @@
 package selinux
 
 import (
+	"bytes"
 	_ "embed"
 	"log"
 	"os"
+	"sync"
 
+	"github.com/pkg/xattr"
 	"github.com/siderolabs/go-procfs/procfs"
-	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
@@ -23,7 +25,7 @@ var policy []byte
 // the kernel command line. It returns true if SELinux is enabled,
 // otherwise it returns false. It also ensures we're not in a container.
 // By default SELinux is disabled.
-func IsEnabled() bool {
+var IsEnabled = sync.OnceValue(func() bool {
 	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


internal/pkg/selinux/selinux.go
```diff
@@ -6,54 +6,84 @@
...
 	"github.com/siderolabs/go-procfs/procfs"
-	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
 //go:embed policy/policy.33
 var policy []byte
 
+var isEnabled, checkedEnabled, isEnforcing, checkedEnforcing bool
```
DmitriyMV, line 23: I know this code is sequential, but I really don't like the idea of the global mutable variables. Also `is|checked` combo can be replaced with `optional.Optional[bool]`.

dsseng: Another option could be defining functions as constants using [OnceValue](https://pkg.go.dev/sync#OnceValue). To me global variable also seems not a really great idea, so perhaps this is more atomic.

albert849: Great job

smira: I think `IsEnabled`/`IsEnforcing` is better as `sync.OnceValue` - we can't guarantee they won't be called concurrently.

DmitriyMV: I agree. I think you can store `sync.OnceValue` inside unexported variable, move logic there and call this variable from the existing function.

dsseng: Yes, will do like this

internal/pkg/selinux/selinux.go
```diff
@@ -6,54 +6,75 @@
...
 	val := procfs.ProcCmdline().Get(constants.KernelParamSELinux).First()
 
 	return val != nil && *val == "1"
+})
+
+// IsEnabled checks if SELinux is enabled on the system by reading
+// the kernel command line. It returns true if SELinux is enabled,
+// otherwise it returns false. It also ensures we're not in a container.
+// By default SELinux is disabled.
+func IsEnabled() bool {
```
smira, line 38: We could probably have var IsEnabled= sync.OnceValue

smira (approved):

PR #9764, opened 2024-11-20
feat: add e2fsprogs to talos rootfs

frezbo:
Bring in e2fsprogs from pkgs.

Initial support for ext* filesystems.

Fixes: #9746


smira (approved):
I guess we need to add initial support to `pkg/makefs` as well

Issue #9616, opened 2024-10-31
generating images with cache


smira:
Update imager to include image cache generated with #9611 to be included into the image (supported output types - ISO, disk image), as a separate partition. 

The partition should be marked so that it is discoverable by Talos.

smira:
If `imager` is given the image cache image (OCI) tarball, it does the following:

## ISO (both SecureBoot and GRUB)

Create a folder in the ISO filesystem root named `imagecache/` and put the Image Cache there.

## Disk Image

Create a partition with a label `IMAGECACHE` and put the image contents there.

PR #9775, opened 2024-11-21
feat: support generating iso with imagecache

frezbo:
Support generating iso with imagecache.

Part-of: #9616

### DIFF
--- a/internal/integration/api/selinux.go
+++ b/internal/integration/api/selinux.go
@@ -12,13 +12,13 @@ import (
 	"io"
 	"maps"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/siderolabs/go-pointer"
 	"github.com/siderolabs/go-procfs/procfs"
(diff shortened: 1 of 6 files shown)


internal/integration/api/selinux.go
```diff
@@ -12,13 +12,13 @@ import (
...
 	"maps"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/siderolabs/go-pointer"
 	"github.com/siderolabs/go-procfs/procfs"
-	"golang.org/x/exp/slices"
```
frezbo, line 21: good catch :sweat_smile:

frezbo:
Can be either:

```bash
make secureboot-iso IMAGER_ARGS="--image-cache=/out/image-cache"
```

or

```bash
make iso IMAGER_ARGS="--image-cache=127.0.0.1:5010/frezbo/cache:latest"
```

smira (approved):

PR #9778, opened 2024-11-21
feat: support generating disk images with image cache

frezbo:
Add support for generating disk images with image cache.

Fixes: #9616

### DIFF
--- a/internal/pkg/partition/constants.go
+++ b/internal/pkg/partition/constants.go
@@ -25,6 +25,7 @@ const (
 	FilesystemTypeZeroes FileSystemType = "zeroes"
 	FilesystemTypeXFS    FileSystemType = "xfs"
 	FilesystemTypeVFAT   FileSystemType = "vfat"
+	FileSystemTypeExt4   FileSystemType = "ext4"
 )
 
 // Partition default sizes.
--- a/internal/pkg/partition/partition.go
+++ b/internal/pkg/partition/partition.go
@@ -90,6 +90,13 @@ func NewPartitionOptions(label string, uki bool) Options {
 			PartitionType:  LinuxFilesystemData,
 			Size:           0,
 		}
+	case constants.ImageCachePartitionLabel:
+		return Options{
+			FormatOptions:  *formatOptions,
+			PartitionLabel: label,
+			PartitionType:  LinuxFilesystemData,
+			Size:           0,
+		}
 	default:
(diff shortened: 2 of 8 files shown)


frezbo:
One thing i can think of is if the image cache is too large, the disk image creation would fail since we set a default sparse file of almost 1.3gig

```yaml
imageOptions:
    diskSize: 1306525696
```

internal/pkg/partition/constants.go
```diff
@@ -25,6 +25,7 @@ const (
 	FilesystemTypeZeroes FileSystemType = "zeroes"
 	FilesystemTypeXFS    FileSystemType = "xfs"
 	FilesystemTypeVFAT   FileSystemType = "vfat"
+	FileSystemTypeExt4   FileSystemType = "ext4"
```
frezbo, line 28: maybe split this into a different PR :thinking:

internal/pkg/partition/partition.go
```diff
@@ -90,6 +90,13 @@ func NewPartitionOptions(label string, uki bool) Options {
 			PartitionType:  LinuxFilesystemData,
 			Size:           0,
 		}
+	case constants.ImageCachePartitionLabel:
+		return Options{
+			FormatOptions:  *formatOptions,
+			PartitionLabel: label,
+			PartitionType:  LinuxFilesystemData,
+			Size:           0,
```
frezbo, line 98: should this be set to the size of the incoming cache?

internal/pkg/partition/constants.go
smira: should be good, it's only used in the installer for now

internal/pkg/partition/partition.go
smira: good question... I guess this makes sense (or specified as a parameter?)

frezbo: actually it shouldn't matter much here, probably the overall profile disk size should be bumped, another option is to increment the default with the size of cache , so it'll grow automatically

frezbo: ahh the image already supports setting disk size, we can document to bump it if using huge image cache

dsseng (approved):

smira (approved):
should be good to go, tested this locally (missing docs, but we miss them everywhere for the Image Cache)

PR #9779, opened 2024-11-21
feat: udev: label device nodes

dsseng:
Use udev rules to assign basic device file labels based on their subsystem


Dockerfile
```diff
@@ -731,6 +731,7 @@ COPY --chmod=0644 hack/containerd.toml /rootfs/etc/containerd/config.toml
 COPY --chmod=0644 hack/cri-containerd.toml /rootfs/etc/cri/containerd.toml
 COPY --chmod=0644 hack/cri-plugin.part /rootfs/etc/cri/conf.d/00-base.part
 COPY --chmod=0644 hack/udevd/80-net-name-slot.rules /rootfs/usr/lib/udev/rules.d/
+COPY --chmod=0644 hack/udevd/xx-selinux.rules /rootfs/usr/lib/udev/rules.d/
```
dsseng, line 734: In what place in order should we place it? Perhaps 99?

internal/pkg/selinux/policy/selinux/services/udev.cil
```diff
@@ -20,6 +20,20 @@
...
+; RTC is important for security verification
+(type rtc_device_t)
+(call protected_device_f (rtc_device_t))
+; Could be exposed firmware storage
+(type mtd_device_t)
+(call protected_device_f (mtd_device_t))
+; Could reset the system
+(type wdt_device_t)
+(call protected_device_f (wdt_device_t))
+; TODO: should pods be able to access TPM?
```
dsseng, line 33: Should they? If so, no reason to label TPM specially if we are not going to use its label.

Dockerfile
dsseng: Set 90 now

hack/udevd/90-selinux.rules
```diff
@@ -0,0 +1,11 @@
+SUBSYSTEM=="*",SECLABEL{selinux}="system_u:object_r:device_t:s0"
```
dsseng, line 1: Label all devices as device_t, overrides are lower in this file

Dockerfile
frezbo: 90 feels good, maybe we should also do a validation that the user doesn't override these?

hack/udevd/90-selinux.rules
frezbo, line 1: are these labels like standard? I guess devices can have standard labels

internal/integration/api/selinux.go
```diff
@@ -117,12 +117,25 @@ func (suite *SELinuxSuite) TestFileMountLabels() {
 	}
 	maps.Copy(expectedLabelsControlPlane, expectedLabelsWorker)
 
-	suite.checkFileLabels(workers, expectedLabelsWorker)
-	suite.checkFileLabels(controlplanes, expectedLabelsControlPlane)
+	// Devices labeled by subsystems, labeled by udev
+	expectedLabelsDevices := map[string]string{
+		"/dev/rtc0":      "system_u:object_r:rtc_device_t:s0",
+		"/dev/tpm0":      "system_u:object_r:tpm_device_t:s0",
```
frezbo, line 123: hmm `/dev/tpm` is not available by default on our qemu tests, right? only for secureboot i believe

dsseng: Yes. I added an option for test to ignore when file doesn't exist

hack/udevd/90-selinux.rules
dsseng: Apparently `(fsuse trans devtmpfs (system_u object_r device_t (systemLow systemLow)))` is not sufficient for this to be default

Dockerfile
dsseng: Well, user is the Talos admin, not the k8s admin who we protect the system from. Perhaps should be okay

frezbo: I guess it's fine, on second thoughts I believe we allow overwriting existing udev rules, so should be okay

internal/integration/api/selinux.go
frezbo: actually let's just enable tpm2 for all tests in CI, it shouldn't hurt, better than trying to match on error, probably can do in a different PR

frezbo (approved):

internal/integration/api/selinux.go
dsseng: Will do

PR #9783, opened 2024-11-22
chore: enable TPM2 by default in CI

dsseng:
With some tests requiring TPM and others benefitting from it (SELinux label test) it's better to always enable it as our CI has swtpm installed.

Do not automatically enable TPM2 in talosctl as it might be a breaking change to users who do not have swtpm available.

Partially reverts: a13f82c59456574238a75959ff395746c93f1cfa regarding tests

### DIFF
--- a/hack/test/e2e-image-factory.sh
+++ b/hack/test/e2e-image-factory.sh
@@ -24,7 +24,7 @@ case "${FACTORY_BOOT_METHOD:-iso}" in
     QEMU_FLAGS+=("--ipxe-boot-script=${FACTORY_SCHEME}://${PXE_FACTORY_HOSTNAME}/pxe/${FACTORY_SCHEMATIC}/${FACTORY_VERSION}/metal-amd64")
     ;;
   secureboot-iso)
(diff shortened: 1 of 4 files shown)


dsseng (commented):
From @frezbo:

> also we might need some tests with tpm2 disabled so that we know our skip PCR extension works

We need to know what test must be run without TPM2 available and add a param to them

hack/test/e2e-image-factory.sh
```diff
@@ -24,7 +24,7 @@ case "${FACTORY_BOOT_METHOD:-iso}" in
     QEMU_FLAGS+=("--ipxe-boot-script=${FACTORY_SCHEME}://${PXE_FACTORY_HOSTNAME}/pxe/${FACTORY_SCHEMATIC}/${FACTORY_VERSION}/metal-amd64")
     ;;
   secureboot-iso)
-    QEMU_FLAGS+=("--iso-path=${FACTORY_SCHEME}://${FACTORY_HOSTNAME}/image/${FACTORY_SCHEMATIC}/${FACTORY_VERSION}/metal-amd64-secureboot.iso" "--with-tpm2" "--encrypt-ephemeral" "--encrypt-state" "--disk-encryption-key-types=tpm")
```
smira, line 27: I'm not sure why it got removed here

dsseng: If we run all tests with TPM2 we don't need it here. However we shouldn't do it for all

smira: ah, I see it now

smira: we have at least in the `misc` section the test with UEFI disabled (BIOS mode), this could be a good place with disabled tpm2

PR #9798, opened 2024-11-26
fix: systemd-udevd restore old naming behavior

frezbo:
Restore old naming behavior as `eudevd`.

Ref:

* https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/
* https://www.freedesktop.org/software/systemd/man/latest/systemd-udevd.service.html
* https://www.freedesktop.org/software/systemd/man/latest/systemd.link.html#


smira (approved):

Towerful:
does this patch negate the note in the upgrade docs:  
https://www.talos.dev/v1.9/talos-guides/upgrading-talos/#before-upgrade-to-v192

I'm on 1.9.1 bare metal, and I would like to undestand how the changing of predictable network names might impact my config for:  

```yaml
bond:
  interfaces:
    - enp2s0f0np0
    - enp2s0f1np1
```

should I upgrade a single node, and see if the name changes and I have to rework machive configs?

smira:
No, it doesn't on its own.  This change makes names same as they were before.

Names _shouldn't_ change in general, and Talos in general can pick up "old" names, so it should be good.

PR #9801, opened 2024-11-26
fix: small logrus fixes

DmitriyMV:
Ensure correct logrus setup.


pkg/machinery/resources/time/condition.go
```diff
@@ -34,6 +34,7 @@ func (condition *SyncCondition) Wait(ctx context.Context) error {
 	_, err := condition.state.WatchFor(
 		ctx,
 		resource.NewMetadata(v1alpha1.NamespaceName, StatusType, StatusID, resource.VersionUndefined),
+		state.WithEventTypes(state.Created, state.Updated),
```
dsseng, line 37: is this relevant?

dsseng (approved):

pkg/machinery/resources/time/condition.go
DmitriyMV: It's a by product of today fixes :)

PR #9803, opened 2024-11-26
fix: fix e2e test with extensions and SELinux

dsseng:
Skip SELinux test when system extensions are active, we'll work on that in a bit. Also elaborate more on SELinux test errors.

### DIFF
--- a/internal/integration/api/selinux.go
+++ b/internal/integration/api/selinux.go
@@ -71,8 +71,10 @@ func (suite *SELinuxSuite) getLabel(nodeCtx context.Context, pid int32) string {
 
 // TestFileMountLabels reads labels of runtime-created files and mounts from xattrs
 // to ensure SELinux labels for files are set when they are created and FS's are mounted with correct labels.
-// FIXME: cancel the test in case system was upgraded.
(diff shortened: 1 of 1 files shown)


internal/integration/api/selinux.go
```diff
@@ -152,6 +153,13 @@ func (suite *SELinuxSuite) checkFileLabels(nodes []string, expectedLabels map[st
 			suite.T().Skip("skipping SELinux test since SELinux is disabled")
 		}
 
+		// TODO: do not skip the test if system extensions are enabled
```
frezbo, line 156: probably easier to check if either `extensions.qemu` or `extensions.nvidia` is set

dsseng: But what if we just run locally? Well, for CI that'd suffice

frezbo: same, those tests gets enabled only if the flag is passed

smira:
Actually this test fails for more integration tests, so I feel it should be skipped until we have a PR which re-enables it and the PR is green on all pipelines. As we're close the the release, we need to have tests working.

PR #9805, opened 2024-11-26
Update local-storage.md

shtouff:
Add an explicit example of kubelet config for the `local-path-provisioner` paragraph.

# Pull Request

## What? (description)

I think the local storage page lacks a clear example of how to integrate `local-path-provisioner`

## Why? (reasoning)

While playing with `local-path-provisioner` on my lab cluster, I had sometimes some dataloss I couldn't explain. After long troubleshooting, I identified some strange behaviour when using `volumeMount` with the `subPath` statement.

I initially thought it could a bug in the provisioner itself, and I proposed an extra test there (https://github.com/rancher/local-path-provisioner/pull/467).

After a while, and lots of tests, I ended up trying to replicate the kubelet config which is shown in the `hostPath` paragraph, and everything went back to the expected behaviour.

Afterwards, it may be seen as obvious, but I really thought that `hostPath` and `local-path-provisioner` were 2 distinct paragraphs, hence my mistake.

Feel free to reject if you think it's a duplicate and/or unnecessary.

## Acceptance

- [ x] you linked an issue (if applicable)
- [ x] you generated documentation (`make docs`)


smira (approved):
thank you, that's a nice catch!

Issue #9808, opened 2024-11-27
save full logs in the crashdump (ci)


smira:
Before changes in #9667 and related PRs, the full logs were saved including Docker logs (for Docker-based) and console logs (for QEMU-based).

Proposal: add to the `support.zip` (or another file) a dump of all console logs resurrecting code that was there for docker & vm provisioners.

PR #9828, opened 2024-11-28
chore: support saving cluster logs on destroy

frezbo:
Support saving cluster logs on destroy

Fixes: #9808


smira (approved):

cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -161,7 +161,6 @@ var (
 	dockerHostIP              string
 	withInitNode              bool
 	customCNIUrl              string
-	crashdumpOnFailure        bool
```
DmitriyMV, line 164: Isn't this an orthogonal change?

pkg/cluster/logsaarchive.go
```diff
@@ -0,0 +1,77 @@
...
+
+	gzipWriter := gzip.NewWriter(logFile)
+	defer gzipWriter.Close() //nolint:errcheck
+
+	r := filemap.Build(logFileMap)
+
+	if _, err := io.Copy(gzipWriter, r); err != nil {
+		fmt.Fprintf(os.Stderr, "error writing log archive: %s\n", err)
+
+		gzipWriter.Close() //nolint:errcheck
```
DmitriyMV, line 75: No need to double close it on error, because defer will work regardless.

pkg/cluster/logsaarchive.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/siderolabs/talos/pkg/imager/filemap"
+)
+
+// SaveClusterLogsArchive saves all logs from the cluster state directory to a gzip archive.
+func SaveClusterLogsArchive(statePath, archivePath string) {
```
DmitriyMV, line 19: I think it's better create a separate version of this function that returns an error, and move all print-on-error calls here, inside this variant.

You could also wrap all closers inside sync.OnceValue and call them at the final `return cmp.Or(closer1(), closer2(),...)`

pkg/provision/providers/docker/destroy.go
```diff
@@ -25,30 +30,36 @@ func (p *provisioner) Destroy(ctx context.Context, cluster provision.Cluster, op
...
 	}
 
+	stateDirectoryPath, err := cluster.StatePath()
+	if err != nil {
+		return err
+	}
+
 	complete := false
-	deleteStateDirectory := func(shouldDelete bool) error {
+	deleteStateDirectory := func(stateDir string, shouldDelete bool) error {
```
DmitriyMV, line 39: Wrap it with sync.OnceValue because we are already here.

DmitriyMV (changes requested):

cmd/talosctl/cmd/mgmt/cluster/create.go
frezbo: yeh, since we always call destroy, let it save all logs

DmitriyMV (approved):

PR #9829, opened 2024-11-28
refactor: run kubelet and etcd in system containerd

dsseng:
This change allows us to ensure their contexts are protected by SELinux policy, as well as making CRI containerd only host containers managed by Kubernetes.


smira:
I'm not sure if we can do that, as system containerd runs from `tmpfs`, so all container state and images should be in memory.

dsseng:
Well, etcd data is on a mount to where it belongs to be

frezbo:
> Well, etcd data is on a mount to where it belongs to be

but the uncompressed layers and other metadata will fillup tmpfs, just wasting memory

dsseng:
yes, perhaps this solution is not really efficient since those images aren't tiny

dsseng:
Idea: we could create a namespace belonging to the system containerd, but with snapshotter configured to store data on the `/var` partition? This should bring together best of both concepts.

smira:
> Idea: we could create a namespace belonging to the system containerd, but with snapshotter configured to store data on the `/var` partition? This should bring together best of both concepts.

that sounds interesting, if we can actually pull it through, but still it's kubelet/etcd only

PR #9832, opened 2024-11-28
chore: deprecate cgroupsv1 in non-container mode

dsseng:
Also fix a typo.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -28,6 +28,12 @@ preface = """
 Talos is built with Go 1.23.3.
 """
 
+    [notes.cgroupsv1]
+        title = "cgroups version 1"
+        description = """\
+Support for cgroupsv1 is deprecated, and will be removed in Talos 1.10 (for non-container mode).
+"""
+
     [notes.machineConfigKubeAPIServer]
         title = "Kube APIServer Authorization Config"
         description = """\
--- a/internal/pkg/mount/v2/cgroups.go
+++ b/internal/pkg/mount/v2/cgroups.go
@@ -14,14 +14,14 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
-// ForceGGroupsV1 returns the cgroup version to be used (only for !container mode).
-func ForceGGroupsV1() bool {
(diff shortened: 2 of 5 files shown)


internal/pkg/mount/v2/cgroups.go
```diff
@@ -14,14 +15,15 @@ import (
...
+// ForceCGroupsV1 returns the cgroup version to be used (only for !container mode).
+func ForceCGroupsV1() bool {
 	return pointer.SafeDeref(procfs.ProcCmdline().Get(constants.KernelParamCGroups).First()) == "0"
 }
 
 // CGroupMountPoints returns the cgroup mount points.
 func CGroupMountPoints() Points {
-	if ForceGGroupsV1() {
+	if ForceCGroupsV1() {
+		log.Println("cgroupsv1 is deprecated and will not be supported in the 1.10 release other than in the container mode")
```
smira, line 26: this logger is not as nice, can you please move this method to startup tasks which calls this method? (it has zap logger)

website/content/v1.9/reference/kernel.md
```diff
@@ -233,6 +233,8 @@ Valid options are:
 
 #### `talos.unified_cgroup_hierarchy`
 
+> Deprecated: From the 1.10 release it is planned that `cgroupsv1` will only be supported in the container mode.
```
smira, line 236: please add a note about this to `hack/release.toml` as well

hack/release.toml
```diff
@@ -25,6 +25,13 @@ preface = """
 * runc: 1.2.1
 
 Talos is built with Go 1.23.3.
+"""
+
+    [notes.cgroupsv1]
+        title = "cgroups version 1 deprecation"
```
smira, line 31:
```suggestion
        title = "cgroups version 1"
```

hack/release.toml
```diff
@@ -25,6 +25,13 @@ preface = """
...
+        description = """\
+Add a warning when `talos.unified_cgroup_hierarchy=0` parameter is passed to notify users of the deprecation.
```
smira, line 33: Support for `cgroupsv1` is deprecated, and will be removed in Talos 1.10 (for non-container mode).

PR #9834, opened 2024-11-28
feat: support overriding base OCI spec for CRI

### LINKED ISSUE #9827: allow to patch containerd CRI base OCI spec
From containerd config:

```
      # base_runtime_spec is a file path to a JSON file with the OCI spec that will be used as the base spec that all
      # container's are created from.
      # Use containerd's `ctr oci spec > /etc/containerd/cri-base.json` to output initial spec file.
      # Spec files are loaded at launch, so containerd daemon must be restarted on any changes to refresh default specs.
      # Still running containers and restarted containers will still be using the original spec from which that container was created.
      base_runtime_spec = ""
```

Idea: Talos generates base OCI spec, and merge machine configuration patches supplied by the user, building a final base OCI spec which will be used by the CRI.

This allows e.g. to change default ulimits for containers, or in general any OCI settings which are not configurable in Kubernetes.

smira:
Fixes #9827

### DIFF
--- a/internal/integration/api/common.go
+++ b/internal/integration/api/common.go
@@ -9,9 +9,12 @@ package api
 import (
 	"context"
 	"strings"
+	"testing"
 	"time"
 
 	"github.com/siderolabs/talos/internal/integration/base"
+	"github.com/siderolabs/talos/pkg/machinery/config/machine"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/v1alpha1"
 )
 
 // CommonSuite verifies some default settings such as ulimits.
@@ -30,7 +33,7 @@ func (suite *CommonSuite) SuiteName() string {
 // SetupTest ...
 func (suite *CommonSuite) SetupTest() {
 	// make sure API calls have timeout
-	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 10*time.Minute)
+	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 15*time.Minute)
 }
 
 // TearDownTest ...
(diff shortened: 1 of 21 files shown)


internal/integration/api/common.go
```diff
@@ -157,6 +160,105 @@ func (suite *CommonSuite) TestDNSResolver() {
...
+	node := suite.RandomDiscoveredNodeInternalIP(machine.TypeWorker)
+
+	k8sNode, err := suite.GetK8sNodeByInternalIP(suite.ctx, node)
+	suite.Require().NoError(err)
+
+	nodeName := k8sNode.Name
+
+	suite.T().Logf("adjusting base OCI specs on %s/%s", node, nodeName)
+
+	suite.AssertRebooted(
```
frezbo, line 182: can't we just restart containerd rather than a reboot?

internal/integration/api/common.go
```diff
@@ -157,6 +160,105 @@ func (suite *CommonSuite) TestDNSResolver() {
...
+	suite.Require().NoError(ociUlimits2PodDef.Create(suite.ctx, 5*time.Minute))
+
+	stdout, stderr, err = ociUlimits2PodDef.Exec(
+		suite.ctx,
+		"ulimit -n",
+	)
+	suite.Require().NoError(err)
+
+	suite.Require().Equal("", stderr)
+	suite.Require().Equal("1048576\n", stdout)
```
frezbo, line 257: i think we have some other code also that verifies default `ulimit` that also needs updating i guess (if that and this runs parallel?)

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -290,6 +290,18 @@ type MachineConfig struct {
...
+	//
+	//    It can be used to set some default container settings which are not configurable in Kubernetes,
+	//    for example default ulimits.
+	//    Note: this change applies to all newly created containers, and it requires a reboot to take effect.
+	//  examples:
+	//    - name: override default open file limit
+	//      value: machineBaseRuntimeSpecOverridesExample()
+	//  schema:
+	//    type: object
+	MachineBaseRuntimeSpecOverrides Unstructured `yaml:"baseRuntimeSpecOverrides,omitempty"`
```
frezbo, line 303: does this apply for both cri and containerd? I guess we only need for cri?

frezbo (approved):

internal/integration/api/common.go
smira: it's in the same file, but they run sequentially

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
smira: only for CRI

internal/integration/api/common.go
smira: so this test first adjusts rlimits, and then reverts it back (and verifies it)

internal/integration/api/common.go
smira: Right now - no, we don't have anything that controls CRI

frezbo: we can fix for 1.10

smira:
Updated with docs and some test cleanups

frezbo (approved):
:cool:

PR #9844, opened 2024-11-29
feat: add support for custom search domains

### LINKED ISSUE #9342: Specify additional DNS search domains
## Feature Request

It should be possible to add arbitrary search domains to the Talos + k8s DNS configurations. This is possible at the moment only by making the node name an FQDN built on the search domain.

DmitriyMV:
New config field `machine.network.searchDomains` supports specifying custom search domains.

For the node it will look something like this:
```
nameserver 127.0.0.53

search my-custom-search-name.com my-custom-search-name2.com
```

For the pods it will look something like this:

```
search default.svc.cluster.local svc.cluster.local cluster.local my-custom-search-name.com my-custom-search-name2.com
nameserver 10.96.0.10
options ndots:5
```

Fixes #9342


internal/app/machined/pkg/controllers/network/etcfile_test.go
```diff
@@ -223,23 +223,27 @@ func (suite *EtcFileConfigSuite) testFiles(resources []resource.Resource, conten
 }
 
 func (suite *EtcFileConfigSuite) TestComplete() {
+	suite.resolverStatus.TypedSpec().SearchDomains = []string{"foo.example.com"}
+
 	suite.testFiles(
 		[]resource.Resource{suite.cfg, suite.defaultAddress, suite.hostnameStatus, suite.resolverStatus, suite.hostDNSConfig},
```
smira, line 229: I guess `hostnameStatus` should be dropped from the test

smira (approved):

internal/app/machined/pkg/controllers/network/etcfile_test.go
DmitriyMV: @smira Nope, I can't even drop `HostnameStatuses.net.talos.dev` resource input from `network.EtcFileController` since it's used in `network.(*EtcFileController).renderHosts` method.

hack/release.toml
```diff
@@ -132,6 +132,30 @@ The new command `talosctl wipe disk` allows to wipe a disk or a partition which
...
 Talos now allows to [modify the OCI base runtime spec for the container runtime](https://www.talos.dev/v1.9/advanced/oci-base-spec/).
+"""
+
+    [notes.custom-search-domains]
+        title = "Custom search domains for Talos nodes"
+        description = """\
+Talos now allows to supports specifying custom search domains for Talos nodes using
+new config field `machine.network.searchDomains`
+
+For the node it will look something like this:
```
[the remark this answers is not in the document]
DmitriyMV: ✅

DmitriyMV:
@flpajany greetings! Thanks for the request. Unfortunately we do not back-port changes that modify machine config or proto resources. Those backports leaves a big chance of accidentally breaking things.

flpajany:
I understand. Is there a way to change the seach path for DNS resolution manually ?

smira:
> I understand. Is there a way to change the seach path for DNS resolution manually ?

This PR is mostly about host search path in the `/etc/resolv.conf`.

See https://github.com/siderolabs/talos/issues/9342 for discussion for K8s pods and others.

flpajany:
It is exactly what I am talking about : a way to change /etc/resolv.conf and add other domains to the search path to the host. 

I was thinking about changing machineconfig to add something like this : 

```
machine:
    files:
        - path: /etc/resolv.conf
          op: append
          content: 'search mydomain.com'
          permissions: 0o644
```

smira:
this won't work most probably. 

you can achieve the same by setting the hostname to the fqdn as well.

flpajany:
> this won't work most probably.
> 
> you can achieve the same by setting the hostname to the fqdn as well.

I understand but unfortunatly it is complicated to do that for some applications. We'll try nonetheless to change it that way and I'll be back here to tell you the result.

smira:
Also keep in mind that k8s pod search path is totally independent of this change.

flpajany:
That would mean to rebuild container images, thing we are not able to do :-(

smira:
Please read #9342 carefully. If you have any specific question - create a GitHub discussion with all relevant information.

flpajany:
OK we'll try this too. Thx

PR #9845, opened 2024-11-29
feat: support image cache copying

### LINKED ISSUE #9615: copying image cache
On install, the image cache should be copied from the ephemeral installation media (e.g. ISO) to the disk.

Proposed flow:

* a user should define a user volume which will contain the image cache on disk, it will be prepared by Talos
* an explicit copy (merge) should happen from discovered image caches to the user volume (?)

smira:
Fixes #9615

The are no integration tests, this is to be addressed later. I did manual tests so far.

Also includes first draft of the documentation.


internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -326,3 +440,105 @@ func (ctrl *ImageCacheConfigController) getImageCacheRoot(ctx context.Context, r
...
+}
+
+func copyFileSafe(src, dst string) error {
+	srcStat, err := os.Stat(src)
+	if err != nil {
+		return fmt.Errorf("error getting source file info: %w", err)
+	}
+
+	dstStat, err := os.Stat(dst)
+	if err == nil && srcStat.Size() == dstStat.Size() {
```
frezbo, line 510: two files could haave the same size, but hashing is expensive

smira: that would be outside of our control I guess, we just try to skip copying what is already there

website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+```bash
+talosctl images default > images.txt
+cat extra-images.txt >> images.txt
+```
+
+Next, prepare an OCI tarball which contains all cached images:
+
+```bash
+cat images.txt | talosctl images cache-create --image-cache-path image-cache.tar --images=-
```
frezbo, line 24: i
```suggestion
cat images.txt | talosctl images cache-create --image-cache-path <directory to store cache> --images=-
```

maybe?

website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+> Note: The `cache-create` supports a `--layer-cache` flag to additionally cache the pulled images layers on the filesystem.
+> This is useful to speed up repeated calls for `cache-create` with the same images.
+
+The image cache tarball might be used as a file (`image-cache.tar`) or pushed itself to a container registry of your choice.
```
frezbo, line 30: same here, it's a folder

website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -0,0 +1,124 @@
...
+
+### ISO
+
+In case of ISO, the image cache is bundled with a Talos ISO image, it will be available for the initial install and (if configured) copied to the
+disk during the installation process.
+
+The ISO image can built with the [imager]({{< relref "../install/boot-assets#imager" >}}) by passing an additional `--image-cache` flag:
+
+```bash
+docker run --rm -t -v $PWD/_out:/secureboot:ro -v $PWD/_out:/out -v /dev:/dev --privileged ghcr.io/siderolabs/imager:{{< release >}} iso --image-cache ./image-cache.tar
```
frezbo, line 45: probably add a note that `_out` should contain the folder

website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -0,0 +1,124 @@
...
+> Note: If the image cache was pushed to a container registry, the `--image-cache` flag should point to the image reference.
+
+### Disk Image
+
+In case of disk image, the image cache is included in the disk image itself, and on boot it would be used immediately by the Talos.
+
+The disk image can be built with the [imager]({{< relref "../install/boot-assets#imager" >}}) by passing an additional `--image-cache` flag:
+
+```bash
+docker run --rm -t -v $PWD/_out:/secureboot:ro -v $PWD/_out:/out -v /dev:/dev --privileged ghcr.io/siderolabs/imager:{{< release >}} metal --image-cache ./image-cache.tar
```
frezbo, line 57: also add a note on bump disk-image-size

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -326,3 +440,105 @@ func (ctrl *ImageCacheConfigController) getImageCacheRoot(ctx context.Context, r
...
+	}
+
+	srcFile, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("error opening source file: %w", err)
+	}
+
+	defer srcFile.Close() //nolint:errcheck
+
+	tempPath := dst + ".tmp"
```
DmitriyMV, line 522: Any reason to create `tmp` file without deleting it in the errors case?

smira: I'm not sure if we care about leftover `.tmp` files? The copy process will be retried automatically, and it will overwrite the file if possible.

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -326,3 +440,105 @@ func (ctrl *ImageCacheConfigController) getImageCacheRoot(ctx context.Context, r
...
+	logger.Info("copying image cache", zap.String("source", source), zap.String("target", target))
+
+	if ctrl.DisableCacheCopy {
+		// used for testing
+		return nil
+	}
+
+	var bytesCopied int64
+
+	if err := filepath.WalkDir(source, func(path string, d fs.DirEntry, err error) error {
```
DmitriyMV, line 454: I guess you could use `os.CopyFS` here and wrap `fs.FS.Open` interface method to record `bytesCopied`.

smira: see the docs: https://pkg.go.dev/os#CopyFS

it doesn't do what we need in case of overwrites/same files

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: added `mkdir`

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: fixed it up

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: fixed it up

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: note added

frezbo (approved):
:cool:

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -287,16 +293,116 @@ func (ctrl *ImageCacheConfigController) createVolumeConfigDisk(ctx context.Conte
...
 	}
 
+	return &imageCacheVolumeStatus{
+		roots:      roots,
+		allReady:   allReady,
+		copyStatus: copyStatus,
+	}, nil
+}
+
+func (ctrl *ImageCacheConfigController) getImageCacheRoot(ctx context.Context, r controller.Reader, volumeStatus *block.VolumeStatus, mountReadyOnly bool) (optional.Optional[string], bool, error) {
```
DmitriyMV, line 405: Why return an `Optional` if you also return a bool value?

DmitriyMV: Ignore. Found my answer.

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -287,16 +293,116 @@ func (ctrl *ImageCacheConfigController) createVolumeConfigDisk(ctx context.Conte
...
+			case VolumeImageCacheISO:
+				isoReady = true
+				copySource = root.ValueOr("")
+			case VolumeImageCacheDISK:
+				diskReady = true
+				copyTarget = root.ValueOr("")
+			}
+		}
+
+		allReady = allReady && ready
```
DmitriyMV, line 367: Since there could only be `VolumeImageCacheDISK` and `VolumeImageCacheISO` volume statuses, doesn't it make sense to move and redo `allReady = allReady && ready`  as `allReady = isoReady && diskReady`? Or I'm missing something.

DmitriyMV: I guess we can also define `copyTarget` and `copySource` as `optional.Optional[string]` and drop `diskReady` and `isoReady` entirely.

smira: I want to leave some room for expansion and support other read-only sources

internal/app/machined/pkg/controllers/cri/image_cache_config.go
smira: because they are not same if you look closely.

It might be ready but missing path. 

Readiness is another dimension, it is ready if e.g. it was found to be missing, so the path is missing, but it's ready in the sense we don't need to wait for it

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -287,16 +293,116 @@ func (ctrl *ImageCacheConfigController) createVolumeConfigDisk(ctx context.Conte
...
+
+	roots := make([]string, 0, len(volumeIDs))
+
+	var (
+		allReady, isoReady, diskReady bool
+		copySource, copyTarget        string
+	)
+
+	// analyze volume statuses, and build the roots
+	for _, volumeStatus := range volumeStatuses {
```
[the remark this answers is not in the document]
smira: yes, there might be just disk image cache. isoPresent is only important for copying.

DmitriyMV (approved):
🆒 but `cri.(*ImageCacheConfigController).copyImageCache` has ~~two~~ too many states and it's not exactly clear how it works. I think it could be simplified, but I don't have any really good ideas right now.

PR #9848, opened 2024-11-29
feat: add filesystem scrubbing controller

### LINKED ISSUE #8314: fstrim and monitoring for disks
## Feature Request
Support periodic FS trimming would not only help bare metal installs, but also thin-provisioned VMs.

hwmon could be useful as well for admins to be able to get CPU/disk/mainboard temperature etc.

### Description
Basically the request. Unsure if these can be enabled via extensions, but perhaps these should be available by default, at least discards/fstrim. Maybe this is achievable using stuff like discard=async as well, but some admins use periodic TRIM.

I could possibly develop fstrim stuff if I have time

### LINKED ISSUE #9545: xfs_scrub controller
## Feature Request

Implement a Talos controller for periodically running `xfs_scrub`. This also needs to be configurable by a machine configuration document.

#### References

- https://man7.org/linux/man-pages/man8/xfs_scrub.8.html
- https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/scrub

dsseng:
# TODO
- [x] factor out Task + TasksController to decouple actual execution from scrub scheduling logic
- [x] tests for scrub controller (unit tests based on task resources)
- [x] report status for monitoring of scheduled scrubs, most recent scrubs date, duration and result.

Fixes #9545


pkg/machinery/config/types/runtime/fs_scrub.go
```diff
@@ -0,0 +1,157 @@
...
+	_ config.Validator     = &FilesystemScrubV1Alpha1{}
+)
+
+// Timeout constants.
+const (
+	MinScrubPeriod     = 10 * time.Second
+	DefaultScrubPeriod = 24 * 7 * time.Hour
+)
+
+// FilesystemScrubV1Alpha1 is a filesystem scrubbing config document.
```
dsseng, line 48: Docs must mention this feature is experimental and perhaps link to kernel docs about this

internal/app/machined/pkg/controllers/runtime/fs_scrub.go
```diff
@@ -0,0 +1,244 @@
...
+
+			// When scheduling the first scrub, we use a random time to avoid all scrubs running in a row.
+			// After the first scrub, we use the period defined in the config.
+			cb := func() {
+				logger.Warn("!!! scrub !!! ding", zap.String("path", mountpoint))
+				ctrl.c <- mountpoint
+				ctrl.schedule[mountpoint].timer.Reset(ctrl.schedule[mountpoint].period)
+			}
+
+			ctrl.schedule[mountpoint] = scrubSchedule{
```
dsseng, line 196: I guess we need to somehow protect from user creating multiple documents referencing the same mountpoint.

internal/app/machined/pkg/controllers/runtime/fs_scrub.go
```diff
@@ -0,0 +1,244 @@
...
+		runner.WithLoggingManager(ctrl.Runtime.Logging()),
+		runner.WithEnv(environment.Get(ctrl.Runtime.Config())),
+		runner.WithOOMScoreAdj(-999),
+		runner.WithDroppedCapabilities(constants.XFSScrubDroppedCapabilities),
+		runner.WithPriority(19),
+		runner.WithIOPriority(runner.IoprioClassIdle, 7),
+		runner.WithSchedulingPolicy(runner.SchedulingPolicyIdle),
+	)
+
+	return r.Run(func(s events.ServiceState, msg string, args ...any) {})
```
smira, line 243: I wonder if we should have a way to cancel running scrub process?

smira (commented):
as `xfs_scrub` is experimental, I'd say let's do it the first thing we merge in 1.10

internal/app/machined/pkg/controllers/runtime/fs_scrub.go
dsseng: We can add a channel to signal this once document is deleted. However I haven't studied whether or not it's okay to abort the process and does it terminate safely

dsseng:
Other things to consider:
- add an option to scrub on boot
- check whether or not scrub may be aborted. If not ensure it's not and we must delay reboot/poweroff to ensure this

pkg/machinery/constants/constants.go
```diff
@@ -1277,6 +1277,38 @@ var DefaultDroppedCapabilities = map[string]struct{}{
 	"cap_sys_module": {},
 }
 
+// XFSScrubDroppedCapabilities is the set of capabilities to drop for xfs_scrub.
```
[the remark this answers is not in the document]
dsseng: Yes, I also want to do such a thing on the level of our interface to libcap. Actually systemd does manage such a thing, since in xfs scrubbing service there is a list of capabilities to give, not take

pkg/machinery/resources/runtime/fs_scrub_config.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// FSScrubConfigType is type of FSScrubConfig resource.
+const FSScrubConfigType = resource.Type("FSScrubConfigs.runtime.talos.dev")
```
smira, line 19: I wonder if we should move this under `block.` (both resources and config)

internal/app/machined/pkg/controllers/runtime/fs_scrub.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+	}
+
+	defer stopTimers()
+
+	ctrl.schedule = make(map[string]scrubSchedule)
+	ctrl.status = make(map[string]scrubStatus)
+	ctrl.c = make(chan string, 5)
+
+	for {
```
smira, line 104: I almost wonder if we should split up this controller into two controllers:

* one which reads FSScrubConfig and outputs FSScrubSchedule resources (builds a schedule, handles jitter, updates schedule, etc.) - this controller can be fully unit-tested
* another which reads FSScrubSchedule, and based on the schedule runs the actual xfs_scrub tasks (or cancels them)

We can test time-based stuff using fake clocks, example https://github.com/siderolabs/talos/blob/f75604313d535180c38b33df53253ad4acba2ec1/internal/app/machined/pkg/controllers/runtime/cri_image_gc_test.go#L33-L52

dsseng: Yes, sane. We already have the config controller which populates the structures looking the same as the yaml source, why not handle scheduling there, yes

smira: No, no... config controller is good, don't touch it ;)

Schedule controller will be fully testable without mocks.

Scrub controller will need a clock mock and an actual scrub mock, and it can be fully tested once again by manipulating schedules.

pkg/machinery/config/types/runtime/fs_scrub.go
dsseng: Not anymore, Linux 6.18 now includes this without experimental status

dsseng:
TODO:
- [ ] fix test flakes
- [ ] add SELinux rules for xfs_scrub, test in enforcing mode

api/resource/definitions/runtime/runtime.proto
```diff
@@ -209,6 +210,24 @@ message ServicePIDSpec {
   string mount_namespace = 2;
 }
 
+// TaskSpec describes a background task to be run by a schedule.
+message TaskSpec {
+  string id = 1;
+  string owner = 2;
+  repeated string args = 3;
+  string selinux_label = 4;
```
dsseng: Add capabilities to be revoked/granted? Not useful yet, but might be when more different background tasks appear

api/resource/definitions/runtime/runtime.proto
```diff
@@ -209,6 +210,24 @@ message ServicePIDSpec {
...
+}
+
+// TaskStatusSpec describes configuration of watchdog timer.
```
dsseng:
```suggestion
// TaskStatusSpec describes result of a background task.
```

smira:
some steps:

- we need a new process runner extracted from `service/runners/process.go` which is self-contained, and use it to implement `service/runners/process.go`
- caps: default to drop all

pkg/machinery/constants/constants.go
```diff
@@ -1501,6 +1511,16 @@ var DefaultDroppedCapabilities = map[string]struct{}{
...
 
+// XFSScrubCapabilities is the set of capabilities xfs_scrub runs with (matching the upstream
+// xfs_scrub systemd service); all other capabilities are dropped (additive model).
+var XFSScrubCapabilities = []string{
+	"cap_dac_override",
+	"cap_dac_read_search",
+	"cap_fowner",
+	"cap_sys_admin",
+	"cap_sys_rawio",
+}
```
dsseng: Experimentally proven that just sys_admin is enough, but I'm trying to break it up into more specific caps with hopes to run a really minimized capability set

dsseng: Not really at least now: https://github.com/torvalds/linux/blob/7d0a66e4bb9081d75c82ec4957c50034cb0ea449/fs/xfs/scrub/scrub.c#L808-L809

dsseng (commented):
worth a release notes point, since a new feature is going to be auto-enabled

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/scrub.go
```diff
@@ -0,0 +1,50 @@
...
+// Scrubbing is enabled by default with the default interval; the global document adjusts or
+// disables it for all volumes, and the per-volume override takes precedence over both.
+//
+// It returns enabled=false (and interval=0) when scrubbing is disabled for the volume.
+func ResolveScrub(cfg configconfig.Config, scrubConfig configconfig.VolumeScrubConfigProvider) (enabled bool, interval time.Duration) {
+	if cfg == nil {
+		return false, 0
+	}
+
+	enabled = true
```
smira: do not enable by default if not set in the machine config (there is no document)

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/scrub.go
```diff
@@ -0,0 +1,50 @@
...
+	interval = constants.DefaultFilesystemScrubInterval
```
smira: or interval should be zero?

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -188,8 +220,17 @@ func (p *processRunner) build(extraLogWriter io.Writer) (commandWrapper, error)
 	}
 
 	// reduce capabilities and assign them to launcher
-	if err := dropCaps(p.opts.DroppedCapabilities, launcher); err != nil {
-		return commandWrapper{}, err
+	switch {
+	case p.opts.AllowedCapabilities != nil && p.opts.DroppedCapabilities != nil:
```
smira: I wonder if we should just use `DroppedCapabilities` and compute the set via `cap.Known()` ?

smira: this would allow us not to implement another branch with `keepCaps`

smira (commented):
we don't (we shouldn't) enable xfs_scrub by default, but in the e2e-qemu I think we should inject xfs_scrub enabled, we could add the doc to `hack/test/patches/ephemeral-nvme.yaml`

internal/app/machined/pkg/controllers/block/fs_scrub.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+// runXFSScrub runs xfs_scrub on the mounted filesystem at the given target path.
+//
+// The process runs with minimal privileges (only the capabilities xfs_scrub needs) and
+// the lowest CPU/IO priority; it is stopped when the context is canceled.
+func (ctrl *FSScrubController) runXFSScrub(ctx context.Context, logger *zap.Logger, target string) error {
+	caps := map[string]struct{}{}
+	for _, capName := range cap.Known() {
+		capName := strings.ToLower(capName)
```
dsseng, line 412: looks a little bit hacky, yet works (containerd package uses uppercase while libcap fails to parse anything but lowercase)

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/scrub.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	interval = 0
+
+	if globalScrub := cfg.FilesystemScrubConfig(); globalScrub != nil {
+		interval = globalScrub.Interval().ValueOr(interval)
+	}
+
+	enabled = interval > 0
+
```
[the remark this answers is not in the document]
dsseng: I should probably remove the `Enabled` field, and treat presence of the document as enabled (with a default interval set therefore)

Will do later today

pkg/machinery/config/types/block/fs_scrub.go
```diff
@@ -0,0 +1,129 @@
...
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/FilesystemScrubConfig
+type FilesystemScrubConfigV1Alpha1 struct {
+	meta.Meta `yaml:",inline"`
+
+	//   description: |
+	//     Enable or disable periodic filesystem scrubbing.
+	//
+	//     If not set, scrubbing is enabled by default.
+	ScrubEnabled *bool `yaml:"enabled,omitempty"`
```
smira: let's drop enabled to keep it in sync with FilesystemTrimConfig

smira: (enabled: document exists && interval > 0)

smira (approved):

PR #9862, opened 2024-12-02
feat: add more scheduling options for process runner

dsseng:
This allows us to run low-priority background tasks like xfs_scrub only while idle.

Example usage in #9848

- **chore: process: minor capability cleanup**
- **feat: allow running processes with custom priority**
- **feat: allow setting ionice on processes**
- **feat: allow setting scheduling class on processes**

### DIFF
--- a/internal/app/machined/pkg/system/runner/process/process.go
+++ b/internal/app/machined/pkg/system/runner/process/process.go
@@ -13,9 +13,9 @@ import (
 	"os"
 	"path"
 	"slices"
-	"strings"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/containerd/cgroups/v3"
 	"github.com/containerd/cgroups/v3/cgroup1"
@@ -104,26 +104,21 @@ type commandWrapper struct {
 }
 
 func dropCaps(droppedCapabilities []string, launcher *cap.Launcher) error {
-	droppedCaps := strings.Join(droppedCapabilities, ",")
-
-	if droppedCaps != "" {
-		caps := strings.Split(droppedCaps, ",")
-		dropCaps := xslices.Map(caps, func(c string) cap.Value {
-			capability, capErr := cap.FromName(c)
-			if capErr != nil {
-				panic(fmt.Errorf("failed to parse capability: %s", capErr))
-			}
-
-			return capability
-		})
-
-		iab := cap.IABGetProc()
-		if err := iab.SetVector(cap.Bound, true, dropCaps...); err != nil {
-			return fmt.Errorf("failed to set capabilities: %w", err)
+	dropCaps := xslices.Map(droppedCapabilities, func(c string) cap.Value {
+		capability, capErr := cap.FromName(c)
+		if capErr != nil {
+			panic(fmt.Errorf("failed to parse capability: %s", capErr))
 		}
 
-		launcher.SetIAB(iab)
+		return capability
+	})
+
+	iab := cap.IABGetProc()
+	if err := iab.SetVector(cap.Bound, true, dropCaps...); err != nil {
+		return fmt.Errorf("failed to set capabilities: %w", err)
 	}
+
+	launcher.SetIAB(iab)
 
 	return nil
 }
@@ -309,7 +304,7 @@ func (p *processRunner) build() (commandWrapper, error) {
 
 // Apply cgroup and OOM score after the process is launched.
 //
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func applyProperties(p *processRunner, pid int) error {
 	if p.opts.CgroupPath != "" {
 		path := cgroup.Path(p.opts.CgroupPath)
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -353,6 +348,70 @@ func applyProperties(p *processRunner, pid int) error {
...
+		return fmt.Errorf("failed to set IO priority of process %s: class %d is not valid", p, ioPriority.Class)
+	}
+
+	if ioPriority.Priority > 7 {
+		return fmt.Errorf("failed to set IO priority of process %s: priority %d is not valid", p, ioPriority.Priority)
+	}
+
+	classPos := 13 // IOPRIO_CLASS_SHIFT
+	priorityValue := ioPriority.Class<<classPos | ioPriority.Priority
+	sysctlWho := uintptr(1) // IOPRIO_WHO_PROCESS, we don't operate on threads or groups
```
smira, line 387: should it be current thread here?

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -353,6 +348,70 @@ func applyProperties(p *processRunner, pid int) error {
...
+
+	ioPriority, ioPrioritySet := p.opts.IOPriority.Get()
+	if ioPrioritySet {
+		err := setIOPriority(p, pid, ioPriority)
+		if err != nil {
+			return err
+		}
+	}
+
+	schedulingPolicy, schedulingPolicySet := p.opts.SchedulingPolicy.Get()
```
smira, line 365: we can merge here and above for readability

```suggestion
	if schedulingPolicy, schedulingPolicySet := p.opts.SchedulingPolicy.Get(); schedulingPolicySet {
```

smira (approved):

internal/app/machined/pkg/system/runner/process/process.go
dsseng: I believe no, since we pass the pid (of already spawned child) here, and want to set priority to all its threads

dsseng (commented):
Unsure how we test this. Perhaps running a unit test on the host won't work (or maybe it'll be okay if we lower priorities or set idle scheduling class), and an integration test would require exposing unneeded APIs

smira:
> Unsure how we test this. Perhaps running a unit test on the host won't work (or maybe it'll be okay if we lower priorities or set idle scheduling class), and an integration test would require exposing unneeded APIs

if we can do a test, that'd be cool.

we could use this for the `dashboard` service - lower the priority to idle on everything.

dsseng:
Hm, but then we'll need to get those parameters from Talos to test. In case of xfs_scrub I just waited for it to run and inspected the process with `chrt`, `ionice` and `htop` from a debug container

smira:
> Hm, but then we'll need to get those parameters from Talos to test. In case of xfs_scrub I just waited for it to run and inspected the process with `chrt`, `ionice` and `htop` from a debug container

dashboard runs, so we can do an integration test?

frezbo:
> > Hm, but then we'll need to get those parameters from Talos to test. In case of xfs_scrub I just waited for it to run and inspected the process with `chrt`, `ionice` and `htop` from a debug container
> 
> dashboard runs, so we can do an integration test?

yeh we could run those tools from a privileged pod in test, we already have helpers for these

smira:
> > > Hm, but then we'll need to get those parameters from Talos to test. In case of xfs_scrub I just waited for it to run and inspected the process with `chrt`, `ionice` and `htop` from a debug container
> > 
> > 
> > dashboard runs, so we can do an integration test?
> 
> yeh we could run those tools from a privileged pod in test, we already have helpers for these

I guess we could simply read from `/proc` ?

dsseng:
I guess nice and maybe scheduling policy could be received using procfs, let's see. Didn't find anything about IO priorities so they're likely fetched using a syscall. But if we can run commands in a privileged pods that could work as well.

dsseng:
I don't think giving dashboard some weird scheduling policy would be great. Idle can make it not informative when system has something to do (it won't be scheduled) while RT-like would increase resource usage by dashboard and raise its priority too much

smira:
> I don't think giving dashboard some weird scheduling policy would be great. Idle can make it not informative when system has something to do (it won't be scheduled) while RT-like would increase resource usage by dashboard and raise its priority too much

Probably not totally idle, but de-prioritizing in favor of anything else I guess might make sense

dsseng:
Well, out of 6 scheduling policies default perhaps shouldn't be changed. Offsetting priority (nice) would be good

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Bump. I want to make sure it's okay to resolve this conversation

internal/app/machined/pkg/system/runner/process/process_test.go
```diff
@@ -219,6 +222,148 @@ func (suite *ProcessSuite) TestStopSigKill() {
 	<-done
 }
 
+func (suite *ProcessSuite) TestPriority() {
+	pidFile := filepath.Join(suite.tmpDir, "talos-test-pid-prio")
```
smira, line 226: do we need to `t.Skip()` this is `os.Getuid() != 0`? (to make `go test ./...` pass without buildkit?)

smira (approved):
should we de-prioritize dashboard?

internal/app/machined/pkg/system/runner/process/process_test.go
dsseng: `go test -v ./internal/app/machined/pkg/system/runner/process` works for me

smira:
> > should we de-prioritize dashboard?
> 
> By how many? Not really sure, as it must still be responsive even under load, so just increase niceness by some points

I'd say it's the last thing that should be responsive (when compared to any other system component).

dsseng:
10 should do fine then. 0 - normal, -19 - max, 20 - least prioritized. Should I do it here as well?

PR #9872, opened 2024-12-03
chore: add integration tests for image-cache

### LINKED ISSUE #9860: integration test for image cache
## Proposal

Build an image cache out of `talosctl images default`. 

Add wrong registry mirrors (point them to some non-existent endpoint, e.g. 127.0.0.1:2005), and boot Talos from ISO with image cache, configuring also the partition on disk.

Config patch could be:

```
machine:
  features:
    imageCache:
      localEnabled: true
---
apiVersion: v1alpha1
kind: VolumeConfig
name: IMAGECACHE
provisioning:
    diskSelector:
        match: 'system_disk'
```

DmitriyMV:
Provide separate `integration/image-cache` tag.

Closes #9860

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -72,6 +72,7 @@ spec:
     - integration-images
     - integration-reproducibility-test
     - integration-cloud-images
+    - integration-image-cache
     - integration-image-factory
     - integration-aws
     - integration-aws-nvidia-oss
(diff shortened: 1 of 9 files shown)


DmitriyMV:
Side note: now I understand how to run (and what to edit) CI pipelines locally. The whole `.kres` (generates) -> `ci.yaml` (uses) -> `Makefile` (calls) -> `/hack/test/*.sh` is actually cool but not easily intuitive.

hack/test/e2e-image-cache.sh
```diff
@@ -0,0 +1,59 @@
...
+
+PROVISIONER=qemu
+CLUSTER_NAME=e2e-image-cache
+
+NODE="172.20.2.2"
+
+function build_special_mirrors {
+  REGISTRY_MIRROR_FLAGS=()
+
+  for registry in docker.io quay.io ghcr.io; do
```
frezbo, line 15: the cache should also work for the installer?

DmitriyMV: I actually don't know how to do in portable way:

* The known part is updating `cache-create` stage: new command should look something like `@( $(TALOSCTL_EXECUTABLE) images default | grep -v 'siderolabs'; echo "$(REGISTRY_AND_USERNAME)/installer:$(IMAGE_TAG)" ) | $(TALOSCTL_EXECUTABLE) images cache-create --image-cache-path=/tmp/cache.tar --images=- --force`
* The unknown part is how to create `image-cache.yaml` where one of the registries names is dependent on `${IMAGE_REGISTRY}` variable.

frezbo:
```
images default | grep -v installer > image.txt
echo ${IMAGE_REGISTRY}/siderolabs/installer:${TAG} >> images.txt
cat images.txt | <cache create>
```

DmitriyMV: Thats already done. What I'm saying is I have the patch file

```yaml
machine:
  features:
    imageCache:
      localEnabled: true
  registries:
    mirrors:
      gcr.io:
        skipFallback: true
        endpoints:
        - http://172.20.0.251:65000
      k8s.gcr.io:
        skipFallback: true
        endpoints:
        - http://172.20.0.251:65000
      registry.k8s.io:
        skipFallback: true
        endpoints:
        - http://172.20.0.251:65000
      <registry-name-should-be-inserted-during-ci-step>:
        skipFallback: true
        endpoints:
        - http://172.20.0.251:65000
---
apiVersion: v1alpha1
kind: VolumeConfig
name: IMAGECACHE
provisioning:
  diskSelector:
    match: 'system_disk'
```

frezbo: ahh can't we use `*` host and add `skipFallback` for it, so it catches all, or just `yq` it in like other steps

DmitriyMV: @frezbo yeah, it worked. Now we only download stuff from the cache.

frezbo (approved):

.kres.yaml
```diff
@@ -1506,6 +1507,54 @@ spec:
...
+          command: cache-create
+          environment:
+            PLATFORM: linux/amd64,linux/arm64
+            IMAGE_REGISTRY: registry.dev.siderolabs.io
+            PUSH: true
+        - name: use-image-cache
+          command: e2e-image-cache
+          withSudo: true
+          environment:
+            IMAGE_REGISTRY: registry.dev.siderolabs.io
```
smira, line 1546: probably should run a short integration test

Makefile
```diff
@@ -459,6 +459,12 @@ uki-certs: talosctl ## Generate test certificates for SecureBoot/PCR Signing
 	@$(TALOSCTL_EXECUTABLE) gen secureboot pcr
 	@$(TALOSCTL_EXECUTABLE) gen secureboot database
 
+.PHONY: cache-create
+cache-create: installer imager ## Generate image cache.
+	@( $(TALOSCTL_EXECUTABLE) images default | grep -v 'siderolabs/installer'; echo "$(REGISTRY_AND_USERNAME)/installer:$(IMAGE_TAG)" ) | $(TALOSCTL_EXECUTABLE) images cache-create --image-cache-path=/tmp/cache.tar --images=- --force
+	@crane push /tmp/cache.tar $(REGISTRY_AND_USERNAME)/image-cache
```
smira, line 465: should use a specific tag here `$(IMAGE_TAG)`, otherwise concurrent tests might override things

.kres.yaml
smira: ah, it doesn't run any (?).

In fact, why create a flake instead of re-using existing `e2e-qemu` like every other test?

the only difference is booting from an ISO and using a config patch, and both should be supported (or easy to add if not yet)

DmitriyMV: @smira fixed

smira:
We still pass `--registry-mirror` to the `e2e-image-cache-step`:

https://github.com/siderolabs/talos/actions/runs/12165563053/job/33930586574?pr=9872#step:13:96

smira:
If you look into the support bundle in artifacts, you will see:

```yaml
metadata:
    namespace: cri
    type: RegistryConfigs.cri.talos.dev
    id: registries
    version: 2
    owner: cri.RegistriesConfigController
    phase: running
    created: 2024-12-04T18:17:02Z
    updated: 2024-12-04T18:17:05Z
spec:
    mirrors:
        '*':
            endpoints:
                - http://127.0.0.1:3172
                - http://172.20.0.251:65000
            skipFallback: true
        docker.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://192.168.118.176:5000
        gcr.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://192.168.118.173:5000
        ghcr.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://192.168.118.165:5000
        k8s.gcr.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://172.20.0.251:65000
            skipFallback: true
        quay.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://192.168.118.179:5000
        registry.k8s.io:
            endpoints:
                - http://127.0.0.1:3172
                - http://192.168.118.148:5000
                - http://172.20.0.251:65000
            skipFallback: true
```

smira (approved):

PR #9874, opened 2024-12-04
feat: pcirebind controller

frezbo:
Add a controller to support rebinding drivers for PCI devices.

Fixes: https://github.com/siderolabs/extensions/pull/488


internal/app/machined/pkg/controllers/runtime/pcirebind.go
```diff
@@ -0,0 +1,224 @@
...
+				if err := c.handlePCIBindToTarget(cfg.TypedSpec().VendorDeviceID, cfg.TypedSpec().TargetDriver); err != nil {
+					return err
+				}
+
+				boundDriver, err := checkDeviceBoundDriver(cfg.TypedSpec().VendorDeviceID)
+				if err != nil {
+					return fmt.Errorf("error checking bound driver for device with id: %s, %w", cfg.TypedSpec().VendorDeviceID, err)
+				}
+
+				if boundDriver != cfg.TypedSpec().TargetDriver { // TODO: this will fail if the module is not loaded
```
frezbo, line 124: should we add a module status and see if it exists?

api/resource/definitions/runtime/runtime.proto
```diff
@@ -104,6 +104,21 @@ message MountStatusSpec {
...
+  string name = 1;
+  string vendor_device_id = 2;
+  string target_driver = 3;
+}
+
+// PCIRebindStatusSpec describes status of rebinded drivers.
+message PCIRebindStatusSpec {
+  string name = 1;
+  string vendor_device_id = 2;
+  string host_driver = 3;
```
dsseng, line 118: maybe default or original driver? as technically vfio would be a host driver as well

api/resource/definitions/runtime/runtime.proto
```diff
@@ -104,6 +104,21 @@ message MountStatusSpec {
   repeated string encryption_providers = 6;
 }
 
+// PCIRebindConfigSpec describes PCI rebind configuration.
+message PCIRebindConfigSpec {
+  string name = 1;
+  string vendor_device_id = 2;
```
dsseng, line 110: Is it just the model ID (like VID:PID)? Maybe there should be an option to select based on controller/port as well?

internal/app/machined/pkg/controllers/runtime/pcirebind.go
```diff
@@ -0,0 +1,224 @@
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		}
+
+		// Skip PCI rebind handling if running in a container or agent mode.
+		if c.V1Alpha1Mode.InContainer() || c.V1Alpha1Mode.IsAgent() {
+			return nil
+		}
```
dsseng, line 91: Move this up? We don't need to wait for udev to tell our platform

internal/app/machined/pkg/controllers/runtime/pcirebind.go
dsseng: Maybe the pcirebind controler could create kernel module configs (automatically request the module to be loaded when user wants to rebind to it), and consume the status from module controller?

internal/app/machined/pkg/controllers/runtime/pcirebind.go
```diff
@@ -0,0 +1,224 @@
...
+	if err := os.WriteFile(fmt.Sprintf(driverUnbindPath, vendorDeviceID), []byte(vendorDeviceID), 0o200); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("error unbinding device with id: %s, %w", vendorDeviceID, err)
+	}
+
+	return handleDriverProbe(vendorDeviceID)
+}
+
+// handlePCIBindToHost unbinds PCI device from a target driver and binds it to the host driver.
+func (c *PCIRebindController) handlePCIBindToHost(vendorDeviceID, hostDriver string) error {
+	if err := handleDriverOverride(vendorDeviceID, hostDriver); err != nil {
```
dsseng, line 183: I guess we typically want to send just "" there and it'll remove the override. Not 100% sure, but try, this could perhaps remove the need to keep a map of old drivers.

internal/app/machined/pkg/controllers/runtime/pcirebind.go
```diff
@@ -0,0 +1,224 @@
...
+		return err
+	}
+
+	if err := os.WriteFile(fmt.Sprintf(driverUnbindPath, vendorDeviceID), []byte(vendorDeviceID), 0o200); err != nil {
```
dsseng, line 187: Is the `vendorDeviceID` a VID:PID:SVID:SPID or like `00:1c.1` (a bus path). If the second, it's good and my comment regarding specifying the device could be invalidated

internal/app/machined/pkg/controllers/runtime/pcirebind_config_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestPCIRebindConfigSuite(t *testing.T) {
+	suite.Run(t, new(PCIRebindConfigSuite))
+}
+
+func (suite *PCIRebindConfigSuite) TestPCIRebindConfig() {
+	suite.Require().NoError(suite.Runtime().RegisterController(&runtimectrls.PCIRebindConfigController{}))
+
+	pciRebindConfig := &pci.RebindConfigV1Alpha1{
+		MetaName:          "ixgbe-bind",
+		PCIVendorDeviceID: "0000:04:00.00",
```
dsseng, line 35: Ah, well, get it now. Then I believe it should be renamed to path or sth like that in other places as well, since this is not specific to device model but its location within the system (controller/port/switches if any)

internal/integration/api/pcirebind.go
```diff
@@ -0,0 +1,173 @@
...
+	suite.Require().NoError(err)
+
+	suite.validateDriver(nodeCtx, pciDeviceID, "virtio-pci")
+
+	suite.RemoveMachineConfigDocuments(nodeCtx, cfgDocument.MetaKind)
+
+	_, err = suite.Client.COSI.WatchFor(nodeCtx, runtimeres.NewPCIRebindStatus(pciDeviceID).Metadata(), state.WithEventTypes(state.Destroyed))
+	suite.Require().NoError(err)
+
+	suite.validateDriver(nodeCtx, pciDeviceID, "virtio-pci")
```
dsseng, line 146: Maybe add transition to `vfio-pci` once more before this, so we're sure it also rebinds back to the original on config document removal? As this doesn't really test this since you first ask it to switch to the original driver by config

internal/pkg/tui/installer/installer.go
```diff
@@ -411,10 +411,6 @@ func (installer *Installer) apply(conn *Connection) error {
 		}
 	}
 
-	if err != nil {
-		return err
-	}
-
```
dsseng, line 417: Irrelevant/debugging?

website/content/v1.9/reference/configuration/pci/pcirebindconfig.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+
+
+{{< highlight yaml >}}
+apiVersion: v1alpha1
+kind: PCIRebindConfig
+name: ixgbe # Name of the config document.
+vendorDeviceID: 0000:04:00.00 # PCI device vendor and device ID.
```
dsseng, line 20: Well, not vendor/device ID but device path obtained from lspci or ID from `PCIDevices.hardware.talos.dev`

internal/integration/api/pcirebind.go
frezbo: yup, good point

internal/pkg/tui/installer/installer.go
frezbo: yeh linter caught unused condition

internal/app/machined/pkg/controllers/runtime/pcirebind.go
dsseng: yes, i guess it should be named path or so

internal/app/machined/pkg/controllers/runtime/pcirebind.go
dsseng: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-pci

> When specified, only a driver with a name matching the value written
		to driver_override will have an opportunity to bind to the
		device.  The override is specified by writing a string to the
		driver_override file (echo pci-stub > driver_override) and
		may be cleared with an empty string (echo > driver_override).
		This returns the device to standard matching rules binding.

...

> This also allows devices to
		opt-out of driver binding using a driver_override name such as
		"none".
		
So I guess clearing the file should restore binding by vendor ID

website/content/v1.9/reference/configuration/pci/pcirebindconfig.md
frezbo: I guess calling it pci id or bdf is the right term then

cmd/talosctl/cmd/docs.go
```diff
@@ -135,6 +136,10 @@ var docsCmd = &cobra.Command{
 					name:    "block",
 					fileDoc: block.GetFileDoc(),
 				},
+				{
+					name:    "pci",
```
smira, line 140: let's name it in some more generic way, e.g. `hardware`, otherwise we'll end up with tons of config folder

internal/app/machined/pkg/controllers/runtime/pcirebind_config_test.go
smira: I think this is usually called just PCI ID? 

it has nothing to do with vendor, but basically a path to the device in the current PCI topology

pkg/machinery/config/types/pci/rebind.go
```diff
@@ -0,0 +1,140 @@
...
+
+		count++
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	if count == 0 {
+		return []string{"IOMMU is not enabled"}, fmt.Errorf("IOMMU is not enabled")
```
smira, line 136: we should either return a warning or an error. but this would prevent Talos from booting completely which I think is not the best UX if IOMMU is disabled and the config is present

pkg/machinery/resources/runtime/pcirebind_config.go
```diff
@@ -0,0 +1,71 @@
...
+	return meta.ResourceDefinitionSpec{
+		Type:             PCIRebindConfigType,
+		DefaultNamespace: NamespaceName,
+		PrintColumns: []meta.PrintColumn{
+			{
+				Name:     "Name",
+				JSONPath: `{.name}`,
+			},
+			{
+				Name:     "VendorDeviceID",
```
smira, line 53: `PCI ID`

internal/app/machined/pkg/controllers/runtime/pcirebind.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Name implements controller.Controller interface.
+func (c *PCIRebindController) Name() string {
+	return "runtime.PCIRebindController"
+}
+
+// Inputs implements controller.Controller interface.
+func (c *PCIRebindController) Inputs() []controller.Input {
+	return nil
```
dsseng, line 44: Doesn't this need to contain `PCIRebindConfig`?

website/content/v1.9/reference/api.md
```diff
@@ -127,6 +127,8 @@ description: Talos gRPC API reference.
 - [resource/definitions/hardware/hardware.proto](#resource/definitions/hardware/hardware.proto)
     - [MemoryModuleSpec](#talos.resource.definitions.hardware.MemoryModuleSpec)
     - [PCIDeviceSpec](#talos.resource.definitions.hardware.PCIDeviceSpec)
+    - [PCIDriverRebindConfigSpec](#talos.resource.definitions.hardware.PCIDriverRebindConfigSpec)
```
smira, line 130: let's please revert all changes in `website/` and re-build docs, they should land in v1.10

hack/release.toml
```diff
@@ -26,6 +26,12 @@ preface = """
 * CoreDNS: 1.12.0
 
 Talos is built with Go 1.23.4.
+"""
+    [notes.driver-rebind]
+        title = "Driver Rebind"
+        description = """\
+Talos 1.9 now supports a new machine config document named `PCIDriverRebindConfig` that allows rebinding the driver of a PCI device to a different target driver.
```
smira, line 33:
```suggestion
Talos 1.10 now supports a new machine config document named `PCIDriverRebindConfig` that allows rebinding the driver of a PCI device to a different target driver.
```

smira (approved):

internal/app/machined/pkg/controllers/hardware/pci_driver_rebind.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Outputs implements controller.Controller interface.
+func (c *PCIDriverRebindController) Outputs() []controller.Output {
+	return []controller.Output{
+		{
+			Type: hardware.PCIDriverRebindStatusType,
+			Kind: controller.OutputExclusive,
+		},
+	}
```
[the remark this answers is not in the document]
frezbo: yeh, I guess better the user handle it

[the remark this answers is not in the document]
frezbo: probably will add a separate documentation for the same

internal/app/machined/pkg/controllers/hardware/pci_driver_rebind.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+
+			boundDriver, err := checkDeviceBoundDriver(cfg.TypedSpec().PCIID)
+			if err != nil {
+				return fmt.Errorf("error checking bound driver for device with id: %s, %w", cfg.TypedSpec().PCIID, err)
+			}
+
+			if boundDriver != cfg.TypedSpec().TargetDriver {
+				logger.Info(
+					"cannot validate if device is bound to target driver, ensure target driver module is loaded",
```
frezbo, line 110: on second thoughts i think we should error, otherwise if the module is loaded later, we cannot guarantee that the driver is rebinded, also this would skip creating the status resource

frezbo: actually this is fine, when module is loaded later on, the driver gets bind

PR #9878, opened 2024-12-04
fix: lock provisioning order of user disk partitions

### LINKED ISSUE #9877: user disks multiple partitions on the same disk get mixed up

smira:
Fixes #9877

As a side-effect, fix alignment of user disks for newer QEMU versions.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -1162,6 +1162,8 @@ func parseCPUShare(cpus string) (int64, error) {
 }
 
 func getDisks() ([]*provision.Disk, error) {
+	const GPTAlignment = 2 * 1024 * 1024 // 2 MB
+
 	// should have at least a single primary disk
 	disks := []*provision.Disk{
 		{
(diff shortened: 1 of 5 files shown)


cmd/talosctl/cmd/mgmt/cluster/create.go
```diff
@@ -1211,8 +1213,9 @@ func getDisks() ([]*provision.Disk, error) {
 		}
 
 		disks = append(disks, &provision.Disk{
-			// add 1 MB to make extra room for GPT and alignment
-			Size:            diskSize + 2*1024*1024,
+			// add 2 MB per partition to make extra room for GPT and alignment
+			// align to 4MB due to QEMU limitations
+			Size:            (diskSize + 2*1024*1024*uint64(len(diskPartitions)+1) + QEMUAlignment - 1) / QEMUAlignment * QEMUAlignment,
```
dsseng, line 1218: Maybe extract this variable and create it step by step like this for clarity:
```go
const GPTExtraSpace = 2*1024*1024
size := diskSize + GPTExtraSpace*uint64(1+len(diskPartitions))
size = (size + QEMUAlignment - 1) / QEMUAlignment * QEMUAlignment
```

dsseng (approved):

cmd/talosctl/cmd/mgmt/cluster/create.go
smira: but I can do the change you proposed as well

dsseng: Okay, then that'd be a different name, yet still not a number

DmitriyMV (approved):

cmd/talosctl/cmd/mgmt/cluster/create.go
smira: I refactored this completely moving the QEMU alignment into QEMU provisioner applying to any disk types, while GPT gap is still here

PR #9918, opened 2024-12-11
fix: power on the machine on reboot request in qemu power api

utkuozdemir:
IPMI powers on a machine when a PowerCycle command is sent, making a prior power on call unnecessary. We probably want to do the same thing in our API-based power implementation.

### DIFF
--- a/pkg/provision/providers/qemu/controller.go
+++ b/pkg/provision/providers/qemu/controller.go
@@ -87,7 +87,11 @@ func (c *Controller) Reboot() error {
 	c.mu.Lock()
 
 	if c.state == PoweredOff {
+		c.state = PoweredOn
+
 		c.mu.Unlock()
+
+		c.commandsCh <- VMCommandStart
 
 		return nil
 	}


pkg/provision/providers/qemu/controller.go
```diff
@@ -87,8 +87,12 @@ func (c *Controller) Reboot() error {
 	c.mu.Lock()
 
 	if c.state == PoweredOff {
+		c.state = PoweredOn
```
frezbo, line 90: this looks weird when we set after `PoweredOff`, maybe add a comment

frezbo: also we don't check if the `VMCommandStart` worked?

utkuozdemir: See the PowerOn code above - I simply do what it does, if the machine is powered off. Because the behavior we want is for it to power on.

smira (approved):

Issue #7881, opened 2023-10-21
Unable to destroy talos cluster on Podman


Chan9390:
### Description

I had setup Talos using talosctl on Podman (v4.7.1) rootful machine. It installed fine but when destroying the cluster using `talosctl cluster destroy` I get the error `strconv.Atoi: parsing "": invalid syntax`

### Logs

```
strconv.Atoi: parsing "": invalid syntax
```

### Environment

- Talos version: v1.5.4
- Kubernetes version: -
- Platform: Mac M2

BobyMCbobs:
I have the same error

sanmai-NL:
More modern versions of Podman have increasingly better CLI compat with Docker Engine CLI. Apart from that, I prefer to issue container management commands manually rather than letting `talosctl cluster` do opaque things.

FalconerTC:
Also seeing this on talosctl v1.7.6, Mac M2, Podman 5.2.2

PR #9920, opened 2024-12-11
fix: Use mtu network option for podman

clkamp:
For podman the network options is not 'com.docker.network.driver.mtu but only mtu.
This patches allow to use podman for local clusters

Fixes: #7881


smira (approved):
thank you!

Issue #9892, opened 2024-12-07
Image cache from ISO not being loaded


rothgar:
## Bug Report

I've been trying to get the image cache to work with an ISO on a bare metal machine.

I created the iso with image cache via
```
docker run --rm -t -v $PWD/image-cache.oci:/tmp/ -v $PWD/_out:/secureboot:ro -v $PWD/_out:/out -v /dev:/dev --privileged ghcr.io/siderolabs/imager:v1.9.0-beta.0 iso --image-cache /tmp/
```
and when I mount the iso on my local machine I can see the image cache directory and blobs

I'm generating a default machine config and applying the following patch
```
machine:
  features:
    imageCache:
      localEnabled: true
---
apiVersion: v1alpha1
kind: VolumeConfig
name: IMAGECACHE
provisioning:
  diskSelector:
    match: "system_disk"
  minSize: 4GB
  maxSize: 4GB
```
After the machine is installed the imagecacheconfig says it was skipped but I'm not sure why

```
node: 192.168.4.26
metadata:
    namespace: cri
    type: ImageCacheConfigs.cri.talos.dev
    id: image-cache
    version: 4
    owner: cri.ImageCacheConfigController
    phase: running
    created: 2024-12-07T01:00:52Z
    updated: 2024-12-07T01:00:57Z
spec:
    status: ready
    copyStatus: skipped
    roots:
        - /system/imagecache/disk
```
When I look through the registryd logs I can see all of the images were not found

```
192.168.4.26: 2024-12-07T01:03:40.209Z ERROR failed to handle request {"error": "1 error occurred:\n\t* stat /system/imagecache/disk/blob/sha256:0bc8e07f93b4cc11ff57d8ea896ed9d3622c4e79bbd7ac3d03d684a269a6ac8a: no such file or directory\n\n"}
192.168.4.26: 2024-12-07T01:03:40.228Z INFO image request {"method": "GET", "url": "/v2/coredns/coredns/blobs/sha256:ceabd226a7a8e199afc46c5f32d5c23fe98001ff25e6a8c7ca5f4490241256c2?ns=registry.k8s.io", "remote_addr": "127.0.0.1:48630", "name": "coredns/coredns", "digest": "sha256:ceabd226a7a8e199afc46c5f32d5c23fe98001ff25e6a8c7ca5f4490241256c2", "is_blob": true, "registry": "registry.k8s.io"}
```

When I look on the filesystem the /system/imagecache/disk directory is empty
```
[...]

PR #9929, opened 2024-12-12
docs: update image-cache doc for iso

rothgar:
Clarifying information from #9892

### DIFF
--- a/website/content/v1.9/talos-guides/configuration/image-cache.md
+++ b/website/content/v1.9/talos-guides/configuration/image-cache.md
@@ -29,6 +29,12 @@ cat images.txt | talosctl images cache-create --image-cache-path ./image-cache.o
 
 The OCI image cache directory might be used directly (`./image-cache.oci`) or pushed itself to a container registry of your choice (e.g. with `crane push`).
 
+Example of pushing the OCI image cache directory to a container registry:
+
+```bash
+crane push ./image-cache.oci my.registry/image-cache:my-cache
+```
+
 ## Building Boot Assets
 
 The image cache is provided to Talos via the boot assets.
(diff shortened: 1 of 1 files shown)


website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -79,21 +82,22 @@ machine:
...
 provisioning:
   diskSelector:
-    match: 'system_disk'
+    match: "system_disk"
   minSize: 2GB
   maxSize: 2GB
 ```
 
-In this example, image cache volume is provisioned on the system disk with a fixed size of 2GB.
+In this example, image cache volume is provisioned on the system disk with a fixed size of 2GB (default is 1GB).
```
frezbo, line 98: default is actually 1.3Gb

website/content/v1.9/talos-guides/configuration/image-cache.md
```diff
@@ -36,14 +36,17 @@ There are two supported boot asset types for the Image Cache: ISO and disk image
 
 ### ISO
 
+> Note: ISO image cache only works with iso9660 filesystems.
```
frezbo, line 39: what does it mean by an `iso9660`, I guess virtual cdroms are not so much used as an option, also newer ISO support something called File System Transposition and are not really supposed to be `dd`'ed to a disk

frezbo: some readup:

* https://github.com/pbatard/rufus/wiki/FAQ#user-content-Why_doesnt_Rufus_recommend_DD_mode_over_ISO_mode_for_ISOHybrid_images_Surely_DD_is_better
* https://github.com/pbatard/rufus/issues/2534#issuecomment-2263949687

smira: I think we need to fix this one way or another, but I'm confused a bit about the method. We should document what is the supported method of moving an ISO to a USB stick, and the one we support here

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: the actual defaults are: https://github.com/siderolabs/talos/blob/470b75563add4ce5bbce312c1e3dc783e63af1fa/internal/app/machined/pkg/controllers/cri/image_cache_config.go#L102-L104

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: ok, we have a plan going to support two things:

* `dd` the ISO to the USB stick (already works without issues)
* copy the ISO contents to the VFAT usb stick (need some changes to support that)

rothgar: I was basing this off what I was seeing on my physical hardware in the various tests. It only seemed to work when the detected filesystem was iso9660 but I didn't exhaustively test what's available.

website/content/v1.9/talos-guides/configuration/image-cache.md
rothgar: How does it pick between min and max?

website/content/v1.9/talos-guides/configuration/image-cache.md
rothgar: > * `dd` the ISO to the USB stick (already works without issues)

When I `dd` the iso to usb the usb drive showed up as sda1 and I don't think the cache was used during provisioning. I can test again if you think this should work.

rothgar:
If we're going to make this work with USB drives before 1.9 comes out then we can close this PR and focus on one with alternate instructions for USB.

website/content/v1.9/talos-guides/configuration/image-cache.md
smira: it picks a value in between leaning towards max if space is available, but it will never allocated less than min

smira:
@rothgar I'm going to update this a bit with recent fixes and push back to your branch.

smira (approved):

frezbo (approved):

Unix4ever (approved):

PR #9941, opened 2024-12-12
fix: support image cache on VFAT USB stick

### LINKED ISSUE #9936: detect FAT filesystem/volume labels

smira:
Scenario: copy contents of the ISO to the USB VFAT stick.

Make sure VFAT filesystem has a label `TALOS_*`.

Fixes #9936


smira:
I got this fully tested e2e with the following flow:

* mount an ISO
* copy everything from an ISO to a "USB stick" FAT-formatted, with volume label `TALOS_*`
* boot Talos with USB stick mounted - image cache is correctly detected, and copied to the disk

internal/app/machined/pkg/controllers/block/internal/volumes/format.go
```diff
@@ -30,7 +30,7 @@ func Format(ctx context.Context, logger *zap.Logger, volumeContext ManagerContex
 		devPath = volumeContext.Status.Location
 	}
 
-	dev, err := blockdev.NewFromPath(devPath, blockdev.OpenForWrite())
+	dev, err := blockdev.NewFromPath(devPath)
```
smira, line 33: dropped write bit here, as it's not needed, and breaks when the media is read-only (we're not writing to it anyways in this function)

internal/app/machined/pkg/controllers/cri/image_cache_config.go
```diff
@@ -198,19 +199,26 @@ func (ctrl *ImageCacheConfigController) Run(ctx context.Context, r controller.Ru
 func (ctrl *ImageCacheConfigController) createVolumeConfigISO(ctx context.Context, r controller.ReaderWriter) error {
 	builder := cel.NewBuilder(celenv.VolumeLocator())
 
-	// volume.name == "iso9660" && volume.label.startsWith("TALOS_")
+	// volume.name in ["iso9660", "vfat"] && volume.label.startsWith("TALOS_")
```
smira, line 202: allow VFAT here to handle USB sticks

pkg/provision/providers/qemu/launch.go
```diff
@@ -448,7 +448,7 @@ func launchVM(config *LaunchConfig) error {
 			)
 		case config.USBPath != "":
 			args = append(args,
-				"-drive", fmt.Sprintf("if=none,id=stick,format=raw,file=%s", config.USBPath),
+				"-drive", fmt.Sprintf("if=none,id=stick,format=raw,read-only=on,file=%s", config.USBPath),
```
smira, line 451: make QEMU USB stick emulation read-only, so that we can mount same file to multiple VMs

frezbo (approved):

rothgar:
Will there be another beta cut before the final release so I can test this or should I build from main?

smira:
> Will there be another beta cut before the final release so I can test this or should I build from main?

There will be no other beta, but we have it fully tested. There's another bug fix in #9954

PR #9948, opened 2024-12-13
fix: image cache integration test

frezbo:
Fix image cache cli integration test.


smira:
@frezbo should we promote to `integration/image-cache` ?

frezbo:
> @frezbo should we promote to `integration/image-cache` ?

na, this just the cli integration fix

frezbo:
`integration/qemu` faisl randomly, so skipping, the failed `TestCacheCreate` now passes.

PR #9954, opened 2024-12-13
fix: generating and serving registries with port

frezbo:
Fix generating and serving registies having port in them. This is needed to copy and serve imagecache from a vfat filesystem.

### DIFF
--- a/internal/app/machined/pkg/system/services/registry/registry.go
+++ b/internal/app/machined/pkg/system/services/registry/registry.go
@@ -123,7 +123,9 @@ func (svc *Service) handler(w http.ResponseWriter, req *http.Request) error {
 	if p.isBlob {
 		s = &singleFileStore{root: svc.root, path: "blob"}
 	} else {
-		s = &singleFileStore{root: svc.root, path: filepath.Join("manifests", ref.Name(), "digest")}
+		refName := handleRegistryWithPort(ref, p)
+
+		s = &singleFileStore{root: svc.root, path: filepath.Join("manifests", refName, "digest")}
 	}
 
 	info, err := s.Info(req.Context(), ref.Digest())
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/registry/registry.go
```diff
@@ -123,7 +123,9 @@ func (svc *Service) handler(w http.ResponseWriter, req *http.Request) error {
 	if p.isBlob {
 		s = &singleFileStore{root: svc.root, path: "blob"}
 	} else {
-		s = &singleFileStore{root: svc.root, path: filepath.Join("manifests", ref.Name(), "digest")}
+		refName := handleRegistryWithPort(ref, p)
```
smira, line 126: I wonder if we could do a simple test (cc @DmitriyMV) here in the `registry_test.go` by providing a fake FS in `testdata` with some structure, but e.g. empty files, and verifying all our conversions for `:` work correctly ?

frezbo: added a test

internal/app/machined/pkg/system/services/registry/registry_test.go
```diff
@@ -145,3 +150,151 @@ func readAll(t *testing.T, resp *http.Response) string {
...
+		tag          string
+	}
+
+	blobDir := filepath.Join(cacheDir, "blob")
+	manifestsDir := filepath.Join(cacheDir, "manifests")
+
+	require.NoError(t, os.MkdirAll(blobDir, 0o755))
+	require.NoError(t, os.MkdirAll(manifestsDir, 0o755))
+
+	for _, data := range []fakeRegistryData{
```
frezbo, line 197: maybe we could actually use image cache code itself to generate the cache, though we don't have a `registry:port` registry in CI

frezbo: or we could use `registry-gcr-io.ci.svc:5000` one of these in CI

frezbo: oh actually I can just use `ghcr.io:443`

smira: yeah, that might work

smira (approved):

frezbo:
now we have good test coverage

PR #9956, opened 2024-12-13
Add lost steps in bootstrap. Add variant to upload image.

PRIHLOP:
# Pull Request

## What? (description)

Added lost steps in the bootstrap process and additional variant for uploading images.

## Why? (reasoning)
An additional tool appears, and a command is added to check the status after the bootstrap process.

### DIFF
--- a/website/content/v1.8/talos-guides/install/cloud-platforms/hetzner.md
+++ b/website/content/v1.8/talos-guides/install/cloud-platforms/hetzner.md
@@ -10,10 +10,11 @@ aliases:
 Hetzner Cloud does not support uploading custom images.
 You can email their support to get a Talos ISO uploaded by following [issues:3599](https://github.com/siderolabs/talos/issues/3599#issuecomment-841172018) or you can prepare image snapshot by yourself.
 
-There are two options to upload your own.
+There are three options to upload your own.
 
 1. Run an instance in rescue mode and replace the system OS with the Talos image
 2. Use [Hashicorp packer](https://www.packer.io/docs/builders/hetzner-cloud) to prepare an image
+3. Use special utility [hcloud-upload-image](https://github.com/apricote/hcloud-upload-image/)
 
 ### Rescue mode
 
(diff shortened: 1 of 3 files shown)


website/content/v1.8/talos-guides/install/cloud-platforms/hetzner.md
```diff
@@ -255,6 +279,13 @@ This can be found by issuing:
 hcloud server list | grep talos-control-plane
 ```
 
+Before the bootstrap process, you need to apply controlplane config to the first node:
```
smira, line 282: the configuration is poassed via `--user-data-from-file`, so looks like something else is going wrong here

smira: ah, you're using metal- image instead of hcloud one (you should use on from the Image Factory), and then these steps are not needed.

PRIHLOP: Ah, probably you are right about the bootstrap process.

However, the part about additional variants of image upload is still present.

PRIHLOP: Oh, I understand my mistake. I will make changes to my suggestion of changes to docs.

website/content/v1.8/talos-guides/install/cloud-platforms/hetzner.md
```diff
@@ -145,6 +146,29 @@ export IMAGE_ID=<image-id-in-packer-output>
...
+Install process described [here](https://github.com/apricote/hcloud-upload-image/?tab=readme-ov-file#getting-started)(you can download binary or build from source, it is also possible to use Docker).
+
+For process simplification you can use this bash script:
+```bash
+#!/usr/bin/env bash
+export TALOS_IMAGE_VERSION=v1.8.4 # You can change to the current version
+export TALOS_IMAGE_ARC=amd64 # You can change to arm architecture
+export HCLOUD_SERVER_ARC=x86 # HCloud server architecture can be x86 or arm
+export PACKAGE_INSTALL_COMMAND="apt install"
+wget https://github.com/siderolabs/talos/releases/download/$TALOS_IMAGE_VERSION/metal-$TALOS_IMAGE_ARC.raw.zst
```
smira, line 160: should use Image Factory hcloud image here

smira (approved):
thank you, I applied small fixups

PR #9968, opened 2024-12-16
feat: support link altnames/aliases

smira:
At the moment, we don't use/support aliases, but we might in the future.

Altnames are filled out by `systemd-udevd`.

This PR has two parts:

* show aliases & altnames in `LinkStatus`
* match links by aliases/altnames when we configure addresses/routes/links

This should make a transition to `systemd-udevd` less painful if the previous link name is in `altNames`.

Forked rtnetlink for https://github.com/jsimonetti/rtnetlink/pull/241


frezbo (approved):
no tests, or no existing plumbing for tests?

smira:
> no tests, or no existing plumbing for tests?

good point, I will see if I can. I don't think I can _create_ altnames, but I should be able to set aliases

smira:
This is how it looks on QEMU:

```yaml
# talosctl -n 172.20.0.5 get links enp0s2 -o yaml
node: 172.20.0.5
metadata:
    namespace: network
    type: LinkStatuses.net.talos.dev
    id: enp0s2
    version: 3
    owner: network.LinkStatusController
    phase: running
    created: 2024-12-16T16:12:26Z
    updated: 2024-12-16T16:12:26Z
spec:
    altNames:
        - enxca1d163e1f18
    index: 8
    type: ether
    linkIndex: 0
    flags: UP,BROADCAST,RUNNING,MULTICAST,LOWER_UP
    hardwareAddr: ca:1d:16:3e:1f:18
    permanentAddr: ca:1d:16:3e:1f:18
    broadcastAddr: ff:ff:ff:ff:ff:ff
    mtu: 1500
    queueDisc: pfifo_fast
    operationalState: up
    kind: ""
    slaveKind: ""
    busPath: "0000:00:02.0"
    driver: virtio_net
    driverVersion: 1.0.0
    productID: "0x1000"
    vendorID: "0x1af4"
    product: Virtio network device
    vendor: Red Hat, Inc.
    linkState: true
    speedMbit: 4294967295
    port: Other
    duplex: Unknown
```

dsseng (approved):

smira:
> > no tests, or no existing plumbing for tests?
> 
> good point, I will see if I can. I don't think I can _create_ altnames, but I should be able to set aliases

added a bunch of unit-tests

internal/app/machined/pkg/controllers/network/route_spec_test.go
```diff
@@ -552,6 +552,123 @@ func (suite *RouteSpecSuite) TestLinkLocalRoute() {
...
+	dummyInterface := suite.uniqueDummyInterface()
+	dummyAlias := suite.uniqueDummyInterface()
+
+	conn, err := rtnetlink.Dial(nil)
+	suite.Require().NoError(err)
+
+	defer conn.Close() //nolint:errcheck
+
+	suite.Require().NoError(
+		conn.Link.New(
```
dsseng, line 565: Won't these affect host? Is this a mock netlink or one actually managing the kernel? If latter is the case this test should be skipped when nl is not accessible (perhaps not root)

frezbo (approved):

internal/app/machined/pkg/controllers/network/route_spec_test.go
smira: It's already the case for all tests in this file, they all do the same thing

smira:
I did more testing, and did even more fixes.

PR #10010, opened 2024-12-20
fix: fix `Failed to initialize SELinux labeling handle` udev error

dsseng:
Create a dummy SELinux file_contexts file to mitigate the warning. We do not rely on libselinux for labeling, so empty file suffices.


frezbo (approved):

hack/udevd/90-selinux.rules
```diff
@@ -1,5 +1,3 @@
-SUBSYSTEM=="*",SECLABEL{selinux}="system_u:object_r:device_t:s0"
-
```
dsseng, line 2: I added this by mistake, when my xattrs implementation failed to report SELinux label on non-labeled device (yet default is always in there). #9788 fixed that

smira (approved):

PR #10015, opened 2024-12-23
release(v1.10.0-alpha.0): prepare release

smira:
This is the official v1.10.0-alpha.0 release.


DmitriyMV (approved):

internal/app/machined/pkg/system/runner/process/process_test.go
```diff
@@ -223,6 +223,10 @@ func (suite *ProcessSuite) TestStopSigKill() {
 }
 
 func (suite *ProcessSuite) TestPriority() {
+	if os.Geteuid() != 0 {
+		suite.T().Skip("skipping test, need root privileges")
+	}
```
dsseng, line 228: Does it break w/o this?

dsseng (approved):

internal/app/machined/pkg/system/runner/process/process_test.go
smira: it breaks outside of container, requires more privs. I do smoke test via `go test ./...`

dsseng: hm, okay, maybe on different systems it's different

frezbo (approved):

PR #10031, opened 2024-12-24
fix: use correct default search domain

smira:
Search domain should be domain name of the hostname, not the FQDN.

Bug introduced in #9844


DmitriyMV:
Can we add what was expected and what was actual as an example?

DmitriyMV (approved):

smira:
> Can we add what was expected and what was actual as an example? 

example.org vs. foo.example.org

frezbo (approved):

PR #10057, opened 2024-12-26
refactor: generate /etc/os-release file static way

### LINKED ISSUE #8898: Remove dynamic /etc/os-release creation
Code: https://github.com/siderolabs/talos/blob/1fb8453c2db1659dd6c1670e4174125b26e777c5/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go#L387-L401

Instead of writing it down dynamic way with a bind mount, `go generate` the file contents, and make it part of `rootfs` and read-only.

Remove the task.

smira:
The file which is exported back to source via `make generate` is using short tag (vX.Y.Z), while the one generated for the actual build comes with full version tag.

Fixes #8898


Unix4ever (approved):

Dockerfile
```diff
@@ -285,6 +287,45 @@ RUN --mount=type=cache,target=/.cache go mod verify
...
+
+FROM embed-generate AS embed-abbrev-generate
+ARG ABBREV_TAG
+RUN echo -n "undefined" > pkg/machinery/gendata/data/sha && \
+    echo -n ${ABBREV_TAG} > pkg/machinery/gendata/data/tag
+RUN mkdir -p _out && \
+    echo PKGS=${PKGS} >> _out/talos-metadata && \
+    echo TAG=${TAG} >> _out/talos-metadata && \
+    echo EXTRAS=${EXTRAS} >> _out/talos-metadata
+COPY --from=pkg-kernel /certs/signing_key.x509 _out/signing_key.x509
```
frezbo, line 320: we can drop this, we don't use this anymore

frezbo (approved):

dsseng (approved):

PR #10061, opened 2024-12-27
feat: bring in partity with sd-257

frezbo:
Bring in parity with systemd 257 by supporting more UKI sections.

The output of `sd-measure` and our measure code will be different until https://github.com/systemd/systemd/pull/35765 is fixed upstream.

Fixes: #10075

### DIFF
--- a/internal/pkg/secureboot/secureboot.go
+++ b/internal/pkg/secureboot/secureboot.go
@@ -14,21 +14,39 @@ const (
 	OSRel   Section = ".osrel"
 	CMDLine Section = ".cmdline"
 	Initrd  Section = ".initrd"
+	Ucode   Section = ".ucode"
 	Splash  Section = ".splash"
 	DTB     Section = ".dtb"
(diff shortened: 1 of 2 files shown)


internal/pkg/secureboot/secureboot.go
```diff
@@ -14,21 +14,39 @@ const (
...
+		Initrd,
+		Ucode,
+		Splash,
+		DTB,
+		Uname,
+		SBAT,
+		PCRPKey,
+		Profile,
+		DTBAuto,
+		HWIDS,
```
smira, line 48: does it affect upgrades with SB in any way?

smira (approved):

internal/pkg/secureboot/secureboot.go
frezbo: nope, these are fields we don't use, so never get parsed, and since when we add `.profile` since the signature is bound to generated UKI, old ones would just work

PR #10062, opened 2024-12-27
refactor: implement volume mount controller

### LINKED ISSUE #9602: implement mount controller

smira:
Fixes #9602

Aggregate incoming volume mount requests, reconcile them with volume status, perform actual mounting, and produce mount status.


smira:
This PR implements only the first part of the volume mount flow, but I would like to split up things a bit to make it easier to get other changes in.

The image cache config controller never did proper unmount of the volumes, so the same state here for now - we don't actually unmount anything, but simply using new controller to mount image cache volumes.

The follow-up PR is planned to support first mounting STATE & EPHEMERAL, and that will include proper refactoring of the volume subsystem to support full unmount flow.

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -0,0 +1,295 @@
...
+				mountTarget := volumeStatus.TypedSpec().MountSpec.TargetPath
+				mountFilesystem := volumeStatus.TypedSpec().Filesystem
+
+				mountCtx, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+
+				// mount hasn't been done yet
+				if !ok {
+					var opts []mount.NewPointOption
+
+					// [TODO]: need to support more mount options:
```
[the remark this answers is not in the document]
smira: no, the next one - so far only Image Cache is handled, it doesn't use project quotas

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -0,0 +1,295 @@
...
+
+	for {
+		select {
+		case <-r.EventCh():
+		case <-ctx.Done():
+			return nil
+		}
+
+		volumeStatuses, err := safe.ReaderListAll[*block.VolumeStatus](ctx, r)
+		if err != nil {
```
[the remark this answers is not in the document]
frezbo: ahh I'm blind

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -0,0 +1,295 @@
...
+		if state.IsNotFoundError(err) {
+			// no mount status, we are done
+			return true, nil
+		}
+
+		return false, fmt.Errorf("failed to teardown mount status %q: %w", mountRequest.Metadata().ID(), err)
+	}
+
+	if !okToDestroy {
+		logger.Warn("waiting for mount status to be torn down")
```
dsseng, line 282: Not a reason for a warning?
```suggestion
		logger.Info("waiting for mount status to be torn down")
```

internal/pkg/mount/v2/mount.go
```diff
@@ -298,6 +313,20 @@ func (p *Point) Move(newTarget string) error {
...
+// RemountReadOnly remounts the mount point as read-only.
+func (p *Point) RemountReadOnly() error {
+	p.flags |= unix.MS_RDONLY
+
+	return p.remount()
+}
+
+// RemountReadWrite remounts the mount point as read-write.
+func (p *Point) RemountReadWrite() error {
+	p.flags &^= unix.MS_RDONLY
```
dsseng, line 325: Is this equivalent to `&= ~unix.MS_RDONLY`?

dsseng (approved):

internal/pkg/mount/v2/mount.go
smira: I think that should be exactly it.

PR #10066, opened 2024-12-28
ceph-with-rook.md: Fixed command to wait for ceph-rook HEALTH_OK

timolson:
Fixed command to wait for ceph-rook HEALTH_OK


smira (approved):
thank you!

PR #10084, opened 2025-01-03
fix: mount selinuxfs only when SELinux is enabled

### LINKED ISSUE #10083: KubeVirt does not work with Talos Linux 1.9
### Logs

virt-handler reports an error during VM creation:
```
failed to configure vmi network: setup failed, err: Critical network error: could not retrieve pid 199735 selinux label: getxattr /proc/199735/attr/current: operation not supported
```

this is happening because of selinux enabled:

```
cat /sys/fs/selinux/enforce
0
```

code:
- https://github.com/kubevirt/kubevirt/blob/9120e62feedcb842be595a9d1069e305c07d01b9/pkg/virt-handler/selinux/context_executor.go#L112
- https://github.com/kubevirt/kubevirt/blob/9120e62feedcb842be595a9d1069e305c07d01b9/cmd/virt-chroot/selinux.go#L102

### Environment

- Talos version: v1.9.1
- Kubernetes version: v1.30.3
- Platform: Cozystack

I tried to disable selinux using machineconfig:

```yaml
machine:
    install:
        extraKernelArgs:
        - -selinux
```

and 

```yaml
machine:
    install:
        extraKernelArgs:
        - selinux=0
```
and running `talosctl upgrade` commad, but it had no effect

Talos Linux 1.8 have no `/sys/fs/selinux/enforce` so this check returns false:
- https://github.com/kubevirt/kubevirt/blob/9120e62feedcb842be595a9d1069e305c07d01b9/cmd/virt-chroot/selinux.go#L27

dsseng:
Having selinuxfs mounted might confuse some software, as conventional Linux systems do not have selinuxfs mounted when SELinux is disabled and no policy is loaded.

Fixes #10083

### DIFF
--- a/internal/pkg/mount/v2/pseudo.go
+++ b/internal/pkg/mount/v2/pseudo.go
@@ -9,6 +9,7 @@ import (
 
 	"golang.org/x/sys/unix"
 
+	"github.com/siderolabs/talos/internal/pkg/selinux"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
@@ -49,8 +50,8 @@ func PseudoSubMountPoints() Points {
 		)
 	}
 
-	if _, err := os.Stat("/sys/fs/selinux"); err == nil {
-		// mount selinuxfs if it exists
+	if selinux.IsEnabled() {
+		// mount selinuxfs if it is enabled, which implies SELinux is the major LSM
 		points = append(points,
 			NewPoint("selinuxfs", "/sys/fs/selinux", "selinuxfs", WithFlags(unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_RELATIME)),
 		)


internal/pkg/mount/v2/pseudo.go
```diff
@@ -48,7 +49,7 @@ func PseudoSubMountPoints() Points {
 		)
 	}
 
-	if _, err := os.Stat("/sys/fs/selinux"); err == nil {
+	if _, err := os.Stat("/sys/fs/selinux"); err == nil && selinux.IsEnabled() {
```
frezbo, line 52: I guess we can then drop the `/sys/fs/selinux` check then?

frezbo: I thought we did the `/sys/fs/selinux` only appeared when `selinux=1` was set

dsseng: Did we add that check for cases like when apparmor is enabled and selinux is then disabled? Then perhaps yes, we can drop that check

frezbo (approved):

internal/pkg/mount/v2/pseudo.go
frezbo: > Did we add that check for cases like when apparmor is enabled and selinux is then disabled?

I believe so

kvaps (approved):

PR #10085, opened 2025-01-03
feat: enable `configfs`

frezbo:
Enable `configfs`.

Ref: https://www.kernel.org/doc/Documentation/filesystems/configfs/configfs.txt

Part of: https://github.com/siderolabs/extensions/issues/562


frezbo:
CONFIGFS is already enabled in Kconfig.

As per docs, bith sysfs and configfs should be enabled together.

dsseng (approved):

PR #10089, opened 2025-01-06
chore: drop deprecated allowSchedulingOnMasters

frezbo:
Drop deprecated `.cluster.allowSchedulingOnMasters` machine config field.

### DIFF
--- a/pkg/machinery/config/generate/init.go
+++ b/pkg/machinery/config/generate/init.go
@@ -190,13 +190,8 @@ func (in *Input) init() ([]config.Document, error) {
 		ClusterInlineManifests: v1alpha1.ClusterInlineManifests{},
 	}
 
-	if in.Options.AllowSchedulingOnControlPlanes {
-		if in.Options.VersionContract.KubernetesAllowSchedulingOnControlPlanes() {
-			cluster.AllowSchedulingOnControlPlanes = pointer.To(in.Options.AllowSchedulingOnControlPlanes)
-		} else {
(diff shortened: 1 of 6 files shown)


utkuozdemir (approved):

pkg/machinery/config/generate/init.go
```diff
@@ -193,9 +193,6 @@ func (in *Input) init() ([]config.Document, error) {
 	if in.Options.AllowSchedulingOnControlPlanes {
 		if in.Options.VersionContract.KubernetesAllowSchedulingOnControlPlanes() {
 			cluster.AllowSchedulingOnControlPlanes = pointer.To(in.Options.AllowSchedulingOnControlPlanes)
```
utkuozdemir, line 195: you can merge these 2 `if`s if you want.

dsseng (approved):

smira:
@frezbo can we please revert this change? 

I think we should keep backwards compatibility for the machine configuration, this field was available, yes, it is deprecated, but removing it would break upgrades from 1.8 to 1.9 for no real reason (I mean technically this extra field doesn't hurt).

frezbo:
> @frezbo can we please revert this change?
> 
> I think we should keep backwards compatibility for the machine configuration, this field was available, yes, it is deprecated, but removing it would break upgrades from 1.8 to 1.9 for no real reason (I mean technically this extra field doesn't hurt).

yeh, I can revert, I was thinking along the lines of keeping it forever :sweat_smile:

smira:
Thank you!

Issue #9633, opened 2024-11-04
unify boot assets in the intaller/image to be UKIs


smira:
## Feature Request

Instead of having separate images for UKIs and raw boot assets, settle down on having UKIs always.

We can have some format to extract necessary bits from the UKI, e.g. `talos.uki:<section_name>`. Support in the `imager`/`installer` to extract a raw bit from the UKI for non-sd-boot systems.

Support generating unsigned UKIs.

In Image Factory, drop the distinction between `installer` and `installer-secureboot` by always enabling SecureBoot for the installer.

### Description

```[tasklist]
### Tasks
- [ ] https://github.com/siderolabs/talos/issues/9688
- [ ] https://github.com/siderolabs/talos/issues/10101
- [x] go-ukify multi profile support
- [ ] https://github.com/siderolabs/talos/issues/10075
- [ ] https://github.com/siderolabs/talos/issues/10323
- [ ] https://github.com/siderolabs/talos/issues/10332
- [ ] https://github.com/siderolabs/talos/issues/10339
- [ ] https://github.com/siderolabs/talos/issues/10349
```

PR #10095, opened 2025-01-07
feat: uki iso

frezbo:
Switch to using sd-boot for newer iso.
New profile `legacy-iso` added to support booting with grub.

Part of: #9633


smira:
Why are we doing this, or in other words, what is the goal?

I don't think our default ISO shouldn't work with BIOS.

PR #10102, opened 2025-01-08
refactor: drop usage of objcopy to generate UKIs

smira:
This brings native implementation without external dependencies.


internal/pkg/secureboot/uki/internal/pe/native.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	sectionAlignment := uint64(header.SectionAlignment - 1)
+	fileAlignment := uint64(header.FileAlignment - 1)
+
+	baseVirtualAddress := uint64(lastSection.VirtualAddress) + uint64(lastSection.VirtualSize)
+	baseVirtualAddress = (baseVirtualAddress + sectionAlignment) &^ sectionAlignment
+
+	newHeader := *header
+	newHeader.MajorLinkerVersion = 2
```
frezbo, line 108: should we just set with whatever objcopy does or set as `0` like ukify does?

internal/pkg/secureboot/uki/internal/pe/native.go
```diff
@@ -0,0 +1,275 @@
...
+	var dosheader [0x40]byte
+
+	_, err = in.ReadAt(dosheader[:], 0)
+	if err != nil {
+		return fmt.Errorf("failed to read DOS header: %w", err)
+	}
+
+	binary.LittleEndian.PutUint32(dosheader[0x3c:], 0x80)
+
+	_, err = out.Write(append(append(dosheader[:], bytes.Repeat([]byte{0}, 0x40)...), []byte("PE\x00\x00")...))
```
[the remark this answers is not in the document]
smira: I made it to match more closely what `objcopy` does - it adds another `0x40` bytes, so this reduces the diff with `objcopy`, otherwise the whole output is shifted 0x40 bytes and hard to compare

frezbo (approved):

internal/pkg/secureboot/uki/internal/pe/native.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// 1. DOS header
+	var dosheader [0x40]byte
+
+	_, err = in.ReadAt(dosheader[:], 0)
+	if err != nil {
+		return fmt.Errorf("failed to read DOS header: %w", err)
+	}
+
+	binary.LittleEndian.PutUint32(dosheader[0x3c:], 0x80)
```
dsseng, line 72: Maybe label constants used here?

dsseng (approved):

internal/pkg/secureboot/uki/internal/pe/native.go
smira: yes, good point, I can do some constants

internal/pkg/secureboot/uki/internal/pe/native.go
smira: fixed this up, ignoring mismatch in the tests

internal/pkg/secureboot/uki/internal/pe/native.go
smira: added constants, thanks

PR #10112, opened 2025-01-09
feat: support generating unsigned UKIs

frezbo:
Support generating unsigned UKI's.

Also plumb in support to `talosctl cluster create` to boot off UKI's. This doesn't work yet as installer needs more work.

Part of #9633

### DIFF
--- a/pkg/imager/profile/default.go
+++ b/pkg/imager/profile/default.go
@@ -55,6 +55,22 @@ var Default = map[string]Profile{
 			},
 		},
 	},
+	"metal-uki": {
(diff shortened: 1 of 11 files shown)


pkg/imager/profile/default.go
```diff
@@ -55,6 +55,14 @@ var Default = map[string]Profile{
 			},
 		},
 	},
+	"metal-uki": {
```
smira, line 58: and (probably?) add a `secureboot-metal-uki` ?

smira (approved):

PR #10116, opened 2025-01-10
fix: extfs repair and resize

### LINKED ISSUE #10103: resize2fs needs e2fsck
### Description

AFAIK, resizing an online ext4 partition typically always needs a preliminary file system check (e2fsck).
The latter however is not implemented in `pkg/makefs/ext4`. AFAICS, `Ext4Resize` was only implemented in context of handling the new image cache in v1.9, which is how I found this out.

### Logs

```
talosctl get volumestatus -{n,e}172.31.32.201 IMAGECACHE -o yaml

(...)

    errorMessage: |+
        error growing filesystem: error growing ext4: exit status 1: resize2fs 1.47.1 (20-May-2024)
        Please run 'e2fsck -f /dev/nvme0n1p6' first.
```

### Environment

- Talos version: 1.9.1
- Kubernetes version: N/A
- Platform: AMD64 baremetal, nvme disk

### Relevant Config

I got this working for now by avoiding the need for a resize by setting the same exact sizes with following configurations:
```
apiVersion: v1alpha1
kind: VolumeConfig
name: IMAGECACHE
provisioning:
  diskSelector:
    match: 'system_disk'
  minSize: 8GiB
  maxSize: 8GiB
  grow: False
```
and for generating the installer artifacts:
```
docker run --rm -t \
	-v ${buildir}:/secureboot:ro \
	-v ${buildir}:/out \
	-v ${buildir}/image-cache.oci:/image-cache.oci:ro \
	-v /dev:/dev \
	--privileged \
	ghcr.io/siderolabs/imager:${TALOS_VERSION} \
	iso \
	--image-disk-size 8GiB \
	--image-cache /image-cache.oci \
	--base-installer-image ${installimage} \
	${sysext}
```
(Keep in mind, setting min/maxSize to `8GB` will also trigger a resize if `--image-disk-size ` is set to 8GiB (giga vs gigi)

### Proposed Solution
Always perform an e2fsck before resize2fs.

smira:
Fixes #10103

### DIFF
--- a/hack/test/patches/image-cache.yaml
+++ b/hack/test/patches/image-cache.yaml
@@ -23,3 +23,4 @@ name: IMAGECACHE
 provisioning:
   diskSelector:
     match: 'system_disk'
+  grow: true
--- a/internal/pkg/mount/v2/repair.go
+++ b/internal/pkg/mount/v2/repair.go
@@ -12,10 +12,21 @@ import (
 
 // repair a filesystem.
 func (p *Point) repair(printerOptions PrinterOptions) error {
-	printerOptions.Printf("filesystem on %s needs cleaning, running repair", p.source)
+	var repairFunc func(partition string) error
 
-	if err := makefs.XFSRepair(p.source, p.fstype); err != nil {
-		return fmt.Errorf("xfs_repair: %w", err)
+	switch p.fstype {
+	case "ext4":
+		repairFunc = makefs.Ext4Repair
+	case "xfs":
+		repairFunc = makefs.XFSRepair
+	default:
+		return fmt.Errorf("unsupported filesystem type for repair: %s", p.fstype)
+	}
+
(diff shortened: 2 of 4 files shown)


frezbo (approved):
interesting this test worked, or it's wrong https://github.com/siderolabs/talos/blob/main/pkg/makefs/ext4_test.go#L64

dsseng (approved):

smira:
> interesting this test worked, or it's wrong [`main`/pkg/makefs/ext4_test.go#L64](https://github.com/siderolabs/talos/blob/main/pkg/makefs/ext4_test.go?rgh-link-date=2025-01-10T17%3A12%3A18Z#L64)

I think it's fine, but the filesystem is/was clean, we have to make it specifically dirty (like unclean reboot I guess)

PR #10117, opened 2025-01-10
chore: uki code restructure

frezbo:
UKI code re-structure, no-op.


internal/pkg/encryption/keys/tpm2.go
```diff
@@ -16,8 +16,8 @@ import (
 	"github.com/siderolabs/go-blockdevice/v2/encryption/luks"
 	"github.com/siderolabs/go-blockdevice/v2/encryption/token"
 
-	"github.com/siderolabs/talos/internal/pkg/secureboot"
-	"github.com/siderolabs/talos/internal/pkg/secureboot/tpm2"
+	"github.com/siderolabs/talos/internal/pkg/uki"
+	"github.com/siderolabs/talos/internal/pkg/uki/secureboot/tpm2"
```
smira, line 20: idk, does it make sense to keep `pkg/secureboot` and extract `pkg/secureboot/uki` to `pkg/uki` ?

smira: as now we have `pkg/uki/uki`

frezbo: okay, that won't work since we import `uki` in `secureboot/internal/measure` and import `measure` back into `uki`, I can renamed the `pkg/uki/uki` to `pkg/uki/gen` since it's more aligned with what it does

smira: can we make `measure` a package on its own? I think it makes sense for the `uki` to import `secureboot`, but `secureboot` is technically independent of UKI?

smira (approved):

Issue #10110, opened 2025-01-09
Authorization Mode Order in kube-apiserver Configuration


Nauno33:
## Bug Report

Expected Behavior:
I expect the authorization-mode to be set in the exact order as specified: Node,Webhook,RBAC.

Actual Behavior:
The authorization-mode changes order after reboot.

### Description
I am trying to configure the authorization-mode in kube-apiserver to prioritize the Webhook authorization mode before RBAC. Here are the steps I followed:
I set the following extra args in my Machineconfiguration:

```
authorization-mode: Node,Webhook,RBAC  
authorization-webhook-version: v1  
authorization-webhook-config-file: /system/config/kubernetes/kube-apiserver/authorization-config.yaml  
```

I am using an external Webhook, which is a pod running Open Policy Agent (OPA) within the cluster.
After rebooting, I noticed that the authorization-mode order is altered as follows:

    authorization-mode: Node,RBAC,Webhook  

The order is not as I configured it.

Steps to Reproduce:

Configure extraArgs for kube-apiserver as follows:

```
    authorization-mode: Node,Webhook,RBAC  
    authorization-webhook-version: v1  
    authorization-webhook-config-file: /system/config/kubernetes/kube-apiserver/authorization-config.yaml  
```

Reboot the node.
Check the effective authorization-mode in the kube-apiserver configuration.

Questions:

Is there a way to ensure the authorization-mode order is respected as configured?

Thank you for your help!

### Environment

- Talos version: [1.9.1]
- Kubernetes version: [1.32.0]
- Platform: vmware

PR #10125, opened 2025-01-14
fix: kube-apiserver authorizers order

frezbo:
Fixes handling of `kube-apiserver` authorization config authorizers order.

Fixes: #10110

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane.go
@@ -131,11 +131,6 @@ func NewControlPlaneAuthorizationController() *ControlPlaneAuthorizationControll
 				var authorizers []k8s.AuthorizationAuthorizersSpec
 
 				for _, authorizer := range cfgProvider.Cluster().APIServer().AuthorizationConfig() {
-					// skip Node and RBAC authorizers as we add them by default later on.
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/controllers/k8s/control_plane.go
```diff
@@ -145,7 +140,29 @@ func NewControlPlaneAuthorizationController() *ControlPlaneAuthorizationControll
 					})
 				}
 
-				res.TypedSpec().Config = slices.Concat(v1alpha1.APIServerDefaultAuthorizationConfigAuthorizers, authorizers)
+				if !slices.ContainsFunc(authorizers, func(a k8s.AuthorizationAuthorizersSpec) bool {
```
smira, line 143: I'm confused why we add them to the tail of the list now, while previously they were at the head? Should we keep it at the head?

frezbo: if we keep at head, does that mean user provided always comes at the end, which is not what we want

frezbo: oh I understand now, keep at head if either `Node` or `RBAC` is not specified by the user, that makes sense

smira (approved):

PR #10128, opened 2025-01-14
chore: copy uki to installer

frezbo:
Add unsigned uki to installer.
Add a QEMU test with unsigned UKI.

Part of: #9633


frezbo:
> I don't quite understand the goal of what we're doing today to be honest.
> 
> I think we should put unconditionally UKI to the installer (either signed or unsigned, depending on imager configuration), and make `installer` automatically extract initramfs/kernel from it when running in GRUB mode.

but imager generates uki, so we can't put in default talos installer target, or we use imager always to generate installer

smira:
> but imager generates uki, so we can't put in default talos installer target, or we use imager always to generate installer

So my thought process is the following:

* in Dockerfile `install-artifacts`, always put UKIs instead of split initramfs/vmlinuz
* in `imager`, when consuming incoming profile, support extracting sections of PE files, with syntax e.g. like `/path/to/some.uki:.kernel`, so that one can reference kernel asset to be extracted from the UKI
* in `imager`, always generate installers with UKIs in artifacts attached
* in `installer`, automatically detect what we should be doing: if using sd-boot and we have UKI, just copy it; if using GRUB and we have UKI, extract section from UKI same way

All of that for Talos 1.10+

Now, in Image Factory, we can generate `installer` with signed UKIs always, and they will work both for SecureBoot and non-SecureBoot.

Does it makes sense? We can talk it through as well

frezbo:
Closing as this is going to be split up

PR #10158, opened 2025-01-17
docs: update SideroLinkConfig example

smira:
The provided example wasn't valid, it should not have a path component.

### DIFF
--- a/website/content/v1.10/reference/configuration/siderolink/siderolinkconfig.md
+++ b/website/content/v1.10/reference/configuration/siderolink/siderolinkconfig.md
@@ -16,14 +16,14 @@ title: SideroLinkConfig
 {{< highlight yaml >}}
 apiVersion: v1alpha1
 kind: SideroLinkConfig
-apiUrl: https://siderolink.api/join?token=secret # SideroLink API URL to connect to.
+apiUrl: https://siderolink.api/jointoken?token=secret # SideroLink API URL to connect to.
 {{< /highlight >}}
 
 
 | Field | Type | Description | Value(s) |
 |-------|------|-------------|----------|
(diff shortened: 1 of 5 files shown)


Unix4ever (approved):

utkuozdemir (approved):

frezbo (approved):

website/content/v1.10/reference/configuration/siderolink/siderolinkconfig.md
```diff
@@ -23,7 +23,7 @@ apiUrl: https://siderolink.api/join?token=secret # SideroLink API URL to connect
 | Field | Type | Description | Value(s) |
 |-------|------|-------------|----------|
 |`apiUrl` |URL |SideroLink API URL to connect to. <details><summary>Show example(s)</summary>{{< highlight yaml >}}
-apiUrl: https://siderolink.api/join?token=secret
+apiUrl: https://siderolink.api/?token=secret
```
rsmitty, line 26: shouldn't this be `https://siderolink.api/?jointoken=secret`?

smira: I guess you're right, thanks. In fact there might be more params there, but we can't get all of that here

rsmitty: nit: if we want to match exactly what would be given by omni, we'd do `https://siderolink.api?jointoken=secret` without the trailing slash. I know it doesn't make any functional difference, but...

smira: I guess nobody needs this with Omni, they should use what Omni says (it might have more arguments). Let's assume this is for manual only

Issue #10009, opened 2024-12-20
Support generating `iqn` and `nqn` during install


frezbo:
## Feature Request

Currently we generate an `iqn` once when the iscsi extension is installed, since `iqn` and `nqn` are pretty standard, move this to talos install process, so that they remain during the lifecycle of the node.

Ref: https://docs.netapp.com/us-en/e-series-santricity/sm-settings/nvme-overview.html

frezbo:
`nvme gen-hostnqn` can be used

frezbo:
As per discussion we'll derive these from the machine id deterministically on each boot

PR #10162, opened 2025-01-17
feat: generate iqn and nqn files.

frezbo:
Generate deterministic IQN and NQN files based on `machine-id`.

Fixes: #10009

### DIFF
--- a/internal/app/machined/pkg/controllers/files/iqn.go
+++ b/internal/app/machined/pkg/controllers/files/iqn.go
@@ -1,0 +1,99 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package files
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/siderolabs/gen/optional"
+	"go.uber.org/zap"
+
+	clusteradapter "github.com/siderolabs/talos/internal/app/machined/pkg/adapters/cluster"
+	runtimetalos "github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/controllers/files/nqn.go
```diff
@@ -0,0 +1,119 @@
...
+		hostID, err := uuid.FromBytes(machineID[:16])
+		if err != nil {
+			return fmt.Errorf("failed to convert machine-id to UUID: %w", err)
+		}
+
+		if err := safe.WriterModify(ctx, r, files.NewEtcFileSpec(files.NamespaceName, "nvme/hostid"),
+			func(r *files.EtcFileSpec) error {
+				spec := r.TypedSpec()
+
+				spec.Contents = []byte(hostID.String())
```
DmitriyMV, line 96: uuid is actually []byte underneath, so you can do a simple assign here I think. Or []byte(hostID)

DmitriyMV: Oh, yeah, my bad. We need string representation stored in byte slice.

internal/app/machined/pkg/controllers/files/iqn.go
```diff
@@ -0,0 +1,98 @@
...
+		_, err := safe.ReaderGetByID[*files.EtcFileStatus](ctx, r, "machine-id")
+		if err != nil {
+			if state.IsNotFoundError(err) {
+				continue
+			}
+
+			return fmt.Errorf("failed to get machine-id etcfile status: %w", err)
+		}
+
+		machineID, err := os.ReadFile("/etc/machine-id")
```
smira, line 79: I don't think it's correct to wait for a file and read it, we should instead use directly the Identity resource:

```go
		if err := safe.WriterModify(ctx, r, cluster.NewIdentity(cluster.NamespaceName, cluster.LocalIdentity), func(r *cluster.Identity) error {
			*r.TypedSpec() = localIdentity
```

https://github.com/siderolabs/talos/blob/399d53b543f6ca99f13d28313ae77b3472b0f728/internal/app/machined/pkg/controllers/cluster/node_identity.go#L103-L110

frezbo: fixed :+1:

smira (approved):

PR #10178, opened 2025-01-20
chore: de-hardcode list of extra images for image-cache test

DmitriyMV:
Get the image list using `registry.k8s.io/conformance` image instead of hardcoding it. Add new command `talosctl image integration` to create a proper list of k8s integration images for `talosctl images cache-create` command.


smira (approved):

cmd/talosctl/cmd/talos/image.go
```diff
@@ -153,6 +156,87 @@ var imageDefaultCmd = &cobra.Command{
 	},
 }
 
+// imageIntegrationCmd represents the integration image command.
```
frezbo, line 159: not sure why we need this in `talosctl` doesn't the list from conformance container image not enough?

DmitriyMV: Not really, because we need to do deduplication, also need images from the `default` command, but not all of them, and also `installer` we have to format with current TAG rather than stored one. All of this could be done using bash script, but I think that simple Go code is better than complex bash script.

frezbo (approved):

PR #10194, opened 2025-01-21
docs: update disk-encryption.md

greenpsi:
# Pull Request

## What? (description)
Add note about required DHCP server when using kms disk encryption

## Acceptance

### DIFF
--- a/website/content/v1.8/talos-guides/configuration/disk-encryption.md
+++ b/website/content/v1.8/talos-guides/configuration/disk-encryption.md
@@ -36,6 +36,8 @@ Talos Linux supports four encryption methods, which can be combined together for
 
 > Note: `nodeID` encryption is not designed to protect against attacks where physical access to the machine, including the drive, is available.
 > It uses the hardware characteristics of the machine in order to decrypt the data, so drives that have been removed, or recycled from a cloud environment or attached to a different virtual machine, will maintain their protection and encryption.
+>
+> Note: When using KMS encryption for `STATE` partition the network configuration can't be provided via the machine configuration, as KMS requires network connectivity before `STATE` partition is unlocked.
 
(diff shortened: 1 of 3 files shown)


DmitriyMV:
Greetings! Thanks for the PR. Can you also update 1.10 docs?

website/content/v1.8/talos-guides/configuration/disk-encryption.md
```diff
@@ -37,6 +37,9 @@ Talos Linux supports four encryption methods, which can be combined together for
 > Note: `nodeID` encryption is not designed to protect against attacks where physical access to the machine, including the drive, is available.
 > It uses the hardware characteristics of the machine in order to decrypt the data, so drives that have been removed, or recycled from a cloud environment or attached to a different virtual machine, will maintain their protection and encryption.
 
+> Note: When using kms encryption, a DHCP server is required for the talos nodes, as the static Network configuration is only loaded after the disk is decrypted.
```
frezbo, line 40: I'm not sure this is entirely correct based on how static network info is provided, if it's kernel cmdline I think it should work, am I right @smira ?

smira: yes, and it also depends on STATE/EPHEMERAL, and network config might come from too many place. but if you have your static network config in STATE, of course using KMS doesn't make sense

smira: so I think we should rephrase this fixing the casing as well:

```suggestion
> Note: When using KMS encryption for `STATE` partition the network configuration can't be provided via the machine configuration, as KMS requires network connectivity before `STATE` is decrypted.
```

smira (approved):

Issue #10097, opened 2025-01-07
1.9.1 installer creates invalid/unbootable GPT partition table


jfroy:
### Description

The 1.9.1 installer creates invalid/unbootable GPT partition tables on one of my metal amd64 nodes. I don't know when this regressed, but it used to work.

I've investigated a little bit and I believe that this node's UEFI firmware (namely the block storage DXE driver) is unable to process the partition table, and therefore is unable to find the EFI system partition to boot.

Here's the partition created by Talos on that node:

```
Disk /dev/nvme8n1: 931.51 GiB, 1000204886016 bytes, 244190646 sectors
Disk model: WD_BLACK SN850X HS 1000GB
Units: sectors of 1 * 4096 = 4096 bytes
Sector size (logical/physical): 4096 bytes / 4096 bytes
I/O size (minimum/optimal): 4096 bytes / 4096 bytes
Disklabel type: gpt
Disk identifier: 5BC2EC55-B0CA-4F84-AD62-27209C6C8175

Device          Start       End   Sectors   Size Type
/dev/nvme8n1p1    256    282111    281856   1.1G EFI System
/dev/nvme8n1p2 282112    282367       256     1M Linux filesystem
/dev/nvme8n1p3 282368    307967     25600   100M Linux filesystem
/dev/nvme8n1p4 307968 244190463 243882496 930.3G Linux filesystem
```

On a hunch, I moved/resized the EFI partition to start at LBA 1024:

```
Disk /dev/nvme8n1: 931.51 GiB, 1000204886016 bytes, 244190646 sectors
Disk model: WD_BLACK SN850X HS 1000GB
Units: sectors of 1 * 4096 = 4096 bytes
Sector size (logical/physical): 4096 bytes / 4096 bytes
I/O size (minimum/optimal): 4096 bytes / 4096 bytes
Disklabel type: gpt
Disk identifier: 5BC2EC55-B0CA-4F84-AD62-27209C6C8175

Device          Start       End   Sectors   Size Type
/dev/nvme8n1p1   1024    282111    281088   1.1G EFI System
/dev/nvme8n1p2 282112    282367       256     1M Linux filesystem
/dev/nvme8n1p3 282368    307967     25600   100M Linux filesystem
/dev/nvme8n1p4 307968 244190463 243882496 930.3G Linux filesystem
```

And recreated the boot entry:

```
[...]

PR #10197, opened 2025-01-22
fix: disks with 4k sector size and systemd-boot

smira:
The previous fix was completely wrong.

See #10097

The core part of the problem was that `dosfstools` was not installed in the `installer`, so this causes a busybox version of it to be used which is totally broken for 4k disks.

Adds an integration test.

go-blockdevice updates revert first partition alignment for 4k disks and allow `talosctl cluster create` to detect properly installation status.


frezbo (approved):

buroa:
Thanks @smira

DmitriyMV (approved):

jfroy:
Thanks indeed. We were so off the mark in the previous "fix". It wasn't a fix at all -- I just masked the problem by re-creating the EFI system partition in a rescue OS 😬

Issue #10206, opened 2025-01-22
Generated `initiatorname.iscsi` IQN is invalid


bzub:
### Description

I upgraded a node to a version of Talos after https://github.com/siderolabs/talos/commit/79987c05dcd39ca646c2d73c1e25488504f13a60 and the generated `/etc/iscsi/initiatorname.iscsi` appears to be invalid.

I see this in the file:
```
InitiatorName=iqn.2017.11.dev.talos:51e4c8772379c1d61fe5b73af861cd83
```

The `2017.11` date should, I believe, be `2017-11` instead.

Reference: https://datatracker.ietf.org/doc/html/rfc7143#section-4.2.7.4

### Logs

A slightly modified [trident CSI](https://github.com/NetApp/trident) reports:

```
API status: failed, Reason: Invalid IQN "iqn.2017.11.dev.talos:51e4c8772379c1d61fe5b73af861cd83": Invalid yyyy-mm date field. An example of a valid IQN is "iqn.1995-08.com.example:string"., Code: 13115
```

### Environment

```
Server:
          NODE:        <IP>
          Tag:         v1.10.0-alpha.0-72-g82c9ec158
          SHA:         82c9ec15
          Built:
          Go version:  go1.23.5
          OS/Arch:     linux/amd64
          Enabled:     RBAC
```

bzub:
Same issue with NQN I'm realizing now.

frezbo:
Ah yes, silly bug, will fixup from the PR, thanks for finding it.

PR #10207, opened 2025-01-22
fix: invalid date field in iqn/nqn

bzub:
This sets the date field to `yyyy-mm` in the generated IQN/NQN.

See: https://datatracker.ietf.org/doc/html/rfc7143#section-4.2.7.4
Signed-off-by: bzub <bzub@users.noreply.github.com>

Fixes: https://github.com/siderolabs/talos/issues/10206


frezbo (approved):
Thank you :heart:

PR #10210, opened 2025-01-23
feat: provide stable symlinks in disk resources

smira:
This allows to grab various `/dev/disk` symlinks,
including in maintenance mode when `talosctl ls` is not allowed.

Samle output:

```yaml
node: 172.20.0.5
metadata:
    namespace: runtime
    type: Disks.block.talos.dev
    id: nvme0n2
    version: 2
    owner: block.DisksController
    phase: running
    created: 2025-01-23T12:57:08Z
    updated: 2025-01-23T12:57:09Z
spec:
    dev_path: /dev/nvme0n2
    size: 5368709120
    pretty_size: 5.4 GB
    io_size: 512
    sector_size: 512
    readonly: false
    cdrom: false
    model: QEMU NVMe Ctrl
    serial: deadbeef
    wwid: nvme.1b36-6465616462656566-51454d55204e564d65204374726c-00000002
    bus_path: /pci0000:00/0000:00:08.0/nvme
    sub_system: /sys/class/block
    transport: nvme
    symlinks:
        - /dev/disk/by-diskseq/11
        - /dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_deadbeef_2
        - /dev/disk/by-id/nvme-nvme.1b36-6465616462656566-51454d55204e564d65204374726c-00000002
        - /dev/disk/by-path/pci-0000:00:08.0-nvme-2
```


internal/app/machined/pkg/controllers/block/symlinks.go
```diff
@@ -0,0 +1,199 @@
...
+		{
+			Type: block.SymlinkType,
+			Kind: controller.OutputExclusive,
+		},
+	}
+}
+
+const baseDevDiskPath = "/dev/disk"
+
+var byDiskPaths = []string{
```
frezbo, line 60: any reference for these, better to add as a comment

smira: good point, I'm not sure if all of them exist always, they seem to be coming from udev rules in the end, and they are loosely defined

smira: I think I will rework them a bit to support better handling

smira: reworked this completely, now we don't hardcode the list

frezbo (approved):
+1 on the better integration test

Issue #10191, opened 2025-01-21
build `installer` artifacts with UKIs always in the `imager`


smira:
> [!IMPORTANT]
> All of that for Talos 1.10+ only!

Rationale: drop `installer`/`installer-secureboot` difference in the Image Factory.

In GRUB bootloader code, support extracing vmlinuz/kernel from UKI using `debug/pe`.

In Talos change default installer image in the generate machine config step to use `factory.talos.dev` empty schematic installer instead of `ghcr.io/siderolabs/installer` image.

PR #10211, opened 2025-01-23
feat: extract kernel/initrd from uki for grub

frezbo:
Extract Kernel, Initrd and Commandline from UKI for GRUB installs.

Fixes: #10191

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
@@ -6,6 +6,7 @@ package grub
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -16,6 +17,7 @@ import (
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/mount"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/options"
 	"github.com/siderolabs/talos/internal/pkg/partition"
+	"github.com/siderolabs/talos/internal/pkg/uki"
 	"github.com/siderolabs/talos/pkg/imager/utils"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
@@ -68,18 +70,43 @@ func (c *Config) install(opts options.InstallOptions) (*options.InstallResult, e
 		return nil, err
 	}
 
-	if err := utils.CopyFiles(
-		opts.Printf,
-		utils.SourceDestination(
-			opts.BootAssets.KernelPath,
-			filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.KernelAsset),
-		),
-		utils.SourceDestination(
-			opts.BootAssets.InitramfsPath,
-			filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset),
-		),
-	); err != nil {
-		return nil, err
+	// if we have a kernel path, assume that the kernel and initramfs are available
+	if _, err := os.Stat(opts.BootAssets.KernelPath); err == nil {
+		if err := utils.CopyFiles(
+			opts.Printf,
+			utils.SourceDestination(
+				opts.BootAssets.KernelPath,
+				filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.KernelAsset),
+			),
+			utils.SourceDestination(
+				opts.BootAssets.InitramfsPath,
+				filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset),
+			),
+		); err != nil {
+			return nil, err
+		}
+	} else {
+		// if the kernel path does not exist, assume that the kernel and initramfs are in the UKI
+		assetInfo, err := uki.Extract(opts.BootAssets.UKIPath)
+		if err != nil {
+			return nil, err
+		}
+
+		defer assetInfo.Close() //nolint:errcheck
+
+		if err := utils.CopyReader(
+			opts.Printf,
+			utils.ReaderDestination(
+				assetInfo.Kernel,
+				filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.KernelAsset),
+			),
+			utils.ReaderDestination(
+				assetInfo.Initrd,
+				filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset),
+			),
+		); err != nil {
+			return nil, err
+		}
 	}
 
 	if err := c.Put(c.Default, opts.Cmdline, opts.Version); err != nil {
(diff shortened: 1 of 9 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
```diff
@@ -62,26 +66,63 @@ func (c *Config) Install(opts options.InstallOptions) (*options.InstallResult, e
...
+			utils.SourceDestination(
+				opts.BootAssets.InitramfsPath,
+				filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset),
+			),
+		); err != nil {
+			return nil, err
+		}
 	}
 
+	// TODO(frezbo): should we merge the cmdline from the UKI file with the one from the options?
```
frezbo, line 124: I think we can ignore the uki cmdline, since this will be an upgrade path in most cases unless booting bios

pkg/imager/imager.go
```diff
@@ -112,7 +112,7 @@ func (i *Imager) Execute(ctx context.Context, outputPath string, report *reporte
 			return "", fmt.Errorf("UKI output is not supported in this Talos version")
 		}
 	case profile.OutKindISO, profile.OutKindImage, profile.OutKindInstaller:
-		needBuildUKI = needBuildUKI && i.prof.SecureBootEnabled()
+		// needBuildUKI = needBuildUKI && i.prof.SecureBootEnabled()
```
frezbo, line 115: not sure what to do here? :thinking:

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
smira: yes, in GRUB case it's explicit I think

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
```diff
@@ -62,26 +66,63 @@ func (c *Config) Install(opts options.InstallOptions) (*options.InstallResult, e
...
-			opts.BootAssets.KernelPath,
-			filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.KernelAsset),
-		),
-		utils.SourceDestination(
-			opts.BootAssets.InitramfsPath,
-			filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset),
-		),
-	); err != nil {
-		return nil, err
+	if quirks.New(opts.Version).TalosVersionSDBootOnly() {
```
smira, line 75: I would probably instead check assets - if we see kernel & initramfs, use them, otherwise use UKI 

I mean check which file exists in the assets, it would be more compatible

pkg/machinery/imager/quirks/quirks.go
```diff
@@ -161,3 +161,15 @@ func (q Quirks) SupportsSELinux() bool {
 
 	return q.v.GTE(minVersionSELinux)
 }
+
+var minTalosVersionSDBootOnly = semver.MustParse("1.10.0")
+
+// TalosVersionSDBootOnly returns true if the Talos version supports only SDBoot for EFI.
+func (q Quirks) TalosVersionSDBootOnly() bool {
```
smira, line 168: I don't think we need `TalosVersion` in quirks, they are always dependent on Talos verion

Probably `UseSDBootForUEFI` ?

pkg/imager/out.go
```diff
@@ -393,17 +393,10 @@ func (i *Imager) outInstaller(ctx context.Context, path string, report *reporter
 
 	printf("generating artifacts layer")
 
+	ukiPath := strings.TrimLeft(fmt.Sprintf(constants.UKIAssetPath, i.prof.Arch), "/")
+
 	if i.prof.SecureBootEnabled() {
```
smira, line 398: I'm confused on this condition - we should generate UKI installer for anything Talos >=1.10, shouldn't depend on SecureBoot?

And we should never put both UKI and kernel/initramfs there

pkg/imager/imager.go
smira: I think we need to split it up:

* for ISO & Image things stay same as before
* for Installer we build UKI if SecureBoot || Talos >= 1.10

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
frezbo: ah okay, makes things easier

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
```diff
@@ -62,26 +66,63 @@ func (c *Config) Install(opts options.InstallOptions) (*options.InstallResult, e
...
+		}
+
+		kernel, err := os.Create(filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.KernelAsset))
+		if err != nil {
+			return nil, err
+		}
+
+		defer kernel.Close() //nolint:errcheck
+
+		initrd, err := os.Create(filepath.Join(opts.MountPrefix, constants.BootMountPoint, string(c.Default), constants.InitramfsAsset))
```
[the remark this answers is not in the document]
frezbo: +1 for this, make more sense

pkg/machinery/constants/constants.go
```diff
@@ -630,11 +630,17 @@ const (
 	RootfsAsset = "rootfs.sqsh"
 
 	// UKIAsset defines a well known name for our UKI filename.
-	UKIAsset = "vmlinuz.efi.signed"
+	UKIAsset = "vmlinuz.efi"
+
+	// UKIAssetSigned defines a well known name for our signed UKI filename.
+	UKIAssetSigned = "vmlinuz.efi.signed"
```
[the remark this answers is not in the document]
frezbo: would making it `vmlinuz.efi` break factory, I guess not?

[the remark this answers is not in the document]
smira: Good question, I think we never stored it in the assets.

But if we generate an installer for Talos < 1.10, we should name it `.signed`, otherwise old secureboot installer won't be able to find it.

internal/pkg/uki/internal/pe/extract.go
```diff
@@ -0,0 +1,47 @@
...
+		case ".initrd":
+			assetInfo.Initrd = section.Open()
+		case ".cmdline":
+			assetInfo.Cmdline = section.Open()
+		case ".linux":
+			assetInfo.Kernel = section.Open()
+		}
+	}
+
+	return assetInfo, nil
```
smira, line 46: I would suggest to return an error early here if e.g. `.linux` or `.initrd` are missing, as otherwise it would be a nil panic later. 

I'm not sure if we need a cmdline? (if not we could skip this)

smira (approved):

internal/pkg/uki/internal/pe/extract.go
frezbo: good point, `cmdline` is for later when we do uki kexec

Issue #10192, opened 2025-01-21
make non-secureboot ISO use sd-boot for UEFI mode


smira:
> [!IMPORTANT]
> Only for Talos 1.10+

Our current non-SB ISO is built using GRUB and uses GRUB in both BIOS and UEFI mode.

Goal: use GRUB only for BIOS, and use sd-boot for UEFI in the same ISO.

PR #10217, opened 2025-01-24
feat: generate iso's with both UKI and grub

frezbo:
Starting with Talos 1.10, the default generated ISO's will use GRUB for BIOS boot and sd-boot for EFI boot.

Fixes: #10192


smira:
We should probably add a note to `hack/release.toml`, this is a big change!

smira:
If this works, it's going to be massive! :tada:

smira (approved):

PR #10220, opened 2025-01-24
refactor: all network merge controllers

smira:
They had lots of common (and broken) code, so use generics to implement the core logic once, and use that in all kinds of merge controllers.

This removes lots of code duplication.

Fixes flaky unit-tests, and also improves controller perfomance on conflicts.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/address_merge_test.go
+++ b/internal/app/machined/pkg/controllers/network/address_merge_test.go
@@ -52,7 +52,7 @@ func (suite *AddressMergeSuite) SetupTest() {
 	suite.runtime, err = runtime.NewRuntime(suite.state, zaptest.NewLogger(suite.T()))
 	suite.Require().NoError(err)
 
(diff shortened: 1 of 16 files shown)


internal/app/machined/pkg/controllers/network/address_merge_test.go
```diff
@@ -152,7 +152,7 @@ func (suite *AddressMergeSuite) TestMerge() {
 	suite.assertAddresses(
 		[]string{
 			"lo/127.0.0.1/8",
-			"eth0/10.0.0.35/32",
+			"eth0/10.0.0.1/8",
```
frezbo, line 155: I guess this was a bug that got fixed?

frezbo (approved):

internal/app/machined/pkg/controllers/network/address_merge_test.go
smira: Yes, I did lots of runs with count 100 to find it )

PR #10224, opened 2025-01-25
fix(ci): k8s integration suite wait for resource.

frezbo:
The code was previously calling `GetUnstructuredResource` which wraps
the k8s error, construct the mapping and call the api directly.


internal/integration/base/k8s.go
```diff
@@ -641,10 +641,14 @@ func (k8sSuite *K8sSuite) WaitForResource(ctx context.Context, namespace, group,
 func (k8sSuite *K8sSuite) WaitForResourceToBeAvailable(ctx context.Context, duration time.Duration, namespace, group, kind, version, resourceName string) error {
 	return retry.Constant(duration).Retry(func() error {
 		_, err := k8sSuite.GetUnstructuredResource(ctx, namespace, group, kind, version, resourceName)
-		if errors.IsNotFound(err) {
+		if err != nil {
```
DmitriyMV, line 644: This will retry for any error. Maybe get the error type using "%T" and try to assert it? Or use a `string.Contains`?

frezbo: oh this was me being dump, I was calling `GetUnstructured` resource and it was wrapping the k8s error.

smira (approved):

Issue #10189, opened 2025-01-21
support `kexec` for UKIs


smira:
On non-SecureBoot systems, we can do `kexec` with UKIs by extracting kernel/initramfs/cmdline and passing raw parts to `kexec`, like we do for ZBOOT kernel on arm64.

frezbo:
how would `extraKernelArgs` work in non-secureboot mode?

smira:
Good question, I would say we either do `loader.conf`, or we skip it for now, and say that kernel args are only modifiable permanently with an upgrade?

I'm leaning towards upgrade path.

For temporary cmdline changes, I guess you can use sd-boot menu.

PR #10228, opened 2025-01-27
feat: support kexec from uki

frezbo:
Support kexec from UKI for non-secureboot by extracting kernel, initramfs and cmdline from UKI

Fixes: #10189

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
@@ -8,6 +8,7 @@ package bootloader
 import (
 	"os"
 
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/options"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot"
@@ -27,6 +28,9 @@ type Bootloader interface {
 	Revert(disk string) error
 	// RequiredPartitions returns the required partitions for the bootloader.
 	RequiredPartitions() []partition.Options
+
+	// KexecLoad does a kexec_file_load using the current entry of the bootloader.
+	KexecLoad(r runtime.Runtime, disk string) error
 }
 
 // Probe checks if any supported bootloaders are installed.
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
@@ -8,8 +8,14 @@ package grub
 import (
 	"errors"
 	"fmt"
+	"log"
+	"os"
 	"path/filepath"
+	"strings"
 
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/kexec"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/bootloader/options"
 	"github.com/siderolabs/talos/internal/pkg/partition"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"github.com/siderolabs/talos/pkg/machinery/version"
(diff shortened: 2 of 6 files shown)


frezbo:
Currently WIP, doesn't work

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1901,7 +1911,113 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
...
+				systemDisk.DevPath,
+				[]bootloadermount.Spec{
+					{
+						PartitionLabel: constants.EFIPartitionLabel,
+						FilesystemType: partition.FilesystemTypeVFAT,
+						MountTarget:    constants.EFIMountPoint,
+					},
+				},
+				func() error {
+					var kernelFd int
```
smira, line 1931: I think we need to refactor this a bit, this already becomes a bit of a function that is super hard to read.

we might want to extract these functions out, we can even make them part of the `bootloader` interface

smira (commented):
I guess we need to promote this to uki pipeline to prove that it works?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo: :+1: yup, makes sense

internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
```diff
@@ -27,6 +28,9 @@ type Bootloader interface {
 	Revert(disk string) error
 	// RequiredPartitions returns the required partitions for the bootloader.
 	RequiredPartitions() []partition.Options
+
+	// KexecLoad does a kexec using the bootloader.
```
dsseng, line 32:
```suggestion
	// KexecLoad does a kexec using the config from bootloader.
```

Maybe another rephrase, but the point is bootloader isn't used, but only its configs

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1877,6 +1872,12 @@ func KexecPrepare(_ runtime.Sequence, data any) (runtime.TaskExecutionFunc, stri
 			}
 		}
 
+		if efi.GetSecureBoot() {
+			log.Print("kexec skipped as secure boot is enabled")
```
dsseng, line 1876: kexec should have secure boot support iirc, yet this can be added later

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -44,6 +56,79 @@ func NewConfig() *Config {
...
 }
 
+// KexecLoad does a kexec using the bootloader.
+func (c *Config) KexecLoad(r runtime.Runtime, disk string) error {
+	return mount.PartitionOp(
+		disk,
+		[]mount.Spec{
+			{
+				PartitionLabel: constants.BootPartitionLabel,
+				FilesystemType: partition.FilesystemTypeXFS,
```
dsseng, line 66: nit: this should perhaps have a constant somewhere, as that's not necessarily XFS but rather we create it as XFS

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
```diff
@@ -44,6 +56,79 @@ func NewConfig() *Config {
...
+				if err != nil {
+					return err
+				}
+
+				defer func() {
+					if fileCloser != nil {
+						fileCloser.Close() //nolint:errcheck
+					}
+				}()
+			}
```
dsseng, line 102: This code is shared with sdboot and can be refactored into a new function

dsseng (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
frezbo: kexec works with secureboot, but that would break disk encryption aka trusted boot, also our UKI doesn't sign kernel if I'm right

frezbo: yeh, it's going to be a mess, since we also extend PCR from init, otherwise we need complex logic to see if we kexeced or not

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
frezbo: didn't quite get that

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
frezbo: but it makes more sense to keep it in here only, since it replaces the `fd` and passing it around might be confusing

smira:
> I guess we need to promote this to uki pipeline to prove that it works?

I don't see any labels so far...

frezbo:
> > I guess we need to promote this to uki pipeline to prove that it works?
> 
> I don't see any labels so far...

promoted the misc-0 and removed when i rebased, though the tests doesn't verify anything since if kexec fails it just continues with a normal reboot, I guess we could store that state in meta or something that it was kexeced

internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
smira: I would say we need to phrase that `KexecLoad` emulates boot as if it was done by the bootloader - we pick the entry to boot based on bootloader information, and then do the boot as if the machine was rebooted

smira (approved):
once it passes the tests

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/grub.go
smira: I think Dmitry's point is that GRUB has this multiple types around the code (as it does similar mount operations), so the whole block might be shared?

PR #10230, opened 2025-01-27
fix: pe uki extract

frezbo:
The PE UKI extract was returning a reader that reads upto the section `Size` which is the size on disk which could be padded, we need a limit reader reading upto `VirtualSize`.


smira (approved):

frezbo:
verified this works with Grub code extracting kernel and cmdline from UKI

dsseng (approved):

PR #10260, opened 2025-01-31
feat: bump pkgs for udev update

dsseng:
Update systemd to 257.2


smira (approved):

smira:
`nvidia-nonfree` often fails because it's out of VMs (fyi), it can be skipped

PR #10263, opened 2025-01-31
fix: do not close client.Client.conn with finalizer

DmitriyMV:
Test no longer pass.

This reverts commit 673ca4bcb2448b3c252fccff0d243932c97fd893.


smira (approved):

smira:
I think the root cause is something like:

```
client.DoSomething(...)
```

At this point nothing references `client` (missing `defer client.Close()`, but that was the whole point of adding a finalizer), so it calls the finalizer, and closes the connection (while the connection is still executing the request).

We can't say we're done with the client if we don't reference it anymore (e.g. we could have started long-running streaming API).

PR #10281, opened 2025-02-03
fix: image cache volume provisioning

smira:
There is a logic failure in the `ImageCacheConfig` controller that `allReady` was always `false` (now should be fixed).

The real issue was that `IMAGECACHE` being in the `SystemDisk` wave, and not being provisioned until `EPHEMERAL` is not provisioned, which blocks booting with Image Cache on and without an ISO, by just providing an empty `IMAGECACHE` disk volume. There isn't much use in this scenario, but still good to make it fixed.


DmitriyMV (approved):

Makefile
```diff
@@ -490,7 +490,7 @@ uki-certs: talosctl ## Generate test certificates for SecureBoot/PCR Signing
 .PHONY: cache-create
 cache-create: installer imager ## Generate image cache.
 	@docker run --entrypoint /usr/local/bin/e2e.test registry.k8s.io/conformance:$(KUBECTL_VERSION) --list-images | \
-		$(TALOSCTL_EXECUTABLE) images integration --installer-tag=$(IMAGE_TAG) --registry-and-user=$(REGISTRY_AND_USERNAME) | \
+		$(TALOSCTL_EXECUTABLE) images integration --installer-tag=$(IMAGE_TAG)-amd64-secureboot --registry-and-user=$(REGISTRY_AND_USERNAME) | \
```
frezbo, line 493: I guess unrelated to this, but to make the default secureboot installer?

frezbo (approved):

Makefile
smira: it is in fact related, the change you did to make ISO sd-boot broke image-cache, as it needs to cache a different installer image

smira: so I think we should figure out a plan vs. the `installer` and start using UKIfied installer everywhere now, but it doesn't have to be the this PR

PR #10283, opened 2025-02-03
fix: ensure proper closure of client.Client.conn with finalizer

DmitriyMV:
As @smira noted in #10263, as soon as compiler can assume that variable is unneeded, it's free to drop it, and runtime if free to GC it. We want to prevent that, because sometimes variable can be dropped while one of its fields methods is called or in-flight (the last method in scope for example).

To do this we need to patch two things:
1. Ensure that we call [runtime.KeepAlive](https://pkg.go.dev/runtime#KeepAlive) in all methods after the main logic path.
2. Replace embedding with unexported fields and add methods (and KeepAlive's) manually. Otherwise, "promoted" methods do not "keep-alive" our object and GC is free to collect it.

Even tho `ClientConnInterface` interface defines only two methods, we don't know how many type assertions happen inside, so we export everything from `*grpc.ClientConn` to our type.

### DIFF
--- a/pkg/machinery/client/grpc_connection_wrapper.go
+++ b/pkg/machinery/client/grpc_connection_wrapper.go
@@ -6,33 +6,108 @@ package client
 
 import (
 	"context"
+	"runtime"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/metadata"
 )
 
 type grpcConnectionWrapper struct {
-	*grpc.ClientConn
+	conn *grpc.ClientConn
 
 	clusterName string
 }
 
 func newGRPCConnectionWrapper(clusterName string, conn *grpc.ClientConn) *grpcConnectionWrapper {
-	return &grpcConnectionWrapper{
-		ClientConn:  conn,
+	res := &grpcConnectionWrapper{
+		conn:        conn,
 		clusterName: clusterName,
 	}
+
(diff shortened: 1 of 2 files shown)


pkg/machinery/client/grpc_connection_wrapper.go
```diff
@@ -6,33 +6,108 @@ package client
...
+	runtime.KeepAlive(c)
+
+	return err
+}
+
+// NewStream begins a streaming RPC.
+func (c *grpcConnectionWrapper) NewStream(ctx context.Context, desc *grpc.StreamDesc, method string, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+	res, err := c.conn.NewStream(c.appendMetadata(ctx), desc, method, opts...)
+
+	runtime.KeepAlive(c)
```
smira, line 46: this starts to feel more and more hacky solution to me, and more fragile.

Also I don't see how it fixes anything, as in this function the result (`grpc.ClientStream`) can run for as long as possible once this function returns, and still the problem is there.

Why can't we spend the effort other way around and find where the connection is not closed? It gives us immediate benefit? Easy ways there, like adding an assertion using `goleak` that no goroutines are left in each unit-test, and just simply inspecting the code. 

Even if we find a finalizer solution which works (which I doubt), it feels that fixing the core problem is more beneficial immediately.

DmitriyMV: Yeah, you right. I think we should try something different.

PR #10284, opened 2025-02-03
chore: track opened grpc connections

DmitriyMV:
This is an alternative to #10283 which doesn't close the connections and instead allows us to easily track the lost grpc connections using pprof.

Fix various grpc leaks while we are at it.


DmitriyMV:
[Explanation](https://medium.com/@cep21/creating-custom-go-profiles-with-pprof-b737dfc58e11)

smira (approved):

DmitriyMV:
@dsseng done as an Example - if those contain `Output:` comment block they execute as tests.

Issue #10133, opened 2025-01-14
Refactor talosctl command


rothgar:
## Feature Request

Update `talosctl` to have the following subcommands structured as **verbs** (apply, edit, gen, get, run), **nouns** (etcd, k8s, node), and **maintenance** (completion, config, help, version).

| command | usage |
|----|----|
| apply, a | used to apply a configuration |
| completion | show shell completion |
| config | manage talosonfig file |
| edit | change in-place configuration |
| etcd | managed etcd |
| gen | generate example configurations |
| get, g | get information about resources |
| help | print help |
| kubernetes, k8s, k | manage kubernetes clusters |
| node, n | node management commands |
| run, r | interactive commands |
| version | print version |

This replaces or moves a significant portion of subcommands to categorize them into one of these other groups.

Here is a high level table of how existing commands would be removed, moved, or modified

| Original Command | New Structure | Status |
|-----------------|---------------|---------|
| `apply-config, apply` | `apply` with `serviceaccount` subcommand | 🔴 Removed `apply-config` |
| `bootstrap` | Moved under `etcd bootstrap` | 🟡 Moved |
| `cluster create/destroy/show` | Moved under `kubernetes cluster` | 🟡 Moved |
| `completion bash/zsh` | `completion bash/zsh` | ⚪ Unchanged |
| `config` (all subcommands) | `config` with added `validate` | 🟡 Modified |
| `conformance` | Moved under `kubernetes ca` | 🟡 Moved |
| `containers, c` | Moved under `get containers` | 🟡 Moved |
| `copy, cp` | Moved under `run` | 🟡 Moved |
| `dashboard` | Moved under `run` | 🟡 Moved |
| `disks` | Moved under `get` | 🟡 Moved |
| `dmesg` | Moved under `run` | 🟡 Moved |
| `edit` | `edit` with `--patch` option | 🟡 Modified |
| `etcd` (all subcommands) | `etcd` with added `bootstrap` | 🟡 Modified |
| `events` | Removed | 🔴 Removed |
| `gen` (all subcommands) | `gen` | ⚪ Unchanged |
| `get` | `get` with expanded subcommands | 🟡 Modified |
| `health` | Moved under `get` | 🟡 Moved |
| `help` | `help` | ⚪ Unchanged |
| `image, images` | Moved under `get` | 🟡 Moved |
| `inject` | Removed | 🔴 Removed |
| `inspect` | Removed | 🔴 Removed |
[...]

PR #10287, opened 2025-02-04
refactor cluster create command

Orzelius:
### Refactor `cluster create` command

slightly related: https://github.com/siderolabs/talos/issues/10133

### Motivation and background

I attempted to make the qemu provider work with macOs, but decided to give the existing logic a brush up before adding another provider.

I had a hard time understanding which options were applicable to which providers, and thus it was also hard to tell what needs and doesn't need to be done to create the relevant providers' clusters. I thought that adding a macOs qemu provider, which would only support a subset of the linux options, would complicate the create logic too much so I decided to give it a rework before I proceed with adding the qemu macOs support. I will create an issue for the macOs provider feature to facilitate further discussion once/if this gets merged.

#### User facing
* Add two sub-commands `cluster create qemu` and `cluster create docker`
  * The commands only have flags relevant to the applicable providers
* Error if invalid provider's flags are passed (previously not all the flags were checked). This is technically a breaking change. Please let me know if it should be a warning at first.
* Add `(QEMU only)` or `(docker only)` at the start of the flag description
* In command description (e.g. `cluster create -h`) group flags by providers and then sort alphabetically. See results in the [reference docs](https://github.com/Orzelius/talos/blob/create-cmd-refactor/website/content/v1.10/reference/cli.md#talosctl-cluster-create)

#### Internal
* Split out common logic
  * cluster request to [`provision.ClusterRequestBase`](https://github.com/Orzelius/talos/blob/create-cmd-refactor/pkg/provision/request.go)
  * Common create command logic to [cluster/create/common.go](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/common.go)
  * Common [flags](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L304) and [options](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L27) 
* Split out provider specific logic
  * Request [vm.ClusterRequest](https://github.com/Orzelius/talos/blob/create-cmd-refactor/pkg/provision/providers/vm/request.go#L20) and [docker.ClusterRequest](https://github.com/Orzelius/talos/blob/create-cmd-refactor/pkg/provision/providers/docker/request.go#L13)
  * Create cmd functions: [CreateDockerCluster](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/docker.go#L18C6-L18C25) and [CreateQemuCluster](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/qemu_linux.go#L56)
  * Flags ([docker](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L227), [qemu](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L243)) and options ([docker](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L118), [qemu](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/create.go#L69))
* Add basic tests for commands and flags
* Add thorough unit tests for common logic 

#### unrelated
* [Use default kubeconfig](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/common.go#L391) instead of the [previous `kubeconfig.SinglePath()`](https://github.com/Orzelius/talos/blob/create-cmd-refactor/cmd/talosctl/cmd/mgmt/cluster/create/common.go#L391) which [errors if multiple configs are defined](https://github.com/siderolabs/go-kubeconfig/blob/main/kubeconfig.go#L34)
* [If KVM is not available, log a warning and proceed without it](https://github.com/Orzelius/talos/pull/4/files#diff-f891534971415a25f2238870c7659304248967ab401f21dd56f1abb33eeaa1b4R31-R39) (instead of a hard error). As I don't have access to a linux box this is the only way to test the qemu functionality for me

### Notes to reviewers

These changes most likely still contain bugs as I am not able to perform thorough testing on my mac machine. I've also failed to get the integration tests running locally, so I'd love to see the results of a ci run and will fix any issues if they come up.

Please let me know if I've missed something or if you have any ideas.


pkg/provision/providers/qemu/create.go
```diff
@@ -66,7 +79,10 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
 
 	fmt.Fprintln(options.LogWriter, "creating load balancer")
 
-	if err = p.CreateLoadBalancer(state, request); err != nil {
+	controlplanes := xslices.Filter(request.Nodes, func(n vm.NodeRequest) bool { return n.Type.IsControlPlane() })
+	controlplaneIps := xslices.Map(controlplanes, func(n vm.NodeRequest) string { return n.IPs[0].String() })
```
Orzelius, line 83: I couldn't think of a way to reuse the [provision.BaseNodeRequests.ControlPlaneNodes()](https://github.com/Orzelius/talos/blob/create-cmd-refactor/pkg/provision/request.go#L73) method with the [vm.NodeRequests](https://github.com/Orzelius/talos/blob/create-cmd-refactor/pkg/provision/providers/vm/request.go#L84) type. Please let me know if you can think of something to reduce this duplication here and in similar places

cmd/talosctl/cmd/mgmt/cluster/create/common_test.go
```diff
@@ -0,0 +1,472 @@
...
+	name       string
+}
+
+var testedFields = []string{}
+
+func TestAllCommonOptionFields(t *testing.T) {
+	for _, fieldTest := range fieldTests {
+		t.Run(fieldTest.name, fieldTest.test)
+	}
+
```
Orzelius, line 438: Test that there are tests for all the common fields.

Please let me know if this is inappropriate or if there's a less hacky way to achieve the same result :)

smira:
Before many any massive changes, it would be best to open a discussion first to make sure changes follow project direction.

I agree with overall direction of the change, but many things in the implementation are not the way they should be, so we could peek some parts as we get to refactoring talosctl, but we can't accept this PR in the way it is.

Orzelius:
> Before many any massive changes, it would be best to open a discussion first to make sure changes follow project direction.
> 
> I agree with overall direction of the change, but many things in the implementation are not the way they should be, so we could peek some parts as we get to refactoring talosctl, but we can't accept this PR in the way it is.

Makes sense haha
I'll open up an issue and lay out my learnings ad ideas to facilitate the discussion further. This pr can be revisited if appropriate.

smira:
First, let's split discussion about changing the provision library (I don't think we need to) and the frontend (CLI).

The provision library is being used in many places, so any change there should be done only if absolutely required. 

The frontend `talosctl cluster create` is a mess right now, and we have several ideas (including splitting up `docker`/`qemu`), creating "simple" version and "expert" versions of it.

Orzelius:
I still like to think that having Docker/Qemu/ProviderX specific options in the generic provision interface and types will lead to confusion, but as you said currently that's far from a problem. I might just not be aware, but the changes I made don't seem to cascade anywhere other than the cli and integration tests. Is there an omni or some other part of the picture I'm not seeing?

By the end of the day I'm just an engineer hungry for some fun open source work so I'll trust you on what's best for the project. I'd love to work on fixing up the fe if you could bother writing up a detailed request for how and what needs to be done. Feel free to continue discussion in the [issue](https://github.com/siderolabs/talos/issues/10289) :)

smira:
> Is there an omni or some other part of the picture I'm not seeing?

yes, in general you might assume that anything in `pkg/` is being used outside of Talos.

PR #10290, opened 2025-02-04
feat: provide initial support for ethtool configuration

smira:
See https://github.com/siderolabs/ethtool - our fork.

This PR covers only configuring rings, follow-up PRs will address other pieces: channels and features.

Example:

```
node: 172.20.0.5
metadata:
    namespace: network
    type: EthernetStatuses.net.talos.dev
    id: enp0s2
    version: 4
    owner: network.EthernetStatusController
    phase: running
    created: 2025-02-04T16:03:14Z
    updated: 2025-02-04T16:04:12Z
spec:
    linkState: true
    port: Other
    duplex: Unknown
    rings:
        rx-max: 256
        tx-max: 256
        rx: 128
        tx: 128
        tx-push: false
        rx-push: false
```


internal/app/machined/pkg/controllers/network/ethernet_status.go
```diff
@@ -0,0 +1,184 @@
...
+// EthernetStatusController reports Ethernet link statuses.
+type EthernetStatusController struct{}
+
+// Name implements controller.Controller interface.
+func (ctrl *EthernetStatusController) Name() string {
+	return "network.EthernetStatusController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *EthernetStatusController) Inputs() []controller.Input {
```
frezbo, line 35: can't we generate this based on ethernet spec inputs?

smira: sorry, I didn't get that, how to generate Inputs? 

this one watches Linux kernel and reports `ethtool` status, so it doesn't actually have any resource inputs.

pkg/machinery/resources/network/ethernet_spec.go
```diff
@@ -0,0 +1,72 @@
...
+
+// EthernetRingsSpec describes config of Ethernet rings.
+//
+//gotagsrewrite:gen
+type EthernetRingsSpec struct {
+	RX           *uint32 `yaml:"rx,omitempty" protobuf:"1"`
+	TX           *uint32 `yaml:"tx,omitempty" protobuf:"4"`
+	RXMini       *uint32 `yaml:"rx-mini,omitempty" protobuf:"2"`
+	RXJumbo      *uint32 `yaml:"rx-jumbo,omitempty" protobuf:"3"`
+	RXBufLen     *uint32 `yaml:"rx-buf-len,omitempty" protobuf:"5"`
```
dsseng, line 37: nit, the order of protobuf fields tx, then rxbuflen seems pretty weird. no practical implications though

internal/app/machined/pkg/controllers/network/ethernet_status.go
frezbo: ah got it, I was wondering why this does not have the spec as an input

dsseng (approved):

pkg/machinery/resources/network/ethernet_spec.go
smira: I tried to follow partially the order in `ethtool` and Linux netlink, but no particular order (fine to change it).

smira:
In the end I skipped the test in the CI, as I can't see from the resource state if this is "good" or "bad" QEMU, so just let it skip in the CI. We'll have more ethtool stuff.

PR #10298, opened 2025-02-05
fix: extracting cmdline multi profile UKIs

frezbo:
When extracting `.cmdline` from multi profile UKI's we only need the first cmdline, rest are profile specific and not the default cmdline.


dsseng (approved):

smira (approved):
should we  do in general for all sections? in case we have multiple kernels/initramfses later

DmitriyMV (approved):

smira (approved):

PR #10305, opened 2025-02-06
chore: upgrade siderolabs/go-loadbalancer

DmitriyMV:
It contains a breaking change, so we also need to update Talos.


DmitriyMV:
Depends on https://github.com/siderolabs/go-loadbalancer/pull/22

Issue #10187, opened 2025-01-21
Reproducible and Hermetic Builds


smira:
Talos builds are already reproducible for the most part, but we need to ensure that remaining missing bits are addressed.

Two main areas:

* make sure `toolchain` is reproducible (bootstrapped from a reproducible bootstrap toolchain)
* drop dependency on `alpine` from Talos `installer`/`imager` and Image Factory transitively

PR #10307, opened 2025-02-06
feat: use bootstrapped packages for building Talos

dsseng:
Update tools, pkgs and extras to use fully bootstrapped [Stageˣ]-derived toolchain for building Talos and its dependencies.

This brings in changes related to root being usrmerged now, so some paths have changed. Extras have been cleaned up: use only the needed package.

Addresses: #10187


hack/cleanup.sh
```diff
@@ -25,16 +22,16 @@ find ${PREFIX} -type f -name \*.la -delete
...
        ${PREFIX}/{share,usr/share}/* \
        ${PREFIX}/usr/lib/cmake \
-       ${PREFIX}/lib/gconv/ \
+       ${PREFIX}/usr/lib/gconv/ \
        ${PREFIX}/usr/libexec/getconf \
        ${PREFIX}/var/db
 
 # Remove contents of /usr/bin except for udevadm
-find ${PREFIX}/usr/bin \( -type f -o -type l \) ! -name udevadm -delete
+# find ${PREFIX}/usr/bin \( -type f -o -type l \) ! -name udevadm -delete
```
dsseng, line 37: Actually, some stuff like `c_rehash` and `openssl` aren't necessary. We should examine and remove unnecessary binaries from packages (or copy `/usr/lib` only for some)

dsseng: (cleanup was disabled as we now install containerd there, as `/bin` is a symlink now)

dsseng: <details>
<summary>squashfs tree (main dirs only)</summary>

```
.
├── bin -> usr/bin
├── boot
│   └── EFI
├── dev
├── etc
│   ├── ca-certificates -> /etc/ssl
│   ├── cni
│   │   └── net.d
│   ├── containerd
│   ├── cri
│   │   └── conf.d
│   ├── iscsi
│   ├── kubernetes
│   │   └── manifests
│   ├── lvm
│   │   └── profile
│   ├── nvme
│   ├── pki -> /etc/ssl
│   ├── selinux
│   │   └── targeted
│   └── ssl
│       └── certs
├── lib -> usr/lib
├── lib64 -> usr/lib
├── mnt
├── opt
│   ├── cni
│   │   └── bin
│   └── containerd
│       ├── bin
│       └── lib
├── proc
├── root
├── run
│   └── lock
├── sbin -> usr/sbin
├── sys
├── system
├── tmp
├── usr
│   ├── bin
│   ├── etc
│   ├── include
│   ├── lib
│   │   ├── cryptsetup
│   │   ├── engines-3
│   │   ├── firmware
│   │   ├── modules
│   │   ├── ossl-modules
│   │   ├── systemd
│   │   ├── udev
│   │   └── xtables
│   ├── lib64 -> lib
│   ├── libexec
│   │   └── kubernetes
│   ├── local
│   │   ├── lib
│   │   └── share
│   ├── sbin
│   ├── share
│   │   ├── ca-certificates -> /etc/ssl
│   │   └── zoneinfo
│   ├── src
│   └── var
│       └── lib
└── var
    ├── lock -> ../run/lock
    ├── log
    ├── mail
    ├── run -> ../run
    ├── spool
    └── tmp
```

</details>

Dockerfile
```diff
@@ -650,36 +645,36 @@ COPY --from=pkg-sd-boot /*.efi.stub /sd-stub-${TARGETARCH}.efi
...
-COPY --from=pkg-kernel-arm64 /lib/modules lib/modules
+COPY --from=pkg-kernel-arm64 /usr/lib/modules usr/lib/modules
 RUN <<EOF
 set -euo pipefail
 
-KERNEL_VERSION=$(ls lib/modules)
+KERNEL_VERSION=$(ls usr/lib/modules)
 
-xargs -a modules-arm64.txt -I {} install -D lib/modules/${KERNEL_VERSION}/{} /build/lib/modules/${KERNEL_VERSION}/{}
+xargs -a modules-arm64.txt -I {} install -D usr/lib/modules/${KERNEL_VERSION}/{} /build/usr/lib/modules/${KERNEL_VERSION}/{}
```
smira, line 671: This change would need a quirk, as it would affect how we merge kernel modules when installing extensions

dsseng: quirk in what part? and yes, this is going to be a breaking change as extensions are now disallowed to have `/lib` on their rootfs

smira: quirk in Talos/imager/Image Factory, as Image Factory handles any version of Talos

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -129,7 +129,7 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: "/dev", Source: "/dev", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "sysfs", Destination: "/sys", Source: "/sys", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: constants.CgroupMountPath, Source: constants.CgroupMountPath, Options: []string{"rbind", "rshared", "rw"}},
-		{Type: "bind", Destination: "/lib/modules", Source: "/lib/modules", Options: []string{"bind", "ro"}},
+		{Type: "bind", Destination: "/usr/lib/modules", Source: "/usr/lib/modules", Options: []string{"bind", "ro"}},
```
smira, line 132:
```suggestion
		{Type: "bind", Destination: "/lib/modules", Source: "/usr/lib/modules", Options: []string{"bind", "ro"}},
```

pkg/machinery/extensions/extensions.go
```diff
@@ -10,12 +10,12 @@ import "path/filepath"
...
-	"/lib/modules",
+	"/usr/lib/firmware",
+	"/usr/lib/modules",
 	// The glibc loader is required by glibc dynamic binaries.
-	"/lib64/ld-linux-x86-64.so.2",
-	// /sbin/ldconfig is required by the nvidia container toolkit.
-	"/sbin/ldconfig",
+	"/usr/lib/ld-linux-x86-64.so.2",
+	// /usr/bin/ldconfig is required by the nvidia container toolkit.
+	"/usr/bin/ldconfig",
```
smira, line 18: @frezbo ?

frezbo: na, we can't do this, it neds to be `/sbin/ldconfig` which is standard expected for glibc

website/content/v1.10/advanced/proprietary-kernel-modules.md
```diff
@@ -55,7 +55,7 @@ aliases:
         ```bash
         INSTALLER_VERSION=<talos version>
         IMAGE_NAME="ghcr.io/your-username/talos-installer:$INSTALLER_VERSION"
-        DOCKER_BUILDKIT=0 docker build --build-arg RM="/lib/modules" -t "$IMAGE_NAME" . && docker push "$IMAGE_NAME"
```
smira, line 58: let's drop this document

pkg/machinery/extensions/extensions.go
smira: the idea is that in runtime we have a symlink, e.g. `/sbin` -> `/usr/bin/ldconfig`

frezbo: hmm that should work then, but I guess extensions should still allow the path :thinking:

hack/release.toml
```diff
@@ -107,6 +107,17 @@ Talos starting with 1.10 will have ISO's that will use GRUB only for legacy BIOS
         description = """\
 Talos now provides `ethtool`-style Ethernet low-level configuration via `network/EthernetConfig` documents.
 Current status of the interface can be read by `talosctl get ethernetstatus`.
+"""
+
+    [notes.toolchain]
+        title = "Fully bootstrapped builds"
+        description = """\
+Talos 1.10 is built with a toolchain based on [\[Stageˣ\]](https://stagex.tools/), which is a project building fully bootstrapped software.
```
dsseng, line 115: unsure about square brackets and how are they escaped when generating changelogs

smira: do `make release-notes` and you'll see :)

internal/integration/api/extensions_qemu.go
```diff
@@ -68,14 +68,14 @@ func (suite *ExtensionsSuiteQEMU) TearDownTest() {
...
 	expectedPaths := []string{
-		"/lib/firmware/amdgpu",
-		"/lib/firmware/amd-ucode",
-		"/lib/firmware/bnx2x",
-		"/lib/firmware/cxgb3",
-		"/lib/firmware/cxgb4/configs",
-		"/lib/firmware/i915",
-		"/lib/firmware/intel/ice/ddp",
-		"/lib/firmware/intel-ucode",
+		"/usr/lib/firmware/amdgpu",
```
frezbo, line 71: I think we can keep existing paths, this also ensures that the symlinks work

dsseng: should do it

smira (approved):
This might break system extensions containing kernel modules, but we can fix it next step.

Issue #10190, opened 2025-01-21
support multi-profile UKIs and options


smira:
To achieve feature parity with GRUB, we need to support multi-profile UKIs to implement custom command line for the following options:

* wipe disk fully (ISO boot)
* wipe system partitions (disk boot)

PR #10316, opened 2025-02-08
feat: multi profile UKIs

frezbo:
Support generating multi profile UKIs.
This PR adds the default wipe options.

Supporting extra profiles via imager would be another PR.

Fixes: #10190


internal/pkg/uki/generate.go
```diff
@@ -183,6 +185,87 @@ func (builder *Builder) generatePCRPublicKey() error {
 	return nil
 }
 
+func (builder *Builder) generateProfiles() error {
+	defaultProfile := Profile{
```
smira, line 189: can we make this more generic e.g. by doing something like

```suggestion
for _, profile := range []struct{
  name string
  label string
  extraArgs []string
}{
{
   name: "main",
},
{
   name: "reset",
   label: "....",
   extraArgs: []string{....},
 }
}
	
```

frezbo: I wanted imager to be able to pass in extra profile, I can make this private though

smira: we can refactor it to be public structure, it's all good, but at least we DRY this piece a bit

internal/pkg/uki/generate.go
```diff
@@ -183,6 +185,59 @@ func (builder *Builder) generatePCRPublicKey() error {
 	return nil
 }
 
+func (builder *Builder) generateProfiles() error {
+	for _, profile := range []Profile{
```
smira, line 189: also I think we should generate profile-less UKIs for Talos < 1.10?

internal/pkg/uki/generate.go
```diff
@@ -206,40 +261,72 @@ func (builder *Builder) generateKernel() error {
...
-	if err != nil {
-		return err
+	if previousProfileIndex != 0 {
+		profileIndexes[len(profileIndexes)-1].End = len(toMeasure)
 	}
 
-	path := filepath.Join(builder.scratchDir, "pcrpsig")
+	for i, profileIndex := range profileIndexes {
+		profileData := xslices.ToMap(
+			slices.Concat(toMeasure[:profileIndexes[0].Start], toMeasure[profileIndex.Start:profileIndex.End]),
```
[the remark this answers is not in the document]
frezbo: no, so we measure all sections before the first `.profile` section upto the start of the next profile, eg if we have:

```text
.text
.cmdline
.initrd
.linux
.profile
.profile
.cmdline
```

the first pcrsig is measured as `.cmdline (the first one) + .initrd + .linux + .profile` and injected after the first `.profile`

the next pcrsig is measured as `.initrd + .linux + .profile (second profile) + .cmdline (second cmdline)` and injected after the second `.cmdline`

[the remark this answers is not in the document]
frezbo:
```text
Note that if the PCR signature mechanism described above is used it is recommended to include a separate .pcrsig PE section in each profile matching precisely the sections that apply to that profile (i.e. the combination of the profile’s own sections and those of the base section).
```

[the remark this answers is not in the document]
frezbo: > This concat is s[:i]+s[i:j], so it's s[:j]?

This is true for first profile, second measurement excludes the previous `.profile` section and it components and so on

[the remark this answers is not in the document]
smira: sorry, I missed the fact that it has different indexes, now I get it. 

So probably we only need to store indexes to the `.profile` sections, and then it becomes `[:profile[0]] + [profile[i-1]+1:[profile[i]]`, but in the end it doesn't mattter

internal/pkg/uki/generate.go
```diff
@@ -206,40 +266,72 @@ func (builder *Builder) generateKernel() error {
...
-	sectionsData := xslices.ToMap(
-		xslices.Filter(builder.sections,
-			func(s section) bool {
-				return s.Measure
-			},
-		),
-		func(s section) (string, string) {
-			return s.Name, s.Path
-		})
+	profileIndexes := []profileIndex{}
```
smira, line 275: does it work with profile-less UKIs? (I don't see a test for it?)

smira (approved):

internal/pkg/uki/generate.go
frezbo: good point, let me add a test

PR #10324, opened 2025-02-10
feat+refactor: ability to skip some of the default checks

shellwhale:
# Pull Request

## What? (description)
Define the default checks in a map instead of directly returning them in a slice.
Add a new function `PreBootSequenceChecksFiltered` to allow users to purposefully skip some checks.

## Why? (reasoning)
While using the Terraform Talos Provider and a custom CNI (Cilium), if installing Cilium via Terraform (Terraform Helm Provider) (not doing it via the machine config file) the [talos_cluster_health](https://registry.terraform.io/providers/siderolabs/talos/latest/docs/data-sources/cluster_health) data source fails.

See https://github.com/siderolabs/terraform-provider-talos/issues/206

This PR would allow to modify the [Talos Terraform Provider](https://github.com/siderolabs/terraform-provider-talos/blob/main/pkg/talos/talos_cluster_health_data_source.go#L130 ) (which use this package) in a way that would provide users with the ability to skip some of the checks defined in `default.go`, one of those checks would be `"kubelet to be healthy"` in the case of https://github.com/siderolabs/terraform-provider-talos/issues/206.

Perhaps better names for each of the individuals checks in this new `map[string]ClusterCheck` would be suitable, or maybe another solution all together.


pkg/cluster/check/default.go
```diff
@@ -13,57 +13,124 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/config/machine"
 )
 
+// checks is a map containing all our cluster checks defined once.
+var checks = map[string]ClusterCheck{
```
smira, line 17: let's not use global maps here, it can be a private function with a big `switch` statement and panic on `default:` case

smira: and check names should be constants

pkg/cluster/check/default.go
```diff
@@ -103,70 +203,40 @@ func ExtraClusterChecks() []ClusterCheck {
...
-			}, time.Minute, 5*time.Second)
-		},
-
-		// wait for kubelet to be healthy on all
-		func(cluster ClusterInfo) conditions.Condition {
-			return conditions.PollingCondition("kubelet to be healthy", func(ctx context.Context) error {
-				return ServiceHealthAssertion(ctx, cluster, "kubelet", WithNodeTypes(machine.TypeInit, machine.TypeControlPlane))
-			}, 5*time.Minute, 5*time.Second)
-		},
+	names := preBootSequenceCheckNames()
```
smira, line 223: this won't pass linting - `make lint-go`

and this should be `return PreBootSequenceChecksFiltered(nil)`

shellwhale:
Hello, is there something that I can do to complete this merge?

smira:
> Hello, is there something that I can do to complete this merge?

to address the review feedback?

shellwhale:
@smira done

shellwhale:
@smira done

smira:
> @smira done

I don't see any changes, still everything is prefixed with `Check`, so it is `check.Check*`

shellwhale:
> > @smira done
> 
> I don't see any changes, still everything is prefixed with `Check`, so it is `check.Check*`

I cannot do that because of 

```go
	case CheckAllNodesMemorySizes // Cannot drop the check word
		return func(cluster ClusterInfo) conditions.Condition {
			return conditions.PollingCondition(CheckAllNodesMemorySizes, func(ctx context.Context) error {
				return AllNodesMemorySizes(ctx, cluster) // Because it would result in the same name as this
			}, 5*time.Minute, 5*time.Second)
		}
```

smira:
> I cannot do that because of

probably these functions can be made private now? they should not be used outside of this package?

shellwhale:
> > I cannot do that because of
> 
> probably these functions can be made private now? they should not be used outside of this package?

Maybe but this is pkg, not internal, so that would be a breaking change, is that what we want?

smira:
> > > I cannot do that because of
> > 
> > 
> > probably these functions can be made private now? they should not be used outside of this package?
> 
> Maybe but this is pkg, not internal, so that would be a breaking change, is that what we want?

Anyways it's a big change, so I'm ok with making it even bigger as long as it is consistent and bigger API

shellwhale:
@smira can you help ?
I don't understand why I can't merge from main to this branch

Issue #10289, opened 2025-02-04
talosctl cluster create command refactor


Orzelius:
related: #10133, #10287

## User facing

* Often no error is provided when wrong provider's flag are passed
* Sometimes it's hard to tell which flags are applicable for which providers

### Developer side

* Both docker and qemu clusters are created via the same flow
  * Same options are crafted and passed in both cases
  * Both providers take the same [clusterRequest](https://github.com/siderolabs/talos/blob/main/pkg/provision/request.go#L24), but result in different clusters
* Currently the logic that crafts the request and executes on it is tightly coupled making it hard to test
* No unified validation step exists making it hard to tell which options are yet to be checked

## Proposed solution

* Group options and flags as follows: common, qemu-specific, docker-specific
* Split out common logic between both providers
* Only pass applicable options to relevant providers
* Split out logic that does validation and the creation of the request data and create tests for it

Here's my go at it https://github.com/siderolabs/talos/pull/10287

Please share any ideas or thoughts about the implementation.
I've already wrapped my head around most of the bits around the cluster creation logic so I'd love to work on this :)

Orzelius:
### Here's my RFC for the implementation:

Please let me know if you have any thoughts or ideas on how to implement this further. The proposed changes only effect the front end cli and require no changes to the provision package. cc @smira 

**1. Split options and flags as follows: common, qemu-specific and docker-specific**
```go
// CommonOps are the options common across all the providers.
type CommonOps struct {
	Talosconfig               string
	KubernetesVersion         string
        ...
}
// qemuOps are the qemu provider specific options
type qemuOps struct {
	nodeInstallImage             string
	nodeVmlinuzPath              string
        ...
}
// dockerOps are the options specific to the docker provider
type dockerOps struct {
	dockerHostIP      string
        ...
}

// Same for flags

// Result:
type createOps struct {
	common *CommonOps
	docker *dockerOps
	qemu   *qemuOps
}

type createFlags struct {
	common *pflag.FlagSet
	docker *pflag.FlagSet
	qemu   *pflag.FlagSet
}
```

**2. Add two more commands under create**
  a. `create qemu` that only takes qemu flags and only creates a qemu cluster
  b. `create docker` that only takes docker flags and only creates a docker cluster
  c. (optional)`create qemu-basic` or something similar. Only takes basic qemu flags and only creates a qemu cluster
The create command itself can be kept as is for backwards compatibility for a while.

[...]

PR #10334, opened 2025-02-11
refactor cluster create command (frontend only)

Orzelius:
closes https://github.com/siderolabs/talos/issues/10289 
related: https://github.com/siderolabs/talos/issues/10133, https://github.com/siderolabs/talos/pull/10287

#### User facing
* Add two sub-commands `cluster create qemu` and `cluster create docker`
  * The commands only have flags relevant to the applicable providers (`cluster create` still functions all the same)
* Error if invalid provider's flags are passed (both new and old `cluster create` commands) (previously not all the flags were checked). This is technically a breaking change. Please let me know if it should be a warning at first.
* Add `(QEMU only)` or `(docker only)` at the start of the flag description
* In command description (e.g. `cluster create -h`) group flags by providers and then sort alphabetically.

See results to the command documentation in the [reference docs](https://github.com/Orzelius/talos/blob/create-refactor-2/website/content/v1.10/reference/cli.md#talosctl-cluster-create)

#### Internal

* Split common code into the [`clustermaker`](https://github.com/Orzelius/talos/tree/create-refactor-2/cmd/talosctl/cmd/mgmt/cluster/create/clustermaker) package
* Split qemu and docker code
* Add basic unit tests for all options
* Split common and provider specific cli options into separate structs and flag groups.

#### other
* [Use default kubeconfig](https://github.com/Orzelius/talos/blob/create-refactor-2/cmd/talosctl/cmd/mgmt/cluster/create/clustermaker/util.go#L56) instead of the previous `kubeconfig.SinglePath()`which [errors if multiple configs are defined](https://github.com/siderolabs/go-kubeconfig/blob/main/kubeconfig.go#L34)

### Notes to reviewers

This might look like a massive change, but it's mostly just moving existing logic around with added tests. I've failed to get the integration tests running locally, so I'd love to see the results of a ci run and will fix any issues if they come up. As I can't perform thorough qemu testing on my mac machine I'm fairly certain some bugs remain, but they should be easy to fix once found thanks to the testable logic and clear separation of concerns.

Please let me know if I've missed something or if you have any ideas <3


cmd/talosctl/cmd/mgmt/cluster/create/clustermaker/maker_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+type fieldTest = struct {
+	fieldNames []string
+	test       func(t *testing.T)
+	name       string
+}
+
+var testedFields = []string{}
+
+func TestAllOptionFields(t *testing.T) {
```
Orzelius, line 423: Test that there are tests for all the common fields.

Please let me know if this is inappropriate or if there's a less hacky way to achieve the same result or if I just change it to good ol' fashioned tests

Orzelius:
Update: rebased latest changes from main and removed the KVM change to reduce the scope of the pr.

Orzelius:
If this is too big of a change to confidentially merge all at once please let me know. It would take some extra work, but I could split it into smaller, easier to review pieces. For example as follows:
1. Split flags and options into separate provider specific structs
2. Dependency inject provider and add tests
3. Split out common logic
4. Split qemu and docker logic
5. Add separate `create docker` and `create qemu` commands

This would result in a easier to review diffs where possible mistakes are easier to spot.

If you generally disagree with the approach or implementation on broader terms please describe a more suitable solution in the https://github.com/siderolabs/talos/issues/10289 issue.

Orzelius:
rebased latest changes

Orzelius:
closing in favor of implementing this in multiple smaller prs to ease the review process and catch more bugs

PR #10337, opened 2025-02-11
feat: support uki profiles via imager

frezbo:
Support passing uki profiles from imager.

Part-of: #10190

### DIFF
--- a/pkg/imager/imager.go
+++ b/pkg/imager/imager.go
@@ -441,6 +441,22 @@ func (i *Imager) buildUKI(ctx context.Context, report *reporter.Reporter) error 
 		OutUKIPath:    i.ukiPath,
 	}
 
+	switch i.prof.Output.Kind { //nolint:exhaustive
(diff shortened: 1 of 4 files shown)


pkg/imager/imager.go
```diff
@@ -441,6 +441,22 @@ func (i *Imager) buildUKI(ctx context.Context, report *reporter.Reporter) error
 		OutUKIPath:    i.ukiPath,
 	}
 
+	switch i.prof.Output.Kind { //nolint:exhaustive
+	case profile.OutKindISO:
+		builder.Profiles = append(builder.Profiles, uki.Profile{
+			ID:      "reset",
+			Title:   "Reset system disk",
+			Cmdline: builder.Cmdline + " talos.experimental.wipe=system",
```
smira, line 449: I believe we have constants for the kernel args...

smira (approved):

PR #10343, opened 2025-02-11
fix: fix diff printing

Mathspy:
The second `@@` was not written to the same writer

This was breaking ability to pipe the diff output to tools that understand [`unidiff`](https://en.wikipedia.org/wiki/Diff#Unified_format)


smira (approved):
thank you! 

this `@@` in the console was bugging me for a long time!

PR #10354, opened 2025-02-12
fix: make ingress firewall filter traffic to nodeports

smira:
This fixes #10347

The core issue was that Talos nftables chain had priority 0 (`filter`), while kube-proxy does DNAT for node ports at priority -110 (before Talos can see source traffic), so Talos rule doesn't match.

Move Talos priority to -140, so it runs before kube-proxy.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -119,6 +119,12 @@ This also changes Talos root filesystem structure for unified /usr, with other d
 System extensions must move their directories accordingly for 1.10.
 """
 
+    [notes.ingress-firewall]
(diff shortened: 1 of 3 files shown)


smira:
`kube-proxy` rule in question:

```
        chain filter-input {
                type filter hook input priority -110; policy accept;
                ct state new jump nodeport-endpoints-check
                ct state new jump service-endpoints-check
        }
```

With the fix above, our rule sits before this one.

frezbo (approved):
release notes missing

Issue #10349, opened 2025-02-12
OS Installer Fails to Set Correct Default Boot Option on UEFI Systems with NVMe Drive


ihelmer07:
### Description
(Trying) to install Talos onto an Beelink EQ14 with an NVMe drive.
After successfully installing Talos via USB (following this [great video](https://www.youtube.com/watch?v=VKfE5BuqlSc)) restarting the computer causes the system to go directly to UEFI shell, without launching Talos.

Seems like a similar issue was already reported here with no resolution: https://github.com/siderolabs/talos/issues/7391
Trying to get some insight on how to solve as I am running into a similar issue but there is no EFI present.

Running `map -r` in the UEFI shell doesn't display any FS drives. Only BLK0.

Any help is appreciated.

### Steps to Reproduce
Get Beelink EQ14 with single NVME drive.
Install Talos via USB
Restart Beelink.

### Logs
N/A
### Environment
N/A

frezbo:
Could you also try new [v1.10.0-alpha.1](https://github.com/siderolabs/talos/releases/tag/v1.10.0-alpha.1) ISO's from Image Factory, just wanted to see if sd-boot fixes this issue

smira:
@frezbo I think we don't see anything related to boot order today in the sd-boot (our) code.

smira:
tl;dr we need to do equivalent of `efibootmgr` to make sure that the disk Talos is installed to is in the EFI boot order

ihelmer07:
Boot Order does nothing. I've switched the boot order around, still goes to the UEFI shell upon restart.

Trying v1.10.0-alpha.1 now.

ihelmer07:
v1.10-alpha.1 version caused the same problem. 

Any ideas on what to do?
It seems that its not installing Talos onto the NVM drive.

ihelmer07:
I'm able to install other OSs on the system just fine. 
Could you guide me on how to supply the relevant logs?

I'm happy and willing to do whatever is needed to troubleshoot!

smira:
I don't know what is available for your for debugging, but Talos to install logs to kernel console - if the install fails, you should see that in the console. If it scrolls too fast, etc., and you have serial console, you can add `console=ttyS0` to the initial boot and look via serial console. 

If you can capture the logs and post it here, it's perfect. If you can't use serial console - even screen recording might work.

ihelmer07:
I was able to figure it out!

[...]

PR #10355, opened 2025-02-12
Update getting-started.md

ihelmer07:
## What? (description)

Clarifies even further that Talos is loaded to RAM until configuration is applied. Really trying to hammer it home. see #10349 

## Why? (reasoning)

Talos loading to RAM first is a unique aspect, outside the typical. Really hammering it home that if one restarts the machine, the OS is 'gone' and needs to be restarted. The reason https://github.com/siderolabs/talos/issues/10349 was because I went in with the assumption that the ISO wasn't needed any more before applying a configuration, which got me into the loop of UEFI shell.

### DIFF
--- a/website/content/v1.9/introduction/getting-started.md
+++ b/website/content/v1.9/introduction/getting-started.md
@@ -86,13 +86,17 @@ The endpoint should be formatted like:
 When Talos boots without a configuration, such as when booting off the Talos ISO, it
 enters maintenance mode and waits for a configuration to be provided.
 
-> A configuration can be passed in on boot via kernel parameters or metadata servers.
-> See [Production Notes]({{< relref "prodnotes#configure-talos" >}}).
+> NOTE: Talos initially loads the OS to RAM, and only installs to disk after the configuration is applied.
+> If you reboot the machine before applying machine config, make sure your boot media is still present.
 
 Unlike traditional Linux, Talos Linux is _not_ configured by SSHing to the server and issuing commands.
 Instead, the entire state of the machine is defined by a `machine config` file which is passed to the server.
 This allows machines to be managed in a declarative way, and lends itself to GitOps and modern operations paradigms.
+
 The state of a machine is completely defined by, and can be reproduced from, the machine configuration file.
+
+> A configuration can be passed in on boot via kernel parameters or metadata servers.
+> See [Production Notes]({{< relref "prodnotes#configure-talos" >}}).
 
 To generate the machine configurations for a cluster, run this command on the workstation where you installed `talosctl`:
 
--- a/website/content/v1.10/introduction/getting-started.md
+++ b/website/content/v1.10/introduction/getting-started.md
@@ -86,13 +86,17 @@ The endpoint should be formatted like:
 When Talos boots without a configuration, such as when booting off the Talos ISO, it
 enters maintenance mode and waits for a configuration to be provided.
 
-> A configuration can be passed in on boot via kernel parameters or metadata servers.
-> See [Production Notes]({{< relref "prodnotes#configure-talos" >}}).
+> NOTE: Talos initially loads the OS to RAM, and only installs to disk after the configuration is applied.
+> If you reboot the machine before applying machine config, make sure your boot media is still present.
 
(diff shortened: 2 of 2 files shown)


website/content/v1.9/introduction/getting-started.md
```diff
@@ -84,16 +85,18 @@ The endpoint should be formatted like:
 ## Configure Talos Linux
 
 When Talos boots without a configuration, such as when booting off the Talos ISO, it
-enters maintenance mode and waits for a configuration to be provided.
-
-> A configuration can be passed in on boot via kernel parameters or metadata servers.
-> See [Production Notes]({{< relref "prodnotes#configure-talos" >}}).
+enters maintenance mode and waits for a configuration to be provided.  
+> NOTE: Talos initially loads the OS to RAM, and only installs to disk after the configuration is applied. Removing the ISO and
```
rothgar, line 89: I don't think your finished the sentence here.

ihelmer07: You're right - embarrassing. Fixed.

website/content/v1.9/introduction/getting-started.md
```diff
@@ -54,7 +54,8 @@ The latest ISO image can be found on the Github [Releases](https://github.com/si
 - ARM64: [https://github.com/siderolabs/talos/releases/download/{{< release >}}/metal-arm64.iso](https://github.com/siderolabs/talos/releases/download/{{< release >}}/metal-arm64.iso)
 
 When booted from the ISO, Talos will run in RAM and will not install to disk until provided a configuration.
-Thus, it is safe to boot any machine from the ISO.
+Thus, it is safe to boot any machine from the ISO. 
+This also means restarting the machine without applying a configuration will restart the process.
```
smira, line 58: this seems to repeat the  sentence at line 56 in different words

website/content/v1.9/introduction/getting-started.md
```diff
@@ -84,16 +85,18 @@ The endpoint should be formatted like:
 ## Configure Talos Linux
 
 When Talos boots without a configuration, such as when booting off the Talos ISO, it
-enters maintenance mode and waits for a configuration to be provided.
-
-> A configuration can be passed in on boot via kernel parameters or metadata servers.
-> See [Production Notes]({{< relref "prodnotes#configure-talos" >}}).
+enters maintenance mode and waits for a configuration to be provided.  
+> NOTE: Talos initially loads the OS to RAM, and only installs to disk after the configuration is applied. Removing the ISO and restarting the machine requires restarting the process completely.
```
smira, line 89: I think it should be reworded, as Talos can be booted via a disk image. 

I don't see why restart is _that_ big of a problem. If the machine was ISO-booted, or PXE-booted, it's natural that on reboot it will boot once again from the same media.

If the machine was booted from a disk image, this note doesn't apply, but it confuses even more.

smira:
To be fair, I can't remember any OS that if booted from an ISO would install itself without any interaction from the user, so I don't see how Talos is unique here

ihelmer07:
I hear with what you're saying.
My admittedly relatively inexperienced perspective is that a 'usual' USB image install process is:
1. Launch iso from usb.
2. Follow prompts if any.
3. Installed & OS displays

Many OS have restart protections 

Talos is:
1. Launch from ISO.
2. No prompts but lots of "installing"
3. Talos Summary page displays. <-- assumed it was 'done'
4. Apply config
5. Talos actually installed

The point is Talos has extra steps - Talos Step 3 is not the same as "usual" step 3. 
My usual 'how-to-install' assumptions were subverted with Talos which is the point caused the confusion. 

I'm only trying to prevent similar situation and make the documentation the best it can be.
Happy to remove/change as needed.

ihelmer07:
@smira - Your wording is better. I've updated accordingly and removed line 58 as well.

Good to hear that 1.11 will make things more straightforward.

website/content/v1.9/introduction/getting-started.md
ihelmer07: reworded.

smira (approved):
thank you!

PR #10370, opened 2025-02-14
feat: support noclooud instance-id from dmi

sergelogvinov:
# Pull Request

## What? (description)

The official documentation has a "i" value that is used as the instance-id meta parameter.
https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html

## Why? (reasoning)

```shell
-smbios type=1,serial=ds=nocloud;h=my-host-name;i=id-1234
```

I know that the cloud (Proxmox) sets a randomly generated instance-id in the metadata file. This value cannot be reliably used in the future by code, as there is no mapping to a real ID, UUID, or other identifiers.

This PR aims to override the randomly generated Proxmox value.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
@@ -248,8 +248,8 @@ func (n *Nocloud) acquireConfig(ctx context.Context, r state.State) (metadataNet
 	}
 
 	var (
-		metaBaseURL, hostname string
-		networkSource         bool
+		metaBaseURL, hostname, instanceID string
+		networkSource                     bool
 	)
 
 	options := strings.Split(s.SystemInformation.SerialNumber, ";")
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
```diff
@@ -296,6 +299,10 @@ func (n *Nocloud) acquireConfig(ctx context.Context, r state.State) (metadataNet
 		metadata.Hostname = hostname
 	}
 
+	if instanceID != "" {
```
smira, line 302: should we overwrite here `metadata.InstanceID` if it's non-empty (something was downloaded?)

sergelogvinov: Goot question. 

It can only be set using the special string format: `-smbios type=1,serial=ds=nocloud;h=my-host-name;i=id-1234`. By default, this value is empty.

I know that the cloud (Proxmox) sets a randomly generated instance-id in the metadata file. This value cannot be reliably used in the future by code, as there is no mapping to a real ID, UUID, or other identifiers.

This PR aims to override the randomly generated Proxmox value.

PS. The hostname in smbios has priority too.

Thanks

smira: sounds good, thanks for clarifying!

smira (approved):

PR #10379, opened 2025-02-14
refactor: ephemeral mount

smira:
Mount EPHEMERAL volume via new controller.

Implement the first cut of volume dependencies from services, refactor
the way system disk wipe works.

Update volume manager controller to destroy volume statuses on shutdown,
which allows to signal mount operations to be terminated.

Lots of WIP ideas still, but I want to complete this PR and move on to
the next one:

* refactor user disks mounts to use new API
* refactor STATE mountes to use new API
* implement directory/overlay mounts


pkg/logging/zap.go
```diff
@@ -138,9 +138,9 @@ func ZapLogger(dests ...*LogDestination) *zap.Logger {
 			dest.level,
 		)
 
-		if dest.suppressThreshold > 0 {
```
smira, line 141: revert me

smira:
The goal of this PR was to change the way EPHEMERAL is mounted, to use new controllers for mounting.

On the way I refactored volume manager to to proper teardown of volume statuses which is not being used actively yet, but it should in the next PR(s).

The next PR should handle directory & overlay volumes refactoring more core parts and specifying correct dependencies for services.

internal/app/machined/pkg/controllers/block/volume_manager.go
```diff
@@ -326,60 +337,70 @@ func (ctrl *VolumeManagerController) Run(ctx context.Context, r controller.Runti
...
+				if len(volumeStatus.TypedSpec().EncryptionFailedSyncs) > 0 {
+					fields = append(fields, zap.Strings("encryptionFailedSyncs", volumeStatus.TypedSpec().EncryptionFailedSyncs))
 				}
 
 				volumeLogger.Info("volume status", fields...)
 			}
 
-			allClosed = allClosed && volumeStatus.Phase == block.VolumePhaseClosed
+			// when closing, ignore META volume, we want it to stay longer, so no problem if is not closed yet
+			allClosed = allClosed && (volumeStatus.TypedSpec().Phase == block.VolumePhaseClosed || vc.Metadata().ID() == constants.MetaPartitionLabel)
```
frezbo, line 385: can't we have a way to say the closing wave for meta needs to be last, so we can do proper dependency?

internal/app/machined/pkg/controllers/runtime/mount_status.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			if err = safe.WriterModify(ctx, r, runtime.NewMountStatus(runtime.NamespaceName, volumeStatus.Metadata().ID()),
+				func(res *runtime.MountStatus) error {
+					res.TypedSpec().Source = mountStatus.TypedSpec().Source
+					res.TypedSpec().Target = mountStatus.TypedSpec().Target
+					res.TypedSpec().FilesystemType = volumeStatus.TypedSpec().Filesystem.String()
+					res.TypedSpec().Encrypted = volumeStatus.TypedSpec().EncryptionProvider != block.EncryptionProviderNone
+
+					// [TODO]: fill in encryption providers
```
frezbo, line 85: still needs to be done, or separate PR?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1713,19 +1713,47 @@ func UnmountStatePartition(runtime.Sequence, any) (runtime.TaskExecutionFunc, st
 // MountEphemeralPartition mounts the ephemeral partition.
 func MountEphemeralPartition(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
 	return func(ctx context.Context, logger *log.Logger, r runtime.Runtime) error {
-		if _, err := waitForVolumeReady(ctx, r, constants.EphemeralPartitionLabel); err != nil {
-			return err
+		mountRequest := blockres.NewVolumeMountRequest(blockres.NamespaceName, constants.EphemeralPartitionLabel)
+		mountRequest.TypedSpec().VolumeID = constants.EphemeralPartitionLabel
+		mountRequest.TypedSpec().Requester = "sequencer"
+		// [TODO]: project quota support
```
frezbo, line 1719: same with the TODO here?

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -184,6 +184,11 @@ func (e *Etcd) DependsOn(runtime.Runtime) []string {
 	return []string{"cri"}
 }
 
+// Volumes implements the Service interface.
+func (e *Etcd) Volumes() []string {
+	return nil
```
frezbo, line 189: doesn't etcd need ephemeral?

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -105,6 +105,11 @@ func (k *Kubelet) DependsOn(runtime.Runtime) []string {
 	return []string{"cri"}
 }
 
+// Volumes implements the Service interface.
+func (k *Kubelet) Volumes() []string {
+	return nil
```
frezbo, line 110: same with kubelet, no ephemeral

internal/app/machined/pkg/controllers/block/volume_manager.go
smira: this is a bit tricky:

* at shutdown, sequencer notifies this controller that all volumes should be closed, and waits for the controller to close them
* so we should close and report that all are closed
* but at the same time META can be accessed after that, and it uses volume to find it
* so this is a bit of a hack, and probably we should do this some other way

internal/app/machined/pkg/controllers/runtime/mount_status.go
smira: yes, need to do this still

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: I put this TODO in too many places, I have it fixed, thanks, I'll remove it

internal/app/machined/pkg/system/services/etcd.go
smira: it needs `/var/lib/etcd`, but I don't have this implemented yet :) so I will add this with directory volumes

internal/app/machined/pkg/system/services/kubelet.go
smira: same here, it needs not EPHEMERAL, but rather some other volumes which I don't have yet

internal/app/machined/pkg/controllers/runtime/mount_status.go
smira: thanks, now fixed

frezbo (approved):

Issue #10332, opened 2025-02-11
disk images and UEFI/BIOS mode


smira:
## Low-level

In the `imager` profile, support configuration for the bootloader in the disk images - GRUB or systemd-boot.

For Talos < 1.10 images, use GRUB for !SecureBoot images (keep same defaults).

## arm64

There is no BIOS boot on arm64, so we can make disk images to be sd-boot always on arm64.

> Neet to verify with SBCs, but should be good for bare-metal arm64.

### Configurable boot loader in the Image Factory

Seems ugly, but we could in theory have `/metal.raw.zst` (GRUB, dual-boot) and `/metal-uefi.raw.zst` (systemd-boot, UEFI only).

The only downside is that it looks ugly.

### Dual-boot disk iamge.

For Talos 1.10+, we could create a dual boot GRUB/systemd-boot disk image:

* make both `BOOT` (xfs, GRUB) and `EFI` (systemd-boot, vfat) partitions big enough
* install GRUB in BIOS mode only
* install sd-boot to EFI partition

On the first boot, Talos should be able to detect such a dual-boot disk image, and wipe the "other" unused bootloader (i.e. if booted in UEFI mode, wipe GRUB). This would prevent a suprise fallback if the machine UEFI/BIOS setting is changed later, and Talos boots into the initially installed image.

PR #10380, opened 2025-02-14
feat: dual boot disk image

frezbo:
Generate disk image with both grub and sd-boot.

Fixes: https://github.com/siderolabs/talos/issues/10332


pkg/imager/profile/testdata/akamai-amd64-1.10.0.yaml
```diff
@@ -0,0 +1,26 @@
...
+  sdBoot:
+    path: /usr/install/amd64/systemd-boot.efi
+  baseInstaller:
+    imageRef: ghcr.io/siderolabs/installer:1.10.0
+output:
+  kind: image
+  imageOptions:
+    diskSize: 2356150272
+    diskFormat: raw
+    bootloader: dual-boot
```
dsseng, line 25: maybe this should be a quirk?

frezbo: these are test stubs, not the actual code

dsseng: ah okay, didn't see it's testdata

pkg/imager/profile/output.go
```diff
@@ -97,3 +105,55 @@ const (
...
+		if secureboot {
+			o.ImageOptions.Bootloader = DiskImageBootloaderSDBoot
+
+			return
+		}
+
+		useSDBoot := quirks.New(version).UseSDBootForUEFI()
+
+		// for arm64, we always use sd-boot
+		if arch == "arm64" && useSDBoot {
```
dsseng, line 143: I'm almost certain it should be
```suggestion
		if arch == "arm64" || useSDBoot {
```

pkg/imager/profile/profile_test.go
```diff
@@ -0,0 +1,63 @@
...
+				p.Output.FillDefaults(arch, version, secureBoot)
+				p.Customization.FillDefaults(p.Output.Kind, arch, version, secureBoot)
+
+				require.NoError(t, p.Validate())
+
+				var profileData strings.Builder
+
+				require.NoError(t, p.Dump(&profileData))
+
+				expectedData, err := os.ReadFile("testdata/" + profile + "-" + arch + "-" + version + ".yaml")
```
dsseng, line 56: Did you generate those with some scripts? Perhaps adding those should be useful for when we add more providers or other targets

frezbo: yeh, I initially started adding them manually until I saw we have too many profiles :sweat_smile: , I'll keep the code commented out to generate

pkg/imager/profile/output.go
frezbo: actually this is correct, we only want sd-boot for arm64, for 1.10 amd64 the default is dual-boot

internal/app/machined/pkg/runtime/v1alpha1/bootloader/dual/dual.go
```diff
@@ -0,0 +1,201 @@
...
+func (c *Config) Install(opts options.InstallOptions) (*options.InstallResult, error) {
+	if !opts.ImageMode {
+		return nil, fmt.Errorf("dual-boot bootloader is only supported in image mode")
+	}
+
+	var installResult *options.InstallResult
+
+	err := mount.PartitionOp(
+		opts.BootDisk,
+		[]mount.Spec{
```
smira, line 49: dumb question - can we use `Install` methods of both GRUB & SdBoot bootloaders here instead of duplicating the code?

frezbo: i could, but it does a little more things, so was trying to avoid it, also grub install is so many flags to pass/not pass based on UEFI

pkg/imager/profile/customization.go
```diff
@@ -0,0 +1,37 @@
...
+	if secureboot || arch == "arm64" {
+		return
+	}
+
+	if outKind == OutKindImage && quirks.New(version).UseSDBootForUEFI() {
+		if c.MetaContents == nil {
+			c.MetaContents = meta.Values{}
+		}
+
+		c.MetaContents = append(c.MetaContents, meta.Value{
```
smira, line 32: I wonder if Image Factory might override this value if schematic might have some meta customization.

I'm torn on whether it should be part of the profile

smira: I mean whether we should just silently write this value in the installer code when the install mode is disk image, and we are using a dual bootloader

frezbo: > I wonder if Image Factory might override this value if schematic might have some meta customization.

isn't this what we want, allow a user to change default to `sd-boot` maybe?

frezbo: > I mean whether we should just silently write this value in the installer code when the install mode is disk image, and we are using a dual bootloader

hmm, this is actually not bad and can avoid profile mods

frezbo: fixed

internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
```diff
@@ -67,14 +74,89 @@ func NewAuto() Bootloader {
 	return grub.NewConfig()
 }
 
-// New returns a new bootloader based on the secureboot flag.
-func New(secureboot bool, talosVersion string) Bootloader {
-	if secureboot {
-		return sdboot.New()
+// New returns a new bootloader based on the secureboot flag and architecture.
+func New(bootloader, talosVersion, arch string) (Bootloader, error) {
+	if arch == "arm64" {
```
smira, line 79: This should still honor the bootloader option? Probably instead of dual it should do sdboot, but it should do grub still if explicitly configured?

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/bootloader/bootloader.go
frezbo: not sure I get it? this honors the option, the profile fills the default to be `dual-boot` but users can change to any other and this still works

smira: I mean here on arm64 it always uses sd-boot, while it should honor setting, and use grub/sd-boot?

PR #10385, opened 2025-02-15
Docs update resetting-a-machine.md to include example of reset cmd

ihelmer07:
## What 
Slight formatting change and provided example of `talosctl reset`.

## Why
Consistency of showing example and made `reset` the focus of the section.


smira (approved):
thank you

PR #10418, opened 2025-02-25
fix: multiple fixes for dashboard/no data

### LINKED ISSUE #10383: Strange values during node boot in dashboard + error in issue template
## Bug Report
Check uptime, please
Image

### Description
Node just rebooted. I got strange uptime
### Environment

```
Client:
        Tag:         v1.9.3
        SHA:         d40df438
        Built:       2025-01-28T18:43:18Z
        Go version:  go1.23.5
        OS/Arch:     linux/amd64
Server:
        NODE:        xxx.xxx.xxx.xxx
        Tag:         v1.9.3
        SHA:         d40df438
        Built:
        Go version:  go1.23.5
        OS/Arch:     linux/amd64
        Enabled:     RBAC

```

I can't report as written in issue template because my kubectl refuses to execute 
`kubectl version --short`

```
$ kubectl version --short
error: unknown flag: --short
See 'kubectl version --help' for usage.

$ kubectl version
Client Version: v1.32.1
Kustomize Version: v5.5.0
Server Version: v1.32.1
```

Let me save you few seconds:
```
$ kubectl version --help
Print the client and server version information for the current context.

Examples:
  # Print the client and server versions for the current context
  kubectl version

Options:
    --client=false:
        If true, shows client version only (no server required).

    -o, --output='':
        One of 'yaml' or 'json'.

Usage:
  kubectl version [flags] [options]

Use "kubectl options" for a list of global command-line options (applies to all commands).
```

- Platform:
AMD, UEFI, no secure boot, bare metal server

smira:
There were multiple issues:

* `hostname` in the header is coming from resources, and the one in API data was unused
* we should update to `n/a` if the value is not available
* if the node is unavailable, the response contains a message with placeholders with a proxy error, and that should be honored to ignore the response; if instead we use it as it is, it puts zero value, which results in funny output.

Fixes #10383

### DIFF
--- a/internal/pkg/dashboard/apidata/source.go
+++ b/internal/pkg/dashboard/apidata/source.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
 	"github.com/siderolabs/talos/internal/pkg/dashboard/resolver"
+	"github.com/siderolabs/talos/pkg/machinery/api/machine"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 )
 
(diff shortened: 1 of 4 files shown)


internal/pkg/dashboard/apidata/source.go
```diff
@@ -85,7 +86,30 @@ func (source *Source) run(dataCh chan<- *Data) {
...
+	defer resultLock.Unlock()
+
+	for _, msg := range resp.GetMessages() {
+		node := source.node(msg)
+
+		if _, ok := nodes[node]; !ok {
+			nodes[node] = &Node{}
+		}
+
+		if msg.GetMetadata().GetError() != "" {
```
smira, line 104: this is the most critical fix, which is hard to see as I de-duplicated all the code

smira:
How it looks now during reboot of the machine:

[asciicast](https://asciinema.org/a/HYYIxroQ3enkixZSxqTWgkLSR)

utkuozdemir (approved):

Issue #10323, opened 2025-02-10
installer-base and installer changes


smira:
Goal: change the default `installer` to contain UKIs.

1. Stop pushing `installer` image in the `Dockerfile`, rename it to `installer-base`, and drop artifacts (should contain a single layer with just "tools").
2. In `imager` profile, default the installerBase image to `installer-base` for Talos >= 1.10 (TBD: also need to update Image Factory)
3. In the CI, in the `images-essential` step, generate UKIfied `installer` image using `imager`. Drop all hacks when we used SecureBoot installer instead of default installer.
4. In the CI, for the `push` step, generate and push `installer` image generated by `imager` (need to make sure it's multi-arch).

Issue #10188, opened 2025-01-21
drop dependency on `alpine` in `installer`/`imager`


smira:
Today we install the following packages from Alpine:

https://github.com/siderolabs/talos/blob/21cff3919b80f33f837b19728500fcb91e7caf8f/Makefile#L131-L149

We probably don't even need `bash` (needs to be verified though).

All other packages should be pulled from `pkgs` instead (build whatever pkg we're missing).

We can do step-by-step by first replacing the bits we already have, and looking into what is missing.

PR #10429, opened 2025-02-26
feat: unified installer

frezbo:
Generate `installer` from `imager` so installer always have UKI's.

Push `installer-base` which just contains base tools to install.

Fixes: #10188
Fixes: #10323


pkg/imager/post.go
```diff
@@ -26,7 +28,59 @@ func (i *Imager) postProcessTar(filename string, report *reporter.Reporter) (str
...
+
+	if err := cmd1.Start(); err != nil {
+		return "", err
+	}
+
+	if err = pipeW.Close(); err != nil {
+		return "", err
+	}
+
+	destination, err := os.OpenFile(outPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
```
smira, line 49: missing destination.Close()?

also probably s/O_APPEND/O_TRUNC/

frezbo: good catch

pkg/imager/out.go
```diff
@@ -417,6 +417,14 @@ func (i *Imager) outInstaller(ctx context.Context, path string, report *reporter
 	newInstallerImg := mutate.MediaType(empty.Image, types.OCIManifestSchema1)
 	newInstallerImg = mutate.ConfigMediaType(newInstallerImg, types.OCIConfigJSON)
 
+	newInstallerImg, err = mutate.ConfigFile(newInstallerImg, &v1.ConfigFile{
```
frezbo, line 420: this is needed if we want to synthesize a proper multiarch manifest. Our previous base image was lacking this info

PR #10436, opened 2025-02-27
fix: handle dynamic HTTP proxy settings for discovery client

### LINKED ISSUE #10136: Member discovery fails: discovery.talos.dev not reachable (1.9.1, Openstack)
Attempting to form a cluster consisting of 1 control plane and 1 worker node fails, because both machines cannot reach discovery.talos.dev for member discovery.

Both are running in Openstack in the same net with an external gateway configured for internet access. An addition, both nodes are configured to use a proxy.
 
Logs:
  `talosctl logs controller-runtime`
 
>   ERROR hello failed {"component": "controller-runtime", "controller": "cluster.DiscoveryServiceController", "error": "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 172.174.35.21:443: connect: no route to host\"", "endpoint": "discovery.talos.dev:443"}
>   ERROR hello failed {"component": "controller-runtime", "controller": "cluster.DiscoveryServiceController", "error": "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 172.174.35.21:443: i/o timeout\"", "endpoint": "discovery.talos.dev:443"}
>   ERROR hello failed {"component": "controller-runtime", "controller": "cluster.DiscoveryServiceController", "error": "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 172.174.35.21:443: i/o timeout\"", "endpoint": "discovery.talos.dev:443"}
>   ..

controlplane.yaml/worker.yaml:
```
machine:
	env:
    HTTP_PROXY: http://$IP
    HTTPS_PROXY: http://$IP
    http_proxy: http://$IP
    https_proxy: http://$IP

```
 
However, if I run the discovery service locally both machines can connect.
Could it be possible that the DiscoveryController ignores the proxy settings?
 
Environment
  Talos version: v1.9.1
  Kubernetes version: v1.32.0
  Platform: Openstack

smira:
Unfortunately, gRPC client library doesn't offer control on picking up HTTP proxy for a connection, and relies on sync.Once protected value which reads proxy variables from the environment just once.

As environment variables can be set dynamically, there is a chance that dial might happen before they are set, so they are never respected after that.

So I had to copy/adapt some part of grpc internal/transport parts to enable dynamic loading of HTTP proxy, and use new dialer in discovery client.

Fixes #10136

Also, while testing this with our small "air-gapped" tool I discovered that our HTTP proxy test implementation is not complete, so this had to be fixed as well.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/debug/air-gapped.go
+++ b/cmd/talosctl/cmd/mgmt/debug/air-gapped.go
@@ -194,14 +194,16 @@ func runHTTPServer(ctx context.Context, certPEM, keyPEM []byte) error {
 }
 
 func handleTunneling(w http.ResponseWriter, r *http.Request) {
-	dst, err := net.DialTimeout("tcp", r.Host, 10*time.Second)
+	addr := r.URL.Host
+
+	dstConn, err := net.DialTimeout("tcp", addr, 10*time.Second)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 
 		return
 	}
 
-	w.WriteHeader(http.StatusOK)
+	dst := dstConn.(*net.TCPConn)
 
 	hijacker, ok := w.(http.Hijacker)
 	if !ok {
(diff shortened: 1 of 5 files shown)


Unix4ever (approved):
I wonder if we could push that upstream 🤔

smira:
> I wonder if we could push that upstream 🤔

my experience with ALPN issue makes me think no... it's crazy hidden inside `internal/`, a set of private functions and variables...

PR #10463, opened 2025-03-04
fix(ci): fix integration tests

frezbo:
Fix CI integration tests.


.kres.yaml
```diff
@@ -1312,7 +1312,7 @@ spec:
           conditions:
             - only-on-schedule
         - name: build-race
-          command: initramfs installer-base
```
smira, line 1315: rekres?

smira (approved):

PR #10483, opened 2025-03-06
chore: Tidy labeled-squashfs.sh

jnohlgard:
Cleaned up the labeled-squashfs.sh script for readability.


hack/labeled-squashfs.sh
```diff
@@ -1,6 +1,22 @@
 #!/bin/bash
+set -euf
```
dsseng, line 2:
```suggestion
set -eufx
```

dsseng (approved):

hack/labeled-squashfs.sh
dsseng: (you don't actually have to apply this, I'll do myself when I merge)

dsseng:
Thanks for the improvements! Could you please pull (or reset to this PR's revision) and run `git commit --amend -s` to sign-off the commit? Our policy requires contributors to sign-off their commits to confirm they are allowed to publish this code (should be the case for the code you wrote yourself)

jnohlgard:
Right, will push an updated commit message including signed-off-by when I am back in front of the keyboard.

PR #10485, opened 2025-03-06
fix: ensure no goroutines escape in dns controller

DmitriyMV:
- Remove all reliance on finalizers.
- Add `Close` method to CoreDNS `Proxy` struct.
- Wait for `Runner.Serve` to complete.


go.mod
```diff
@@ -7,7 +7,7 @@ replace (
 	cloud.google.com/go => cloud.google.com/go v0.100.2
 
 	// forked coredns so we don't carry caddy and other stuff into the Talos
-	github.com/coredns/coredns => github.com/siderolabs/coredns v1.12.50
```
smira, line 10: are we going back in time here?

smira (approved):

PR #10502, opened 2025-03-10
fix: qemu: archive cluster logs only after stopping VMs

dsseng:
Otherwise the archive is corrupted due to QEMU still writing to the logs while they are being archived

ci: enable --with-json-logs for any e2e-qemu runs

### DIFF
--- a/hack/test/e2e-qemu.sh
+++ b/hack/test/e2e-qemu.sh
@@ -41,8 +41,10 @@ case "${WITH_VIRTUAL_IP:-false}" in
     ;;
 esac
 
-case "${WITH_JSON_LOGS:-false}" in
-  true)
+case "${WITH_JSON_LOGS:-true}" in
+  false)
+    ;;
+  *)
     QEMU_FLAGS+=("--with-json-logs")
     ;;
 esac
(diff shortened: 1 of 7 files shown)


hack/test/e2e-qemu.sh
```diff
@@ -275,6 +269,7 @@ function create_cluster {
     --cidr=172.20.1.0/24 \
     --install-image="${INSTALLER_IMAGE}" \
     --with-init-node=false \
+    --with-json-logs=true \
```
smira, line 272: we can do this, but I'm not super excited about it, as this changes the defaults for all runs, and makes our CI run as if logs are always enabled, so we don't test the case when the logs are not enabled. 

so I would leave at least a single job with disabled logs

smira (approved):

pkg/provision/providers/qemu/destroy.go
```diff
@@ -89,6 +83,12 @@ func (p *provisioner) Destroy(ctx context.Context, cluster provision.Cluster, op
 		return err
 	}
 
+	if options.SaveClusterLogsArchivePath != "" {
```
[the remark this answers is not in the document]
dsseng: it's right before removing logs. then I see only networks etc being removed, nothing that writes logfiles

hack/test/e2e-qemu.sh
dsseng: I do not know, but a job that never reboots should be the perfect candidate to do this, both testing without the option and not losing any warnings. Do you know such jobs?

smira: Any integration-misc might be fine, like a disk image one.

PR #10505, opened 2025-03-11
fix: correct indentation for 'name: ci' in controlplane and worker test data

fullstackjam:
# Pull Request

## What? (description)
correct indentation for 'name: ci' in controlplane and worker test data
## Why? (reasoning)
more user-friendly to expand
## Acceptance

- [x ] you linted your code (`make lint`)


smira:
I don't think it's a reasonable change, as it's test data

PR #10509, opened 2025-03-11
refactor: mount operation for STATE and user disks

smira:
Use new controller for user disk and STATE mounts, drop old code in the sequencer.

Also support mounts with parent (when e.g. `/var/lib` is mounted on top of `/var`).


Makefile
```diff
@@ -420,31 +420,31 @@ talosctl-image: ## Builds the talosctl container image and outputs it to the reg
 	@$(MAKE) registry-talosctl
 
 talosctl-all:
-	@$(MAKE) local-talosctl-all DEST=$(ARTIFACTS) PUSH=false NAME=Client
```
smira, line 423: I believe this was never actually used, but this causes very fun side-effects when running integration tests locally

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -291,7 +257,157 @@ func (ctrl *MountController) tearDownMountStatus(ctx context.Context, r controll
...
+	default:
+		return fmt.Errorf("unsupported volume type %q", volumeStatus.TypedSpec().Type)
+	}
+}
+
+func (ctrl *MountController) handleDirectoryMountOperation(
+	_ string,
+	_ *block.VolumeStatus,
+) error {
+	// [TODO]: implement me
```
smira, line 286: for the next PR, right now it's only used for Talos in docker

frezbo (approved):

PR #10517, opened 2025-03-12
fix: reconnect on SideroLink tunnel on/off change

smira:
The issue is not so easy to fix, as GRPC tunnel on/off change requires two different flow for the link (interface):

* no tunnel -> Talos link controller should create in-kernel `wireguard` link and no userspace components
* tunnel on -> Talos link controller should never create the link, and only adjust WG settings via UAPI, while the actual link is created by the userspace implementation (it's a `tun` device)

Transition between those two links is impossible for the link controller to distinguish, as it doesn't know that it has to drop old link and skip creating new one based on the information available.

So, instead, use different names for the link in two states: `siderolink` for the kernel flow, and `siderolinktun` for the userspace flow. This fixes the issue of proper link cleanup/re-creation.

Add integration tests.


Unix4ever (approved):

smira:
Our tests are broken:

```
=== RUN   TestIntegration/api.SideroLinkSuite/TestTunnelSettingFlip
    siderolink.go:55: testing SideroLink on node 172.20.1.6
    siderolink.go:64: skipping the test since SideroLink is not enabled
```

I guess the issue is that UKI doesn't inherit siderolink args. I should rework this to use config patches

smira:
> Our tests are broken:

should be fixed now, reworked `talosctl cluster create` to use config patches instead

smira:
> should be fixed now, reworked talosctl cluster create to use config patches instead

I guess not fixed, need to do both kernel args and config 🤦‍♂️ 

Otherwise it won't work with apply-config..

utkuozdemir (approved):

Issue #10339, opened 2025-02-11
.machine.install.extraKernelArgs with UKIs


smira:
As it becomes a no-op with UKIs, it'd be nice to:

1. Document it in the reference & release notes.
2. Add a runtime validation - if we are booted via an UKI (or using sd-boot), add a warning that non-empty extra kernel args have no effect.

PR #10519, opened 2025-03-13
chore: extraKernelArgs validation for UKI's

frezbo:
Add validation for `.machine.install.extraKernelArgs`.

Fixes: #10339


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -812,6 +812,7 @@ type InstallConfig struct {
 	//     Allows for supplying extra kernel args via the bootloader.
 	//     Existing kernel args can be removed by prefixing the argument with a `-`.
 	//     For example `-console` removes all `console=<value>` arguments, whereas `-console=tty0` removes the `console=tty0` default argument.
+	//     If booted using UEFI and using UKI's this will be ignored.
```
smira, line 815:
```suggestion
	//     If Talos is using systemd-boot as a bootloader (default for UEFI) this setting will be ignored.
```

smira (approved):

smira:
I think we should also update the hack/release.toml and docs.

We should say in clear that:

* if upgrading to 1.10, nothing will change for you
* if installing 1.10 from scratch on UEFI system, systemd-boot will be used, and install.extraKernelArgs is now no-op

We should also say that Talos no longer preserves the args as well (in systemd-boot), like siderolink.api on upgrade/install, so the proper args should be in the installer.

frezbo:
> I think we should also update the hack/release.toml and docs.
> 
> We should say in clear that:
> 
>     * if upgrading to 1.10, nothing will change for you
> 
>     * if installing 1.10 from scratch on UEFI system, systemd-boot will be used, and install.extraKernelArgs is now no-op
> 
> 
> We should also say that Talos no longer preserves the args as well (in systemd-boot), like siderolink.api on upgrade/install, so the proper args should be in the installer.

ahh I lost my release toml changes :joy:

PR #10520, opened 2025-03-13
test: allocate more resources for conformance runs

smira:
There was a wrong variable, but also always use more resources.


frezbo (approved):

dsseng (approved):

smira:
Added a containerd patch via https://github.com/siderolabs/pkgs/pull/1185

PR #10523, opened 2025-03-13
feat: deprecate `.machine.install.extensions`

frezbo:
Fully deprecate `.machine.install.extensions`.


frezbo:
> it feels like we're removing support instead of deprecating it?

yeh removing it

> We could keep this PR till 1.11, and if we do remove, we can cut some `installer` dependencies, like cpio and friends which were used to recompress initramfs, in fact we should probably need almost no dependencies at all.

I already dropped those in the installer PR

smira:
hm, I still see all of that:

```
FROM scratch AS installer-base-image
ARG TARGETARCH
ENV TARGETARCH=${TARGETARCH}
COPY --link --from=pkg-fhs / /
COPY --link --from=pkg-ca-certificates / /
COPY --link --from=pkg-musl / /

COPY --link --from=pkg-cpio / /
COPY --link --from=pkg-dosfstools / /
COPY --link --from=pkg-grub / /
COPY --link --from=pkg-grub-amd64 /usr/lib/grub /usr/lib/grub
COPY --link --from=pkg-grub-arm64 /usr/lib/grub /usr/lib/grub
COPY --link --from=pkg-kmod / /
COPY --link --from=pkg-libattr / /
COPY --link --from=pkg-libinih / /
COPY --link --from=pkg-liblzma / /
COPY --link --from=pkg-liburcu / /
COPY --link --from=pkg-openssl / /
COPY --link --from=pkg-xfsprogs / /
COPY --link --from=installer-image-gen /rootfs /
```

I think we should deprecate fully before removing.

And right now we should warn like you did if the config is being used, in 1.11 we make it hard failure if install.extensions is not empty

frezbo:
> hm, I still see all of that:
> 
> ```
> FROM scratch AS installer-base-image
> ARG TARGETARCH
> ENV TARGETARCH=${TARGETARCH}
> COPY --link --from=pkg-fhs / /
> COPY --link --from=pkg-ca-certificates / /
> COPY --link --from=pkg-musl / /
> 
> COPY --link --from=pkg-cpio / /
> COPY --link --from=pkg-dosfstools / /
> COPY --link --from=pkg-grub / /
> COPY --link --from=pkg-grub-amd64 /usr/lib/grub /usr/lib/grub
> COPY --link --from=pkg-grub-arm64 /usr/lib/grub /usr/lib/grub
> COPY --link --from=pkg-kmod / /
> COPY --link --from=pkg-libattr / /
> COPY --link --from=pkg-libinih / /
> COPY --link --from=pkg-liblzma / /
> COPY --link --from=pkg-liburcu / /
> COPY --link --from=pkg-openssl / /
> COPY --link --from=pkg-xfsprogs / /
> COPY --link --from=installer-image-gen /rootfs /
> ```
> 
> I think we should deprecate fully before removing.
> 
> And right now we should warn like you did if the config is being used, in 1.11 we make it hard failure if install.extensions is not empty

those are mostly for grub, now, squashfstools is only in imager, so it already got dropped from installer

frezbo:
I was thinking in terms of we've had a deprecation for extensions for a while, i think last two releases, so makes sense to only keep the machine config field as no-op

smira:
> those are mostly for grub, now, squashfstools is only in imager, so it already got dropped from installer

we definitely don't need `cpio` in this case

smira:
> I was thinking in terms of we've had a deprecation for extensions for a while, i think last two releases, so makes sense to only keep the machine config field as no-op

ok, I'm convinced now. let's drop and do a scream test. this would allow us to implement a proper installer API in the next release, with copying assets from an ISO or something like that. probably even installing from an Image Factory directly.

frezbo:
> > those are mostly for grub, now, squashfstools is only in imager, so it already got dropped from installer
> 
> we definitely don't need `cpio` in this case

something was using it, let me go back and test again

frezbo:
> > > those are mostly for grub, now, squashfstools is only in imager, so it already got dropped from installer
> > 
> > 
> > we definitely don't need `cpio` in this case
> 
> something was using it, let me go back and test again

yup not needed, fixed it

smira (approved):

PR #10527, opened 2025-03-14
fix: kexec when using sd-boot

frezbo:
https://github.com/siderolabs/talos/pull/10443 broke kexec and `callback` was never being called.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
@@ -65,11 +65,28 @@ func New() *Config {
 }
 
 // ProbeWithCallback probes the sd-boot bootloader, and calls the callback function with the Config.
+//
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -65,12 +65,29 @@ func New() *Config {
...
 		return nil, nil
 	}
 
+	// here we need to read the EFI vars to see if we have any defaults
+	// and populate config accordingly
+	// https://www.freedesktop.org/software/systemd/man/systemd-boot.html#LoaderEntryDefault
+	// this should be set on install/upgrades
+	efiCtx := efivario.NewDefaultContext()
+
+	bootedEntry, err := ReadVariable(efiCtx, LoaderEntrySelectedName)
```
smira, line 82: would it still work if I we boot off initramfs/kernel directly?

smira (approved):

PR #10532, opened 2025-03-14
docs: fix typo cluser -> cluster

ojsef39:
# Pull Request

## What? (description)

I'm not sure if it's intentional to show it as "cluser" jn the example but just in case this is a typo i opened this PR :)

## Acceptance

If its meant as an example maybe it should be changed to example.local? or corrected to cluster.local


smira (approved):
thank you!

ojsef39:
@smira 
Thanks for merging this small change so fast and sorry i missed the other references, ill be more careful on my next PR :)

smira:
> @smira Thanks for merging this small change so fast and sorry i missed the other references, ill be more careful on my next PR :)

It's not your fault, this is actually generated reference, and that is not obvious.

ojsef39:
> It's not your fault, this is actually generated reference, and that is not obvious.

Ahh, okay, thanks for clarifying :)

Issue #7184, opened 2023-05-05
Support for routing rules and tables


dedene:
## Feature Request

Support for routing rules and tables in the machine config

### Description

For various setups on Bare Metal, more advanced routing configuration possibilities are needed. This would allow for things like 

```
ip route add default via 46.105.135.110 dev eth0 table 100
ip rule add from 46.105.135.96/28 table 100
```

See [this discussion](https://github.com/siderolabs/talos/discussions/7177) for an example use case.

smira:
Some notes from implementation point of view:

* routes already support tables, but they are not exposed in the machine config now (easy change)
* routing rules are not exposed as resources yet

ffetech:
Would be great if also something like this would be possible:
`iptables -t nat -A POSTROUTING -s '10.0.0.0/16' -o eth0 -j MASQUERADE`
… to get a talos node to work as NAT gateway.

smira:
> Would be great if also something like this would be possible: `iptables -t nat -A POSTROUTING -s '10.0.0.0/16' -o eth0 -j MASQUERADE` … to get a talos node to work as NAT gateway.

This makes sense, but it is totally different from this issue. This should go to #4421

camrossi:
I need to set a "blackhole" route in my cluster, in general ip route support doing this:
```
ip route add blackhole 1.1.1.1/32
```
So as part of this feature request we could also add `route-type`  support.
For now I did this workaround in my cluster but would like to understand if this is ok or if I am gonna end up with issues down the road:
```
    network:
        interfaces:
            - interface: eth0 # The interface name.
              routes:
                - network: 10.56.0.0/16 
                  gateway: 0.0.0.0 
              dhcp: true

talosctl -n 192.168.11.90  get route | grep 10.56.0.
NODE            NAMESPACE   TYPE          ID                                        VERSION   DESTINATION   GATEWAY  LINK
192.168.11.90   network     RouteStatus   inet4//10.56.0.0/16/1024  1                10.56.0.0/16                         eth0 
```

And it seems to have done the trick but I am not really sure if this is something that is actually supported or not.

kvaps:
Hey there, I just want to note that I'm also looking the way for doing this, I can sponsor on writing implementation.

my issue: https://github.com/siderolabs/talos/issues/8738

maxpain:
It would be useful for Hetzner vSwitch and OVH vRack.

danieljkemp:
Another use case:

[...]

PR #10548, opened 2025-03-17
feat: support specifying route tables

tobiaskohlbau:
# Pull Request

## What? (description)

This is a stab at #7184 to support routing tables within machine configuration or route specs.

## Why? (reasoning)

There are instances where multiple routes need to be specified for e.g. multiple default gateways for example in a vlan situation. In such cases it's helpful if it's possible to specify the routing table where a defined route should be applied.

## Acceptance


pkg/machinery/nethelpers/routingtable.go
```diff
@@ -4,9 +4,26 @@
 
 package nethelpers
 
+import "strconv"
+
 // RoutingTable is a routing table ID.
 type RoutingTable uint32
 
+func (rt RoutingTable) String() string {
```
smira, line 12: these are certainly wrong changes, we don't need any of that

pkg/machinery/resources/network/route_spec_test.go
```diff
@@ -42,7 +42,7 @@ dst: 192.168.3.4/25
 src: 1.1.1.1
 gateway: 2.2.2.2
 outLinkName: eth0
-table: local
```
smira, line 45: these changes are wrong as well

smira (commented):
I don't think it's valuable without support for routing rules. 

We have this on the radar with new network machine configuration.

tobiaskohlbau:
I would like to get some feedback and communication about this feature and how solving it should be approached. It's most likely a breaking change as the serialization is changed from string to integer values within the definitions. I'm not sure how this should be approached.

pkg/machinery/resources/network/route_spec_test.go
tobiaskohlbau: How would you like to approach the situation, routing tables are integers not strings.  Should the definition treat this as strings and convert it internally as integers?

tobiaskohlbau:
> I don't think it's valuable without support for routing rules.
> 
> We have this on the radar with new network machine configuration.

Thanks for your input. So you would like to ditch a possible implementation without routing rules in favor of the machine configuration added in the feature?

pkg/machinery/resources/network/route_spec_test.go
smira: just keep things as they are, and if the change is made, it should be backwards compatible

tobiaskohlbau:
> How would you use route specs without routing rules? Any example?

Im quite sure I'm not following along. That's why I asked what exactly is meant by routing rules, exposing routing definitions within the network section of a machine configuration independent of the individual interface definition?

I went down the route of implementing your suggestion of keeping backward compatibility. Always the best option for the user. Not exactly sure why I went done the other path.

smira:
Routing rules in Linux is https://man7.org/linux/man-pages/man8/ip-rule.8.html

internal/app/machined/pkg/controllers/network/route_config.go
```diff
@@ -279,7 +280,26 @@ func (ctrl *RouteConfigController) processDevicesConfiguration(logger *zap.Logge
...
+				route.Table = nethelpers.TableMain
+			case "local":
+				route.Table = nethelpers.TableLocal
+			default:
+				tableId, err := strconv.ParseUint(routeTable, 10, 32)
+				if err != nil {
+					return route, fmt.Errorf("error parsing route table id: %w", err)
+				}
+				route.Table = nethelpers.RoutingTable(uint32(tableId))
+			}
```
tobiaskohlbau, line 300: @smira would this be an option for now until the machine configuration is done?

Would allow specifying routing tables either by name for the default ones or by stringified id in other cases.

tobiaskohlbau:
> Routing rules in Linux is https://man7.org/linux/man-pages/man8/ip-rule.8.html

Thanks, I was not aware of the nuances. Always was under the impression the routing table approach is the same as the rule approach within the kernel (Thought the rule inserts into the tables directly). TIL that these are handled differently by the kernel.

internal/app/machined/pkg/controllers/network/route_config.go
tobiaskohlbau: Sure let's do it. What's the approach discussing in the issue? This was only a first playground. Never the idea of pushing it through.

smira:
My point is that we know that Talos lacks this feature, but we need to implement both routing rules and table support.

Just one without another imho doesn't provide any value.

PR #10559, opened 2025-03-18
fix: pass /usr/etc/in-container to apid, trustd and extension containers

dsseng:
Fixes SELinux denials due to apid trying to relabel its socket while in a container (observed on Talos, likely to be an issue on any SELinux host)


internal/app/machined/pkg/system/services/apid.go
```diff
@@ -170,6 +170,7 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 	// Set the mounts.
 	mounts := []specs.Mount{
 		{Type: "bind", Destination: "/etc/ssl", Source: "/etc/ssl", Options: []string{"bind", "ro"}},
+		{Type: "bind", Destination: "/usr/etc", Source: "/usr/etc", Options: []string{"bind", "ro"}},
```
smira, line 173: I would rather bind mount a specific path (could be conditional) in order not to expose anything else in `/usr/etc` accidentally

smira (approved):

PR #10566, opened 2025-03-19
fix: kexec with smbios type 11 string

frezbo:
sd-boot supports passing in extra kernel cmdline via type 11 smbios string. Talos should do the same when kexec'ing.


pkg/provision/providers/qemu/launch.go
```diff
@@ -323,7 +323,6 @@ func launchVM(config *LaunchConfig) error {
 		"-no-reboot",
 		"-boot", fmt.Sprintf("order=%s,reboot-timeout=5000", bootOrder),
 		"-smbios", fmt.Sprintf("type=1,uuid=%s", config.NodeUUID),
-		"-smbios", fmt.Sprintf("type=11,value=io.systemd.stub.kernel-cmdline-extra=%s", config.sdStubExtraCmdline),
```
frezbo, line 326: this was broken already, since we append the `sdStubExtraCmdlineConfig` later down below only, so first boot always missed `talos.config`

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -216,6 +218,29 @@ func (c *Config) KexecLoad(r runtime.Runtime, disk string) error {
 			return fmt.Errorf("failed to read cmdline from uki: %w", err)
 		}
 
+		smbiosInfo, err := smbiosinternal.GetSMBIOSInfo()
```
smira, line 221: I know we don't support kexec today with SecureBoot, but I think it'd be nice still to add here `if !SecureBoot`

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -216,6 +218,29 @@ func (c *Config) KexecLoad(r runtime.Runtime, disk string) error {
...
+		if err == nil {
+			for _, structure := range smbiosInfo.Structures {
+				if structure.Header.Type != 11 {
+					continue
+				}
+
+				if structure.Strings != nil {
```
smira, line 228: this should be shorter:

```suggestion
        const kernelCmdlineExtra = "io.systemd.stub.kernel-cmdline-extra="
	for _, s := range structure.Strings {
            if strings.HasPrefix(s, kernelCmdlineExtra) {
                cmdline.WriteString(" " + s[len(kernelCmdlineExtra):])
            }
        }
```

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: this doesn't get called at all if secureboot is enabled, it shortcuts way before in sequencer

PR #10567, opened 2025-03-19
fix: shrink installer and imager images

smira:
Remove some files we don't need, remove some tools, move tools around.

E.g. we don't need grub foreign architecture for the installer, as it only runs for a specific arch.

Move kmod to imager.

Drop `ctr` from containerd, as it was added to pkgs.


smira:
`installer-base` now for amd64: ~24 MiB compressed, 75 MiB uncompressed

frezbo (approved):

Issue #10577, opened 2025-03-20
kexec without ever running sd-boot fails


smira:
1. Boot from initramfs/kernel with bootloader enabled.
2. First reboot goes fine via kexec.
3. Attempt to upgrade fails to kexec, as now we have two UKIs, but booted entry is not set.

We can write the booted entry EFI variable (if it's missing) on the first kexec when we boot the only UKI.

Skip kexec if the booted entry is empty.

Verify with:

* upgrades with kexec enabled
* upgrades with kexec disabled 

.. both should boot correct Talos version after an upgrade.

PR #10581, opened 2025-03-21
fix: upgrades with kexec

frezbo:
Fix by setting the default booted entry as like sd-boot.

Fixes: #10577

### DIFF
--- a/cmd/installer/pkg/install/errata.go
+++ b/cmd/installer/pkg/install/errata.go
@@ -5,71 +5,21 @@
 package install
 
 import (
-	"bytes"
-	"compress/gzip"
 	"context"
 	"fmt"
-	"io"
 	"log"
 	"os"
-	"runtime"
 	"time"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 
-	pkgkernel "github.com/siderolabs/talos/pkg/kernel"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/compatibility"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
-	"github.com/siderolabs/talos/pkg/machinery/kernel"
 	"github.com/siderolabs/talos/pkg/machinery/role"
 )
-
-// errataArm64ZBoot handles the case when upgrading from Talos < 1.8.0 on arm64 to compressed kernel.
-func errataArm64ZBoot() {
-	if runtime.GOARCH != "arm64" {
-		return
-	}
-
-	oldConfig, err := os.OpenFile("/proc/config.gz", os.O_RDONLY, 0)
-	if err != nil {
-		log.Printf("failed to open /proc/config.gz: %s", err)
-
-		return
-	}
-
-	defer oldConfig.Close() //nolint:errcheck
-
-	r, err := gzip.NewReader(oldConfig)
-	if err != nil {
-		log.Printf("failed to read /proc/config.gz: %s", err)
-
-		return
-	}
-
-	contents, err := io.ReadAll(r)
-	if err != nil {
-		log.Printf("failed to read /proc/config.gz: %s", err)
-
-		return
-	}
-
-	if bytes.Contains(contents, []byte("CONFIG_ARM64_ZBOOT=y")) {
-		// nothing to do
-		return
-	}
-
-	log.Printf("disabling kexec due to upgrade to the compressed kernel")
-
-	if err = pkgkernel.WriteParam(&kernel.Param{
-		Key:   "proc.sys.kernel.kexec_load_disabled",
-		Value: "1",
-	}); err != nil {
-		log.Printf("failed to disable kexec: %s", err)
-	}
-}
 
 // errataNetIfnames appends the `net.ifnames=0` kernel parameter to the kernel command line if upgrading
 // from an old enough version of Talos.
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
@@ -77,16 +77,18 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
 
 	// here we need to read the EFI vars to see if we have any defaults
 	// and populate config accordingly
-	// https://www.freedesktop.org/software/systemd/man/systemd-boot.html#LoaderEntryDefault
-	// this should be set on install/upgrades
+	// https://www.freedesktop.org/software/systemd/man/latest/systemd-boot.html#LoaderEntryDefault
+	// this is set by systemd-boot.
 	efiCtx := efivario.NewDefaultContext()
 
-	bootedEntry, err := ReadVariable(efiCtx, LoaderEntrySelectedName)
+	// first we start by checking if we have a Default entry
+	// this is set by installer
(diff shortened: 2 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -77,16 +77,16 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
...
+	// https://www.freedesktop.org/software/systemd/man/latest/systemd-boot.html#LoaderEntrySelected
+	// this is set by systemd-boot.
 	efiCtx := efivario.NewDefaultContext()
 
 	bootedEntry, err := ReadVariable(efiCtx, LoaderEntrySelectedName)
 	if err != nil {
 		return nil, err
 	}
 
-	config := &Config{}
```
frezbo, line 89: this was really dumb mistake when implementing kexec, always returned a non-nil even if probe was unsuccessful

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -116,20 +116,22 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
 				return err
 			}
 
+			// here we handle a case when we boot of just kernel+initrd/uki and we don't have a booted entry
+			// then we know we're booted of UKI
+			if bootedEntry == "" && len(ukiFiles) == 1 {
+				sdbootConf = &Config{}
```
smira, line 122: I'm a bit confused here, should we just set `bootedEntry` to `filepath.Base(ukiFiles[0])`, and the loop below will set the config's Default?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -163,9 +165,17 @@ func Probe(disk string, options options.ProbeOptions) (*Config, error) {
 //nolint:gocyclo
 func (c *Config) KexecLoad(r runtime.Runtime, disk string) error {
 	_, err := ProbeWithCallback(disk, options.ProbeOptions{}, func(conf *Config) error {
+		efiCtx := efivario.NewDefaultContext()
+
+		// we read the next entry to kexec as set by installer
```
smira, line 170: also confused here on the logic. 

when systemd-boot works, which entry does it kexec to?

Default or Selected?

As I don't think `KexecLoad` should set the EFI variables, but rather installing the bootloader should set the entry?

cmd/installer/pkg/install/errata.go
```diff
@@ -61,6 +61,11 @@ func errataArm64ZBoot() {
 		return
 	}
 
+	if bytes.Contains(contents, []byte("CONFIG_EFI_ZBOOT=y")) {
```
frezbo, line 64: kernel config changed upstream, so now we check both `CONFIG_EFI_ZBOOT` and `CONFIG_ARM64_ZBOOT`

smira: nice catch!

and I think we can probably remove this altogether if we don't support upgrading from a version that first introduced ZBOOT?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -77,16 +77,27 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
...
+	// this is set by systemd-boot when we select an entry
 	bootedEntry, err := ReadVariable(efiCtx, LoaderEntrySelectedName)
 	if err != nil {
 		return nil, err
 	}
 
-	config := &Config{}
+	if bootedEntry == "" {
+		// if we don't have a selected entry, we check the default entry
+		// this is set by systemd-boot when we boot
```
smira, line 93: is it _set_ by us? (or should be?) or whoever installs UKIs?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -235,6 +250,13 @@ func (c *Config) KexecLoad(r runtime.Runtime, disk string) error {
 			}
 		}
 
+		efiCtx := efivario.NewDefaultContext()
+
+		// clear the selected entry
+		if err := WriteVariable(efiCtx, LoaderEntrySelectedName, ""); err != nil {
```
smira, line 256: why do we clear selected entry in KexecLoad? should we clear this on bootloader install?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: yeh install code sets this

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: ah good point

frezbo: that doesn't seem to work when booting off iso, since sd-boot already set that and we can't clear it, so doing at kexec makes sense, since we only need to be cleared for kexec

frezbo:
```text
updating EFI variables
[    3.975603] Error: failed to install bootloader: efivario/set: write /sys/firmware/efi/efivars/LoaderEntrySelected-4A67B082-0A4C-41CF-B6C7-440B29BB8C4F: invalid argument

```

frezbo: hmmm weird this only fails when booting off iso

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -77,16 +77,18 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
...
-	// https://www.freedesktop.org/software/systemd/man/systemd-boot.html#LoaderEntryDefault
-	// this should be set on install/upgrades
+	// https://www.freedesktop.org/software/systemd/man/latest/systemd-boot.html#LoaderEntryDefault
+	// this is set by systemd-boot.
 	efiCtx := efivario.NewDefaultContext()
 
-	bootedEntry, err := ReadVariable(efiCtx, LoaderEntrySelectedName)
+	// first we start by checking if we have a Default entry
+	// this is set by installer
+	bootedEntry, err := ReadVariable(efiCtx, LoaderEntryDefaultName)
```
frezbo, line 86: we don't need to rely on the selected entry anymore since the condition below would pick the default uki if there's only one, allowing booting from iso, uki and initrd/kernel combo

cmd/installer/pkg/install/errata.go
smira: double checked, this errata is only for Talos < 1.8.0, while minimum host to upgrade to 1.10.0 is Talos 1.8.0, so we can drop this whole errata function altogether

smira (approved):

cmd/installer/pkg/install/errata.go
frezbo: dropped :+1:

PR #10599, opened 2025-03-26
Update hetzner.md

gecube:
fix indentation - as it contradicts the same in another place of doc.


smira (approved):
thank you!

smira:
Note: build fails because of https://github.com/golang/vulndb/issues/3578, but this should clear soon

Issue #9955, opened 2024-12-13
lock xfs filesystem creation options to match Talos version


smira:
See e.g. #9913 

As the Image Factory runs, it needs to create disk images for older Talos releases, while having the most recent `xfsprogs` in its hands.

The idea is provide a quirk which maps Talos version to xfsprogs config file which is based on Linux LTS version. To keep downgrades possible, we could say that the first Talos release which ships the next Linux LTS is using previous LTS config, while the next one jumps to the new LTS.

E.g. Talos 1.9 ships Linux 6.12, but uses 6.6 xfs settings, while Talos 1.10 shifts to 6.12 xfs settings.

PR #10623, opened 2025-04-01
feat: support xfs mkfs config version

frezbo:
Use xfs mkfs config versions based on Talos versions.

Fixes: #9955

### DIFF
--- a/hack/cleanup.sh
+++ b/hack/cleanup.sh
@@ -27,7 +27,7 @@ find ${PREFIX}/usr/bin -type f -exec strip --strip-all {} ';' || true
 # Remove header files, man files, and any other non-runtime dependencies.
 rm -rf ${PREFIX}/usr/lib/pkgconfig/ \
        ${PREFIX}/{include,usr/include}/* \
-       ${PREFIX}/{share,usr/share}/* \
+       ${PREFIX}/share/* \
        ${PREFIX}/usr/lib/cmake \
        ${PREFIX}/usr/lib/gconv/ \
        ${PREFIX}/usr/libexec/getconf \
--- a/pkg/machinery/imager/quirks/quirks.go
+++ b/pkg/machinery/imager/quirks/quirks.go
@@ -222,3 +222,22 @@ func (q Quirks) SupportsUnifiedInstaller() bool {
 
 	return q.v.GTE(minTalosVersionUnifiedInstaller)
 }
+
+// XFSMkfsConfig returns the mkfs.xfs config for the given Talos version.
+func (q Quirks) XFSMkfsConfig() string {
+	switch version := q.v; {
+	// if the version doesn't parse, we assume it's latest Talos
+	// update when we have a new LTS config
+	case version == nil:
+		return "/usr/share/xfsprogs/mkfs/lts_6.12.conf"
+	// add new version once we have a new LTS config
+	case version.GTE(semver.MustParse("1.10.0")):
+		return "/usr/share/xfsprogs/mkfs/lts_6.12.conf"
+	case version.GTE(semver.MustParse("1.8.0")) && version.LT(semver.MustParse("1.10.0")):
+		return "/usr/share/xfsprogs/mkfs/lts_6.6.conf"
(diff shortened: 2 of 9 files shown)


pkg/machinery/imager/quirks/quirks.go
```diff
@@ -222,3 +222,20 @@ func (q Quirks) SupportsUnifiedInstaller() bool {
...
+// XFSMkfsConfig returns the mkfs.xfs config for the given Talos version.
+func (q Quirks) XFSMkfsConfig() string {
+	switch version := q.v; {
+	// if the version doesn't parse, we assume it's latest Talos
+	case version == nil:
+		return "/usr/share/xfsprogs/mkfs/lts_6.12.conf"
+	case version.GTE(semver.MustParse("1.10.0")):
+		return "/usr/share/xfsprogs/mkfs/lts_6.12.conf"
+	case version.GTE(semver.MustParse("1.9.0")):
+		return "/usr/share/xfsprogs/mkfs/lts_6.6.conf"
```
smira, line 235: I'm confused a bit on the LTS versions here, 1.9 & 1.10 are both LInux 6.12.x based, and 1.6-1.8 should be 6.6.x ?

frezbo: yeh, but isn't the plan to use 6.6x settings for 1.9.x?

smira: probably yes, makes sense, but we'd need to document it here so it's clear what the rules are

frezbo: should we add a quirk to only enable this starting from 1.10, maybe older xfsprogs doesn't have mkfs config files shipped :thinking:

frezbo:
> volumemanager doesn't use the config file and uses the xfs defaults

also should we need to do this for volumes created with volumemanager, then we just need to get the current running version in the controller

pkg/machinery/imager/quirks/quirks.go
smira: > should we add a quirk to only enable this starting from 1.10, maybe older xfsprogs doesn't have mkfs config files shipped 🤔

hold on, this code will run in two contexts:

* in Talos itself, it will run for 1.10+ only
* in the Image Factory, it will run for any version of Talos, and this is where it becomes actually important, and we provide proper defaults.

smira: > actually scratch that, downgrades shouldn't re-create STATE/EPHEMERAL so I guess defualts is good

my point they should be able to read STATE/EPHEMERAL, but they should be created on older version though.

smira:
> also should we need to do this for volumes created with volumemanager, then we just need to get the current running version in the controller

we can use `nil` (latest version) there

pkg/machinery/imager/quirks/quirks.go
smira: so my point is the following (roughly) to stay compatbile:

| Talos version | Linux | xfs compat |
|----------|-----|--------|
| 1.10 | 6.12 | 6.12 |
| 1.9 | 6.12 | 6.6 |
| 1.8 | 6.6 | 6.6 |
| 1.7 | 6.6 | 6.1 |
| 1.6 | 6.1 | 6.1 |
| 1.5 | 6.1 | 6.1 |

pkg/makefs/xfs.go
```diff
@@ -46,8 +46,11 @@ func XFS(partname string, setters ...Option) error {
 	opts := NewDefaultOptions(setters...)
 
 	// The ftype=1 naming option is required by overlayfs.
-	// The bigtime=1 metadata option enables timestamps beyond 2038.
```
frezbo, line 49: `bigtime=1` is present on all xfs config files for `lts6_*` configs

hack/cleanup.sh
```diff
@@ -27,7 +27,7 @@ find ${PREFIX}/usr/bin -type f -exec strip --strip-all {} ';' || true
 # Remove header files, man files, and any other non-runtime dependencies.
```
frezbo, line 27: I think we can drop this script almost with `--exclude` directives in dockerfile

smira (approved):

PR #10629, opened 2025-04-02
feat: update dependencies

smira:
Brings in Linux 6.12.21, go 1.24.2.

Also updates Go dependencies, golangci-lint, etc.

The configuration was migrated, fix new linting errors.


internal/pkg/ntp/internal/spike/spike.go
```diff
@@ -74,7 +74,8 @@ func (d *Detector) IsSpike(sample Sample) bool {
 	var j float64
 
 	for i := range d.samples {
-		j += math.Pow(d.samples[i].Offset-d.samples[indexMin].Offset, 2)
+		offsetDiff := d.samples[i].Offset - d.samples[indexMin].Offset
+		j += offsetDiff * offsetDiff
```
dsseng, line 78: nit: Wouldn't `math.Pow` look better here?

dsseng (approved):

internal/pkg/ntp/internal/spike/spike.go
smira: the linter thinks unrolled version would work faster, which I think is true

dsseng: Valid unless the compiler unrolls it too

smira: I can't find any proof that it'd be unrolled, at least it's not a special case in the function itself.

frezbo (approved):

PR #10633, opened 2025-04-03
test: more extension modules

smira:
Update with the fix https://github.com/siderolabs/pkgs/pull/1200, load explicitly `xdma` and `ena` drivers.


frezbo (approved):
Should we try adding the ena extensions in our AWS test?

smira:
> Should we try adding the ena extensions in our AWS test?

yep, we can, in fact there's a base `ena` driver in the kernel, so our extension brings an out-of-tree "newer" version from Amazon

PR #10636, opened 2025-04-03
fix: race in the volume mount status handling

smira:
Wait for the volume mount status to be properly destroyed on service shutdown, and fix the controller to destroy it as well.


internal/app/machined/pkg/system/volumes.go
```diff
@@ -71,13 +78,14 @@ func (cond *volumeMountedCondition) Wait(ctx context.Context) error {
 }
 
 func (cond *volumeMountedCondition) String() string {
-	return fmt.Sprintf("volume %q to be mounted", cond.id)
+	return fmt.Sprintf("volume %q to be mounted", cond.volumeID)
```
smira, line 81: this allows cleans up the `dmesg` output a bit to be less confusing about which volume a service is waiting for

utkuozdemir (approved):

frezbo (approved):

Issue #10620, opened 2025-04-01
a resource to know that UKI (systemd-boot) is being used


smira:
So that e.g. Omni knows if it can/should use Image Factory's extra kernel args to modify kernel args via an upgrade.

PR #10639, opened 2025-04-04
feat: expose if system is booted with UKI

frezbo:
Expose if system is booted with a UKI in `securitystate` resource.

Fixes: #10620

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -1182,6 +1182,7 @@ spec:
             SHORT_INTEGRATION_TEST: yes
             WITH_UEFI: false
             IMAGE_REGISTRY: registry.dev.siderolabs.io
+            EXTRA_TEST_ARGS: "-talos.verifyukibooted=false"
         - name: e2e-bios-iso
           command: e2e-qemu
           withSudo: true
@@ -1192,6 +1193,7 @@ spec:
             VIA_MAINTENANCE_MODE: true
             WITH_ISO: true
             IMAGE_REGISTRY: registry.dev.siderolabs.io
+            EXTRA_TEST_ARGS: "-talos.verifyukibooted=false"
         - name: e2e-disk-image
           command: e2e-qemu
           withSudo: true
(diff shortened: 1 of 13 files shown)


internal/app/machined/pkg/controllers/runtime/security_state.go
```diff
@@ -92,6 +95,13 @@ func (ctrl *SecurityStateController) Run(ctx context.Context, r controller.Runti
 				secureBootState = true
 			}
 
+			defaultEntry, err := sdboot.ReadVariable(efivario.NewDefaultContext(), sdboot.LoaderEntryDefaultName)
+			if err == nil {
+				if strings.HasPrefix(defaultEntry, "Talos-") {
+					bootedWithUKI = true
+				}
+			}
```
dsseng, line 103: Is there by chance a better option like getting some info about the running kernel from the sysfs or procfs? None known to me, just an idea to detect this by the current state and not boot config (while the current solution is fine)

dsseng (approved):

internal/app/machined/pkg/controllers/runtime/security_state.go
frezbo: ideal would have been to see if the StubInfo has been set https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html#StubInfo, but since when we do kexec and skip sd-boot completely the best is to see if the Default entry has been set, which the installer always does

smira (approved):
👍 , I think it'd be cool to add it to the integration test as well

.kres.yaml
```diff
@@ -1182,6 +1182,7 @@ spec:
             SHORT_INTEGRATION_TEST: yes
             WITH_UEFI: false
             IMAGE_REGISTRY: registry.dev.siderolabs.io
+            EXTRA_TEST_ARGS: "-talos.verifyukibooted=false"
```
frezbo, line 1185: hopefully I caught all cases where the test needs to be disabled

PR #10640, opened 2025-04-04
fix: Avoid printing terminating null byte in SELinux context

jnohlgard:
SELinux context (file extended attribute `security.selinux`) is a null-terminated string. We need to properly convert the C-string inside the byte slice into a Golang string to avoid also printing the null byte in the ls command output.

Discovered when piping talosctl ls output to grep:

    $ talosctl ls /lib/modules/ -r -l | grep '\.ko'
    grep: (standard input): binary file matches

### DIFF
--- a/cmd/talosctl/cmd/talos/list.go
+++ b/cmd/talosctl/cmd/talos/list.go
@@ -16,6 +16,7 @@ import (
 
 	humanize "github.com/dustin/go-humanize"
 	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
 	machineapi "github.com/siderolabs/talos/pkg/machinery/api/machine"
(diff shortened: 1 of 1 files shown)


frezbo:
@dsseng I thought we already had this fixed :thinking:

dsseng:
> @dsseng I thought we already had this fixed 🤔

I also thought so

dsseng (approved):
Thanks for the fix.

We're currently doing the same for process labels, and your solution looks cleaner: https://github.com/siderolabs/talos/blob/73c9e91c6ab7e237ea939d8bb858b81414844a16/internal/pkg/miniprocfs/processes.go#L167

frezbo:
> Thanks for the fix.
> 
> We're currently doing the same for process labels, and your solution looks cleaner:
> 
> https://github.com/siderolabs/talos/blob/73c9e91c6ab7e237ea939d8bb858b81414844a16/internal/pkg/miniprocfs/processes.go#L167

ahh so we fixed somewhere else, good catch

PR #10642, opened 2025-04-04
chore: small developer improvements

frezbo:
- Use `make` variables to auto-detect target platform and not hardcode it.
- Separate one of the aws tests to a little different cron time so it works (running both parallely never worked).
- Pass in `--base-installer-image` so when a developer runs `make image-installer` with some extensions `IMAGE_REGISTRY` is honored.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -2482,7 +2482,7 @@ spec:
         - self-hosted
         - generic # we can use generic here since the tests run against a remote talos cluster
       crons:
-        - '30 7 * * *'
+        - '30 5 * * *'
       triggerLabels:
         - integration/aws-nvidia-oss
       steps:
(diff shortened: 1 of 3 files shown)


frezbo:
this has been bothering me for a while and I had to `s|linux/amd64|linux/arm64` for a while :sweat_smile:

smira (approved):

.kres.yaml
```diff
@@ -2482,7 +2482,7 @@ spec:
         - self-hosted
         - generic # we can use generic here since the tests run against a remote talos cluster
       crons:
-        - '30 7 * * *'
+        - '30 5 * * *'
```
dsseng, line 2485: nice catch

dsseng (approved):

PR #10644, opened 2025-04-04
fix: handling of override path for registry mirrors

smira:
See https://github.com/siderolabs/talos/discussions/10641

The problem is that `overridePath` config option should be per-endpoint, not a global one.

As we can't easily change v1alpha1 config, change the interface instead, and update the controller which generates final registry config to use per-endpoint setting.

Ensure that `registryd` endpoint never uses `overridePath`.

E.g., with this PR:

```toml
[host]
  [host.'http://127.0.0.1:3172']
    capabilities = ['pull', 'resolve']
  [host.'http://172.20.0.1:5006/v2/']
    capabilities = ['pull', 'resolve']
    override_path = true
```


hack/test/patches/image-cache.yaml
```diff
@@ -2,20 +2,6 @@ machine:
   features:
     imageCache:
       localEnabled: true
-  registries:
```
frezbo, line 5: was this intentional, this was added to make sure we never reach out to the internet

internal/pkg/containers/cri/containerd/hosts.go
```diff
@@ -110,16 +110,16 @@ func GenerateHosts(cfg config.Registries, basePath string) (*HostsConfig, error)
 		var hostsConfig HostsConfiguration
 
 		for _, endpoint := range endpoints.Endpoints() {
-			u, err := url.Parse(endpoint)
+			u, err := url.Parse(endpoint.Endpoint())
```
frezbo, line 113: feels super weird `endpoints.Endpoints()[*].EndPoint()` :sweat_smile:

frezbo (approved):

hack/test/patches/image-cache.yaml
smira: oops, I should revert

PR #10645, opened 2025-04-04
feat: implement user volumes

smira:
User volumes are identified by a short name which serves both
as a `/var/mnt` mount point and a partition label.

User volumes can be added and removed on the fly, and they are
automatically propagated into the `kubelet` mount namespace.

Also deprecate `.machine.disks`.


smira:
Should be good now, missing stuff:

- [ ] add support for encryption (I would like this to be a separate PR)
- [ ] documentation updates (this PR)

smira:
> The 'diskPartitions' slice is allocated but never populated; consider replacing 'len(diskPartitions)' with the actual count of user volumes or using a more appropriate metric for computing the GPT alignment.

nice catch, and fixed

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -0,0 +1,218 @@
...
+//nolint:gocyclo,cyclop
+func (ctrl *UserVolumeConfigController) Run(ctx context.Context, r controller.Runtime, _ *zap.Logger) error {
+	for {
+		select {
+		case <-r.EventCh():
+		case <-ctx.Done():
+			return nil
+		}
+
+		cfg, err := safe.ReaderGetByID[*config.MachineConfig](ctx, r, config.ActiveID)
```
frezbo, line 81: how would a try mode apply for a uservolume config work, or we don't support that?

internal/pkg/partition/constants.go
```diff
@@ -14,6 +14,7 @@ const (
 	EFISystemPartition  Type = "C12A7328-F81F-11D2-BA4B-00A0C93EC93B"
 	BIOSBootPartition   Type = "21686148-6449-6E6F-744E-656564454649"
 	LinuxFilesystemData Type = "0FC63DAF-8483-4772-8E79-3D69D8477DE4"
+	UserVolumePartition Type = "DAD6FD75-7A68-42AE-A92D-A9877FB083CF"
```
frezbo, line 17: is this a custom uuid?

frezbo (approved):

internal/pkg/partition/constants.go
smira: Yes, it's a random one, any suggestions on what we could reuse?

internal/app/machined/pkg/controllers/block/user_volume_config.go
smira: It should work the same way, it will create the volume and then unmount it if the config is reverted.

internal/pkg/partition/constants.go
frezbo: I think we should use the same as linux filesystems type since in the end it's a linux partition , actually we can use the variable data partition uuid defined in UAPI https://uapi-group.org/specifications/specs/discoverable_partitions_specification/#partition-attribute-flags

internal/app/machined/pkg/controllers/block/user_volume_config.go
frezbo: So a dry run sets the active id temporarily, that was what confused me 👍

smira: yeah, it does, and then "reverts" it unless someone applies the machine config in non-try mode

internal/pkg/partition/constants.go
smira: yep, thanks, makes sense, I will update

smira: I think I can go back to `LinuxFilesystemData`, as there is no direct match

smira:
Added documentation.

PR #10662, opened 2025-04-07
fix: uki boot detection

frezbo:
fix: uki boot detection
    
Fix UKI boot detection
    
Also fix bug introduced by #10640 which imported the unix package making talosctl non-unix builds broken


frezbo:
AWS CI fixes will be handled in a different PR

smira (approved):

Issue #10650, opened 2025-04-06
Update hugo (current version doesn't support arm)


Orzelius:
Hugo is currently at version 0.111 from two years ago. The newest version is 0.145

I ran into this whilst trying to run the docs server locally, but failed to do so as the 0.111 container doesn't support arm.

Stuff that should be done:
- [ ] change to a different image (current image is no longer maintained) `hugomods/hugo` seems to be the new way to go
- [ ] switch to [hugo modules](https://gohugo.io/hugo-modules/introduction/). This would allow to remove the [docsy clone](https://github.com/siderolabs/talos/tree/main/website/themes/docsy) and just add it as a module https://www.docsy.dev/docs/updating/convert-site-to-module/
- [ ] Go through breaking changes and fix issues that will inevitably come up

If this seems like an okay plan I could work on this as I've already spent a little time researching the topic

PR #10672, opened 2025-04-08
update hugo

Orzelius:
closes [#10650](https://github.com/siderolabs/talos/issues/10650)

Ended up being a slightly more of a pain that anticipated :p

I know having one commit per pr is standard at sidero, but I think for this case it at least makes sense to have the changes split up into separate commits for the ease of review. If need be this can be squashed into one commit before merging

The looks are slightly different. Mainly due to the underlying docsy and bootstrap version changes (will depend on your screen size)

New
image
image

Old
image
image


rothgar:
I agree both of those options are not very useful. If they can be removed without making custom theme files (disabled in config) even better.

Orzelius:
I've tamed netlify and this is now ready for review! Please look at the commits one by one, I put my heart into each and every :p

website/hugo.toml
```diff
@@ -78,7 +77,7 @@ style = "solarized-dark"
 # Comment out if you don't want the "print entire section" link enabled.
 [outputs]
 home = ["HTML", "RSS", "netlify_redirect"]
-section = ["HTML", "print", "RSS"]
```
frezbo, line 81: what was print? saving to pdf?

website/hugo.toml
```diff
@@ -189,7 +188,7 @@ version = "v1.0"
 #  Set to true to disable breadcrumb navigation.
 breadcrumb_disable = false
 # Set to true to disable the About link in the site footer
```
frezbo, line 190: comment needs updation

website/package-lock.json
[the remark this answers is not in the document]
frezbo: ahh I see the `package.json` below, ignore

frezbo (approved):

website/hugo.toml
Orzelius: good catch! fixed in [ab9e6ca](https://github.com/siderolabs/talos/pull/10672/commits/ab9e6ca21aaf97083a5052289f46f7e270fef26e)

rothgar:
Is there a netlify preview for this?

Orzelius:
> Is there a netlify preview for this?

Yes, you can find it under the workflows below!
https://deploy-preview-10672--talos-dev.netlify.app/

smira:
Thank you, I did some small fixups, looks like `relref` works a bit different now.

PR #10699, opened 2025-04-09
Support RSA keys for the Talos API

### LINKED ISSUE #10668: Support RSA certificates for Talos API
## Feature Request

It would be nice we could use RSA mTLS certificates for the Talos API, since not all ecosystems support Ed25519 (e.g. C#/dotnet does not). 

It is currently possible to configure an RSA CA for this, and it works fine for the controlplane, but workers only do Ed25519 CSRs (which fails because they want a Ed25519 signature which the RSA CA cannot provide).

I think there are only only two things necessary to make this work:

A change to the logic here to make a RSA CSR if the issuing CA is RSA: https://github.com/siderolabs/talos/blob/e8e7f75c77c9670b7fb2172b968085214714aa5b/internal/app/machined/pkg/controllers/secrets/api.go#L348-L352

And to support that the information needs to be available somewhere, maybe the `OSRootSpec`, since the issuing CA kinda gets removed from it here: https://github.com/siderolabs/talos/blob/e8e7f75c77c9670b7fb2172b968085214714aa5b/internal/app/machined/pkg/controllers/secrets/root.go#L164-L173

Alternatively, since my tooling only directly needs to talk to the controlplane, it would also be fine (and perhaps ideal) if the Talos nodes use Ed25519 internally but also have RSA available on the controlplane. But that seems like a lot more work.

gwenya:
# Pull Request

Fixes #10668.

The code does not pass conformance checks has no tests yet, and compilation probably breaks with non-default build flags. 

That will of course be fixed, but I wanted to propose this here first before doing anything major.

## What? (description)

1. Make worker nodes generate RSA keys for the Talos API if the CA uses an RSA certificate
2. Add a cli flag `--rsa` to talosctl commands that generate secrets, to generate an RSA CA instead of an Ed25519 one for the Talos API

## Why? (reasoning)

As mentioned in #10668, some languages/environments don't have Ed21159 support, which makes it difficult to write tooling that consumes the Talos API in such languages. 

## Acceptance

### DIFF
--- a/pkg/machinery/config/generate/secrets/bundle.go
+++ b/pkg/machinery/config/generate/secrets/bundle.go
@@ -21,12 +21,12 @@ import (
 )
 
 // NewBundle creates secrets bundle generating all secrets.
-func NewBundle(clock Clock, versionContract *config.VersionContract) (*Bundle, error) {
+func NewBundle(clock Clock, rsa bool, versionContract *config.VersionContract) (*Bundle, error) {
 	bundle := &Bundle{
 		Clock: clock,
 	}
 
-	err := bundle.populate(versionContract)
+	err := bundle.populate(rsa, versionContract)
 	if err != nil {
 		return nil, err
 	}
@@ -140,7 +140,7 @@ func NewBundleFromKubernetesPKI(pkiDir, bootstrapToken string, versionContract *
 		},
 	}
 
-	err = bundle.populate(versionContract)
+	err = bundle.populate(false, versionContract)
 	if err != nil {
 		return nil, err
 	}
@@ -191,7 +191,7 @@ func NewBundleFromConfig(clock Clock, c config.Config) *Bundle {
 // populate fills all the missing fields in the secrets bundle.
 //
 //nolint:gocyclo,cyclop
-func (bundle *Bundle) populate(versionContract *config.VersionContract) error {
+func (bundle *Bundle) populate(rsa bool, versionContract *config.VersionContract) error {
 	if bundle.Clock == nil {
 		bundle.Clock = NewClock()
 	}
@@ -259,7 +259,7 @@ func (bundle *Bundle) populate(versionContract *config.VersionContract) error {
 	}
 
 	if bundle.Certs.OS == nil {
-		talosCA, err := NewTalosCA(bundle.Clock.Now())
+		talosCA, err := NewTalosCA(bundle.Clock.Now(), rsa)
 		if err != nil {
 			return err
 		}
--- a/pkg/machinery/resources/secrets/os_root.go
+++ b/pkg/machinery/resources/secrets/os_root.go
@@ -33,6 +33,7 @@ type OSRootSpec struct {
 	AcceptedCAs     []*x509.PEMEncodedCertificate     `yaml:"acceptedCAs" protobuf:"5"`
 	CertSANIPs      []netip.Addr                      `yaml:"certSANIPs" protobuf:"2"`
 	CertSANDNSNames []string                          `yaml:"certSANDNSNames" protobuf:"3"`
+	Rsa             bool                              `yaml:"rsa" protobuf:"6"`
 
 	Token string `yaml:"token" protobuf:"4"`
 }
--- a/cmd/talosctl/cmd/mgmt/cluster/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create.go
@@ -200,6 +200,7 @@ var (
 	withIOMMU                 bool
 	configInjectionMethodFlag string
 	mountOpts                 opts.MountOpt
+	rsa                       bool
 )
 
 // createCmd represents the cluster up command.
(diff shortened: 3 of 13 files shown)


smira:
I think optionally supporting RSA for Talos API is a good change, but advertising it any way (e.g. having a flag to enable it) is a bad idea - as RSA is worse in all dimensions:

* security
* key size
* performance

So if someone is forced to use RSA, they could, but this should be discouraged.

smira:
> I completely agree that it should be avoided, and I'm fine with throwing the talosctl flag out again, the main reason I added it was for testing if the server side stuff actually works with rsa.

We can even have a flag, but mark it as hidden for example (for testing). But I think in this case unit-tests should be enough to handle this case.

gwenya:
@smira Can you give me some hints on how to unit test this? Since the worker needs a controlplane to talk to for getting its CSR signed. I see that there is `LocalGenerator` that can presumably be used to substitute  `RemoteGenerator` in a unit test, but I have no clue how to pass that into the function, since the entrypoint to the reconciler is `Run` which is part of a foreign interface.

smira:
> @smira Can you give me some hints on how to unit test this? Since the worker needs a controlplane to talk to for getting its CSR signed. I see that there is `LocalGenerator` that can presumably be used to substitute `RemoteGenerator` in a unit test, but I have no clue how to pass that into the function, since the entrypoint to the reconciler is `Run` which is part of a foreign interface.

Right now we're busy with 1.10 release, but I can take a look once we have time

pkg/machinery/config/generate/secrets/bundle.go
```diff
@@ -21,12 +21,12 @@ import (
 )
 
 // NewBundle creates secrets bundle generating all secrets.
-func NewBundle(clock Clock, versionContract *config.VersionContract) (*Bundle, error) {
+func NewBundle(clock Clock, rsa bool, versionContract *config.VersionContract) (*Bundle, error) {
```
smira, line 24: `rsa` should an functional option, let's keep the API clean

pkg/machinery/resources/secrets/os_root.go
```diff
@@ -33,6 +33,7 @@ type OSRootSpec struct {
 	AcceptedCAs     []*x509.PEMEncodedCertificate     `yaml:"acceptedCAs" protobuf:"5"`
 	CertSANIPs      []netip.Addr                      `yaml:"certSANIPs" protobuf:"2"`
 	CertSANDNSNames []string                          `yaml:"certSANDNSNames" protobuf:"3"`
+	Rsa             bool                              `yaml:"rsa" protobuf:"6"`
```
smira, line 36: I don't think we should have this field ever, as it's already in `IssuingCA` indirectly

smira (commented):
overall, lack of unit-tests covering the actual flow with RSA

pkg/machinery/resources/secrets/os_root.go
smira: Worker nodes don't have `IssuingCA` at all - otherwise they would have issued a certificate for themselves.

gwenya: The configs that talosctl generates give the workers an `IssuingCA` but with only the `Crt` set, and `Key` empty. 
The worker then appends that `Crt` to `AcceptedCAs` and clears `IssuingCA`: https://github.com/siderolabs/talos/blob/c55af59a2583b12d4953c8459fc52f7b22cae9c5/internal/app/machined/pkg/controllers/secrets/root.go#L164-L173

My change stores the signature algorithm of said `Crt` in the `OSRootSpec` struct so that this information can later be used to create the appropriate CSR: https://github.com/gwenya/talos/blob/a3672d11482c29e6d798f397d1a329cde6ed6ed3/internal/app/machined/pkg/controllers/secrets/root.go#L170-L175

smira: Storing algorithm sounds better now, but we should probably use proper type to make it look nice in the resource itself, I can work it out later on.

gwenya: I initially used the `x509.SignatureAlgorithm` type from the standard library, but `make docs` failed to generate protobuf for that, so I switched it to `int`. If there is some way to make that work I would prefer it.

gwenya:
@smira Could you look into why `make docs` and `make conformance` fail? 
What do I need to do to make my gpg identity not be an "unknown entity"? I added it to my github account but that doesn't seem to have helped.
The branch is currently 3 commits behind `main`, I don't understand why it says 5406 ahead :sweat_smile:, but even when it was freshly rebased it complained about being 3 commits ahead.

gwenya:
Hi, is there anything I can do to help get this merged?

smira:
> Hi, is there anything I can do to help get this merged?

no, sorry for the delay, but we prioritize fixing bugs over new features. I plan to figure out a clean way to implement and merge this

smira:
See https://github.com/siderolabs/crypto/pull/39 - this allows a more generic and easier approach to solve this.

I'm closing this PR, as we're going to print new crypto library soon.

Issue #10574, opened 2025-03-20
Add a resource with Talos version


smira:
A trivial resource under `runtime` which reports Talos version. 

It's created only once on boot, and never modified.

Reason: make it easier for Omni to watch everything via resources, skipping other API calls.

PR #10703, opened 2025-04-10
feat: add a version resource

frezbo:
```text
❯ talosctl -n 10.5.0.3 get version
NODE       NAMESPACE   TYPE      ID        VERSION   VERSION
10.5.0.3   runtime     Version   version   1         v1.10.0-alpha.3-33-g84f69f043-dirty
```

Fixes: #10574


smira:
let's please also add a test to `integration/api` - we already have a test on `Version()` gRPC API, and we can cross-check via this resource

PR #10711, opened 2025-04-11
test: add negative tests for SELinux

### LINKED ISSUE #10615: SELinux negative tests
When `enforcing=1`, I _can't_ from a privileged pod:

* read machine config via `/system/state/config.yaml` (even when mounting it) 
* disallow mounting STATE from a privileged pod
* accessing machined API file socket
* ptrace `init`
* ...

dsseng:
Make sure a privileged pod cannot violate some of the important security rules enforced by SELinux.

Fixes #10615


internal/integration/api/selinux.go
```diff
@@ -336,7 +337,121 @@ func (suite *SELinuxSuite) TestSecurityState() {
 // TODO: test for system and CRI container labels
 // TODO: test labels for unconfined system extensions, pods
 // TODO: test for no avc denials in dmesg
-// TODO: start a pod and ensure access to restricted resources is denied
+
+// TestNoPtrace confirms ptracing system processes is prohibited in enforcing mode.
```
dsseng, line 341: I'm not sure, but perhaps we should create one pod and run all the assertions against it, so these tests run faster

internal/integration/base/k8s.go
```diff
@@ -266,6 +268,10 @@ func (p *pod) Create(ctx context.Context, waitTimeout time.Duration) error {
 	return p.suite.WaitForPodToBeRunning(ctx, waitTimeout, p.namespace, p.name)
 }
 
+func (p *pod) SetQuiet(quiet bool) {
```
dsseng, line 271: This is to silence stderr output from tests like this, which expect commands to fail

internal/integration/api/selinux.go
```diff
@@ -336,7 +337,121 @@ func (suite *SELinuxSuite) TestSecurityState() {
...
+	ctx, cancel := context.WithTimeout(suite.ctx, time.Second*5)
+	defer cancel()
+
+	_, stderr, err = ptraceTestPodDef.Exec(
+		ctx,
+		"strace -p 1",
+	)
+
+	// in case of successful attach it will be context.DeadlineExceeded
+	suite.Assert().NotNil(err)
```
smira, line 373:
```suggestion
	suite.Require().Error(err)
```

(this would abort execution in case error is `nil`, which might panic below)

internal/integration/api/selinux.go
```diff
@@ -336,7 +337,121 @@ func (suite *SELinuxSuite) TestSecurityState() {
...
+	suite.Assert().Contains(err.Error(), "command terminated with exit code 1")
```
smira, line 374:
```suggestion
	suite.Assert().ErrorContains(err, "command terminated with exit code 1")
```

internal/integration/base/k8s.go
```diff
@@ -236,13 +236,15 @@ type podInfo interface {
 	WithNodeName(nodeName string) podInfo
 	Create(ctx context.Context, waitTimeout time.Duration) error
 	Delete(ctx context.Context) error
+	SetQuiet(quiet bool)
```
smira, line 239: nit, to match previous `With*` methods, I'd make it:

```suggestion
	WithQuiet(quiet bool) podInfo
```

This allows method chaining

smira (approved):

internal/integration/api/selinux.go
smira: I think this one is fine

PR #10713, opened 2025-04-11
test: attempt to debug/fix NVIDIA tests

smira:
Add more logging output.


hack/test/e2e.sh
```diff
@@ -37,7 +37,6 @@ export KUBERNETES_VERSION=${KUBERNETES_VERSION:-1.33.0-rc.0}
 
 export NAME_PREFIX="talos-e2e-${SHA}-${PLATFORM}"
 export TIMEOUT=1200
-export NUM_NODES=${TEST_NUM_NODES:-6}
```
smira, line 40: I couldn't find it being used anywhere

hack/test/e2e-aws-prepare.sh
```diff
@@ -9,6 +9,12 @@ REGION="us-east-1"
 function cloud_image_upload() {
   CLOUD_IMAGES_EXTRA_ARGS=("--name-prefix=${1}" "--target-clouds=aws" "--architectures=amd64" "--aws-regions=${REGION}")
 
+  case "${1}" in
+    talos-e2e-nvidia-oss)
```
smira, line 13: need to invent another hack to make controlplanes (regular AMIs) also use force BIOS mode

hack/test/e2e.sh
frezbo: It used to be used, maybe not anymore

hack/test/e2e-aws-prepare.sh
smira: scratch that, the test is skipped, so it doesn't matter if I update all machines or just one

frezbo (approved):

Issue #10696, opened 2025-04-08
"Interface route" cannot be modified and is set to 0 metric


ArcherSeven:
### Description

Talos creates both a default route with metric 1024 and what I will naively call an "interface route" set to metric of 0 for each configured interface.

If multiple interfaces connect to the same subnet, it is possible to modify the default/gateway route, but not the interface route, and since the interface route is set to metric 0, it cannot be overwritten.

Example config (where an unconfigured interface is left to use DHCP)

```
    network:
        interfaces:
            - addresses:
                - 172.22.1.21/24
              deviceSelector:
                busPath: 0000:02:00.0
              routes:
                - gateway: 172.22.1.1
                  metric: 2048
                  network: 0.0.0.0/0
                - network: 172.22.1.0/24
                  metric: 2048
```

Resulting routes:

```
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         172.22.1.1      0.0.0.0         UG    1024   0        0 eno1
0.0.0.0         172.22.1.1      0.0.0.0         UG    2048   0        0 enp2s0
10.244.1.0      0.0.0.0         255.255.255.0   U     0      0        0 cni0
172.22.1.0      0.0.0.0         255.255.255.0   U     0      0        0 eno1
172.22.1.0      0.0.0.0         255.255.255.0   U     0      0        0 enp2s0
172.22.1.0      0.0.0.0         255.255.255.0   U     2048   0        0 enp2s0
```

This means that the actual order of preference for the routes is based on order of creation, so in the above situation, when the router reboots, the routes end up as:

```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         172.22.1.1      0.0.0.0         UG    1024   0        0 eno1
0.0.0.0         172.22.1.1      0.0.0.0         UG    2048   0        0 enp2s0
172.22.1.0      0.0.0.0         255.255.255.0   U     0      0        0 enp2s0
172.22.1.0      0.0.0.0         255.255.255.0   U     0      0        0 eno1
172.22.1.0      0.0.0.0         255.255.255.0   U     2048   0        0 enp2s0
```

At which point no destination is pingable (since `enp2s0` is configured, but the interface isn't actually plugged in)

We need to either be able to change these "interface" routes, or have them set to a higher metric so we can override them.

### Environment

- Talos version: v1.9.2 (other versions 1.7+ also tested)
- Kubernetes version: v1.27.15
[...]

PR #10732, opened 2025-04-15
feat: support address priority

smira:
See https://github.com/jsimonetti/rtnetlink/pull/256

See https://github.com/siderolabs/talos/issues/10696

Support setting address priority, this is important to ensure that the prefix route has appropriate priority.

For now, we don't have it exposed in the machine config except for the DHCP4 operator, so now both routes created explicitly by DHCP and routes created implicitly have same metric/priority:

```
172.20.0.2   network     RouteStatus   inet4//172.20.0.0/24/1024                                1         172.20.0.0/24                                enp0s2      1024
172.20.0.2   network     RouteStatus   inet4/172.20.0.1//1024                                   1                                         172.20.0.1   enp0s2      1024
```

Before this change, the first route would have a metric zero.

### DIFF
--- a/api/resource/definitions/network/network.proto
+++ b/api/resource/definitions/network/network.proto
@@ -18,6 +18,7 @@ message AddressSpecSpec {
   uint32 flags = 5;
   bool announce_with_arp = 6;
   talos.resource.definitions.enums.NetworkConfigLayer config_layer = 7;
+  uint32 priority = 8;
 }
 
 // AddressStatusSpec describes status of rendered secrets.
@@ -32,6 +33,7 @@ message AddressStatusSpec {
   talos.resource.definitions.enums.NethelpersFamily family = 8;
   talos.resource.definitions.enums.NethelpersScope scope = 9;
   uint32 flags = 10;
+  uint32 priority = 11;
 }
 
 // BondMasterSpec describes bond settings if Kind == "bond".
--- a/internal/app/machined/pkg/controllers/network/nftables_chain_test.go
+++ b/internal/app/machined/pkg/controllers/network/nftables_chain_test.go
@@ -8,6 +8,7 @@ import (
 	"net/netip"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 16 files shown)


api/resource/definitions/network/network.proto
```diff
@@ -18,6 +18,7 @@ message AddressSpecSpec {
   uint32 flags = 5;
   bool announce_with_arp = 6;
   talos.resource.definitions.enums.NetworkConfigLayer config_layer = 7;
+  uint32 priority = 8;
```
dsseng, line 21: Should we maybe adhere to the kernel (and IP stack generally) name `metric` here?

internal/app/machined/pkg/controllers/network/nftables_chain_test.go
```diff
@@ -351,12 +353,19 @@ func (s *NfTablesChainSuite) TestClampMSS() {
...
 		type filter hook input priority filter; policy accept;
 		meta nfproto ipv4 tcp flags syn / syn,rst tcp option maxseg size > 1380 tcp option maxseg size set 1380
 		meta nfproto ipv6 tcp flags syn / syn,rst tcp option maxseg size > 1360 tcp option maxseg size set 1360
 	}
+}`, `table inet talos-test {
+	chain test1 {
+		type filter hook input priority filter; policy accept;
+		meta nfproto ipv4 tcp flags & (syn | rst) == syn tcp option maxseg size > 1380 tcp option maxseg size set 1380
+		meta nfproto ipv6 tcp flags & (syn | rst) == syn tcp option maxseg size > 1360 tcp option maxseg size set 1360
+	}
```
dsseng, line 368: Is this unrelated and just a test fix?

dsseng (approved):

api/resource/definitions/network/network.proto
smira: kernel uses priority actually, and we already do use `priority` for the RouteSpec

internal/app/machined/pkg/controllers/network/nftables_chain_test.go
smira: yes, see the comment, `nft` CLI of different versions decodes the rules a bit different way

api/resource/definitions/network/network.proto
dsseng: Okay

dsseng (approved):

PR #10747, opened 2025-04-17
feat: support encryption config for user volumes

### LINKED ISSUE #10719: add encryption support for user volumes
Support encryption configuration for user volumes.

smira:
No big changes, mostly wiring things together:

* implement encryption config (identical to existing v1alpha1 one) for user volume configuration
* provide validation, some small fixes
* add support for encrypted user volumes in `talosctl cluster create`

Fixes #10719


pkg/machinery/config/types/block/encryption.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	var errs error
+
+	switch s.EncryptionProvider {
+	case block.EncryptionProviderLUKS2:
+	case block.EncryptionProviderNone:
+		fallthrough
+	default:
+		errs = errors.Join(errs, fmt.Errorf("unsupported encryption provider: %s", s.EncryptionProvider))
```
dsseng, line 157: Won't this cause default to be always called? https://go.dev/wiki/Switch#fall-through

I might be wrong, this just feels a bit odd

smira: Go is different from C, it defaults not to fallthrough, so it should be equivalent to expect a single value to be ok

I put it this way, as the linter ensures I cover all type constant values, if we add more values in the future

frezbo (approved):

PR #10750, opened 2025-04-17
fix: sync PCR extension with volume provisioning lifecycle

### LINKED ISSUE #10665: sync PCR extension with volume provisioning
Ensure that volume encryption happens only before is PCR extended finally.

Document proper behavior, that TPM encryption might need a reboot.

smira:
Ensure volumes are not locked to the wrong value of PCR.

Fixes #10665


internal/app/machined/pkg/controllers/hardware/pcr_status.go
```diff
@@ -0,0 +1,171 @@
...
+
+	runtimetalos "github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/internal/pkg/secureboot"
+	"github.com/siderolabs/talos/internal/pkg/secureboot/tpm2"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+	"github.com/siderolabs/talos/pkg/machinery/resources/block"
+	"github.com/siderolabs/talos/pkg/machinery/resources/hardware"
+)
+
+// PCRStatusController populates PCI device information.
```
frezbo, line 25:
```suggestion
// PCRStatusController populates PCR status 
```

internal/app/machined/pkg/controllers/hardware/pcr_status.go
frezbo, line 1: this is nice, this allows us to extend with filesystem information and further allow user volumes to be locked to PCR

frezbo (approved):

internal/app/machined/pkg/controllers/hardware/pcr_status.go
smira: thanks, fixed

PR #10758, opened 2025-04-19
docs: add a dark theme

Orzelius:
* add a dark theme
* fix the language tag 1px offset issue

This took a little longer than anticipated, but I defeated the rebeliant stylesheets and I think even the search still works :p

## new
image
image

## old
image


website/layouts/partials/hooks/body-end.html
```diff
@@ -1,8 +1,3 @@
-<!--- https://aaronluna.dev/blog/add-copy-button-to-code-blocks-hugo-chroma/ -->
-{{ if (findRE "<pre" .Content 1) }}
```
frezbo, line 2: :+1: on removing this

frezbo (approved):

PR #10765, opened 2025-04-21
docs: add caveat for kubespan and host ports

rsmitty:
This PR adds a note to the limitations for Kubespan to point out that allowing 51280 in the port range for pods can cause confusing conflicts on a node where the port is assigned to the pod.


frezbo (approved):

dsseng (approved):
v1.7 is missing despite 1.5 and 1.6 being modified here, maybe you accidentally forgot that

rsmitty:
Yep, missed that one. Thanks @dsseng. Will amend to fix then merge.

Issue #10501, opened 2025-03-10
logToConsole: true in extension service does not work


abckey:
### Description
Defining `logToConsole: true` in extension service does not work as expected, logs of the extension service do not show up in kernel logs from output of `talosctl dmesg` 

I tested with a simple `ext-echo` extension service with echo.yaml below:

echo.yaml
```
name: echo
container:
  entrypoint: /bin/sh
  args:
    - -c
    - while true;do echo "echotest on $(date)";sleep 10;done
restart: always
logToConsole: true
```

### Logs

logs can be seen from `talosctl logs ext-echo`:
```
 # talosctl logs ext-echo
10.129.163.32: echotest on Mon Mar 10 11:49:25 UTC 2025
10.129.163.32: echotest on Mon Mar 10 11:49:35 UTC 2025
10.129.163.32: echotest on Mon Mar 10 11:49:45 UTC 2025
10.129.163.32: echotest on Mon Mar 10 11:49:55 UTC 2025
10.129.163.32: echotest on Mon Mar 10 11:50:05 UTC 2025
10.129.163.32: echotest on Mon Mar 10 11:50:15 UTC 2025
```

but not from `talosctl dmesg|grep echo`
```
 # talosctl dmesg |grep echo
10.129.163.32: user: warning: [2025-03-10T11:49:23.3510105Z]: [talos] [initramfs] enabling system extension echo 1.0
10.129.163.32: user: warning: [2025-03-10T11:49:23.8943865Z]: [talos] service[ext-echo](Starting): Starting service
10.129.163.32: user: warning: [2025-03-10T11:49:23.8944995Z]: [talos] service[ext-echo](Waiting): Waiting for service "containerd" to be "up"
10.129.163.32: user: warning: [2025-03-10T11:49:24.9044095Z]: [talos] service[ext-echo](Preparing): Running pre state
10.129.163.32: user: warning: [2025-03-10T11:49:24.9052655Z]: [talos] service[ext-echo](Preparing): Creating service runner
10.129.163.32: user: warning: [2025-03-10T11:49:25.1094805Z]: [talos] service[ext-echo](Running): Started task ext-echo (PID 4513) for container ext-echo
[...]

PR #10770, opened 2025-04-22
fix: extension services logging to console

frezbo:
Fixes: #10501

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd.go
@@ -9,7 +9,7 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
+	"log"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


utkuozdemir:
Any idea why this didn't work? I remember testing and confirming this. Furthermore, it was `os.Stdout` since the beginning - only the check has changed from `c.debug` to `c.logToConsole`.

Also, does this require the same change? https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/system/runner/process/process.go#L205

frezbo:
> Any idea why this didn't work? I remember testing and confirming this. Furthermore, it was os.Stdout since the beginning - only the check has changed from c.debug to c.logToConsole.

yes, so with existing code it only logs to `serial console` and not dmesg, so if you look for extension logs in `talosctl dmesg` you wouldn't see it since it's kmsg, which the log.Writer handles

smira (approved):

frezbo:
> Also, does this require the same change? https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/system/runner/process/process.go#L205

I believe so, updating so we're consistent

PR #10776, opened 2025-04-22
fix: Set media type to OCI for image cache layer

jnohlgard:
The other components of the image cache image are using OCI media types, setting the layer to OCI instead of the default Docker layer type is more technically correct. This also works around a bug in skopeo that refuses to recognize Docker layers inside OCI images.


smira (approved):
thank you!

Issue #10777, opened 2025-04-22
provision test to test upgrades with BIOS


smira:
We can drop one `stable-to-current` test with "preserve", as Talos defaults to preserve always.

And instead add a test which does upgrade current-to-current with BIOS boot.

PR #10779, opened 2025-04-22
chore(ci): add provision test with bios

frezbo:
Add a provision test without UEFI boot.
Drop preserve upgrade test since it's the default now.

Fixes: #10777

### DIFF
--- a/internal/integration/provision/upgrade.go
+++ b/internal/integration/provision/upgrade.go
@@ -35,9 +35,9 @@ type upgradeSpec struct {
 	ControlplaneNodes int
 	WorkerNodes       int
 
-	UpgradePreserve bool
-	UpgradeStage    bool
-	WithEncryption  bool
+	UpgradeStage   bool
+	WithEncryption bool
+	WithBios       bool
 }
 
 const (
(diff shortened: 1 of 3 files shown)


internal/integration/provision/upgrade.go
```diff
@@ -285,7 +291,7 @@ func init() {
 		&UpgradeSuite{specGen: upgradePreviousToStable, track: 0},
 		&UpgradeSuite{specGen: upgradeStableToCurrent, track: 1},
 		&UpgradeSuite{specGen: upgradeCurrentToCurrent, track: 2},
-		&UpgradeSuite{specGen: upgradeStableToCurrentPreserve, track: 0},
+		&UpgradeSuite{specGen: upgradeCurrentToCurrentBios, track: 2},
```
smira, line 294:
```suggestion
		&UpgradeSuite{specGen: upgradeCurrentToCurrentBios, track: 0},
```

I would probably keep it in the same track, as track 2 runs all Kubernetes versions, which is already 40 minutes long.

internal/integration/provision/upgrade.go
```diff
@@ -38,6 +38,7 @@ type upgradeSpec struct {
 	UpgradePreserve bool
```
smira, line 38: let's please remove this as well, as it has no effect now

internal/integration/provision/upgrade.go
```diff
@@ -133,29 +133,33 @@ func upgradeCurrentToCurrent() upgradeSpec {
...
 	return upgradeSpec{
-		ShortName: fmt.Sprintf("prsrv-%s-%s", stableRelease, DefaultSettings.CurrentVersion),
+		ShortName: fmt.Sprintf("%s-same-ver-bios", DefaultSettings.CurrentVersion),
 
-		SourceKernelPath:     helpers.ArtifactPath(filepath.Join(trimVersion(stableRelease), constants.KernelAsset)),
-		SourceInitramfsPath:  helpers.ArtifactPath(filepath.Join(trimVersion(stableRelease), constants.InitramfsAsset)),
-		SourceInstallerImage: fmt.Sprintf("%s:%s", "ghcr.io/siderolabs/installer", stableRelease),
-		SourceVersion:        stableRelease,
-		SourceK8sVersion:     stableK8sVersion,
+		SourceKernelPath:     helpers.ArtifactPath(constants.KernelAssetWithArch),
```
smira, line 148: do we need to start with a disk image, or this would have already reproduced the bug you fixed?

smira (approved):
approved, but promoting to all provision pipelines

smira:
oops, I was too late to promote, I hope they still work

internal/integration/provision/upgrade.go
frezbo: It might not, I'll confirm once I'm back

frezbo: it does work, so all good

frezbo: actually scratch that, I did a wrong test, will do a different PR to fix

PR #10792, opened 2025-04-24
fix: handle correctly changing platform network config

### LINKED ISSUE #10787: Talos 1.10.0 beta.1 static IP address not applied
After booting Talos 1.10 beta 1, trying to apply static IP address fails.

I'm editing the configuration through F3. Hostname and all the IP addresses are not applied.

Image

Image

I can't get logs since I can't connect to the machine.
Provided is a screenshot from the machine's screen.

Image

I have tested also with 1.9.5 and 1.9.2 which both work without issues.

### Environment

- Talos version: 1.10.0.1-beta.1
- Platform: bare metal server - SuperMicro X11SPW-CTF
1x Intel Xeon Silver 4208 CPU
4x 32GB DDR4 RAM
1x 256GB nvme 
1x HDD 1TB Sata (directly connected to motherboard, no additional RAID controller)
1x nVidia P4 GPU

smira:
The bug was with incorrect condition: if `activeNetworkConfig` was ever set to non-nil value, it was stuck with this value forever, despite new network config being available with `networkConfig`.

In `talosctl dashboard` case, Talos `metal` platform always reports initial data (before META is available) which doesn't have any network config, but later on sends updates (if something updates META), so this bug leads to Talos being stuck with initial empty network config.

Fixes #10787

### DIFF
--- a/internal/app/machined/pkg/controllers/network/platform_config.go
+++ b/internal/app/machined/pkg/controllers/network/platform_config.go
@@ -168,7 +168,7 @@ func (ctrl *PlatformConfigController) Run(ctx context.Context, r controller.Runt
 		}
 
 		// prefer live network config over cached config always
-		if activeNetworkConfig == nil {
+		if networkConfig != nil {
 			activeNetworkConfig = networkConfig
 		}
 
--- a/internal/integration/api/platform.go
+++ b/internal/integration/api/platform.go
@@ -8,6 +8,7 @@ package api
 
 import (
 	"context"
+	"net/netip"
 	"time"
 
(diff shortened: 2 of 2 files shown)


frezbo (approved):
Maybe some tests

smira:
> Maybe some tests

yep, I thought about it, but couldn't figure out at first what it might be, but then I realized I could do integration test with QEMU, that would be way easier.

smira:
> Maybe some tests

I added a proper integration test now for metal/QEMU.

frezbo (approved):
:cool:

Issue #10754, opened 2025-04-18
Add bound driver to `pcidevice`


jfroy:
## Feature Request

`pcidevice` resources do not provide the driver/module bound to the device, if any. This would be useful for clusters where devices can be dynamically bound. This happens for nvme devices (binding to vfio for SPDK) and Nvidia GPUs (binding to vfio or nvidia).

PR #10793, opened 2025-04-24
chore: show bound driver in pcidevices info

frezbo:
Fixes: #10754


internal/app/machined/pkg/controllers/hardware/pcidevices.go
frezbo, line 1: the other change is moving out the code block outside the `for` since we run this only once anyways

smira (approved):

smira:
The only problem here I guess is that we don't refresh this info on driver change (if it happens).

frezbo:
> The only problem here I guess is that we don't refresh this info on driver change (if it happens).

yeh, we probably need to figure out how to refresh this

smira:
Would you be able to also incorporate a bump to https://github.com/siderolabs/go-pcidb/releases/tag/v0.3.1 ?

PR #10803, opened 2025-04-25
feat: allow SideroLink unique token in machine config

### LINKED ISSUE #10570: A configuration document to SideroLink unique token
Make it possible to set node's unique token (SideroLink) via machine configuration.

Use-case: Omni provider setting this by supplying a machine config document in "userdata".

smira:
Allow unique token to be specified in machine config, this way we can workaround a problem with META being non-persistend/wiped.

Fixes #10570


Unix4ever (approved):
❤️

PR #10815, opened 2025-04-28
fix: preserve kubelet image suffix

### LINKED ISSUE #10488: Upgrade k8s should take image tag into account (fat kubelet)
## Feature Request
To be able to use iscsi in a Talos cluster one needs to use the fat kubelet image. All is well until nodes using iscsi are rebooted after upgrading through talosctl upgrade-k8s as the kubelet image no longer refers to the fat image.

### Description

I envision two options as a fix: 
1. make upgrade-k8s take a seperate argument to ensure the image tag is in format $(version)-fat (f.e. --use-fat-kubelet or --release-tag=fat)

2. Make upgrade-k8s detect the fact that a fat kubelet is used on nodes and upgrade kubelet to the fat version on those nodes

smira:
Preserve the `-fat`, `-slim` kubelet image suffixes while upgrading `kubelet` images.

Fixes #10488

### DIFF
--- a/pkg/cluster/kubernetes/kubelet.go
+++ b/pkg/cluster/kubernetes/kubelet.go
@@ -167,6 +167,16 @@ func upgradeKubeletOnNode(ctx context.Context, cluster UpgradeProvider, options 
 	return nil
 }
 
+func extractKubeletVersionSuffix(imageRef string) string {
+	for _, suffix := range []string{"-fat", "-slim"} {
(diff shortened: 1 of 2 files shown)


utkuozdemir (approved):
Should we have a test for this?

smira:
> Should we have a test for this?

yes, good point, I will figure out something

PR #10820, opened 2025-04-30
chore: make qemu provider build on darwin

Orzelius:
This is the first step in splitting up the darwin qemu provider feature into smaller, easier to review bits.

Make the minimal changes required to have the qemu provider build on darwin.

* move linux specific logic into `*_linux.go` files (without making any changes to the logic)
* use a newer commit of the dhcpd fork that has the darwin compatible logic (see https://github.com/insomniacslk/dhcp/pull/550)
* make sure that the (broken) darwin qemu logic only runs when `DARWIN_QEMU` env var is set
* avoid using the linux specific cni.ns package (this can be changed later, just has to be done for now so the thing compiles)
* use a cross-platform fallocate package https://github.com/detailyang/go-fallocate

The following changes will arrive in the "first to run" order. Currently the next step that needs fixing is the preflight checks. After that come the various networking bits. This order will assure that each change can be run and tested before merging.


pkg/provision/providers/qemu/launch.go
```diff
@@ -94,7 +83,7 @@ type LaunchConfig struct {
 	// filled by CNI invocation
 	tapName string
 	vmMAC   string
-	ns      ns.NetNS
+	nsPath  string
```
frezbo, line 86: this seems not to be used now :thinking:

Orzelius: It's still used, but the logic has just moved to a function called `startQemuCmd`, which is implemented separately on darwin and linux. The linux implementation still runs the command in the new qemu network namespace.
https://github.com/siderolabs/talos/blob/665774cda8d8dbc82a2f53041504af563fb8b6b9/pkg/provision/providers/qemu/launch.go#L329-L331
https://github.com/siderolabs/talos/blob/665774cda8d8dbc82a2f53041504af563fb8b6b9/pkg/provision/providers/qemu/launch_linux.go#L218-L226

Maybe with the current name `startQemuCmd` it's unclear that on linux the command is started in a different ns. Perhaps it'd be more intuitive if I renamed it to `startQemuCmdOnNetwork` or something like that.

frezbo (approved):

pkg/provision/providers/qemu.go
```diff
@@ -2,17 +2,26 @@
...
+	"os"
+	"runtime"
 
 	"github.com/siderolabs/talos/pkg/provision"
 	"github.com/siderolabs/talos/pkg/provision/providers/qemu"
 )
 
 func newQemu(ctx context.Context) (provision.Provisioner, error) {
+	// TODO: remove once the provisioner works on darwin
+	_, darwinQemuEnvIsSet := os.LookupEnv("DARWIN_QEMU")
```
smira, line 21: I think we can skip this for now, as the code panics on Darwin, it's good enough for `main` (as we're WIP on adding Darwin support)

smira (approved):
👍 let's make sure the CI pipeline is green (looks like some flakes right now)

(you can do `git commit --amend` and force-push to re-trigger full run)

Orzelius:
I've repushed the commit with the `DARWIN_QEMU` env var thingy removed. The checks seem to pass

PR #10824, opened 2025-04-30
fix: relax etcd APIs RBAC requirements

### LINKED ISSUE #10821: Add etcd status API access to `os:etcd:backup` role
## Feature Request

Extend the `os:etcd:backup` role to allow access to the etcd status API.

### Description

Since the etcd snapshot command is a blocking one, it is prudent to check the etcd status and ensure the snapshot is taken from a follower node. Currently at least `os:operator` role permissions are required for this but it feels like it is something (read-only) that the `os:etcd:backup` role should be able to do too.

smira:
Allow more read-only APIs to be used with `os:reader`/`os:etcd:backup`.

Fixes #10821


frezbo (approved):

utkuozdemir (approved):

TimJones (approved):
:heart:

PR #10827, opened 2025-04-30
split the `cluster create` command into it's own package

Orzelius:
Part of #10289

This is the first change in the cluster create command refactor project. The changes will be split into small parts to make review and catching bugs easier.

The command has to be imported in cmd/root.go because it can't be referenced in the mgmt/cluster package as that would create a cyclic import.
This can be fixed later by separating the logic between a third package.

I split this into two commits so the diff is easier to review. Please see 8b2b46a63e32f7ebe8c11b39aab60ec77d153f5b for the logic changes.

### DIFF
--- a/pkg/provision/providers/factory.go
+++ b/pkg/provision/providers/factory.go
@@ -12,14 +12,35 @@ import (
 	"github.com/siderolabs/talos/pkg/provision/providers/docker"
 )
 
+const (
+	// QemuProviderName is the name of the qemu provider.
+	QemuProviderName = "qemu"
+	// DockerProviderName is the name of the docker provider.
+	DockerProviderName = "docker"
+)
+
 // Factory instantiates provision provider by name.
 func Factory(ctx context.Context, name string) (provision.Provisioner, error) {
-	switch name {
-	case "docker":
-		return docker.NewProvisioner(ctx)
-	case "qemu":
-		return newQemu(ctx)
-	default:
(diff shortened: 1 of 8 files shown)


pkg/provision/providers/factory.go
```diff
@@ -12,14 +12,34 @@ import (
...
+	case DockerProviderName:
 		return docker.NewProvisioner(ctx)
-	case "qemu":
+	case QemuProviderName:
 		return newQemu(ctx)
-	default:
-		return nil, fmt.Errorf("unsupported provisioner %q", name)
 	}
+
+	return nil, nil
```
smira, line 35: should it still return an error/panic here?

pkg/provision/providers/factory.go
```diff
@@ -12,14 +12,34 @@ import (
...
+}
+
+// IsValidProvider returns an error if the passed provider doesn't exist.
+func IsValidProvider(name string) error {
+	if name != QemuProviderName && name != DockerProviderName {
```
frezbo, line 40: easier to use a switch statement here

Orzelius:
That you for the feedback! I've addressed the comments and have squashed the commits for your convenience.

Orzelius:
Ah it seems I have lint issues. I need to get my lint running again, it doesn't seem to work after the v2 update

Orzelius:
should be good now, sorry for the trouble!

PR #10844, opened 2025-04-30
docs: add registryd debug command

rothgar:
Additional command for registry troubleshooting.

#10841


frezbo (approved):

smira:
NB: should be copied to v1.11 docs before merging

PR #10849, opened 2025-05-01
chore(ci): add extensions test for Youki runtime

frezbo:
Add extensions test for Youki container runtime.


Orzelius (approved):
youki can mean both "a container" and "cheerful" in Japanese depending on the symbols used to write it. What a clever name

smira (approved):

PR #10850, opened 2025-05-01
chore: cross platform qemu preflight checks

Orzelius:
* split platform specific preflight checks into platform specific files
* error on darwin if target is amd64
* error on darwin if host platform is amd64
* add darwin specific var-file and pflash file locations
* remove hard kvm requirement (so linux logic can be tested on machines without kvm)

### DIFF
--- a/pkg/provision/providers/qemu/arch_linux.go
+++ b/pkg/provision/providers/qemu/arch_linux.go
@@ -1,0 +1,18 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package qemu
+
+import "runtime"
+
+const accelerator = "kvm"
+
+func (arch Arch) acceleratorAvailable() bool {
+	if err := checkKVM(); err != nil {
+		return false
+	}
+
+	// kvm only supports emulating native architectures
+	return string(arch) == runtime.GOARCH
+}
--- a/pkg/provision/providers/qemu/create.go
+++ b/pkg/provision/providers/qemu/create.go
@@ -27,8 +27,8 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
 	}
 
 	arch := Arch(options.TargetArch)
-	if !arch.Valid() {
-		return nil, fmt.Errorf("unsupported arch: %q", options.TargetArch)
+	if err := arch.Valid(); err != nil {
+		return nil, err
 	}
 
 	if err := p.preflightChecks(ctx, request, options, arch); err != nil {
--- a/pkg/provision/providers/qemu/preflight_linux.go
+++ b/pkg/provision/providers/qemu/preflight_linux.go
@@ -1,0 +1,154 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package qemu
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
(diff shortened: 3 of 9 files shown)


pkg/provision/providers/qemu/arch_linux.go
```diff
@@ -0,0 +1,19 @@
...
+
+const accelerator = "kvm"
+
+func (arch Arch) acceleratorAvailable() bool {
+	err := checkKVM()
+	if err != nil {
+		return false
+	}
+
+	// kvm only supports emulating native architevtures
```
frezbo, line 17:
```suggestion
	// kvm only supports emulating native architectures
```

pkg/provision/providers/qemu/arch_linux.go
```diff
@@ -0,0 +1,19 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package qemu
+
+import "runtime"
+
+const accelerator = "kvm"
+
+func (arch Arch) acceleratorAvailable() bool {
+	err := checkKVM()
```
frezbo, line 12:
```suggestion
	if err := checkKVM(); err != nil {
	    return false
	}
```

pkg/provision/providers/qemu/create.go
```diff
@@ -27,7 +27,7 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
 	}
 
 	arch := Arch(options.TargetArch)
-	if !arch.Valid() {
+	if err := arch.Valid(); err != nil {
```
frezbo, line 30: the returned error message is not used at all, so why change the signature to return an error

pkg/provision/providers/qemu/preflight_linux.go
```diff
@@ -0,0 +1,154 @@
...
+	if check.options.TPM2Enabled {
+		if _, err := exec.LookPath("swtpm"); err != nil {
+			return fmt.Errorf("swtpm not found in PATH, please install swtpm-tools with the package manager: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (check *preflightCheckContext) checkKVM(ctx context.Context) error {
```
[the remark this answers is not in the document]
frezbo: or at least just throw a warning and not error

pkg/provision/providers/qemu/preflight_linux.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, cniDir := range cniDirs {
+		st, err := os.Stat(cniDir)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("error checking CNI directory %q: %w", cniDir, err)
+			}
+
+			fmt.Fprintf(check.options.LogWriter, "creating %q\n", cniDir)
+
+			err = os.MkdirAll(cniDir, 0o777)
```
frezbo, line 51: seems weird a pre-flight check is creating directories, I guess we can move this out somewhere else just before creating the actual cluster, not for this PR

frezbo (approved):

pkg/provision/providers/qemu/preflight_linux.go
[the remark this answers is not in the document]
Orzelius: > do we need this check now, since we optionally support acceleration if kvm available?

I think it's still worth to warn the user as the actual performance of the machines will be extremely slow without kvm

> or at least just throw a warning and not error

👍  This is exactly what I did if you see the code below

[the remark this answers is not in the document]
frezbo: oh I missed that, :+1:

pkg/provision/providers/qemu/create.go
Orzelius: Ah nice catch. I just forgot to pass the error here 🤦

smira (approved):

PR #10854, opened 2025-05-01
release(v1.11.0-alpha.0): prepare release

smira:
This is the official v1.11.0-alpha.0 release.


hack/fdspy/fdspy
frezbo, line 1: what's this?

smira: 💩 I'll remove

smira: it was "file descriptor spy" for fd leak effort

frezbo (approved):

Issue #10537, opened 2025-03-16
MacOS Qemu Provider


Orzelius:
## MacOS Qemu Provider

Add native qemu support for MacOs.

### Motivation

* Allows users to test out Talos on their machines locally with qemu. Talos running on Qemu is faster and more similar to production than Talos running in Docker. This would also allow the development and testing of various CI/CD workflows that utilize the Cluster provision features locally on Apple machines
* Would make contribution to the Talos project more approachable by removing the requirement to have a linux machine/dev-server
  * Allows the development of Talos itself locally on Apple machines.
  * Allows the development of Omni locally on Apple machines

### Technical implementation

I have a working POC which I've been using for further Talos and omni development on my Mac machine without problems for a couple of weeks here https://github.com/Orzelius/talos/pull/6. Please see the PR for more technical details for the implementation.

Currently the branch is based on https://github.com/siderolabs/talos/pull/10334, but if #10334 it gets rejected this can still be merged with minimal changes

Feel free to test out the POC and please let me know what you think!
Thank you in advance

smira:
Before we move into the actual implementation, we should make a plan about what are the features of OS X we use for it (i.e. virtualization framework), and how do we provide networking. This should be compatible with the way `talosctl cluster create` works to provide actual feature parity. E.g. we should be able to assign IPs to the VMs, provide connectivity between VMs, connectivity to host workloads (i.e. registry mirrors) and connectivity to the Internet.

Once those questions are answered, we can take incremental approach to implement this, based on discussions.

Orzelius:
> Before we move into the actual implementation, we should make a plan about what are the features of OS X we use for it (i.e. virtualization framework), and how do we provide networking.

Yep, completely agree. This is why I made this issue first instead of just submitting the PR.

In my POC I went with the qemu [HVF](https://wiki.qemu.org/Features/HVF) accelerator which uses the [Apple Hypervisor.framework](https://developer.apple.com/documentation/hypervisor). This seems to be the  fastest and most modern solution. The downside is that different cpu architectures can't be emulated.

The nodes run in vmnet shared mode which allows traffic originating from the vmnet interface to reach the Internet through NAT. The vmnet interface can communicate with the host and with other shared mode interfaces on the same subnet.
DHCP is handled via bootpd (apple's native dhcp server). Nodes have to be made sequentially. Node has to be assigned an ip before the following nodes are created to assure correct IPs. This process is a little bit of a hassle, but as far as I can tell this is the only option.

[...]

PR #10862, opened 2025-05-02
chore: calculate vmnet interface name preemptively

Orzelius:
part of #10537

this will be used for the dhcpd server assignment later

next step will be making the loadbalancer work

### DIFF
--- a/pkg/provision/providers/vm/network.go
+++ b/pkg/provision/providers/vm/network.go
@@ -1,0 +1,43 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package vm
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"sync"
+)
+
+var vmnetInterfaceRegex = sync.OnceValue(func() *regexp.Regexp { return regexp.MustCompile(`\bbridge(1\d\d)\b`) })
+
+// GetVmnetInterfaceName returns the name of the interface that will be assigned to the next vmnet interface.
+// The name is assigned incrementing by one, starting from "bridge100".
+func GetVmnetInterfaceName(allCurrentInterfaces []string) (string, error) {
+	vmnetInterfaceFound := false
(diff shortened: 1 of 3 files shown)


pkg/provision/providers/vm/network.go
```diff
@@ -0,0 +1,41 @@
...
+	"regexp"
+	"strconv"
+)
+
+// getVmnetInterfaceName returns the name of the interface that will be assigned to the next vmnet interface.
+// The name is assigned incrementing by one, starting from "bridge100".
+func getVmnetInterfaceName(allCurrentInterfaces []string) (string, error) {
+	vmnetInterfaceFound := false
+	largestVmnetIfIndex := 100
+	vmnetInterfaceRegex := regexp.MustCompile(`\bbridge(1\d\d)\b`)
```
frezbo, line 18: let's move this into a `var` declaration so it doesn't get compiled every time this function is called

frezbo: we could also now write a test for this

frezbo: nvm, saw the test below

pkg/provision/providers/vm/network_test.go
```diff
@@ -0,0 +1,31 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package vm //nolint:testpackage
```
frezbo, line 5: we can make the function public and name this `vm_test`

pkg/provision/providers/vm/network.go
smira: the good practice for regex:

```go
// global var
var vmnetInterfaceRegex = sync.OnceValue(func() *regex.Regex { return regexp.MustCompile(...) } )
```

this ensures it's compiled once and not compiled until first fused

smira (approved):

pkg/provision/providers/vm/network.go
Orzelius: Ah genius! Thank you both!

Issue #10847, opened 2025-04-30
bootloop after upgrade to 1.10 (boots OK on 1.9.5)


woodshoes:
## Bug Report
Unable to secureboot on 1.10 - works fine on 1.9.5

### Description

After running 
talosctl upgrade --talosconfig ./talos/talosconfig --nodes talos02 --image factory.talos.dev/installer-secureboot/36cd6536eaec8ba802be2d38974108359069cedba8857302f69792b26b87c010:v1.10.0 --preserve

on my HP EliteDesk 800 G2 i5-6500T 
It starts bootlooping shortly into the boot.  See attached pic.  

Reverting back to 1.9.5 (installed w same schematic "36cd6536eaec8ba802be2d38974108359069cedba8857302f69792b26b87c010") then it boots fine.  

Looks like:  

[talos] [initramfs] failed to see from the TPM: error getting TPM capabilities: unrecognized error code (0xa)
[talos] [initramfs] recovered from: early boot failed: failed to extend PCR 11 with enter-initrd: unrecognized error code (oxa)

[talos] [initramfs] rebooting in 10 seconds

The upgrade works fine on my other nodes in the cluster - different model though "HP EliteDesk 800 G4 i7-8500T"
### Logs

Unable to pull logs, system does not fully boot.  See attached pic.

![Image](https://github.com/user-attachments/assets/94797c38-8eb2-47a1-8fa3-18cf99bdd29a)

### Environment

- Talos version: (working output prior to upgrade)
Server:
        NODE:        talos02
        Tag:         v1.9.5
        SHA:         d07f6daa
        Built:
        Go version:  go1.23.7
        OS/Arch:     linux/amd64
        Enabled:     RBAC
- Kubernetes version: Server Version: v1.32.3
- talosctl -n talos02 get securitystate
NODE      NAMESPACE   TYPE            ID              VERSION   SECUREBOOT   UKISIGNINGKEYFINGERPRINT   PCRSIGNINGKEYFINGERPRINT
talos02   runtime     SecurityState   securitystate   1         true                                    9C:42:05:91:48:E1:57:A0:30:F5:ED:C5:1B:D4:96:7A:2A:3B:1B:C6:4C:DD:48:94:1A:3E:CE:3C:3F:DC:03:2F
- 
- Platform:

tholinka:
I was running into this as well, I switched from platform TPM to firmware TPM in the bios and my node started.

woodshoes:
[...]

PR #10888, opened 2025-05-05
fix: skip PCR extension if TPM1.2 is found

frezbo:
When extending PCR or trying to seed entropy pool from TPM if the found device is a TPM1.2 device, skip it, since Talos only supports TPM2.0

Fixes: #10847

### DIFF
--- a/internal/pkg/rng/tpm.go
+++ b/internal/pkg/rng/tpm.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/google/go-tpm/tpm2"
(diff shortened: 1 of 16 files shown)


internal/pkg/rng/tpm.go
```diff
@@ -22,8 +21,8 @@ import (
 func TPMSeed() error {
 	t, err := tpm.Open()
 	if err != nil {
-		// if the TPM is not available or not a TPM 2.0, we can skip the PCR extension
```
frezbo, line 25: this error is now not returned anymore by the library

internal/pkg/secureboot/tpm2/pcr.go
```diff
@@ -79,6 +78,20 @@ func PCRExtend(pcr int, data []byte) error {
 
 	defer t.Close() //nolint:errcheck
 
+	// now we need to check if the TPM is a 2.0 device
+	// we can do this by checking the manufacturer,
+	// if it fails, we can skip the PCR extension
```
frezbo, line 83: this is the actual fix, other are just cosmetic changes and for allowing to test the scenario in CI

frezbo:
@smira I hope this doesn't affect how pcr status controller works

PR #10898, opened 2025-05-05
fix: replace downloaded asset paths correctly in cluster create cmd

utkuozdemir:
This is a recently introduced regression during the refactoring - the downloaded asset URLs were set on the copy struct instead of the original one, causing the command to pass the URL as-is to qemu, resulting in QEMU failing to find the kernel to boot.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/create.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/create.go
@@ -74,8 +74,12 @@ const (
 	vipOffset = 50
 )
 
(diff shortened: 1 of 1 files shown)


Unix4ever (approved):

smira (approved):

Orzelius (approved):
Ah, my bad. Thanks for the quick fix!

utkuozdemir:
> Ah, my bad. Thanks for the quick fix!

No worries, happens. Good hint that we need to make this code unit-test friendly and add some tests.

Orzelius:
> Good hint that we need to make this code unit-test friendly and add some tests.

That's exactly what I'm trying to accomplish with the refactor 👍

PR #10900, opened 2025-05-05
chore: make qemu lb address bind work on darwin

Orzelius:
part of https://github.com/siderolabs/talos/issues/10537

The loadbalancer is bound to all interfaces on macos. The bridge interface address is not used as the bridge is not yet created at this stage. Multiple loadbalancers can be assigned via different ports.

### DIFF
--- a/pkg/provision/providers/vm/loadbalancer.go
+++ b/pkg/provision/providers/vm/loadbalancer.go
@@ -38,7 +38,7 @@ func (p *Provisioner) CreateLoadBalancer(state *State, clusterReq provision.Clus
 
 	args := []string{
 		"loadbalancer-launch",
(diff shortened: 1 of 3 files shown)


pkg/provision/providers/vm/loadbalancer _linux.go
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
frezbo, line 1: also doesn't this file need go buildtags?

Orzelius: Nope, the following suffixes are picked up from the file names automatically
```
*_GOOS
*_GOARCH
*_GOOS_GOARCH
```
https://pkg.go.dev/cmd/go#hdr-Build_constraints

smira (approved):

PR #10906, opened 2025-05-05
fix: deny apply config requests without v1alpha1 in "normal" mode

### LINKED ISSUE #10897: UX: forbid removing v1alpha1 config via apply-config API
While it's technically acceptable to remove `v1alpha1` config today, but it leads to unexpected results right now.

People often miss the `patch mc` vs. `apply-config` and end up bricking their machines.

Return an error if apply-config would result in v1alpha1 config being removed (if it previous was on the machine, and now it's being removed).

smira:
In maintenance mode, we still accept any config.

Fixes #10897

As "normal" mode requires v1alpha1 config today, it should be an easy fix to require it part of the applied config always.

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -180,6 +180,13 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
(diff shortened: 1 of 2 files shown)


smira:
```shell
$ talosctl -n 172.20.0.2 apply-config -f ephemeral-nvme.yaml
error applying new configuration: 1 error occurred:
        * 172.20.0.2: rpc error: code = InvalidArgument desc = the applied machine configuration doesn't contain v1alpha1 config, did you mean to patch the machine config instead?
```

frezbo (approved):

PR #10913, opened 2025-05-06
docs: remove azure image gallery instructions

rsmitty:
This PR removes the references to azure image gallery, as we're stopping the publishing of assets there.


frezbo (approved):

ruudbleeker:
@rsmitty may I ask why the images were removed from the Azure Community Gallery? My company relied on these images for a Managed Application in the Azure Marketplace. I need to update a bunch of ARM templates now and find a new way to image the OS disk of our Azure VMs that are part of this Managed Application.

smira:
> @smira we are not a paying customer of Sidero Labs, no. We are just a big fan of open source software in general and Talos Linux in particular and were using the Talos images as made publicly available through the Azure Community Gallery as a base for the Kubernetes cluster that we run our software solution on.

that makes sense, thanks for clarifying.

Unfortunately, the cost of Azure Image Gallery was too high for us as a company, and we never had any feedback on it being useful to anyone. You can still upload images to your own Image Gallery.

ruudbleeker:
@smira thank you for your quick response. I had no idea that there was a cost involved having your images published through the Azure Community Gallery. I'm finding that out now as I am indeed looking into uploading our own Talos image to the Gallery. 

Thankfully we only need to publish one, specific to our needs. Once the initial bootstrap of the cluster is done we have our own software solution to manage the cluster which updates it with images directly from Sidero Labs' Image Factory when needed.

ruudbleeker:
Apologies for posting yet another message to an otherwise closed issue, especially after 2 weeks have gone by again. I've enjoyed a much needed vacation in the meantime. 

I was wondering if you guys from Sidero Labs have any automation laying around to create the Azure Image Gallery that I could re-use. Considering it's an Azure thing it's probably in the form of an ARM template. I could probably create it myself, but it would be quicker and easier to have a good starting point.

Thanks in advance for anything you can share.

frezbo:
> Apologies for posting yet another message to an otherwise closed issue, especially after 2 weeks have gone by again. I've enjoyed a much needed vacation in the meantime.
> 
> I was wondering if you guys from Sidero Labs have any automation laying around to create the Azure Image Gallery that I could re-use. Considering it's an Azure thing it's probably in the form of an ARM template. I could probably create it myself, but it would be quicker and easier to have a good starting point.
> 
> Thanks in advance for anything you can share.

You can see the code here: https://github.com/siderolabs/talos/blob/release-1.9/hack/cloud-image-uploader/azure.go

PR #10972, opened 2025-05-08
feat(machined): support compressed kernel modules

jfroy:
# Pull Request

## What? (description)

Use a custom kmod InitFunc that just uses `finit_module(2)` and always passes the `MODULE_INIT_COMPRESSED_FILE` flag.

This requires a kernel with the `CONFIG_MODULE_DECOMPRESS` option enabled.

## Why? (reasoning)

Smaller images!

## Acceptance

I have tested this end-to-end with a custom Talos build (from the grounds up -- custom kernel, pkgs, extensions, Talos), with zstd compressed modules.

Additionally requires editing the modules list in hack.

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/kernel_module_spec.go
+++ b/internal/app/machined/pkg/controllers/runtime/kernel_module_spec.go
@@ -8,12 +8,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/safe"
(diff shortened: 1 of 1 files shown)


frezbo:
I'm not sure if this makes sense, the initramfs in already compressed, so modules also get compressed, same when bringing modules as extensions, so they are anyways compressed in initramfs, so this wouldn't help reduce Talos size at all

jfroy:
> I'm not sure if this makes sense, the initramfs in already compressed, so modules also get compressed, same when bringing modules as extensions, so they are anyways compressed in initramfs, so this wouldn't help reduce Talos size at all

Duh, that makes sense. It might only help a tiny bit build time if copies are involved, but probably not worth it.

PR #10991, opened 2025-05-12
chore: make qemu config server bind work on darwin

Orzelius:
A part of https://github.com/siderolabs/talos/issues/10537

This very is simuilar to ab7e693. The api is bound to all interfaces on macos. The bridge interface address is not used as the bridge is not yet created at this stage.

### DIFF
--- a/pkg/provision/providers/qemu/node_darwin.go
+++ b/pkg/provision/providers/qemu/node_darwin.go
@@ -1,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package qemu
+
+import (
+	"net"
+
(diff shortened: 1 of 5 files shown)


pkg/provision/providers/qemu/node_darwin.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"net"
+
+	"github.com/siderolabs/talos/pkg/provision"
+)
+
+// findAPIBindAddrs returns the 0.0.0.0 address to bind to all interfaces on macos with a random port on macos.
+// The bridge interface address is not used as the bridge is not yet created at this stage.
+func (p *provisioner) findAPIBindAddrs(_ provision.ClusterRequest) (*net.TCPAddr, error) {
```
frezbo, line 15: i think we should just return a string here and create the listener after this is called, also we're returning a `Close()` wouldn't that already close the listener? :thinking:

frezbo: ahh I see, so just a way to get random port

PR #11000, opened 2025-05-13
chore: dump qemu pachine ipam records on darwin

Orzelius:
* rename `config.vmMAC` -> `config.VMMAC` so it can be marshaled and passed to the launch process
* implement withNetworkContext which just runs the function on the host network on darwin


smira:
unit-test failure is a flake I can't figure out yet

smira (approved):

Unix4ever (approved):

PR #11001, opened 2025-05-13
feat: increase boot partition to 2 GiB

smira:
See https://github.com/siderolabs/talos/discussions/10994


pkg/machinery/imager/quirks/partitions.go
```diff
@@ -0,0 +1,53 @@
...
+	return 100 * mib
+}
+
+// GrubBIOSSize return BIOS GRUB partition size.
+func (p PartitionSizes) GrubBIOSSize() uint64 {
+	return 1 * mib
+}
+
+// GrubBootSize return boot partition size for GRUB layout.
+func (p PartitionSizes) GrubBootSize() uint64 {
```
frezbo, line 28: wouldn't this return zero?

frezbo: okay I see it handled in `quirks.go`

frezbo (approved):

PR #11015, opened 2025-05-14
chore: add cross-platform copatible disk probing to qemu

Orzelius:
previous implementation only worked on linux due to relying on various unix systems here https://github.com/siderolabs/go-blockdevice/blob/v2/blkid/blkid_linux.go

Utilize the https://github.com/diskfs/go-diskfs package


go.mod
```diff
@@ -246,9 +246,12 @@ require (
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.6 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/diskfs/go-diskfs v1.6.0 // indirect
```
frezbo, line 249: why is this indirect? missed a `go mod tidy`?

Orzelius: Ah yes, you're exactly right 🤦

pkg/provision/providers/qemu/launch.go
```diff
@@ -461,3 +462,32 @@ func waitForFileToExist(path string, timeout time.Duration) error {
 		time.Sleep(100 * time.Millisecond)
 	}
 }
+
+func checkPartitions(config *LaunchConfig) (diskBootable bool, err error) {
+	f, err := os.Open(config.DiskPaths[0])
```
frezbo, line 467: This `f` is not used elsewhere

pkg/provision/providers/qemu/launch.go
```diff
@@ -461,3 +462,32 @@ func waitForFileToExist(path string, timeout time.Duration) error {
 		time.Sleep(100 * time.Millisecond)
 	}
 }
+
+func checkPartitions(config *LaunchConfig) (diskBootable bool, err error) {
```
frezbo, line 466: probably rename to `checkPartition` and pass in a string since this is checking only a string

frezbo (approved):

pkg/provision/providers/qemu/launch.go
Orzelius: Yep you're right. Below, I was also leaving the closing of the disk to the garbage collector which is not the smartest way to close an open file.

pkg/provision/providers/qemu/launch.go
Orzelius: Renamed 👍  The function is utilizing both `config.DiskPaths` and `config.DiskBlockSizes` so I think it makes sense to keep the config as a param

pkg/provision/providers/qemu/launch.go
```diff
@@ -505,3 +505,27 @@ func dumpIpam(config LaunchConfig) error {
 
 	return nil
 }
+
+func checkBootDiskPartition(config *LaunchConfig) (diskBootable bool, err error) {
```
smira, line 509: I would prefer not to introduce yet another dependency into the mix, and instead make `go-blockdevice` work good enough on Darwin. It should be easy, as we are probing a file, not a real block device

Orzelius:
closing in favor of a new pr with the native implementation in go-blockdevice library

Issue #10965, opened 2025-05-08
better detection of SELinux


frezbo:
## Bug Report

On clusters that have `security=none` set upgrades to 1.10 would fail since setting `selinux=1` won't enable SELinux. Fix by checking for the existence of `/sys/fs/selinux`

PR #11018, opened 2025-05-14
fix: selinux detection

frezbo:
Fix SELinux detection.

Fixes: #10965

### DIFF
--- a/internal/pkg/selinux/selinux.go
+++ b/internal/pkg/selinux/selinux.go
@@ -34,7 +34,13 @@ var IsEnabled = sync.OnceValue(func() bool {
 
 	val := procfs.ProcCmdline().Get(constants.KernelParamSELinux).First()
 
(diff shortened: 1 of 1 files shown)


internal/pkg/selinux/selinux.go
```diff
@@ -34,7 +34,13 @@ var IsEnabled = sync.OnceValue(func() bool {
 
 	val := procfs.ProcCmdline().Get(constants.KernelParamSELinux).First()
 
-	return val != nil && *val == "1"
+	var selinuxFSPresent bool
+
+	if _, err := os.Stat("/sys/fs/selinux"); err == nil {
```
smira, line 39: just checking, does it depend only on `sysfs` being mounted?

frezbo: yeh, I checked an upgrade path from 1.9 with `security=none` set

Issue #10921, opened 2025-05-07
containerd 2.1: use transfer service to pull images


smira:
If we have feature parity (need to research) with our current pull method (via containerd API), use the transfer service.

See https://github.com/containerd/containerd/releases/tag/v2.1.0-rc.1

See https://github.com/containerd/containerd/pull/8515

PR #11024, opened 2025-05-15
feat: update containerd to 2.1.0

smira:
Update configuration.

See also https://github.com/siderolabs/talos/issues/10921


smira:
❓  https://github.com/containerd/containerd/blob/dc795bfc1f3fec423e7c5b7d243f9600eb744453/internal/cri/server/images/image_pull.go#L182

smira:
> ❓ [containerd/containerd@`dc795bf`/internal/cri/server/images/image_pull.go#L182](https://github.com/containerd/containerd/blob/dc795bfc1f3fec423e7c5b7d243f9600eb744453/internal/cri/server/images/image_pull.go#L182)

I wonder if we should first disable pull via the transfer service for now, and keep local image pull for now (?)

frezbo:
> > ❓ [containerd/containerd@`dc795bf`/internal/cri/server/images/image_pull.go#L182](https://github.com/containerd/containerd/blob/dc795bfc1f3fec423e7c5b7d243f9600eb744453/internal/cri/server/images/image_pull.go#L182)
> 
> I wonder if we should first disable pull via the transfer service for now, and keep local image pull for now (?)

I guess once the TODO are done we can use this

PR #11051, opened 2025-05-19
test: fix DHCP unicast failures in QEMU environment

smira:
The fix is based on the https://github.com/kubevirt/kubevirt/pull/1766, and for the background see
https://github.com/gregoryolsen/lxc-devuan/issues/1.

Note this is not an issue in Talos itself, but only for QEMU VMs created with `talosctl cluster create`.


smira:
I don't think I understand why it affects _only_ unicast packets

PR #11066, opened 2025-05-21
chore: add vmnet support (darwin/qemu)

Orzelius:
* split the networking config into common and platform specific structs
* add logic to get the config file server for darwin
* generate a random mac address for the darwin qemu machines
* generate the darwin specific `-netdev` qemu params

With this change the qemu machines are successfully started, but DHCPd is still todo, so they get wrong IPs assigned.

example of the generated netdev params:
`-netdev vmnet-shared,id=net0,start-address=10.5.0.1,end-address=10.5.0.3,subnet-mask=255.255.255.0 -device virtio-net-pci,netdev=net0,mac=7e:96:ad:7c:0a:71 `

part of #10537

### DIFF
--- a/pkg/provision/providers/qemu/launch.go
+++ b/pkg/provision/providers/qemu/launch.go
@@ -18,11 +18,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/containernetworking/cni/libcni"
 	"github.com/google/uuid"
 	"github.com/siderolabs/go-blockdevice/v2/blkid"
 
-	"github.com/siderolabs/talos/pkg/provision"
 	"github.com/siderolabs/talos/pkg/provision/providers/vm"
 )
 
@@ -57,18 +55,6 @@ type LaunchConfig struct {
 	// Talos config
 	Config string
 
-	// Network
-	BridgeName        string
-	NetworkConfig     *libcni.NetworkConfigList
-	CNI               provision.CNIConfig
-	IPs               []netip.Addr
-	CIDRs             []netip.Prefix
-	NoMasqueradeCIDRs []netip.Prefix
-	Hostname          string
-	GatewayAddrs      []netip.Addr
-	MTU               int
-	Nameservers       []netip.Addr
-
 	// PXE
 	TFTPServer       string
 	BootFilename     string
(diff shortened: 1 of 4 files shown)


pkg/provision/providers/qemu/launch_darwin.go
```diff
@@ -6,9 +6,60 @@ package qemu
...
+	subnetMask := fmt.Sprintf("%d.%d.%d.%d", m[0], m[1], m[2], m[3])
+	netDevArg += fmt.Sprintf(",start-address=%s,end-address=%s,subnet-mask=%s", networkConfig.StartAddr, networkConfig.EndAddr, subnetMask)
+
+	return netDevArg
+}
+
+// getConfigServerAddr returns the ip accessible to the VM that will route to the config server.
+// hostAddrs is the address on which the server is accessible from the host network.
+func getConfigServerAddr(hostAddrs net.Addr, config LaunchConfig) (net.Addr, error) {
+	split := strings.Split(hostAddrs.String(), ":")
```
smira, line 51: this way doesn't work for IPv6 addresses which look like `[2000::1]:4000`, so please use https://pkg.go.dev/net#SplitHostPort

but in general, there's a better experience with https://pkg.go.dev/net/netip#AddrPort, so I would suggest to use it here, as we're dealing with IP and port, not a random string.

It looks like all we do is just replace the address

pkg/provision/providers/qemu/launch.go
```diff
@@ -81,10 +67,10 @@ type LaunchConfig struct {
 	sdStubExtraCmdline       string
 	sdStubExtraCmdlineConfig string
 
-	// filled by CNI invocation
-	tapName string
-	VMMac   string
-	nsPath  string
+	// platform specific Network configuration
+	Network networkConfig
```
smira, line 71: this is just a suggesion, but we could probably make the diff smaller if simply embed the struct here, i.e. 
```suggestion
   networkConfig
```

pkg/provision/providers/qemu/node.go
```diff
@@ -368,3 +365,30 @@ func (p *provisioner) createMetalConfigISO(state *vm.State, nodeName, config str
...
+		local     = 0b10
+		multicast = 0b1
+	)
+
+	buf := make([]byte, 6)
+	rand.Read(buf) //nolint:errcheck
+	// clear multicast bit (&^), ensure local bit (|)
+	buf[0] = buf[0]&^multicast | local
+
+	return fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x", buf[0], buf[1], buf[2], buf[3], buf[4], buf[5])
```
smira, line 393: https://pkg.go.dev/net#HardwareAddr

```suggestion
	return net.HardwareAddr(buf).String()
```

smira (approved):

pkg/provision/providers/qemu/launch.go
Orzelius: I kinda like it as is because the struct that has platform specific types is clearly separated from the rest

Issue #11074, opened 2025-05-22
Talos iso is not reproducible anymore


frezbo:
## Bug Report

Non-secureboot iso's should be reproducible

Also validate and add tests to catch these early on

PR #11079, opened 2025-05-22
fix: metal-iso reproducibility

frezbo:
Fix `metal-iso` reproducibility.

Fixes: #11074


Makefile
```diff
@@ -728,6 +728,13 @@ reproducibility-test-local-%:
 	@find _out1/ -type f | xargs -IFILE diffoscope FILE `echo FILE | sed 's/_out1/_out2/'`
 	@rm -rf _out1/ _out2/
 
+reproducibility-test-iso:
```
smira, line 731: looks like it's not being called?

frezbo: ahh forgot it

frezbo: fixed

PR #11082, opened 2025-05-23
chore: use custom dhcpd server on macos qemu

Orzelius:
add usage of our in-house dhcpd server instead of apple's bootpd

part of #10537 

There's stuff to be ironed out, but with this change basic cluster creation should work on darwin as well.

### DIFF
--- a/pkg/provision/providers/qemu/create.go
+++ b/pkg/provision/providers/qemu/create.go
@@ -86,18 +86,19 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
 		}
 	}
 
-	fmt.Fprintln(options.LogWriter, "creating dhcpd")
-
-	if err = p.CreateDHCPd(state, request); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/provision/providers/qemu/create.go
```diff
@@ -100,6 +94,12 @@ func (p *provisioner) Create(ctx context.Context, request provision.ClusterReque
 		return nil, err
 	}
 
+	fmt.Fprintln(options.LogWriter, "creating dhcpd")
```
frezbo, line 97: moving this code to after creating nodes intentional?

Orzelius: Yep, as qemu creates the bridge interface the dhcpd server is attached to, at least one machine has to be created first

Orzelius: I added a comment above the line for extra clarity as it's definitely confusing at the first glance

frezbo (approved):

Issue #11133, opened 2025-06-02
disable IMA framework


smira:
It have never been properly used in Talos, more of a future development.

But it causes real performance regressions, see https://github.com/siderolabs/talos/issues/11129, and a previous one with xfs (I don't have the reference).

maelle-le-bon:
Its been months I struggle to work with seaweedFS on talos, and I just discovered this series of bug related to IMA.

I just tested Talos 1.11.0-alpha.1 and I confirm that seaweedfs mounts now work as expected. A simple `head` on a 1GB file used to be very slow, it is now working very fast.

Thank you for your fix.

PR #11137, opened 2025-06-02
feat: drop IMA support

frezbo:
Fixes: #11133


pkg/machinery/kernel/kernel.go
```diff
@@ -32,9 +32,6 @@ func DefaultArgs(quirks quirks.Quirks) []string {
 		"nvme_core.io_timeout=4294967295",
 		// Disable rate limited printk
 		"printk.devkmsg=on",
-		"ima_template=ima-ng",
```
smira, line 35: I think this should be a `quirk` so that we don't change old Talos versions kernel args?

pkg/imager/imager_test.go
```diff
@@ -116,7 +116,7 @@ func TestImager(t *testing.T) {
 				Version: "1.10.0",
 			},
 
-			expected: "talos.platform=metal console=ttyAMA0 console=tty0 init_on_alloc=1 slab_nomerge pti=on consoleblank=0 nvme_core.io_timeout=4294967295 printk.devkmsg=on ima_template=ima-ng ima_appraise=fix ima_hash=sha512 selinux=1", //nolint:lll
```
smira, line 119: specifically this test should stay unchanged for Talos < 1.11

pkg/machinery/imager/quirks/quirks.go
```diff
@@ -241,3 +241,15 @@ func (q Quirks) XFSMkfsConfig() string {
 		return "/usr/share/xfsprogs/mkfs/lts_6.1.conf"
 	}
 }
+
+var maxTalosVersionIMASupported = semver.MustParse("1.10.0")
```
smira:
```suggestion
var maxTalosVersionIMASupported = semver.MustParse("1.10.99")
```

this should be like this, otherwise the quirk won't trigger for say 1.10.1 (probably good to add a unit-test)

smira: or make it > 1.11.0-alpha.0, but idk if it's going to work without a unit-test

frezbo: that won't work since `quirks.New` only picks up major minor patch

smira: yeah, so probably my first suggestion

PR #11139, opened 2025-06-02
fix: rework the way CRI config generation is waited for

### LINKED ISSUE #11132: `machine.files` causes unsuccessful Talos node deployment (`talos_machine_bootstrap: bootstrap is not available yet`)
### Description

I ran into a weird `machine.files` bug, when trying to make our [Cilium/Tetragon Talos Linux conformance test Terraform setup](https://github.com/isovalent/terraform-aws-talos/tree/main/example) configurable with regard to Talos' containerd NRI hook configuration. Our Terraform setup is highly based on your [siderolabs/contrib](https://github.com/siderolabs/contrib/blob/main/examples/terraform/aws/main.tf) AWS example.

When I make the [Talos containerd NRI runtime hook](https://www.talos.dev/v1.10/talos-guides/configuration/containerd/#enabling-nri-plugins) configurable (via `machine.files` and `disable = ${var.disable_containerd_nri_plugins}`) and set the same value as the default (`disable = true`), it breaks the cluster setup (`talos_machine_bootstrap: rpc error: code = FailedPrecondition desc = bootstrap is not available yet`). Even without the TF variable, this breaks:
```
machine = {
  files = [
    {
      content = <<EOF
[plugins]
  [plugins."io.containerd.nri.v1.nri"]
    disable = true
EOF
      path : "/etc/cri/conf.d/20-customization.part",
      op : "create"
    }
  ]
}
```
While interestingly, when I change the value to `disable = false` (double negative, yay 😅), it works and the cluster is created normally:
```
machine = {
  files = [
    {
      content = <<EOF
[plugins]
  [plugins."io.containerd.nri.v1.nri"]
    disable = false
EOF
      path : "/etc/cri/conf.d/20-customization.part",
      op : "create"
    }
  ]
}
```

### Logs

```bash
[   64.512737] [talos] task writeUserFiles (1/1): writeUserFiles failed, rebooting in 35 minutes. You can use talosctl apply-config or talosctl edit mc to fix the issues, error:
[   64.514572] 1 error occurred:
[   64.514959]  * context deadline exceeded
```

[talos_node_startup.log](https://github.com/user-attachments/files/20547791/talos_node_startup.log)

### Environment

- Talos version: `v1.10.3`
- Talos TF provider: Tested with `0.8.1` and `v0.9.0-alpha.0`
- Kubernetes version: `1.33.1`
- Platform: AWS EC2 instances

smira:
Instead of relying on the fact that CRI patch should modify the generated final CRI config, rely on the specific checksum of the CRI patch to be included into the generated CRI config.

This also to resolve Talos hanging on boot when a CRI patch is a no-op (it doesn't change the generated config).

Fixes #11132


smira:
> Maybe an integration test too with the CRI patch that was reported

added an integration test for it, and verified that this test fails on `main` Talos

Issue #11129, opened 2025-06-01
Kernel bug: Ceph on Talos should be up to 60x faster, I submitted a patch


twelho:
**Link to patch, also confirmed by maintainers: https://lore.kernel.org/lkml/20250529174453.293618-1-twelho@welho.tech/T/#u**

After a year of profiling distributed filesystems on Talos, with a focus on Rook Ceph CSI, I finally figured out why its performance is nowhere near expectations. Mounting the CephFS filesystem on an external machine in the same network produced more than 10x faster results than on the Talos nodes themselves.

Talos enables and configures the Integrity Measurement Architecture (IMA) in the kernel, which is seemingly used for boot-time integrity measurement. Every time files are opened with `O_RDONLY` or `O_RDWR`, the default is to [measure them](https://github.com/siderolabs/talos/blob/a90c936a16756cfe5fe451258f0022b808be17d2/internal/app/machined/pkg/startup/ima.go#L50-L51) with [SHA-512](https://github.com/siderolabs/pkgs/blob/8aad53bab3201d7f87d39ab61953e04392402efc/kernel/build/config-amd64#L6504), but files on regular filesystems, including CephFS, are [correctly excluded](https://github.com/siderolabs/talos/blob/a90c936a16756cfe5fe451258f0022b808be17d2/internal/app/machined/pkg/startup/ima.go#L46).

However, the CephFS kernel driver forgets to set the `s_magic` field in its superblock, which causes the `fsmagic` for CephFS files in the IMA comparison to be zero. This makes the `dont_measure` policy a no-op and results in CephFS files getting hashed constantly, with colossal performance implications.

Tests performed on a 3x replicated Ceph v19.3.0 cluster across three i5-7200U nodes each equipped with one Micron 7400 MAX M.2 disk (BlueStore) and Gigabit ethernet, on Talos Linux v1.10.2:

![Image](https://github.com/user-attachments/assets/a5dae0e1-1f0a-4e79-be8f-4d4c92be0b9f)

![Image](https://github.com/user-attachments/assets/d705ac9c-90f5-4f76-8028-ffd05ff00839)

![Image](https://github.com/user-attachments/assets/83a73c42-0cd1-49c6-9976-26f2a37cdce3)

[IO500](https://github.com/IO500/io500) 2fcd6d6 results (benchmarks within variance omitted):

| IO500 benchmark   | 6.12.27-talos  | +twelho patch  | Speedup   |
|:------------------|:---------------|:---------------|:----------|
[...]

PR #11143, opened 2025-06-03
fix: use correct FUSE magic for IMA `fsmagic` matching

twelho:
`FUSE_CTL_SUPER_MAGIC` is the superblock magic of the [control filesystem](https://www.kernel.org/doc/html/v6.12/filesystems/fuse.html#control-filesystem), actual FUSE mount points have `FUSE_SUPER_MAGIC` in their superblock. This mismatch essentially makes the `dont_measure` rule a no-op, which causes all files under FUSE mount points to be measured with SHA-512 on every `open(2)`, equivalently to https://github.com/siderolabs/talos/issues/11129, causing major performance overhead.

As https://github.com/siderolabs/talos/issues/11133 disables IMA globally for future releases, this is only relevant for v1.10 and before, see https://github.com/siderolabs/talos/issues/11129#issuecomment-2931688318 for details.

## Acceptance


smira (approved):
thank you!

PR #11160, opened 2025-06-05
arm64 support for platform vmware

jonkerj:
# Pull Request

## What? (description)

ARM64 support for VMWare platform config/metadata provider 

## Why? (reasoning)

Talos was lacking support for this

## Acceptance

## Caveat

There is one caveat: the library does not have a non-destructive test to detect if we are a VMWare VM on ARM (it does on x86), and the hypercall causes a SIGILL when running on a different platforms. In other words, supplying `talos.platform=vmware` on an ARM non-VM will probably make Talos die due that SIGILL.


internal/app/machined/pkg/runtime/v1alpha1/platform/vmware/vmware_supported.go
```diff
@@ -13,11 +14,12 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"log/slog"
 
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/go-procfs/procfs"
-	"github.com/vmware/vmw-guestinfo/rpcvmx"
-	"github.com/vmware/vmw-guestinfo/vmcheck"
+	"github.com/siderolabs/talos-vmtoolsd/pkg/hypercall"
```
smira: https://github.com/siderolabs/talos-vmtoolsd/blob/a7fd85604772e331e43929b2b287d5e8415aafa0/go.mod#L12

`talos-vmtoolsd` has a dependency on Talos machinery, so if we do import this way, we create a circular dependency which might be hard to break if we do incompatible change on one of the side.

I would suggest to move `hypercall`/`nanotoolbox` into a separate library to break this cycle

jonkerj: Good suggestion, thanks for that. We've externalised the library as [go-vmw-guestrpc](https://github.com/equinix-ms/go-vmw-guestrpc).

jonkerj:
I externalized the hypercall/nanotoolbox into go-vmw-guestinfo, which is used in this PR and talos-vmtoolsd. I have tested (with success) that provisioning through both OVF env and extraConfig works (on AMD64, don't know how to feed these properties into VMWare Fusion), so I'm think we're good to go.

I have signed the commit using my SSH key, but conformance does not like it. Should I sign using GPG instead?

smira:
> I have signed the commit using my SSH key, but conformance does not like it. Should I sign using GPG instead?

we will re-sign before merging, please don't worry about it

smira:
@jonkerj and same here for 'edit from maintainers' please

jonkerj:
Same here, due to GH feature/bug I can't enable this on org forks. I'll close and reopen personally

jonkerj:
Followed up in #11209

PR #11167, opened 2025-06-05
docs: add example for fluentbit config

rothgar:
Added example to enable HTTP server and listen port.

### DIFF
--- a/website/content/v1.11/talos-guides/configuration/logging.md
+++ b/website/content/v1.11/talos-guides/configuration/logging.md
@@ -239,6 +239,9 @@ config:
       Daemon        Off
       Log_Level     warn
       Parsers_File  custom_parsers.conf
+      HTTP_Server   On
+      HTTP_Listen   0.0.0.0
+      HTTP_Port     2020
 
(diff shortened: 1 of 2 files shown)


website/content/v1.11/talos-guides/configuration/logging.md
```diff
@@ -239,6 +239,9 @@ config:
       Daemon        Off
       Log_Level     warn
       Parsers_File  custom_parsers.conf
+      HTTP_Server   On
```
smira, line 242: is it safe to expose this?

rothgar: It exposes metrics about fluentbit and not actual logs so I assume it would be safe and necessary if someone is scraping it with prometheus

alongwill: We use a Talos `NetworkRuleConfig` to restrict access to this port to the Pod network CIDR.

Unix4ever:
@rothgar maybe needs a rebase

Issue #9372, opened 2024-09-25
IPv6-only support / improvements


lion7:
# Bug Report

## Description
DNS not working when Talos is installed on a machine in an IPv6-only network.
I needed to do multiple customizations and workarounds which I think should not be needed.

Some things that I noticed:

### Default nameservers
**Issue**: 
The default nameservers are `1.1.1.1` and `8.8.8.8`, which cannot be reached from IPv6.

**Workaround**: 
Configure an IPv6 nameserver using the dashboard (requires a local screen/keyboard or BMC). 
Alternatively configure a DNS server using the `ip=:::::::<dns0-ip>:<dns1-ip>:<ntp0-ip>` kernel arg.

**Better solution**: 
Talos should use the IPv6 equivalent when no IPv4 address is configured on any interface.

### Nameservers announced by ND are ignored
**Issue**:
DNS servers that are announced via ND (SLAAC setup) are currently ignored by Talos.

**Workaround**: 
Override DNS servers using machine configuration.

**Better solution**:
Talso should use the announced DNS servers from ND similar to how it does for DNS servers announced via IPv4 DHCP.

### IPv6 endpoint for Image Factory
**Issue**:
Talos Image Factory (`factory.talos.dev`) has no IPv6 addresses configured.

**Workaround**:
Run a pull-through registry on a dual-stack server.

**Better solution**:
Already tracked by https://github.com/siderolabs/image-factory/issues/60

### DNS forwarding to host
**Issue**:
DNS requests are forwarded to the host using a hardcoded `169.254.116.108` IPv4 address, but there is no matching IPv4 route in a IPv6-only environment.

**Workaround**:
Disable DNS forwarding.

**Better solution**:
Also use a hardcoded IPv6 address as a fallback.

## Logs
N/A

## Output of `talosctl get routes | grep -v veth`
```
NODE                                      NAMESPACE   TYPE          ID                                                                   VERSION   DESTINATION                                   GATEWAY                     LINK      METRIC
2a02:****:****:****:d250:99ff:fefa:a836   network     RouteStatus   enp38s0/inet6//2a02:****:****:****::/64/256                          1         2a02:****:****:****::/64                                                  enp38s0   256
2a02:****:****:****:d250:99ff:fefa:a836   network     RouteStatus   enp38s0/inet6//fe80::/64/256                                         1         fe80::/64                                                                 enp38s0   256
[...]

PR #11168, opened 2025-06-06
feat(machined): use configured resolvers only instead of appending to defaults

kema-dev:
# Pull Request

## What? (description)

- Do not use default resolvers when user-configured ones are provided

## Why? (reasoning)

- User provided resolvers should override defaults. This is especially useful for IPv6-only networks where [default nameservers](https://github.com/siderolabs/talos/blob/a791bbe568bba4f9675ef820fd4ce4671a481f5a/internal/app/machined/pkg/controllers/network/resolver_config.go#L180) might be unreachable, as well as many other cases such as air-gapped environments.
- Related to #9372 

## Acceptance

- [ ] you ran conformance (`make conformance`) (used SSH key for signing, will rewrite if a mergeable fix is found)

## Notes

- This does not look sufficient for [kernel command line](https://www.talos.dev/v1.10/reference/kernel/#ip), default resolvers are still used . I must have missed something
- Further documentation on IPv6-only network would be useful

### DIFF
--- a/internal/app/machined/pkg/controllers/network/resolver_config.go
+++ b/internal/app/machined/pkg/controllers/network/resolver_config.go
@@ -225,6 +225,10 @@ func (ctrl *ResolverConfigController) parseMachineConfiguration(logger *zap.Logg
 		return spec, false
 	}
 
+	if len(resolvers) > 0 {
+		spec.DNSServers = make([]netip.Addr, 0, len(resolvers))
+	}
+
 	for _, resolver := range resolvers {
 		server, err := netip.ParseAddr(resolver)
 		if err != nil {


internal/app/machined/pkg/controllers/network/resolver_config.go
```diff
@@ -225,6 +225,10 @@ func (ctrl *ResolverConfigController) parseMachineConfiguration(logger *zap.Logg
 		return spec, false
 	}
 
+	if len(resolvers) > 0 {
+		spec.DNSServers = make([]netip.Addr, 0, len(resolvers))
```
smira, line 229: this change seems to be irrelevant, or AI slop

kema-dev: The intended goal is to remove default resolvers since user provider some, I might have missed something and would be happy to know how to

smira: You should probably start with an issue and a way to reproduce the problem, but this code change doesn't make any sense to me.

PR #11169, opened 2025-06-06
docs: developing talos on darwin

Orzelius:
* add docs about developing talos on macos
* detect arch in the makefile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -944,6 +944,8 @@ ENTRYPOINT ["/sbin/init"]
 
 # The installer target generates an image that can be used to install Talos to
 # various environments.
+
+# Make the installer binary.
 FROM base AS installer-build
 ARG GO_BUILDFLAGS
 ARG GO_LDFLAGS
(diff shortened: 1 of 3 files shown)


website/content/v1.11/advanced/developing-talos.md
```diff
@@ -64,7 +63,7 @@ bash hack/start-registry-proxies.sh
 Start your local cluster with:
 
 ```bash
-sudo --preserve-env=HOME _out/talosctl-linux-amd64 cluster create \
+sudo --preserve-env=HOME _out/talosctl-<YOUR FLAWOR> cluster create \
```
Unix4ever:
```suggestion
sudo --preserve-env=HOME _out/talosctl-<YOUR FLAVOR> cluster create \
```

website/content/v1.11/advanced/developing-talos.md
```diff
@@ -6,8 +6,7 @@ aliases:
 ---
 
 This guide outlines steps and tricks to develop Talos operating systems and related components.
-The guide assumes Linux operating system on the development host.
-Some steps might work under Mac OS X, but using Linux is highly advised.
+The guide assumes MacOS or a Linux operating system on the development host.
```
dsseng:
```suggestion
The guide assumes macOS or a Linux operating system on the development host.
```

smira: yep, we need to pick one of three ways to spell `Mac OS X`, `macOS` or `OS X`: https://en.wikipedia.org/wiki/MacOS

dsseng: The next rename is likely to be soon since macOS is a rather old name from 2016, so it could be rational to move that into a variable /s

macOS probably covers all the versions that work for Talos development though

darwin/arm64 is valid as long as it's used by Go

Dockerfile
```diff
@@ -972,36 +975,45 @@ FROM install-artifacts-${TARGETARCH} AS install-artifacts-targetarch
 
 FROM install-artifacts-${INSTALLER_ARCH} AS install-artifacts
 
+# Add the installer with a simlink as 'imager' and a /rootfs dir containing only the installer.
```
smira:
```suggestion
# Add the installer with a symlink as 'imager' and a /rootfs dir containing only the installer.
```

smira (approved):

PR #11178, opened 2025-06-09
fix: correctly predict interface name on darwin

Orzelius:
Old implementation didn't work if the interface to be created wasn't the biggest index.

For example if interfaces `bridge100` and `bridge102` already existed, vmnet would create a `bridge 101`, but the old logic expected a `bridge103`.

This also slightly simplifies the code

### DIFF
--- a/pkg/provision/providers/vm/network.go
+++ b/pkg/provision/providers/vm/network.go
@@ -5,39 +5,20 @@
 package vm
 
 import (
+	"errors"
 	"fmt"
-	"regexp"
-	"strconv"
-	"sync"
+	"slices"
 )
-
-var vmnetInterfaceRegex = sync.OnceValue(func() *regexp.Regexp { return regexp.MustCompile(`\bbridge(1\d\d)\b`) })
 
 // GetVmnetInterfaceName returns the name of the interface that will be assigned to the next vmnet interface.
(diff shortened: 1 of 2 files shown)


pkg/provision/providers/vm/network.go
```diff
@@ -16,28 +16,21 @@ var vmnetInterfaceRegex = sync.OnceValue(func() *regexp.Regexp { return regexp.M
 // GetVmnetInterfaceName returns the name of the interface that will be assigned to the next vmnet interface.
 // The name is assigned incrementing by one, starting from "bridge100".
 func GetVmnetInterfaceName(allCurrentInterfaces []string) (string, error) {
-	vmnetInterfaceFound := false
```
smira, line 19: I wonder if we can simplify this (removing regexp) to:

```go

for i := 100; i < 200; i++ {
   interfaceName := fmt.Sprintf("bridge%d")
   if slices.Index(allCurrentInterfaces, interfaceName) == -1 {
        return interfaceName, nil
   }
}

return nil, errors.New("all interface names seem to be already in use")
}
```

smira (approved):

pkg/provision/providers/vm/network.go
Orzelius: ah this is very clean, thanks!

PR #11158, opened 2025-06-04
feat: implement swap support

smira:
Fixes #10674

Provide a way to see current swap status, configure additional swap devices (block) and de-configure them on the fly.

PR #11181, opened 2025-06-09
fix: correctl close encrypted volumes

smira:
Wrong name was used to close encrypted volumes, so that the close was no-op.


Unix4ever (approved):

smira:
The test in #11158 fails without the fix, so it serves as a regression test.

PR #11199, opened 2025-06-11
feat: implement zswap support

### LINKED ISSUE #10675: Support for `zswap`

smira:
Zswap allows to compress pages in memory before they hit the actual swap device.

Both swap and zswap (or either one of these) can be enabled.

Fixes #10675


website/content/v1.11/talos-guides/configuration/swap.md
```diff
@@ -0,0 +1,79 @@
...
+This configuration enables zswap with a maximum pool size of 20% of the total system memory.
+To check the current zswap status, you can use the `talosctl get zswapstatus` command:
+
+```shell
+$ talosctl -n 172.20.0.5 get zswapstatus
+NODE         NAMESPACE   TYPE          ID      VERSION   TOTAL SIZE   STORED PAGES   WRITTEN BACK   POOL LIMIT HIT
+172.20.0.5   runtime     ZswapStatus   zswap   1         0 B          0              0              0
+```
+
+Removing a `ZswapConfig` document will disable zswap on the system.
```
smira, line 84: TODO: add a note about Kubernetes & swap, `talosctl cgroups --preset swap`

internal/app/machined/pkg/controllers/block/zswap_config.go
```diff
@@ -0,0 +1,115 @@
...
+						p.TypedSpec().Value = "N"
+					}
+
+					return nil
+				}); err != nil {
+				return fmt.Errorf("error setting zswap config: %w", err)
+			}
+		}
+
+		if err = safe.CleanupOutputs[*runtime.KernelParamSpec](ctx, r); err != nil {
```
frezbo, line 111: is cleaning up enough, or do we need to explicitly set `sys.module.zswap.parameters.enabled` to `N`?

website/content/v1.11/talos-guides/configuration/swap.md
```diff
@@ -0,0 +1,79 @@
...
+Current swap status can be checked using `talosctl get swap` command:
+
+```shell
+$ talosctl -n 172.20.0.5 get swap
+NODE         NAMESPACE   TYPE         ID               VERSION   DEVICE           SIZE      USED     PRIORITY
+172.20.0.5   runtime     SwapStatus   /dev/nvme0n2p2   1         /dev/nvme0n2p2   3.9 GiB   100 MiB  -2
+```
+
+Removing a `SwapVolumeConfig` document will remove the swap device from the system, but the partition will remain on the disk.
+To
```
frezbo: I guess the `To` is to add docs about removing the partition?

frezbo (approved):

internal/app/machined/pkg/controllers/block/zswap_config.go
smira: yes, cleaning up is enough - the kernel param controller reverts the value to the previous setting ('N') if the spec is removed

website/content/v1.11/talos-guides/configuration/swap.md
smira: yep, thanks, fixed

website/content/v1.11/talos-guides/configuration/swap.md
smira: done

PR #11200, opened 2025-06-11
docs: set apiServer.extraArgs.enable-aggregator-routing

Blarc:
If you are not running kube-proxy on a host running the API server, then you must make sure that the system is enabled with the following kube-apiserver flag -https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/.

# Pull Request

## Acceptance

### DIFF
--- a/website/content/v1.10/kubernetes-guides/network/deploying-cilium.md
+++ b/website/content/v1.10/kubernetes-guides/network/deploying-cilium.md
@@ -43,6 +43,10 @@ cluster:
   network:
     cni:
       name: none
+  apiServer:
+    extraArgs:
+      # https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/
+      enable-aggregator-routing: true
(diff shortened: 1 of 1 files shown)


smira:
But this is only applicable if you want to use aggregation layer, and most people do not use this I guess?

And those who do want to use it, should be comfortable setting this up correctly?

Blarc:
Hey @smira,

I just think it would be maybe wise to mention this somewhere. I needed to add this in order for cert-manager to start working. I'd think a lot of people use cert-manager.

smira:
> I just think it would be maybe wise to mention this somewhere. I needed to add this in order for cert-manager to start working. I'd think a lot of people use cert-manager.

Probably yes, not sure if this is the right place/context, as otherwise we end up rebuilding Cilium/cert-manager docs in Talos

smira:
Documentation moved to siderolabs/docs repository.

PR #11204, opened 2025-06-12
fix: userspace wireguard library overrides

smira:
Drop using the fork https://github.com/siderolabs/wireguard-go which had a single change to the variable.

Instead, patch the variable using Go linker flags and only for `machined`.

This fixes the issue with `wg` utility not working for `talosctl cluster create --with-siderolink` clsuters, as it was using incompatible path.

We do need to override the path for Talos only, as in Talos `/var/run` is not always available, but outside of Talos itself we never need this change at all.


smira:
So to make it more clear:

* when we use userspace Wireguard _inside_ Talos, we do need to override UAPI socket path to `/system/...`
* when we do anything outside Talos, e.g. in `talosctl`, this hurts, as makes it incompatible with CLIs like `wg`

utkuozdemir (approved):

PR #11208, opened 2025-06-12
feat: enforce Kubernetes version compatibility

### LINKED ISSUE #11198: talosctl Should not create an impossible situation to recover from (upgrade k8s past os supported version)
### Description

I have upgraded k8s on my cluster beyond the supported version by the OS.  k8s is working fine, but now I need to upgrade the OS because of a requirement in the cilium upgrade path.  Upgrading the OS requires a reboot, upgrading k8s doesn't, so because of persistent disks in my setup which require a rebuild if a node reboots, I avoid the OS upgrade unless necessary.

`talosctl upgrade-k8s` does not warn you when you're going beyond a supported OS version, but `talosctl upgrade` has a hard check you cannot override for the k8s version.

IMO Either:
- `upgrade-k8s` should perform the same check for k8s version as `upgrade` does to stop you going beyond a supported version **or**
- `upgrade` should have a method for saying "k8s that is running on that node is fine for the version we're passing through, go ahead" (I expected `--force` to do this but it does not)

#### real example:
talos 1.6.4 (probably `talosctl` 1.9.4 because I had a mix of node OS versions) happily upgraded to k8s 1.32.5 with `upgrade-k8s` (was running 1.32.2 for a while, but because I added a node with the wrong version, everything got downgraded and back because it's not possible to just upgrade one node either)
[supported upgrade path](https://www.talos.dev/v1.10/talos-guides/upgrading-talos/#supported-upgrade-paths): ->1.6.8->1.7.x->1.8.x->1.9.4 - does **not** work because the pre-flight checks run on upgrade reject the upgrade because 1.32.x is not supported on 1.6.8
direct upgrade path - straight to 1.9.4 (first version that supports 1.32.x of k8s) refuses to work because the pre-flight checks include a current version check that 1.6.4 fails.
Catch 22.

### Logs
Difficult to get because the error only happens in the dashboard and when it fails, it reboots and is gone.  Here's the OS upgrade that cannot be passed :
```
 user: warning: [2025-06-11T14:30:48.597486002Z]: pre-flight checks failed: host version 1.6.4 is too old to upgrade to Talos 1.9.6
 user: warning: [2025-06-11T14:30:48.771210002Z]: [talos] task upgrade (1/1): failed: task "upgrade" failed: exit code 1
 user: warning: [2025-06-11T14:30:48.944778002Z]: [talos] phase upgrade (10/14): failed
 user: warning: [2025-06-11T14:30:49.001256002Z]: [talos] upgrade sequence: failed
```
And here's trying to upgrade from 1.6.4 to 1.6.8 :
```
 user: warning: [2025-06-11T15:35:19.905903132Z]: pre-flight checks failed: component kubelet version issue: version of Kubernetes 1.32.5 is too new to be used with Talos 1.6.8
 user: warning: [2025-06-11T15:35:20.071014132Z]: [talos] task upgrade (1/1): failed: task "upgrade" failed: exit code 1
 user: warning: [2025-06-11T15:35:21.809719132Z]: [talos] phase upgrade (10/14): failed
 user: warning: [2025-06-11T15:35:21.809821132Z]: [talos] upgrade sequence: failed
 user: warning: [2025-06-11T15:35:21.809840132Z]: [talos] upgrade failed: error running phase 10 in upgrade sequence: task 1/1: failed, task "upgrade" failed: exit code 1
 user: warning: [2025-06-11T15:35:21.810406132Z]: [talos] service[apid](Stopping): Sending SIGTERM to task apid (PID 1594, container apid)
 user: warning: [2025-06-11T15:35:22.169574132Z]: [talos] service[dashboard](Stopping): Sending SIGTERM to Process(["/sbin/dashboard"])
```

### Environment

- Talos version: any lower than the k8s supported version and more than a minor step to the next version that supports the k8s version (e.g. 1.6.4->1.9.4)
- Kubernetes version: [`kubectl version`] any higher than the supported version of the OS version (e.g. 1.32.5)
- Platform: presumably all but in this case `metal-amd64`

smira:
Fixes #11198

We should enforce in following places:

* before starting `upgrade-k8s`, check that all Talos machines would end up with a valid version
* validate in Talos machine configuration, this will cover both upgrades, new installs, and any machine configuration manual edits


smira:
I removed the upgrade pre-checks for Kubernetes version, as now they are handled by the config validation anyways - config validation with new `installer` will ensure that Kubernetes version used in the cluster will match new Talos version

pkg/machinery/config/types/v1alpha1/v1alpha1_validation.go
```diff
@@ -985,7 +987,70 @@ func (c *Config) RuntimeValidate(ctx context.Context, st state.State, mode valid
...
+	if err != nil {
+		return fmt.Errorf("failed to parse Kubernetes version from image reference %q: %w", imageRef, err)
+	}
+
+	return k8sVersion.SupportedWith(currentTalosVersion)
+}
+
+// KubernetesVersionFromImageRef parses the Kubernetes version from the image reference.
+func KubernetesVersionFromImageRef(ref string) (*compatibility.KubernetesVersion, error) {
+	idx := strings.LastIndex(ref, ":v")
```
utkuozdemir, line 1044: I'm curious, should we be doing this assumption (that the tag will follow a specific format, being honest about the version)? I guess we do it in more places anyway, right?

smira: yes, this creates a bit of ambiguity, as the tag is ignored if the hash is present, but we can't do anything better here to guess the version

smira: this check got _moved_ though, so it existed before this PR, but it was used only in Talos upgrade path

utkuozdemir (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_validation_test.go
```diff
@@ -2069,3 +2073,158 @@ func TestValidateCNI(t *testing.T) {
...
+
+func TestKubernetesVersionFromImageRef(t *testing.T) {
+	t.Parallel()
+
+	for _, test := range []struct {
+		imageRef string
+
+		expectedVersion string
+	}{
+		{
```
frezbo, line 2085: maybe also have ones without tag and one with just sha and no tag

smira: we don't allow no tag, as we won't be able to guess the version this way....

smira: but this might be a breaking change, so I will add a note that the tag is required

PR #11262, opened 2025-06-27
fix: set default MTU on Azure to 1400

smira:
Azure documentation is very confusing, but the tests and the issues we observed show that 1400 is a more safe default than 1500, reducing packet fragmentation and timeouts.

See e.g. https://www.reddit.com/r/networking/comments/puowuy/do_anyone_know_the_default_tcp_mss_between_two/


internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
```diff
@@ -160,6 +161,22 @@ func (a *Azure) ParseMetadata(metadata *ComputeMetadata, interfaceAddresses []Ne
 		return nil, err
 	}
 
+	networkConfig.Operators = append(networkConfig.Operators, network.OperatorSpecSpec{
```
smira, line 164: adding DHCP4 explicitly here, as previously it was implicitly ran due to "no config for eth0", and now we do explicit config (with MTU)

Unix4ever (approved):

frezbo (approved):

PR #11263, opened 2025-06-27
fix: treat context canceled as expected error on image pull

smira:
The containerd pull process might internally generate context canceled error while our "external" `ctx` is not canceled.

Don't treat context canceled as non-retriable, as `retry` package stops retrying when the outer `ctx` is canceled, so we don't need to handle it in any specific way.


smira:
This bug manifests in Talos stopping to retry image pull while it could have continued the attempts.

PR #11286, opened 2025-07-01
chore: improve cloud image uploader resilience

smira:
It turns out some S3 regions can't upload a Talos image in 15 minutes, so rework the code once again to make it more resilient to failures:

* if one region/arch fails, ignore it and go forward
* bump S3 upload timeout to 30 minutes
* previously, a failure to upload failed the errgroup which in turn cancels all other uploads, and the whole upload fails.

### DIFF
--- a/hack/cloud-image-uploader/aws.go
+++ b/hack/cloud-image-uploader/aws.go
@@ -249,7 +249,7 @@ func (au *AWSUploader) registerAMI(ctx context.Context, region string, svc *ec2.
 		g.Go(func() error {
 			err = au.registerAMIArch(ctx, region, svc, arch, bucketName, uploader)
 			if err != nil {
-				return fmt.Errorf("error registering AMI for %s: %w", arch, err)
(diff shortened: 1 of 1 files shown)


hack/cloud-image-uploader/aws.go
```diff
@@ -348,9 +348,7 @@ func (au *AWSUploader) registerAMIArch(ctx context.Context, region string, svc *
 		return retry.ExpectedErrorf("task status not found")
 	})
 	if err != nil {
-		log.Printf("WARNING: aws: ignoring failure to import snapshot into %s/%s: %s", region, arch, err)
-
-		return nil //nolint:nilerr
+		return fmt.Errorf("failed to wait for import task: %w", err)
```
frezbo, line 351: so we now fail on import, right?

frezbo (approved):

hack/cloud-image-uploader/aws.go
smira: I moved the ignore line "up" the stack to the place where this whole function is called.

PR #11290, opened 2025-07-02
feat: add talosctl mulitarch bundle image

### LINKED ISSUE #11281: build an image with all `talosctl` binaries
The end goal is to allow Image Factory to serve a `talosctl` binary for each version of Talos.

Today we have `talosctl` image which has amd64/arm64 `talosctl` to be used with `docker run`, the idea is to add another `talosctl-all` image which will contain all `talosctl`s built for all platforms/arches combined together, so that we can offer in Image Factory a URL like `https://factory.talos.dev/talosctl/v1.11.0/talosctl-windows-arm64.exe`

shanduur:
New makefile target, which triggers registry build of a `talosctl-multiarch-bundle` Dockerfile target.
This target builds single image with all possible os/arch combinations of `talosctl`.

Fixes #11281

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,6 +426,9 @@ talos: ## Builds the Talos container image and outputs it to the registry.
 talosctl-image: ## Builds the talosctl container image and outputs it to the registry.
 	@$(MAKE) registry-talosctl
 
+talosctl-all-image:
+	@$(MAKE) registry-talosctl-all
+
 talosctl-all:
 	@$(MAKE) local-talosctl-all DEST=$(ARTIFACTS) PUSH=false
 
@@ -689,6 +692,7 @@ push: ## Pushes the installer, imager, talos and talosctl images to the configur
 	@$(MAKE) installer PUSH=true
 	@$(MAKE) talos PUSH=true
 	@$(MAKE) talosctl-image PUSH=true
+	@$(MAKE) talosctl-all-image PUSH=true
 
 push-%: ## Pushes the installer, imager, talos and talosctl images to the configured container registry with the specified tag (e.g. push-latest).
 	@$(MAKE) push IMAGE_TAG_OUT=$*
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -669,7 +669,7 @@ COPY --from=talosctl-windows-arm64-build /talosctl-windows-arm64.exe /talosctl-w
 
 FROM --platform=${BUILDPLATFORM} talosctl-${TARGETOS}-${TARGETARCH} AS talosctl-targetarch
 
-FROM scratch AS talosctl-all
```
smira: I wonder if we should keep it `talosctl-all` as the name, and ... (see below in the `Makefile`)

Makefile
```diff
@@ -426,6 +426,9 @@ talos: ## Builds the Talos container image and outputs it to the registry.
 talosctl-image: ## Builds the talosctl container image and outputs it to the registry.
 	@$(MAKE) registry-talosctl
 
+talosctl-multiarch-bundle:
```
smira: ... and name this target `talosctl-all-image`, which will do `registry-talosctl-all` (so that the image is `talosctl-all`)

Dockerfile
smira: (as we have `talosct-all` target still in the `Makefile` which outputs to the filesystem)

.kres.yaml
```diff
@@ -151,7 +151,7 @@ spec:
         - not-on-tag
       steps:
         - name: build
-          command: talosctl-all kernel sd-boot sd-stub initramfs installer-base imager talos
+          command: talosctl-multiarch-bundle talosctl-all kernel sd-boot sd-stub initramfs installer-base imager talos
```
smira: probably we don't need to build an image here, and just leave `talosctl-all` which will save `talosctl` binary in the artifacts

(building on `push` should be enough)

Dockerfile
shanduur: Thanks! I've updated all the files, to roll back to the `talosctl-all` name, and include `talosct-all-image` target in the `Makefile`.

Makefile
```diff
@@ -689,6 +692,7 @@ push: ## Pushes the installer, imager, talos and talosctl images to the configur
 	@$(MAKE) installer PUSH=true
 	@$(MAKE) talos PUSH=true
 	@$(MAKE) talosctl-image PUSH=true
+	@$(MAKE) talosctl-all-image PUSH=true
```
smira, line 695: still need to add to the `image-list` target so that it gets signed (see `make sign-images`)

.kres.yaml
```diff
@@ -177,7 +177,7 @@ spec:
         - only-on-tag
       steps:
         - name: build
-          command: talosctl-all kernel sd-boot sd-stub initramfs installer-base imager talos talosctl-cni-bundle
+          command: talosctl-all-image talosctl-all kernel sd-boot sd-stub initramfs installer-base imager talos talosctl-cni-bundle
```
smira: I would drop it from here, as `push` step would trigger it anyways

smira (approved):

Issue #11135, opened 2025-06-02
write the ESP location to EFI boot order vars


smira:
When installing systemd-boot, we create an `EFI` partition which should be discoverable automatically by the firmware, but it might be still not found if there is old explicit location, or if the firmware is outdated.

See e.g. https://github.com/siderolabs/talos/discussions/11127, but there were some more reported in the past.

PR #11305, opened 2025-07-03
feat: support writing EFI boot order

frezbo:
Write Talos as the first in the EFI boot order.

Fixes: https://github.com/siderolabs/talos/issues/11135

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -74,7 +74,6 @@ require (
 	github.com/docker/docker v28.3.0+incompatible
 	github.com/docker/go-connections v0.5.0
 	github.com/dustin/go-humanize v1.0.1
-	github.com/ecks/uefi v0.0.0-20221116212947-caef65d070eb
 	github.com/elastic/go-libaudit/v2 v2.6.2
 	github.com/equinix-ms/go-vmw-guestrpc v0.1.1
 	github.com/fatih/color v1.18.0
@@ -82,6 +81,7 @@ require (
 	github.com/foxboron/go-uefi v0.0.0-20250625111927-a3183a1bfc84
 	github.com/freddierice/go-losetup/v2 v2.0.1
 	github.com/fsnotify/fsnotify v1.9.0
+	github.com/g0rbe/go-chattr v1.0.1
 	github.com/gdamore/tcell/v2 v2.8.1
 	github.com/gertd/go-pluralize v0.2.1
 	github.com/gizak/termui/v3 v3.1.0
@@ -196,7 +196,6 @@ require (
 
 require (
 	cel.dev/expr v0.24.0 // indirect
-	github.com/0x5a17ed/itkit v0.6.0 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.1 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/security/keyvault/internal v1.2.0 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
@@ -357,6 +356,7 @@ require (
 	gopkg.in/evanphx/json-patch.v4 v4.12.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gotest.tools/v3 v3.4.0 // indirect
 	k8s.io/cli-runtime v0.34.0-alpha.2 // indirect
 	k8s.io/kube-openapi v0.0.0-20250610211856-8b98d1ed966a // indirect
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397 // indirect
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
@@ -5,25 +5,22 @@
 package sdboot
 
 import (
-	"errors"
+	"fmt"
+	"slices"
 
-	"github.com/ecks/uefi/efi/efiguid"
-	"github.com/ecks/uefi/efi/efivario"
+	"github.com/siderolabs/gen/xslices"
+	"github.com/siderolabs/go-blockdevice/v2/blkid"
 	"golang.org/x/sys/unix"
-	"golang.org/x/text/encoding/unicode"
 
+	"github.com/siderolabs/talos/internal/pkg/efivarfs"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
 
-const (
-	// SystemdBootGUIDString is the GUID of the SystemdBoot EFI variables.
-	SystemdBootGUIDString = "4a67b082-0a4c-41cf-b6c7-440b29bb8c4f"
-	// SystemdBootStubInfoPath is the path to the SystemdBoot StubInfo EFI variable.
-	SystemdBootStubInfoPath = constants.EFIVarsMountPoint + "/" + "StubInfo-" + SystemdBootGUIDString
-)
+// TalosBootEntryDescription is the description of the Talos Linux UKI UEFI boot entry.
+const TalosBootEntryDescription = "Talos Linux UKI"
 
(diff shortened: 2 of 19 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
```diff
@@ -62,7 +57,7 @@ func ReadVariable(c efivario.Context, name string) (string, error) {
 }
 
 // WriteVariable reads a SystemdBoot EFI variable.
-func WriteVariable(c efivario.Context, name, value string) error {
+func WriteVariable(name, value string) error {
 	// mount EFI vars as rw
 	if err := unix.Mount("efivarfs", constants.EFIVarsMountPoint, "efivarfs", unix.MS_REMOUNT, ""); err != nil {
```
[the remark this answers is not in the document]
frezbo: that code is already present at line 66

go.mod
```diff
@@ -192,11 +191,11 @@ require (
 	kernel.org/pub/linux/libs/security/libcap/cap v1.2.76
 	sigs.k8s.io/hydrophone v0.7.0
 	sigs.k8s.io/yaml v1.5.0
+	source.monogon.dev v0.0.0-20250707205020-c4c89601ff80
```
smira: my concern with this one is that it's a monorepo, and it's quite huge, so we might have issues due to e.g. mixing up dependencies (like in this PR, e.g. nftables update), so we probably need a way to extract this library as a standalone repo

frezbo: so we create a `go-efivarfs` under siderolabs?

smira: if the license allows it - or we retain original copyright, but we do need to have something smaller than this

go.mod
```diff
@@ -74,21 +74,22 @@ require (
...
+	github.com/g0rbe/go-chattr v1.0.1
 	github.com/gdamore/tcell/v2 v2.8.1
 	github.com/gertd/go-pluralize v0.2.1
 	github.com/gizak/termui/v3 v3.1.0
 	github.com/godbus/dbus/v5 v5.1.0
-	github.com/golang/mock v1.6.0
+	github.com/golang/mock v1.7.0-rc.1
 	github.com/google/cadvisor v0.53.0
 	github.com/google/cel-go v0.25.0
+	github.com/google/go-cmp v0.7.0
```
frezbo: `go mod why` does say this is only used in docgen, so I'm not sure why it gets imported here

frezbo:
```fish
go mod why github.com/golang/mock
# github.com/golang/mock
(main module does not need package github.com/golang/mock)

go mod why github.com/google/go-cmp
# github.com/google/go-cmp
(main module does not need package github.com/google/go-cmp)
```

internal/pkg/efivarfs/efivarfs.go
```diff
@@ -0,0 +1,194 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Copyright The Monogon Project Authors.
+// SPDX-License-Identifier: Apache-2.0
```
dsseng, line 6: Does this mean AND or OR dual-licensing? Maybe this should be just left Apache-2.0, as original. Is this file mostly copied from Monogon?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
```diff
@@ -81,5 +74,99 @@ func WriteVariable(c efivario.Context, name, value string) error {
...
+
+	if partitionUUID == nil {
+		return fmt.Errorf("EFI partition UUID not found on install disk %q", installDisk)
+	}
+
+	if err := efivarfs.SetBootEntry(talosBootIndex, &efivarfs.LoadOption{
+		Description: TalosBootEntryDescription,
+		FilePath: efivarfs.DevicePath{
+			&efivarfs.HardDrivePath{
+				PartitionNumber:     1,
```
dsseng: Is this dependent on ESP being the first partition? Perhaps it'd be better to get this from somewhere

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
```diff
@@ -81,5 +74,99 @@ func WriteVariable(c efivario.Context, name, value string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get current BootOrder: %w", err)
+	}
+
+	if currentBootOrder[0] == uint16(talosBootIndex) {
+		// Talos Linux UKI boot entry is already first in the BootOrder
+		return nil
+	}
+
+	if err := efivarfs.SetBootOrder(slices.Concat([]uint16{uint16(talosBootIndex)}, currentBootOrder)); err != nil {
```
dsseng, line 171: Not really sure, but could this concatenated slice be longer than the variable? Perhaps that needs checking if I'm not wrong

dsseng (commented):
Could this be tested e.g. in QEMU?

frezbo:
> Could this be tested e.g. in QEMU?

yeh, this already goes through the normal qemu test

internal/pkg/efivarfs/efivarfs.go
frezbo: hmm, I'm not sure @smira wdyt?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
frezbo: that's a good point, there's a chance it could be not the first partition, let me fix that

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
frezbo: it's a uint16/2 so we shouldn't be worried

dsseng:
> > Could this be tested e.g. in QEMU?
> 
> yeh, this already goes through the normal qemu test

Not sure if it's needed, maybe we could read those vars in integration. But well, if that's incorrect then OVMF won't boot the image, so perhaps that's okay

internal/pkg/efivarfs/efivarfs.go
smira: I'd say as we keep changing these files (and probably we did some minimal changes anyways), we should comply with the original LICENSE:

* keep the Copyright in the file (doesn't have to be at the top)
* put a README to the folder saying that these files were acquired from that source, and copy the original LICENSE there

go.mod
smira: it might be pulled via the test, probably revert `golang/mock` update, I think it's unrelated?

internal/pkg/efivarfs/efivarfs.go
dsseng: IIRC these are Apache requirements. Not sure if we may relicense it as MPL, keeping notices of original being Apache. Certainly works for e.g BSD->GPLv3 or Apache->GPLv3. Keeping original license should be better I guess

go.mod
frezbo: it gets updated, so I'm a bit lost there

smira: revert of all of `go.mod` back to `main` and run `go mod tidy`

go.mod
```diff
@@ -74,21 +74,22 @@ require (
...
 	github.com/fsnotify/fsnotify v1.9.0
+	github.com/g0rbe/go-chattr v1.0.1
 	github.com/gdamore/tcell/v2 v2.8.1
 	github.com/gertd/go-pluralize v0.2.1
 	github.com/gizak/termui/v3 v3.1.0
 	github.com/godbus/dbus/v5 v5.1.0
 	github.com/golang/mock v1.6.0
 	github.com/google/cadvisor v0.53.0
 	github.com/google/cel-go v0.25.0
+	github.com/google/go-cmp v0.7.0
```
frezbo: do we use docgen as a library, `go mod why` shows `mvdan.cc/gofumpt/format` imports it, but only docgen uses that library

PR #11309, opened 2025-07-03
fix: provide FIPS 140-3 compliance

smira:
Provide compliant unit-test runs.


pkg/makefs/ext4.go
```diff
@@ -32,7 +33,7 @@ func Ext4(partname string, setters ...Option) error {
 			return err
 		} else if ok {
 			// ref: https://gitlab.archlinux.org/archlinux/archiso/-/merge_requests/202/diffs
-			detUUID := uuid.NewSHA1(uuid.MustParse("93a870ff-8565-4cf3-a67b-f47299271a96"), []byte(fmt.Sprintf("%d ext4 hash seed", epoch)))
+			detUUID := uuid.NewHash(sha256.New(), uuid.MustParse("93a870ff-8565-4cf3-a67b-f47299271a96"), []byte(fmt.Sprintf("%d ext4 hash seed", epoch)), 5)
```
smira, line 36: @frezbo that is going to change the UUID, but we can't use sha1 in FIPS mode

I don't think we had much reproducible ext4 though (probably only Image Cache?)

frezbo: yeh, let's not worry too much here, as long as the new code generates a consistent uuid

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -98,6 +100,12 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
 		log.Printf("race detection enabled with halt_on_error=1")
 	}
 
+	if val := procfs.ProcCmdline().Get("talos.fips140"); val != nil && fipsmode.Enabled() {
```
smira, line 103: no documentation here yet, this is mostly for testing only - we don't provide a FIPS build yet

internal/pkg/mount/switchroot/switchroot.go
```diff
@@ -98,6 +100,12 @@ func Switch(prefix string, mountpoints mount.Points) (err error) {
...
+		if pointer.SafeDeref(val.First()) == "strict" {
+			envv = append(envv, fipsmode.StrictEnvironmentVariable())
```
frezbo, line 105: I guess `envv` is a previous typo

smira: nope, it's correct arg name here: https://pkg.go.dev/golang.org/x/sys/unix#Exec (goes from the C naming)

PR #11318, opened 2025-07-06
feat: allow writing /etc/nfs.conf

buroa:
Need to be able to overwrite/mount `/etc/nfs.conf` for nfs-utils.

Supplemental PR to https://github.com/siderolabs/extensions/pull/753


smira:
This doesn't allow writing `nfs.conf`, does it?

PR #11323, opened 2025-07-07
feat: include packages into SBOM

dsseng:
Also generate multiple SBOMs for variants including different
sets of packages and different architectures.

Generated SBOMs (you can use this [web app](https://apps.rancher.io/sbom-viewer) to view them):

[talos-container-arm64.spdx.json](https://github.com/user-attachments/files/21107055/talos-container-arm64.spdx.json)
[talos-container-amd64.spdx.json](https://github.com/user-attachments/files/21107057/talos-container-amd64.spdx.json)
[talos-arm64.spdx.json](https://github.com/user-attachments/files/21107058/talos-arm64.spdx.json)
[talos-amd64.spdx.json](https://github.com/user-attachments/files/21107059/talos-amd64.spdx.json)

Scan: `grype sbom:talos-container-arm64.spdx.json`: `No vulnerabilities found`

Then patch with this to replace versions with known bad ones (only change format for Linux as NVD will report even current as vulnerable, if you scan a full image) (v prefix in version doesn't matter to Grype):

<details>
<summary>Details</summary>

```patch
19c19
<    "versionInfo": "v3.1.7",
---
>    "versionInfo": "3.1.5",
31c31
<      "referenceLocator": "cpe:2.3:a:apparmor:apparmor:v3.1.7:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:a:apparmor:apparmor:3.1.5:*:*:*:*:*:*:*"
36c36
<      "referenceLocator": "cpe:2.3:a:canonical:apparmor:v3.1.7:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:a:canonical:apparmor:3.1.5:*:*:*:*:*:*:*"
105c105
<    "versionInfo": "v2.1.3",
---
>    "versionInfo": "1.6.0",
117c117
<      "referenceLocator": "cpe:2.3:a:containerd:containerd:v2.1.3:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:a:containerd:containerd:1.6.0:*:*:*:*:*:*:*"
122c122
<      "referenceLocator": "cpe:2.3:a:linuxfoundation:containerd:v2.1.3:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:a:linuxfoundation:containerd:1.6.0:*:*:*:*:*:*:*"
127c127
<      "referenceLocator": "pkg:github/containerd/containerd@v2.1.3"
---
>      "referenceLocator": "pkg:github/containerd/containerd@1.6.0"
134c134
<    "versionInfo": "2.7.5",
---
>    "versionInfo": "v2.4.1",
146c146
<      "referenceLocator": "cpe:2.3:a:cryptsetup_project:cryptsetup:2.7.5:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:a:cryptsetup_project:cryptsetup:v2.4.1:*:*:*:*:*:*:*"
8538c8538
<    "versionInfo": "6.12.35",
---
>    "versionInfo": "v6.12.35",
8550c8550
<      "referenceLocator": "cpe:2.3:o:linux:linux_kernel:6.12.35:*:*:*:*:*:*:*"
---
>      "referenceLocator": "cpe:2.3:o:linux:linux_kernel:v6.12.35:*:*:*:*:*:*:*"
```

</details>

Scan once more - grype sees the packages, which means it recognized them by their CPEs and read the version correctly:

```
 ✔ Scanned for vulnerabilities     [8 vulnerability matches]  
   ├── by severity: 1 critical, 2 high, 5 medium, 0 low, 0 negligible
   └── by status:   8 fixed, 0 not-fixed, 0 ignored 
NAME                INSTALLED  FIXED IN                TYPE            VULNERABILITY   SEVERITY  EPSS %  RISK   
containerd-aarch64  1.6.0      1.4.13, 1.5.10, 1.6.1   UnknownPackage  CVE-2022-23648  High      90.37   4.4    
containerd-aarch64  1.6.0      1.5.16, 1.6.12          UnknownPackage  CVE-2022-23471  Medium    42.24   0.1    
apparmor-aarch64    3.1.5      2.13.10, 3.0.12, 3.1.6                  CVE-2016-1585   Critical  24.04   < 0.1  
containerd-aarch64  1.6.0      1.5.18, 1.6.18          UnknownPackage  CVE-2023-25153  Medium    31.08   < 0.1  
containerd-aarch64  1.6.0      1.5.13, 1.6.6           UnknownPackage  CVE-2022-31030  Medium    30.06   < 0.1  
cryptsetup-aarch64  v2.4.1     2.3.7, 2.4.3                            CVE-2021-4122   Medium    15.82   < 0.1  
containerd-aarch64  1.6.0      1.5.18, 1.6.18          UnknownPackage  CVE-2023-25173  High      4.27    < 0.1  
containerd-aarch64  1.6.0      1.6.38, 1.7.27, 2.0.4   UnknownPackage  CVE-2024-40635  Medium    0.62    < 0.1
```

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -207,7 +207,10 @@ spec:
               - metal-arm64-uki.efi
               - metal-amd64.raw.zst
               - metal-arm64.raw.zst
-              - sbom.json
+              - talos-arm64.spdx.json
+              - talos-amd64.spdx.json
+              - talos-container-arm64.spdx.json
+              - talos-container-amd64.spdx.json
               - talosctl-cni-bundle-amd64.tar.gz
               - talosctl-cni-bundle-arm64.tar.gz
               - talosctl-darwin-amd64
--- a/Dockerfile
+++ b/Dockerfile
@@ -449,30 +449,6 @@ FROM scratch AS microsoft-db-keys
 COPY --from=microsoft-secureboot-database /DB/Certificates/MicCor*.der /db/
 COPY --from=microsoft-secureboot-database /DB/Certificates/microsoft*.der /db/
 
-FROM build-go AS sbom-generate
-COPY ./tools ./tools
-
-ARG SOURCE_DATE_EPOCH
-ENV SYFT_FORMAT_SPDX_JSON_CREATED_TIME=${SOURCE_DATE_EPOCH}
-ENV SYFT_FORMAT_PRETTY=1
-ENV SYFT_FORMAT_SPDX_JSON_DETERMINISTIC_UUID=1
-
-ARG NAME
-ARG TAG
-
-RUN mkdir -p sbom-src /usr/share/sbom
-# TODO: copy pkgs SBOMs to sbom-src when we generate them
-RUN cp go.mod go.sum sbom-src
-RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool -modfile=tools/go.mod \
-    github.com/anchore/syft/cmd/syft \
-    scan --from dir sbom-src \
-    --select-catalogers "+sbom-cataloger,go" \
-    --source-name "${NAME}" --source-version "${TAG}" \
-    -o spdx-json > /usr/share/sbom/sbom.json
-
-FROM scratch AS sbom
-COPY --from=sbom-generate /usr/share/sbom/sbom.json /
-
 FROM --platform=${BUILDPLATFORM} scratch AS generate
 COPY --from=proto-format-build /src/api /api/
 COPY --from=generate-build /api/common/*.pb.go /pkg/machinery/api/common/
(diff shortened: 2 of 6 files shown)


.kres.yaml
```diff
@@ -207,7 +207,10 @@ spec:
               - metal-arm64-uki.efi
               - metal-amd64.raw.zst
               - metal-arm64.raw.zst
-              - sbom.json
+              - talos-arm64.spdx.json
+              - talos-amd64.spdx.json
+              - talos-container-arm64.spdx.json
+              - talos-container-amd64.spdx.json
```
dsseng, line 213: @smira hopefully I added these to the right place. `make local-sbom PROGRESS=plain PLATFORM=linux/amd64 DEST=./` produces these files in the current dir, so that should work

hack/sbom.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+SYFT_FORMAT_PRETTY=1 SYFT_FORMAT_SPDX_JSON_DETERMINISTIC_UUID=1 \
+	go tool -modfile=tools/go.mod \
+	github.com/anchore/syft/cmd/syft \
+	scan --from dir /tmp/sbom-src \
```
dsseng: nit: this can be parameterized if that reduces clutter and not increases it

Dockerfile
```diff
@@ -893,13 +869,68 @@ RUN <<END
...
-FROM rootfs-base-${TARGETARCH} AS rootfs-base
+FROM build-go AS build-sbom
+ARG SOURCE_DATE_EPOCH
+ENV SYFT_FORMAT_SPDX_JSON_CREATED_TIME=${SOURCE_DATE_EPOCH}
+ARG NAME
+ARG TAG
+
+COPY ./hack/sbom.sh /usr/bin/sbom.sh
+COPY ./tools ./tools
+
```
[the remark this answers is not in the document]
dsseng: Only these two files are there now

Dockerfile
```diff
@@ -893,13 +869,68 @@ RUN <<END
...
+FROM scratch AS sbom-container-arm64
+COPY --from=sbom-container-arm64-generate /rootfs/usr/share/spdx/talos-container-arm64.spdx.json /
+
+FROM build-sbom AS sbom-container-amd64-generate
+COPY --from=rootfs-base-amd64 /rootfs/usr/share/spdx /tmp/sbom-src/
+RUN --mount=type=cache,target=/.cache,id=talos/.cache sbom.sh "$NAME (amd64 container)" talos-container-amd64.spdx.json
+
+FROM scratch AS sbom-container-amd64
+COPY --from=sbom-container-amd64-generate /rootfs/usr/share/spdx/talos-container-amd64.spdx.json /
+
```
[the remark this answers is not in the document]
dsseng: Need to generate all SBOMs anyway

Dockerfile
```diff
@@ -893,13 +869,68 @@ RUN <<END
...
+FROM scratch AS sbom
+COPY --from=sbom-container-arm64 / /
+COPY --from=sbom-container-amd64 / /
+COPY --from=sbom-arm64 / /
+COPY --from=sbom-amd64 / /
+
+# Add resulting SBOMs to the container rootfs
+FROM rootfs-base-arm64 AS rootfs-sbom-arm64
+COPY --from=sbom-container-arm64 / /rootfs/usr/share/spdx/
```
smira: I guess we could skip that extra layer with:

```
FROM sbom-container-${TARGETARCH} AS sbom-container-targeteach

....

FROM rootfs-base-${TARGETARCH} as rootfs-base
COPY --from=sbom-container-targeteach / /rootfs/usr/share/spdx
```

(buildkit doesn't allow variables in `--from`, but with `FROM` alias we can use vars)

Dockerfile
```diff
@@ -449,30 +449,6 @@ FROM scratch AS microsoft-db-keys
 COPY --from=microsoft-secureboot-database /DB/Certificates/MicCor*.der /db/
```
smira, line 449: we should be somwhere removing SBOMs from the rootfs, and then adding back merged ones? I don't see it?

Dockerfile
dsseng: Ah, should optimize the line count I guess. It'll still be dropped out when buildkit tree-shakes the build, but you approach is cleaner, will fix, thanks

Dockerfile
dsseng: Hmm? I build SBOMs as a separate layers, and then add it in `rootfs-base`, `rootfs-squashfs-arm64`, `rootfs-squashfs-amd64` steps (copy from the generated layer into the image layer). This allows an `sbom` target to export all SBOMs in the tag build process

smira: yes,  but SBOMs are already in say `pkg-kernel`, so does the cleanup script remove it?

dsseng: I decided to not remove those yet, since we do not copy their source file data (downloaded source archive URLs and hashes) into the resulting SBOM. If it's not wanted I can remove those before copying the final SBOM in. Not sure what'd be better

smira: ok, so I'll take a look at the final SBOMs or even try to present them as resources to see how it looks

dsseng: It might be great though to create some separation to indicate talos.spdx.json and later extensions' SBOMs are the primary ones, and other ones are sort of sub-SBOMs

smira: I don't think we need to ship _too many_ SBOMs (at least for now) - we should have a clear core Talos SBOM. If it's missing something right now -- we can go back and add stuff there, but the core SBOM should be a single file imho

dsseng: Because you have all necessary info about Talos in its SBOM, and e.g. containerd's one gives you what containerd was built with. Or maybe we should drop them from the image altogether since that information is of not much use and can still be followed to by reading sources

smira: I mean there should be one for core Talos and others for extensions

dsseng: Yes, so I'll remove pkgs' ones now. Should it be okay to rm -rf it, or I better provide some separation via `COPY --filter` and create a separate target like `rootfs-arm64-sboms`? The latter seems to carry an extra risk of adding a package but forgetting to add its SBOM, so I guess that won't be a huge issue if we remove SBOM directories in `rootfs-squashfs-arm64` and so on

dsseng: `--exclude` and `--from` seem to conflict, weird. Did `rm`

PR #11325, opened 2025-07-08
feat: export SBOM as resources

### LINKED ISSUE #10942: API to retrieve SPDX
Split SPDX into resources per package and present them in `talosctl get sbom` (or some better name).

smira:
Fixes #10942


smira:
pending fixes:

- [x] drop arch from `pkgs` SBOMs, so we can assert on them easily in the test, and also assert e.g. that linux is in non-container mode, and vice versa

smira:
I've added a bit more tests, but waiting for @dsseng to finalize SBOMs, so that I can assert even more

PR #11333, opened 2025-07-09
feat: bump Linux, Go and other packages

dsseng:
Include all core packages into SBOM, make sure Talos is built with the
same Go versions as pkgs.


Dockerfile
```diff
@@ -877,6 +887,14 @@ ENV SYFT_FORMAT_SPDX_JSON_CREATED_TIME=${SOURCE_DATE_EPOCH}
 ARG NAME
 ARG TAG
 
+COPY --from=pkg-containerd-amd64 /usr/bin/containerd-shim-runc-v2 /tmp
```
smira: I would add some comment here for the future readers on what's going on here :)

dsseng: Ah, yeah, worth it!

hack/release.toml
```diff
@@ -18,15 +18,20 @@ preface = """
...
         description = """\
-Linux: 6.12.35
+Linux: 6.12.36
 Kubernetes: 1.34.0-alpha.2
 runc: 1.3.0
 containerd: 2.1.3
 Flannel CNI plugin: 1.7.1-flannel1
 Flannel: 0.27.0
 CoreDNS: 1.12.2
+ipxe: c21443f
```
smira: we don't usually announce iPXE updates in Talos

`ipxe` doesn't ship with Talos (it's actually only part of `talosctl cluster create`)

others I think are good

hack/release.toml
```diff
@@ -101,6 +106,12 @@ The SBOM is available in the `/usr/share/sbom` directory on the machine and can
         title = "Disk Wipe"
         description = """\
 Talos now supports `talosctl disk wipe` command in maintenance mode (`talosctl disk wipe <disk> --insecure`).
+"""
+
+    [notes.sbom]
+        title = "SBOM generation"
```
smira: please see line 93/98: we already have it.

dsseng:
> if we have Go now as SBOM, can we add it to the test in `internal/integration/api/sbom.go` please?

Yes. Btw a great catch: we don't, I forgot to copy it over

Dockerfile
```diff
@@ -727,8 +729,10 @@ COPY --link --exclude=usr/bin/ctr --from=pkg-containerd-amd64 / /rootfs
 COPY --link --from=pkg-dosfstools-amd64 / /rootfs
 COPY --link --from=pkg-e2fsprogs-amd64 / /rootfs
 COPY --link --exclude=usr/share --from=pkg-systemd-udevd-amd64 / /rootfs
+COPY --link --exclude=usr/share --from=pkg-systemd-udevd-amd64 /usr/share/spdx/systemd.spdx.json /rootfs/usr/share/spdx/systemd.spdx.json
```
frezbo: would this work an `exclude` of `usr/share` and then copying from there itself, I think we don't need the exclude directive here

Dockerfile
frezbo: can;t we just do this as a unit test, way easier and not affected by this layer being cached

Dockerfile
```diff
@@ -933,8 +951,6 @@ FROM rootfs-base-arm64 AS rootfs-squashfs-arm64
 RUN rm -rf /rootfs/usr/share/spdx/*
 COPY --from=sbom-arm64 / /rootfs/usr/share/spdx/
 ARG ZSTD_COMPRESSION_LEVEL
-RUN find /rootfs -print0 \
```
frezbo, line 936: not needed anymore?

Dockerfile
dsseng: Makes sense, thanks. It works though

Dockerfile
dsseng: Yes, new mksquashfs honors SOURCE_DATE_EPOCH. I should run reproducibility test though

Makefile
```diff
@@ -25,9 +25,9 @@ DEBUG_TOOLS_SOURCE := scratch
 EMBED_TARGET ?= embed
 
 TOOLS_PREFIX ?= ghcr.io/siderolabs/tools
-TOOLS ?= v1.11.0-alpha.0-3-g1dfd14b
+TOOLS ?= v1.11.0-alpha.0-6-g4818702
 PKGS_PREFIX ?= ghcr.io/siderolabs
-PKGS ?= v1.11.0-alpha.0-43-g2537e61
+PKGS ?= v1.11.0-alpha.0-45-g90f7c65
```
dsseng: This'll need one more bump to remove PURLs which confuse scanners into treating some deps, e.g. containerd, cni and runc as Go modules which they in this context aren't:

```
[0000] TRACE searching for vulnerability matches package=pkg:golang/github.com/containernetworking/plugins@0.6.1
[0000] TRACE fetched affected package record distro=none duration=127.41µs pkg=package(name=cni, ecosystem=go-module) records=0 vulns=any
```

Dockerfile
```diff
@@ -311,6 +313,8 @@ ENV GOMODCACHE=/.cache/mod
 ENV PROTOTOOL_CACHE_PATH=/.cache/prototool
 ARG SOURCE_DATE_EPOCH
 ENV SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH}
+# Go standard library is shipped with Talos, thus it must be tracked in SBOM
+RUN mkdir -p /rootfs/usr/share/spdx/ && cp /usr/share/spdx/golang.spdx.json /rootfs/usr/share/spdx/golang.spdx.json
```
frezbo: can't we just do a `COPY --from=tools` as a step in initramfs step, also avoids `RUN` commands which speeds up builds

Dockerfile
dsseng: Not sure. Cache should be discarded as needed, so this would fail the build once currently selected sources have this mismatch. Could be moved to an integration test though, not sure about unit

Dockerfile
dsseng: Perhaps a `COPY --from=tools` here, to make this a shared layer and avoid doing this thrice for two initramfs and container rootfs. Will fix, thx

Dockerfile
smira: I would say a unit-test might be easier, as the unit-test has Talos rootfs, so it can even programmatically read Go version from any binary in the rootfs. 

We can use [BuildInfo](https://pkg.go.dev/debug/buildinfo#ReadFile) which contains Go version

dsseng: Oh, great, yes, I guess I should do a unit test then, if we've got rootfs there. Thanks for linking to buildinfo, will implement

Dockerfile
```diff
@@ -311,6 +313,8 @@ ENV GOMODCACHE=/.cache/mod
 ENV PROTOTOOL_CACHE_PATH=/.cache/prototool
 ARG SOURCE_DATE_EPOCH
 ENV SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH}
+# Go standard library is shipped with Talos, thus it must be tracked in SBOM
+COPY --from=tools /usr/share/spdx/golang.spdx.json /rootfs/usr/share/spdx/golang.spdx.json
```
frezbo: missing `--link`

PR #11337, opened 2025-07-10
fix: change module instance evaluation 

shanduur:
A minor change to validation of LoadedKernelModules (Instances field).
This field should be non-negative, zero should be treaded as correct value.


frezbo (approved):

internal/integration/api/extensions_qemu.go
```diff
@@ -600,7 +600,7 @@ func (suite *ExtensionsSuiteQEMU) TestLoadedKernelModule() {
 			asrt.NotEmpty(res.TypedSpec().Name, "kernel module name should not be empty")
 			asrt.NotEmpty(res.TypedSpec().Size, "kernel module size should not be empty")
 			asrt.NotEmpty(res.TypedSpec().Address, "kernel module address should not be empty")
-			asrt.NotZero(res.TypedSpec().Instances, "kernel module instances should not be zero")
```
smira, line 603: probably message is outdated now? as we assert it's not negative?

smira (approved):

dsseng (approved):

PR #11346, opened 2025-07-10
fix: rename instances to referenceCount

shanduur:
`ReferenceCount` is a more common name, so replacing `Instances` will provide better UX and will match online resources.

Additionally drop `Name` which is duplicated from ID.

### DIFF
--- a/pkg/machinery/resources/runtime/loaded_kernel_module.go
+++ b/pkg/machinery/resources/runtime/loaded_kernel_module.go
@@ -23,12 +23,11 @@ type LoadedKernelModule = typed.Resource[LoadedKernelModuleSpec, LoadedKernelMod
 //
 //gotagsrewrite:gen
 type LoadedKernelModuleSpec struct {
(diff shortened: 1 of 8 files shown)


pkg/machinery/resources/runtime/loaded_kernel_module.go
```diff
@@ -49,10 +48,6 @@ func (LoadedKernelModuleExtension) ResourceDefinition() meta.ResourceDefinitionS
 		Aliases:          []resource.Type{"module", "modules"},
 		DefaultNamespace: NamespaceName,
 		PrintColumns: []meta.PrintColumn{
-			{
```
frezbo, line 52: was this intentional?

frezbo (approved):

pkg/machinery/resources/runtime/loaded_kernel_module.go
shanduur: Yes. Name is duplicated in ID, getting rid of it to simplify the output.

PR #11347, opened 2025-07-10
feat: implement raw volume support

### LINKED ISSUE #10931: `RawVolumeConfig`: support for unformatted volumes
Similar to user volumes in the way they work, but the volume should not be formatted and mounted.

Probably should support disk encryption as well.

See https://github.com/siderolabs/talos/discussions/10795

smira:
Fixes #10931


website/content/v1.11/talos-guides/configuration/disk-management.md
smira, line 1: I refactored by splitting into parts, and fixing small things/expanding, also adding raw volume docs.

frezbo (approved):

PR #11363, opened 2025-07-11
feat(talosctl): support tpm operation on mac

frezbo:
Support TPM operations when developing on Mac.

`brew install swtpm` is all that is needed.


Orzelius:
Just works out of the box? nice!

frezbo:
> Just works out of the box? nice!

yeh, secureboot is still pending, since the brew qemu/ovmf doesn't ship a vars file that has secureboot support, I couldn't find info on how to support it

PR #11377, opened 2025-07-15
docs: add release notes about v3.6.x bug

shanduur:
Talos won't default to ETCD v3.6.x in the next release due to bug in the v3.6.x series that causes data corruption in some cases.
The default version will be kept at v3.5.x for now, but users can still use v3.6.x by explicitly specifying it in the configuration.

X-Ref: https://github.com/etcd-io/etcd/issues/20340

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -113,6 +113,15 @@ Talos now supports `talosctl disk wipe` command in maintenance mode (`talosctl d
 Talos now supports [raw user volumes](https://www.talos.dev/v1.11/talos-guides/configuration/disk-management/raw/), allowing to allocate unformatted disk space as partition.
 """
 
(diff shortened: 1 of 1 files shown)


ahrtr:
The issue has been resolved.

etcd v3.6.4 was just released with the fix.

https://github.com/etcd-io/etcd/releases/tag/v3.6.4

smira:
> The issue has been resolved.
> 
> etcd v3.6.4 was just released with the fix.
> 
> [etcd-io/etcd@`v3.6.4` (release)](https://github.com/etcd-io/etcd/releases/tag/v3.6.4)

thank you, we will give it a spin and make a decision if we go with 3.6.x for Talos 1.11

Issue #11344, opened 2025-07-10
use/set selected entry in addition to default one


smira:
When probing, check selected entry (EFI var) in addition to default one.

On kexec clear selected entry or set it to the entry being booted.

frezbo:
We cannot set/delete `LoaderEntrySelected` since systemd writes it as a volatile variable, Ref: https://github.com/systemd/systemd/blob/e65455feade65c798fd1742220768eba7f81755b/src/boot/boot.c#L2752, this only causes a minor UX issue if someone manually reads `LoaderEntrySelected` otherwise, Talos will never use this variable.

smira:
I'm confused a bit here:

* let's say I have two UKIs, A & B 
* the default entry is B
* I boot into B, it's broken
* I got to bootloader menu, and select A - A is selected now
* A boots just fine
* I do a talos upgrade to C, followed by kexec:
  * Talos should remove UKI (B), not A from `EFI` partition
  * On next boot, Talos should understand that `C` got booted, not `A`

frezbo:
Okay, this is doable, then we need to cleanup based on `LoaderEntrySelected` and `LoaderEntryDefault`, though I think we never supported this with GRUB

Issue #11343, opened 2025-07-10
never overwrite booted UKIs


smira:
If I install Talos v1.10.0, and upgrade to 1.10.0, the UKI filename will be same `Talos-v1.10.0.efi`, so it will get overwritten.

Never overwrite UKIs, add a suffix instead.

PR #11380, opened 2025-07-16
fix: sd-boot kexec with disk images

frezbo:
Fixes: #11343
Fixes: #11344

Also simplify the probing code since that only runs during `Upgrade`, `Kexec` or `Reboot`.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/foxboron/go-uefi/efi"
@@ -67,6 +68,7 @@ func New() *Config {
 }
 
 // ProbeWithCallback probes the sd-boot bootloader, and calls the callback function with the Config.
+// this is called when we upgrade, do KexecLoad, or for reverting the bootloader.
 //
 //nolint:gocyclo
 func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(*Config) error) (*Config, error) {
@@ -76,19 +78,6 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
 
 		return nil, nil
 	}
-
-	// here we need to read the EFI vars to see if we have any defaults
-	// and populate config accordingly
-	// https://www.freedesktop.org/software/systemd/man/latest/systemd-boot.html#LoaderEntryDefault
-	// this is set by systemd-boot.
-	// first we start by checking if we have a Default entry
-	// this is set by installer
-	bootedEntry, err := ReadVariable(LoaderEntryDefaultName)
-	if err != nil {
-		return nil, err
-	}
-
-	options.Log("sd-boot: booted entry: %q", bootedEntry)
 
 	var sdbootConf *Config
 
@@ -111,7 +100,7 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
 			}
 
 			if len(files) == 0 {
-				return fmt.Errorf("no boot*.efi files found in %q", filepath.Join(constants.EFIMountPoint, "EFI", "boot"))
+				return fmt.Errorf("no boot*.efi files found in %s", filepath.Join(constants.EFIMountPoint, "EFI", "boot"))
 			}
 
 			// list existing UKIs, and check if the current one is present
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -120,29 +109,59 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
...
+			}
+
+			options.Log("sd-boot: LoaderEntryDefault: %s", bootEntry)
+
+			if bootEntry == "" {
+				// If we booted of a Disk image, only `LoaderEntrySelected` will be set until we do an upgrade
+				// which will set the `LoaderEntryDefault` to the UKI file name.
+				// So for reboot with Kexec we will have to read the `LoaderEntrySelected`
+				// upgrades will always have `LoaderEntryDefault` set to the UKI file name.
+				loaderEntrySelected, err := ReadVariable(LoaderEntrySelectedName)
```
smira, line 134: I'm confused here, if I selected a non-default entry, should the selected take over?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -120,29 +109,59 @@ func ProbeWithCallback(disk string, options options.ProbeOptions, callback func(
...
+				if err != nil {
+					return err
+				}
 
-				sdbootConf = &Config{
-					Default: filepath.Base(ukiFiles[0]),
+				if loaderEntrySelected == "" {
```
smira, line 139: what if I boot a UKI directly? none of the variables are set? and that is fine?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -340,8 +359,10 @@ func (c *Config) install(opts options.InstallOptions) (*options.InstallResult, e
 		return nil, err
 	}
 
-	// writing UKI by version-based filename here
-	ukiPath := fmt.Sprintf("%s-%s.efi", "Talos", opts.Version)
+	// writing UKI by version-based filename + a 7 digit timestamp
+	// this is to avoid collisions with the same version UKI being installed multiple times
+	// hash is calculated from current timestamp
+	ukiPath := fmt.Sprintf("%s-%s-%07d.efi", "Talos", opts.Version, time.Now().Unix())
```
smira: I think it'd be nice to do the suffix only if we have a collision, and avoid it otherwise?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: this code is never used during install, so it doesn't matter

frezbo: and on upgrades the installer will set the `LoaderEntryDefault` to the newly copied UKI file from the installer

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: yeh, since installer code never calls this, though I forgot to test with a UKI, I'll test that locally

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
frezbo: good point, one thing I noticed is that if we do it always the disk image will be non-reproducible, since the timestamp always gets added, so doing this will fix that too.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
smira: but see my point in the issue about which one we should actually overwrite

frezbo: I think we can do better during install, on install/upgrade we set `LoaderEntryOneShot` and if the next one boots to completion during the stage we write meta we can write the current value of `LoaderEntryOneShot` if kexec'ed or `LoaderEntrySelected` if a normal reboot to the `LoaderEntryDefault` value.

smira (approved):

PR #11382, opened 2025-07-17
chore(machined): remove deprecated Endpoints

### LINKED ISSUE #11322: stop using deprecated v1.Endpoint Kubernetes resource
Deprecated since Kubernetes 1.33, a replacement EndpointSlice is supported for all supported releases of Kubernetes.

Might be non-trivial move, but we need to move.

For `kubeaccess.EndpointController`, it might make sense to clean up outdated `Endpoint` resource.

shanduur:
Remove deprecated core/v1.Endpoints in favor of discovery/v1.EndpointSlices.

Fixes #11322

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -137,7 +137,7 @@ require (
 	github.com/siderolabs/crypto v0.6.3
 	github.com/siderolabs/discovery-api v0.1.6
 	github.com/siderolabs/discovery-client v0.1.12
-	github.com/siderolabs/gen v0.8.4
+	github.com/siderolabs/gen v0.8.5
 	github.com/siderolabs/go-api-signature v0.3.6
 	github.com/siderolabs/go-blockdevice v0.4.8
 	github.com/siderolabs/go-blockdevice/v2 v2.0.18
@@ -189,6 +189,7 @@ require (
 	google.golang.org/protobuf v1.36.6
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/klog/v2 v2.130.1
+	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
 	kernel.org/pub/linux/libs/security/libcap/cap v1.2.76
 	sigs.k8s.io/hydrophone v0.7.0
 	sigs.k8s.io/yaml v1.5.0
@@ -359,7 +360,6 @@ require (
 	gotest.tools/v3 v3.4.0 // indirect
 	k8s.io/cli-runtime v0.34.0-beta.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20250710124328-f3f2b991d03b // indirect
-	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397 // indirect
 	kernel.org/pub/linux/libs/security/libcap/psx v1.2.76 // indirect
 	rsc.io/qr v0.2.0 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
--- a/internal/app/machined/pkg/controllers/k8s/endpoint.go
+++ b/internal/app/machined/pkg/controllers/k8s/endpoint.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"net/netip"
 	"slices"
-	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/safe"
@@ -17,10 +16,10 @@ import (
 	"github.com/siderolabs/gen/optional"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
@@ -119,29 +118,22 @@ func (ctrl *EndpointController) watchEndpointsOnWorker(ctx context.Context, r co
 
 	defer client.Close() //nolint:errcheck
 
-	ticker := time.NewTicker(10 * time.Minute)
-	defer ticker.Stop()
+	r.QueueReconcile()
 
 	for {
-		// unfortunately we can't use Watch or CachedInformer here as system:node role is only allowed verb 'Get'
-		endpoints, err := client.CoreV1().Endpoints(corev1.NamespaceDefault).Get(ctx, "kubernetes", v1.GetOptions{})
-		if err != nil {
-			return fmt.Errorf("error getting endpoints: %w", err)
-		}
-
-		if err = ctrl.updateEndpointsResource(ctx, r, logger, endpoints); err != nil {
-			return err
-		}
-
 		select {
+		case <-r.EventCh():
 		case <-ctx.Done():
 			return nil
-		case <-ticker.C:
-		case <-r.EventCh():
+		}
+
+		if err = ctrl.watchKubernetesEndpointSlices(ctx, r, logger, client); err != nil {
+			return err
 		}
 	}
 }
 
+//nolint:gocyclo
 func (ctrl *EndpointController) watchEndpointsOnControlPlane(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
 	if err := r.UpdateInputs([]controller.Input{
 		{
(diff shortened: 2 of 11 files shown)


internal/app/machined/pkg/controllers/k8s/endpoint.go
```diff
@@ -124,12 +125,12 @@ func (ctrl *EndpointController) watchEndpointsOnWorker(ctx context.Context, r co
 
 	for {
 		// unfortunately we can't use Watch or CachedInformer here as system:node role is only allowed verb 'Get'
-		endpoints, err := client.CoreV1().Endpoints(corev1.NamespaceDefault).Get(ctx, "kubernetes", v1.GetOptions{})
+		endpointSlice, err := client.DiscoveryV1().EndpointSlices(corev1.NamespaceDefault).Get(ctx, "kubernetes", v1.GetOptions{})
```
smira: looking at sample output:

```yaml
# kubectl get endpointslices kubernetes -o yaml 
addressType: IPv4
apiVersion: discovery.k8s.io/v1
endpoints:
- addresses:
  - 172.20.0.2
  conditions:
    ready: true
- addresses:
  - 172.20.0.3
  conditions:
    ready: true
kind: EndpointSlice
metadata:
  creationTimestamp: "2025-07-17T11:29:24Z"
  generation: 2
  labels:
    kubernetes.io/service-name: kubernetes
  name: kubernetes
  namespace: default
  resourceVersion: "766"
  uid: d5cc8407-2c21-4d25-8a44-ec357877fa97
ports:
- name: https
  port: 6443
  protocol: TCP
```

I wonder, that it seems to be per-IP-type, so should we instead list by label and aggregate them?looking at sample output:

internal/app/machined/pkg/controllers/k8s/endpoint.go
```diff
@@ -276,7 +277,7 @@ func kubernetesEndpointWatcher(ctx context.Context, logger *zap.Logger, client *
 		}),
 	)
 
-	notifyCh := make(chan *corev1.Endpoints, 1) //nolint:staticcheck
+	notifyCh := make(chan *discoveryv1.EndpointSlice, 1)
 
 	informer := informerFactory.Core().V1().Endpoints().Informer()
```
smira: same here: can we/should we use a filter by label?

smira: and, as there is ipv4 and ipv6 slice, we need to take care of correctly updating the result

internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
```diff
@@ -197,17 +204,17 @@ func (ctrl *EndpointController) ensureTalosService(ctx context.Context, client *
 }
 
 //nolint:gocyclo
-func (ctrl *EndpointController) ensureTalosEndpoints(ctx context.Context, logger *zap.Logger, client *kubernetes.Client, endpointAddrs k8s.EndpointList) error {
+func (ctrl *EndpointController) ensureTalosEndpointSlices(ctx context.Context, logger *zap.Logger, client *kubernetes.Client, endpointAddrs k8s.EndpointList) error {
 	for {
-		oldEndpoints, err := client.CoreV1().Endpoints(constants.KubernetesTalosAPIServiceNamespace).Get(ctx, constants.KubernetesTalosAPIServiceName, metav1.GetOptions{})
+		oldEndpointSlice, err := client.DiscoveryV1().EndpointSlices(constants.KubernetesTalosAPIServiceNamespace).Get(ctx, constants.KubernetesTalosAPIServiceName, metav1.GetOptions{})
```
smira: same ipv4/ipv6 question here

internal/app/machined/pkg/controllers/k8s/endpoint.go
shanduur: I've split that into two separate EndpoinSlices. Those are named `talos-ipv4` and `talos-ipv6` respectively.

smira: what about this one? it's what Kubernetes generates!

internal/app/machined/pkg/controllers/k8s/endpoint.go
shanduur: I've added a note to that part, we need a separate handler logic for that in the future. As of now, as we discussed, I'm leaving Endpoints in the k8s. Also I've opened PR to K/K to add missing RBAC rules.

PR: https://github.com/kubernetes/kubernetes/pull/133057
Slack 🧵: https://kubernetes.slack.com/archives/C0EN96KUY/p1752845789653939

internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
shanduur: Fixed!

internal/app/machined/pkg/controllers/k8s/endpoint.go
shanduur: I've discussed that, and this part is being dropped. Kubelet should (and will not have) endpoint permissions in future releases. We need to add the permissions ourselves.

internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
```diff
@@ -197,70 +206,169 @@ func (ctrl *EndpointController) ensureTalosService(ctx context.Context, client *
...
 
 		if apierrors.IsNotFound(err) {
-			newEndpoints = &corev1.Endpoints{ //nolint:staticcheck
+			newEndpointSlice = &discoveryv1.EndpointSlice{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      constants.KubernetesTalosAPIServiceName,
+					Name:      name,
 					Namespace: constants.KubernetesTalosAPIServiceNamespace,
 					Labels: map[string]string{
 						"provider":  constants.KubernetesTalosProvider,
```
smira: hmm... I thought endpoint slices had a label which map it to the service?

internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
```diff
@@ -197,70 +206,169 @@ func (ctrl *EndpointController) ensureTalosService(ctx context.Context, client *
...
-			return fmt.Errorf("error updating Kubernetes Talos API endpoints: %w", err)
+			return fmt.Errorf("error updating Kubernetes Talos API endpoint slices: %w", err)
+		}
+	}
+}
+
+//nolint:gocyclo
+func (ctrl *EndpointController) cleanupTalosEndpoints(ctx context.Context, logger *zap.Logger, client *kubernetes.Client) error {
+	for {
+		err := client.CoreV1().Endpoints(constants.KubernetesTalosAPIServiceNamespace).Delete(ctx, constants.KubernetesTalosAPIServiceName, metav1.DeleteOptions{})
```
smira, line 356: does controller-manager remap endpointslice  back to endpoints, or am I thinking about it too much? :)

shanduur: No, it's a one way operation.

pkg/algorithm/slices.go
```diff
@@ -0,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smira: let's move it here please: https://github.com/siderolabs/gen/tree/main/xslices 

it would be a better home for it

shanduur: https://github.com/siderolabs/gen/pull/29

go.mod
```diff
@@ -137,7 +137,7 @@ require (
 	github.com/siderolabs/crypto v0.6.3
 	github.com/siderolabs/discovery-api v0.1.6
 	github.com/siderolabs/discovery-client v0.1.12
-	github.com/siderolabs/gen v0.8.4
+	github.com/siderolabs/gen v0.8.5-0.20250721103949-044d921685bb
```
shanduur: Do we need to release the `github.com/siderolabs/gen` before merging this PR?

internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
shanduur: But we have full control over names here right?

go.mod
smira: discussed in private - we need a tag, as we're very close to beta

go.mod
```diff
@@ -137,7 +137,7 @@ require (
 	github.com/siderolabs/crypto v0.6.3
 	github.com/siderolabs/discovery-api v0.1.6
 	github.com/siderolabs/discovery-client v0.1.12
-	github.com/siderolabs/gen v0.8.4
+	github.com/siderolabs/gen v0.8.5
```
smira, line 140: please bump also in `pkg/machinery/go.mod`, so that they are in sync

go.mod
```diff
@@ -194,6 +194,8 @@ require (
 	sigs.k8s.io/yaml v1.5.0
 )
 
+require k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
```
smira: let's move it into the block above

internal/app/machined/pkg/controllers/k8s/endpoint.go
```diff
@@ -216,7 +229,36 @@ func (ctrl *EndpointController) updateEndpointsResource(ctx context.Context, r c
...
+				addrIPv6 = addrs
+
+			case discoveryv1.AddressTypeFQDN:
+				fallthrough
+
+			default:
+				// ignore all other cases
+			}
+
+			r.TypedSpec().Addresses = append(addrIPv4, addrIPv6...) //nolint:gocritic
```
smira: nit, we could use `slices.Concat` here as I think it matches more closely semantic here (we don't want to mutate `addrIPv4)

internal/app/machined/pkg/controllers/k8s/endpoint.go
```diff
@@ -216,7 +229,36 @@ func (ctrl *EndpointController) updateEndpointsResource(ctx context.Context, r c
 				logger.Debug("updated controlplane endpoints", zap.Any("endpoints", addrs))
 			}
 
-			r.TypedSpec().Addresses = addrs
+			var (
+				addrIPv4 = []netip.Addr{}
```
smira: nit, we can even not allocate them to empty and leave as `nil`:

```suggestion
	var			addrIPv4, addrIPv6 netip.Addr
```

smira:
on thing I noticed, I don't see an e2e full test for `kubeaccess` test in `integration/api/serviceaccount.go`, so I wonder if we should add something like that, e.g. that `talosctl version` completes successfully, by running a Job (?) 

we could use I guess `talosctl` container image for the pod

PR #11388, opened 2025-07-18
test: use Grype to scan SBOM for vulnerabilities

dsseng:
Add a CI job to build the current VEX file and scan SBOM.

This should enable automatic detection of new vulnerabilities once
information on them becomes available.


.kres.yaml
```diff
@@ -258,6 +258,30 @@ spec:
             artifactPath: _out
         - name: ci-temp-release-tag
         - name: lint
+    - name: grype-scan
```
dsseng, line 261: TODO: add a cron job as well

Dockerfile
```diff
@@ -933,6 +933,31 @@ COPY --from=sbom-amd64 / /
...
 FROM sbom-container-${TARGETARCH} AS sbom-container-target
 
+# Use an unpinned latest version, because we want to use the latest advisories
+FROM ghcr.io/siderolabs/generate-vex:latest AS talos-vex
+
+FROM build-go AS vex-generate
+ARG TAG
+COPY --from=talos-vex /generate-vex /generate-vex
+# We do not need reproducible timestamps here, as we only use VEX for scanning
+# FIXME: add an output option to not require a shell in this step
```
dsseng: Well, might just drop this: by the time we scan we need a tools image anyway

Dockerfile
```diff
@@ -933,6 +933,31 @@ COPY --from=sbom-amd64 / /
...
+RUN /generate-vex gen --target-version $TAG > /talos.vex.json
+
+FROM scratch AS vex
+COPY --from=vex-generate /talos.vex.json /talos.vex.json
+
+FROM build-go AS grype-scan
+COPY --from=sbom-arm64 /talos-arm64.spdx.json /talos-arm64.spdx.json
+COPY --from=vex /talos.vex.json /talos.vex.json
+COPY ./tools ./tools
+# TODO: should we cache DB, how often is it updated?
```
dsseng: Currently: not caching

Dockerfile
```diff
@@ -933,6 +933,31 @@ COPY --from=sbom-amd64 / /
...
+RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool -modfile=tools/go.mod \
+    github.com/anchore/grype/cmd/grype sbom:/talos-arm64.spdx.json \
+    --vex /talos.vex.json -vv --fail-on negligible 2>&1 | tee /grype-scan.log
```
dsseng: Here we need to both fail and produce a readable log file

frezbo: make this a local target which passes, and we can parse the output file to see if we need to fail

dsseng: This makes the most sense I guess

Dockerfile
dsseng: That'd be great if we at least cache between the two runs (list vulns and later fail on unknown ones)

.kres.yaml
```diff
@@ -258,6 +258,34 @@ spec:
...
+        - name: download-artifacts
+          artifactStep:
+            type: download
+            artifactName: talos-artifacts
+            artifactPath: _out
+        - name: ci-temp-release-tag
+        - name: local-grype-scan-result
+          environment:
+            DEST: _out
+        # Fail the run if found vulnerabilities which do not have a statement
```
dsseng, line 281: Here's an explanation:

Listed vulnerabilities have been acknowledged in https://github.com/siderolabs/talos-vex, so they do not fail the build. The report in artifacts, however, does list those, because they're `affected` or `under_investigation` and are of the interest to users. However, users are acknowledged of those by the advisories, so maintenance duties on the side of Talos are done (as those are tracked now) and the CI should pass again.

image

Makefile
```diff
@@ -28,6 +28,8 @@ TOOLS_PREFIX ?= ghcr.io/siderolabs/tools
 TOOLS ?= v1.11.0
 PKGS_PREFIX ?= ghcr.io/siderolabs
 PKGS ?= v1.11.0
+GENERATE_VEX_PREFIX ?= ghcr.io/dsseng/generate-vex
```
shanduur:
```suggestion
GENERATE_VEX_PREFIX ?= ghcr.io/siderolabs/generate-vex
```

tools/go.mod
```diff
@@ -2,26 +2,42 @@ module github.com/siderolabs/talos/tools
...
-tool github.com/anchore/syft/cmd/syft
+tool (
+	github.com/anchore/grype/cmd/grype
+	github.com/anchore/syft/cmd/syft
+)
 
 replace (
-	github.com/anchore/syft => github.com/dsseng/syft v1.26.2-0.20250610160657-e11602f64241
+	github.com/anchore/grype => github.com/dsseng/grype v0.96.1-0.20250718094733-2197792a3b26
+	github.com/anchore/syft => github.com/dsseng/syft v1.26.2-0.20250703101014-f39c35d156d9
```
shanduur: Can you maybe leave a comment why is there a replace for Grype and Syft?

Makefile
dsseng: Yeah, that's in a HACK commit I'll drop once talos-vex changes are ready. This is just to test CI through. Thanks for reminder so I don't squash that in accidentally

tools/go.mod
dsseng: Yes, I should. They are replaced with my not reviewed yet PRs in Syft (reproducible IDs and SOURCE_DATE_EPOCH support) and Grype (allow applying VEX to an SBOM source)

PR #11413, opened 2025-07-22
fix: do not download artifacts for cron Grype scan

dsseng:
SBOM will be generated in the same Docker run.


.kres.yaml
```diff
@@ -261,20 +261,12 @@ spec:
     - name: grype-scan
       buildxOptions:
         enabled: true
-      depends:
-        - default
```
smira: should we keep the dependency here more to "serialize" the steps, so that we don't do buildx build connurrently?

smira (approved):

smira (approved):

smira (approved):

.kres.yaml
dsseng: I can keep it, if that helps performance

PR #11419, opened 2025-07-23
chore(ci): nvidia try us-west-2 region

frezbo:
Try us-west-2 region for better instance availability.


frezbo:
Fixed issue in us-east-1

PR #11430, opened 2025-07-25
added docs for insecure, etcd-metrics, inlinemanifests and extraManifests

Iheanacho-ai:
# Pull Request

## What? (description)
I added the docs for insecure, etcd metrics, inlineManifests and extraManifests

## Why? (reasoning)
Because the customers were asking for them and we did not have any docs on them

## Acceptance

### DIFF
--- a/website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
+++ b/website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
@@ -1,0 +1,62 @@
+---
+title: "Expose the Etcd Metrics Endpoint"
+description: "Learn how to the expose etcd metrics endpoint."
+---
+
+To allow monitoring tools to collect metrics from your etcd database, you need to explicitly expose the etcd metrics endpoint.
+
+Here's how to do it:
+
+1. Create a patch file named `etcd-metrics-patch.yaml` that exposes the etcd metrics endpoint on `port:2381`, accessible from all network interfaces
+
+    ```shell
+    cat << EOF > etcd-metrics-patch.yaml
+    - op: add
+    path: /cluster/etcd/extraArgs
+    value:
+        listen-metrics-urls: http://0.0.0.0:2381
+    EOF
+    ```
+
+1. Create a `CP_IPS` variable that contains the IP addresses of your control plane nodes:
+
+    ```bash
+    CP_IPS="<control-plane-ip-1>,<control-plane-ip-2>,<control-plane-ip-3>"
+    ```
+
+   **Note**: Don’t use a load balancer or VIP endpoint for the control-plane-ips.
+   It should be calling the Talos API directly.
+
+1. Run this command to export your `TALOSCONFIG` variable.
+   You can skip this step if you've already done it:
+
+    ```bash
+    mkdir -p ~/.talos
+    cp ./talosconfig ~/.talos/config
+    ```
+
+1. Apply the `etcd-metrics-patch.yaml` patch to your control plane nodes:
+
+    ```bash
+    talosctl patch machineconfig \
+    --patch @etcd-metrics-patch.yaml \
+    --endpoints $CP_IPS \
+    --nodes $CP_IPS
+    ```
+
+1. Reboot the nodes.
+   Note that if you have only one control plane node, rebooting it will cause cluster downtime.
+
+    ```bash
+    for NODE in $(echo "${CP_IPS}" | tr ',' ' '); do
+        echo "Rebooting control plane node: $NODE"
+        talosctl reboot --endpoints "$NODE" --nodes "$NODE" --wait
+    done
+    ```
+
+1. After the node reboots, run the following command to confirm that the etcd metrics endpoint is accessible:
+
+    ```bash
+    CP_IP=$(echo $CP_IPS | cut -d',' -f1)
+    curl "${CP_IP}:2381/metrics"
+    ```
--- a/website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
+++ b/website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
@@ -1,0 +1,138 @@
+---
+title: "inlineManifests and extraManifests"
+description: "Learn what inlineManifests and extraManifests are, how they differ, and why they matter."
+---
+
+`inlineManifests` and `extraManifests` allow you to automatically apply Kubernetes resources to your cluster during installation and upgrades.
+
+Both are designed to automate the provisioning of components like CNIs and other static infrastructure, but they differ in how the manifest content is sourced and applied.
+They are not meant for deploying applications or frequently changing services.
+For those, it's better to use a GitOps or CI/CD tool.
+
+## inlineManifests
+
+`inlineManifests` are defined directly within the machine configuration file.
+The YAML content is embedded inside the `inlineManifests` section, making it ideal for tightly coupled resources that need to be provisioned as soon as the node boots up.
+
+Here’s an example of how to configure a cluster using an `inlineManifest`:
+
+```yaml
+cluster:
+  inlineManifests:
+    - name: my-app
+      contents: |
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: my-application
+        spec:
+          # ... deployment specification
+```
+
+## extraManifests
+
+`extraManifests` are Kubernetes manifests fetched from external, unauthenticated HTTP sources such as GitHub, raw file servers, or gists.
+You define them in the `extraManifests` section of the machine configuration.
+They’re best suited for shared, versioned, or centrally managed resources.
+These manifests are pulled directly by the node during configuration.
+If the node doesn’t have network access to the HTTP endpoint hosting the manifest the installation will fail.
+
+Similarly, if the endpoint is down or returns an error, the manifest will not be applied, and the machine configuration will fail as a result.
+
+Here’s how to reference `extraManifests`:
+
+```bash
+cluster:
+  extraManifests:
+    - "https://raw.githubusercontent.com/example/repo/main/manifest.yaml"
+    - "https://gist.githubusercontent.com/user/gist-id/raw/manifest.yaml"
+```
+
+## Usage Example
+
+A common use case for `inlineManifests` or `extraManifests` is to install a GitOps controller like Flux or ArgoCD.
+Once the controller is running, it connects to your Git repository and automatically applies the rest of your Kubernetes configuration.
+
+Here's how to install the Flux GitOps controller using an `extraManifest`:
+
+1. Create a patch file named `flux-extra-manifest.yaml` that automatically downloads and applies the Flux installation manifest from GitHub:
+
+    ```shell
+    cat << EOF > flux-extra-manifest.yaml
+    cluster:
+    extraManifests:
+        - "https://github.com/fluxcd/flux2/releases/latest/download/install.yaml"
+    EOF
+    ```
+
+1. Create a `CP_IPS` variable that contains the IP addresses of your control plane nodes:
+
+    ```bash
+    CP_IPS="<control-plane-ip-1>,<control-plane-ip-2>,<control-plane-ip-3>"
+    ```
+
+1. Run this command to export your `TALOSCONFIG` variable.
+You can skip this step if you've already done it:
+
+    ```bash
+    mkdir -p ~/.talos
+    cp ./talosconfig ~/.talos/config
+    ```
+
+1. Apply the `flux-extra-manifest.yaml` patch to your control plane nodes:
+
+    ```bash
+    talosctl patch machineconfig \
+    --patch @flux-extra-manifest.yaml \
+    --endpoints $CP_IPS \
+    --nodes $CP_IPS
+    ```
+
+1. Reboot the nodes.
+Note that if you have only one control plane node, rebooting it will cause cluster downtime.
+
+    ```bash
+    for NODE in $(echo "${CP_IPS}" | tr ',' ' '); do
+        echo "Rebooting control plane node: $NODE"
+        talosctl reboot --endpoints "$NODE" --nodes "$NODE" --wait
+    done
+    ```
+
+1. Wait a few seconds and check for the Flux pods:
+
+    ```bash
+    kubectl get pods -n flux-system -w
+    ```
+
+## Omni Patches
+
+You can also apply `inlineManifests` or `extraManifests` patches to Talos clusters managed by Omni.
+
+Refer to [Create a Patch For Cluster Machines](https://omni.siderolabs.com/how-to-guides/create-a-patch-for-cluster-machines?q=inline+manifest&ask=true) to learn how to create and apply the patches.
+
+## Summary: inlineManifests vs extraManifests
+
+Here’s a quick overview of the key differences between `inlineManifests` and `extraManifests`:
+
+|                        | `inlineManifests`                              | `extraManifests`                                             |
+| ---------------------- | ---------------------------------------------- | ------------------------------------------------------------ |
+| Source                 | Defined directly in the machine configuration  | Pulled from external URLs (GitHub gists, web servers, gists) |
+| Configuration Location | Under the `inlineManifests` section.           | Under the `extraManifests` section                           |
+| Usecase                | Early bootstrapping of critical resources      | For reusable, version-controlled, or shared manifests        |
+| Benefits               | No external dependencies                       | Centrally managed                                            |
+| Disadvantages          | Difficult to maintain and format embedded YAML | Requires external HTTP server                                |
+
+## Key Considerations
+
+One important thing to note is that Talos treats upgrades the same way it handles fresh installs.
+During an upgrade, Talos installs the new version onto a separate partition and then switches to it.
+As a result, any `inlineManifests` or `extraManifests` defined in the machine configuration will be re-applied.
+This can unintentionally overwrite or downgrade components that were manually updated during the cluster's lifetime.
+
+### Recommendations
+
+To manage these current limitations, we recommend that you:
+
+- Avoid placing version-sensitive applications in `inlineManifests` or `extraManifests`, unless absolutely necessary.
+- Only include manifests for components that are unlikely to change throughout the life of your cluster.
+- Apply the manifests once, then remove them from the machine configuration and manage them as standard Kubernetes resources going forward.
--- a/website/content/v1.10/talos-guides/configuration/insecure.md
+++ b/website/content/v1.10/talos-guides/configuration/insecure.md
@@ -1,0 +1,104 @@
+---
+title: "The insecure flag"
+description: "Learn how to use the insecure flag."
+---
+
+The `--insecure` flag is a per-command argument that allows you to communicate with the Talos API when a node is in maintenance mode, that is, before a node has been configured with a machine configuration or TLS certificates.
+
+This is necessary because the machine configuration contains all the security credentials and PKI setup that secure communication between client and node.
+Without these credentials, nodes in this maintenance state can only accept API calls through the insecure interface.
+
+Only a small subset of Talos API commands support the `--insecure` flag, specifically those required for initial setup and maintenance operations.
+However, once you've applied a machine config, you must stop using the `--insecure` flag for all subsequent operations.
+The node will now expect secure communication through a talosconfig file.
+
+**Note**: The `--insecure` flag is used in a different context by the `talosctl image cache-create` command.
+This command is not used for interacting with the Talos node, but for allowing access to insecure image registries that do not support TLS.
+
+## Supported Commands With the insecure Flag
+
+The following commands can be used with the `--insecure` flag:
+
+`talosctl apply-config`
+
+Use this command alongside the `--insecure` flag to apply a machine configuration for the first time.
+
+`talosctl version`
+
+Check the Talos version running on the node.
+
+`talosctl get`
+
+Retrieves resources from the node.
+
+**Note**: You **cannot** retrieve the following resources in insecure mode:
+
+| **Resource Name**                                | **Aliases**                            |
+| ------------------------------------------------ | -------------------------------------- |
+| apicertificates.secrets.talos.dev                | apicertificate ac acs                  |
+| auditpolicyconfigs.kubernetes.talos.dev          | auditpolicyconfig apc apcs             |
+| certsans.secrets.talos.dev                       | certsan csan csans                     |
+| deviceconfigspecs.net.talos.dev                  | deviceconfigspec dcs                   |
+| discoveryconfigs.cluster.talos.dev               | discoveryconfig dc dcs                 |
+| etcdrootsecrets.secrets.talos.dev                | etcdrootsecret ers                     |
+| etcdsecrets.secrets.talos.dev                    | etcdsecret es                          |
+| kubeletsecrets.secrets.talos.dev                 | kubeletsecret ks                       |
+| kubernetesdynamiccerts.secrets.talos.dev         | kubernetesdynamiccert kdc kdcs         |
+| kubernetesrootsecrets.secrets.talos.dev          | kubernetesrootsecret krs               |
+| kubernetessecrets.secrets.talos.dev              | kubernetessecret ks                    |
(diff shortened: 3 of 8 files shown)


website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,138 @@
...
+| Configuration Location | Under the `inlineManifests` section.           | Under the `extraManifests` section                           |
+| Usecase                | Early bootstrapping of critical resources      | For reusable, version-controlled, or shared manifests        |
+| Benefits               | No external dependencies                       | Centrally managed                                            |
+| Disadvantages          | Difficult to maintain and format embedded YAML | Requires external HTTP server                                |
+
+## Key Considerations
+
+One important thing to note is that Talos treats upgrades the same way it handles fresh installs.
+During an upgrade, Talos installs the new version onto a separate partition and then switches to it.
+As a result, any `inlineManifests` or `extraManifests` defined in the machine configuration will be re-applied.
```
smira: this is wrong. 

Talos logic is the following:

* it only creates missing Kubernetes resources
* Talos never updates or deletes Kubernetes resources

A sync for updated (but not removed) resources is described in https://www.talos.dev/v1.10/kubernetes-guides/upgrading-kubernetes/

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "The insecure flag"
+description: "Learn how to use the insecure flag."
+---
+
+The `--insecure` flag is a per-command argument that allows you to communicate with the Talos API when a node is in maintenance mode, that is, before a node has been configured with a machine configuration or TLS certificates.
```
smira: This gives the wrong impression about the security.

`--insecure` in this context means "don't verify the server's TLS certificate". Even in maintenance mode API runs over TLS, but the identity of both client and server are not verified.

Probably this section should start with a note that regular Talos API uses mutual TLS verification, while in maintenance mode client doesn't present any certificate, while Talos (server) uses self-signed TLS certificate, which `--insecure` mode skips a check for.

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,104 @@
...
+
+`talosctl version`
+
+Check the Talos version running on the node.
+
+`talosctl get`
+
+Retrieves resources from the node.
+
+**Note**: You **cannot** retrieve the following resources in insecure mode:
```
smira: I think this list will be always outdated, so it doesn't make any sense to include it in the docs.

`talosctl get rd` gives a list of resource definitions, those which have `sensitive: true` are not available

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,104 @@
...
+| volumeconfigs.block.talos.dev                    | volumeconfig vc vcs                    |
+
+`talosctl meta`
+
+Manages key-value pairs in the META partition.
+
+- `talosctl meta write`: Writes a key-value pair to the META partition.
+- `talosctl meta delete`: Deletes keys from the META partition.
+
+`talosctl reset`
```
smira, line 83: we should not document it here, or we should properly document it works over SideroLink only, but I would drop these commands

website/content/v1.11/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,104 @@
...
+
+`talosctl reset`
+
+Used alongside the `--insecure` flag to reset nodes in Omni.
+
+`talosctl upgrade`
+
+Used alongside the `--insecure` flag to upgrade Talos in Omni.
+
+Refer to the [CLI reference]({{< relref "../../reference/cli" >}}) for full CLI details.
```
smira: `talosctl wipe disk` is missing here in the list (for 1.11)

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,138 @@
...
+cluster:
+  inlineManifests:
+    - name: my-app
+      contents: |
+        apiVersion: apps/v1
+        kind: Deployment
+        metadata:
+          name: my-application
+        spec:
+          # ... deployment specification
```
jimmyjones2, line 29: Would it be useful to explain that despite being an array, you cannot rely on resources being applied in order. However you can use multi-doc syntax `---` if necessary

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
Iheanacho-ai: Oh my bad. Just to be sure, during upgrades:
1. Talos will create resources defined in your manifests if they do not exist in the cluster
2. Even if the resources exists, but has deviated from the manifest definition, Talos will leave the resources completely unchanged.
3. Talos WILL NOT delete resources from the cluster, even if they are no longer present in your manifest configuration.

smira: It is not connected in any way to upgrades directly, Talos attempts a sync on each boot, and if you change manifests.

Everything else is correct.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
Iheanacho-ai: Do you mean for the patches ?

jimmyjones2: Yes, like https://www.talos.dev/v1.10/kubernetes-guides/network/deploying-cilium/#method-5-using-a-job

website/content/v1.10/talos-guides/configuration/insecure.md
rothgar: Oh that's cool. I didn't know that.

website/content/v1.10/talos-guides/configuration/insecure.md
rothgar: Is there a way to filter based on that property? This is a common question we get in slack and it would be helpful if we could do something like
```
talosctl get rd --insecure
```
to see resources that are available in insecure mode. Otherwise we could write a quick for loop that gets them.

Iheanacho-ai: Yeah thats what i did

smira: You can do `talosctl get rd --insecure`, but it means getting resource definitions in insecure mode. You can do `-o yaml` and filter with `yq`, but I don't think anybody needs this for real in the docs. It will be outdated anyways. If they hit a resource which is not available, it's not available, and the message is clear about it.

website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
```diff
@@ -0,0 +1,62 @@
...
+    EOF
+    ```
+
+1. Create a `CP_IPS` variable that contains the IP addresses of your control plane nodes:
+
+    ```bash
+    CP_IPS="<control-plane-ip-1>,<control-plane-ip-2>,<control-plane-ip-3>"
+    ```
+
+   **Note**: Don’t use a load balancer or VIP endpoint for the control-plane-ips.
```
smira: this is a fine note, but I don't think it belongs here. In fact people don't use VIP for `--nodes`, they might use VIP for `--endpoints` (both are bad)

website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
```diff
@@ -0,0 +1,62 @@
...
+   It should be calling the Talos API directly.
+
+1. Run this command to export your `TALOSCONFIG` variable.
+   You can skip this step if you've already done it:
```
smira: does it belong here? should we just reference how to handle `talosconfig`?

website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
```diff
@@ -0,0 +1,62 @@
...
+        talosctl reboot --endpoints "$NODE" --nodes "$NODE" --wait
+    done
+    ```
+
+1. After the node reboots, run the following command to confirm that the etcd metrics endpoint is accessible:
+
+    ```bash
+    CP_IP=$(echo $CP_IPS | cut -d',' -f1)
+    curl "${CP_IP}:2381/metrics"
+    ```
```
smira, line 54: I think this tutorial ends before actually saying why you're doing all of this - if the goal is to add it to Prometheus - how to do that?

also this exposes metric to the world if the CP_IP is a public IP, so might not be be the best way

smira: or let me rephrase this - we should link to the Ingress Firewall docs

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
+---
+title: "inlineManifests and extraManifests"
+description: "Learn what inlineManifests and extraManifests are, how they differ, and why they matter."
+---
+
+`inlineManifests` and `extraManifests` allow you to automatically apply Kubernetes resources to your cluster during installation and upgrades.
```
smira:
```suggestion
`inlineManifests` and `extraManifests` allow you to automatically apply Kubernetes resources to your cluster during initial bootstrap.
```

probably better than upgrades

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
...
+        kind: Deployment
+        metadata:
+          name: my-application
+        spec:
+          # ... deployment specification
+```
+
+## extraManifests
+
+`extraManifests` are Kubernetes manifests fetched from external, unauthenticated HTTP sources such as GitHub, raw file servers, or gists.
```
smira, line 34: Talos actually supports sending custom HTTP headers, so it's not exactly unauthenticated always, but probably we can leave it like this

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
...
+  extraManifests:
+    - "https://raw.githubusercontent.com/example/repo/main/manifest.yaml"
+    - "https://gist.githubusercontent.com/user/gist-id/raw/manifest.yaml"
+```
+
+## Resource Ordering Considerations
+
+You must note that although `inlineManifests` and `extraManifests` are defined as arrays, Talos does not guarantee that the items will be applied in the order they appear.
+
+If your resources depend on one another and must be applied in a specific order, define them in a single manifest entry and separate each resource using the `---` YAML document separator.
```
smira: this whole section is not how Talos works I have no idea where it came from.

Talos sorts manifests in the following order:

* namespaces
* CRDs
* everything else in alphabetical order of `name:` property

Kubernetes doesn't technically require a specific order of manifests to apply, so it should be fine.

Plus even if one manifest fails to apply before another one is applied, Talos anyways retries the whole process, so it will eventually apply.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
...
+    ```
+
+1. Create a `CP_IPS` variable that contains the IP addresses of your control plane nodes:
+
+    ```bash
+    CP_IPS="<control-plane-ip-1>,<control-plane-ip-2>,<control-plane-ip-3>"
+    ```
+
+1. Run this command to export your `TALOSCONFIG` variable.
+You can skip this step if you've already done it:
```
smira, line 85: same comment as before, are we going to put this to every documentation page?

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
...
+| ---------------------- | ---------------------------------------------- | ------------------------------------------------------------ |
+| Source                 | Defined directly in the machine configuration  | Pulled from external URLs (GitHub gists, web servers, gists) |
+| Configuration Location | Under the `inlineManifests` section.           | Under the `extraManifests` section                           |
+| Usecase                | Early bootstrapping of critical resources      | For reusable, version-controlled, or shared manifests        |
+| Benefits               | No external dependencies                       | Centrally managed                                            |
+| Disadvantages          | Difficult to maintain and format embedded YAML | Requires external HTTP server                                |
+
+## How Talos Handles Manifest Resources During Upgrades
+
+During upgrades, Talos follows a conservative, additive-only approach when processing your `inlineManifests` and `extraManifests`.
```
smira: This should be rephrases, Talos reconciles manifests on every boot, on every failure to apply, on every change to the manifests in the machine config.

So "upgrade" there is irrelevant.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,176 @@
...
+Here’s what that means in practice:
+
+* **Creates missing resources**: If a resource defined in your manifests doesn't exist in the cluster, Talos will create it.
+
+* **Preserves existing resources**: Resources that already exist in the cluster are left completely unchanged, regardless of any differences between the current state and the manifest definition.
+
+* **Never deletes resources**: Talos will not remove resources from the cluster, even if they're no longer present in your manifest configuration
+
+This means that any manual changes, updates, or customizations made to these resources after the initial deployment will persist through Talos upgrades.
```
smira: missing a reference to the Upgrade Kubernetes docs on how to apply updated manifests which I provided in the previous round of reviews

website/content/v1.10/talos-guides/configuration/insecure.md
smira: drop it here... it doesn't work with Omni via CLI, Omni uses this API internally.

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,109 @@
...
+
+`talosctl meta`
+
+Manages key-value pairs in the META partition.
+
+`talosctl reset`
+
+Resets the nodes in Omni.
+
+`talosctl upgrade`
```
smira: drop it

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,109 @@
...
+
+Upgrades the Talos versions  in the nodes in Omni.
+
+`talosctl wipe disk`
+
+Erase data from disk partitions on a Talos node.
+
+Refer to the [CLI reference](https://www.talos.dev/v1.10/reference/cli/) for full CLI details.
```
smira: this should be relref... Please don't use any direct links ever inside the docs

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
rothgar: Does it also do that sorting for extraManifests? I didn't realized it would sort resources. That's really cool.

website/content/v1.10/kubernetes-guides/configuration/etcd-metrics.md
rothgar: Good point. We should link to those docs on how to secure the port.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
smira: it doesn't really matter where the manifests came from - Talos built-in, inline, extra - they are all processed in the end by extracting all resources and then applying resource-by-resource in the order above

rothgar: We'll make sure that info gets added to the docs

Iheanacho-ai: I got it from a comment by @jimmyjones2 in the previous round of reviews. I will delete

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
Iheanacho-ai: I don't like that its there, but Justin made a point that I should add it, as they cant successfully run this command if they do not export the TALOSCONFIG.  

But I will rephrase

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
Iheanacho-ai: Or I could just delete that paragraph since we are no longer talking about upgrades anymore

website/content/v1.10/talos-guides/configuration/insecure.md
Iheanacho-ai: My apologies, I added it as an external link in my Google docs, so when I exported as markdown I missed that link

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,149 @@
...
+```yaml
+cluster:
+  extraManifests:
+    - "https://raw.githubusercontent.com/example/repo/main/manifest.yaml"
+    - "https://gist.githubusercontent.com/user/gist-id/raw/manifest.yaml"
+```
+
+## Resource Ordering Considerations
+
+Talos automatically sorts all manifests, including `inlineManifests`, `extraManifests`, or `built-in manifests`, before applying them in the following order:
```
rothgar: What are built-in manifests?

smira: they should not be in backticks, and we should probably clarify that Talos has some built-in manifests, e.g. kubelet bootstrap token, CoreDNS, etc.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,149 @@
...
+1. Apply the `flux-extra-manifest.yaml` patch to your control plane nodes:
+
+    ```bash
+    talosctl patch machineconfig \
+    --patch @flux-extra-manifest.yaml \
+    --endpoints $CP_IPS \
+    --nodes $CP_IPS
+    ```
+
+1. Reboot the nodes.
```
rothgar: Can you confirm if a reboot is required for this patch? Because extra and inline manifests only apply during install or upgrade applying the patch and rebooting I don't think will trigger them to be installed if Talos is already installed and running.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,149 @@
...
+
+    ```bash
+    kubectl get pods -n flux-system -w
+    ```
+
+## Omni Patches
+
+You can also apply `inlineManifests` or `extraManifests` patches to Talos clusters managed by Omni.
+
+Refer to [Create a Patch For Cluster Machines](https://omni.siderolabs.com/how-to-guides/create-a-patch-for-cluster-machines?q=inline+manifest&ask=true) to learn how to create and apply the patches.
```
rothgar: This has a search query `?q` I think you want to remove.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,149 @@
...
+
+Here's what that means in practice:
+
+* **Creates missing resources**: If a resource defined in your manifests doesn't exist in the cluster, Talos will create it.
+
+* **Preserves existing resources**: Resources that already exist in the cluster are left completely unchanged, regardless of any differences between the current state and the manifest definition.
+
+* **Never deletes resources**: Talos will not remove resources from the cluster, even if they're no longer present in your manifest configuration.
+
+This means that any manual changes, updates, or customizations made to these resources after the initial deployment will persist through [Talos upgrades]({{< relref "../../kubernetes-guides/upgrading-kubernetes">}}).
```
smira: the link seems not to match the text. 

Talos upgrade is irrelevant here.

We should say that if you want to sync/update manifests after the initial cluster creation, one can use Kubernetes upgrade process by specifying upgrade to the same version as the running cluster, so that only manifests are procesed.

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,105 @@
...
+* The node uses a self-signed TLS certificate.
+* The client (talosctl) does not present a certificate.
+* Neither side can verify the other's identity.
+
+In this case, the `--insecure` flag tells `talosctl` to skip verifying the server’s certificate, allowing the connection to proceed.
+
+Only a small subset of Talos API commands support the --insecure flag, specifically those required for initial setup and maintenance operations.
+
+However, once you've applied a machine config, you must stop using the `--insecure` flag for all subsequent operations.
+The node will now expect secure communication through a talosconfig file.
```
rothgar: I would say "using certificates stored in"

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,105 @@
...
+
+**Note**: The `--insecure` flag is used in a different context by the `talosctl image cache-create` command.
+This command is not used for interacting with the Talos node, but for allowing access to insecure image registries that do not support TLS.
+
+## In Omni-Managed Clusters
+
+The `--insecure` flag works differently when you're using Omni to manage Talos clusters.
+
+Here, the flag is used for nodes that haven't joined a cluster yet.
+These nodes will only listen for communication over the SideroLink connection, a secure VPN point-to-point connection between Omni and the Talos node.
```
rothgar: "secure, point-to-point" add comma and remove VPN

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,105 @@
...
+
+So the SideroLink connection is the only way you can run commands against a node in insecure mode.
```
rothgar:
```suggestion
So the SideroLink connection is the only way you can run commands against a node connected to Omni.
```

website/content/v1.10/talos-guides/configuration/insecure.md
```diff
@@ -0,0 +1,105 @@
...
+
+This architecture provides a unique security advantage because if a machine is managed by Omni, you cannot send configurations to it from another machine, even if they are on the same network.
```
rothgar:
```suggestion
This architecture provides a unique security advantage because if a machine is managed by Omni, you cannot send configurations to it from another machine without authentication, even if they are on the same network.
```

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
smira: If you add a new manifest, not reboot is ever required, Talos will deploy it immediately.

website/content/v1.10/kubernetes-guides/configuration/inlinemanifests.md
```diff
@@ -0,0 +1,139 @@
...
+Here's what that means in practice:
+
+* **Creates missing resources**: If a resource defined in your manifests doesn't exist in the cluster, Talos will create it.
+
+* **Preserves existing resources**: Resources that already exist in the cluster are left completely unchanged, regardless of any differences between the current state and the manifest definition.
+
+* **Never deletes resources**: Talos will not remove resources from the cluster, even if they're no longer present in your manifest configuration.
+
+If you need to reapply or update these manifests after the cluster has been created, use the [Kubernetes upgrade]({{< relref "../../kubernetes-guides/upgrading-kubernetes">}}) process.
+Specify the same Kubernetes version that’s currently running in your cluster to ensure that only the manifests are reapplied, without performing a version upgrade."
```
smira:
```suggestion
Specify the same Kubernetes version that’s currently running in your cluster to ensure that only the manifests are reapplied, without performing a version upgrade.
```

smira (approved):

Issue #8972, opened 2024-07-05
Security Guarantees of LUKS Encryption using TPM for User Data on EPHEMERAL


stereobutter:
## Feature Request
Seal the LUKS encryption keys for the `EPHEMERAL` partition using a TPM register that depends on confidential information from the `STATE` partition.

### Description
Consider a [scenario](https://en.wikipedia.org/wiki/Evil_maid_attack) where an attacker has physical access to a cluster (as opposed to access to a lone disk removed from the cluster). I'm currently unsure whether using TPM-based LUKS encryption for both the `STATE` and `EPHEMERAL` partitions is sufficient to guarantee that user data on `EPHEMERAL` will not be readable by the attacker.

The boot process, as I understand it using SecureBoot and LUKS with TPM-based encryption, looks roughly as follows:

1. At system startup, UEFI is loaded. The UEFI firmware itself and its configuration are measured into PCR[0] and PCR[1].
2. UEFI measures the bootloader and its configuration into PCR[4] and PCR[5]. If PCR measurements up until this point are as expected, the bootloader is loaded.
3. The bootloader (systemd-boot) runs. The UKI content (includes the kernel, its config, the initrd, etc.) are measured into PCR[11].
4. If the PCR measurements up to this point are as expected, the kernel and initrd are loaded.
5. The kernel initializes hardware, etc., and may measure data into PCR[10].
6. If the PCR measurements up to this point are as expected, the TPM releases the encryption key for the STATE partition. The partition is decrypted and mounted.
7. The kernel executes `init` and talos takes over.
8. Talos runs through its start-up phases and [measures progress into PCR[11]](https://github.com/search?q=repo%3Asiderolabs%2Ftalos+PCRExtent&type=code) and can read the machine config from the now decrypted `STATE` partition.
9. If the PCR measurements up to this point are as expected, the TPM releases the encryption key for `EPHEMERAL`. The partition is decrypted and mounted as `/var`.

If I'm not mistaken, all PCR measurements up to step 9, where the encryption key for `EPHEMERAL` is released by the TPM, depend either on the physical device (CPU, TPM-Chip, etc.) or the current talos installer image used on the machine (Kernel, Bootloader, SecureBoot signature, etc.) and not on the *identity* of the machine/cluster. I believe this would enable an attacker to overwrite the `STATE` with their talos `STATE` partition (created by the attacker using the same installer image).

[...]

PR #11435, opened 2025-07-25
feat: implement encryption locking to STATE

### LINKED ISSUE #10676: Lock encrypted partition key to STATE encryption
Generate another unique token and store in `STATE`, it is different from the existing unique token in a way that it doesn't get exposed via Kubernetes (as `machine-id`).

Optionally lock the encryption key to this value.

See https://github.com/siderolabs/talos/issues/8972

smira:
Fixes #10676


website/content/v1.11/reference/configuration/block/volumeconfig.md
```diff
@@ -174,6 +174,7 @@ EncryptionKey represents configuration for disk encryption key.
 |`nodeID` |<a href="#VolumeConfig.encryption.keys..nodeID">EncryptionKeyNodeID</a> |Deterministically generated key from the node UUID and PartitionLabel.  | |
 |`kms` |<a href="#VolumeConfig.encryption.keys..kms">EncryptionKeyKMS</a> |KMS managed encryption key.  | |
 |`tpm` |<a href="#VolumeConfig.encryption.keys..tpm">EncryptionKeyTPM</a> |Enable TPM based disk encryption.  | |
```
jimmyjones2, line 176: Should this be updated to say tpm is not recommended for EPHEMERAL and UserVolumes due to the risk described in #8972?

smira: No, it shouldn't be updated this way, in fact with what this PR implements, it's completely safe to use it

internal/app/machined/pkg/adapters/secrets/encryption_salt.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+}
+
+type encryptionSalt struct {
+	*secrets.EncryptionSaltSpec
+}
+
+// Generate new encryption salt.
+func (a encryptionSalt) Generate() error {
+	buf := make([]byte, constants.DiskEncryptionSaltSize)
```
frezbo, line 30: i know systemd also extends the PCR with the disk uuid of the state, would it offer us anything extra?

pkg/machinery/config/types/block/volume_config.go
```diff
@@ -177,6 +178,13 @@ func (s *VolumeConfigV1Alpha1) Validate(validation.RuntimeMode, ...validation.Op
 		if !s.ProvisioningSpec.IsZero() {
 			validationErrors = errors.Join(validationErrors, fmt.Errorf("provisioning config is not allowed for the %q volume", s.MetaName))
 		}
+
+		for _, key := range s.EncryptionSpec.EncryptionKeys {
+			if pointer.SafeDeref(key.KeyLockToSTATE) {
+				// state-locked keys are not allowed
+				validationErrors = errors.Join(validationErrors, fmt.Errorf("state-locked key is not allowed for the %q volume", s.MetaName))
```
frezbo, line 185: so not supported for additional volumes?

website/content/v1.11/talos-guides/configuration/disk-encryption.md
```diff
@@ -126,8 +127,16 @@ Talos supports two kinds of keys:
 - `kms` which is sealed with the network KMS.
 - `tpm` which is sealed using the TPM and protected with SecureBoot.
 
-> Note: Use static keys only if your STATE partition is encrypted and only for the EPHEMERAL partition.
-> For the STATE partition it will be stored in the META partition, which is not encrypted.
+> Note: Use static keys only if your STATE partition is encrypted and only for the EPHEMERAL volume or user volumes.
```
frezbo: this is a bit confusing, probably reword to:

```
Use static keys for anything other than `STATE`, only if `STATE` is encrypted with anything other than static keys.
```

Ahh this is more confusing

internal/app/machined/pkg/adapters/secrets/encryption_salt.go
dsseng: I'd also consider this. Basically some of those PCR extensions act as a "no way back" to make sure after you've used the TPM-backed secret you progress those PCRs further to make the secret inaccessible.

pkg/machinery/config/types/block/volume_config.go
smira: not supported for `STATE` itself, you can't lock `STATE` to `STATE` for obvious reasons :)

internal/app/machined/pkg/adapters/secrets/encryption_salt.go
smira: I'm not sure how this is going to work for multiple volumes, and I think this is a bit different than this PR?

website/content/v1.11/talos-guides/configuration/disk-encryption.md
smira: yeah, need to rephrase, I agree with you

internal/app/machined/pkg/adapters/secrets/encryption_salt.go
smira: > I'd also consider this. Basically some of those PCR extensions act as a "no way back" to make sure after you've used the TPM-backed secret you progress those PCRs further to make the secret inaccessible.

we already do this - the PCRs are progressed once the TPM-backed volumes are unlocked

what this PR is trying to solve is a specific issue like this:

* I have TPM-encrypted STATE & EPHEMERAL
* SecureBoot is enabled, everything is good
* Malicious user can't boot any other OS, or access the data (Talos and Kubernetes API are not accessible)
* Now malicious user gets physical access to the disk of the machine
* They can now replace `STATE` with non-encrypted version of it with their own PKI
* Talos has no way to tell on boot that `STATE` has been replaced, so it will boot fine, and `EPHEMERAL` will be unlocked as well
* At this point I have access with Talos API, and I can access contents of EPHEMERAL

This PR blocks this, as replacing STATE will render any locked to STATE partitions unusable.

website/content/v1.11/talos-guides/configuration/disk-encryption.md
smira: rewrote this sentence

PR #11451, opened 2025-07-28
test: wait for service account test job longer

smira:
It seems we hit the 30s timeout with network chaos tests.


smira:
This seems to help:

```
        --- PASS: TestIntegration/api.ServiceAccountSuite/TestValid (39.71s)
```

Issue #11458, opened 2025-07-29
detect and report correctly booted entry


smira:
TBD: discover booted entry with kexec and with bootloader menu selection correctly

PR #11482, opened 2025-08-01
feat: bootedentry resource

frezbo:
Add a new bootedentry resource that shows the current booted UKI.

Fixes: #11458


internal/app/machined/pkg/controllers/runtime/booted_entry.go
```diff
@@ -0,0 +1,147 @@
...
+// Run implements controller.Controller interface.
+//
+//nolint:gocyclo,cyclop
+func (ctrl *BootedEntryController) Run(ctx context.Context, r controller.Runtime, _ *zap.Logger) error {
+	// If we're booted in Container mode, short-circuit the controller.
+	if ctrl.V1Alpha1Mode == machineruntime.ModeContainer {
+		return nil
+	}
+
+	for {
```
dsseng, line 64: Maybe this could just run one-shot somehow? Just wait for the first SecurityState appearing, and then create the resource and exit?

dsseng (approved):
Maybe this could be exercised in integration tests (at least partially)? Would be great

internal/app/machined/pkg/controllers/runtime/booted_entry.go
frezbo: this only runs once

dsseng: ah, okay

PR #11486, opened 2025-08-01
fix: one more attempt to fix volume mount race on restart

smira:
The issue seems to be around still racy service restarts which leads to a potential hang on a conflicting state.

By not re-using the mount request IDs on each restart, hopefully we can improve on it.


Orzelius (approved):
🤞

frezbo (approved):

PR #11503, opened 2025-08-04
docs: added the deploy application docs

Iheanacho-ai:
# Pull Request

## What? (description)
i added the docs on deploying a first workload

## Why? (reasoning)
To help users get started
## Acceptance


website/content/v1.10/introduction/deploy-first-workload.md
```diff
@@ -0,0 +1,67 @@
...
+1. Verify that your pods and services are running:
+
+    ```bash
+    kubectl get pods,services # Lists the deployed pods and services
+    ```
+
+1. Create a NODE_IP variable by retrieving an IP address of any one of your nodes:
+
+    ```bash
+    NODE_IP=$(kubectl get nodes -o jsonpath='{.items[0].status.addresses[?(@.type=="InternalIP")].address}'; echo)
```
rothgar, line 28: Will this work if they don't schedule workloads on control plane nodes and the use an IP address from one of the control plane nodes? I don't think kube-proxy will be running so the node IP wouldn't work.

We could try to filter this to worker nodes but that makes the assumption that they have worker nodes. Does anyone have ideas to make this consistently work for single node clusters, multi-node with CP scheduling, multi-node without CP workload scheduling?

Iheanacho-ai: Will it be too much to ask them to input the node IP address themselves? I think in this scenario its okay.

rothgar: Let's just leave it as is.

My other idea was to get the node where the pod is running and then get the IP address of that node. It would work consistently, but is a bit complicated and hopefully the command you have works most of the time.

PR #11520, opened 2025-08-06
chore: rekres to configure slack notify workflow for CI failures

oguzkilcan:
Rekres to configure slack notify workflow for CI failures on main and scheduled jobs


smira (approved):
I'll fix grype in a moment

smira:
@oguzkilcan I fixed grype scane

PR #11553, opened 2025-08-11
chore: refactor how tools are being installed

smira:
Move stuff into `tools/go.mod`.

Also fix linting issues on the way (updating golangci-lint).


pkg/machinery/api/cluster/cluster_grpc.pb.go
```diff
@@ -90,7 +90,7 @@ type UnsafeClusterServiceServer interface {
 }
 
 func RegisterClusterServiceServer(s grpc.ServiceRegistrar, srv ClusterServiceServer) {
-	// If the following call panics, it indicates UnimplementedClusterServiceServer was
+	// If the following call pancis, it indicates UnimplementedClusterServiceServer was
```
oguzkilcan, line 93: How did this happen?

smira: I guess a bug in the proto generator, as it's generated code

smira:
> Suggestion: maybe `.git-blame-ignore-revs` could be created to hide such changes from blame?

In fact, this won't work in this PR, as I need to put there a revision which itself would change the commit sha :)

PR #11558, opened 2025-08-12
chore: re-enable vulncheck

smira:
This reverts commit 6186d182189d229e3065631076f435d34bfc4f53.

It was disabled to make v1.11.0-beta.2 out, depends on https://github.com/golang/vulndb/issues/3860

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -561,7 +561,7 @@ lint-%: ## Runs the specified linter. Valid options are go, protobuf, and markdo
 	@$(MAKE) target-lint-$* PLATFORM=linux/$(ARCH)
 
 lint: ## Runs linters on go, vulncheck, deadcode, protobuf, and markdown file types.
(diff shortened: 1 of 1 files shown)


shanduur:
We could ignore that CVE, as according to [GHSA](https://github.com/moby/moby/security/advisories/GHSA-4vq8-7jfc-9cvp), it is affecting only <28.0.0? 

Reusing the following: https://github.com/siderolabs/image-factory/blob/9760ab0/hack/govulncheck.sh

smira:
> We could ignore that CVE, as according to [GHSA](https://github.com/moby/moby/security/advisories/GHSA-4vq8-7jfc-9cvp), it is affecting only <28.0.0?
> 
> Reusing the following: https://github.com/siderolabs/image-factory/blob/9760ab0/hack/govulncheck.sh

I hope we can still have it fixed upstream and skip hacks.

PR #11559, opened 2025-08-12
fix: dial with proxy

### LINKED ISSUE #11536: talosctl fails with multiple endpoints with proxy is being used
See https://github.com/siderolabs/talos/discussions/11529

Reproducer: use a `talosconfig` with multiple endpoints.

Start a mock proxy with `talosctl debug airgapped`

```shell
$ HTTP_PROXY=http://localhost:8002 HTTPS_PROXY=http://localhost:8002 talosctl -n 172.20.0.2 get manifests
rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for talos-default-controlplane-1, not 172.20.0.2:50000,172.20.0.3:50000,172.20.0.4:50000"
```

The difference between one endpoint and multiple endpoints comes from https://github.com/siderolabs/talos/blob/d306713a13a18d7af6caffd5890d54d91d22cad7/pkg/machinery/client/connection.go#L155-L170

It looks like identity for TLS check comes from some bug related to mix of proxy and loadbalancer.

Also we have a fork of a proxy dialer, see https://github.com/siderolabs/talos/blob/d306713a13a18d7af6caffd5890d54d91d22cad7/pkg/grpc/dialer/proxy.go#L41-L46

shanduur:
Fixes #11536


cmd/talosctl/cmd/talos/root.go
```diff
@@ -35,16 +36,20 @@ var GlobalArgs global.Args
 
 const pathAutoCompleteLimit = 500
 
+var defaultDialOpts = []grpc.DialOption{
+	grpc.WithContextDialer(dialer.DynamicProxyDialer),
```
smira: I wonder if we should inject that in the client itself (?) or is it going to be too much?

smira: I mean to make it a default option for the `pkg/machinery/client.New` so that it applies to everyone, including non-talosctl users

shanduur: This would require breaking change in `pkg` (which is public), because (IMHO) the best solution is to move the dialer package to `pkg/machinery` which is separate module.

smira: yes, let's move dialer into machinery, and make a change to enable this by default with an option to disable it.

the only case we don't need it is when we dial to a unix socket I guess

PR #11579, opened 2025-08-14
fix: kubernetes upgrade options

alperencelik:
## What? (description)

The PR aims to fix and enhance the Kubernetes upgrade options.

## Why? (reasoning)

Currently, the upgradeKubelet option is tied to the worker nodes, which I believe is a mistake. Instead, it should be tied to the upgradeKubelet function, with a separate option added to toggle the upgrade of worker nodes.

## Acceptance

### DIFF
--- a/pkg/cluster/kubernetes/talos_managed.go
+++ b/pkg/cluster/kubernetes/talos_managed.go
@@ -75,7 +75,7 @@ func (options *UpgradeOptions) Validate() error {
 
 // Upgrade the Kubernetes control plane components, manifests, kubelets.
 //
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func Upgrade(ctx context.Context, cluster UpgradeProvider, options UpgradeOptions) error {
 	if err := options.Validate(); err != nil {
 		return fmt.Errorf("invalid upgrade options: %w", err)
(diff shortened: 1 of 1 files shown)


smira (commented):
I see your point: even if `upgrade-kubelet` is false, it will still upgrade it on controlplanes, but I think what you propose is even more confusing.

I'm fine to make the command skip kubelet completely even on controlplanes if `upgrade-kubelet` is false

alperencelik:
Actually, the reason I’d like to add a separate flag for the worker nodes is to be able to upgrade them individually. I agree that this doesn’t address the upgradeKubelet issue, but I thought it could still be useful feature for certain edge cases.

If you think that's unnecessary, I can update the PR for only upgradeKubelet condition fix.

smira:
I mean that it makes it even more confusing I think on what is going to be done with so many flags.

In general, kubelet update is the least scary part as it's graceful and kubelet <> kube-apiserver have API compatibility across versions.

alperencelik:
I think I get what you mean, and it sounds reasonable. For now, I’ve just updated the upgradeKubelet options to be checked before upgradeKubelet function.  The only problem that I see right now is checking on the same condition twice but I wasn't able to find another solution tbh.

Issue #10677, opened 2025-04-08
Drop PCR7 from the TPM encryption policy


smira:
### Description

Ref: https://github.com/systemd/systemd/commit/4b840414be3b2d6520599d86d2b718a37574aabf

From Talos side we could also lock to the state of just secureboot is enabled/disabled

### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`]
- Kubernetes version: [`kubectl version`]
- Platform:

PR #11582, opened 2025-08-15
fix: tpm2 luks header metadata

frezbo:
Talos TPM2 based disk encryprion was locking to both PCR7 and PCR11, but we were only showing PCR11 in the luks header metadata. Fix it to include both PCR7 and PCR11. This data was never used anywhere, but we would need this for implementing #10677.


smira:
I guess for backwards compatibility we would need to add some "version" field so that we can correctly interpret "legacy" metadata with wrong PCR list.

frezbo:
> I guess for backwards compatibility we would need to add some "version" field so that we can correctly interpret "legacy" metadata with wrong PCR list.

another option is to add a new resource, so we can know we're upgrading from or use a quirk?

smira:
Not sure about a resource, I guess the only piece of information we have is the LUKS token, and we need to figure out from it only how to unseal/unlock it?

frezbo:
Closing, going to use v1 and v2 versions for the luks token header

smira:
I think we don't need this anymore.

Issue #11608, opened 2025-08-18
stop mounting `STATE` volume/partition to the mountpoint


smira:
Only Talos consumes `STATE` contents, so we can open the filesystem without ever mounting it to the `/system/state` mountpoint, and pass around opentree fd in the `MountStatus` resource so that consumers can write/read to `STATE`, and `STATE` is never exposed in the root mount namespace.

PR #11620, opened 2025-08-19
chore: use new filesystem api on STATE partition

shanduur:
Prepares for mount refactoring (part of https://github.com/siderolabs/talos/issues/11608)

### DIFF
--- a/internal/pkg/xfs/opentree_test.go
+++ b/internal/pkg/xfs/opentree_test.go
@@ -1,0 +1,133 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build unix
+
+package xfs_test
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/blang/semver/v4"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+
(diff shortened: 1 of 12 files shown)


internal/pkg/xfs/root.go
```diff
@@ -0,0 +1,253 @@
...
+	"math/rand"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+var random = struct {
```
smira: I think this might better be `sync.OnceValue` to avoid the `init()` block

and another question - why can't we use `math/rand/v2` which I believe is always initialized and good to go?

internal/pkg/xfs/opentree_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func hasKernel(tb testing.TB, kernelVersion string) bool {
+	tb.Helper()
+
+	capabilityAt, err := semver.Parse(kernelVersion)
+	require.NoError(tb, err)
+
+	buf := new(unix.Utsname)
+
+	err = unix.Uname(buf)
```
frezbo, line 126: we can just read from constants.Kernel?

shanduur: This is unit test, and runs on the dev host / GitHub Actions host.

frezbo: oops, I missed that completely 👍

frezbo (approved):

PR #11628, opened 2025-08-20
fix: make --with-uuid-hostnames functionality available to qemu provider

oguzkilcan:
Make --with-uuid-hostnames flag functionality available to qemu provider on `talosctl cluster create`

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/create_test.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/create_test.go
@@ -6,6 +6,7 @@ package create //nolint:testpackage
 
 import (
 	"net/netip"
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


Unix4ever (approved):

Orzelius (commented):
Thank you! Would you be so kind and update the test as well please

cmd/talosctl/cmd/mgmt/cluster/create/create_test.go
```diff
@@ -178,3 +179,36 @@ func TestCreateNodeRequestsNames(t *testing.T) {
...
+	nodeIps1, err := getIps(cidr1, cOps)
+	assert.NoError(t, err)
+	nodeIps2, err := getIps(cidr2, cOps)
+	assert.NoError(t, err)
+
+	controlplanes, workers, err := createNodeRequests(cOps, resources, resources, [][]netip.Addr{nodeIps1, nodeIps2})
+	assert.NoError(t, err)
+
+	assert.Equal(t, 1, len(controlplanes))
+	assert.Equal(t, 1, len(workers))
```
Orzelius: I'm sorry for being a pain but I think I'd result in cleaner code if you either extracted the setup into a helper func or just added the assertions to the `TestCreateNodeRequestsNames` above

Orzelius:
> I'm a bit confused, this option was there, but it got accidentally disconnected?

yep, the flag itself was there but I accidentally removed the internal logic in the recent refactor 🤦

Orzelius (approved):

PR #11632, opened 2025-08-20
test: fix the Image Factory PXE boot test

smira:
The shell script expected `ipxe`, but we passed `pxe`, and it silently worked.


frezbo (approved):
💩 , good catch

PR #11623, opened 2025-08-19
feat: add a user-facing create qemu command 

Orzelius:
part of [#11404](https://github.com/siderolabs/talos/issues/11404)

I've split the changes into two commits for ease of review. Will squash before merging.

Stuff yet to do:
* extract more common logic and create a separate simplified code path for the this command (like already done for the docker command in [create_docker.go](https://github.com/siderolabs/talos/blob/main/cmd/talosctl/cmd/mgmt/cluster/create/create_docker.go))
* Change applicable e2e tests to use this command instead
* documentation updates (also need to update the docker docs)

PR #11640, opened 2025-08-21
feat: add human readable byte size cli flags

Orzelius:
Add flags for memory and disk sizes where people can specify the capacity freely in gb, mb tb etc. This change is backwards compatible and the values without units used previously still work.

requested here https://github.com/siderolabs/talos/pull/11623#discussion_r2287800355

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
@@ -13,6 +13,7 @@ import (
 
 	clustercmd "github.com/siderolabs/talos/cmd/talosctl/cmd/mgmt/cluster"
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/mgmt/helpers"
+	"github.com/siderolabs/talos/pkg/bytesize"
 	"github.com/siderolabs/talos/pkg/cli"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 )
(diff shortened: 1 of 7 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/create.go
```diff
@@ -92,7 +93,7 @@ func createNodeRequests(cOps commonOps, controlplaneRes, workerRes parsedNodeRes
 			Name:     machineName,
 			IPs:      ips,
 			Type:     machineType,
-			Memory:   controlplaneRes.memoryMb,
+			Memory:   int64(controlplaneRes.memory.Bytes()),
```
Orzelius, line 95: This was a bug, the memory value should have been passed in bytes

cmd/talosctl/cmd/mgmt/cluster/create/bytesize/bytesize.go
```diff
@@ -0,0 +1,92 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package bytesize
```
Orzelius: please let me know if the package should be located somewhere else

smira: we have `cmd/talosctl/pkg/`  which might be a better place

cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
```diff
@@ -131,19 +132,21 @@ func addTalosconfigDestinationFlag(flagset *pflag.FlagSet, bind *string, flagNam
...
-	flagset.IntVar(bind, flagName, 2048, "the limit on memory usage in MB (each control plane/VM)")
+func addControlPlaneMemoryFlag(flagset *pflag.FlagSet, bind *bytesize.ByteSize, flagName string) {
+	cli.Should(bind.Set("2.0gb")) // set default value
+	flagset.Var(bind, flagName, "the limit on memory usage for each control plane/VM")
 }
 
-func addWorkersMemoryFlag(flagset *pflag.FlagSet, bind *int, flagName string) {
-	flagset.IntVar(bind, flagName, 2048, "the limit on memory usage in MB (each worker/VM)")
+func addWorkersMemoryFlag(flagset *pflag.FlagSet, bind *bytesize.ByteSize, flagName string) {
+	cli.Should(bind.Set("2.0gb")) // set default value
```
smira: I think this should be 
```suggestion
	cli.Should(bind.Set("2.0GiB")) // set default value
```

cmd/talosctl/cmd/mgmt/cluster/create/cmd_qemu.go
```diff
@@ -58,7 +57,7 @@ func init() {
 	getQemuFlags := func() *pflag.FlagSet {
 		qemu := pflag.NewFlagSet("qemu", pflag.PanicOnError)
 
-		addDisksFlag(qemu, &ops.qemu.disks, []string{"virtio:" + strconv.Itoa(10*1024), "virtio:" + strconv.Itoa(6*1024)})
+		addDisksFlag(qemu, &ops.qemu.disks, []string{"virtio:10gb", "virtio:6gb"})
```
smira:
```suggestion
		addDisksFlag(qemu, &ops.qemu.disks, []string{"virtio:10GB", "virtio:6GB"})
```

cmd/talosctl/cmd/mgmt/cluster/create/bytesize/bytesize.go
```diff
@@ -0,0 +1,92 @@
...
+func (bs *ByteSize) Megabytes() uint64 {
+	return bs.valBytes / (1000 * 1000)
+}
+
+// Gigabytes returns the value in gigabytes.
+func (bs *ByteSize) Gigabytes() uint64 {
+	return bs.valBytes / (1000 * 1000 * 1000)
+}
+
+// IMegabytes returns the value in binary megabytes.
```
shanduur: Maybe worth naming it Mebibyte? Also same for other `I`-prefixed methods.

Orzelius: Ah didn't know they were called that. Thanks!

Issue #10899, opened 2025-05-05
ETCD encryption configuration


oliverbaehler:
## Feature Request

Hi, I was playing around with etcd encryption and Talos, or rather trying to embed a kmsv2 provider to encrypt all `secrets` and `configmaps` at REST. However i have noticed, that currently it's only possible to pass one key to the template here: https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/k8s/templates/kube-system-encryption-config-template.yaml

If I understood correctly, it's also not possible to overwrite that file. This causes some problems with ETCD encryptions:

- no other resources than `secrets` can be encrypted
- key rotation is not possible
- other kms than static keys on the controlplanes are not possible.

I was wondering if we could improve this by either:
- Being able to supply a custom-path to a user-defined encryption-configuration
- Having the Encryption-Config inline [as with the AuditPolicy](https://www.talos.dev/v1.3/introduction/what-is-new/#kube-apiserver-custom-audit-policy)

Wondering what your thoughts are about this. I can probably contribute the feature, if we agree on an implementation.

smira:
I think customizing encryption configuration would be a great additional feature!

frezbo:
ETCD specific machine config document would be great

oliverbaehler:
@frezbo Just for clarification, i would propose something like this:

```yaml
cluster:
    apiServer:
        encryptionProviderConfig:
            apiVersion: apiserver.config.k8s.io/v1
            kind: EncryptionConfiguration
            resources:
              - resources:
                  - secrets
                  - configmaps
                providers:
                  - aescbc:
                      keys:
                        - name: key1
                          secret: "${ETCD_ENCRYPTION_KEY_1}"
                        - name: key2
                          secret: "${ETCD_ENCRYPTION_KEY_2}"
```

I would place it under `cluster.apiServer.encryptionProviderConfig` because it should always be the same for all the cps. From the kubernetes-docs:

> For cluster configurations with two or more control plane nodes, the encryption configuration should be identical across each control plane node. If there is a difference in the encryption provider configuration between control plane nodes, this difference may mean that the kube-apiserver can't decrypt data.

If that's alright I will try to get back with a PR soon. 🍀

frezbo:
> [@frezbo](https://github.com/frezbo) Just for clarification, i would propose something like this:
> 
> cluster:
>     apiServer:
>         encryptionProviderConfig:
>             apiVersion: apiserver.config.k8s.io/v1
[...]

PR #11641, opened 2025-08-21
Feat/k8s config etcd encryption

clementnuss:
# Pull Request

## Note to the Contributor

I started implementing this feature after reading issue #10899 and more specifically @smira's [comment](https://github.com/siderolabs/talos/issues/10899#issuecomment-2851581462).
As suggested, I created a new `k8s` document config package, in which I added the new `EtcdEncryptionConfig` config.

I made a few assumptions that require validation:
- only 1 `EtcdEncryptionConfig` can be specified per machine config
- I do not validate the content of the custom `EtcdEncryptionConfig` (i.e. if the user uploads a broken etcd config, etcd will crash as a result). we could import k8s upstream types and try to marshal, let me know if it should be done.
- when an `EtcdEncryptionConfig` is specified, talos default etcd `encryptionconfig.yaml` is completely bypassed

something else that got me thinking: if there are multiple documents to be added in the future to permit customize all k8s moving parts, wouldn't be smart if instead of creating multiple talos configs to embed the corresponding k8s config we directly allowed specifying the k8s config? with an example, instead of:

```yaml
apiVersion: v1alpha1
kind: EtcdEncryptionConfig
config: |4
    apiVersion: apiserver.config.k8s.io/v1
    kind: EncryptionConfiguration
    ...
```
directly offer the possibility to use as a patch document

```yaml
apiVersion: apiserver.config.k8s.io/v1
kind: EncryptionConfiguration
...
```

## What? (description)

offers a config document based solution for the user to provide a custom
etcd encryption configuration. when specified, the custom encryption
config is applied as-is band it bypases the talos etcd encryption config

## Why? (reasoning)

a custom etcd configuration is needed as soon as you want to rotate, or if you want to have more specific parameters than talos defaults (e.g. other encryption scheme, key names, resources to be encrypted, etc)

## Acceptance

⚠️  still draft, I have some open questions before finalizing the PR and going through the checklist again.

- [x] you ran conformance (`make conformance`) (GPG key missing)


clementnuss:
I've implemented the another version where we directly use the upstream apiversion and kind as config document, making it transparent for users willing to configure that aspect of the apiserver.

I think this PR will be discarded in favor of #12785

smira:
Should be superseded now with #13253

PR #11647, opened 2025-08-22
chore: refactor mount package

### LINKED ISSUE #10951: refactor mount/v2 package to use new mount APIs
Use new Linux 5.15+ APIs to handle mounts better.

First stage - just supply new APIs.

Second stage - make mount remember the `opentree` handle and handle mounts on top of each other and mounts of top of each other correctly.

shanduur:
Fixes #10951 

This extensive refactoring introduces a new `mount/v3` package, migrating the mount logic from the previous `mount/v2` implementation to leverage the modern Linux Mount API. Direct use of the Linux Mount API offers access to advanced mount features.

**Key Changes and Rationale:**

*   **Introduction of `mount/v3`:** The core mount logic has been rewritten in the `internal/pkg/mount/v3` package. This new package directly utilizes the `fsopen(2)`, `fsmount(2)`, and `mount_setattr(2)` syscalls, providing a more granular and powerful control over mounting and unmounting operations.
*   **`Manager` and `Managers` Types:**
    *   A new `Manager` struct is introduced to encapsulate the configuration and state of a single mount point. It now uses an `xfs.FS` interface (backed by `fsopen.FS`) to represent the filesystem being mounted.
    *   A `Managers` type (a slice of `*Manager`) facilitates managing multiple mount points, ensuring proper mounting and unmounting order, and consolidating error handling.
*   **Enhanced Mount Options:** The `ManagerOption` type replaces `NewPointOption` and `OperationOption` from `v2`, offering a more comprehensive and type-safe way to configure mount behavior. New options include:
    *   `WithTarget()`: Specifies the mount target path.
    *   `WithKeepOpenAfterMount()`: Controls whether the underlying file descriptor for the mount should remain open after the mount operation.
    *   `WithSelinuxLabel()`: Sets the SELinux label for the mount.
    *   `WithShared()`: Enables shared propagation for the mount.
    *   `WithPrinter()`: Provides a flexible way to inject a logging function for mount operations.
    *   `WithSkipIfMounted()`: Prevents remounting if the target is already mounted.
    *   `WithFsopen()`: Configures the filesystem opener using the `fsopen` package, allowing for detailed filesystem-specific parameters.
    *   `WithMountAttributes()`: Directly sets mount attributes via `MOUNT_ATTR_*` flags.
    *   `WithReadOnly()`: Sets the mount as read-only.
    *   `WithExtraDirs()`: Specifies additional directories to be created before mounting.
    *   `WithExtraUnmountCallbacks()`: Allows for custom cleanup logic to be executed during unmount.
    *   `WithRepairOnUnmount()`: Triggers a filesystem repair during unmount if the filesystem type supports it.
*   **Updated Pseudo-Filesystem Mounting:** The functions `Pseudo()`, `PseudoLate()`, and `PseudoSub()` are updated to return `Managers` instances, aligning with the new mount architecture.
*   **Improved Error Handling and Retries:** The `retry` mechanism is integrated within the `Point` struct, providing more robust handling of transient mount/unmount errors, including automatic filesystem repair attempts for supported filesystems.
*   **`fsopen` Integration for Filesystem Parameters:** The `internal/pkg/xfs/fsopen` package is now used to build the filesystem configuration passed to the `fsopen(2)` syscall. This allows for a more structured and extensible way to specify filesystem-specific parameters.
    *   `WithStringParameter`, `WithBoolParameter`, and `WithBinaryParameters` are introduced to replace the generic `WithData` from `v2`.
*   **Simplified `Squashfs` Mounting:** The `Squashfs` helper now returns a `Manager`, encapsulating the loop device attachment and squashfs-specific mount options.
*   **Consolidated `Cgroup2` Mounting:** The `CGroupMountPoints()` function is replaced by `NewCgroup2()`, providing a dedicated manager for cgroup2 mounts.
*   **Removal of `mount/v2`:** The entire `internal/pkg/mount/v2` package has been deleted, ensuring a clean transition to the new `v3` implementation.


internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -548,20 +564,20 @@ func (ctrl *MountController) handleOverlayMountOperation(
 		return fmt.Errorf("overlay mount is not supported for %q", volumeStatus.TypedSpec().ParentID)
 	}
 
-	mountpoint := mount.NewVarOverlay(
+	manager := mount.NewVarOverlay(
 		[]string{mountTarget},
 		mountTarget,
-		mount.WithFlags(unix.MS_I_VERSION),
```
shanduur, line 554: `MS_I_VERSION` is no longer a valid flag. It was replaced by `iversion` fsconfig parameter, which is accepted by configs of:
- `ext4` (default enabled)
- `xfs` (always enabled)
- `btrfs`
- `f2fs`

internal/app/machined/pkg/runtime/v1alpha1/bootloader/mount/mount.go
```diff
@@ -56,13 +58,25 @@ func PartitionOp(
...
-					mount.NewPoint(
-						partitioning.DevName(disk, partition.PartitionIndex),
-						spec.MountTarget,
-						spec.FilesystemType,
-						newPointOptions...,
-					),
+				manager := mount.NewManager(slices.Concat(
+					[]mount.ManagerOption{
+						mount.WithTarget(spec.MountTarget),
+						// mount.WithRepairOnUnmount(),
```
frezbo: is this commented code meant to be for something in the future?

internal/pkg/mount/v3/all.go
```diff
@@ -112,7 +112,7 @@ func readMountInfo() ([]mountInfo, error) {
 			mntInfo.MountPoint = pre[4]
 		}
 
-		if len(post) >= 1 {
+		if len(post) >= 2 {
```
frezbo, line 115: since we changed this, maybe have a test for this?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/mount/mount.go
shanduur: No, I missed that during cleanup. That was part of the debugging of the imager issue, both function and call will be dropped.

Thanks!

internal/pkg/mount/v3/helpers.go
```diff
@@ -0,0 +1,313 @@
...
+				fsopen.WithPrinter(printer),
+				fsopen.WithStringParameter("mode", "0755"),
+			),
+		),
+		newManager(
+			always,
+			WithTarget("/tmp"),
+			WithMountAttributes(unix.MOUNT_ATTR_NOSUID|unix.MOUNT_ATTR_NOEXEC|unix.MOUNT_ATTR_NODEV),
+			WithFsopen(
+				"tmpfs",
```
frezbo, line 235: https://www.kernel.org/doc/html/latest/filesystems/tmpfs.html I wonder if we should also do tmpfs quotas, something for later not this PR, FYI @smira

internal/pkg/mount/v3/manager.go
```diff
@@ -0,0 +1,236 @@
...
+func (m *Manager) Mount() (*Point, error) {
+	printer := discard
+	if m.printer != nil {
+		printer = m.printer
+	}
+
+	for _, dir := range m.extraDirs {
+		printer("creating directory tree %q", dir)
+
+		if err := os.MkdirAll(dir, 0o755); err != nil {
```
frezbo, line 61: should we make sure the caller calls this code from `/`? otherwise the code could create a directory under `/foo` for example

internal/pkg/mount/v3/all.go
shanduur: Previously this might have caused out-of-bounds error. If we never hit that, I'm not sure we need to worry about it here?

internal/pkg/mount/v3/manager.go
shanduur: Do you mean by validating that the `dir` is absolute path?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/mount/mount.go
shanduur: Removed.

internal/pkg/mount/v3/manager.go
frezbo: I mean if there is some code that does `os.Chddir` and calls this api doesn't `os.Mkdirall` creates under that folder 🤔

shanduur: I created check if the path is absolute, otherwise it will return error.

internal/app/init/main.go
```diff
@@ -124,20 +125,22 @@ func mountRootFS() error {
...
 		}
 	}
 
 	// if no extensions found use plain squashfs mount
 	if len(extensionsConfig.Layers) == 0 {
-		squashfs, err := mount.Squashfs(constants.NewRoot, "/"+constants.RootfsAsset)
+		squashfs, err := mount.Squashfs(constants.NewRoot, "/"+constants.RootfsAsset, log.Printf)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to initialize sqashfs: %w", err)
```
smira:
```suggestion
			return fmt.Errorf("failed to initialize squashfs: %w", err)
```

internal/app/init/main.go
```diff
@@ -124,20 +125,22 @@ func mountRootFS() error {
...
 		}
 
-		_, err = squashfs.Mount()
+		if _, err := squashfs.Mount(); err != nil {
+			return fmt.Errorf("failed to mount sqashfs: %w", err)
```
smira:
```suggestion
			return fmt.Errorf("failed to mount squashfs: %w", err)
```

internal/pkg/mount/v3/all.go
smira: yep, it was a bug, thanks for catching that!

internal/pkg/mount/v3/helpers.go
```diff
@@ -0,0 +1,313 @@
...
+
+func hasEFIVars() bool {
+	_, err := os.Stat(constants.EFIVarsMountPoint)
+
+	if err == nil || errors.Is(err, os.ErrNotExist) {
+		return err == nil
+	}
+
+	// this means something else is wrong, let's panic
+	// as this shoul never happen
```
smira:
```suggestion
	// as this should never happen
```

internal/pkg/mount/v3/helpers.go
smira: yeah, quotas might be cool, not sure if we have any way to enforce them? (most should come from `root` anyways?)

internal/pkg/mount/v3/manager.go
```diff
@@ -0,0 +1,227 @@
...
+	opts := Options{
+		Printer:         printer,
+		Shared:          m.shared,
+		SkipIfMounted:   m.skipIfMounted,
+		MountAttributes: m.mountattr,
+	}
+
+	printer("mkdirAll %q", m.target)
+
+	if err := os.MkdirAll(m.target, 0o755); err != nil {
```
smira, line 88: not for this PR, but I wish we dropped this and did more explicit mkdir based on specific settings

smira (approved):
🆒  🎉

internal/pkg/mount/v3/helpers.go
shanduur: Fixed

PR #11656, opened 2025-08-25
feat: support configuring PCR states to bind disk encryption

frezbo:
feat: support configuring PCR states to bind disk encryption

See https://github.com/systemd/systemd/commit/4b840414be3b2d6520599d86d2b718a37574aabf for more information.

Talos versions prior to 1.12 locked to PCR 7 state and PCR 11 for signed policies.

In-order for backwards compatibility newer installs will still default to PCR 7 state. Locking to PCR 7 can be disabled by passing an empty list.

Fixes: #10677


internal/pkg/secureboot/tpm2/pcr.go
```diff
@@ -151,15 +151,6 @@ func validatePCRBanks(t transport.TPM) error {
 		return err
 	}
 
-	pcrValue, err = ReadPCR(t, SecureBootStatePCR)
```
frezbo, line 154: i think we can still keep this check, doesn't hurt, thoughts?

shanduur (approved):

internal/pkg/secureboot/tpm2/pcr.go
smira: yeah, why do we drop the check?

frezbo: since new version won't rely on PCR 7, I'll add back and add some release notes

smira:
Should we keep old behavior available for backward compatibility? (or people who want to explicitly lock to PCR 7?)

internal/pkg/secureboot/tpm2/unseal.go
```diff
@@ -224,25 +224,27 @@ func Unseal(sealed SealedResponse) ([]byte, error) {
 		return nil, fmt.Errorf("failed to execute policy authorize: %w", err)
 	}
 
-	secureBootStatePCRSelector, err := CreateSelector([]int{SecureBootStatePCR})
```
smira, line 227: probably keeping the constant here will make it more readable?

frezbo: yup, reverted to using the constant

frezbo:
> Should we keep old behavior available for backward compatibility? (or people who want to explicitly lock to PCR 7?)

I think this is way better, old ones would fail to unlock even for a dbx update or some small change in certs enrolled, existing installs would unlock with old schema, and this provides a way to rotate to just using PCR 11

smira:
> > Should we keep old behavior available for backward compatibility? (or people who want to explicitly lock to PCR 7?)
> 
> I think this is way better, old ones would fail to unlock even for a dbx update or some small change in certs enrolled, existing installs would unlock with old schema, and this provides a way to rotate to just using PCR 11

I wonder if we should optionally support locking to PCR7, even if it is disabled by default.

frezbo:
> > > Should we keep old behavior available for backward compatibility? (or people who want to explicitly lock to PCR 7?)
> > 
> > 
> > I think this is way better, old ones would fail to unlock even for a dbx update or some small change in certs enrolled, existing installs would unlock with old schema, and this provides a way to rotate to just using PCR 11
> 
> I wonder if we should optionally support locking to PCR7, even if it is disabled by default.

I can add that for the new encryption config

website/content/v1.12/schemas/config.schema.json
```diff
@@ -104,9 +104,16 @@
...
-          "x-intellij-html-description": "\u003cp\u003eCheck that Secureboot is enabled in the EFI firmware.\nIf Secureboot is not enabled, the enrollment of the key will fail. As the TPM key is anyways bound to the value of PCR 7, changing Secureboot status or configuration after the initial enrollment will make the key unusable.\u003c/p\u003e\n"
+          "description": "Check that Secureboot is enabled in the EFI firmware.\nIf Secureboot is not enabled, the enrollment of the key will fail.\n",
+          "markdownDescription": "Check that Secureboot is enabled in the EFI firmware.\nIf Secureboot is not enabled, the enrollment of the key will fail.",
+          "x-intellij-html-description": "\u003cp\u003eCheck that Secureboot is enabled in the EFI firmware.\nIf Secureboot is not enabled, the enrollment of the key will fail.\u003c/p\u003e\n"
+        },
+        "lockToSecureBootState": {
+          "type": "boolean",
+          "title": "lockToSecureBootState",
+          "description": "Lock the TPM key to the current Secureboot state. changing Secureboot status or configuration after the initial enrollment will make the key unusable.\n",
```
dsseng: Maybe keep notes on which PCRs we bind in each case?

dsseng (approved):
is there some compatibility during an upgrade process so that old partitions either continue to unlock or their headers are upgraded with PCR11-only unlock?

frezbo:
> is there some compatibility during an upgrade process so that old partitions either continue to unlock or their headers are upgraded with PCR11-only unlock?

yeh, old luks header won't have version field, so will fall to using pcr 7 also to unlock, newer versions have the field or if explicitly requested to lock to PCR 7 will have both 7 and 11 in the pcr's list

website/content/v1.12/schemas/config.schema.json
frezbo: yeh, not sure if most people care 😅

dsseng: if they do it should be findable in source for them ;)

smira (commented):
as a result of discussion:

* we should keep backwards compatible config, i.e. if the default previously was to lock to PCR 7, new machine config should lock to PCR 7 by default as well
* (?) should we allow to configure a set of PCRs to lock to (with the default being `[]7` and an option to lock to none of them, or have a simple option: "lockToSecureBootState"

internal/pkg/secureboot/tpm2/unseal.go
```diff
@@ -224,25 +225,23 @@ func Unseal(sealed SealedResponse) ([]byte, error) {
...
-		PCRSelections: []tpm2.TPMSPCRSelection{
-			{
-				Hash:      tpm2.TPMAlgSHA256,
-				PCRSelect: secureBootStatePCRSelector,
-			},
-		},
-	})
-	if err != nil {
-		return nil, fmt.Errorf("failed to calculate policy PCR digest: %w", err)
+	// this handles the case when talos is upgraded from pre Talos 1.12 and we had PCRs field
```
frezbo, line 228: i wonder if this can be simplified a bit

pkg/provision/providers/qemu/launch.go
```diff
@@ -243,7 +243,6 @@ func launchVM(config *LaunchConfig) error {
 			fmt.Sprintf("dir=%s,mode=0644", config.TPMConfig.StateDir),
 			"--ctrl",
 			fmt.Sprintf("type=unixio,path=%s", tpm2SocketPath),
-			// "--tpm2",
```
utkuozdemir, line 246: unintentional?

frezbo: no, removed the  commented code

cmd/talosctl/cmd/mgmt/cluster/create/create_dev.go
```diff
@@ -815,7 +815,11 @@ func convertEncryptionKeys(keys []*v1alpha1.EncryptionKey) []block.EncryptionKey
 		}
 
 		if k.KeyTPM != nil {
-			r.KeyTPM = pointer.To(block.EncryptionKeyTPM(*k.KeyTPM))
+			encrypttionKeyTPM := block.EncryptionKeyTPM{
```
smira:
```suggestion
			encrypttonKeyTPM := block.EncryptionKeyTPM{
```

dsseng:
```suggestion
			encryptionKeyTPM := block.EncryptionKeyTPM{
```

internal/app/machined/pkg/controllers/block/volume_config.go
```diff
@@ -124,6 +124,8 @@ func convertEncryptionConfiguration(in cfg.EncryptionConfig, out *block.VolumeCo
 		case key.TPM() != nil:
 			out.Encryption.Keys[i].Type = block.EncryptionKeyTPM
 			out.Encryption.Keys[i].TPMCheckSecurebootStatusOnEnroll = key.TPM().CheckSecurebootOnEnroll()
+			out.Encryption.Keys[i].TPMPCRs = key.TPM().PCRs()
+			out.Encryption.Keys[i].PubKeyPCRs = []int{constants.UKIPCR}
```
smira: it seems that we don't consume this value anywhere?

smira (approved):
let's add an upgrade test with encryption.. probably with next alpha of Talos (?)

internal/app/machined/pkg/controllers/block/volume_config.go
frezbo: correct, but I would like this to be reported

smira: I mean in the tpm2 code, we don't use this value, but rather use the constant once again - my point is - should we actually use value from the resource? (I'm all for exposing this in a resource)

frezbo: I'll fix it to use this from a named method on tpm, and have the default there, also saw that I need to propagate these values to the volumestatus resource too, otherwise the test is not verifying the disk was unlocked with the right set of PCR combinations

.kres.yaml
```diff
@@ -744,11 +744,11 @@ spec:
             PLATFORM: linux/amd64,linux/arm64
             IMAGER_ARGS: "--extra-kernel-arg=console=ttyS0 --extra-kernel-arg=enforcing=1"
             IMAGE_REGISTRY: registry.dev.siderolabs.io
-        - name: integration-trusted-boot
+        - name: integration-trusted-boot-enforcing
```
frezbo, line 747: minor fix

Dockerfile
```diff
@@ -351,7 +351,7 @@ RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool golang.org/x/tools
 RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool mvdan.cc/gofumpt -w ./pkg/
 WORKDIR /src/pkg/machinery
 RUN --mount=type=cache,target=/.cache,id=talos/.cache go generate ./...
-RUN go tool github.com/siderolabs/talos/tools/gotagsrewrite .
+RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool github.com/siderolabs/talos/tools/gotagsrewrite .
```
frezbo, line 354: this was missing and was downloading for every run

Dockerfile
```diff
@@ -1185,6 +1185,17 @@ RUN --mount=type=cache,target=/.cache,id=talos/.cache GOOS=linux GOARCH=arm64 go
 FROM scratch AS integration-test-linux-arm64
 COPY --from=integration-test-linux-arm64-build /src/integration.test /integration-test-linux-arm64
 
+FROM base AS integration-test-darwin-arm64-build
```
frezbo, line 1188: makes it easier to run integration tests on M-series Mac

internal/app/machined/pkg/controllers/block/volume_config.go
frezbo: fixed

PR #11692, opened 2025-08-28
test: ignore one leaking goroutine

smira:
I believe it's a bug in the `suture` library itself, relevant code is:

https://github.com/thejerf/suture/blob/8a2561c661dce2a30c484b08e865676059c1e63e/supervisor.go#L779-L816

If the global context is canceled, nothing consumes from the notification channel, so the goroutine trying to send there will never have a chance to complete.

Also avoid parallel issues by moving goleak check into `TestMain`.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/dns_resolve_cache_test.go
+++ b/internal/app/machined/pkg/controllers/network/dns_resolve_cache_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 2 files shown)


frezbo:
should we create an upstream issue?

smira:
> should we create an upstream issue?

I would rather rewrite the whole piece of code and get rid of this code.

frezbo (approved):

frezbo:
> > should we create an upstream issue?
> 
> I would rather rewrite the whole piece of code and get rid of this code.



> > should we create an upstream issue?
> 
> I would rather rewrite the whole piece of code and get rid of this code.

+1 for that

PR #11696, opened 2025-08-28
feat: use the id of the `VolumeConfig` in the mapped name when decrypting

### LINKED ISSUE #11662: Encrypted `RawVolumeConfigs` should have a mapped path based on the resource name
### Description

As far as I can tell, though the raw volume gets a known label, the virtual block device is only accessible at `/dev/mapper/<partition-name>-encrypted` and `/dev/dm-#`, where `<partition-name>` is something like `nvme0n1p5` and `/dev/dm-#` is arbitrarily numbered.

This doesn't feel intentional, am I missing something? It makes it quite difficult to refer to the encrypted volume by the resource name.

Should the mapped partition have a path like `/dev/mapper/<name>-decrypted`?

### Environment

- Talos version: v1.11.0-rc.0
- Kubernetes version: v1.34-rc.1
- Platform: metal

michaelbeaumont:
# Pull Request

## What? (description)

Fixes #11662 so that not only the raw partition has a name based on the encrypted RawVolumeConfig resource's name but also the mapped partition that ends up being used by the user. The new mapped path has the format `/dev/mapper/r-<name>-decrypted`.

I suspect this changes the mapped name for both `VolumeConfigs` and `UserVolumeConfigs`, is that a problem? Maybe we can filter for RawVolumeConfigs based on labels?

## Why? (reasoning)

The way to refer to a non-encrypted RawVolumeConfig is by the partition label `r-<name>`. But the point of an encrypted RawVolumeConfig is to provide the user with a block device _on top_ of the partition created by Talos. The underlying device is uninteresting. At the moment the only options to refer to this virtual device are the arbitrarily numbered name `/dev/dm-##` and the `/dev/mapper/<partition>-encrypted` where `partition` is the path-based name `/dev/mapper/nvme0n1p5-encrypted`. This PR changes the latter mapped name to be based on the name of the resource rather than the underlying partition path. This PR also changes the suffix `-encrypted` to `-decrypted` because this virtual device is the _decrypted_ block device, i.e. it contains the decrypted data.

## Acceptance

  - no new tests
  - but I get failing tests on `main` so not sure

### DIFF
--- a/internal/app/machined/pkg/controllers/block/internal/volumes/close.go
+++ b/internal/app/machined/pkg/controllers/block/internal/volumes/close.go
@@ -7,7 +7,6 @@ package volumes
 import (
 	"context"
 	"fmt"
-	"path/filepath"
 
 	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
@@ -52,15 +51,15 @@ func CloseWithHandler(ctx context.Context, logger *zap.Logger, volumeContext Man
 	ctx, cancel := context.WithTimeout(ctx, encryptionTimeout)
 	defer cancel()
 
-	encryptedName := filepath.Base(volumeContext.Status.Location) + "-encrypted"
+	mappedName := handler.Name() + "-" + volumeContext.Cfg.Metadata().ID()
 
-	if err := handler.Close(ctx, encryptedName); err != nil {
-		return xerrors.NewTaggedf[Retryable]("error closing encrypted volume %q: %w", encryptedName, err)
+	if err := handler.Close(ctx, mappedName); err != nil {
+		return xerrors.NewTaggedf[Retryable]("error closing encrypted volume mapped to %q: %w", mappedName, err)
 	}
 
 	volumeContext.Status.Phase = block.VolumePhaseClosed
 
-	logger.Info("encrypted volume closed", zap.String("name", encryptedName))
+	logger.Info("encrypted volume closed", zap.String("name", mappedName))
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/block/internal/volumes/close.go
```diff
@@ -52,15 +52,15 @@ func CloseWithHandler(ctx context.Context, logger *zap.Logger, volumeContext Man
 	ctx, cancel := context.WithTimeout(ctx, encryptionTimeout)
 	defer cancel()
 
-	encryptedName := filepath.Base(volumeContext.Status.Location) + "-encrypted"
+	mappedName := filepath.Base(volumeContext.Cfg.TypedSpec().Provisioning.PartitionSpec.Label) + "-decrypted"
```
smira: this certainly doesn't make any sense, as partition label is not required at all.

if we were to change anything here, the only way we can go with is to use volume's ID.

michaelbeaumont: It doesn't make _any_ sense _certainly_? It seems to be set by the control plane for all the configs I've seen. 

The problem with the ID, if you're referring to the resource ID, afaict is that it's not necessarily unique across resources. This could be either a UserVolumeConfig or a RawVolumeConfig.

We could use the labels to get the resource type perhaps, i.e. recreate the partition label

michaelbeaumont: OK I see now we have access via `Cfg.Metadata()` to the id of the common VolumeConfig, I think, changed that

frezbo (approved):

michaelbeaumont:
Thanks!! :pray:

shanduur (approved):

Issue #11404, opened 2025-07-21
User facing `talosctl cluster create commands`


Orzelius:
Currently the talosctl cluster create command is mainly aimed for internal usage and the development of Talos. As we already have advanced automation for creation of production-like local cluster the functionality could also be beneficial to general users of Talos Linux.

PR #11706, opened 2025-08-29
refactor create cluster commands internal logic

Orzelius:
Split common code between `create docker` and `create qemu` commands into a helper. The dev command stays as is due to big differences between the user oriented commands.

part of #11404

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
@@ -84,15 +84,31 @@ type commonOps struct {
 }
 
 func getDefaultCommonOptions() commonOps {
+	memory2GB := bytesize.WithDefaultUnit("MiB")
+	cli.Should(memory2GB.Set("2.0GiB"))
+	defaultResources := nodeResources{
+		cpu:    "2.0",
+		memory: *memory2GB,
+	}
+
 	return commonOps{
-		controlplanes:      1,
-		networkMTU:         1500,
-		clusterWaitTimeout: 15 * time.Minute,
-		clusterWait:        true,
-		dnsDomain:          "cluster.local",
-		controlPlanePort:   constants.DefaultControlPlanePort,
-		rootOps:            &clustercmd.PersistentFlags,
-		networkIPv4:        true,
+		controlplanes:         1,
+		controlplaneResources: defaultResources,
+		workers:               1,
+		workerResources:       defaultResources,
+
+		networkCIDR:            "10.5.0.0/24",
+		kubernetesVersion:      constants.DefaultKubernetesVersion,
+		networkMTU:             1500,
+		clusterWaitTimeout:     20 * time.Minute,
+		clusterWait:            true,
+		dnsDomain:              "cluster.local",
+		controlPlanePort:       constants.DefaultControlPlanePort,
+		rootOps:                &clustercmd.PersistentFlags,
+		networkIPv4:            true,
+		kubePrismPort:          constants.DefaultKubePrismPort,
+		enableClusterDiscovery: true,
+		talosVersion:           helpers.GetTag(),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/create_qemu.go
```diff
@@ -0,0 +1,119 @@
...
+			if err != nil {
+				return nil, nil, err
+			}
+
+			for i := range cp {
+				cp[i].Disks = primaryDisks
+			}
+
+			for i := range w {
+				w[i].Disks = append(primaryDisks, workerDisks...) //nolint:gocritic
```
smira:
```suggestion
				w[i].Disks = slices.Concat(primaryDisks, workerDisks)
```

cmd/talosctl/cmd/mgmt/cluster/create/create_qemu.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return createClusterRequest(createClusterRequestOps{
+		commonOps:   cOps,
+		provisioner: provisioner,
+		withExtraGenOpts: func(cr provision.ClusterRequest) []generate.Option {
+			genOptions := []generate.Option{
+				generate.WithInstallImage(qOps.nodeInstallImage),
+			}
+
+			endpointList := make([]string, cOps.controlplanes)
```
smira:
```suggestion
			endpointList := xslices.Map(cOps.controlplanes, func(n NodeRequest) string { return node.IPs[0].String })
```

cmd/talosctl/cmd/mgmt/cluster/create/create_qemu.go
```diff
@@ -0,0 +1,119 @@
...
+
+			for i := range cOps.controlplanes {
+				endpointList[i] = cr.Nodes[i].IPs[0].String()
+			}
+
+			genOptions = append(genOptions, getWithAdditionalSubjectAltNamesGenOps(endpointList)...)
```
smira: I think we don't need, it's legacy

Orzelius:
After some discussion with @smira we've decided that it's probably smarter to move the logic directly to the ops types as methods and have it in a separate package. This will create a nice public api which and will aid with testability. That being said this pr is still a move towards a cleaner state so it still makes sense to merge it and move on from there.

cmd/talosctl/cmd/mgmt/cluster/create/create_qemu.go
Orzelius: Removed for qemu but kept for docker as it stopped working without it.

cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
```diff
@@ -133,22 +152,18 @@ func addTalosconfigDestinationFlag(flagset *pflag.FlagSet, bind *string, flagNam
 }
 
 func addControlplaneCpusFlag(flagset *pflag.FlagSet, bind *string, flagName string) {
-	flagset.StringVar(bind, flagName, "2.0", "the share of CPUs as fraction for each control plane/VM")
+	flagset.StringVar(bind, flagName, _dco.controlplaneResources.cpu, "the share of CPUs as fraction for each control plane/VM")
```
smira: I'm confused with `default` value and `bind` being different here? 

Should the default be `*bind` ?

cmd/talosctl/cmd/mgmt/cluster/create/cmd_dev.go
```diff
@@ -193,15 +191,16 @@ func getCreateCmd() *cobra.Command {
...
-		qemu:   qemuOps{},
+		common: getDefaultCommonOptions(),
+		qemu:   getDefaultQemuOptions(),
 	}
 	legacyOps := legacyOps{}
 
 	getCommonFlags := func() *pflag.FlagSet {
 		common := pflag.NewFlagSet("common", pflag.PanicOnError)
 
+		dco := getDefaultCommonOptions()
```
smira: why another variable here, why can't we use `opts.common.xxx`?

cmd/talosctl/cmd/mgmt/cluster/create/cmd_dev.go
```diff
@@ -218,91 +217,93 @@ func getCreateCmd() *cobra.Command {
...
-		common.BoolVar(&ops.common.skipInjectingConfig, skipInjectingConfigFlag, false, "skip injecting config from embedded metadata server, write config files to current directory")
-		common.BoolVar(&ops.common.enableClusterDiscovery, withClusterDiscoveryFlag, true, "enable cluster discovery")
-		common.BoolVar(&ops.common.enableKubeSpan, enableKubeSpanFlag, false, "enable KubeSpan system")
-		common.IntVar(&ops.common.kubePrismPort, kubePrismFlag, constants.DefaultKubePrismPort, "KubePrism port (set to 0 to disable)")
-		common.BoolVar(&ops.common.skipK8sNodeReadinessCheck, skipK8sNodeReadinessCheckFlag, false, "skip k8s node readiness checks")
-		common.BoolVar(&ops.common.withJSONLogs, withJSONLogsFlag, false, "enable JSON logs receiver and configure Talos to send logs there")
-		common.BoolVar(&ops.common.withUUIDHostnames, withUUIDHostnamesFlag, false, "use machine UUIDs as default hostnames")
+		common.StringVar(&ops.common.networkCIDR, networkCIDRFlagName, dco.networkCIDR, "CIDR of the cluster network (IPv4, ULA network for IPv6 is derived in automated way)")
```
smira:
```suggestion
		common.StringVar(&ops.common.networkCIDR, networkCIDRFlagName, ops.common.networkCIDR, "CIDR of the cluster network (IPv4, ULA network for IPv6 is derived in automated way)")
```

?

cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
Orzelius: My thinking was that this is would result in a shorter line, but on a second thought it's unnecessary. I've changed the code such that the default values come from the existing option variables and binds.

PR #11747, opened 2025-09-03
feat: add a userspace OOM controller

### LINKED ISSUE #11588: experimental OOM handler
## Trigger

* Monitor memory PSI information via cgroups
* We can tune which PSI is more important:
   * global cgroup
   * or high-priority cgroups: `init`/`podruntime`/`runtime` (?)
 * Threshold: which value of PSI triggers OOM

## Pick target

* First, sort by high-level group:
   * `kubepods` (workloads)
   * `podruntime` (CRI, kubelet, etcd)
   * `runtime` (core containerd, system services)
   * `init`
 * Second, inside `kubepods` we have [QoS groups](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/):
   * first priority: BestEffort
   * second: Burstable
   * last: Guaranteed
  * Third, look into other attributes, e.g. OOM score.
  * Fourth, look into memory max - memory current (if memory max is set).

Pick a target(s) to kill

## Kill

Kill the whole cgroup.

## Cooldown

Wait some time before, next kill, go to the beginning monitoring PSI.

dsseng:
OOM controller helps mitigate out-of-memory by monitoring PSI, selecting
and killing cgroups to free up memory when RAM is near full.

Fixes #11588


internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+)
+
+const (
+	sampleInterval  = 5000 * time.Millisecond
+	mempressureProp = "memory.pressure"
+	pressureType    = "full"
+	pressureSpan    = "avg10"
+	psiThresh       = 12
+	cooldownTimeout = time.Second * 20
```
frezbo:
```suggestion
	cooldownTimeout = 20 * time.Second
```

to be consistent

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+
+type oomRankedCgroup struct {
+	Class         int
+	Path          string
+	MemoryCurrent int64
+	MemoryPeak    int64
+	MemoryMax     int64
+}
+
+// OOMController is a controller that publishes Talos SBOMs as resources.
```
frezbo: not sbom 😅

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// Outputs implements controller.Controller interface.
+func (ctrl *OOMController) Outputs() []controller.Output {
+	return nil
+}
+
+// Run implements controller.Controller interface.
+func (ctrl *OOMController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	ticker := time.NewTicker(sampleInterval)
```
frezbo, line 98: sampling needs to be this fast?

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		case <-tickerC:
+		}
+
+		node, err := cgroups.GetCgroupProperty(constants.CgroupMountPath, mempressureProp)
+		if err != nil {
+			fmt.Println("cannot read memory pressure", err)
```
frezbo: we should be using the `logger`

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: yeah, prototype quality, gotta refactor it first

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: once in 5 seconds seems to be okay, but perhaps that should be configurable

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+			continue
+		}
+
+		fmt.Println(node.MemoryPressure)
```
smira: we have `logger` - please use it (even for debug output, but certainly not one per 5seconds in the final version)

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+		if !ok {
+			fmt.Println("cannot find memory pressure span:", pressureSpan)
+			continue
+		}
+
+		if !value.IsSet || value.IsMax {
+			continue
+		}
+
+		val, err := strconv.ParseFloat(value.String(), 32)
```
smira: nit: let's implement (and add a test for): `value.Float64()`

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: Yes, most of these logs should and will be removed, errors directed to proper logger

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+
+			for _, prop := range []string{"memory.current", "memory.peak", "memory.max"} {
+				err := cgroups.ReadCgroupfsProperty(&node, leafDir, prop)
+				if err != nil {
+					fmt.Println("cannot read property for cgroup", leafDir, prop, err)
+					continue
+				}
+			}
+
+			ranking = append(ranking, oomRankedCgroup{
```
smira: from testability perspective, I think we should extract this type to some `internal/oom` subpackage (as a public type there), implement a `Compare` function on it (whether we use CEL or do some builtin weighting), and add unit-tests for it.

It's harder (but still possible) to test the code which gathers this info from cgroupfs, but the code which makes a decision is absolutely possible to test.

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: Yes, certainly would be a reusable method, so let's move this part to cgroups

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,208 @@
...
+			return a.MemoryCurrent > b.MemoryCurrent
+		}
+
+		return a.Class < b.Class
+	})
+
+	fmt.Println(ranking)
+	fmt.Println("SENDING SIGKILL TO CGROUP", ranking[0])
+
+	os.WriteFile(filepath.Join(ranking[0].Path, "cgroup.kill"), []byte{'1'}, 0o644)
```
smira: I wonder if we should also create a resource(s) to store something like ten last killed cgroup names as the output of the controller.

We can also store there PSI at the moment we killed, the CGroup properties - why it was picked, etc.

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: Gathering info should also be possible with a mock fs or storing mock data in a temporary dir

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: We could keep such a log I think. Mostly that would be pods, not sure if it's of much use for the users, but should help in development I think

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,324 @@
...
+			logger.Error("cannot parse memory pressure:", zap.String("pressureSpan", pressureSpan), zap.Error(err))
+			continue
+		}
+
+		evalContext := map[string]any{
+			"memory_full_avg10":  float64(val),
+			"time_since_trigger": time.Since(ctrl.ActionTriggered),
+		}
+
+		// FIXME: remove hot-path logging
```
dsseng: .

shanduur (approved):

internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go
```diff
@@ -0,0 +1,197 @@
...
+// RankedCgroup contains information about a cgroup used for OOM handling.
+type RankedCgroup struct {
+	Class         runtime.QoSCgroupClass
+	Path          string
+	MemoryCurrent cgroups.Value
+	MemoryPeak    cgroups.Value
+	MemoryMax     cgroups.Value
+}
+
+func cgroupValueToFloat64(v cgroups.Value, evalContext map[string]any, key string) {
```
smira: I'm confused either about the name, or about that this function does, as it seems to convert to `uint`, not `float64`?

internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go
```diff
@@ -0,0 +1,197 @@
...
+		"class": int(cgroup.Class),
+		"path":  cgroup.Path,
+	}
+
+	cgroupValueToFloat64(cgroup.MemoryCurrent, evalContext, "memory_current")
+	cgroupValueToFloat64(cgroup.MemoryPeak, evalContext, "memory_peak")
+	cgroupValueToFloat64(cgroup.MemoryMax, evalContext, "memory_max")
+
+	result, err = expr.EvalDouble(celenv.OOMCgroupScoring(), evalContext)
+	log.Printf("Evaluating CalculateScore: evalContext = %v, result = %v, err = %v", evalContext, result, err)
```
smira: drop this or move to debug, but I think even for debug logging this might be too much

internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go
```diff
@@ -0,0 +1,197 @@
...
+		evalContext[key] = types.OptionalNone
+	} else {
+		evalContext[key] = types.OptionalOf(types.Uint(v.Val))
+	}
+}
+
+// CalculateScore calculates the score of the cgroup for OOM handling.
+//
+// Higher score means the cgroup is more likely to be killed.
+func (cgroup *RankedCgroup) CalculateScore(expr *cel.Expression) (result float64, err error) {
```
smira: let's avoid using named params, we can simple return `expr.EvalDouble(...)` here

```suggestion
func (cgroup *RankedCgroup) CalculateScore(expr *cel.Expression) (float64, err) {
```

internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return
+}
+
+// EvaluateTrigger is a method obtaining data and evaluating the trigger expression.
+// When the result is true, designated OOM action is to be executed.
+func EvaluateTrigger(triggerExpr cel.Expression, evalContext map[string]any, cgroup string) (bool, error) {
+	err := PopulatePsiToCtx(cgroup, evalContext)
+	if err != nil {
+		return false, fmt.Errorf("cannot populate context: %w", err)
```
smira: nit to make error a bit more descrivpitve: 

```suggestion
		return false, fmt.Errorf("cannot populate PSI context: %w", err)
```

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,245 @@
...
+
+	processes := oom.ListCgroupProcs(cgroupPath)
+	logger.Info("victim processes:", zap.Any("processes", processes))
+
+	// Open pidfd's of all the processes in cgroup to accelerate kernel
+	// garbage-collecting those processes via mrelease.
+	pidfds := []int{}
+
+	for _, pid := range processes {
+		pidfd, err := unix.PidfdOpen(pid, 0)
```
smira, line 220: I would add a comment here that it looks like `close-on-exec` is set here by default (as otherwise it might leak if concurrent goroutine does fork/exec).

internal/app/machined/pkg/controllers/runtime/internal/oom/oom.go
```diff
@@ -0,0 +1,197 @@
...
+func ListCgroupProcs(cgroupPath string) []int {
+	processes := []int{}
+
+	// Ignore errors, find as many processes as possible
+	//nolint:errcheck
+	filepath.WalkDir(cgroupPath, func(path string, d fs.DirEntry, walkErr error) error {
+		if walkErr != nil {
+			return walkErr
+		}
+
```
smira, line 176: I think we should skip non-directories here? as we are only interested in paths which are directories

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,245 @@
...
+
+	ticker := time.NewTicker(sampleInterval)
+	tickerC := ticker.C
+
+	if ctrl.CgroupRoot == "" {
+		ctrl.CgroupRoot = constants.CgroupMountPath
+	}
+
+	for {
+		// the controller runs a single time
```
smira: I guess this comment is not accurate? it runs in a loop

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	err := reapCg(logger, cgroupToKill.Path)
+	if err != nil {
+		logger.Error("cannot reap cgroup", zap.String("cgroup", cgroupToKill.Path), zap.Error(err))
+	}
+}
+
+func reapCg(logger *zap.Logger, cgroupPath string) error {
+	logger.Info("Sending SIGKILL to cgroup", zap.String("cgroup", cgroupPath))
```
smira:
```suggestion
	logger.Warn("Sending SIGKILL to cgroup", zap.String("cgroup", cgroupPath))
```

pkg/machinery/cel/celenv/celenv_test.go
```diff
@@ -74,3 +75,26 @@ func TestVolumeLocator(t *testing.T) {
 		})
 	}
 }
+
+func TestOOMCgroupScoring(t *testing.T) {
```
smira, line 79: need to add `TestOOMTrigger` simple test

smira (approved):
lgtm to me, with a couple of follow-ups which we can do in separate PRs:

* "log" kills, i.e. create (no more than N) output resources when the cgroup was reaped, with the name, score, matching PIDs, etc.
* integration test where we create stress-ng deployment, scale it above limits, wait N seconds, observer OOM kills (see above), delete scale-ng deployment, and assert cluster is healthy
* documentation - what are the expression inputs, etc. 

as a side note, we could also build default expressions programmatically, but I don't know if we should or not

dsseng:
> as a side note, we could also build default expressions programmatically, but I don't know if we should or not

Not really sure, that could be easier if users want to keep the trigger logic mostly the same, but configure thresholds? Can be done via e.g. Go templating on the CEL expression, but not sure if that would be any easier than copying the default from the docs and modifying it. Maybe we can keep that to a follow-up

smira:
> Not really sure, that could be easier if users want to keep the trigger logic mostly the same, but configure thresholds? Can be done via e.g. Go templating on the CEL expression, but not sure if that would be any easier than copying the default from the docs and modifying it. Maybe we can keep that to a follow-up

No, you can ignore my comment for now.

I mean instead of putting them as string constants we could build them via AST, so we know they are always good.

But he expressions are static here, so it doesn't really matter.

PR #11753, opened 2025-09-03
docs: add docs on multihoming

Iheanacho-ai:
moved around the multihoming docs to reduce cognitive load

### DIFF
--- a/website/content/v1.12/introduction/prodnotes.md
+++ b/website/content/v1.12/introduction/prodnotes.md
@@ -23,7 +23,7 @@ Follow the steps below to build a production-grade Talos cluster that is highly 
 
 ## Step 1: Prepare Your Infrastructure
 
-To create your production cluster infrastructure :
+To create your production cluster infrastructure:
 
 1. Boot your machines using the Talos ISO image
 1. Ensure network access on your nodes.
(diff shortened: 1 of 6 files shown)


website/content/v1.12/introduction/prodnotes.md
```diff
@@ -265,7 +263,12 @@ Follow these steps to patch your machine configuration:
 
 Additionally, you can learn more about [patches]({{< relref "../talos-guides/configuration/patching/" >}}) from the configuration patches documentation.
 
-## Step 10: Apply the Machine Configuration
+## Step 10: Configure Your Multihomed Machines
+
+If your machines are multihomed, i.e., they have more than one IPv4 and/or IPv6 addresses other than loopback, then additional configuration is required.
```
smira, line 268: this should come before applying machine configuration (you need to patch it for multi-homing), but idk what reads better

smira (approved):

website/content/v1.12/introduction/prodnotes.md
Iheanacho-ai: > this should come before applying machine configuration (you need to patch it for multi-homing), but idk what reads better

Yes this step does come before the apply machine configuration step in the docs also.

PR #11757, opened 2025-09-04
feat: stop mounting state partition

### LINKED ISSUE #11608: stop mounting `STATE` volume/partition to the mountpoint

shanduur:
Fixes #11608

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/create_dev.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/create_dev.go
@@ -10,6 +10,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net"
 	"net/netip"
 	"net/url"
@@ -781,7 +782,7 @@ func mergeKubeconfig(ctx context.Context, clusterAccess *access.Adapter) error {
 
 	_, err = os.Stat(kubeconfigPath)
 	if err != nil {
-		if !os.IsNotExist(err) {
+		if !errors.Is(err, fs.ErrNotExist) {
 			return err
 		}
 
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -6,7 +6,9 @@ package block
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
@@ -19,14 +21,14 @@ import (
 	"github.com/siderolabs/gen/xslices"
 	"github.com/siderolabs/go-blockdevice/v2/swap"
 	"go.uber.org/zap"
-	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/internal/pkg/mount/v3"
 	"github.com/siderolabs/talos/internal/pkg/selinux"
-	"github.com/siderolabs/talos/internal/pkg/xfs/fsopen"
 	"github.com/siderolabs/talos/pkg/filetree"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"github.com/siderolabs/talos/pkg/machinery/resources/block"
+	"github.com/siderolabs/talos/pkg/xfs"
+	"github.com/siderolabs/talos/pkg/xfs/fsopen"
 )
 
 type mountContext struct {
@@ -232,6 +234,15 @@ func (ctrl *MountController) Run(ctx context.Context, r controller.Runtime, logg
 						mountStatus.TypedSpec().EncryptionProvider = volumeStatus.TypedSpec().EncryptionProvider
 						mountStatus.TypedSpec().ReadOnly = mountRequest.TypedSpec().ReadOnly
 						mountStatus.TypedSpec().ProjectQuotaSupport = volumeStatus.TypedSpec().MountSpec.ProjectQuotaSupport
+						mountStatus.TypedSpec().Detached = mountRequest.TypedSpec().Detached
+
+						// This needs to be set through accessor, and is not guaranteed to resolve to a valid root.
+						mount, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+						if ok && mount.point != nil {
+							mountStatus.TypedSpec().SetRoot(mount.point.Root())
+						} else {
+							mountStatus.TypedSpec().SetRoot(&xfs.OSRoot{Shadow: filepath.Join(rootPath, mountTarget)})
+						}
 
 						return nil
 					},
@@ -341,7 +352,7 @@ func (ctrl *MountController) handleSymlinkMountOperation(
 	targetPath := filepath.Join(rootPath, target)
 
 	st, err := os.Lstat(targetPath)
-	if err != nil && !os.IsNotExist(err) {
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
 		return fmt.Errorf("failed to stat target path: %w", err)
 	}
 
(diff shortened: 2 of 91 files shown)


pkg/xfs/opentree_test.go
```diff
@@ -107,10 +107,6 @@ func TestOpentree(t *testing.T) {
 		err = rwRoot.OpenFS()
 		require.NoError(t, err)
 
-		t.Cleanup(func() {
```
frezbo, line 110: no need to close anymore?

shanduur: It's closed twice, no need to do it twice (in the `root.Close()`)

pkg/machinery/config/configloader/configloader.go
```diff
@@ -48,8 +49,8 @@ func newConfig(r io.Reader, opt ...Opt) (config config.Provider, err error) {
 }
 
 // NewFromFile will take a filepath and attempt to parse a config file from it.
-func NewFromFile(filepath string) (config.Provider, error) {
-	f, err := os.Open(filepath)
+func NewFromFile(root xfs.Root, filepath string) (config.Provider, error) {
```
smira: let's please keep this backwards compatible (i.e. add a new function which accepts xfs.Root), as this is used in tons of places, and we shouldn't force a change for everyone.

tools/structprotogen/proto/proto.go
```diff
@@ -352,7 +352,7 @@ func PrepareProtoData(pkgsTypes slices.Sorted[*types.Type], constants consts.Con
 func mustFormatTypeName(fieldTypePkg string, fieldType string, declPkg string) (string, string) {
 	importPath, name := formatTypeName(fieldTypePkg, fieldType, declPkg)
 	if name == "" {
-		panic(fmt.Errorf("unknown type %s.%s", fieldTypePkg, fieldType))
+		panic(fmt.Errorf("huh? unknown type %s.%s", fieldTypePkg, fieldType))
```
smira: do we need this?

shanduur: Whoopsie, left this when trying to find why the `structprotogen` was trying to generate spec for unexported fields.

pkg/machinery/resources/block/mount_status.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/cosi-project/runtime/pkg/resource/typed"
 
 	"github.com/siderolabs/talos/pkg/machinery/proto"
+	"github.com/siderolabs/talos/pkg/xfs"
```
smira: this is not possible to have - we can't have `machinery` import `talos`.

so we have two options here - make it `any`, or make it some interface, also see my other comment about adapters

internal/app/machined/pkg/controllers/secrets/encryption_salt.go
```diff
@@ -86,16 +91,22 @@ func (ctrl *EncryptionSaltController) Run(ctx context.Context, r controller.Runt
...
 func (ctrl *EncryptionSaltController) establishEncryptionSalt(ctx context.Context, r controller.ReaderWriter, logger *zap.Logger, mountStatus *block.VolumeMountStatus) error {
-	root := &xfs.UnixRoot{FS: opentree.NewFromPath(mountStatus.TypedSpec().Target)}
-	if err := root.OpenFS(); err != nil {
-		return fmt.Errorf("error opening filesystem: %w", err)
-	}
-
-	defer func() {
-		if err := root.Close(); err != nil {
-			logger.Error("error closing filesystem", zap.Error(err))
+	//nolint:dupl
```
smira: I wonder if we should move this common code to the adapters (we have that as a bridge between resources and additional code we don't want to go into machinery), I mean a function like:

`func(mountStatus *block.VolumeMountStatus, func(root xfs.Root) error) error` which handles getting the root properly

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -232,6 +234,15 @@ func (ctrl *MountController) Run(ctx context.Context, r controller.Runtime, logg
...
 						mountStatus.TypedSpec().ReadOnly = mountRequest.TypedSpec().ReadOnly
 						mountStatus.TypedSpec().ProjectQuotaSupport = volumeStatus.TypedSpec().MountSpec.ProjectQuotaSupport
+						mountStatus.TypedSpec().Detached = mountRequest.TypedSpec().Detached
+
+						// This needs to be set through accessor, and is not guaranteed to resolve to a valid root.
+						mount, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+						if ok && mount.point != nil {
+							mountStatus.TypedSpec().SetRoot(mount.point.Root())
+						} else {
+							mountStatus.TypedSpec().SetRoot(&xfs.OSRoot{Shadow: filepath.Join(rootPath, mountTarget)})
```
smira, line 244: quick question - why can't we use `mount.point.Root()` here always?

smira: can we in other words unify all mount statuses to have a valid root?

shanduur: I've done it like this, because `VolumeTypeDirectory` is never registered as active mount.

https://github.com/siderolabs/talos/blob/0b129f9efdf57dd9692f7cece6b97719a7ccf80e/internal/app/machined/pkg/controllers/block/mount.go#L303-L326

cmd/talosctl/cmd/mgmt/cluster/create/create_dev.go
```diff
@@ -781,7 +782,7 @@ func mergeKubeconfig(ctx context.Context, clusterAccess *access.Adapter) error {
 
 	_, err = os.Stat(kubeconfigPath)
 	if err != nil {
-		if !os.IsNotExist(err) {
+		if !errors.Is(err, fs.ErrNotExist) {
```
frezbo, line 785: why change, is it deprecated? It's easy for new code to miss using `io/fs` and just continue using `os.IsNotExist`

shanduur: It's not deprecated, but it's suggested to switch to `errors.Is(err, fs.ErrNotExist)`

https://pkg.go.dev/os#IsNotExist

> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

frezbo: 🆒 👍

internal/app/machined/pkg/controllers/block/mount.go
smira: ok, got you... probably we should do this refactoring to make all of them have this (doesn't have to be this PR of course0

internal/app/machined/pkg/controllers/secrets/encryption_salt.go
shanduur: I've added adapter for it.

pkg/machinery/resources/block/mount_status.go
shanduur: I think we can be safe enough with `any` here - reading the actual value should be only done through adapter, and setting it is done only in controller.

pkg/machinery/config/configloader/configloader.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/config/configloader/internal/decoder"
 	"github.com/siderolabs/talos/pkg/machinery/config/container"
 	_ "github.com/siderolabs/talos/pkg/machinery/config/types" // import config types to register them
+	"github.com/siderolabs/talos/pkg/xfs"
```
smira: we still have the issue of importing `talos/pkg` from `pkg/machinery`... I guess we need to put in an interface stub here

smira (approved):
👍  with the note of import inversion

PR #11759, opened 2025-09-04
feat: create detached tmpfs

### LINKED ISSUE #10159: Use a detached `tmpfs` mount in `EtcFileController`
## Feature Request

use a detached `tmpfs` mount created used `fsopen` and use the resulting fd to write etc files in there and `bind` mount them as read only at `/etc`

shanduur:
Fixes #10159

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -441,11 +441,11 @@ ARG GOAMD64
 RUN --mount=type=cache,target=/.cache,id=talos/.cache GOOS=linux GOARCH=amd64 GOAMD64=${GOAMD64} go build ${GO_BUILDFLAGS} -ldflags "${GO_LDFLAGS} ${GO_MACHINED_LDFLAGS} -dumpdep" ./internal/app/machined \
     |& go tool github.com/aarzilli/whydeadcode > deadcode.txt
 RUN if [[ -s deadcode.txt ]]; then \
-        echo "Dead code elimination problem found:"; \
-        cat deadcode.txt; \
-        exit 1; \
+    echo "Dead code elimination problem found:"; \
+    cat deadcode.txt; \
+    exit 1; \
     else \
-        echo "No dead code elimination issues found"; \
+    echo "No dead code elimination issues found"; \
     fi
 
 # The init target builds the init binary.
--- a/internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
+++ b/internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
@@ -40,6 +40,8 @@ import (
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
 	runtimelogging "github.com/siderolabs/talos/internal/app/machined/pkg/runtime/logging"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/system"
+	"github.com/siderolabs/talos/internal/pkg/xfs"
+	"github.com/siderolabs/talos/internal/pkg/xfs/fsopen"
 	"github.com/siderolabs/talos/pkg/logging"
 	talosconfig "github.com/siderolabs/talos/pkg/machinery/config/config"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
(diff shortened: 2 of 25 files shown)


internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -87,6 +89,44 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
 		return err
 	}
 
+	var (
+		etcRoot        xfs.Root
+		networkEtcRoot xfs.Root
+	)
+
+	opentreeOnAnonymous, err := xfs.HasKernel("6.15.0")
```
smira: I think we should put this under a check that we are in a container mode, as if we run with Talos kernel, we should never try to fallback to non-secure tmpfs

internal/app/machined/pkg/controllers/files/etcfile_fd.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return mount.BindReadonlyFd(int(f.Fd()), dst)
+}
+
+// UpdateFileFd is like `os.WriteFile`, but it will only update the file if the
+// contents have changed.
+func UpdateFileFd(root xfs.Root, filename string, contents []byte, mode os.FileMode, selinuxLabel string) error {
+	oldContents, err := xfs.ReadFile(root, filename)
+	if err == nil && bytes.Equal(oldContents, contents) {
+		return selinux.FSetLabel(root, filename, selinuxLabel) // TODO(shanduur): failing
```
frezbo: this TODO still valid?

shanduur: No, it's fixed now.

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
shanduur: > as if we run with Talos kernel

Sure, but we can also check for it if we run in container, and only then fallback to insecure. This should work just fine in container - works for me on `6.16.x` kernel.

dsseng: Agreed, I guess Andrey meant only query kernel version in container mode, and when we have both an old kernel and container mode do a fallback. We should make use of detached tmpfs in containers for sure

internal/pkg/selinux/selinux.go
```diff
@@ -104,6 +105,79 @@ func SetLabel(filename string, label string, excludeLabels ...string) error {
 	return nil
 }
 
+// FGetLabel gets label for file, directory or symlink (not following symlinks) using provided root.
```
dsseng, line 110: Does it actually not follow symlinks?

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
dsseng: And maybe we need a comment here: what features we need from the kernel

internal/pkg/selinux/selinux.go
shanduur: I've added `O_NOFOLLOW` to the open file call, to not follow symlinks.

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
shanduur: I've moved the check to runtime package, and I am only checking it if the Talos is running in Docker, PTAL! ~thanks

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -87,6 +89,63 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
...
+			"tmpfs",
+			fsopen.WithStringParameter("mode", "0755"),
+			fsopen.WithStringParameter("size", "4M"),
+		),
+	}
+
+	networkBindMountTarget = constants.SystemResolvedPath
+
+	// While running in container, we don't have controll over kernel version
+	// shipped with the machine. If the kernell is not supportin open_tree call
```
dsseng: nit:

```suggestion
	// While running in container, we don't have control over the kernel version
	// shipped with the machine. If the kernel does not support open_tree syscall
```

shanduur: Fixed, thanks!

internal/app/machined/pkg/runtime/v1alpha2/v1alpha2_controller.go
```diff
@@ -87,6 +89,63 @@ func (ctrl *Controller) Run(ctx context.Context, drainer *runtime.Drainer) error
...
+	}
+
+	networkBindMountTarget = constants.SystemResolvedPath
+
+	// While running in container, we don't have control over kernel version
+	// shipped with the machine. If the kernel does not support open_tree syscall
+	// on anonymous filesystem file descriptors, we need to fallback to the classic,
+	// less secure mode. This capability was added in kernel 6.15.0.
+	if ctrl.v1alpha1Runtime.State().Platform().Mode().InContainer() {
+		opentreeOnAnonymous, err := runtime.HasKernel("6.15.0")
```
frezbo: I wonder if there a better way to check if the kernel supports syscall, seems we can check the return code and the error returned, this should also handle cases where kernel has custom names

shanduur: So something like this?

```
runtime.KernelCapabilities().OpentreeOnAnonymousFS()
```

And then inside create tmpfs, try to opentree on it, close it and validate all the errors?

smira:
I like the code, but it seems to fail for me locally with:

```
[   25.192694] [talos] controller failed {"component": "controller-runtime", "controller": "files.CRIRegistryConfigController", "error": "bind mount failed for \"/etc/cri/conf.d/hosts\": no such file or directory"}
```

shanduur:
@smira I think I got this. At some point I messed `src` and `dst` when changing code around bind mounting the directory.

Dockerfile
```diff
@@ -1129,8 +1129,8 @@ ARG TESTPKGS
 ENV PLATFORM=container
 ARG GO_LDFLAGS
 RUN --security=insecure --mount=type=cache,id=testspace,target=/tmp --mount=type=cache,target=/.cache,id=talos/.cache go test \
-    -ldflags "${GO_LDFLAGS}" \
-    -covermode=atomic -coverprofile=coverage.txt -coverpkg=${TESTPKGS} -p 4 ${TESTPKGS}
+    -ldflags "${GO_LDFLAGS}"\
+    -v -covermode=atomic -coverprofile=coverage.txt -coverpkg=${TESTPKGS} -p 4 ${TESTPKGS}
```
smira: we dropped `-v` to make logs a bit more readable (only failures are shown)

shanduur: Fixed.

smira (approved):

PR #11762, opened 2025-09-04
feat: implement multi-doc StaticHostConfig

### LINKED ISSUE #10953: Network Config:  `ExtraHostConfig`
### Static Hosts

```yaml
version: v1alpha1
kind: StaticHostConfig
name: 10.3.4.5 # IPv4 or IPv6, unique
hostnames:
  - my-server1
  - my-server1.example.org
```

This literally renders to `/etc/hosts`.

**Replaces** `machine.network.extraHostEntries` (can be merged with)

smira:
Fixes #10953

This the very first migrated document. I picked up a very simple configuration document to build some initial idea how we handle migration of the config documents to the new format.


pkg/machinery/config/types/v1alpha1/v1alpha1_network_bridge_test.go
frezbo, line 1: why this file named `network_bridge_test`?

smira: There is a comment down below, but it's testing a "bridge" between legacy v1alpha1 monolithic config and new multi-doc.

Verifying that from the config interface point of view, there is no difference between using old config and new config, and you can even mix them together (in this case).

frezbo (approved):
🆒

pkg/machinery/config/types/v1alpha1/v1alpha1_network_bridge_test.go
frezbo: I assumed it had something to do with network bridges 😂

PR #11764, opened 2025-09-04
docs: clarify vmware cni limitation

Iheanacho-ai:
clarified that using VMXNET network interfaces in VMware can cause networking issues with CNIs that rely on packet encapsulation.

sorts #10131

### DIFF
--- a/website/content/v1.10/talos-guides/install/virtualized-platforms/vmware.md
+++ b/website/content/v1.10/talos-guides/install/virtualized-platforms/vmware.md
@@ -66,8 +66,16 @@ $ talosctl validate --config worker.yaml --mode cloud
 worker.yaml is valid for cloud mode
 ```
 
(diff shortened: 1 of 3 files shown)


smira:
Talos now supports disabling checksum offloading, which is a better fix than any mentioned here

Iheanacho-ai:
> oading, which is a better fix than any mentioned here

this is the only resource i found that mentioned disabling checksum offloading: https://www.talos.dev/v1.11/talos-guides/network/ethernet-config/ 

But it does not mention how to do so. How do we disable checksum offloading?

By patching the machine configuration?  @smira

smira:
> > oading, which is a better fix than any mentioned here
> 
> this is the only resource i found that mentioned disabling checksum offloading: [talos.dev/v1.11/talos-guides/network/ethernet-config](https://www.talos.dev/v1.11/talos-guides/network/ethernet-config/)
> 
> But it does not mention how to do so. How do we disable checksum offloading?
> 
> By patching the machine configuration? @smira

Yes, I guess you need to try and verify the fix.

I think it does workaround issue with vxlan.

PR #11765, opened 2025-09-04
docs: clarify that talos does not support intermediate ca

Iheanacho-ai:
explained that you would need reissue cluster certificates if you are migrating from kubeadm as talos does not support intermediate ca

sorts #11691

### DIFF
--- a/website/content/v1.10/talos-guides/configuration/certificate-authorities.md
+++ b/website/content/v1.10/talos-guides/configuration/certificate-authorities.md
@@ -5,7 +5,7 @@ aliases:
   - ../../guides/configuring-certificate-authorities
 ---
 
-## Appending the Certificate Authority
+## Appending the Certificate Authority (CA)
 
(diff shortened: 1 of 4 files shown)


website/content/v1.10/talos-guides/configuration/certificate-authorities.md
```diff
@@ -26,3 +26,17 @@ This configuration can be also applied in maintenance mode.
 
 Please note that if the `STATE` partition is encrypted, the CA certificates will be only be loaded after the partition is unlocked.
 So the encryption method should allow unlocking the partition without the need for a CA certificate.
+
+### Limitations
```
smira: I think this is a wrong document. Talos does support intermediate CAs for generic trust of external services (which is what this file is about).

you want instead to add it to https://www.talos.dev/v1.11/advanced/migrating-from-kubeadm/

Iheanacho-ai: moved it

smira (approved):

PR #11789, opened 2025-09-08
docs: change the get IP addressses command

Iheanacho-ai:
changed the command to a more robust command

sorts siderolabs/docs#89

### DIFF
--- a/website/content/v1.10/talos-guides/install/virtualized-platforms/vagrant-libvirt.md
+++ b/website/content/v1.10/talos-guides/install/virtualized-platforms/vagrant-libvirt.md
@@ -128,7 +128,7 @@ worker-node-1             running (libvirt)
 Find out the IP addresses assigned by the libvirt DHCP by running:
 
 ```bash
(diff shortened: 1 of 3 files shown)


website/content/v1.10/talos-guides/install/virtualized-platforms/vagrant-libvirt.md
```diff
@@ -128,7 +128,7 @@ worker-node-1             running (libvirt)
 Find out the IP addresses assigned by the libvirt DHCP by running:
 
 ```bash
-virsh list | grep vagrant | awk '{print $2}' | xargs -t -L1 virsh domifaddr
```
smira, line 131: this doesn't look to be equivalent, grep `vagrant` is lost

rothgar: `vagrant` is intentionally missing from the command from the original bug report. The text "vagrant" only showed up in the VM name if the directory where the original nodes were created were in a folder called "vagrant"

Hopefully, this command is more robust for people that name the folder something else.

PR #11792, opened 2025-09-09
fix: re-create cgroups when restarting runners

### LINKED ISSUE #11785: Containerized services have no resource limits after they restart
### Description

If a service like Kubelet restarts, containerd (system engine) removes its cgroup and then re-creates it, without keeping limits configured by machined.

A similar problem exists for services started by the Process runner: if a cgroup is killed via `cgroup.kill`, it will fail to restart, because it must first be recreated. This problem will be addressed as well.

### Logs

Nothing special, but comparing cgroups overviews can show the limits are lost after a service restart.

### Environment

- Talos version: [`talosctl version --nodes <problematic nodes>`] e5acb10fcceba69060507a35caea21281bdc71cc
- Kubernetes version: [`kubectl version`] v1.34.0
- Platform: QEMU

dsseng:
Make sure processes are only launched into freshly-created cgroups
with all limits set when they are restarted.

This also allows processes to restart after cgroup being killed via the
cgroup.kill mechanism.

Fixes #11785

### DIFF
--- a/internal/app/machined/pkg/system/runner/containerd/containerd.go
+++ b/internal/app/machined/pkg/system/runner/containerd/containerd.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"os"
 	"syscall"
 	"time"
 
@@ -167,6 +168,21 @@ func (c *containerdRunner) Run(eventSink events.Recorder) error {
 	if err != nil {
 		return fmt.Errorf("error creating log: %w", err)
 	}
+
+	cg, err := cgroup.CreateCgroup(c.opts.CgroupPath)
+	if err != nil {
+		return fmt.Errorf("error creating cgroup: %w", err)
+	}
+
+	// If the task is not cleaned up by containerd or another error
+	// happens during the lifecycle, remove the cgroup before exiting
+	// if one still exists
+	defer func() {
+		err := cg.Delete()
+		if err != nil && !os.IsNotExist(err) {
+			eventSink(events.StateStopping, "Failed to remove cgroup for %s, %s", c, err)
+		}
+	}()
 
 	defer logW.Close() //nolint:errcheck
 
--- a/internal/app/machined/pkg/system/runner/process/process.go
+++ b/internal/app/machined/pkg/system/runner/process/process.go
@@ -413,7 +413,20 @@ func setSchedulingPolicy(p *processRunner, pid int, schedulingPolicy uint) error
 	return nil
 }
 
+//nolint:gocyclo
 func (p *processRunner) run(eventSink events.Recorder) error {
+	cg, err := cgroup.CreateCgroup(p.opts.CgroupPath)
+	if err != nil {
+		return fmt.Errorf("error creating cgroup: %w", err)
+	}
+
+	defer func() {
+		err := cg.Delete()
+		if err != nil {
+			eventSink(events.StateStopping, "Failed to remove cgroup for %s, %s", p, err)
+		}
+	}()
+
 	cmdWrapper, err := p.build()
 	if err != nil {
 		return fmt.Errorf("error building command: %w", err)
--- a/internal/app/machined/pkg/system/services/apid.go
+++ b/internal/app/machined/pkg/system/services/apid.go
@@ -177,12 +177,7 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: filepath.Dir(constants.APISocketPath), Source: filepath.Dir(constants.APISocketPath), Options: []string{"rbind", "rw"}},
 	}
 
-	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
-		mounts = append(
-			mounts,
-			specs.Mount{Type: "bind", Destination: "/usr/etc/in-container", Source: "/usr/etc/in-container", Options: []string{"bind", "ro"}},
-		)
-	}
+	mounts = bindMountContainerMarker(mounts)
 
 	env := []string{
 		constants.TcellMinimizeEnvironment,
--- a/internal/app/machined/pkg/system/services/extension.go
+++ b/internal/app/machined/pkg/system/services/extension.go
@@ -176,12 +176,7 @@ func (svc *Extension) Runner(r runtime.Runtime) (runner.Runner, error) {
 
 	mounts := append([]specs.Mount{}, svc.Spec.Container.Mounts...)
 
-	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
-		mounts = append(
-			mounts,
-			specs.Mount{Type: "bind", Destination: "/usr/etc/in-container", Source: "/usr/etc/in-container", Options: []string{"bind", "ro"}},
-		)
-	}
+	mounts = bindMountContainerMarker(mounts)
 
 	envVars, err := svc.parseEnvironment()
 	if err != nil {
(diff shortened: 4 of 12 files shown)


internal/app/machined/pkg/system/runner/containerd/containerd.go
```diff
@@ -168,6 +171,21 @@ func (c *containerdRunner) Run(eventSink events.Recorder) error {
...
+	if err != nil {
+		return fmt.Errorf("error creating cgroup: %w", err)
+	}
+
+	// If the task is not cleaned up by containerd or another error
+	// happens during the lifecycle, remove the cgroup before exiting
+	// if one still exists
+	defer func() {
+		err := cg.Delete()
+		if err != nil && !strings.Contains(err.Error(), "no such file or directory") {
```
frezbo: can't we check for os.IsnotExist err?

smira: yes, please...

smira (approved):

internal/app/machined/pkg/system/runner/process/process.go
```diff
@@ -413,7 +413,22 @@ func setSchedulingPolicy(p *processRunner, pid int, schedulingPolicy uint) error
...
 func (p *processRunner) run(eventSink events.Recorder) error {
+	if cgroups.Mode() == cgroups.Unified {
+		cg, err := cgroup.CreateCgroupV2(p.opts.CgroupPath)
+		if err != nil {
+			return fmt.Errorf("error creating cgroup: %w", err)
+		}
+
+		defer func() {
+			err := cg.Delete()
+			if err != nil {
```
frezbo: no need for the os.IsNotExist error here?

shanduur (approved):

internal/app/machined/pkg/system/runner/containerd/containerd.go
```diff
@@ -168,6 +170,23 @@ func (c *containerdRunner) Run(eventSink events.Recorder) error {
 		return fmt.Errorf("error creating log: %w", err)
 	}
 
+	if cgroups.Mode() == cgroups.Unified {
```
smira: I think we need a bit of layer of abstraction here:

* we want here to create a cgroup and clean it up afterwards, but we don't care if it's cgroupsv1 or cgroupsv2
* I would like to have a simple package I can call into, `cg, err := cgroup.Create(path)\n defer cg.Destroy()`

smira: right now it has a bug that we only do it for cgroupsv2

internal/pkg/cgroup/cgroup.go
```diff
@@ -56,3 +62,133 @@ func Path(cgroupPath string) string {
...
+
+		return zeroT
+	}
+
+	return v
+}
+
+// CreateCgroupV2 creates a cgroupv2, with resources limits if configured.
+func CreateCgroupV2(name string) (*cgroup2.Manager, error) {
+	groups := map[string]*cgroup2.Resources{
```
smira: we could refactor this into a function

```go
func GetCgroupV2Resources(name string) *cgroupv2.Resources {
if inContainer || ... { return nil }
switch name {
name1: return ...
case name2: ...
default: return nil
```

Creating a map to select from it looks odd to me

internal/pkg/cgroup/cgroup.go
```diff
@@ -56,3 +62,133 @@ func Path(cgroupPath string) string {
...
+				Swap: pointer.To[int64](0),
+			},
+			CPU: &cgroup2.CPU{
+				Weight: pointer.To[uint64](MillicoresToCPUWeight(MilliCores(constants.CgroupTrustdMillicores))),
+			},
+		},
+	}
+
+	inContainer := false
+	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
```
smira: should we extract this into a single place, and use `sync.OnceValue` for it? 

we already have like 6 references to it

internal/app/machined/pkg/system/runner/process/process.go
dsseng: Yes, in case of containerd the containerd will have removed the cgroup on container exit (but not if we haven't spawned the container and exited early due to an error). I cannot see in what situation might the cgroup be cleaned up in case of a directly launched process

internal/app/machined/pkg/system/runner/containerd/containerd.go
dsseng: > right now it has a bug that we only do it for cgroupsv2

cgroupv1 doesn't seem to need it, as we neither set resources for it nor have cgroup.kill there. So I intend to not change cgroupv1 behavior at all, and I do not think the bug affects cgv1.

I guess no more abstraction is needed, it's just create, destroy, handle errors. However that's guarded behing cgroupv2 check. Maybe we could move the check into the cgroup package, but that would look like cgv1 is now also re-created

internal/pkg/cgroup/cgroup.go
dsseng: needs to be done, what package should it be in? Machinery? Also perhaps needs to be constified

internal/app/machined/pkg/system/runner/containerd/containerd.go
smira: it's just hard to read and use - we have three callers of this code right now, and every place does pre-checks

internal/pkg/cgroup/cgroup.go
smira: certainly not machinery, as it's Talos-internal, it doesn't make sense outside of Talos, so could be `internal/containermode` ??

dsseng: Would make sense, will place under internal/ directly

internal/app/machined/pkg/system/services/trustd.go
```diff
@@ -154,10 +154,10 @@ func (t *Trustd) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: filepath.Dir(constants.TrustdRuntimeSocketPath), Source: filepath.Dir(constants.TrustdRuntimeSocketPath), Options: []string{"rbind", "ro"}},
 	}
 
-	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
+	if _, err := os.Stat(constants.ContainerMarkerFilePath); err == nil {
```
frezbo: should be using `containermode.InContainer`

internal/app/machined/pkg/system/services/extension.go
```diff
@@ -176,10 +176,10 @@ func (svc *Extension) Runner(r runtime.Runtime) (runner.Runner, error) {
 
 	mounts := append([]specs.Mount{}, svc.Spec.Container.Mounts...)
 
-	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
+	if _, err := os.Stat(constants.ContainerMarkerFilePath); err == nil {
```
frezbo: should be using `containermode.InContainer`

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -177,10 +177,10 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: filepath.Dir(constants.APISocketPath), Source: filepath.Dir(constants.APISocketPath), Options: []string{"rbind", "rw"}},
 	}
 
-	if _, err := os.Stat("/usr/etc/in-container"); err == nil {
+	if _, err := os.Stat(constants.ContainerMarkerFilePath); err == nil {
```
frezbo: should be using `containermode.InContainer`

dsseng: Would make sense, yeah. We'll still reference the file in next lines though. Will do (and maybe this common code can be shared), thanks

Issue #9925, opened 2024-12-11
image-cache: allow updating on and accessing from a live cluster


srgvg:
### Description

I just did some initial tests of Talos `v1.9.0-beta1` which includes the image-cache feature, and this seems to work as expected.

To be able to fully manage a stand-alone, isolated and air-gapped cluster, it would be neat to have some extra features:
* update (push new  and delete old from) the cache on a live cluster
* access the cache as a proper registry/oci registry over http, to allow caching more than just container images: e.g. Helm charts or FluxCD artifacts. The FluxCD source controller OCIRepository object can be configured with a "proxy server". This proxy server should point to the local cache.

### Related
* Whilst adding non container images like Helm charts, to the cache is currently possible, it does not work with the `talosctl images cache-create` `--image-layer-cache-path` switch. I tried looking at the `crane` project to find out if that misses some switch of its own, or is a bug - as they do support other OCI types - but couldn't exactly pinpoint it.

smira:
Please keep in mind that you can also deploy a registry mirror in the environment as well which would support all of the above in a more native way.

I don't think we would allow external network access to the image cache for security reasons. 

Updating the cache on a live cluster might be on a radar eventually.

a1994sc:
The project [`zarf`](https://github.com/zarf-dev/zarf) does a very similar thing of creating an intra-cluster oci registry; I am very much looking forward to trying this feature out in the near future.... having core images cached on disk sounds amazing.

marcopornic:
Hello everyone,

In this video comments [https://www.youtube.com/watch?v=Ir2m11Cijas](https://www.youtube.com/watch?v=Ir2m11Cijas), "SideroLabs" says 

> you can update the images on the ISO and plug it into the system. If you reboot with updated images the cache will be updated

I have generated a new ISO with new images and mount it to my VM as a DVD drive. Talos try to copy images in cache but destination file /system/disk/blob/ is read-only file system. 

Did I get it wrong ?

Thx !

PR #11802, opened 2025-09-10
docs: add docs about live image cache update

Iheanacho-ai:
explained that you cannot truly update the image cache on a live node

sorts #9925


website/content/v1.10/talos-guides/configuration/image-cache.md
```diff
@@ -121,6 +121,15 @@ If the disk image is used, the `IMAGECACHE` volume doesn't need to be configured
 
 See [disk management]({{< relref "./disk-management#machine-configuration" >}}) for more information on volume configuration.
 
+## Updating the Cache
+
+The image cache is populated during installation from the boot media and stored on disk.
+
+Once Talos is running, this cache becomes read-only and cannot be updated on a live node.
```
smira, line 128: why is that? do you have this tested that it doesn't work?

Iheanacho-ai: i havent really tried this one yet, i was going off your comment, here:
https://github.com/siderolabs/talos/issues/9925#issuecomment-2536544813

smira: That comment is very old, I haven't tried, but I do believe if you mount an ISO/USB stick which looks like an Image Cache, modern Talos should copy its contents over in live mode (for a running machine).

PR #11823, opened 2025-09-11
docs: add docs on updating image cache

Iheanacho-ai:
sorts ##9925

### DIFF
--- a/website/content/v1.10/talos-guides/configuration/image-cache.md
+++ b/website/content/v1.10/talos-guides/configuration/image-cache.md
@@ -121,6 +121,29 @@ If the disk image is used, the `IMAGECACHE` volume doesn't need to be configured
 
 See [disk management]({{< relref "./disk-management#machine-configuration" >}}) for more information on volume configuration.
 
+## Updating the Image Cache
+
+The image cache is initially populated during installation from the boot media (ISO or disk image) and stored on disk.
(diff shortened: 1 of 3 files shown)


website/content/v1.10/talos-guides/configuration/image-cache.md
```diff
@@ -121,6 +121,22 @@ If the disk image is used, the `IMAGECACHE` volume doesn't need to be configured
...
 
+## Updating the Image Cache
+
+The image cache is initially populated during installation from the boot media (ISO or disk image) and stored on disk.
+
+To update the image cache on a live node, attach new cache media (ISO or USB) as a secondary device by:
+
+* Booting the node from its installed disk.
+* [Bundling the updated image cache into a Talos ISO image]({{< relref "#iso" >}}).
+* Inserting the bundled ISO into the Talos node.
```
rothgar: I think this is supposed to work from a USB drive too. If they create an image cache on a USB drive and the drive has the proper `TALOS_` label I think it's supposed to copy over updated images. I haven't personally tested that though.

smira: it should work with anything we describe in the Image Cache docs - it doesn't need to have Talos installation or be bootable itself.

Iheanacho-ai: @smira done

PR #11840, opened 2025-09-16
feat: enforce more KSPP and hardening sysctls

smira:
Source: https://github.com/a13xp0p0v/kernel-hardening-checker/blob/76821aabeb62a9bf4130bae65d467ca2561bebd5/kernel_hardening_checker/checks.py

Filter some sysctls (if we don't need them because of kernel config), or try not to break things too much.


smira:
> Feels like we should document what all sysctls we enable in a page

I think that might get outdated, but we can say that `talosctl get kernelparamstatus` always gives you a list

PR #11845, opened 2025-09-17
chore: drop docs folder

frezbo:
Drop the `docs` folder, we are not using it at all.


smira (approved):

docs/community/project-layout.md
```diff
@@ -1,56 +0,0 @@
-# Project Layout Standards
```
Orzelius, line 1: I found this doc useful when I first got interested in talos. I think it still makes sense to have this information somewhere for potential contributors

PR #11850, opened 2025-09-17
feat: ship nft binary with Talos rootfs

### LINKED ISSUE #11846: nft binary missing in node $PATH
### Description
Talos is missing the `nft` binary to interact with directly with nftables.

This is required by the https://github.com/kubernetes-sigs/knftables Go library that's used in core Kubernetes components.

In particular I found this issue when I was trying to configure the `portmap` CNI plugin to use nftables.

I wanted to do this because I needed to add a filter to `portmap` that can't be expressed as a single iptables rule:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  labels:
    k8s-app: flannel
    tier: node
  name: kube-flannel-cfg
  namespace: kube-system
data:
  cni-conf.json: |
    {
      "name": "cbr0",
      "cniVersion": "1.0.0",
      "plugins": [
        {
          "type": "flannel",
          "delegate": {
            "hairpinMode": true,
            "isDefaultGateway": true
          }
        },
        {
          "type": "portmap",
          "capabilities": {"portMappings": true},
          "backend": "nftables"
        }
      ]
    }
  net-conf.json: |-
    {
      "Network": "10.244.0.0/16",
      "Backend": {
        "Type": "vxlan",
        "Port": 4789
      }
    }
```

### Logs

After applying and restarting, I run a pod with a hostPort and I get this event:

`hostport-test        1s (x17 over 3m33s)      Warning   FailedCreatePodSandBox    Pod/hostport-test                                   (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "b2f90857da3f8367b0782f972fac5c148b4be82860fdfb477978debb61a8163f": plugin type="portmap" failed (add): could not find nftables binary: exec: "nft": executable file not found in $PATH`

### Environment

- Talos version: v1.10.7
- Kubernetes version: v1.34.0
- Platform: bare metal amd64

Thanks

smira:
Fixes #11846

Also bump Linux to 6.16.7 via pkgs.


oguzkilcan (approved):

internal/app/machined/pkg/controllers/network/nftables_chain_test.go
```diff
@@ -29,6 +29,13 @@ type NfTablesChainSuite struct {
 
 func (s *NfTablesChainSuite) nftOutput() string {
 	out, err := exec.CommandContext(s.T().Context(), "nft", "list", "table", "inet", "talos-test").CombinedOutput()
+	if err != nil {
```
frezbo, line 32: is this related to bringing in the new binary? or something changed?

smira: these tests now actually started running in the CI (previously `nft` binary was missing, and the tests got skipped), so it revealed a race that the controller might create a table not immediately.

previously I only ran them locally, so now we have a better flow I would say, so I had to fix the tests

PR #11858, opened 2025-09-18
feat: implement link configuration

### LINKED ISSUE #10957: Network Config: `LinkConfig`
### Link Configuration

```yaml
version: v1alpha1
kind: LinkConfig
name: en0sp2 # either primary name or an alias
mtu: 1500 # optional
up: true # defaults to true, can be brought down with `false`
addresses:
  - 10.3.5.1/24
  - 20a1:fe1d::1/64
routes:
  - destination: 10.3.4.0/24 # route destination, if not set defaults to * (default gateway)
    gateway: 10.3.4.1 # if not set, creates "on-link" route
    source: 10.3.5.1 # optional
    metric: 1024 # optional, defaults to 1024
    mtu: 1500 # optional, defaults to unset
    table: main # optional, defaults to main
```

This configures basic settings for the link, including static addressing.

- `name` name of the link (or alias); `LinkConfig` never creates links, it only applies to already existing links
- `mtu` overrides link MTU
- `up` defaults to `true` (if the `LinkConfig` document is present), can be used to force link down with `false`
- `addresses` is a list of static addresses assigned to the link
- `routes` block configures static routes, almost identical to existing `v1alpha1.Config`

**Replaces** `machine.network.interfaces.{addresses,routes,up,mtu}` and other nested configuration.

It should be machine configuration validation error to have a `LinkConfig` and `BondConfig`, `VLANConfig` , `BridgeConfig` for the same link name.

smira:
New multi-doc way to apply static configuration to network links.

Fixes #10957

Also implements related changes:

* new-style auto-DHCP mode - run DHCP as long as there is no new-style
  config document
* refactoring of controllers/tests which were touched in this PR
* implemented embedded config structs handling in docgen -> Markdown and
  docgen -> JSON schema


internal/app/machined/pkg/controllers/network/operator_config.go
```diff
@@ -215,77 +230,61 @@ func (ctrl *OperatorConfigController) Run(ctx context.Context, r controller.Runt
 			}
 		}
 
-		// build configuredInterfaces from linkSpecs in `network-config` namespace
-		// any link which has any configuration derived from the machine configuration or platform configuration should be ignored
-		configuredInterfaces := map[string]struct{}{}
+		// run default DHCP operators is enabled
```
frezbo:
```suggestion
		// run default DHCP operators if enabled
```

frezbo (approved):
🆒

internal/app/machined/pkg/controllers/network/operator_config.go
smira: thank you, fixed

PR #11875, opened 2025-09-22
feat: release cloud image using factory

### LINKED ISSUE #11454: Published Talos AMIs should include ecr-credential-helper extension
## Feature Request

When publishing AMIs to AWS we should include the ecr-credential-helper by default. It's a safe assumption that users running in AWS will want to use ECR in a variety of ways (e.g. pull through cache for k8s/talos, application workloads).

A user can leave the helper unconfigured and it wouldn't affect image pulls, but the helper is required even if they give the instance access to ECR.

### Description

We currently don't include the extension so users have to build their own AMI or perform an upgrade to include the extension before pulling images from ECR.

shanduur:
Rather than building images manually, retrieve them from the image factory to guarantee that the correct base schematic is applied and that all required default extensions are consistently included.

Fixes #11454

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2025-09-25T16:16:48Z by kres fdbc9fc.
+# Generated on 2025-09-29T11:30:50Z by kres eb905b6-dirty.
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
@@ -5030,16 +5030,10 @@ jobs:
           PLATFORM: linux/amd64,linux/arm64
         run: |
           make images
-      - name: cloud-images
-        env:
-          PLATFORM: linux/amd64,linux/arm64
-        run: |
-          make cloud-images
       - name: Install Cosign
         uses: sigstore/cosign-installer@v3
       - name: Sign artifacts
         run: |
-          cosign sign-blob --output-signature _out/cloud-images.json.sig --yes _out/cloud-images.json
           cosign sign-blob --output-signature _out/initramfs-amd64.xz.sig --yes _out/initramfs-amd64.xz
           cosign sign-blob --output-signature _out/initramfs-arm64.xz.sig --yes _out/initramfs-arm64.xz
           cosign sign-blob --output-signature _out/metal-amd64.iso.sig --yes _out/metal-amd64.iso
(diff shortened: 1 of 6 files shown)


pkg/machinery/gendata/data/ecr-credential-provider
```diff
@@ -0,0 +1 @@
+v1.34.0
```
smira: we can't/shouldn't pin it this way, I don't think we can or should even pin in Talos itself

The only supported way is https://github.com/siderolabs/extensions#installing-extensions

And the only way this would work reliably is for a release of Talos (once the extensions are tagged as well).

smira: we could probably run the `imager` in the CI with `make image-aws` and pass extension as extra imager args, but getting correct version/digest is tricky due to the tagging problem above

smira:
After thinking more about this, I don't see a good way.

If we add extension via an `imager`, we create a path which is not good from Image Factory perspective - no embedded schematic, etc.

If we build using Image Factory, we need:

* in release path, before cloud-images run - we need to ensure images are tagged, so we need to pause the workflow
* in non-release path (no tag) - we can't add via Image Factory at all for "development" version

If we don't follow Image Factory, it would be harder for people to upgrade their machine in the future.

frezbo:
> After thinking more about this, I don't see a good way.
> 
> If we add extension via an `imager`, we create a path which is not good from Image Factory perspective - no embedded schematic, etc.
> 
> If we build using Image Factory, we need:
> 
>     * in release path, before cloud-images run - we need to ensure images are tagged, so we need to pause the workflow
> 
>     * in non-release path (no tag) - we can't add via Image Factory at all for "development" version
> 
> 
> If we don't follow Image Factory, it would be harder for people to upgrade their machine in the future.

hmm can't we just split out the `cloud-images` as a manually step after release is done? then we make sure we can build with a tagged extension release.

From image factory we add the schematic always for aws

smira:
> > After thinking more about this, I don't see a good way.
> > If we add extension via an `imager`, we create a path which is not good from Image Factory perspective - no embedded schematic, etc.
> > If we build using Image Factory, we need:
> > ```
> > * in release path, before cloud-images run - we need to ensure images are tagged, so we need to pause the workflow
> > 
> > * in non-release path (no tag) - we can't add via Image Factory at all for "development" version
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If we don't follow Image Factory, it would be harder for people to upgrade their machine in the future.
> 
> hmm can't we just split out the `cloud-images` as a manually step after release is done? then we make sure we can build with a tagged extension release.
> 
> From image factory we add the schematic always for aws

yes, we could, but there will be no good way to test it as well... or make it a separate workflow

.github/workflows/ci.yaml
```diff
@@ -4858,6 +4858,91 @@ jobs:
...
+          cosign sign-blob --output-signature sha512sum.txt.sig --yes sha512sum.txt
+      - name: release
+        uses: crazy-max/ghaction-github-release@v2
+        with:
+          body_path: _out
+          draft: "true"
+          files: |-
+            _out/cloud-images.json
+            _out/sha*.txt
+            _out/*.sig
```
shanduur: This would override existing `sha*.txt` and `sha*.txt.sig` files, TODO.

shanduur: I've dropped them as of now.

.kres.yaml
```diff
@@ -216,6 +212,25 @@ spec:
...
             generateSignatures: true
             releaseNotes: RELEASE_NOTES.md
+    - name: cloud-images
+      dispatchable: true
+      sops: true
+      buildxOptions:
+        enabled: true
+      runnerGroup: generic # this is not compute intensive - we are pulling images from factory
+      conditions:
+        - only-on-tag # TODO: only after publishing release
```
smira: I wonder if we should make `TAG` to be an input here, so we can test this flow by running it e.g. right now for alpha.0?

shanduur:
I trimmed it a little, but should work now, and is not required to run on any particular release/tag/commit - it takes two arguments, which are defining what tag to pull and release, and to which release attach new generated file.

- https://github.com/shanduur/siderolabs-talos-dispatch-test/actions/runs/18100085741
- https://github.com/shanduur/siderolabs-talos-dispatch-test/releases/tag/v1.11.2

smira (approved):
👍  let's test it once merged for 1.12.0-alpha.0!

.kres.yaml
```diff
@@ -216,6 +212,28 @@ spec:
...
+            CLOUD_IMAGES_EXTRA_ARGS: |-
+              --use-factory
+            PLATFORM: linux/amd64,linux/arm64
+            TAG: ${{ github.event.inputs.tag }}
+        - name: attach-images
+          nonMakeStep: true
+          environment:
+            RELEASE: ${{ github.event.inputs.release }}
+            GH_TOKEN: ${{ github.token }}
+          command: |-
```
frezbo, line 234: we could sign this too if needed

.kres.yaml
```diff
@@ -216,6 +212,28 @@ spec:
...
+          environment:
+            CLOUD_IMAGES_EXTRA_ARGS: |-
+              --use-factory
+            PLATFORM: linux/amd64,linux/arm64
+            TAG: ${{ github.event.inputs.tag }}
+        - name: attach-images
+          nonMakeStep: true
+          environment:
+            RELEASE: ${{ github.event.inputs.release }}
+            GH_TOKEN: ${{ github.token }}
```
frezbo, line 233: is `secrets.GITHUB_TOKEN` and `github.token` the same?

shanduur: Yes.

.kres.yaml
shanduur: I'm not sure we need to sign this JSON.

PR #11876, opened 2025-09-22
feat: support setting wake-on-lan for Ethernet

### LINKED ISSUE #11866: feat(wake-on-lan)
## Feature Request

Support enabling Wake On LAN (WoL)

### Description

Some NICs reset their WoL configuration when the computer starts. This can be [reconfigured with ethtool](https://wiki.archlinux.org/title/Wake-on-LAN#Enable_WoL_on_the_network_adapter) through `ethtool -s interface wol g
`

Considering ethtool seems to be used to configure other parameters, could WoL be supported ?

smira:
Fixes #11866


internal/app/machined/pkg/controllers/cluster/node_identity_test.go
```diff
@@ -87,6 +87,10 @@ func (suite *NodeIdentitySuite) TestLoad() {
 func TestNodeIdentitySuite(t *testing.T) {
 	t.Parallel()
 
+	if os.Geteuid() != 0 {
```
smira, line 90: I added these `Skip` so that `go test ./...` is clean with non-root (on the host)

dsseng (approved):

PR #11880, opened 2025-09-24
refactor: efivars mock and tests

smira:
Refactor efivarfs and add tests.


frezbo (approved):

internal/pkg/efivarfs/efivarfs.go
```diff
@@ -87,8 +90,48 @@ func varPath(scope uuid.UUID, varName string) string {
 	return fmt.Sprintf("/sys/firmware/efi/efivars/%s-%s", varName, scope.String())
 }
 
+// ReaderWriter is an interface for reading and writing EFI variables.
+type ReaderWriter interface {
```
shanduur, line 94: nit: this should be named `ReadWriter`[^1]

[^1]: https://go.dev/doc/effective_go#interface-names

internal/pkg/efivarfs/efivarfs.go
```diff
@@ -87,8 +90,48 @@ func varPath(scope uuid.UUID, varName string) string {
...
+
+// FilesystemReaderWriter implements ReaderWriter using the efivars Linux filesystem.
+type FilesystemReaderWriter struct {
+	write bool
+}
+
+// NewFilesystemReaderWriter creates a new FilesystemReaderWriter.
+func NewFilesystemReaderWriter(write bool) (*FilesystemReaderWriter, error) {
+	if write {
+		if err := unix.Mount("efivarfs", constants.EFIVarsMountPoint, "efivarfs", unix.MS_REMOUNT, ""); err != nil {
```
shanduur, line 109: nit: this could be anonymous mount, without mount point in fs tree, if we use `xfs` and `mount/v3` package.

internal/pkg/efivarfs/efivarfs.go
frezbo: will fix in next PR

internal/pkg/efivarfs/efivarfs.go
frezbo: do we support remount with anonymous mounts? Also the library expects `/sys/fs/firmware/efivars` as an actual path

frezbo: we could refactor it to support xfs inputs, but maybe not for a backport PR

PR #11886, opened 2025-09-25
fix: default gateway as string

sergelogvinov:
# Pull Request

## What? (description)

According to the documentation, the default gateway can be specified using the string "default".

## Why? (reasoning)

https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html

```yaml
routes:
 - to: default  # could be 0.0.0.0/0 optionally
   via: 10.23.2.1
   metric: 3
```
## Acceptance


smira:
It looks like the fix was wrong, but I'm going to update it

Issue #11829, opened 2025-09-12
Unable to boot Talos on Intel NUC 8 fresh install of v1.11.1


Diaoul:
## Bug Report
👉 After booting on an USB image, wiping the disk and installing Talos v1.11.1 with secure boot, I'm unable to boot on it.
In the boot menu, the entry does not even appear.

I've then tried the exact same with v1.10.7 and everything works fine.
I suspect something is broken between the 2 and in particular the size increase to 2GB of the EFI partition, specifically for older systems.

I could reproduce on 2 distinct machines and 2 different BIOS versions, excluding any possible hardware/BIOS cause.

### Environment
- Talos version: v1.11.1
- Kubernetes version: v1.34.1
- Platform: Bare metal NUC8i5BEH

Diaoul:
It also fails to upgrade from v1.10.7 to v1.11.1 with the same symptoms of not being able to boot. I'm not sure it's the partitioning if upgrade does not resize partitions 🤷

Diaoul:
Just inspected the partitioning and it seems the size is preserved so we can exclude the partitioning to be the culprit here. Maybe something with SecureBoot then 🤔

Diaoul:
I managed to boot **without** secureboot on v1.11.1 after a wipe and it has a 2GB EFI partition. Maybe the issue is with more with secureboot then, or a combination of both.

frezbo:
Could you provide the console logs while talos is getting installed? Also there should be a UEFI entry named `Talos Linux UKI` can you find that in the UEFI bot entries?

Diaoul:
- I cannot see that in the boot menu no
- In the BIOS however, I can see many UEFI entries
- [Here](https://github.com/user-attachments/assets/254c0306-c19a-4a2f-9b9e-80c062bac370) is a video with the installation logs (scrappy solution since it does not boot afterwards..)
  - As you can see in the end it goes straight to PXE boot which is the last boot entry in the list because it did not find any other
  - No problem with v1.10.7, it reboots to Talos just fine

![Image](https://github.com/user-attachments/assets/51cee989-3245-46ea-8c19-58d58fc57678)
![Image](https://github.com/user-attachments/assets/ebbe4ac0-fd97-471d-95ae-2283da456977)

Diaoul:
For all my tests I'm using the same schematic id `664afad155864e8bdeca25cfd6078305fae06e7373251a1be7e88120eabd47ec` with this config and no kernel args:
```yaml
[...]

PR #11888, opened 2025-09-25
fix: uefi boot entry handling logic

frezbo:
Fix how we handle adding, deleting and updating UEFI boot entries and `BootOrder` EFI variables.

Fixes: #11829

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
@@ -7,6 +7,9 @@ package sdboot
 import (
 	"errors"
 	"fmt"
+	"io/fs"
+	"maps"
+	"math"
 	"os"
 	"slices"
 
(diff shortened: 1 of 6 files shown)


internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/efivars.go
```diff
@@ -93,97 +96,145 @@ func WriteVariable(name, value string) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to read BootOrder: %w", err)
+	partitionUUID := efiPartInfo[0].PartitionUUID
+
+	if partitionUUID == nil {
+		return fmt.Errorf("EFI partition UUID not found on install disk %q", blkidInfo.Name)
 	}
 
-	bootOrder, err := efivarfs.UnmarshalBootOrder(rawBootOrderData)
+	printf("using disk %s with partition %d and UUID %s", blkidInfo.Name, efiPartInfo[0].PartitionIndex, partitionUUID.String())
```
dsseng, line 121: should we use logger, or there's no easy way to get it here?

frezbo: oh this printf is the logger passed in here

shanduur (approved):

PR #11897, opened 2025-09-26
fix: provide refreshing CA pool (resolvers)

smira:
When a registry has _some_ TLS config included, the refreshing CA pool was overwritten with the result returned from the config provider.

Ensure that is is restored back to the default value (unless explicitly set by the provider if the registry CA is set).

### DIFF
--- a/internal/pkg/containers/image/resolver.go
+++ b/internal/pkg/containers/image/resolver.go
@@ -60,6 +60,11 @@ func RegistryHosts(reg config.Registries) docker.RegistryHosts {
 				if err != nil {
 					return nil, fmt.Errorf("error preparing TLS config for %q: %w", u.Host, err)
 				}
+
+				// set up refreshing Root CAs if none were provided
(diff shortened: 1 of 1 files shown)


internal/pkg/containers/image/resolver.go
```diff
@@ -60,6 +60,11 @@ func RegistryHosts(reg config.Registries) docker.RegistryHosts {
 				if err != nil {
 					return nil, fmt.Errorf("error preparing TLS config for %q: %w", u.Host, err)
 				}
+
+				// set up refreshing Root CAs if none were provided
+				if transport.TLSClientConfig.RootCAs == nil {
+					transport.TLSClientConfig.RootCAs = httpdefaults.RootCAs()
```
smira, line 66: to make it more clear: `newTransport()` sets up default pool correctly, but this flow overwrites it

smira:
There was no report for this issue, just found looking into the code.

oguzkilcan (approved):

Issue #11900, opened 2025-09-28
Broadcast added to non-broadcast addresses


chris-sanders:
Description

  When configuring a VIP on a single-node Talos cluster, the VIP is incorrectly assigned a broadcast address equal to the IP itself for /32 addresses. This causes the Linux kernel to treat the address as both unicast and broadcast, leading to sporadic routing failures where packets destined for the default gateway get incorrectly broadcast and black-holed.

  Symptoms:
  - Routes that should go through the gateway (e.g., 10.255.255.1) intermittently fail
  - `ip route get 1.1.1.1` or anything that goes through the gateway intermittently returns 'broadcast'
  - Network connectivity becomes unreliable when VIP is enabled, periodically returning Network Unreachable for all network connections.

  Example Configuration:
  - Primary interface: 10.255.255.231/24 (broadcast: 10.255.255.255)
  - VIP: 10.255.255.230/32 (broadcast incorrectly set to: 10.255.255.230)

  Root Cause:
  The broadcastAddr() function in /internal/app/machined/pkg/controllers/network/address_spec.go calculates broadcast addresses for all IPv4 prefixes, including /32 and /31:
  - For /32: netmask is 255.255.255.255, inverted becomes 0.0.0.0, resulting in broadcast = IP
  - For /31: similar issue per RFC 3021 (point-to-point links shouldn't have broadcast)

  This creates a conflict where the kernel sees the same IP as both a unicast VIP and a broadcast address, confusing routing decisions.

  Expected Behavior:
  - /32 addresses should not have a broadcast address set (host routes)
  - /31 addresses should not have a broadcast address set (RFC 3021 point-to-point)
  - This matches standard Linux behavior when adding addresses with ip addr add

  Logs

  When VIP is configured and routing fails:
  ip addr show
  [interface]: <BROADCAST,MULTICAST,UP,LOWER_UP>
      inet 10.255.255.231/24 brd 10.255.255.255 scope global
      inet 10.255.255.230/32 brd 10.255.255.230 scope global

  Routing intermittently fails to gateway
  Packets to 10.255.255.1 sometimes get broadcast instead of routed

  Environment

  - Talos version: v1.11.1
  - Kubernetes version: 1.34.0

  Additional Context

  - Manually adding the VIP without broadcast (ip addr add 10.255.255.230/32 dev eth0) works correctly
  - I have not tested a /31 but due to the non-broadcast nature of the address I expect it to be a similar issue

chris-sanders:
[...]

PR #11902, opened 2025-09-28
fix(broadcast): don't set broadcast for /31 and /32 addresses

chris-sanders:
RFC 3021 defines /31 as point-to-point links without broadcast. /32 addresses are host routes that don't need broadcast addresses.

Setting broadcast=IP for these prefixes confuses kernel routing decisions, causing packets to be incorrectly broadcast instead
of routed through the gateway.

On a single node test cluster this results in intermittent failure of all network connections.

See Issue: https://github.com/siderolabs/talos/issues/11900

TL'DR: Broadcast address is getting set to the IP address on the VIP which is causing intermittent failures on at least single node clusters. Omitting the broadcast on a `/32` works fine and would seem to be the expected configuration. This small patch applies no broadcast for either /31 or /32

- [x ] you linked an issue (if applicable)
- [x ] you included tests (if applicable)
- [x ] you ran conformance (`make conformance`)
- [x ] you formatted your code (`make fmt`)
- [x ] you linted your code (`make lint`)
- [x ] you generated documentation (`make docs`)

I was having issues getting podman to run unit-tests sorry about that.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/address_spec.go
+++ b/internal/app/machined/pkg/controllers/network/address_spec.go
@@ -20,9 +20,9 @@ import (
 	"github.com/mdlayher/arp"
 	"github.com/siderolabs/go-pointer"
 	"go.uber.org/zap"
-	"go4.org/netipx"
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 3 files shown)


smira (approved):
thank you!

PR #11931, opened 2025-10-01
fix(ci): try mode tests

frezbo:
Bump Try mode timeout so we've more time to assert.


smira (approved):
let's also run with regular `integration-qemu` to make sure it's not broken

PR #11941, opened 2025-10-01
feat: drop and lock deprecated features

smira:
These features don't make any sense right now, so remove them from the machine config to reduce clutter.


frezbo (approved):
nice 👍

PR #11942, opened 2025-10-01
feat: create merge patch from diff of two machine configs

oguzkilcan:
Added `configdiff.Patch` function to talos machinery. This function takes two machine configs in the form of `config.Provider` and returns a slice of Patches capable of converting the original config to the modified config when applied in order.


pkg/machinery/config/configdiff/mergepatch.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func setValue(mergePatch *unstructured, key string, value any, shouldSet bool) {
+	if shouldSet {
+		(*mergePatch)[key] = value
+	}
+}
+
+// nolint: gocyclo
+func matchesValue(aV, bV any) bool {
```
smira: can we use `reflect.DeepEqual` ?

smira:
we could probably add a bit of a dynamic test - generate base machine config (with the current/latest version of Talos), apply a couple of patches to it, and then verify that patch diff can be computed and applied?

PR #11943, opened 2025-10-01
fix: support secure HTTP proxy with gRPC dial

smira:
Most of the work is to add proper test environment for more cases.

Include a test for pulling an image

### DIFF
--- a/internal/app/machined/pkg/controllers/cluster/discovery_service.go
+++ b/internal/app/machined/pkg/controllers/cluster/discovery_service.go
@@ -235,6 +235,12 @@ func (ctrl *DiscoveryServiceController) Run(ctx context.Context, r controller.Ru
 				return fmt.Errorf("error initializing AES cipher: %w", err)
 			}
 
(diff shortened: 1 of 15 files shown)


internal/app/machined/pkg/controllers/cluster/discovery_service.go
```diff
@@ -235,6 +235,12 @@ func (ctrl *DiscoveryServiceController) Run(ctx context.Context, r controller.Ru
 				return fmt.Errorf("error initializing AES cipher: %w", err)
 			}
 
+			tlsConfigFunc := func() *tls.Config {
+				return &tls.Config{
+					RootCAs: httpdefaults.RootCAs(),
```
frezbo, line 240: shouldn't we use the CA reloader from crypto package?

smira: this is the CA reloader actually, we don't have one in the crypto package

frezbo (approved):

PR #11950, opened 2025-10-02
fix: minor improvements to mounts

shanduur:
- Do not create target dir for detached mounts;
- Use 'ro' flag on ReadOnly mount requests;

### DIFF
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -495,6 +495,7 @@ func (ctrl *MountController) handleDiskMountOperation(
 
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
+			fsOpts = append(fsOpts, fsopen.WithBoolParameter("ro"))
 		}
 
 		if mountRequest.TypedSpec().Detached {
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -495,6 +495,7 @@ func (ctrl *MountController) handleDiskMountOperation(
 
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
```
frezbo, line 497: this not needed then?

frezbo (approved):

frezbo:
added provision-1 to see if this is solved https://github.com/siderolabs/talos/actions/runs/18183453005/job/51763518772

internal/app/machined/pkg/controllers/block/mount.go
smira: my understanding is that opening fs as ro mounting as read-only are different concepts

Mateusz, what I meant is that can we in mount package if I say `mount.ReadOnly` enforce fsopen with `ro`?

shanduur: Separately one over another? Unfortunately no, I kept that deliberately to not pass `ro` (or any `fsopen` specific flags) to the Mount manager itself, as I want to keep this separate to allow other implementations - e.g. `opentree`.

frezbo: > my understanding is that opening fs as ro mounting as read-only are different concepts
> 
> Mateusz, what I meant is that can we in mount package if I say `mount.ReadOnly` enforce fsopen with `ro`?

ahh I get it now, we need both I guess

smira (approved):

PR #11959, opened 2025-10-03
fix: provide nocloud metadata with missing network config

smira:
Even if network config is missing, provide metadata always.

This way `PlatformMetadata` resource is properly populated, even if it just contains the platform information.

See https://github.com/siderolabs/omni/discussions/1633#discussioncomment-14577048


Orzelius (approved):

smira:
I got a confirmation that the fix works.

PR #11962, opened 2025-10-03
feat: add static registry to talosctl

### LINKED ISSUE #11928: Create a repository of images from the list
- download each image from the list (all arches), place it to the filesystem for later serving (borrow code from image-cache create)
   - optionally, verify existing signature in a connected environment, and re-sign with a private key (to support offline signature verification)

### LINKED ISSUE #11929: Serve a static read-only image registry
Serving a repository: `talosctl image repository serve --path=/var/images` which acts as a read-only registry
 
(optional) support containerd `?ns=ghcr.io` to servie multiple registry endpoints from a single repository endpoint in a safe way

shanduur:
`talosctl` includes two new subcommands `talosctl image registry create` and `talosctlimage registry serve`.

* `talosctl image registry create` allows creating a local OCI image registry containing Talos images for speeding up installations, upgrades, and deployments in air-gapped environments.

* `talosctl image registry serve` allows serving the created OCI image registry over HTTP. It is a read-only registry, meaning images cannot be pushed to it, but the backing storage can be updated by re-running the `create` command.

Fixes #11928
Fixes #11929

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -7,10 +7,15 @@ package talos
 import (
 	"bufio"
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
+	"net"
+	"net/http"
 	"os"
+	"os/signal"
 	"slices"
 	"strings"
 	"text/tabwriter"
@@ -18,8 +23,15 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/dustin/go-humanize"
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/olareg/olareg"
+	"github.com/olareg/olareg/config"
+	"github.com/siderolabs/go-pointer"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/artifacts"
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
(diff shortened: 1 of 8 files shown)


cmd/talosctl/pkg/talos/artifacts/copy.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/siderolabs/go-retry/retry"
+)
+
+// Mirror mirrors the given images to the destination registry.
+func Mirror(ctx context.Context, options []crane.Option, images []string, destRegistry string) error {
+	// TODO:
```
frezbo, line 21: TODO needs to be done for future?

shanduur: I want to copy signatures in this PR, rest will be in the following RPs.

shanduur: Done, signatures are now pulled with images wherever they are present.

cmd/talosctl/cmd/talos/image.go
```diff
@@ -394,16 +418,202 @@ var imageCacheCreateCmdFlags struct {
...
-	provisioner pflag.Value
-}{
-	provisioner: helpers.StringChoice(provisionerInstaller, provisionerDocker, provisionerAll),
+		transport := remote.DefaultTransport.(*http.Transport).Clone()
+		transport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true, //nolint: gosec
+		}
+
+		options = append(options,
+			crane.WithJobs(0),
```
frezbo, line 434: crane cli usually works with `--insecure` for localhost, not sure if this is really needed

shanduur: That's exactly what crane `--insecure` flag is doing under the hood - this parameter is added when you call crane with it.

cmd/talosctl/pkg/talos/artifacts/copy.go
```diff
@@ -0,0 +1,222 @@
...
+
+	if err := r.RetryWithContext(
+		ctx, func(ctx context.Context) error {
+			options = append(options,
+				crane.WithContext(ctx),
+			)
+
+			var err error
+
+			digest, err = crane.Digest(
```
smira: I think we should find a signature for the exactly same digest as we mirror, to close the gap between resolving a tag and fetching a signature

shanduur: I think I've fixed it to what you've meant.

smira:
As another question more higher level:

* we already have image pull process for image cache
* we already have `registryd` which can serve such image cache as a registry

Do we need to pull another/incompatible way with `olareg`? Can we unify both paths? 

With image cache e.g. we already solved the problem of running on non-Linux and being able to handle paths with something like `foo.bar:5000`

shanduur:
`registryd` is not OCI compatible and I'm not sure how to test it (even manually) - I can't just `crane pull` from it - we need a custom client for tests. This might cause friction for users in the future. Additionally `registryd` is not able to handle multi-arch manifests as of now, which are required for image factory (unless we don't want to distribute image factory using the cache). Olareg is just compliant plug-and-play option.

shanduur:
Alternative approach using image cache and registryd: https://github.com/siderolabs/talos/pull/12014

smira:
> `registryd` is not OCI compatible and I'm not sure how to test it (even manually) - I can't just `crane pull` from it - we need a custom client for tests. This might cause friction for users in the future. Additionally `registryd` is not able to handle multi-arch manifests as of now, which are required for image factory (unless we don't want to distribute image factory using the cache). Olareg is just compliant plug-and-play option.

I think `registryd` is dumb enough that it doesn't care what it serves :) So it should serve multi-arch as long as the directory contains multi-arch.

If `crane` fails to pull from it, we could try to fix it? 

But my point is not to have two implementations, but rather just one (if we can).

PR #11964, opened 2025-10-03
fix: make Akamai platform usable

Xaerxess:
Akamai images were unusable (see siderolabs/docs#92), so this PR addresses issues with Akamai by:

- clarifying that Direct Disk must be used to avoid kernel panic
- handling errors better while fetching metadata
- clarifying documentation

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/akamai/akamai.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/akamai/akamai.go
@@ -17,6 +17,7 @@ import (
 	"github.com/siderolabs/go-procfs/procfs"
 
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/errors"
 	"github.com/siderolabs/talos/internal/app/machined/pkg/runtime/v1alpha1/platform/internal/netutils"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"github.com/siderolabs/talos/pkg/machinery/imager/quirks"
@@ -159,7 +160,11 @@ func (a *Akamai) Configuration(ctx context.Context, r state.State) ([]byte, erro
 		return nil, fmt.Errorf("get user data: %w", err)
 	}
 
-	return []byte(userData), err
+	if userData == "" {
+		return nil, errors.ErrNoConfigSource
+	}
+
+	return []byte(userData), nil
 }
 
 // Mode implements the platform.Platform interface.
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/akamai/akamai.go
```diff
@@ -205,3 +224,75 @@ func (a *Akamai) NetworkConfiguration(ctx context.Context, _ state.State, ch cha
...
+	} else {
+		// SystemInformation not available yet, assume we need override
+		shouldOverride = true
+		currentUUID = "not-available-yet"
+	}
+
+	if shouldOverride && linodeID > 0 {
+		generatedUUID := generateLinodeUUID(linodeID)
+
+		uuidKey := runtimeres.NewMetaKey(runtimeres.NamespaceName, runtimeres.MetaKeyTagToID(meta.UUIDOverride))
```
smira: my only concern here is that this somewhat breaks the access flow - this resource represents the state of `META`, but creating this resource doesn't write to META.

so I wonder what is the proper answer here, and I'm not sure if we should do it this way, or instead let the platform publish an UUID override which will be picked up and published in SystemInformation

Xaerxess: @smira As discussed on [Slack](https://taloscommunity.slack.com/archives/C04D4PDAJT0/p1759404072816299?thread_ts=1759329780.355509&cid=C04D4PDAJT0), if Omni resolves UUID conflicts properly, this could even be removed or separated info different PR, since the Talos-only functionality for Akamai image should be fixed by null-check in lines 166-170. 

All-zeros UUID were causing issues with both self-hosted and cloud Omni, but with a "strict" `--join-tokens-mode` rollout, it should be better.

Regarding your note — what's the best way to publish the override like you described? I don't have much experience in Go and even less with Talos architecture/codebased.

smira: I don't have a ready answer here, but I think it should publish this to platform metadata, and this should later be picked up by other paths, but we can discuss this internally. 

Strict join tokens are now on by default for new Omni accounts, and on-prem you can enable it as you wish.

Xaerxess:
@smira I removed the UUID resetting part from this PR. It now aims to make Akamai images usable. I'd appreciate any feedback (is new PR required?), thanks!

Xaerxess:
I think it makes sense to release it in 1.11.x patch release, should I also make similar changes to v1.11 docs?

smira:
@Xaerxess sorry, as we were on this PR, Talos docs got moved to a different repo. I will merge this PR as it is, and schedule a backport for 1.11.x.

@Iheanacho-ai should take care of the documentation.

PR #11965, opened 2025-10-03
feat: implement link aliasing

### LINKED ISSUE #10956: Network Config: `LinkAliasConfig`
### Link Aliasing

Link aliasing is optional, Talos provides predictable interface names by default, and these can be used directly.

There are two main reasons for link aliasing:

- interface name is not known, but some properties are known, e.g. the MAC address or the driver
- renaming links to have logical names, e.g. `int0` and `ext0` to distinguish private/public links

As soon as the link was aliased, its alias can be used interchangeably with the primary name in any of the configuration documents.

Only a single link should match for the alias to be in place.

```yaml
version: v1alpha1
kind: LinkAliasConfig
name: int0 # alias name
match:
  selector: link.hardware_addr.startswith("01:02:03:ee:") && link.physical
```

- `name` specific the alias to be created
- `match.selector` specifies a CEL expression (similar to disk match) to select a link, the details of the expression are TBD at the moment; selector only matches physical links, i.e. it never matches VLANs, bonds, etc.

> [!NOTE]  
> Link aliasing should get better support in general, e.g. showing link aliases in `get links`, `get addresses`, etc.
It should be nice, easy and intuitive to use link aliases.

smira:
Fixes #10956


tools/structprotogen/proto/proto.go
```diff
@@ -468,15 +468,13 @@ func getProtoBasicName(typ string) string {
 	switch typ {
 	case "bool":
 		return "bool"
-	case "int8", "int16":
-		return "fixed32"
-	case "int32":
+	case "int8", "int16", "int32":
```
smira, line 474: note: this is a big issue in `.proto` generation from resources

but this way it matches what `protoenc` does

smira: fix in https://github.com/siderolabs/talos/pull/11979

PR #11977, opened 2025-10-06
fix: add dashboard spinner

smira:
This is handy to detect frozen machines when looking at the dashboard output video console.

[asciicast](https://asciinema.org/a/ta8PAxXSYVAbtViRwQ3Yz864a)

### DIFF
--- a/internal/pkg/dashboard/components/header.go
+++ b/internal/pkg/dashboard/components/header.go
@@ -38,6 +38,7 @@ type Header struct {
 
 	selectedNode string
 	nodeMap      map[string]*headerData
+	spinnerPos   int
 }
 
 // NewHeader initializes Header.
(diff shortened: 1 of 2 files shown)


frezbo (approved):

internal/pkg/dashboard/components/header.go
```diff
@@ -90,6 +91,13 @@ func (widget *Header) OnAPIDataChange(node string, data *apidata.Data) {
 	}
 }
 
+// OnTick implements the TickerListener interface.
+func (widget *Header) OnTick() {
```
dsseng, line 95: If this runs in talosctl client, and the server freezes, it'll perhaps keep spinning, no? Is this feature only designated for VT console on the node itself?

smira: it is to show that `talosctl` is running itself - if we want to handle missing data, it should be handled other ways

dsseng (approved):

PR #11979, opened 2025-10-06
fix: resource proto definitions

smira:
There was a mismatch between how Talos encodes resource specs to protobuf on the wire, and `.proto` files generated by `structprotogen` for the same specs.

This only affected `LinkStatus`/`LinkSpec` resources, and it doesn't affect Talos/Omni as they go via `protoenc`, bypassing `.proto` files.

Also if default proto library is used, it also works correctly.

The only place where it fails is a strict `vtprotobuf` encoding.

Also imports `protoenc` 0.2.4 skipping encoding of zero fields, which should make COSI responses smaller for resources with lots of zero fields.


api/lock.binpb
frezbo, line 1: is this new?

frezbo (approved):

api/lock.binpb
smira: This is protobuf check for incompatible changes, but this is an incompatible change, so I have to "reset" it

PR #11981, opened 2025-10-07
chore(ci): run only nvidia tests for NVIDIA workflows

frezbo:
Run only NVIDIA related tests for NVIDIA workflows.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -2735,6 +2735,7 @@ spec:
           command: e2e-aws
           environment:
             EXTRA_TEST_ARGS: "-talos.extensions.nvidia -talos.verifyukibooted=false"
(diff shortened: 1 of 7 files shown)


.kres.yaml
```diff
@@ -2734,7 +2734,7 @@ spec:
         - name: e2e-aws-nvidia-oss-lts
           command: e2e-aws
           environment:
-            EXTRA_TEST_ARGS: "-talos.extensions.nvidia -talos.verifyukibooted=false"
+            EXTRA_TEST_ARGS: "-talos.extensions.nvidia -talos.verifyukibooted=false -test.run=TestIntegration/api.ExtensionsSuiteNVIDIA"
```
smira: I thought we always send `-test.run` which comes from the environment variable?

frezbo: yeh, this is wrong

frezbo: fixed

PR #11985, opened 2025-10-08
docs: add redirects to the new site in the netlify.toml file

Iheanacho-ai:
add redirects for the new site


rothgar:
This will be opened against the new talos-website netlify project instead of here. Keeping the old talos.dev site functioning as is.

PR #12000, opened 2025-10-10
feat: add --airgapped flag to talosctl cluster create

shanduur:
Add new `--airgapped` flag to talos cluster create (qemu)
to disable NAT in the VMs to effectively become airgapped.

---

1. Generating minimal image cache can be done using following chain of commands:
    ```bash
    talosctl images default |\
      talosctl images cache-create --images=- --image-cache-path=/tmp/cache --layout=flat
    ```

1. Additionally, it needs a set of certificates. You can generate them (and required patch) using the following command:
    ```bash
    talosctl image cache-cert-gen --advertise-address=172.20.0.1 \
      --tls-ca-file=/tmp/ca.crt \
      --tls-cert-file=/tmp/tls.crt \
      --tls-key-file=/tmp/tls.key
    ```

1. After that you can start a cluster using the following command. There is no need to launch the image-cache manually, it's handled once `--image-cache-path` is passed. The directory must exist, and must contain image cache with flat layout.
    ```bash
    sudo --preserve-env=HOME talosctl cluster create \
      --with-cluster-discovery=false --provisioner=qemu \
      --cidr=172.20.0.0/24 \
      --install-image="172.20.0.1:12000/siderolabs/installer:${TALOS_VERSION}" \
      --controlplanes 3 \
      --workers 2 \
      --with-bootloader=false \
      --disk-preallocate=false \
      --with-cluster-discovery=false \
      --airgapped \
      --config-patch='@hack/test/patches/airgapped-timesync.yaml' \
      --config-patch='@image-cache-patch.yaml' \
      --image-cache-path=/tmp/cache \
      --image-cache-tls-cert-file=/tmp/tls.crt \
      --image-cache-tls-key-file=/tmp/tls.key \
      --image-cache-port=12000 \
      --registry-mirror=docker.io=https://172.20.0.1:12000 \
      --registry-mirror=registry.k8s.io=https://172.20.0.1:12000 \
      --registry-mirror=gcr.io=https://172.20.0.1:12000 \
      --registry-mirror=ghcr.io=https://172.20.0.1:12000
    ```

---

Patches explained:

1. `image-cache-patch.yaml` - Talos requires HTTPS to pull installer image. In airgapped setups, images are hosted in an internal OCI registry using a self-signed or private TLS certificate. Because Talos does not trust this certificate by default, it will fail with `x509: certificate signed by unknown authority`. This patch adds the certificate of private registry to Talos trusted CAs so it can securely pull images.
    ```yaml
    apiVersion: v1alpha1
    kind: TrustedRootsConfig
    name: image-cache-ca
    certificates: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
    ```

3. `airgapped-timesync.yaml` - Talos uses time.cloudflare.com by default for NTP. In airgapped environments there is no internet access, which means NTP sync fails and time drifts. That breaks TLS validation and can block Talos bootstrap. This patch replaces the public NTP server with a local time source like `/dev/ptp0` (bare metal/QEMU) or an internal NTP server.
    ```yaml
    machine:
      time:
        servers:
          - /dev/ptp0 # or internal NTP server
    ```

4. `registry-mirror` - Talos and Kubernetes components normally pull images from public registries like `docker.io`, `ghcr.io`, and `registry.k8s.io`. These are unreachable in airgapped environments. This patch tells Talos to redirect all image pulls to local registry mirror at `https://172.20.0.1:12000`. This is the same registry as the one hosting installer image, thus we need to specify HTTPS and not HTTP as it's often showed in Talos docs.
    ```yaml
    machine:
      registries:
        mirrors:
          docker.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          gcr.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          ghcr.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
          registry.k8s.io:
            endpoints:
              - https://172.20.0.1:12000 # note that it is HTTPS not HTTP
    ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -482,10 +482,15 @@ uki-certs: talosctl ## Generate test certificates for SecureBoot/PCR Signing
 	@$(TALOSCTL_EXECUTABLE) gen secureboot pcr
 	@$(TALOSCTL_EXECUTABLE) gen secureboot database
 
-.PHONY: cache-create
-cache-create: installer imager ## Generate image cache.
+.PHONY: integration-images-list
+integration-images-list: ## Generate list of integration images.
 	@docker run --entrypoint /usr/local/bin/e2e.test registry.k8s.io/conformance:$(KUBECTL_VERSION) --list-images | \
-		$(TALOSCTL_EXECUTABLE) images integration --installer-tag=$(IMAGE_TAG_IN) --registry-and-user=$(REGISTRY_AND_USERNAME) | \
+		$(TALOSCTL_EXECUTABLE) images integration --installer-tag=$(IMAGE_TAG_IN) --registry-and-user=$(REGISTRY_AND_USERNAME) \
+		> $(ARTIFACTS)/integration-images.txt
+
+.PHONY: cache-create
+cache-create: installer imager integration-images-list ## Generate image cache.
+	@cat $(ARTIFACTS)/integration-images.txt | \
 		$(TALOSCTL_EXECUTABLE) images cache-create --image-cache-path=/tmp/cache.tar --images=- --force
 	@crane push /tmp/cache.tar $(REGISTRY_AND_USERNAME)/image-cache:$(IMAGE_TAG_OUT)
 	@$(MAKE) image-iso IMAGER_ARGS="--image-cache=$(REGISTRY_AND_USERNAME)/image-cache:$(IMAGE_TAG_OUT) --extra-kernel-arg='console=ttyS0'"
--- a/cmd/talosctl/cmd/mgmt/cluster/create/cmd_dev.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/cmd_dev.go
@@ -98,6 +98,11 @@ func getCreateCmd() *cobra.Command {
 		withUUIDHostnamesFlag         = "with-uuid-hostnames"
 		withSiderolinkAgentFlag       = "with-siderolink"
 		configInjectionMethodFlag     = "config-injection-method"
+		airgappedFlag                 = "airgapped"
+		imageCachePathFlag            = "image-cache-path"
+		imageCacheTLSCertFileFlag     = "image-cache-tls-cert-file"
+		imageCacheTLSKeyFileFlag      = "image-cache-tls-key-file"
+		imageCachePortFlag            = "image-cache-port"
 
 		// The following flags are the gen options - the options that are only used in machine configuration (i.e., not during the qemu/docker provisioning).
 		// They are not applicable when no machine configuration is generated, hence mutually exclusive with the --input-dir flag.
@@ -131,6 +136,7 @@ func getCreateCmd() *cobra.Command {
 		packetReorderFlag,
 		packetCorruptFlag,
 		bandwidthFlag,
+		airgappedFlag,
 
 		// The following might work but need testing first.
 		configInjectionMethodFlag,
@@ -209,7 +215,7 @@ func getCreateCmd() *cobra.Command {
 		qemu.MarkHidden("with-debug-shell") //nolint:errcheck
 		qemu.StringSliceVar(&qOps.ExtraUEFISearchPaths, extraUEFISearchPathsFlag, qOps.ExtraUEFISearchPaths, "additional search paths for UEFI firmware (only applies when UEFI is enabled)")
 		qemu.StringSliceVar(&qOps.NetworkNoMasqueradeCIDRs, networkNoMasqueradeCIDRsFlag, qOps.NetworkNoMasqueradeCIDRs, "list of CIDRs to exclude from NAT")
-		qemu.StringSliceVar(&qOps.Nameservers, nameserversFlag, qOps.Nameservers, "list of nameservers to use")
+		qemu.StringSliceVar(&qOps.Nameservers, nameserversFlag, qOps.Nameservers, "list of nameservers to use, by default use embedded DNS forwarder")
 		qemu.UintVar(&qOps.DiskBlockSize, diskBlockSizeFlag, qOps.DiskBlockSize, "disk block size")
 		qemu.StringVar(&qOps.TargetArch, targetArchFlag, qOps.TargetArch, "cluster architecture")
 		qemu.StringSliceVar(&qOps.CniBinPath, cniBinPathFlag, qOps.CniBinPath, "search path for CNI binaries")
@@ -239,6 +245,11 @@ func getCreateCmd() *cobra.Command {
 			"enables the use of siderolink agent as configuration apply mechanism. `true` or `wireguard` enables the agent, `tunnel` enables the agent with grpc tunneling")
 		qemu.StringVar(&qOps.ConfigInjectionMethod,
 			configInjectionMethodFlag, qOps.ConfigInjectionMethod, "a method to inject machine config: default is HTTP server, 'metal-iso' to mount an ISO")
+		qemu.BoolVar(&qOps.Airgapped, airgappedFlag, qOps.Airgapped, "limit VM network access to the provisioning network only")
+		qemu.StringVar(&qOps.ImageCachePath, imageCachePathFlag, qOps.ImageCachePath, "path to image cache")
+		qemu.StringVar(&qOps.ImageCacheTLSCertFile, imageCacheTLSCertFileFlag, qOps.ImageCacheTLSCertFile, "path to image cache TLS cert")
+		qemu.StringVar(&qOps.ImageCacheTLSKeyFile, imageCacheTLSKeyFileFlag, qOps.ImageCacheTLSKeyFile, "path to image cache TLS key")
+		qemu.Uint16Var(&qOps.ImageCachePort, imageCachePortFlag, qOps.ImageCachePort, "port on which to serve image cache")
 
 		return qemu
 	}
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -10,6 +10,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
+	"net"
 	"os"
 	"os/signal"
 	"slices"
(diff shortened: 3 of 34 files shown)


Orzelius:
I'll take a look tomorrow

cmd/talosctl/cmd/mgmt/cluster/create/cmd_dev.go
```diff
@@ -98,6 +98,8 @@ func getCreateCmd() *cobra.Command {
 		withUUIDHostnamesFlag         = "with-uuid-hostnames"
 		withSiderolinkAgentFlag       = "with-siderolink"
 		configInjectionMethodFlag     = "config-injection-method"
+		airgappedFlag                 = "airgapped"
+		airgappedConnectedBridgesFlag = "airgapped-connected-bridges"
```
Orzelius: what do you think about `airgapped-connect-bridges` instead? It's a bit shorter and imperative

pkg/provision/providers/vm/dnsd.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// DestroyDNSd destoys load balancer.
+func (p *Provisioner) DestroyDNSd(state *State) error {
+	pidPath := state.GetRelativePath(dnsPid)
+
+	return StopProcessByPidfile(pidPath)
+}
+
+func dumbHandler(config *dns.ClientConfig) func(w dns.ResponseWriter, r *dns.Msg) {
```
Orzelius: typo or is it literally **dumb**Handler?

Orzelius (commented):
What about updating the airgapped integration tests as well? 

On macos I get `listen udp 10.5.0.1:53: bind: can't assign requested address`, but the dns still works. I think there's a dns proxy to the host by default so that's why it keeps working. I couldn't find anything actually bound to the port so I can't tell why the permission errors occur without looking into it further.

pkg/provision/providers/vm/dnsd.go
shanduur: I might need a better name here. `simpleHandler`? `forwardHandler`?

shanduur:
I'm not sure if it will even work on MacOS. It will require stopping `mDNSResponder`, which according [to this Apple discussion](https://discussions.apple.com/thread/8313713) is a required system component.

This might also explain why are you still able to resolve.

pkg/provision/providers/vm/dnsd.go
Orzelius: Ah I see, I immediately assumed it to be a type of dump and got a bit confused

hack/test/e2e.sh
```diff
@@ -141,7 +141,31 @@ function dump_cluster_state {
...
 function build_registry_mirrors {
+  if [[ "${WITH_AIRGAPPED:-false}" != "false" ]]; then
+    build_image_cache
+    serve_image_cache
+
+    REGISTRY_MIRROR_FLAGS=()
+
+    for registry in docker.io registry.k8s.io quay.io gcr.io ghcr.io; do
+      addr="???" # This should be Gateway IP?
+      REGISTRY_MIRROR_FLAGS+=("--registry-mirror=${registry}=http://${addr}:5000")
```
shanduur: Not sure how to get Gateway IP here.

hack/test/e2e.sh
```diff
@@ -141,7 +141,31 @@ function dump_cluster_state {
...
 }
 
+function build_image_cache {
+  "${TALOSCTL}" images default | \
+    "${TALOSCTL}" image cache-create --images=- --image-cache-path="${TMP}/image-cache" --layout=flat
+}
+
+function serve_image_cache {
+  "${TALOSCTL}" image cache-server --address=:5000 --image-cache-path="${TMP}/image-cache" &
+  # TODO: handle logs
```
shanduur: `2>&1 > "${TMP}/image-cache.log"` ? We _probably_ want it saved somewhere?

hack/test/e2e.sh
```diff
@@ -141,7 +141,31 @@ function dump_cluster_state {
   ${KUBECTL} get pods --all-namespaces -o wide
 }
 
+function build_image_cache {
+  "${TALOSCTL}" images default | \
+    "${TALOSCTL}" image cache-create --images=- --image-cache-path="${TMP}/image-cache" --layout=flat
+}
+
+function serve_image_cache {
```
shanduur: I'm not sure if this should be started using `image cache-server` directly, or thorugh a talosctl flag, e.g. `--airgapped-cache-path`, that (if not empty) will launch the registry?

shanduur: Moved it to the Airgapped code as of now.

hack/test/e2e.sh
shanduur: Hardcoded `172.20.2.1`, similarly as we do with other IPs.

pkg/provision/providers/vm/dnsd.go
```diff
@@ -0,0 +1,89 @@
...
+		c.Net = "udp"
+
+		var (
+			resp *dns.Msg
+			err  error
+		)
+
+		for _, serverAddr := range config.Servers {
+			resp, _, err = c.Exchange(r, net.JoinHostPort(serverAddr, config.Port))
+			if err == nil {
```
smira: on `NXDOMAIN` we should stop here

pkg/provision/providers/vm/image_cache.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	defer logFile.Close() //nolint:errcheck
+
+	gatewayAddrs := xslices.Map(clusterReq.Network.GatewayAddrs, netip.Addr.String)
+	gatewayAddrs = xslices.Map(gatewayAddrs, func(s string) string { return net.JoinHostPort(s, "5000") })
+
+	args := []string{
+		"image", "cache-serve",
+		"--addr", gatewayAddrs[0],
```
shanduur: Should I add support for multiple addresses to cache-serve?

pkg/provision/providers/vm/dnsd.go
shanduur: Fixed

pkg/provision/providers/vm/image_cache.go
Orzelius: I think it's good as-is. If we ever need/want to test dual-stack ipv4 and ipv6 setup we might need to expand this.

shanduur:
Back to draft, failing conformance 😮‍💨

cmd/talosctl/cmd/talos/image.go
```diff
@@ -460,6 +468,68 @@ var imageCacheServeCmdFlags struct {
 	tlsKeyFile     string
 }
 
+// imageCacheCertGenCmd represents the image cache tls certificate generation command.
+var imageCacheCertGenCmd = &cobra.Command{
+	Use:     "cache-cert-gen",
+	Short:   "Generate TLS certificates and CA patch required for securing image cache to Talos communication",
+	Long:    `Generate TLS certificates and CA patch required for securing image cache to Talos communication`,
+	Example: ``,
```
Orzelius, line 477: I think you can just omit `Long` and `Example` here

cmd/talosctl/cmd/talos/image.go
```diff
@@ -460,6 +468,68 @@ var imageCacheServeCmdFlags struct {
...
+		if err := os.WriteFile(imageCacheCertGenCmdFlags.tlsKeyFile, keyPEM, 0o600); err != nil {
+			return err
+		}
+
+		return nil
+	},
+}
+
+func generateConfigPatch(caPEM []byte) error {
+	patch2 := security.NewTrustedRootsConfigV1Alpha1()
```
Orzelius: Probably doesn't need to be named patch2, right?

cmd/talosctl/cmd/talos/image.go
shanduur: Technically yes, although I'm following the pattern of other commands.

cmd/talosctl/cmd/talos/image.go
shanduur: Fixed

pkg/provision/providers/vm/dnsd.go
```diff
@@ -0,0 +1,113 @@
...
+			resp *dns.Msg
+			err  error
+		)
+
+		for _, serverAddr := range config.Servers {
+			log.Debug("making DNS request", "target", serverAddr+":"+config.Port)
+
+			resp, _, err = c.Exchange(r, net.JoinHostPort(serverAddr, config.Port))
+			if err == nil {
+				log.Debug("DNS request failed", "error", err)
```
smira, line 96: looks like the log line doesn't match the if condition

smira: of the `if` condition should be `err != nil`, otherwise `if resp` below doesn't make sense

Makefile
```diff
@@ -482,6 +482,12 @@ uki-certs: talosctl ## Generate test certificates for SecureBoot/PCR Signing
...
 
+.PHONY: integration-images-list
+integration-images-list: ## Generate list of integration images.
+	@docker run --entrypoint /usr/local/bin/e2e.test registry.k8s.io/conformance:$(KUBECTL_VERSION) --list-images | \
+		$(TALOSCTL_EXECUTABLE) images integration --installer-tag=$(IMAGE_TAG_IN) --registry-and-user=$(REGISTRY_AND_USERNAME) \
+		> $(ARTIFACTS)/integration-images.txt
+
 .PHONY: cache-create
 cache-create: installer imager ## Generate image cache.
 	@docker run --entrypoint /usr/local/bin/e2e.test registry.k8s.io/conformance:$(KUBECTL_VERSION) --list-images | \
```
smira: should we re-use the target above here?

pkg/provision/providers/vm/dnsd.go
shanduur: Fixed

Issue #11989, opened 2025-10-08
Set `module.sig_enforce` in talos images


frezbo:
### Description

As per https://github.com/siderolabs/talos/issues/11675#issuecomment-3382138668 we'll be disabling `CONFIG_MODULE_SIG_FORCE`.

See https://docs.kernel.org/admin-guide/kernel-parameters.html

```text
module.sig_enforce
                        [KNL] When CONFIG_MODULE_SIG is set, this means that
                        modules without (valid) signatures will fail to load.
                        Note that if CONFIG_MODULE_SIG_FORCE is set, that
                        is always true, so this option does nothing.
```

Also revert https://github.com/siderolabs/pkgs/pull/1332

frezbo:
Pkgs reverted https://github.com/siderolabs/pkgs/pull/1346

PR #12001, opened 2025-10-10
feat: support optionally disabling module sig verification

frezbo:
Support disabling kernel module signature verification. Note that this does not work when SecureBoot is enabled.

Fixes: #11989

### DIFF
--- a/internal/integration/api/security.go
+++ b/internal/integration/api/security.go
@@ -1,0 +1,66 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_api
+
+package api
+
+import (
+	"context"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/resource"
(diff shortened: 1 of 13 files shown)


smira:
Oh, and more important thing: let's add current status to `SecurityState` and assert that it is enforced in all our tests.

frezbo:
> Oh, and more important thing: let's add current status to `SecurityState` and assert that it is enforced in all our tests.

yes, I also need to update some docs in developing talos

internal/integration/api/security.go
```diff
@@ -0,0 +1,66 @@
...
+func (suite *SecuritySuite) SuiteName() string {
+	return "api.SecuritySuite"
+}
+
+// SetupTest sets up the test.
+func (suite *SecuritySuite) SetupTest() {
+	// make sure API calls have timeout
+	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 1*time.Minute)
+
+	if suite.Cluster == nil || suite.Cluster.Provisioner() != base.ProvisionerQEMU {
```
smira, line 40: why do we skip it e.g. for AWS? it should have this enforced as well?

frezbo: yes, should be skipped for docker provisioner only

PR #12002, opened 2025-10-10
feat: update uefi image to talos linux logo

rothgar:
Updated to the latest logo.

Image size changed slightly from 369x400 to 400x400 but I can crop if we need it to be the same.


smira:
https://github.com/user-attachments/assets/8b189fd7-7e8b-472b-9f21-5f18f99a0fbd

PR #12009, opened 2025-10-13
talosctl/cluster: use `/etc/resolv.conf` nameservers during cluster creation (for QEMU)

laurazard:
# Pull Request

## What? (description)

This commit adds logic to read and parse configuration from `/etc/resolv.conf`. The parsing logic is adapted from the [Go stdlib](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/dnsconfig_unix.go;l=18).

The current implementation does stream the file (I was thinking that in case something is really wrong and /etc/resolv.conf is way bigger than it should be, this would be better) but it might be better to keep it simple and just read the file and then pass it along for parsing (plus, if something is that wrong with your `/etc/resolv.conf`, you probably have bigger issues 😅).

If anything fails w/ reading/parsing, the previously configured defaults are used.

## Why? (reasoning)

`talosctl cluster create` uses the Google and CloudFlare nameservers as defaults for the QEMU config. This is generally fine, but can cause issues if these are unreachable from the user's network (see https://github.com/siderolabs/docs/issues/94), and it would generally be nicer to attempt to use the configured nameservers.

## Acceptance

## Bonus doc picture

Screenshot 2025-10-13 at 15 15 37

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
@@ -110,6 +110,7 @@ type Qemu struct {
 	ExtraUEFISearchPaths      []string
 	NetworkNoMasqueradeCIDRs  []string
 	Nameservers               []string
+	AddHostNameservers        bool
 	Disks                     flags.Disks
 	DiskBlockSize             uint
 	PreallocateDisks          bool
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu.go
@@ -5,11 +5,13 @@
 package makers
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
 	"net/netip"
 	"net/url"
+	"os"
 	"slices"
 	"strings"
 
@@ -215,6 +217,13 @@ func (m *Qemu) ModifyClusterRequest() error {
 	nameserverIPs, err := getNameserverIPs(m.EOps.Nameservers)
 	if err != nil {
 		return err
+	}
+
+	if m.EOps.AddHostNameservers {
+		hostNameserverIPs, err := ParseHostDNSConfig("/etc/resolv.conf")
+		if err == nil {
+			nameserverIPs = append(hostNameserverIPs, nameserverIPs...)
+		}
 	}
 
 	noMasqueradeCIDRs := make([]netip.Prefix, 0, len(m.EOps.NetworkNoMasqueradeCIDRs))
@@ -665,4 +674,45 @@ func getNameserverIPs(nameservers []string) ([]netip.Addr, error) {
 	}
 
 	return nameserverIPs, nil
+}
+
+const resolvNameserverKey = "nameserver"
+
+// ParseHostDNSConfig parses the system's configured DNS resolvers.
+// It is adapted from the stdlib:
+// https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/dnsconfig_unix.go;l=18
+func ParseHostDNSConfig(path string) ([]netip.Addr, error) { //nolint:gocyclo
+	var dnsServers []netip.Addr
+
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("read %s: %w", path, err)
+	}
+	defer file.Close() //nolint:errcheck
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		if len(dnsServers) > 2 { // small, but the standard limit
+			break
+		}
+
+		line := scanner.Text()
+		if len(line) > 0 && (line[0] == ';' || line[0] == '#') {
+			// comment
+			continue
+		}
+
+		fields := strings.Fields(line)
+		// len(fields) > 1 because we need a value too
+		if len(fields) > 1 && fields[0] == resolvNameserverKey {
+			// make sure server name is an IP address (otherwise we need DNS to
+			// look it up) and not a loopback address (like systemd-resolved,
+			// which will be unreachable from the VM)
+			if addr, err := netip.ParseAddr(fields[1]); err == nil && !addr.IsLoopback() {
+				dnsServers = append(dnsServers, addr)
+			}
+		}
+	}
+
+	return dnsServers, nil
 }
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu_test.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu_test.go
@@ -5,8 +5,11 @@
 package makers_test
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/siderolabs/talos/cmd/talosctl/cmd/mgmt/cluster/create/clusterops"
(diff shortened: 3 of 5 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options_test.go
```diff
@@ -0,0 +1,115 @@
...
+nameserver 10.10.1.4
+nameserver 10.10.1.5
+nameserver 10.10.1.6
+`,
+			expected: []string{
+				"10.10.1.2:53",
+				"10.10.1.3:53",
+				"10.10.1.4:53",
+			},
+		},
```
dsseng: Nit: would be cool to test comments

```suggestion
		},
		{
			doc: "comments",
			fileContents: `options edns0
search .
nameserver 10.10.1.2
#nameserver 10.10.1.3
nameserver 10.10.1.4
;nameserver 10.10.1.5
nameserver 10.10.1.6
`,
			expected: []string{
				"10.10.1.2:53",
				"10.10.1.4:53",
				"10.10.1.6:53",
			},
		},
```

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
```diff
@@ -192,6 +197,68 @@ func GetQemu() Qemu {
...
+		if len(line) > 0 && (line[0] == ';' || line[0] == '#') {
+			// comment
+			continue
+		}
+
+		fields := strings.Fields(line)
+		if len(fields) > 1 && fields[0] == "nameserver" { // len(fields) > 1 because we need a value too
+			// make sure server name is just an
+			// IP address. Otherwise we need DNS
+			// to look it up.
```
dsseng: Not sure if there's a rule, but subjectively it looks more straightforward if a big comment is before if

```suggestion
		    // len(fields) > 1 because we need a value to make sure server name
			// is just an IP address. Otherwise we need DNS to look it up.
			if len(fields) > 1 && fields[0] == "nameserver" { 
```

dsseng:
Regarding the CI breakage: it looks like docs generator also finds the default using resolv.conf now. I think it can be solved by making the default to be empty, and if we encounter an empty nameserver list then we call the function to attempt getting this from resolv.conf or fallback to default. Maybe there's some other option to fix that, @Orzelius should know better

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options_test.go
laurazard: Good catch! Totally missed that in the test cases, ty.

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
laurazard: I touched it up – lmk if that looks good :)

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
```diff
@@ -192,6 +197,72 @@ func GetQemu() Qemu {
 	}
 }
 
+func getSystemDNSResolvers() []string {
+	dnsResolvers, err := parseDNSConfig("/etc/resolv.conf")
+	if err != nil {
+		cli.Warning("failed to read system DNS resolvers, using defaults...")
+
+		return DefaultQEMUNameservers
+	}
```
shanduur: I'm not sure we ever had this command working on Windows, but is there a way to make it more portable?

smira: For the QEMU variant, it only works on Linux/OS X. 

Docker variant shouldn't even have this flag (it inherits settings from docker).

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
```diff
@@ -192,6 +197,72 @@ func GetQemu() Qemu {
...
+	"8.8.8.8",
+	"1.1.1.1",
+	"2001:4860:4860::8888",
+	"2606:4700:4700::1111",
+}
+
+// getSystemDNSConfig gets the system's configured DNS resolvers.
+// It is adapted from the stdlib:
+// https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/net/dnsconfig_unix.go;l=18
+func parseDNSConfig(path string) ([]string, error) {
```
smira: I think this might not work with `systemd`, that is my `/etc/resolv.conf`:

```
nameserver 127.0.0.53
```

It is not routable from a VM. 

So I don't think in general we can/should use host's DNS.

laurazard: Hmm, I think that's a fair point, this might break w/ people running local resolvers (like systemd-resolved). What about using the host DNS if it's not a loopback address (like the whole 127.0.0.0/8 block)? Wouldn't that still be a better default than 8.8.8.8/1.1.1.1?

Or, validating the resolvers from /etc/resolv.conf and only using those if they actually work?

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
laurazard: I don't think so 😅 I did do this under the assumption that this wouldn't be used under Windows – reading `/etc/resolv.conf` should in general be an okay solution for unix/BSD family things, but I think for Windows it'd require some more platform specific API calls.

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
dsseng: Validating would be hard, since they are available to the talosctl actually. I was going to suggest removing loopback addresses, yeah. I guess resolved only binds to localhost, so unlikely to be accessible from QEMU, so then we should fallback to the default

laurazard: > Validating would be hard, since they are available to the talosctl actually.
Right, of course 🤦‍♀️ 

In addition to excluding loopback, we can always keep 8.8.8.8/1.1.1.1, just add the ones from the host in case those don't work. That'd still be a better experience in a few cases imo.

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
```diff
@@ -199,3 +204,71 @@ func GetDocker() Docker {
...
+func getSystemDNSResolvers() []string {
+	dnsResolvers, err := parseDNSConfig("/etc/resolv.conf")
+	if err != nil {
+		cli.Warning("failed to read system DNS resolvers, using defaults...")
+
+		return DefaultQEMUNameservers
+	}
+
+	if len(dnsResolvers) == 0 {
+		cli.Warning("no system DNS resolvers found, using defaults...")
```
Orzelius: This will be logged twice (there are two qemu commands, for both of which `clusteroptions.GetQemu()` is called)

Orzelius: I think this warning is not too useful and can be omitted

laurazard:
> Although the parsing fails the command returns with the error instead of using the defaults.

~~Yeah sorry, I was making some changes for loopback exclusion in the middle of some other changes to fix the flag description/docs and pushed an incomplete/bad commit 😭 Will fix it in a sec.~~

Nevermind, that was a separate issue, I hadn't actually e2e tested ipv6 and it seems the `[2404:1a8:7f01:b::3]:53` format doesn't get parsed correctly later down the line. It didn't just fallback to the default resolvers because the parsing works correctly when creating the QEMU config (for which nameservers are just a `[]string`) but fails when those strings are parsed later.

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/options.go
laurazard: Yeah, initially my idea was to have this as some sort of debug log, but I don't think we have a debug flag for talosctl `talosctl --debug [...]`), do we?

Regardless, I'm changing the implementation to add resolvers parsed from resolv.conf (instead of replacing the defaults), so this log won't make sense either way.

laurazard:
I pushed a bunch of changes that I think make sense after all the discussion:
- *Add* (don't replace) nameservers parsed from resolv.conf to the default list
- Added (non-Go version-breakingly) a `AddHostNameservers` field to the `Qemu` config struct, and changed `Qemu.ModifyClusterRequest()` to only add host nameservers if that's `true`
- Added logic to `talosctl cluster create` to set `Qemu.AddHostNameservers` to true if the `--nameservers` argument wasn't changed from it's default value
- Moved all the host nameserver fetching logic out of `GetQemu()` and into the `makers` package – this consolidates some of the nameserver handling into one place *and* moving it out of `GetQemu()` solves the docs issue, too

Regardless, from chatting with @smira I think we probably want to do the proper solution instead and proxy DNS, so I'll close this for now.

PR #12029, opened 2025-10-15
feat: unify cmdline handling GRUB/systemd-boot

### LINKED ISSUE #12019: unify cmdline handling for GRUB and systemd-boot
1. Deprecate and hide `machine.install.extraKernelArgs`.
2. Add new flag `machine.install.grubCmdline` (???? what is the name) which is enabled by default for Talos 1.12+. It should be not valid to have both extraKernelArgs and this new flag.
3. If this flag is enabled, installer with GRUB now behaves like systemd-boot: cmdline is not computed on the host, but instead simply extracted from the UKI and saved to grub config without any modifications.

Reasons:

* remove a gap between two bootloaders in how cmdline is handled
* allow for any changes to the cmdline via schematic
* this flag will not be enabled for upgrades, so for existing installs it will be no-op

smira:
Use cmdline from the UKI in Talos 1.12+ by default for new installs.

This brings GRUB in line with systemd-boot vs. cmdline behavior.

Fixes #12019

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -127,6 +127,16 @@ By default module signature verification is enabled (`module.sig_enforce=1`).
 When using Factory or Imager supply as `-module.sig_enfore module.sig_enforce=0` kernel parameters to disable module signature enforcement.
 """
 
+    [notes.grub]
+        title = "GRUB"
+        description = """\
+Talos Linux introduces new machine configuration option `.machine.install.grubUseUKICmdline` to control whether GRUB should use the kernel command line 
+provided by the boot assets (UKI) or to use the command line constructed by Talos itself (legacy behavior).
+
+This option defaults to `true` for new installations, which means that GRUB will use the command line from the UKI, making it easier to customize kernel parameters via boot asset generation.
+For existing installations upgrading to v1.12, this option will default to `false` to preserve the legacy behavior.
+"""
+
 [make_deps]
 
     [make_deps.tools]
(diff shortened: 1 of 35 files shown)


hack/release.toml
```diff
@@ -103,6 +103,15 @@ These fields were removed from the default machine configuration schema in v1.12
 etcd container image is now pulled from `registry.k8s.io/etcd` instead of `gcr.io/etcd-development/etcd`.
 """
 
+    [notes.grub]
+        title = "GRUB"
+        description = """\
+Talos Linux introduces new machine configuration option `.machine.install.grubUseUKICmdline` to control whether GRUB should use the kernel command line
```
frezbo, line 133: should we actually start a new install multi doc config and add it there, this way we can start slowly moving away from `.machie.install` to a new InstallConfig. This could be even a completely separate document. Feels like we should not add more stuff to `.machine.install`

pkg/machinery/constants/constants.go
```diff
@@ -1313,6 +1313,9 @@ const (
 	DefaultOOMCgroupRankingExpression = `memory_max.hasValue() ? 0.0 :
 		{Besteffort: 1.0, Burstable: 0.5, Guaranteed: 0.0, Podruntime: 0.0, System: 0.0}[class] *
 		   double(memory_current.orValue(0u)) / double(memory_peak.orValue(0u) - memory_current.orValue(0u))`
+
```
frezbo, line 1320: https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html#SMBIOS%20Type%2011%20Strings comment should update it's for systemd-stub, not systemd-boot

hack/release.toml
smira: I had discussion with @utkuozdemir about it, and my point is:

* yes, we want to drop `.machine.install`
* if we drop it, we need to replace it with something new (probably it's not even a config, but arguments to the new API call)
* if we want to create new "MachineInstallConfig" document, I would rather move _all_ options into it, rather than introduce a new machine config document with a single flag, while all other flags stay in `.machine.install`

So I feel at this moment, as we're not replacing `.machine.install` for 1.12, the least evil is to add a new field there. If we move to Install API, we don't need this machine config at all.

pkg/machinery/constants/constants.go
smira: ah yeah, thanks, I will fix it

PR #12038, opened 2025-10-16
feat: implement new DHCP network configuration

### LINKED ISSUE #10958: Network Config: DHCP
### DHCP

Enable DHCP (v4/v6) on the link. DHCPv4 is run by default on links if no network configuration is not submitted.

```yaml
version: v1alpha1
kind: DHCPv4Config
name: en0sp2 # name of the link to run DHCP on
routeMetric: 1024 # optional, set route metric for routes received from DHCP
receiveHostname: true # defaults to true, can be used to disable receiving hostname from DHCP
```

```yaml
version: v1alpha1
kind: DHCPv6Config
name: en0sp2 # name of the link to run DHCP on
routeMetric: 1024 # optional, set route metric for routes received from DHCP
duid: hex # DHCPv6 client DUID (optional)
receiveHostname: true # defaults to true, can be used to disable receiving hostname
```

### LINKED ISSUE #11661: Feature request: Support 'dhcp-identifier: mac' in machine network config
## Feature Request
We are evaluating Talos as a way to simplify and standardize our Kubernetes clusters.
Currently we use kubeadm on standard Linux VMs. Our corporate IT environment makes it difficult to manage OS variance, so the Talos model (immutable, declarative, API-driven) is very attractive for us.

In Kubernetes we need worker nodes to always get the same IP address.
In our corporate VM environment, the DHCP infrastructure achieves this by requiring DHCPv4 ClientIdentifier = MAC. Standard Linux distributions expose this option via systemd-networkd ([DHCPv4] ClientIdentifier=mac). Talos does not currently expose any way to set this.

### Description
As an experiment, I rebuilt metal-amd64.iso with a modification to dhcp4.go to force ClientIdentifier=mac. It worked for a plain ISO boot, but once installed to disk the networking path changed and the patch was no longer effective. Rebuilding the full image pipeline became too complex to pursue further.

We would like to request a supported feature in machineconfig similar to:
```
network:
  hostname: uni-s1-w3
  interfaces:
    - interface: ens160
      dhcp: true
      dhcp-identifier: mac
```
This would allow DHCP reservations in our environment to function correctly. As far as we can tell, there is currently no way to achieve this behavior with Talos.

Because we want to stay on supported upstream releases and upgrade Talos over time, a one-off rebuild is not a viable long-term solution.

Thank you for considering this request.

smira:
Fixes #11661

Fixes #10958

This also implement proper client identifier handling.


pkg/provision/providers/qemu/qemu.go
```diff
@@ -48,21 +55,52 @@ func (p *provisioner) GenOptions(networkReq provision.NetworkRequest, contract *
...
-			v1alpha1.WithNetworkInterfaceDHCPv4(virtioSelector, hasIPv4),
-			v1alpha1.WithNetworkInterfaceDHCPv6(virtioSelector, hasIPv6),
-		),
+	}
+
+	var bundleOpts []bundle.Option
+
+	if contract.MultidocNetworkConfigSupported() {
+		aliasConfig := networkcfg.NewLinkAliasConfigV1Alpha1("net0")
+		aliasConfig.Selector = networkcfg.LinkSelector{
```
smira, line 66: use new network config for our QEMU provisioner:

create an alias for the only link to be `net0`, and run DHCP4/6 on it as requested

go.mod
```diff
@@ -23,10 +23,6 @@ replace (
 	gopkg.in/yaml.v3 => github.com/unix4ever/yaml v0.0.0-20220527175918-f17b0f05cf2c
 )
 
-// fd-leak related replacements: https://github.com/siderolabs/talos/issues/9412
-// https://github.com/insomniacslk/dhcp/pull/550
```
smira, line 27: this PR got merged, so we can stop using that fork now

frezbo (approved):

PR #12039, opened 2025-10-16
docs/test: improvements for OOM controller

dsseng:
- Disable OOM controller in container mode
- Log OOM events
- Add an integration test verifying the system can sustain an OOM event


internal/integration/k8s/testdata/oom.yaml
```diff
@@ -0,0 +1,32 @@
...
+    matchLabels:
+      app: stress-mem
+  template:
+    metadata:
+      labels:
+        app: stress-mem
+    spec:
+      containers:
+        - name: stress-mem
+          image: ghcr.io/dsseng/stress-ng:v1
```
dsseng: I believe this image should be replaced or moved to @siderolabs org

I'll take a look if I can find an alternative, as this boils down to just installing stress-ng on top of Alpine image

internal/integration/k8s/testdata/oom-50-replicas.yaml
```diff
@@ -0,0 +1,3 @@
+---
+spec:
+  replicas: 50
```
dsseng, line 3: This replica count is enough to overload 3 GB VMs I use, so should be fine for 2/4 GB we have in CI. 50 per 100 MB each should attempt to consume up to 5 GiB RAM

pkg/machinery/resources/runtime/oom_action.go
```diff
@@ -0,0 +1,72 @@
...
+			{
+				Name:     "Time",
+				JSONPath: `{.time}`,
+			},
+			{
+				Name:     "Score",
+				JSONPath: `{.score}`,
+			},
+		},
+		Sensitivity: meta.Sensitive,
```
dsseng, line 58: Considering this sensitive as it exposes process names and full cmdlines

internal/integration/k8s/oom.go
```diff
@@ -0,0 +1,127 @@
...
+	suite.Assert().True(suite.waitForOOMKilled(ctx, 15*time.Second, "stress-ng"))
+
+	// Scale to 1, wait for deployment to scale down, proving system is operational
+	suite.PatchK8sObject(ctx, "default", "apps", "Deployment", "v1", "stress-mem", oom1ReplicaPatch)
+	suite.Require().NoError(suite.WaitForDeploymentAvailable(ctx, time.Minute, "default", "stress-mem", 1))
+
+	suite.APISuite.AssertClusterHealthy(ctx)
+}
+
+// Waits for a period of time and return
```
dsseng:
```suggestion
// Waits for a period of time and returns whether or not OOM events containing a specified process have been observed
```

Oops, haven't completed this. Here I want to let it spin the overloaded workload for 15 seconds, regardless of how soon the first OOM kill happens. This is to ensure nothing in the OS glitches up due to memory pressure

internal/integration/k8s/testdata/oom.yaml
frezbo: you can use nixery https://nixery.dev/

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -213,8 +273,17 @@ func reapCg(logger *zap.Logger, cgroupPath string) error {
 	// Open pidfd's of all the processes in cgroup to accelerate kernel
 	// garbage-collecting those processes via mrelease.
 	pidfds := []int{}
+	cmdlines := []string{}
 
 	for _, pid := range processes {
+		cmdBytes, err := os.ReadFile(path.Join("/proc", strconv.Itoa(pid), "cmdline"))
```
smira: nit:

```suggestion
		cmdBytes, err := os.ReadFile(filepath.Join("/proc", strconv.Itoa(pid), "cmdline"))
```

internal/app/machined/pkg/controllers/runtime/oom.go
```diff
@@ -134,10 +156,46 @@ func (ctrl *OOMController) Run(ctx context.Context, r controller.Runtime, logger
...
+
+		if err = safe.CleanupOutputs[*runtimeres.OOMAction](ctx, r); err != nil {
+			return err
+		}
+
 		// TODO: evaluate on different cgroups, not only root. E.g. action when podruntime experiences high PSI.
 		if trigger {
+			score, processes := ctrl.OomAction(logger, ctrl.CgroupRoot, scoringExpr)
+
+			evalContext["time_since_trigger"] = time.Since(ctrl.ActionTriggered).String()
```
smira: should it be `.String()` or `time.Duration` ?

internal/integration/k8s/testdata/oom.yaml
smira: I've copied it over: https://github.com/orgs/siderolabs/packages/container/package/stress-ng

internal/integration/k8s/oom.go
```diff
@@ -0,0 +1,127 @@
...
+
+// TestOom verifies that system remains stable after handling an OOM event.
+func (suite *OomSuite) TestOom() {
+	if suite.Cluster == nil {
+		suite.T().Skip("without full cluster state reaching out to the node IP is not reliable")
+	}
+
+	if suite.Cluster.Provisioner() != base.ProvisionerQEMU {
+		suite.T().Skip("skipping OOM test since provisioner is not qemu")
+	}
```
smira, line 57:
```go
if testing.Short() {
   t.Log("skipping in short mode"
}
```

this way it will not run for short tests, only for full runs

pkg/machinery/resources/runtime/oom_action.go
```diff
@@ -0,0 +1,72 @@
...
+const OOMActionType = resource.Type("OOMActions.talos.dev")
+
+// OOMAction is the OOM action record resource.
+type OOMAction = typed.Resource[OOMActionSpec, OOMActionExtension]
+
+// OOMActionSpec describes the OOM action record resource properties.
+//
+//gotagsrewrite:gen
+type OOMActionSpec struct {
+	Time           time.Time      `yaml:"time" protobuf:"1"`
```
smira: we could probably skip this, we have `.Metadata().CreatedAt()` which is basically same?

internal/app/machined/pkg/controllers/runtime/oom.go
dsseng: time.Duration fails to serialize (or either de-serialize on the talosctl end, I don't recall exactly)

internal/integration/k8s/oom.go
```diff
@@ -0,0 +1,128 @@
...
+
+	timeoutCh := time.After(timeout)
+	ret := false
+
+	for {
+		select {
+		case <-timeoutCh:
+			return ret
+		case ev := <-watchCh:
+			if ev.Type != state.Created /*|| ev.Resource.Metadata().Created().Before(startTime)*/ {
```
smira: this should be uncommented?

dsseng: oops, correct. thanks!

Issue #11782, opened 2025-09-08
support "panic" reboot mode


smira:
Support for panic-style reboot when Talos doesn't try to shutdown any workloads/wait for kubelet to terminate.

See e.g. https://github.com/siderolabs/talos/issues/11775

dsseng:
Other name options: `nowait`, `force`, `immediate`

especially-relative:
I'm not sure what is happening but I have a node that crashes periodically. My suspicion is a rogue OOMKill happening but have not been able to get conclusive logs just yet. In all cases I've left a monitor/keyboard connect and the node is unresponsive to all input.

An automatic panic reboot mode would be a huge help. I'd much rather a node reboot than hang forever. Should be configurable, obvious as some situations would not prefer that. 

(At some point I'll get a PiKVM or simlar going so I can kick the node but hopefully I manage to actually fix the real issue. Next up is trying to enable swap so memory spikes are "softer")

steverfrancis:
> I'm not sure what is happening but I have a node that crashes periodically. My suspicion is a rogue OOMKill happening but have not been able to get conclusive logs just yet. In all cases I've left a monitor/keyboard connect and the node is unresponsive to all input.
> 
> An automatic panic reboot mode would be a huge help. I'd much rather a node reboot than hang forever. Should be configurable, obvious as some situations would not prefer that.
> 
> (At some point I'll get a PiKVM or simlar going so I can kick the node but hopefully I manage to actually fix the real issue. Next up is trying to enable swap so memory spikes are "softer")

I think you are looking for https://www.talos.dev/v1.11/advanced/watchdog/
No?

especially-relative:
Yeah! Maybe! I had missed that, thanks. I will give it a shot at least. 

I don't know enough about how that works to say if there are some panics that the initial suggestion would be helpful with that the watchdog doesn't cover.

especially-relative:
Update: I applied the a patch to enable the watchdog, It took a few days but the node crashed again. No luck on the watchdog trigger, and when I went to look the dashboard screen was also not responsive

Issue #11775, opened 2025-09-06
Talos node stuck in rebooting (or shutting down) state


B0bby31:
### Description

I am running an HA kubernetes cluster and it had been work for a while. At one point, the base kubernetes services (i.e. api-server, controller, scheduler) got stuck in a terminating state on one of the nodes. To remedy this, I decided to use talosctl to reboot the node, but it got stuck doing this:

 stopAllPods: failed waiting for kubelet lifecycle finalizers: context deadline exceeded.

Now, it is perpetually stuck. I can still interact a bit with the nodes via talosctl, but I am unable to reset or shutdown the node. Since I am not on site, I cannot reboot the node manually.

Also, when interacting with the node, it only allows certain operations. For instance, I tried to run talos support, but this fails on the stuck node.

I suppose my main question is why is there no way to force a reboot or shutdown. I have tried resetting the node, but, now that it is stuck shutting down, it will not work.

### Logs

I managed to get these logs from the api server on the stuck node before deleting it:

192.168.231.8: {"ts":1757010279252.6155,"caller":"collectors/resource_metrics.go:161","msg":"Error getting summary for resourceMetric prometheus endpoint","err":"failed to list pod stats: failed to get pod or container map: failed to list all containers: rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
192.168.231.8: {"ts":1757010283211.4668,"caller":"kubelet/kubelet.go:2460","msg":"Skipping pod synchronization","err":"PLEG is not healthy: pleg was last seen active 16h14m34.439643899s ago; threshold is 3m0s","errCauses":[{"error":"PLEG is not healthy: pleg was last seen active 16h14m34.439643899s ago; threshold is 3m0s"}]}
192.168.231.8: {"ts":1757010283489.8174,"caller":"server/server.go:336","msg":"Unable to authenticate the request due to an error","err":"[invalid bearer token, service account token has been invalidated]","errCauses":[{"error":"[invalid bearer token, service account token has been invalidated]"}]}
192.168.231.8: {"ts":1757010288212.0938,"caller":"kubelet/kubelet.go:2460","msg":"Skipping pod synchronization","err":"PLEG is not healthy: pleg was last seen active 16h14m39.440232139s ago; threshold is 3m0s","errCauses":[{"error":"PLEG is not healthy: pleg was last seen active 16h14m39.440232139s ago; threshold is 3m0s"}]}
[...]

PR #12049, opened 2025-10-20
feat(machined): add panic/force mode reboot

laurazard:
# Pull Request

closes https://github.com/siderolabs/talos/issues/11782

## What? (description)

Add a forceful mode to the reboot sequence (`talosctl reboot --mode force`) that bypasses graceful userspace teardown.

## Why? (reasoning)

In certain situations, Talos's shutdown/reboot sequence hangs while waiting for services/mounts to be gracefully stopped (see: https://github.com/siderolabs/talos/issues/11775).

------------

I left a few questions as TODOs, PTAL at them if you review! Mainly I'm wondering about `*machine.RebootRequest` vs. `runtime.XxxOptions`, and whether to error out if we don't get a `*machine.RebootRequest` or just silently continue (which is fine for this, since any hypothetical older client wouldn't support force-reboot anyway, and the default is "graceful reboot".

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/reboot.go
+++ b/cmd/talosctl/cmd/talos/reboot.go
@@ -39,6 +39,8 @@ var rebootCmd = &cobra.Command{
 		// skips kexec and reboots with power cycle
 		case "powercycle":
 			opts = append(opts, client.WithPowerCycle)
+		case "force":
+			opts = append(opts, client.WithForce)
 		case "default":
 		default:
 			return fmt.Errorf("invalid reboot mode: %q", rebootCmdFlags.mode)
@@ -85,7 +87,7 @@ func rebootGetActorID(opts ...client.RebootMode) func(ctx context.Context, c *cl
 }
 
 func init() {
-	rebootCmd.Flags().StringVarP(&rebootCmdFlags.mode, "mode", "m", "default", "select the reboot mode: \"default\", \"powercycle\" (skips kexec)")
+	rebootCmd.Flags().StringVarP(&rebootCmdFlags.mode, "mode", "m", "default", "select the reboot mode: \"default\", \"powercycle\" (skips kexec), \"force\" (skips graceful teardown)")
 	rebootCmdFlags.addTrackActionFlags(rebootCmd)
 	addCommand(rebootCmd)
 }
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
@@ -381,7 +381,14 @@ func (c *Controller) phases(seq runtime.Sequence, data any) ([]runtime.Phase, er
 
 		phases = c.s.Shutdown(c.r, in)
 	case runtime.SequenceReboot:
-		phases = c.s.Reboot(c.r)
+		var in *machine.RebootRequest
+		if req, ok := data.(*machine.RebootRequest); ok {
+			in = req
+		} else {
+			log.Printf("warning: API reboot missing reboot request")
+		}
+
+		phases = c.s.Reboot(c.r, in)
 	case runtime.SequenceUpgrade:
 		in, ok := data.(*machine.UpgradeRequest)
 		if !ok {
--- a/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller_test.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller_test.go
@@ -44,7 +44,7 @@ func (m *mockSequencer) Install(r runtime.Runtime) []runtime.Phase {
 	return m.phases[runtime.SequenceInstall]
 }
 
-func (m *mockSequencer) Reboot(r runtime.Runtime) []runtime.Phase {
+func (m *mockSequencer) Reboot(r runtime.Runtime, _ *machine.RebootRequest) []runtime.Phase {
 	return m.phases[runtime.SequenceReboot]
 }
 
--- a/internal/integration/api/reboot.go
+++ b/internal/integration/api/reboot.go
@@ -10,12 +10,14 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/siderolabs/go-retry/retry"
 
 	"github.com/siderolabs/talos/internal/integration/base"
+	machineapi "github.com/siderolabs/talos/pkg/machinery/api/machine"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/config/machine"
 )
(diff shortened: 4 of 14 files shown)


pkg/machinery/client/events.go
```diff
@@ -248,6 +249,7 @@ func UnmarshalEvent(event *machineapi.Event) (*Event, error) {
 	}
 
 	if msg == nil {
+		log.Printf("event not supported: %s", typeURL)
 		// We haven't implemented the handling of this event yet.
 		return nil, ErrEventNotSupported
```
shanduur, line 253: Do we need to log here, or can we just wrap the returned error with the `typeURL`?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -381,7 +381,19 @@ func (c *Controller) phases(seq runtime.Sequence, data any) ([]runtime.Phase, er
...
 	case runtime.SequenceReboot:
-		phases = c.s.Reboot(c.r)
+		// TODO: we're both using *machine.XxxRequest and
+		// runtime.XxxOptions types here. Is one of these preferred?
+		// Might be kind of confusing to be passing a request here
+		// when the request also gets passed to the specific tasks
+		// via controller.runTask().
+		var in *machine.RebootRequest
+		if req, ok := data.(*machine.RebootRequest); ok {
+			// TODO: complain if missing? are there older clients
```
shanduur: We should probably at least log here.

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller_test.go
```diff
@@ -44,7 +44,7 @@ func (m *mockSequencer) Install(r runtime.Runtime) []runtime.Phase {
 	return m.phases[runtime.SequenceInstall]
 }
 
-func (m *mockSequencer) Reboot(r runtime.Runtime) []runtime.Phase {
+func (m *mockSequencer) Reboot(r runtime.Runtime, req *machine.RebootRequest) []runtime.Phase {
```
shanduur: Do we need `req *machine.RebootRequest` here? It appears unused.

pkg/machinery/client/events.go
laurazard: Yeah, I was going to ask about this – I'm not sure about in general what we tend to do in `talosctl`, and (from what I saw) logging seems kind of inconsistent – I had to add this to even figure out what was going wrong because I was just getting very generic error output.

I'm happy to change the error returned, (after checking whether we're using `ErrEventNotSupported` isn't being used as a sentinel anywhere/compared to directly).

(As a sidenote, I wish we had a `talosctl --debug ...` and debug level logging.)

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller_test.go
laurazard: it needs to satisfy the `Sequencer` interface, and the interface had to be changed so that we could skip the graceful teardown phases: https://github.com/siderolabs/talos/blob/8e9d12f18bee0f247d434ee53490a690c7b9af0c/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go#L236-L251

In the current Sequencer flow, the request gets passed to individual tasks, which is how the powercycle flag is currently working: https://github.com/siderolabs/talos/blob/8e9d12f18bee0f247d434ee53490a690c7b9af0c/internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go#L1629-L1637

However, `Sequencer.Reboot` did not receive the request before, and I thought adding it there and just skipping the other phases was more readable than adding 10x `if req.Force { //skip this }` to the tasks.

laurazard: (but we can have a 
```suggestion
func (m *mockSequencer) Reboot(r runtime.Runtime, _ *machine.RebootRequest) []runtime.Phase {
```

instead, which now I'm wondering if that's what you meant 😅 🤦‍♀️ )

shanduur: I think linters should scream at you for not doing `_ *machine.RebootRequest` 💀

shanduur: FYI we have a make `lint` target (which I keep forgetting about).

laurazard: > I think linters should scream at you for not doing _ *machine.RebootRequest 💀

YEAH I was surprised it didn't scream it – I did run `make lint` 😅 (the PR template always reminds me)

Maybe it has to do with the integration build tags/etc. I'll take a sec to check that later.

pkg/machinery/client/events.go
laurazard: It feels weird to wrap `ErrEventNotSupported` with the typeURL since the typeURL is a part of why that error happened, so I pushed a commit changing `ErrEventNotSupported` into a
```go
type ErrEventNotSupported struct {
  TypeURL string
}
...
```
However, in practice we can't/shouldn't do this outside of a major version, since it changes a public API (hence me leaving it separate). In the meantime, we can either keep it as-is, or wrap it like suggested.

laurazard: Oh, I didn't realize having an extra commit meant CI would complain/tests won't run 😭 I'll remove it then and just skip it for now.

cmd/talosctl/cmd/talos/reboot.go
smira, line 90: should update the help message here

and also add a short note please to `hack/release.toml`

internal/integration/api/reboot.go
```diff
@@ -71,6 +73,71 @@ func (suite *RebootSuite) TestRebootNodeByNode() {
...
+		suite.T().Skip("cluster doesn't support reboots")
+	}
+
+	nodes := suite.DiscoverNodeInternalIPs(suite.ctx)
+	suite.Require().NotEmpty(nodes)
+
+	for _, node := range nodes {
+		suite.T().Log("force rebooting node", node)
+
+		nodeCtx := client.WithNodes(suite.ctx, node)
```
smira:
```suggestion
		nodeCtx := client.WithNode(suite.ctx, node)
```

laurazard: Good catch, thanks!

cmd/talosctl/cmd/talos/reboot.go
laurazard: I added a small note, can you (and/or others) check if it looks good/conforms to how you usually write there?

smira: release.toml is good, but please update the help message for the flag on line 88/90

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -381,7 +381,19 @@ func (c *Controller) phases(seq runtime.Sequence, data any) ([]runtime.Phase, er
 
 		phases = c.s.Shutdown(c.r, in)
 	case runtime.SequenceReboot:
-		phases = c.s.Reboot(c.r)
+		// TODO: we're both using *machine.XxxRequest and
+		// runtime.XxxOptions types here. Is one of these preferred?
+		// Might be kind of confusing to be passing a request here
+		// when the request also gets passed to the specific tasks
+		// via controller.runTask().
```
laurazard: Anyone have thoughts on this?

smira: I think it was an ugly workaround to pass this any -> specific type, and I don't think any change can make it worse or better.

As it works here, and enables us to do a proper implementation, I would say 👍

smira:
Looks like "regular" reboot is broken now (I was looking at `integration-qemu` logs), @laurazard please let me know if you need help to troubleshoot/read logs.

pkg/machinery/client/events.go
smira: tests should still run, it only fails `conform` checks

rcdailey:
@laurazard thank you so much for doing this!

PR #12050, opened 2025-10-20
feat: use image signer

frezbo:
Use `image-signer` to sign images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ GENERATE_VEX_PREFIX ?= ghcr.io/siderolabs/generate-vex
 GENERATE_VEX ?= latest
 
 KRES_IMAGE ?= ghcr.io/siderolabs/kres:latest
+IMAGE_SIGNER_IMAGE ?= ghcr.io/siderolabs/image-signer:latest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -663,9 +664,7 @@ image-list: ## Prints a list of all images built by this Makefile with digests.
 .PHONY: sign-images
 sign-images: ## Run cosign to sign all images built by this Makefile.
 	@for image in $(shell $(MAKE) --quiet image-list REGISTRY_AND_USERNAME=$(REGISTRY_AND_USERNAME) IMAGE_TAG_IN=$(IMAGE_TAG_IN)); do \
-		echo '==>' $$image; \
-		cosign verify $$image --certificate-identity-regexp '@siderolabs\.com$$' --certificate-oidc-issuer https://accounts.google.com || \
-			cosign sign --yes $$image; \
+	        docker run --pull=always --rm --net=host $(IMAGE_SIGNER_IMAGE) sign $$image; \
```
smira: does it go one-by one here?

frezbo: ahh I can make it one time and drop the loop, forgot

PR #12056, opened 2025-10-21
fix: drop 'ro' falg from defaults

shanduur:
Flag 'ro' should not be applied as default to ReadOnly filesystems.

### DIFF
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -495,7 +495,6 @@ func (ctrl *MountController) handleDiskMountOperation(
 
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -495,7 +495,6 @@ func (ctrl *MountController) handleDiskMountOperation(
 
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
-			fsOpts = append(fsOpts, fsopen.WithBoolParameter("ro"))
```
smira, line 498: do we need to force it for iso9660? 

let's please promote to `integration/misc`

laurazard (approved):
LGTM

internal/app/machined/pkg/controllers/block/mount.go
shanduur: Yes, we need it in `iso9660`, but it's handled by the package already.

PR #12058, opened 2025-10-22
feat: support bootloader option for ISO

frezbo:
Support selecting bootloader option for ISO.


pkg/imager/profile/output.go
```diff
@@ -27,6 +27,11 @@ type Output struct {
 	//  * .xz - output xz archive
 	//  * .gz - output gz archive
 	OutFormat OutFormat `yaml:"outFormat"`
+	// Bootloader is the bootloader to use for the image.
+	// If not set, it defaults to dual-boot.
+	// Secureboot always uses sd-boot.
+	// ARM64 uses sd-boot for Talos >= 1.10.
+	Bootloader ImageBootloader `yaml:"bootloader"`
```
smira: This feels like a breaking change to me... 

As much as I hate it, I feel we might need to duplicate this in Image and ISO options :(

frezbo: I went that route first and felt we were duplicating a lot of code sadly, I can go back to that

smira: I guess the only duplication will be guessing the bootloader type?

We could probably be a bit more smart and provide "auto/none" bootloader type, so that default value is "empty", and we guess like before, but you can override this choice if you want to.

PR #12060, opened 2025-10-22
fix: validate provisioner when destroying local clusters

Orzelius:
Return an error when attempting to destroy a cluster that was created with a different provisioner.

This fixes a bug where the qemu cluster state is removed without any actual cleanup being done when running `talosctl cluster destroy` (provisioner defaults to docker).

* move the the `vm.State` logic into -> `provision.State` as it's now reused by the docker provisioner as well.
* relax the validation of the omni-api-endpoint flag


cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/helpers.go
```diff
@@ -25,9 +21,5 @@ func ParseOmniAPIUrl(urlIn string) (*url.URL, error) {
 		return nil, err
 	}
```
smira: https://github.com/siderolabs/talos/blob/1cf12ea9cd9f054987edd17f89cb001a0bdfe243/pkg/machinery/config/types/siderolink/siderolink.go#L131-L132

Use this method, please don't do any other validations.

If you want to validate, just put it into the config document and run validate on it

cmd/talosctl/cmd/mgmt/cluster/destroy.go
```diff
@@ -39,6 +39,16 @@ func destroy(ctx context.Context) error {
 
 	defer provisioner.Close() //nolint:errcheck
 
+	state, err := provision.ReadState(ctx, PersistentFlags.ClusterName, PersistentFlags.StateDir)
+	if err != nil {
+		return fmt.Errorf("failed to read cluster state: %w", err)
+	}
+
+	if state.ProvisionerName != provisionerName {
```
smira: can we just guess provisioner from the cluster name automatically, given that we have the state now?

website/content/v1.12/reference/cli.md
```diff
@@ -327,7 +327,7 @@ talosctl cluster create [flags]
 
 ## talosctl cluster destroy
 
-Destroys a local docker-based or firecracker-based kubernetes cluster
+Destroys a local Talos kubernetes cluster
```
smira, line 329: while at it, let's please hide here `talosctl cluster create` (dev) if we can 

and document available `--presets` values for `cluster create qemu`

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/helpers.go
Orzelius: The method doesn't seem to check for the presence of the `jointoken` though. Is there a case where the join token is not necessary?

cmd/talosctl/cmd/mgmt/cluster/destroy.go
Orzelius: Sure! It'd be a breaking change, but should I remove the provisioner flag from the command as well?

website/content/v1.12/reference/cli.md
Orzelius: > while at it, let's please hide here talosctl cluster create (dev) if we can

I'll give it a try but won't spend too much time on it if there's no clean way to hide the parent command.

> and document available --presets values for cluster create qemu

The presents are already documented in the [cli reference](https://github.com/siderolabs/talos/blob/main/website/content/v1.12/reference/cli.md#talosctl-cluster-create-qemu), [cmd help](https://github.com/siderolabs/talos/blob/main/cmd/talosctl/cmd/mgmt/cluster/create/cmd_qemu.go#L49-L58) and [the qemu guide](https://github.com/siderolabs/docs/blob/main/public/talos/v1.12/platform-specific-installations/local-platforms/qemu.mdx#presets). Is there any other place they should be documented?

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/helpers.go
smira: Let's not build assumptions that might change in the future - e.g. we might allows tokenless joins to Omni.

What you're accepting is a SideroLink URL, and it should be valid from the perspective of the machine config.

cmd/talosctl/cmd/mgmt/cluster/destroy.go
smira: I'd say hide/ignore it. You have state, so name uniquely identifies a cluster.

website/content/v1.12/reference/cli.md
smira: I would put them as a list to the flag help (just a list) if possible... I always miss long help, look at the flag value and miss it :)

cmd/talosctl/cmd/mgmt/cluster/destroy.go
Orzelius: I've marked it deprecated (hides it and logs a message if it's provided)

PR #12074, opened 2025-10-24
feat: add directory backed UserVolumes

### LINKED ISSUE #11848: support directory-based user volumes
The problem: a mixed cluster with new-style `UserVolumes` on additional disks and single-disk machines which have only `EPHEMERAL`.

Support user volumes which are backed by a directory, so `/var/mnt/<something>` is just a directory on EPHEMERAL (`/var`).

shanduur:
New field in UserVolumeConfig - `volumeType` that defaults to `partition`, but can be set to `directory`.
When set to `directory`, provisioning and filesystem operations are skipped and a directory is created under `/var/mnt/<name>`.

The `directory` type enables lightweight storage volumes backed by a host directory, instead of requiring a full block device partition.

When `volumeType = "directory"`:
- A directory is created at `/var/mnt/<metadata.name>`;
- `provisioning`, `filesystem` and `encryption` are prohibited.

Note: this mode does not provide filesystem-level isolation and inherits the EPHEMERAL partition capacity limits.
It should not be used for workloads requiring predictable storage quotas.

```yaml
apiVersion: v1alpha1
kind: UserVolumeConfig
name: test
volumeType: directory
```

Resolves #11848

### DIFF
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -293,7 +293,7 @@ func (ctrl *MountController) handleMountOperation(
 ) error {
 	switch volumeStatus.TypedSpec().Type {
 	case block.VolumeTypeDirectory:
-		return ctrl.handleDirectoryMountOperation(rootPath, mountTarget, volumeStatus)
+		return ctrl.handleDirectoryMountOperation(logger, rootPath, mountTarget, mountRequest, volumeStatus)
 	case block.VolumeTypeOverlay:
 		return ctrl.handleOverlayMountOperation(logger, filepath.Join(rootPath, mountTarget), mountRequest, volumeStatus)
 	case block.VolumeTypeSymlink:
@@ -312,8 +312,10 @@ func (ctrl *MountController) handleMountOperation(
 }
 
 func (ctrl *MountController) handleDirectoryMountOperation(
+	logger *zap.Logger,
 	rootPath string,
 	target string,
+	mountRequest *block.MountRequest,
 	volumeStatus *block.VolumeStatus,
 ) error {
 	targetPath := filepath.Join(rootPath, target)
@@ -333,7 +335,91 @@ func (ctrl *MountController) handleDirectoryMountOperation(
 		}
 	}
 
+	if volumeStatus.TypedSpec().MountSpec.BindTarget != nil {
+		if err := ctrl.handleBindMountOperation(
+			logger,
+			rootPath, target, *volumeStatus.TypedSpec().MountSpec.BindTarget,
+			mountRequest, volumeStatus,
+		); err != nil {
+			return fmt.Errorf("target path %q is not a directory", targetPath)
+		}
+	}
+
 	return ctrl.updateTargetSettings(targetPath, volumeStatus.TypedSpec().MountSpec)
+}
+
+func (ctrl *MountController) handleBindMountOperation(
+	logger *zap.Logger,
+	rootPath string,
+	source string,
+	bindTarget string,
+	mountRequest *block.MountRequest,
+	volumeStatus *block.VolumeStatus,
+) error {
+	_, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+
+	// mount hasn't been done yet
+	if !ok {
+		mountSource := filepath.Join(rootPath, source)
+		mountTarget := filepath.Join(rootPath, bindTarget)
+
+		if err := os.Mkdir(mountTarget, volumeStatus.TypedSpec().MountSpec.FileMode); err != nil {
+			if !os.IsExist(err) {
+				return fmt.Errorf("failed to create target path: %w", err)
+			}
+
+			st, err := os.Stat(mountTarget)
+			if err != nil {
+				return fmt.Errorf("failed to stat target path: %w", err)
+			}
+
+			if !st.IsDir() {
+				return fmt.Errorf("target path %q is not a directory", mountTarget)
+			}
+		}
+
+		var opts []mount.ManagerOption
+
+		opts = append(opts,
+			mount.WithSelinuxLabel(volumeStatus.TypedSpec().MountSpec.SelinuxLabel),
+		)
+
+		manager := mount.NewManager(slices.Concat(
+			[]mount.ManagerOption{
+				mount.WithTarget(mountTarget),
+				mount.WithOpentreeFromPath(mountSource),
+				mount.WithPrinter(logger.Sugar().Infof),
+			},
+			opts,
+		)...)
+
+		mountpoint, err := manager.Mount()
+		if err != nil {
+			return fmt.Errorf("failed to mount %q: %w", mountRequest.Metadata().ID(), err)
+		}
+
+		if !mountRequest.TypedSpec().ReadOnly && !mountRequest.TypedSpec().Detached {
+			if err = ctrl.updateTargetSettings(mountTarget, volumeStatus.TypedSpec().MountSpec); err != nil {
+				manager.Unmount() //nolint:errcheck
+
+				return fmt.Errorf("failed to update target settings %q: %w", mountRequest.Metadata().ID(), err)
+			}
+		}
+
+		logger.Info("bind mount",
+			zap.String("volume", volumeStatus.Metadata().ID()),
+			zap.String("source", mountSource),
+			zap.String("target", mountTarget),
+		)
+
+		ctrl.activeMounts[mountRequest.Metadata().ID()] = &mountContext{
+			point:     mountpoint,
+			readOnly:  mountRequest.TypedSpec().ReadOnly,
+			unmounter: manager.Unmount,
+		}
+	}
+
+	return nil
 }
 
 //nolint:gocyclo
@@ -645,7 +731,7 @@ func (ctrl *MountController) handleUnmountOperation(
 ) error {
 	switch volumeStatus.TypedSpec().Type {
 	case block.VolumeTypeDirectory:
-		return nil
+		return ctrl.handleDirectoryUnmountOperation(logger, mountRequest, volumeStatus)
 	case block.VolumeTypeTmpfs:
 		return fmt.Errorf("not implemented yet")
 	case block.VolumeTypeDisk, block.VolumeTypePartition, block.VolumeTypeOverlay:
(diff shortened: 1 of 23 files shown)


pkg/machinery/config/types/block/user_volume_config.go
```diff
@@ -168,7 +181,22 @@ func (s *UserVolumeConfigV1Alpha1) Validate(validation.RuntimeMode, ...validatio
 		validationErrors = errors.Join(validationErrors, errors.New("name can only contain lowercase and uppercase ASCII letters, digits, and hyphens"))
 	}
 
-	extraWarnings, extraErrors := s.ProvisioningSpec.Validate(true)
+	if s.VolumeType != nil && !slices.Contains(
+		[]block.VolumeType{block.VolumeTypePartition, block.VolumeTypeDirectory},
+		*s.VolumeType,
+	) {
+		validationErrors = errors.Join(validationErrors, fmt.Errorf("unsupported volume type: %s", s.VolumeType))
```
smira: please add a validation unit-test for it (we already have validation unit-test)

pkg/machinery/config/types/block/user_volume_config.go
```diff
@@ -168,7 +181,22 @@ func (s *UserVolumeConfigV1Alpha1) Validate(validation.RuntimeMode, ...validatio
...
+	}
+
+	provisioningSpecRequired := true
+	if s.VolumeType != nil {
```
smira, line 196: if volume type is a directory, we should not have provisioning instructions at all? they are not only not required, but it should be empty?

pkg/machinery/config/types/block/user_volume_config.go
smira: can you add here please another example for directory volumes?

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -343,6 +363,26 @@ func (ctrl *UserVolumeConfigController) handleUserVolumeConfig(
...
+	v.TypedSpec().Type = block.VolumeTypeDirectory
+	v.TypedSpec().Locator.Match = labelVolumeMatch(volumeID)
+	v.TypedSpec().Mount = block.MountSpec{
+		TargetPath:          userVolumeConfig.Name(),
+		ParentID:            constants.UserVolumeMountPoint,
+		SelinuxLabel:        constants.EphemeralSelinuxLabel,
+		FileMode:            0o755,
+		UID:                 0,
+		GID:                 0,
+		ProjectQuotaSupport: userVolumeConfig.Filesystem().ProjectQuotaSupport(),
```
smira: this probably should be removed, doesn't make sense for directories

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -343,6 +363,26 @@ func (ctrl *UserVolumeConfigController) handleUserVolumeConfig(
 	return nil
 }
 
+func (ctrl *UserVolumeConfigController) handleDirectoryUserVolumeConfig(
+	userVolumeConfig configconfig.UserVolumeConfig,
+	v *block.VolumeConfig,
+	volumeID string,
+) error {
+	v.TypedSpec().Type = block.VolumeTypeDirectory
+	v.TypedSpec().Locator.Match = labelVolumeMatch(volumeID)
```
smira: this should be removed as well

pkg/machinery/config/types/block/user_volume_config.go
smira: so it should be all empty except for the type, every other field should be empty

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -301,6 +301,26 @@ func (ctrl *UserVolumeConfigController) handleUserVolumeConfig(
...
+) error {
+	switch userVolumeConfig.Type().ValueOr(block.VolumeTypePartition) {
+	case block.VolumeTypePartition:
+		return ctrl.handlePartitionUserVolumeConfig(userVolumeConfig, v, volumeID)
+
+	case block.VolumeTypeDirectory:
+		return ctrl.handleDirectoryUserVolumeConfig(userVolumeConfig, v, volumeID)
+
+	case block.VolumeTypeDisk, block.VolumeTypeTmpfs, block.VolumeTypeSymlink, block.VolumeTypeOverlay:
+		fallthrough
```
Orzelius, line 314: This case doesn't seem to be used and can just be removed

shanduur: I added it due to the [exhausting linter](https://golangci-lint.run/docs/linters/configuration/#exhaustive), which will complain if we don't check all values. I don't want to add `//nolint:exhaustive` here.

smira: it has a benefit that if we add new volume types, linter will alert us on updating this switch one way or another

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -343,6 +363,24 @@ func (ctrl *UserVolumeConfigController) handleUserVolumeConfig(
 	return nil
 }
 
+func (ctrl *UserVolumeConfigController) handleDirectoryUserVolumeConfig(
```
smira, line 367: let's please add a unit-test for this too

internal/app/machined/pkg/controllers/block/user_volume_config.go
Orzelius: Ah I see. Thank you

shanduur:
We pass `/var/mnt` to kubelet as read-only, need some different handling of mount here - e.g. with bind?

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -687,6 +767,37 @@ func (ctrl *MountController) handleDiskUnmountOperation(
 	return nil
 }
 
+// TODO: @shanduur - fix this.
```
smira: do we need to fix this?

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -565,6 +567,82 @@ func (ctrl *MountController) handleDiskMountOperation(
...
+	rootPath string,
+	target string,
+	mountRequest *block.MountRequest,
+	volumeStatus *block.VolumeStatus,
+) error {
+	_, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+
+	// mount hasn't been done yet
+	if !ok {
+		// TODO:
```
smira: do we need to implement this? 

not sure why we have `/var/mnt` at this low level?

internal/app/machined/pkg/controllers/block/mount.go
shanduur: Ay, fixed it on Friday, and forgot to remove the comment. Thanks for catching this!

internal/app/machined/pkg/controllers/block/mount.go
shanduur: Same as the above, forgot to delete my notes 🤦🏻

pkg/machinery/config/types/block/user_volume_config.go
```diff
@@ -68,6 +73,14 @@ type UserVolumeConfigV1Alpha1 struct {
 	//     lowercase and uppercase ASCII letters, digits, and hyphens.
 	MetaName string `yaml:"name"`
 	//   description: |
+	//     Volume type.
+	//   values:
+	//     - partition
+	//     - bind
```
smira: that would be less confusing for users:

```suggestion
	//     - directory
```

pkg/machinery/resources/block/volumetype.go
```diff
@@ -17,4 +17,5 @@ const (
 	VolumeTypeDirectory                   // directory
 	VolumeTypeSymlink                     // symlink
 	VolumeTypeOverlay                     // overlay
+	VolumeTypeBind                        // bind
```
smira: Merge into `VolumeTypeDirectory` with non-empty `MountSpec`, so that mount controller would pick it up and bind-mount the directory as expected?

smira: The idea: volumeType == directory, but `mountSpec.bindTarget` specifies we should bind mount over target (in this case, over itself)

pkg/machinery/config/types/block/user_volume_config.go
```diff
@@ -168,25 +192,53 @@ func (s *UserVolumeConfigV1Alpha1) Validate(validation.RuntimeMode, ...validatio
...
 
-	extraWarnings, extraErrors = s.EncryptionSpec.Validate()
-	warnings = append(warnings, extraWarnings...)
-	validationErrors = errors.Join(validationErrors, extraErrors)
+		extraWarnings, extraErrors = s.EncryptionSpec.Validate()
+		warnings = append(warnings, extraWarnings...)
+		validationErrors = errors.Join(validationErrors, extraErrors)
+
+	case block.VolumeTypeBind:
+		if !s.ProvisioningSpec.IsZero() {
```
smira, line 216: for directory volumes, don't allow non-empty (non-zero) encryption & filesystem specs

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -687,6 +763,36 @@ func (ctrl *MountController) handleDiskUnmountOperation(
...
+	mountCtx, ok := ctrl.activeMounts[mountRequest.Metadata().ID()]
+	if !ok {
+		return nil
+	}
+
+	if err := mountCtx.unmounter(); err != nil { // TODO: filter out error is not mounted
+		return fmt.Errorf("failed to unmount %q: %w", mountRequest.Metadata().ID(), err)
+	}
+
+	if err := os.RemoveAll(mountCtx.point.Source()); err != nil {
```
smira: don't clear user volumes on unmount, this might be unexpected

pkg/machinery/config/types/block/user_volume_config.go
shanduur: Fixed.

pkg/machinery/resources/block/volumetype.go
shanduur: Done.

PR #12083, opened 2025-10-27
feat: bump deps

frezbo:
Bump dependencies.


pkg/machinery/constants/constants.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	// DefaultKernelVersion is the default Linux kernel version.
-	DefaultKernelVersion = "6.17.4-talos"
+	DefaultKernelVersion = "6.17.5-talos"
```
smira, line 17: `hack/release.toml` not updated

PR #12101, opened 2025-10-29
fix: userspace wireguard handling

smira:
This is a fix for the wrong fix in #11204, which was wrong in two ways:

* the ldflags -X override had a wrong variable name, so it had no effect
* but the above even if it worked, only covered "management" part of things, while `wgctrl-go` which configures things still has a hardcoded location of `/var/run/`.

So the fix is two ways:

* replace the location where the socket is created properly
* use updated forked wgctrl-go which looks in both locations

This keeps all fixes of #11204 - `talosctl cluster create` siderolink agent works properly with `wg` on the host, and Talos uses proper location.

Before the fix the location was actually `/var/run` and it randomly failed depending on the race condition of Talos booting up and managing `/var`.

### DIFF
--- a/internal/app/machined/pkg/controllers/network/link_alias_spec.go
+++ b/internal/app/machined/pkg/controllers/network/link_alias_spec.go
@@ -107,8 +107,6 @@ func (ctrl *LinkAliasSpecController) Run(ctx context.Context, r controller.Runti
 				continue
 			}
 
-			logger.Debug("checking link for alias", zap.String("link", link.Attributes.Name), zap.Any("link", link))
-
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/network/link_alias_spec.go
```diff
@@ -107,8 +107,6 @@ func (ctrl *LinkAliasSpecController) Run(ctx context.Context, r controller.Runti
 				continue
 			}
 
-			logger.Debug("checking link for alias", zap.String("link", link.Attributes.Name), zap.Any("link", link))
```
smira, line 110: this was a debug log which was too noisy, dropped it

PR #12104, opened 2025-10-30
feat: add full disk volumes

### LINKED ISSUE #11847: [talos v1.11.0] UserVolume on Logical Volume(LVM) does not work
### Description

machine config:
```
apiVersion: v1alpha1
kind: UserVolumeConfig
name: nova
provisioning:
  diskSelector: 
    match: disk.dev_path == '/dev/vg-nova/lv-nova'
  minSize: 100GB
```

### Logs

```
10.10.10.10: user: warning: [2025-09-17T12:33:08.887711414Z]: [talos] volume status {"component": "controller-runtime", "controller": "block.VolumeManagerController", "volume": "u-nova", "phase": "failed -> failed", "error": "error probing disk: open /dev/dm-0p1: no such file or directory", "location": "/dev/dm-0p1", "parentLocation": "/dev/dm-0"}
```

```
# lvdisplay /dev/vg-nova/lv-nova
  --- Logical volume ---
  LV Path                /dev/vg-nova/lv-nova
  LV Name                lv-nova
  VG Name                vg-nova
  LV UUID                zR5ofm-fIhJ-frx8-G3Zl-ZAT3-eqpY-i7ohta
  LV Write Access        read/write
  LV Creation host, time ulegcpcibarium, 2025-09-17 08:13:09 +0000
  LV Status              available
  # open                 0
  LV Size                8.73 TiB
  Current LE             2289267
  Segments               1
  Allocation             inherit
  Read ahead sectors     auto
  - currently set to     768
  Block device           251:0

```

```
# talosctl  ls -lH -t l /dev/vg-nova/
NODE            MODE         UID   GID   SIZE(B)   LASTMOD          LABEL                           NAME
10.10.10.10   Lrwxrwxrwx   0     0     7 B       37 minutes ago   system_u:object_r:device_t:s0   lv-nova -> ../dm-0

 # talosctl get disk dm-0 -o yaml
node: 10.10.10.10
metadata:
    namespace: runtime
    type: Disks.block.talos.dev
    id: dm-0
    version: 4
    owner: block.DisksController
    phase: running
    created: 2025-09-17T12:03:57Z
    updated: 2025-09-17T12:03:57Z
spec:
    dev_path: /dev/dm-0
    size: 9601881735168
    pretty_size: 9.6 TB
    io_size: 65536
    sector_size: 512
    readonly: false
    cdrom: false
    bus_path: /virtual
    sub_system: /sys/class/block
    secondary_disks:
        - nvme1n1
        - nvme2n1
        - nvme3n1
    symlinks:
        - /dev/disk/by-diskseq/27
        - /dev/disk/by-id/dm-name-vg--nova-lv--nova
        - /dev/disk/by-id/dm-uuid-LVM-5ZT2lilTazxDeYJIb21NnpKlfYyK6zH5zR5ofmfIhJfrx8G3ZlZAT3eqpYi7ohta

# talosctl get discoveredvolume dm-0  -o yaml
node: 10.10.10.10
metadata:
    namespace: runtime
    type: DiscoveredVolumes.block.talos.dev
    id: dm-0
    version: 4
    owner: block.DiscoveryController
    phase: running
    created: 2025-09-17T12:03:57Z
    updated: 2025-09-17T12:35:17Z
spec:
    dev_path: /dev/dm-0
    type: disk
    device_path: /sys/devices/virtual/block/dm-0
    size: 9601881735168
    pretty_size: 9.6 TB
    sector_size: 512
    io_size: 65536
    name: ""
```

### Environment

- Talos version: v1.11.0
- Kubernetes version: v1.33.4
- Platform: metal

shanduur:
`volumeType` in UserVolumeConfig can be set to `disk`.
When set to `disk`, a full block device is used for the volume.

When `volumeType = "disk"`:
- Size specific settings are not allowed in the provisioning block (`minSize`, `maxSize`, `grow`).

Closes #11847

### DIFF
--- a/internal/app/machined/pkg/controllers/block/internal/volumes/locate.go
+++ b/internal/app/machined/pkg/controllers/block/internal/volumes/locate.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/google/cel-go/cel"
 	"github.com/siderolabs/gen/value"
 	"github.com/siderolabs/gen/xerrors"
 	"github.com/siderolabs/go-blockdevice/v2/partitioning"
@@ -44,8 +45,18 @@ func LocateAndProvision(ctx context.Context, logger *zap.Logger, volumeContext M
 
 	// attempt to discover the volume
 	for _, dv := range volumeContext.DiscoveredVolumes {
-		matchContext := map[string]any{
-			"volume": dv,
+		var locator *cel.Env
+
+		matchContext := map[string]any{}
+
+		switch volumeType { //nolint:exhaustive // we do not need to repeat exhaustive check here
+		case block.VolumeTypeDisk:
+			locator = celenv.DiskLocator()
+
+		case block.VolumeTypePartition:
+			locator = celenv.VolumeLocator()
+
+			matchContext["volume"] = dv
 		}
 
 		// add disk to the context, so we can use it in CEL expressions
@@ -63,7 +74,7 @@ func LocateAndProvision(ctx context.Context, logger *zap.Logger, volumeContext M
 			}
 		}
 
-		matches, err := volumeContext.Cfg.TypedSpec().Locator.Match.EvalBool(celenv.VolumeLocator(), matchContext)
+		matches, err := volumeContext.Cfg.TypedSpec().Locator.Match.EvalBool(locator, matchContext)
 		if err != nil {
 			return fmt.Errorf("error evaluating volume locator: %w", err)
 		}
@@ -125,6 +136,10 @@ func LocateAndProvision(ctx context.Context, logger *zap.Logger, volumeContext M
 
 	if len(matchedDisks) == 0 {
 		return fmt.Errorf("no disks matched selector for volume")
+	}
+
+	if volumeType == block.VolumeTypeDisk && len(matchedDisks) > 1 {
+		return fmt.Errorf("multiple disks matched selector for disk volume; matched disks: %v", matchedDisks)
 	}
 
 	logger.Debug("matched disks", zap.Strings("disks", matchedDisks))
--- a/internal/integration/api/volumes.go
+++ b/internal/integration/api/volumes.go
@@ -265,32 +265,7 @@ func (suite *VolumesSuite) TestLVMActivation() {
 
 	suite.Require().Contains(stdout, "Logical volume \"lv1\" created.")
 
-	defer func() {
-		suite.T().Logf("removing LVM volumes %s/%s", node, nodeName)
-
-		deletePodDef, err := suite.NewPrivilegedPod("pv-destroy")
-		suite.Require().NoError(err)
-
-		deletePodDef = deletePodDef.WithNodeName(nodeName)
-
-		suite.Require().NoError(deletePodDef.Create(suite.ctx, 5*time.Minute))
-
-		defer deletePodDef.Delete(suite.ctx) //nolint:errcheck
-
-		if _, _, err := deletePodDef.Exec(
-			suite.ctx,
-			"nsenter --mount=/proc/1/ns/mnt -- vgremove --nolocking --yes vg0",
-		); err != nil {
-			suite.T().Logf("failed to remove pv vg0: %v", err)
-		}
-
-		if _, _, err := deletePodDef.Exec(
-			suite.ctx,
-			fmt.Sprintf("nsenter --mount=/proc/1/ns/mnt -- pvremove --nolocking --yes %s", userDisksJoined),
-		); err != nil {
-			suite.T().Logf("failed to remove pv backed by volumes %s: %v", userDisksJoined, err)
-		}
-	}()
+	defer suite.deleteLVMVolumes(node, nodeName, userDisksJoined)
 
 	suite.T().Logf("rebooting node %s/%s", node, nodeName)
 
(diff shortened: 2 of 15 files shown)


internal/integration/api/volumes.go
```diff
@@ -650,8 +650,189 @@ func (suite *VolumesSuite) TestUserVolumesPartition() {
 		})
 }
 
-// TestUserVolumesBind performs a series of operations on user volumes (bind type): creating, destroying, verifying, etc.
-func (suite *VolumesSuite) TestUserVolumesBind() {
+// TestUserVolumesDisk performs a series of operations on user volumes (disk type): creating, destroying, verifying, etc.
+func (suite *VolumesSuite) TestUserVolumesDisk() {
```
shanduur, line 656: There needs to be another test for LVM instead of raw disk.

shanduur: I'll add them in a follow up, let's keep the scope of this PR to `volumeType = "disk"` only. I had that test partially written, but it looked bad, and I want to spend a little bit more time on it (e.g. implement LVM controller), to make the tests less hacky.

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -364,6 +367,49 @@ func (ctrl *UserVolumeConfigController) handlePartitionUserVolumeConfig(
...
+	volumeID string,
+) error {
+	diskSelector, ok := userVolumeConfig.Provisioning().DiskSelector().Get()
+	if !ok {
+		// this shouldn't happen due to validation
+		return fmt.Errorf("disk selector not found for volume %q", volumeID)
+	}
+
+	v.TypedSpec().Type = block.VolumeTypeDisk
+	v.TypedSpec().Locator.Match = labelVolumeMatch(volumeID)
```
smira: I think this is unused/should be dropped? as it's a disk, we can't match on label

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -364,6 +367,49 @@ func (ctrl *UserVolumeConfigController) handlePartitionUserVolumeConfig(
...
+	v *block.VolumeConfig,
+	volumeID string,
+) error {
+	diskSelector, ok := userVolumeConfig.Provisioning().DiskSelector().Get()
+	if !ok {
+		// this shouldn't happen due to validation
+		return fmt.Errorf("disk selector not found for volume %q", volumeID)
+	}
+
+	v.TypedSpec().Type = block.VolumeTypeDisk
```
smira: one thought - should we fail in the `VolumeManager` controller if the disk selector spec for VolumeTypeDisk  matches more than a single disk? as this might lead to undefined behavior e.g. on reboot?

internal/app/machined/pkg/controllers/block/user_volume_config.go
shanduur: We need some kind of matcher, but I think we can reuse disk locator here? This should be sufficient.

internal/app/machined/pkg/controllers/block/user_volume_config.go
shanduur: Right, I will add condition for this!

internal/app/machined/pkg/controllers/block/user_volume_config.go
smira: I'm not sure, a good way is to reboot Talos and see what it does?

shanduur: I haven't seen any issues, however it goes through the provisioning flow each time - reusing disk locator should cut that, right?

shanduur: https://github.com/shanduur/talos/blob/lvm-volumes/internal/app/machined/pkg/controllers/block/internal/volumes/locate.go#L45-L83

internal/app/machined/pkg/controllers/block/user_volume_config.go
shanduur: https://github.com/shanduur/talos/blob/lvm-volumes/internal/app/machined/pkg/controllers/block/internal/volumes/locate.go#L130-L132

internal/app/machined/pkg/controllers/block/user_volume_config.go
```diff
@@ -364,6 +367,49 @@ func (ctrl *UserVolumeConfigController) handlePartitionUserVolumeConfig(
...
+	v.TypedSpec().Locator.Match = diskSelector
```
smira: this does work, but by chance a bit (with a specific form of disk selector).

we have a problem here, the disk selector is evaluated in:

https://github.com/shanduur/talos/blob/0d8c0228615d4353f9f5ded9d8b682200ee7e342/pkg/machinery/cel/celenv/celenv.go#L34-L61

while volume locator in:

https://github.com/shanduur/talos/blob/0d8c0228615d4353f9f5ded9d8b682200ee7e342/pkg/machinery/cel/celenv/celenv.go#L62-L86

So anything which does simple `disk.<something>` works as they both have same. 

But e.g. `system_disk` is not available in volume locator, also volume locator runs for partitions as well, so it might match it by accident.

So I think we need a separate `locator.DiskMatch` or something close to that for disk volumes. 

Open to any other ideas as well

shanduur: Let me know if the changes I did make any sense. I've added different `*cel.Env` that is selected based on volume type: https://github.com/siderolabs/talos/pull/12104/files#diff-29f79fc583c81df9b1167eb597fb3642587aa952b153ee5c3c5190005577bd6bR54-R61

Orzelius (approved):
Very cool
Love the thorough tests!

internal/app/machined/pkg/controllers/block/internal/volumes/locate.go
```diff
@@ -63,7 +74,7 @@ func LocateAndProvision(ctx context.Context, logger *zap.Logger, volumeContext M
 			}
 		}
 
-		matches, err := volumeContext.Cfg.TypedSpec().Locator.Match.EvalBool(celenv.VolumeLocator(), matchContext)
+		matches, err := volumeContext.Cfg.TypedSpec().Locator.Match.EvalBool(locator, matchContext)
```
smira, line 77: I think there is also `system_disk: boolean` in the DiskLocator environment

smira (approved):

internal/app/machined/pkg/controllers/block/internal/volumes/locate.go
shanduur: `system_disk` is invalid here, as UserVolume with type `disk` cannot take control over the `system_disk`.

PR #12106, opened 2025-10-30
feat: implement bond multi-doc configuration

### LINKED ISSUE #10960: Network Config: `BondConfig`
### Bonds

Bond configuration document creates and configures the bond, it also embeds `LinkConfig` fields.

```yaml
version: v1alpha1
kind: BondConfig
name: bond0 # name of the link
links: # links which are bonded, NB: order is important
  - int0
  - int1
settings: # Note: more settings follow, see existing configuration
  mode: 802.3ad
  lacpRate: fast
  ...
# embedded from LinkConfig, optional
mtu: 1500
up: true
addresses: [...]
routes: [...]
```

`BondConfig` conflicts with `VLANConfig`, `BridgeConfig`, `WireguardConfig` and `LinkConfig` with the same `name`.

### LINKED ISSUE #12048: [FR] Support for bond arpIPTarget
## Feature Request
Hi,

In the network config doc : 
```
arpIPTarget	[]string	A bond option.
Please see the official kernel documentation.
Not supported at the moment.
```

I'd love to see that supported.

### Description

I'm trying to setup an active-backup bond on my nodes, so I can connect them to two switches that sadly don't support any fancy MLAG.
That works fine with miimon, which is a good start, but my understanding is that could still lead to weird behavior when the switches are rebooting if the link is still up while the switch is not yet connected to it's uplink, which arpIPTarget could detect.

It'd be great to add support for those config keys.

Thanks

smira:
Also expand internal bond configuration to cover missing fields. They are not going to be exposed in legacy configuration.

Fixes #10960

Fixes #12048


pkg/machinery/config/types/network/bond.go
```diff
@@ -0,0 +1,352 @@
...
+	//   description: |
+	//     Names of the parent links (interfaces) on which the bond link will be created.
+	//     Link aliases can be used here as well.
+	//   examples:
+	//    - value: >
+	//       []string{"enp0s3", "enp0s8"}
+	//   schemaRequired: true
+	ParentLinks []string `yaml:"parentLinks,omitempty"`
+	//   description: |
+	//     Bond ID to be used for the Bond link.
```
frezbo: Bond type

pkg/machinery/resources/network/link.go
```diff
@@ -28,30 +28,236 @@ type VLANSpec struct {
...
+	if spec.MissedMax != other.MissedMax {
+		return false
+	}
+
+	return true
+}
+
+// IsZero checks if the BondMasterSpec is zero value.
+//
+//nolint:gocyclo,cyclop
```
frezbo, line 231: I wonder if we can do some codegen for this, if new fields ever pop up

smira: I tried to find something, but I failed to... still looking if I can drop this or rework

pkg/machinery/config/types/network/bond.go
smira: thanks, fixed

pkg/machinery/resources/network/link.go
smira: I haven't found any good replacement.

PR #12107, opened 2025-10-30
fix: improve OOM controller stability and make test strict on false positives

### LINKED ISSUE #12077: improve cooldown of OOM trigger
Once the PSI goes down, but absolute value stays high, stop doing OOM kills.

dsseng:
- Add d_* PSI derivative values to the trigger expression context
- Only trigger OOM action while PSI is rising
- Make OOM test fail if controller kills a cgroup without stress-ng

Fixes #12077


smira (approved):
🆒

frezbo (approved):

smira:
I don't know if we should relax our assertions a bit by allowing non-stressng kills?

smira (approved):

Issue #11992, opened 2025-10-08
VolumeConfig resources are processed after RawVolumeConfig resources


Drakonis:
### Description

When both a `VolumeConfig` resource modifying `EPHEMERAL` and a `RawVolumeConfig` resource are applied at machine creation, the raw volume is created first, then `EPHEMERAL` is created, ensuring that `EPHEMERAL` cannot be created if the volume preceding it uses all storage, as well as preventing the resizing of the newly created raw volume due to `EPHEMERAL`'s location in the disk partitions.

`VolumeConfig` should be handled first to avoid this situation.
These manifests were used to partition the machines
```
---
apiVersion: v1alpha1
kind: VolumeConfig
name: EPHEMERAL
provisioning:
    grow: false
    maxSize: 50GiB
---
apiVersion: v1alpha1
kind: RawVolumeConfig
name: local-data
provisioning:
    diskSelector:
        match: system_disk
    grow: true
    maxSize: 180GiB
```
When first initializing with a `VolumeConfig` resource modifying `EPHEMERAL` results in the following layout:
<img width="591" height="221" alt="Image" src="https://github.com/user-attachments/assets/9ed0b8bc-d8d3-4a31-a0fd-7a1c2ff6bc0a" />

Then with a `RawVolumeConfig` resource setting up `local-data` results in:
<img width="588" height="243" alt="Image" src="https://github.com/user-attachments/assets/2b51809f-2c8e-4529-b69d-a48cc1e8db47" />

When both resources are used, following incorrect layout is generated:
<img width="594" height="245" alt="Image" src="https://github.com/user-attachments/assets/b426b5ad-3739-416a-864f-a3c42178c14a" />

### Logs

While partitioning using a `RawVolumeConfig` with a size greater than the remaining available space will result in `EPHEMERAL` not being created, as demonstrated here:
```
user: warning: [2025-10-08T21:31:35.679923253Z]: [talos] volume status {"component": "controller-runtime", "controller": "block.VolumeManagerController", "volume": "EPHEMERAL", "phase": "failed -> failed", "error": "no disks matched for volume (1 matched selector): 1 have not enough space, 0 have wrong format, 0 have other issues"}
```
```
NODE                   NAMESPACE   TYPE           ID                                  VERSION   TYPE        PHASE    LOCATION    SIZE
talos-infra-worker-4   runtime     VolumeStatus   /etc/cni                            2         overlay     ready
[...]

PR #12122, opened 2025-11-03
Fix volume creation race + refactor (User)VolumeConfigController

laurazard:
# Pull Request

Fixes https://github.com/siderolabs/talos/issues/11992

## What? (description)

Merges `(System)VolumeConfig` and `UserVolumeConfig` controllers into a single controller, w/ extensive refactors:
- reuse "transformer" pattern (machine config -> `VolumeConfig`) for system+user volumes
- reuses `handleCustomVolumeConfig` (now `createVolume`) for both as well
- introduces a small internal type `volumeResource` to replace all the arguments passed to `handleCustomConfig`/`createVolume`
- Merged tests from `user_volume_config_test.go` into `volume_config_test.go`
- Removed the `r.StartTrackingOutput`/`safe.CleanupOutputs` logic, and reused the custom logic from the `UserVolumeConfig` controller for keeping track of `VolumeConfig`s/removing unused ones (added a label, `block.SystemVolumeLabel` for tracking)

Additionally, a small, internal `volumeConfigBuilder` convenience wrapper for building `VolumeConfig` resource modifier funcs is added:

```go
newVolumeConfigBuilder().
	WithType(block.VolumeTypeDirectory).
	WithMount(block.MountSpec{
		TargetPath:   userVolumeConfig.Name(),
		ParentID:     constants.UserVolumeMountPoint,
		SelinuxLabel: constants.EphemeralSelinuxLabel,
		FileMode:     0o755,
		UID:          0,
		GID:          0,
		BindTarget:   pointer.To(userVolumeConfig.Name()),
	})
```

## Why? (reasoning)

Merging `VolumeConfig` and `UserVolumeConfig` controllers removes the race condition between system and user volume config creation, and these controllers are quite similar already.

## Acceptance


laurazard:
`TestLVMActivation` is failing 🤔 I wonder if it's my changes

smira:
> `TestLVMActivation` is failing 🤔 I wonder if it's my changes

that one is flaky if it fails saying 'locking failed'...

internal/app/machined/pkg/controllers/block/user_volumes.go
```diff
@@ -0,0 +1,231 @@
...
+
+var userVolumeTransformers = []volumeConfigTransformer{
+	userVolumeTransformer,
+	rawVolumeTransformer,
+	existingVolumeTransformer,
+	swapVolumeTransformer,
+}
+
+var (
+	userVolumeTransformer = func(c configconfig.Config) ([]volumeResource, error) {
```
smira, line 34: dumb question, what is the benefit of this vs.

```go
func userVolumeTransformer(c configconfig.Config) ([]volumeResource, error)
```

smira: I don't think we need this right now, but we could also move these functions into `internal/...` sub-package, make them public, so we could unit-test them in isolation as well 

this might provide some better grouping, remove some code from this package into smaller sub-packages

smira (approved):
🆒

internal/app/machined/pkg/controllers/block/user_volumes.go
laurazard: I thought about doing that too! I think I have a branch locally with a change similar to that + some unit tests, but then I decided I had to draw a line somewhere and leave that for another PR 😅

I'll push those changes anyway so I don't forget.

laurazard: > dumb question, what is the benefit of this vs.

No real benefit (at least, I don't think there is), just style – I'm happy to change it to either form. I think had it this way initially to kind of emphasize the fact that these should be grouped together. I also thought about adding a dummy `type ConfigResourceTransformer struct{}` that just does these, which is good for grouping, but I don't love the unnecessary bloat.

Moving these to another package like you suggested is a good approach, so I'll do that (in a separate PR, if that's alright).

laurazard: Opened https://github.com/siderolabs/talos/pull/12196

smira (approved):

PR #12124, opened 2025-11-03
fix: update YAML library

smira:
Update COSI, and stop using a fork of `gopkg.in/yaml.v3`, now we use new supported for of this library.

Drop `MarshalYAMLBytes` for the machine config, as we actually marshal config as a string, and we don't need this at all.

Make `talosctl` stop doing hacks on machine config for newer Talos, keep hacks for backwards compatibility.


cmd/talosctl/cmd/talos/output/yaml.go
```diff
@@ -40,7 +40,11 @@ func (y *YAML) WriteHeader(definition *meta.ResourceDefinition, withEvents bool)
 
 // WriteResource implements output.Writer interface.
 func (y *YAML) WriteResource(node string, r resource.Resource, event state.EventType) error {
-	if r.Metadata().Type() == config.MachineConfigType {
+	if r.Metadata().Type() == config.MachineConfigType && r.Metadata().Annotations().Empty() {
+		// use a temporary wrapper to adjust YAML marshaling
```
smira, line 44: this change is going to break Talos 1.11 talosctl against Talos 1.12 when using YAML + machine config, but there is backwards compatibility (Talos 1.12 talosctl will talk to Talos 1.11)

Unix4ever (approved):
Awesome to drop this bandaid finally :)

Will need to do the same in all dependencies.

PR #12157, opened 2025-11-06
fix: stop attaching to tearing down mount parents

smira:
This bug showed up as a random deadlock on kubelet restart (might be any other service though).

With a chain of mount requests, like `/var/log` -> `/var/log/containers`, there was a chance that a new generation of mount requests might try to pick up a tearing down parent of the previous generation leading to a deadlock when the mount can't proceed for the parent.


frezbo (approved):
wonder if we can have tests to catch these

shanduur (approved):

Issue #10962, opened 2025-05-08
Network Config: `BridgeConfig`


smira:
### Bridges

Bridge configuration document creates and configures the bridge, it also embeds `LinkConfig` fields.

```yaml
version: v1alpha1
kind: BridgeConfig
name: bridge0 # bridge name
links: # links which are bridged
  - en0sp2
  - int0.17
  - bond1
macAddress: 00:aa:... # optional, set the MAC
stp: # Spanning Tree Protocol settings
  enabled: true # defaults to false
vlan:
  filtering: true # defaults to false
# embedded from LinkConfig, optional
mtu: 1500
up: true
addresses: [...]
routes: [...]
```

`BridgeConfig` conflicts with `BondConfig`, `VLANConfig`, `WireguardConfig` and `LinkConfig` with the same `name`.

PR #12158, opened 2025-11-06
feat: implement bridge multi-document config

smira:
Fixes https://github.com/siderolabs/talos/issues/10962


smira:
On top of #12106 to minimize conflicts

PR #12169, opened 2025-11-08
fix: only set default bootloader if none is set

frezbo:
Only set a default bootloader if none is set, secureboot still always return `sd-boot`.


frezbo:
> I feel like something is wrong in this code and we should rewrite it...

agree, it all seems a bit messy, too much conditionals, I tried cleaning up how we select the bootloader for iso and decided to leave it as is for now

PR #12199, opened 2025-11-11
feat: implement time and resolvers multi-doc configuration

### LINKED ISSUE #10954: Network Config: `ResolverConfig`
### Resolvers

```yaml
version: v1alpha1
kind: ResolverConfig
resolvers:
  - address: 1.1.1.1
    protocol: DNS # default
  - address: 8.8.8.8
    hostname: dns.google
    protocol: DoT # DNS over TLS
searchDomains:
  disableDefault: false # default value
  extra: # optional
    - example.org
```

> [!WARNING]  
> We don’t have DNS over TLS implemented yet

- `resolvers` is a list of DNS resolvers to use, `protocol` can be omitted, in that case only `address` should be specified (DNS over TLS is mentioned here for completeness only, see https://github.com/siderolabs/talos/issues/9999)
- `searchDomains` configures search domains in `/etc/resolv.conf`, they don’t affect DNS resolution, but affect client resolver inside a program, this section is optional
    - `disableDefault` disables default extraction of the domain of the hostname (if specified as FQDN)
    - `extra` specifies a list of additional entries to add to search domains

**Replaces** `machine.network.nameservers`, `machine.network.searchDomains` , `machine.network.disableSearchDomain`

### LINKED ISSUE #10955: Network Config: `TimeSyncConfig`
### Time Sync

```yaml
version: v1alpha1
kind: TimeSyncConfig
enabled: true # default value, set to false to disable
bootTimeout: 2m # unset means no boot timeout
ntp: # mutually exclusive with ptp
  servers:
    - time.cloudflare.com
ptp: # mutually exclusive with ntp
  devices:
    - /dev/ptp0
```

> [!WARNING]  
> Time sync config might go to its own namespace `time`, but documented here as it’s still closely related to network, and it’s critical for initial boot.

- `enabled` defaults to `true`, can be disabled e.g. to run your own NTP server
- `bootTimeout` is used to not to block the boot forever if the time can’t be synced, after specified time is assumed to be in sync despite no sync being done
- `ntp` and `ptp` specify either network or local ways to sync the time.

**Replaces** `machine.time` (conflicts)

smira:
Fixes #10954

Fixes #10955


smira:
> don;t we need a default generated config for `gen config`?

timeservers/nameservers are not there by default which is good. 

I forgot hack/release.toml note, it's the last network config piece

Issue #12090, opened 2025-10-28
change `talosctl reset --system-labels-to-wipe` to wipe partitions


smira:
This would allow to re-create partitions with Omni or without Omni.

pranav767:
Are there any code changes expected here/ just the cmd doc change
Can see
```

--system-labels-to-wipe strings           if set, just wipe selected system disk partitions by label but keep other partitions intact

```
Don't feel that any code change might be required (dont see any system-labels-to-wipe for omni/non omni contraint)

smira:
> Are there any code changes expected here/ just the cmd doc change Can see

Omni here is an example. Talos will not only erase partitions, but also drop partitions themselves. This allows e.g. EPHEMERAL to be relocated after wipe with new machine config.

PR #12207, opened 2025-11-12
feat: drop partitions on reset with system partitions wipe

laurazard:
# Pull Request

Closes https://github.com/siderolabs/talos/issues/12090

## What? (description)

When resetting+wiping system partitions (`talosctl reset --system-labels-to-wipe ...`), also drop partitions.

## Why? (reasoning)

This enables usecases such as relocating EPHEMERAL, etc. as described in https://github.com/siderolabs/talos/issues/12090.

## Acceptance

### DIFF
--- a/internal/pkg/partition/wipe.go
+++ b/internal/pkg/partition/wipe.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/siderolabs/go-blockdevice/v2/block"
+	"github.com/siderolabs/go-blockdevice/v2/partitioning/gpt"
 
 	blockres "github.com/siderolabs/talos/pkg/machinery/resources/block"
 )
@@ -19,6 +20,8 @@ type VolumeWipeTarget struct {
 	label string
 
 	parentDevName, devName string
+
+	partitionIndex int // partitionIndex is 1-based, but decrement before using
 }
 
 // VolumeWipeTargetFromVolumeStatus creates a new VolumeWipeTarget from a VolumeStatus.
@@ -30,9 +33,10 @@ func VolumeWipeTargetFromVolumeStatus(vs *blockres.VolumeStatus) *VolumeWipeTarg
 	}
 
 	return &VolumeWipeTarget{
-		label:         vs.Metadata().ID(),
-		parentDevName: parentDevName,
-		devName:       vs.TypedSpec().Location,
+		label:          vs.Metadata().ID(),
+		parentDevName:  parentDevName,
+		devName:        vs.TypedSpec().Location,
+		partitionIndex: vs.TypedSpec().PartitionIndex,
 	}
 }
 
@@ -45,9 +49,10 @@ func VolumeWipeTargetFromDiscoveredVolume(dv *blockres.DiscoveredVolume) *Volume
 	}
 
 	return &VolumeWipeTarget{
-		label:         dv.TypedSpec().PartitionLabel,
-		parentDevName: parentDevName,
-		devName:       dv.TypedSpec().DevPath,
+		label:          dv.TypedSpec().PartitionLabel,
+		parentDevName:  parentDevName,
+		devName:        dv.TypedSpec().DevPath,
+		partitionIndex: int(dv.TypedSpec().PartitionIndex),
 	}
 }
 
@@ -62,8 +67,9 @@ func (v *VolumeWipeTarget) String() string {
 }
 
 // Wipe implements runtime.PartitionTarget.
+// Asides from wiping the device, Wipe() also drops the partition.
 func (v *VolumeWipeTarget) Wipe(ctx context.Context, log func(string, ...any)) error {
-	parentBd, err := block.NewFromPath(v.parentDevName)
+	parentBd, err := block.NewFromPath(v.parentDevName, block.OpenForWrite())
 	if err != nil {
 		return fmt.Errorf("error opening block device %q: %w", v.parentDevName, err)
 	}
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1253,10 +1253,21 @@ func ResetSystemDiskSpec(_ runtime.Sequence, data any) (runtime.TaskExecutionFun
...
 		for _, target := range in.GetSystemDiskTargets() {
+			// TODO(laurazard): these are only used here (and, in general, runtime.PartitionTarget)
+			// is only used here/for system volumes. Should wipe/drop partition be together?
 			if err = target.Wipe(ctx, logger.Printf); err != nil {
 				return fmt.Errorf("failed wiping partition %s: %w", target, err)
 			}
+
+			if err = target.Drop(ctx, logger.Printf); err != nil {
+				return fmt.Errorf("failed dropping partition %s: %w", target, err)
+			}
```
laurazard: @smira thoughts?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1253,10 +1253,21 @@ func ResetSystemDiskSpec(_ runtime.Sequence, data any) (runtime.TaskExecutionFun
 			return errors.New("unexpected runtime data")
 		}
 
+		// TODO(laurazard): should we have better error handling here?
+		// if we manage to wipe STATE, then get an error wiping EPHEMERAL,
+		// we currently just return the error, and won't delete
+		// StateEncryptionConfig from META.
```
laurazard: And just in general wondering about `ResetSystemDiskSpec` – not related to this PR in particular, but maybe something we can improve later (or not, if it doesn't make sense!)

smira: yes, it's a good question, we could probably do some kind of multiError and track things better. I agree with you that if we managed to wipe STATE in any way, we should drop encryption config from META

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: yes, I think we could merge them, but up to you

smira: I think merging would have a benefit of locking the parent only once

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
laurazard: I'll open a draft so I don't forget 😅

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
laurazard: Yeah, my question was mostly in terms of the interface, if we would ever want to just wipe without dropping. For now, I've merged it.

internal/integration/api/reset.go
```diff
@@ -233,6 +233,37 @@ func (suite *ResetSuite) TestResetWithSpecStateAndUserDisks() {
...
+		Graceful: true,
+		SystemPartitionsToWipe: []*machineapi.ResetPartitionSpec{
+			{
+				Label: constants.EphemeralPartitionLabel,
+				Wipe:  true,
+			},
+		},
+	}, true)
+
+	rtestutils.AssertNoResource[*block.DiscoveredVolume](nodeCtx, suite.T(), suite.Client.COSI, constants.EphemeralPartitionLabel)
```
smira: I _think_ this won't work as the ResetNode above will only return after reboot 

I don't think we can assert on this condition (and I think we already have a test which wipes by system labels)

internal/pkg/partition/wipe.go
```diff
@@ -76,14 +82,54 @@ func (v *VolumeWipeTarget) Wipe(ctx context.Context, log func(string, ...any)) e
...
+	}
+
+	if err := v.dropParentLocked(parentBd, log); err != nil {
+		return fmt.Errorf("error dropping partition: %w", err)
+	}
+
+	return nil
+}
+
+func (v *VolumeWipeTarget) wipeParentLocked(log func(string, ...any)) error {
```
smira: nit, should it be better with `wipeWithParentLocked` (and same for drop), as I read it first as "wipe parent"?

internal/integration/api/reset.go
laurazard: Yeah, I wanted to ask about that – this was my super naive attempt (that I wasn't sure would work), but I think the proper way would be to somehow create this cluster with no config, then wipe, then maintenance-mode do the equivalent of a `talosctl get --insecure discoveredvolumes`.

laurazard: For the record, this did (somehow) work locally, but I was assuming the test wasn't actually testing anything. I could instead change the other test, but regardless if we were testing this we'd want to actually test that it gets dropped. I guess we can also test this later with an actual usecase (like relocating a system volume).

laurazard: I'll drop the test for now – I think that should be fine, we have other tests to at least ensure that the change doesn't break anything.

smira: Yep, I think it's fine to skip the explicit test for now.

It might work due to the fact that EPHEMERAL hasn't been created yet after a reboot, but it's racy

PR #12208, opened 2025-11-12
feat: relative volume size

### LINKED ISSUE #11744: support for specifying relative volume `maxSize`
Imagine a following scenario: I want to deploy Talos to 10 machines which have all a single huge disk, but disk size varies across machines.

I want EPHEMERAL to occupy most of the disk, but not all of it (reserve some space for swap/user volumes, etc.)

So instead of specifying exact maxSize which might vary, support the following:

* percent-based `maxSize`, e.g. use 50% of unallocated space
* grow to maximum size, but leave "X GiB" unallocated

shanduur:
Include percent-based `maxSize`, e.g. use 50% of space.

Closes #11744


website/content/v1.12/reference/configuration/block/rawvolumeconfig.md
```diff
@@ -78,10 +78,10 @@ ProvisioningSpec describes how the volume is provisioned.
 |-------|------|-------------|----------|
 |`diskSelector` |<a href="#RawVolumeConfig.provisioning.diskSelector">DiskSelector</a> |The disk selector expression.  | |
 |`grow` |bool |Should the volume grow to the size of the disk (if possible).  | |
-|`minSize` |ByteSize |The minimum size of the volume.<br><br>Size is specified in bytes, but can be expressed in human readable format, e.g. 100MB. <details><summary>Show example(s)</summary>{{< highlight yaml >}}
+|`minSize` |Size |The minimum size of the volume.<br><br>Size is specified in bytes, but can be expressed in human readable format, e.g. 100MB. <details><summary>Show example(s)</summary>{{< highlight yaml >}}
```
smira: should we update the doc here/provide examples with relative size?

internal/app/machined/pkg/controllers/block/internal/volumes/partition.go
```diff
@@ -63,8 +63,8 @@ func CreatePartition(ctx context.Context, logger *zap.Logger, diskPath string, v
 
 	available := pt.LargestContiguousAllocatable()
 
-	size := volumeCfg.TypedSpec().Provisioning.PartitionSpec.MinSize
```
smira, line 66: how does it work for minSize if we're picking up a disk (before we create a partition?)

smira:
I like it, I probably have a few questions here and there, let's get it backported to 1.12 once it's ready

website/content/v1.12/reference/configuration/block/rawvolumeconfig.md
shanduur: Done!

internal/app/machined/pkg/controllers/block/internal/volumes/partition.go
shanduur: I dropped support for relative `minSize`.

pkg/machinery/config/types/block/volume_config.go
```diff
@@ -230,7 +232,7 @@ func (s ProvisioningSpec) Validate(required bool, sizeSupported bool) ([]string,
 	}
 
 	if sizeSupported {
-		if !s.ProvisioningMinSize.IsZero() && !s.ProvisioningMaxSize.IsZero() {
+		if !s.ProvisioningMinSize.IsZero() && !s.ProvisioningMaxSize.IsZero() && !s.ProvisioningMaxSize.IsRelative() {
```
shanduur, line 235: Not sure if we want to skip the test here, but we don't have a way to pass the data about available space here, right?

smira: yep, exactly, we might be validating on the client machine with `talosctl`.

we have also `RuntimeValidate` interface which runs on Talos itself, but once again we'd have to pick the disk which would be also not the exact choice

PR #12209, opened 2025-11-12
chore: improve error handling for system disk reset

laurazard:
# Pull Request

## What? (description)

Change system disk reset flow to collects errors, checking whether META/STATE were successfully wiped along the way, and unconditionally runs the `delete state encryption info from META if STATE was wiped and META wasn't` flow.

## Why? (reasoning)

Previously, this code returned immediately on first error.

This means that if:
1. multiple system partitions are being wiped (i.e. STATE and EPHEMERAL)
2. `STATE` is wiped
3. `EPHEMERAL` fails to be wiped

Then the state encryption info won't be deleted from `META`.

## Acceptance


smira (approved):
👍

PR #12218, opened 2025-11-13
feat: update dependencies

smira:
Many small changes, linting fixes, migration to moby/moby client package.


internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
```diff
@@ -58,7 +58,7 @@ type LoadBalancerMetadata struct {
 			FrontendIPAddress string `json:"frontendIpAddress,omitempty"`
 			PrivateIPAddress  string `json:"privateIpAddress,omitempty"`
 		} `json:"publicIpAddresses,omitempty"`
-	} `json:"loadbalancer,omitempty"`
```
frezbo, line 61: intentional?

go.mod
```diff
@@ -344,22 +345,21 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	gotest.tools/v3 v3.4.0 // indirect
 	k8s.io/cli-runtime v0.35.0-alpha.3 // indirect
 	k8s.io/kube-openapi v0.0.0-20250910181357-589584f1c912 // indirect
-	kernel.org/pub/linux/libs/security/libcap/psx v1.2.76 // indirect
+	kernel.org/pub/linux/libs/security/libcap/psx v1.2.77 // indirect
 	rsc.io/qr v0.2.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/knftables v0.0.18 // indirect
```
frezbo, line 367: let's also bump the indirect selinux package

internal/app/machined/pkg/runtime/v1alpha1/platform/azure/azure.go
smira: yes, it's what linter found - this `omitempty` has no effect. In fact we never marshal this struct, so all of them should be no-op

go.mod
smira: bumped

PR #12223, opened 2025-11-13
fix: split volume/disk locators

smira:
Don't guess based on the volume type, but use explicit fields for different locators.

IMAGECACHE-ISO is a disk volume, but uses full volume locator (by filesystem type, etc.)


shanduur (approved):
Do we need integration tests change? I suspect that this needs adjustment and/or another test?

https://github.com/siderolabs/talos/blob/93f2e87c2d00c69aacc5f4422182db01b9e617fd/internal/integration/api/volumes.go#L701-L703

smira:
> Do we need integration tests change? I suspect that this needs adjustment and/or another test?

No, I don't think so, I only changed internal `VolumeConfig` resource, I haven't made any change to the user config.

PR #12233, opened 2025-11-16
feat: implement multicast setting

### LINKED ISSUE #12232: Allow multicast to be configured on links
## Feature Request
The `LinkConfig` document should have an option to enable/disable multicast traffic.

### Description
Disabling multicast can be done on Linux using the command `ip link set multicast off dev $IFACE`. Disabling multicast can be desired in numerous situations, such as when using Talos on the edge, where multicast traffic should not exist.

6ixfalls:
# Pull Request

Implemented as optional to preserve existing behavior, where providing the parameter in the LinkSpec will override the value. A multicast parameter was added to CommonLinkConfig that allows users to set the value if needed. This is my first PR to Talos, so please verify that I covered everything related to implementing this (but I think I covered everything). In addition, the test behavior is machine-dependent - there may need to be changes to check what the actual kernel is set to - but in 99% of cases, the runner should be using a kernel with multicast enabled anyway.

## What? (description)
Add a multicast option to LinkSpec that can be configured to explicitly declare the state of the multicast flag. Fixes #12232.

## Why? (reasoning)
Multicast is currently not configurable (uses the kernel default). Allows users to configure multicast, which can be useful in larger networks where multicast is forbidden.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/network/link_config.go
+++ b/internal/app/machined/pkg/controllers/network/link_config.go
@@ -408,7 +408,7 @@ func (ctrl *LinkConfigController) processDevicesConfiguration(
 	}
 }
 
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (ctrl *LinkConfigController) processLinkConfigs(logger *zap.Logger, linkMap map[string]*network.LinkSpecSpec, cfg *config.MachineConfig, linkNameResolver *network.LinkResolver) {
 	if cfg == nil {
 		return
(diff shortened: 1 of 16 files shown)


6ixfalls:
Hitting https://github.com/siderolabs/conform/issues/272 - do I need to re-sign with a GPG key instead? Don't have one setup.

shanduur:
@6ixfalls do not worry, this is internal thing, we will resign the commit if we will be merging it.

internal/app/machined/pkg/controllers/network/link_config.go
```diff
@@ -439,6 +439,11 @@ func (ctrl *LinkConfigController) processLinkConfigs(logger *zap.Logger, linkMap
 			linkMap[linkName].HardwareAddress = nil
 		}
 
+		if multicast, ok := linkConfig.Multicast().Get(); ok {
+			linkMap[linkName].Multicast = new(bool)
```
smira, line 443: I guess it should be `linkMap[linkName].Multicast = pointer.To(multicast)` ?

6ixfalls: Yeah, that sounds correct. Do you need me to patch this, or are you able to do that?

shanduur: @6ixfalls if you want to create PR, please do 😄

PR #12244, opened 2025-11-17
feat: support TALOS_HOME env var

### LINKED ISSUE #12222: Allow changing the talos directory via environment variable
## Feature Request
Add the `TALOSHOME` environment variable, which overrides the default location returned by [`GetTalosDirectory`](https://github.com/siderolabs/talos/blob/93f2e87c2d00c69aacc5f4422182db01b9e617fd/pkg/machinery/client/config/path.go#L23).

```go
// constants.go
const TalosHomeEnvVar = "TALOSHOME"

// path.go
func GetTalosDirectory() (string, error) {
	if path, ok := os.LookupEnv(constants.TalosHomeEnvVar); ok {
		return path, nil
	}
	home, err := os.UserHomeDir()
	if err != nil {
		return "", err
	}

	return filepath.Join(home, constants.TalosDir), nil
}
```

### Description
Currently `talosctl` always puts state data in the `~/.talos` directory.
This makes it difficult to test infrastructure tooling in an isolated manner, because the contents of the state directory are not fully isolated from the rest of the system.

It also makes it difficult to have multiple different filesystem locations for different state directories.
For example, a user might have two different disk drives, each containing talos state data for different clusters.
Currently it is impossible to use both state directories at the same time in different ttys (as the current workaround requires a symlink), but with this feature each tty can define the environment variable to use the corresponding location.

### Alternative
A command-line flag would also work for the same reasons as above.

Skyb0rg007:
# Pull Request

## What? (description)
Use TALOS_HOME for state instead of ~/.talos, when defined to an absolute filepath.
This resolves #12222.

## Why? (reasoning)
Currently talosctl always puts state data in the ~/.talos directory.
This makes it difficult to test infrastructure tooling in an isolated manner, because the contents of the state directory are not fully isolated from the rest of the system.

It also makes it difficult to have multiple different filesystem locations for different state directories.
For example, a user might have two different disk drives, each containing talos state data for different clusters.
Currently it is impossible to use both state directories at the same time in different ttys (as the current workaround requires a symlink), but with this feature each tty can define the environment variable to use the corresponding location.

The path referenced must be an abolute path to help avoid user error.

## Acceptance
The conformance fails GPG Identity; I'm not sure if this is because the commit is not signed-off by a sidero developer.

### DIFF
--- a/cmd/talosctl/cmd/completion.go
+++ b/cmd/talosctl/cmd/completion.go
@@ -38,11 +38,11 @@ Note for zsh users: [1] zsh completions are only supported in versions of zsh >=
 ## Load the talosctl completion code for bash into the current shell
 	source <(talosctl completion bash)
 ## Write bash completion code to a file and source if from .bash_profile
-	talosctl completion bash > ~/.talos/completion.bash.inc
-	printf "
+	talosctl completion bash > "${TALOS_HOME:-$HOME/.talos}/completion.bash.inc"
+	printf '
 		# talosctl shell completion
-		source '$HOME/.talos/completion.bash.inc'
-		" >> $HOME/.bash_profile
+		source "${TALOS_HOME:-$HOME/.talos}/completion.bash.inc"
+		' >> $HOME/.bash_profile
 	source $HOME/.bash_profile
 # Load the talosctl completion code for fish[1] into the current shell
 	talosctl completion fish | source
--- a/pkg/machinery/client/config/path.go
+++ b/pkg/machinery/client/config/path.go
@@ -20,8 +20,12 @@ type Path struct {
 	WriteAllowed bool
 }
 
-// GetTalosDirectory returns path to Talos directory (~/.talos).
+// GetTalosDirectory returns path to Talos directory ($TALOS_HOME or ~/.talos).
 func GetTalosDirectory() (string, error) {
+	if path, ok := os.LookupEnv(constants.TalosHomeEnvVar); ok && filepath.IsAbs(path) {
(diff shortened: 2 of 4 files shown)


shanduur:
Please run `make generate docs` 😄 Other than that, everything looks nice!

Skyb0rg007:
Unfortunately the `make docs` command does not succeed for me. I'm not sure why, since the logs do not show any errors.

```
------
Dockerfile:357
--------------------
 355 |     RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool mvdan.cc/gofumpt -w ./pkg/
 356 |     WORKDIR /src/pkg/machinery
 357 | >>> RUN --mount=type=cache,target=/.cache,id=talos/.cache go generate ./...
 358 |     RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool github.com/siderolabs/talos/tools/gotagsrewrite .
 359 |     RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool golang.org/x/tools/cmd/goimports -w -local github.com/siderolabs/talos ./
--------------------
ERROR: failed to build: failed to solve: process "/bin/bash -c go generate ./..." did not complete successfully: exit code: 1
make[2]: *** [Makefile:315: target-generate] Error 1
make[2]: Leaving directory '/home/skye/Projects/Forks/talos'
make[1]: *** [Makefile:322: local-generate] Error 2
make[1]: Leaving directory '/home/skye/Projects/Forks/talos'
make: *** [Makefile:341: generate] Error 2
```

Issue #12229, opened 2025-11-14
🥼  test: `misc-1` / `min-requirements`


smira:
There seems to be a bump in memory usage of Kubernetes e2e tests in the last release(s), which makes running e2e impossible for 1 GiB worker nodes.

We need to conditionally disable `talosctl health --run-e2e` for this specific test.

PR #12247, opened 2025-11-17
chore: disable k8s integration tests for 1GiB worker nodes

laurazard:
# Pull Request

## What? (description)

Disable `talosctl health --run-e2e` for worker nodes with <= 1GiB of memory.

Closes https://github.com/siderolabs/talos/issues/12229.

## Why? (reasoning)

Kubernetes E2E tests have had a bump in memory usage, making running these on 1GiB worker nodes unreliable.

## Acceptance

### DIFF
--- a/hack/test/e2e-qemu.sh
+++ b/hack/test/e2e-qemu.sh
@@ -359,7 +359,10 @@ case "${TEST_MODE:-default}" in
   *)
     get_kubeconfig
     run_talos_integration_test
-    run_kubernetes_integration_test
+
+    if [[ ${QEMU_MEMORY_WORKERS:-2048} -le 1024 ]]; then
+        run_kubernetes_integration_test
+    fi
 
     if [ "${WITH_TEST:-none}" != "none" ]; then
       "${WITH_TEST}"


shanduur (approved):

hack/test/e2e-qemu.sh
```diff
@@ -359,7 +359,10 @@ case "${TEST_MODE:-default}" in
   *)
     get_kubeconfig
     run_talos_integration_test
-    run_kubernetes_integration_test
+
+    if [[ ${QEMU_MEMORY_WORKERS:-2048} -le 1024 ]]; then
```
frezbo, line 363: so it's only misc-1 that uses 1Gig ram, maybe we add `SKIP_K8S_INTEGRATION_TEST` just for that?

laurazard: I thought about it, but I thought this was a better solution – my reasoning is that the actual thing we want to do is skip if the worker node doesn't have enough memory, so adding a flag `SKIP_K8S_INTEGRATION_TEST` is just an indirection of what we're actually trying to do.

Looking at`if [[ ${QEMU_MEMORY_WORKERS:-2048} -le 1024 ]]` if it's obvious why it's skipped (we only want the test to run if nodes have more memory), but looking at a `if [[ ! -z "$SKIP_K8S_INTEGRATION_TEST" ]]` or the like still forces me to go dig more/find the PR/commit to understand why.

I'm happy either way though, if others prefer it the other way.

frezbo: makes sense

PR #12258, opened 2025-11-20
feat: extend flags of cache-cert-gen

shanduur:
Add flags to add SAN names to generated certificate.

Additionally bump `golang.org/x/crypto` due to Security Advisory.


tools/go.mod
```diff
@@ -546,7 +546,7 @@ require (
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
 	go.yaml.in/yaml/v4 v4.0.0-rc.3 // indirect
 	go4.org v0.0.0-20230225012048-214862532bf5 // indirect
-	golang.org/x/crypto v0.44.0 // indirect
+	golang.org/x/crypto v0.45.0 // indirect
```
frezbo, line 549: let's do multiple commits now

PR #12261, opened 2025-11-21
fix: add more resilient move

### LINKED ISSUE #12259: images cache-create fails to copy tmp folder
## Bug Report

When using talosctl 1.12.0-beta.0 this command
```
cat images.txt | ./talosctl images cache-create --layout flat --image-cache-path ./image-cache.flat --images=-
```
fails with the following error
```
error generating cache: rename /tmp/talos-image-cache-gen370585548 ./image-cache.flat: invalid cross-device link
```
If I change the cache path to `/tmp/image-cache.flat` the command works. I also tried specifying the full path to $PWD and I get the same error.

### Environment

- Platform: linux

shanduur:
Fixes #12259

### DIFF
--- a/pkg/imager/cache/cache.go
+++ b/pkg/imager/cache/cache.go
@@ -11,10 +11,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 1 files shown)


pkg/imager/cache/cache.go
```diff
@@ -185,6 +187,78 @@ func Generate(images []string, platforms []string, insecure bool, imageLayerCach
 	return removeAll()
 }
 
+func move(src, dest string) error {
+	if err := os.Rename(src, dest); err == nil {
+		return nil
+	} else if !errors.Is(err, syscall.EXDEV) {
```
frezbo, line 193: TIL 😅

frezbo (approved):

pkg/imager/cache/cache.go
shanduur: Yeah, same 🥲

PR #12262, opened 2025-11-21
fix: add CA subject to generated certificate

shanduur:
Self-signed certificates are missing Subject/Issuer info, which are not present in CA. This sometimes might be causing issues as it is invalid format.


cmd/talosctl/pkg/mgmt/helpers/airgapped.go
```diff
@@ -13,7 +13,11 @@ import (
 
 // GenerateSelfSignedCert generates self-signed certificate.
 func GenerateSelfSignedCert(sanIPs []net.IP, sanNames []string) ([]byte, []byte, []byte, error) {
-	ca, err := x509.NewSelfSignedCertificateAuthority(x509.ECDSA(true))
+	ca, err := x509.NewSelfSignedCertificateAuthority(
+		x509.ECDSA(true),
+		x509.Organization("test"),
```
frezbo: let's use `talos.dev` since this is not a test file

frezbo (approved):

PR #12265, opened 2025-11-21
feat: generate mirrors patch

shanduur:
Upon starting a image cache server, attempt to generate patch for Talos.

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"log"
 	"net"
+	"net/url"
 	"os"
 	"os/signal"
 	"slices"
@@ -21,6 +22,7 @@ import (
 
 	"github.com/blang/semver/v4"
 	"github.com/dustin/go-humanize"
+	"github.com/siderolabs/gen/ensure"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/talos/image.go
```diff
@@ -443,6 +448,14 @@ var imageCacheServeCmd = &cobra.Command{
 			return fmt.Errorf("failed to create development logger: %w", err)
 		}
 
+		if err = generateMirrorsConfigPatch(
+			imageCacheServeCmdFlags.address,
+			imageCacheServeCmdFlags.mirrors,
+			imageCacheServeCmdFlags.tlsCertFile != "" && imageCacheServeCmdFlags.tlsKeyFile != "",
+		); err != nil {
+			development.Error("failed to generate Talos config patch for registry mirrors", zap.Error(err))
```
shanduur, line 456: This is not a fatal error, we don't want to crash image-cache server here, so just log it.

cmd/talosctl/cmd/talos/image.go
```diff
@@ -462,9 +475,88 @@ var imageCacheServeCmd = &cobra.Command{
...
+	if err != nil {
+		return err
+	}
+
+	patchBytes, err := ctr.EncodeBytes(encoder.WithComments(encoder.CommentsDisabled))
+	if err != nil {
+		return err
+	}
+
+	const patchFile = "image-cache-mirrors-patch.yaml"
```
frezbo, line 549: should we make this configurable?

frezbo (approved):

cmd/talosctl/cmd/talos/image.go
shanduur: Most of our patches are not configurable now. I want to change it at some point, but make it all at once.

PR #12288, opened 2025-11-27
chore(ci): skip multipath extension tests

frezbo:
Skip multipath and trident specific extensions from tests. We could add a multipath test later on.


smira:
promote to extensions?

frezbo:
> promote to extensions?

we're skipping those so didn't make any sense I guess

smira:
> > promote to extensions?
> 
> we're skipping those so didn't make any sense I guess

we need to prove that extensions test work with this change :)

smira (approved):

PR #12115, opened 2025-10-31
feat: implement logs persistence

smira:
Implement a log persistence controller, rotate logs and bufferize writes.

Fixes #11461

PR #12291, opened 2025-11-27
feat: present kernel log as `talosctl logs kernel`

smira:
Extracted from #12115

The idea is that kernel log can be delivered/persisted along with any other service logs.


smira:
> Should I update this branch to include the kernel log integration test I have in #12115?

let me pick it up, thanks, I forgot about it!

PR #12292, opened 2025-11-27
chore: update pkgs and tools version

shanduur:
Update pkgs and tools version


pkg/machinery/constants/constants.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	// DefaultKernelVersion is the default Linux kernel version.
-	DefaultKernelVersion = "6.17.7-talos"
```
smira, line 17: `hack/release.toml` update missing here

PR #12314, opened 2025-12-01
feat: rename `image default` and `image source-bundle`

rothgar:
s/default/k8s-bundle
s/source-bundle/talos-bundle

For UX consistency when generating lists of images used by talos and kubernetes. Also makes arguments to source-bundle optional and uses current version. It won't work for local, untagged builds, but should work the same with releases.

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -138,11 +138,12 @@ var imagePullCmd = &cobra.Command{
 	},
 }
 
-// imageDefaultCmd represents the image default command.
+// imageDefaultCmd represents the image k8s-bundle command.
 var imageDefaultCmd = &cobra.Command{
-	Use:   "default",
-	Short: "List the default images used by Talos",
-	Long:  ``,
+	Use:     "k8s-bundle",
+	Aliases: []string{"default"},
+	Short:   "List the default Kubernetes images used by Talos",
+	Long:    ``,
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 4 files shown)


smira:
Needs `make docs` at least, and it should go through image-cache, airgapped builds to verify the change

shanduur:
This needs few more changes.

`k8s-bundle` includes Talos installer, so we need to drop it from that command.

cmd/talosctl/cmd/talos/image.go
```diff
@@ -167,10 +168,6 @@ var imageDefaultCmd = &cobra.Command{
 		fmt.Printf("%s\n", images.KubeProxy)
 		fmt.Printf("%s\n", images.Kubelet)
 
-		if slices.Contains([]string{provisionerInstaller, provisionerAll}, imageDefaultCmdFlags.provisioner.String()) {
-			fmt.Printf("%s\n", images.Installer)
-		}
```
shanduur, line 172: The other check below must be gone as well.

rothgar: I think I removed the correct check. Let me know if there's something else.

hack/release.sh
```diff
@@ -22,7 +22,7 @@ function release-notes {
   release-tool "${2}" --gfm > "${1}"
 
   echo -e '\n## Images\n\n```' >> ${1}
-  ${ARTIFACTS}/talosctl-linux-amd64 image default >> ${1}
```
shanduur, line 25: This will create slightly different changelog now, but I think that's okay-ish.

PR #12317, opened 2025-12-02
chore: fix longhorn test

smira:
The condition added in #12288 was wrong - we shouldn't include trident-iscsi-tools.

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -2196,7 +2196,7 @@ spec:
         - name: installer extensions
           command: installer-with-extensions
           environment:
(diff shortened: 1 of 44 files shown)


.kres.yaml
```diff
@@ -2196,7 +2196,7 @@ spec:
         - name: installer extensions
           command: installer-with-extensions
           environment:
-            EXTENSIONS_FILTER_COMMAND: "grep -E 'iscsi-tools|util-linux-tools|trident-iscsi-tools'"
+            EXTENSIONS_FILTER_COMMAND: "grep -E '/iscsi-tools|util-linux-tools'"
```
smira, line 2199: this is the actual fix

PR #12324, opened 2025-12-03
fix: update the config schema $id path

Slessi:
Update the path in the $id of the config.schema.json

### DIFF
--- a/pkg/machinery/config/schemas/config.schema.json
+++ b/pkg/machinery/config/schemas/config.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://talos.dev/v1.12/schemas/config.schema.json",
+  "$id": "https://raw.githubusercontent.com/siderolabs/talos/v1.12/pkg/machinery/config/schemas/config.schema.json",
   "$defs": {
     "block.DiskSelector": {
       "properties": {
--- a/tools/docgen/schema.go
+++ b/tools/docgen/schema.go
@@ -21,7 +21,7 @@ import (
 	orderedmap "github.com/wk8/go-ordered-map/v2"
 )
 
-const ConfigSchemaURLFormat = "https://talos.dev/%s/schemas/%s"
+const ConfigSchemaURLFormat = "https://raw.githubusercontent.com/siderolabs/talos/%s/pkg/machinery/config/schemas/%s"
 
 // SchemaWrapper wraps jsonschema.Schema to provide correct YAML unmarshalling using its internal JSON marshaller.
 type SchemaWrapper struct {
--- a/website/content/v1.12/schemas/config.schema.json
+++ b/website/content/v1.12/schemas/config.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://talos.dev/v1.12/schemas/config.schema.json",
+  "$id": "https://raw.githubusercontent.com/siderolabs/talos/v1.12/pkg/machinery/config/schemas/config.schema.json",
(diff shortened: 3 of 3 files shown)


pkg/machinery/config/schemas/config.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://talos.dev/v1.12/schemas/config.schema.json",
+  "$id": "https://raw.githubusercontent.com/siderolabs/talos/v1.12/pkg/machinery/config/schemas/config.schema.json",
```
Slessi, line 3: @utkuozdemir doesnt actually exist though 👀 I guess expected cause no `v1.12` tag yet?

https://raw.githubusercontent.com/siderolabs/talos/v1.12/pkg/machinery/config/schemas/config.schema.json

smira: I don't like this publishing schema. It should be published under talos.dev back I think

Slessi: @smira the talos.dev links dont work, that is why I made this change

e.g.

- https://www.talos.dev/v1.11/schemas/config.schema.json
- https://www.talos.dev/v1.11.5/schemas/config.schema.json

I am just fixing the links to the ones that work, matching how it is specified in omni's `.kres.yaml`

https://github.com/siderolabs/omni/blob/60cb92a12558481fa3f870dc9220c30a55c016e1/.kres.yaml#L515

utkuozdemir: major.minor will never work anymore, see: https://raw.githubusercontent.com/siderolabs/talos/v1.11/pkg/machinery/config/schemas/config.schema.json

we could point it:
- either to the release branch: https://raw.githubusercontent.com/siderolabs/talos/release-1.12/pkg/machinery/config/schemas/config.schema.json
- or to a major.minor.patch[-alpha/beta etc]: https://raw.githubusercontent.com/siderolabs/talos/v1.12.0-beta.1/pkg/machinery/config/schemas/config.schema.json

I lean towards the release branch idea, as it would mean less updates, and config are stable for minor versions.

smira: @Slessi you can make them work by pushing config schema to talos.dev repository I guess, and I think it makes sense for historical schemas as well. It was one of the issues on the rework of the docs that got never fixed.

smira: https://github.com/siderolabs/talos-website

Slessi: @smira yea, but then it would be a task that requires more time investment than I expected. Just something I saw in passing that looked quick and easy to fix.

I could create a ticket to fix this the way you are describing. In the meantime we could merge this so at least its pointing to something real, or just close and they will one day work in the future, WDYT?

smira: Why can't you just open a PR to talos-website and copy schemas there?

Slessi: IDK isn't this an auto-generated file that may have changes between talos releases? You wan't me to manually copy paste it once? I thought it would require the creation of some kind of generation step to pull it from talos automatically or something. That's what I don't wanna invest time in.

smira: It's stable once we cut beta, or if we do change after that, we can handle sync. 

This file is totally frozen for any Talos version at this moment.

Slessi: https://talossystems.slack.com/archives/C0723NUK7K6/p1764788274345549?thread_ts=1764785965.447009&cid=C0723NUK7K6

PR #12348, opened 2025-12-05
docs: add omni join token example to create qemu command

Orzelius:
this will hopefully reduce confusion on what the flag value should look like

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
@@ -139,7 +139,8 @@ func addDisksFlag(flagset *pflag.FlagSet, bind *flags.Disks) {
 }
 
 func addOmniJoinTokenFlag(cmd *cobra.Command, bindAPIEndpoint *string, cfgPatchAllFlagName, cfgPatchWorkersFlagName, cfgPatchCPsFlagName string) {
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/cmd.go
```diff
@@ -139,7 +139,8 @@ func addDisksFlag(flagset *pflag.FlagSet, bind *flags.Disks) {
 }
 
 func addOmniJoinTokenFlag(cmd *cobra.Command, bindAPIEndpoint *string, cfgPatchAllFlagName, cfgPatchWorkersFlagName, cfgPatchCPsFlagName string) {
-	cmd.Flags().StringVar(bindAPIEndpoint, omniAPIEndpointFlagName, *bindAPIEndpoint, "the Omni API endpoint (must include a scheme, a port and a join token)")
+	cmd.Flags().StringVar(bindAPIEndpoint, omniAPIEndpointFlagName, *bindAPIEndpoint,
+		"the Omni API endpoint (must include a scheme, a port and a join token, e.g. 'grpc://omni.com:8090?jointoken=foobar')")
```
smira: let's do an example which matches more production Omni
```suggestion
		"the Omni API endpoint (must include a scheme, a port and a join token, e.g. 'https://siderolink.omni.example?jointoken=foobar')")
```

Orzelius: Is having no port number more usual?

smira: yep, it should be proper `https://` usually.

smira (approved):

PR #12362, opened 2025-12-08
feat: implement SSA sync for manifests

### LINKED ISSUE #12191: use server-side apply in Talos manifest apply and go-kubernetes
In Talos (https://github.com/siderolabs/talos/blob/2816466656a3d86ab4894328f38d038ae18912f2/internal/app/machined/pkg/controllers/k8s/manifest_apply.go#L186-L191) change to use SSA to create manifests, but keep the logic same - only create missing manifests. Populate inventory in Kubernetes.

In go-kubernetes library (https://github.com/siderolabs/go-kubernetes/blob/8454fe9977f5240a1251c2df1b4f93ca73b869a7/kubernetes/manifests/sync.go#L41) use SSA to update/sync manifests, including handling deletion of removed manifests.

Populate inventory gracefully - if something is not in the inventory, but it is in Kubernetes, add it to the inventory (to handle upgrades to new versions of Talos).

Do an integration test in Talos for that. 

> Note: This is for Talos 1.13. In Talos code, make sure we keep the RAM usage to minimum.

smira:
* add SSA via the new go-kubernetes library implementation to talosctl `upgrade-k8s` command
* add SSA via direct ResourceInterface call into talos (machined) with a manual inventory update
* add an integration test for ssa functionality
* only support ssa if both client and server are newer than v1.13.0

Fixes #12191


internal/app/machined/pkg/controllers/k8s/manifest_apply.go
```diff
@@ -160,7 +167,20 @@ func (ctrl *ManifestApplyController) Run(ctx context.Context, r controller.Runti
...
+				if err != nil {
+					return err
+				}
+
+				applyErr := ctrl.apply(ctx, logger, mapper, dyn, manifests, inv)
+
+				// update inventory even if the apply process failed half way through
+				err = inventoryClient.CreateOrUpdate(ctx, inv, inventory.UpdateOptions{})
+				if err != nil {
+					return err
```
smira: I think we should join the `err` and `applyErr` with `errors.Join()` to make both errors available

smira:
```suggestion
					err = fmt.Errorf("updating inventory failed: %w", err)
```

internal/app/machined/pkg/controllers/k8s/manifest_apply.go
```diff
@@ -160,7 +167,20 @@ func (ctrl *ManifestApplyController) Run(ctx context.Context, r controller.Runti
...
+				}
+
+				return applyErr
```
smira:
```suggestion
				return errors.Join(applyErr, err)
```

smira (commented):
make the output more closely resemble previous version (diffs, etc.)

add a note to `hack/release.toml`

hack/release.toml
```diff
@@ -53,6 +53,14 @@ If the kernel parameter is removed, default behavior is restored, allowing acces
         title = "Container Image Decompression"
         description = """\
 Talos now ships with `igzip` (amd64) and `pigz` (arm64) to speed up container image decompression.
+"""
+
+    [notes.k8s_ssa]
+        title = "Kubernetes server-side apply"
+        description = """\
+Talos now uses inventory backed server-side apply when applying bootsrap manifests (including `extraManifests` and `inlineManifests`).
+Purging of unneeded manifests is automatically performed.
```
smira, line 95:
```suggestion
Purging of unneeded manifests is automatically performed via `talosctl upgrade-k8s` command.
```

pkg/cluster/kubernetes/compat_test.go
```diff
@@ -0,0 +1,135 @@
...
+		name     string
+		versions []kubernetes.NodeVersion
+		want     *compatibility.TalosVersion
+		wantErr  bool
+	}{
+		{
+			name:     "empty",
+			versions: []kubernetes.NodeVersion{},
+			want:     nil,
+			wantErr:  false,
```
shanduur, line 41: nit: no need to set those, as they are defaults.

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -157,12 +159,24 @@ func Upgrade(ctx context.Context, cluster UpgradeProvider, options UpgradeOption
...
+	}
+
+	talos113version, err := semver.ParseTolerant("v1.13.0-alpha.0")
 	if err != nil {
 		return err
 	}
 
-	return syncManifests(ctx, objects, cluster, options)
+	// Use server-side apply for Talos versions >= v1.13.0
+	if semverMinV.Compare(talos113version) <= 0 {
```
smira: looks like the condition is inverted? 

one safe way to compare all pre-release versions without surprises is to get rid of `.Pre` part: https://github.com/siderolabs/talos/blob/b7c49777fe645b85ce9ec72c93922d2b154512f7/pkg/machinery/imager/quirks/quirks.go#L22-L28

and then you can compare with v1.13.0

pkg/cluster/kubernetes/talos_managed.go
```diff
@@ -157,12 +159,24 @@ func Upgrade(ctx context.Context, cluster UpgradeProvider, options UpgradeOption
...
 
-	objects, err := getManifests(ctx, cluster)
+	useSSA := false
+
+	semverMinV, err := semver.ParseTolerant(minTalosVersion.String())
+	if err != nil {
+		return fmt.Errorf("failed to parse minimum talos version %q for %w", minTalosVersion.String(), err)
+	}
+
+	talos113version, err := semver.ParseTolerant("v1.13.0-alpha.0")
```
shanduur: nit, future idea: we can have more generic `quirks` package that holds our talos version compatibility gates?

pkg/cluster/kubernetes/talos_managed.go
shanduur: Also, this should be `talos113version.GTE(semverMinV)`

pkg/cluster/kubernetes/talos_managed.go
smira: we already have config version contract which is super close, and I think we should actually use it here, as e.g. Omni will need to do same check, and we won't have to duplicate it there if we have it in the version contract

Orzelius: Okay, I'll work on implementing the logic into version contract, shouldn't take too long for what's needed here

pkg/cluster/kubernetes/talos_managed.go
smira: refactored to use TalosVersion from compatibility

pkg/cluster/kubernetes/talos_managed.go
smira: I moved this into TalosVersion

smira:
I did some small fixes to the integration test - applying machine config doesn't mean that manifest resources are updated immediately, so we need to wait for the resources are updated, otherwise the sync might pick up "old" manifests breaking the test

PR #12387, opened 2025-12-10
chore: prepare talos for 1.13

shanduur:
Add compatibility with 1.13 and fix/upgrade tests.


hack/release.toml
```diff
@@ -29,195 +29,9 @@ CoreDNS: 1.13.1
 etcd: 3.6.6
 Flannel: 0.27.4
```
smira: we can trim all of the above, as most of it went into 1.12.0

pkg/machinery/compatibility/kubernetes_version_test.go
```diff
@@ -389,8 +389,8 @@ func TestKubernetesCompatibilityUnsupported(t *testing.T) {
 	for _, tt := range []kubernetesVersionTest{
 		{
 			kubernetesVersion: "1.25.0",
-			target:            "1.13.0-alpha.0",
```
smira, line 392: I usually add an explicit test for each Talos version, like the one above for 1.12

pkg/machinery/compatibility/talos_version_test.go
```diff
@@ -400,9 +400,9 @@ func TestTalosUpgradeCompatibility111(t *testing.T) {
 			expectedError: `host version 1.8.0 is too old to upgrade to Talos 1.11.0`,
 		},
 		{
-			host:          "1.13.0-alpha.0",
-			target:        "1.11.0",
-			expectedError: `host version 1.13.0-alpha.0 is too new to downgrade to Talos 1.11.0`,
```
smira, line 405: same here in terms of tests

pkg/machinery/gendata/data/tag
```diff
@@ -1 +1 @@
-v1.12.0-alpha.2
```
smira, line 1: this should be reverted until we actually release v1.13

website/content/v1.13/reference/configuration/block/_index.md
```diff
@@ -0,0 +1,8 @@
+---
+description: |
```
smira, line 2: we should drop v1.12 reference and update `Dockerfile` to output to v1.13/ folder

PR #12391, opened 2025-12-11
fix: do not override DNS on MacOS

shanduur:
When creating Talos with QEMU on Mac, do not override default DNS settings to Gateway IPs.

Reverts (partially): https://github.com/siderolabs/talos/commit/c0772b8eda429675a06899b9c4a4d1dd7d5f6a5f#diff-9e7463328a82c08aae57d1c6a1d157c13da886a3161d319178a156f265e44152L181-L184


smira:
Can we pass it through airgapped pipeline to prove that it still works on Linux?

oguzkilcan (approved):

PR #12392, opened 2025-12-11
feat: `talosctl debug`

### LINKED ISSUE #8720: Ability to debug nodes with running debug container
### Description

It would be nice to relaise API interface and command for talosctl to debug.

It might be done the same way as kubectl debug:

```
# Create an interactive debugging session in pod mypod and immediately attach to it.
talosctl debug mypod -n 11.22.33.44 -e 11.22.33.44 -it --image=busybox -- /bin/sh
```

Or to proxy CRI socket the same way how ssh agent works:

```
 eval `talosctl proxy-cri -n 11.22.33.44 -e 11.22.33.44`
```

which outputs:

```
export CONTAINER_RUNTIME_ENDPOINT=unix://var/run/talosctl/38ab4953-994e-4517-bf4f-43ad8b5d2b38.sock
```

then use crictl to run debug CRI:

```
# crictl ps | grep kube-apiserver
3ff4626a9f10e       e7972205b6614       6 hours ago         Running             kube-apiserver         0                   215107b47bd7e       kube-apiserver-talos-rzq-nkg
```

Or run a container:

```
# Run a container:
crictl run docker.io/library/busybox:latest

# Attach a shell to a running container:
crictl exec -it my-container sh
```

laurazard:
Fixes #8720


laurazard:
@smira when you get a sec can you TAL and see if you have any early feedback/ideas?

internal/app/apid/main.go
```diff
@@ -129,6 +129,7 @@ func apidMain() error {
 		"/machine.MachineService/ImageList",
 		"/machine.MachineService/Kubeconfig",
 		"/machine.MachineService/List",
+		"/machine.MachineService/DebugContainer",
```
smira, line 132: you don't need this

api/machine/machine.proto
```diff
@@ -103,6 +103,10 @@ service MachineService {
   rpc ImageList(ImageListRequest) returns (stream ImageListResponse);
   // ImagePull pulls an image into the CRI.
   rpc ImagePull(ImagePullRequest) returns (ImagePullResponse);
+  // DebugContainerCreate
```
smira: leaving some notes after the call:

probably we can move this out to a separate `DebugService`  (I think it makes sense to be `machine.DebugService`) ?

internal/app/apid/pkg/director/director.go
```diff
@@ -84,6 +84,10 @@ func (r *Router) Director(ctx context.Context, fullMethodName string) (proxy.Mod
 			return proxy.One2One, nil, status.Error(codes.InvalidArgument, "one-2-many proxying is not supported for COSI methods")
 		}
 
+		if strings.HasPrefix(fullMethodName, "/machine.MachineService/DebugContainer") {
```
smira: turn this inside out, specify methods which _support_ one2many proxying (probably `/machine.MachineService` and a couple of others ? and everything else should return an error

api/machine/machine.proto
smira: I don't remember, but I think `apid_test` should scream at you as you haven't specified RBAC for it. 

If it doesn't scream, let's make sure it screams, and specify RBAC (os:admin).

internal/app/machined/internal/server/v1alpha1/v1alpha1_debug.go
```diff
@@ -0,0 +1,530 @@
...
+	b := make([]byte, 8)
+	if _, err := rand.Read(b); err != nil {
+		return "", fmt.Errorf("failed to generate random ID: %w", err)
+	}
+
+	return fmt.Sprintf("debug-%s", hex.EncodeToString(b)), nil
+}
+
+// DebugContainerRun implements the machine.MachineServer interface.
+func (s *Server) DebugContainerRun(srv machine.MachineService_DebugContainerRunServer) error { //nolint:gocyclo
```
smira: as we separated the protos, we can separate the implementation into its own `DebugServer` Go struct which we can now "mix in" into the maintenance/normal mode client.

We should probably double-assert here on `os:admin` RBAC, as in maintenance mode we assign "os:admin" to SideroLink activated connections, and there is no middleware for RBAC

api/machine/machine.proto
smira: oops, I'm blind, I see RBAC changes, but let's double check if we move it to a separate proto, that the test picks it up

internal/app/machined/internal/server/v1alpha1/v1alpha1_debug.go
```diff
@@ -0,0 +1,530 @@
...
+	mu:         sync.Mutex{},
+	images:     map[containerdapi.Image]chan struct{}{},
+	containers: map[containerdapi.Container]chan struct{}{},
+}
+
+// DebugContainerCreate implements the machine.MachineServer interface.
+func (s *Server) DebugContainerCreate(srv machine.MachineService_DebugContainerCreateServer) error { //nolint:gocyclo
+	ctx := srv.Context()
+
+	client, err := containerdapi.New(constants.SystemContainerdAddress,
```
smira: food for thought: should we pick up the containerd to use automatically? if the "CRI" one is up, use it always? and use "system" one only if CRI is not up yet?

smira:
TODO: 

- [x] integration tests for CLI
- [x] figure out problems with non-terminal input
- [x] figure out cgroup 
- [x] add support for running in different containerd instances

internal/integration/api/debug.go
```diff
@@ -0,0 +1,144 @@
...
+
+// TearDownTest ...
+func (suite *DebugSuite) TearDownTest() {
+	if suite.ctxCancel != nil {
+		suite.ctxCancel()
+	}
+}
+
+// TestRunAlpine tests running a simple alpine container via DebugService.
+func (suite *DebugSuite) TestRunAlpine() {
```
frezbo, line 48: I wonder if we can use this for all other tests that does nsenter

smira: yep, we could I guess - I can create an issue to transform this, we might still nsenter due to different mount namespace

PR #12396, opened 2025-12-11
fix: disable kexec in talosctl cluster create on arm64

### LINKED ISSUE #12393: disable kexec in talosctl cluster create/arm64
For unknown reasons, kexec broke on QEMU/arm64 (tested on Mac OS X).

The breakage is between Talos v1.11.3 and Talos v1.11.4, most probably caused by Linux 6.12.53 .. 6.12.57 change.

At the same time test on AWS/arm64 shows that kexec works, meaning it might be a specific issue with QEMU/arm64 or QEMU/OS X and Linux kernel. So we can do a workaround for now.

smira:
Fixes #12393


smira:
> Backport to 1.11 as well?

good point

PR #12406, opened 2025-12-12
feat: rootless imager

frezbo:
feat: imager support rootless builds
    
Also changes the bootloader interface.
Disks are formatted/created with pre-populated source directories in Install/Image mode.


cmd/installer/pkg/install/install.go
```diff
@@ -230,6 +233,99 @@ func NewInstaller(ctx context.Context, cmdline *procfs.Cmdline, mode Mode, opts
...
+	switch mode {
+	case ModeInstall, ModeUpgrade:
+		info, err := blkid.Probe(bd.File(), blkid.WithSkipLocking(true))
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to probe blockdevice %s: %w", i.options.DiskPath, err)
+		}
+
+		return bd, info, nil
+	case ModeImage:
+		// in image mode, we don't need to probe the disk
```
smira: the comment doesn't match code? we still probe the image?

Dockerfile
```diff
@@ -1053,6 +1053,7 @@ COPY --link --from=pkg-fhs / /
 COPY --link --from=pkg-ca-certificates / /
 COPY --link --exclude=**/*.a --exclude=**/*.la --exclude=usr/include --from=pkg-musl / /
 COPY --link --from=pkg-dosfstools / /
+COPY --link --from=pkg-mtools / /
```
smira: let's move it back...

Makefile
```diff
@@ -424,7 +424,7 @@ image-%: ## Builds the specified image. Valid options are aws, azure, digital-oc
 	@docker pull $(REGISTRY_AND_USERNAME)/imager:$(IMAGE_TAG_IN)
 	@for platform in $(subst $(,),$(space),$(PLATFORM)); do \
 		arch=$$(basename "$${platform}") && \
-		docker run --rm -t -e GITHUB_TOKEN -v /dev:/dev -v $(PWD)/$(ARTIFACTS):/secureboot:ro -v $(PWD)/$(ARTIFACTS):/out -e SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) --network=host --privileged $(REGISTRY_AND_USERNAME)/imager:$(IMAGE_TAG_IN) $* --arch $$arch --base-installer-image $(REGISTRY_AND_USERNAME)/installer-base:$(IMAGE_TAG_IN) $(IMAGER_ARGS) ; \
+		docker run --rm -t -e GITHUB_TOKEN -v $(PWD)/$(ARTIFACTS):/secureboot:ro -v $(PWD)/$(ARTIFACTS):/out -e SOURCE_DATE_EPOCH=$(SOURCE_DATE_EPOCH) --network=host $(REGISTRY_AND_USERNAME)/imager:$(IMAGE_TAG_IN) $* --arch $$arch --base-installer-image $(REGISTRY_AND_USERNAME)/installer-base:$(IMAGE_TAG_IN) $(IMAGER_ARGS) ; \
```
smira: 🎉

pkg/imager/out.go
```diff
@@ -316,42 +311,7 @@ func (i *Imager) buildImage(ctx context.Context, path string, printf func(string
 		return err
 	}
 
-	printf("attaching loopback device")
```
smira, line 319: 🎉

cmd/installer/cmd/imager/root.go
```diff
@@ -181,23 +179,6 @@ var rootCmd = &cobra.Command{
 					prof.Input.ImageCache.ForceInsecure = cmdFlags.Insecure
 				}
 
-				if cmdFlags.ImageDiskSize != "" {
```
frezbo, line 184: this is dropped since now we calculate disk size based on the input image cache size, this was added when I implemented image cache

smira: I understand that we might not need it, but should we keep it for someone who wants to bump the disk image size (for whatever reason)? Or is it confusing to keep?

cmd/installer/pkg/install/install.go
```diff
@@ -90,6 +97,9 @@ func (m Mode) IsImage() bool {
 
 const typeGPT = "gpt"
 
+// luckily `talos-image-disk` is 16 bytes long.
+var diskImageGPTUUID = uuid.Must(uuid.FromBytes([]byte(`talos-image-disk`)))
```
smira: I wonder if we should do something like https://github.com/google/uuid/blob/v1.6.0/hash.go#L33 so that the UUID has a valid type and variant set.

cmd/installer/pkg/install/install.go
```diff
@@ -249,130 +351,214 @@ func (i *Installer) Install(ctx context.Context, mode Mode) (err error) {
...
+
+		metaFile, err := os.Create(filepath.Join(i.options.MountPrefix, "meta.img"))
 		if err != nil {
-			return fmt.Errorf("failed to probe bootloader on upgrade: %w", err)
+			return fmt.Errorf("failed to create meta image file: %w", err)
 		}
+
+		defer metaFile.Close() //nolint:errcheck
+
+		pr, err := pt.PartitionReader(metaPartitionIndex)
```
smira: we probably don't need to read it back here, as it should be just all zeroes, we can create an empty file of 1 MB?

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
```diff
@@ -0,0 +1,72 @@
...
+// - core image embedded blocklist continuation at core.img offset 0x1F4.
+func PatchBlocklistsForDiskImage(efiPartitionSizeBytes uint64, mountPrefix string) error {
+	if efiPartitionSizeBytes == 0 {
+		return fmt.Errorf("EFI partition size must be set to patch GRUB blocklists")
+	}
+
+	// Talos partition layout (GPT): EFI (gpt1, efiPartitionSizeBytes), BIOS (gpt2, 1MiB), BOOT (gpt3)
+	// BIOS boot partition starts immediately after the EFI partition.
+	const (
+		firstPartitionStart       = 2048  // sector 2048 (1MiB)
```
smira: I don't remember, but I think it might be different for 4k sector size, should we get it from the GPT partition table? (It has this as a field in the header)

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
```diff
@@ -0,0 +1,72 @@
...
+	bootImgPath := filepath.Join(mountPrefix, "boot.img")
+	coreImgPath := filepath.Join(mountPrefix, "core.img")
+
+	bootImg, err := os.ReadFile(bootImgPath)
+	if err != nil {
+		return fmt.Errorf("failed to read boot.img: %w", err)
+	}
+
+	// Patch 1: tell boot.img where to find core.img (LE64 sector number at 0x5C)
+	binary.LittleEndian.PutUint64(bootImg[bootImgKernelSectorOffset:], biosBootStartSector)
```
smira, line 62: is it LE even on arm64?

pkg/imager/out.go
```diff
@@ -414,6 +371,24 @@ func (i *Imager) buildImage(ctx context.Context, path string, printf func(string
...
 
 		opts.ImageCachePath = imageCacheDir
+
+		imageCacheSize, err := calculateDirectorySizeWithOverhead(imageCacheDir, 20)
+		if err != nil {
+			return fmt.Errorf("failed to calculate image cache size: %w", err)
+		}
+
+		// Align to 512-byte sector boundary to match GPT partition allocation
+		imageCacheSize = (imageCacheSize + 511) &^ 511
```
smira: I think this should be sector size

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
frezbo: this code is not used for ARM64, since MBR is BIOS only

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
frezbo: I'll manually check

cmd/installer/pkg/install/install.go
frezbo: ah yes

frezbo:
> https://github.com/siderolabs/talos/pull/12406#discussion_r2640163034

This was added just with image cache, since now we calcuate it's size and align it, it should be good I guess, less confusing flags

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
smira: ah yeah 🤦

pkg/makefs/xfs.go
```diff
@@ -64,8 +64,21 @@ func XFS(partname string, setters ...Option) error {
 		args = append(args, "--unsupported")
 	}
 
+	if opts.SourceDirectory != "" {
+		args = append(args, "-p", fmt.Sprintf("file=%s", opts.SourceDirectory))
+	}
```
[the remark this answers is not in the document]
frezbo: this is not correct, xfs 6.17.0 added support for populating from a directory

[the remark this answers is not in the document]
frezbo: see https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/mkfs/xfs_mkfs.c?h=v6.17.0#n1229

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/blocklist.go
frezbo: yes, it's 2048 start for 512 and 256 for 4k

cmd/installer/pkg/install/install.go
```diff
@@ -398,65 +584,326 @@ func (i *Installer) Install(ctx context.Context, mode Mode) (err error) {
...
+
+	w, size, err := pt.PartitionWriter(biosPartitionIndex)
+	if err != nil {
+		return fmt.Errorf("failed to get partition writer for partition %s: %w", biosPartitionInfo[0].Label, err)
+	}
+
+	if size != int(biosPartitionInfo[0].Size) {
+		return fmt.Errorf("partition size mismatch for partition %s: expected %d, got %d", biosPartitionInfo[0].Label, biosPartitionInfo[0].Size, size)
+	}
+
```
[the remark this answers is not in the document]
frezbo: as per:

```
go doc io.WriterAt
package io // import "io"

type WriterAt interface {
	WriteAt(p []byte, off int64) (n int, err error)
}
    WriterAt is the interface that wraps the basic WriteAt method.

    WriteAt writes len(p) bytes from p to the underlying data stream at offset
    off. It returns the number of bytes written from p (0 <= n <= len(p)) and
    any error encountered that caused the write to stop early. WriteAt must
    return a non-nil error if it returns n < len(p).

    If WriteAt is writing to a destination with a seek offset, WriteAt should
    not affect nor be affected by the underlying seek offset.

    Clients of WriteAt can execute parallel WriteAt calls on the same
    destination if the ranges do not overlap.

    Implementations must not retain p.
```

WriteAt doesn't change the Seek offset

cmd/installer/pkg/install/install.go
```diff
@@ -398,65 +584,326 @@ func (i *Installer) Install(ctx context.Context, mode Mode) (err error) {
...
+					return fmt.Errorf("failed to format partition %s: %w", partitionImageFile, err)
+				}
+			}
+
+			partitionData, err := os.Open(partitionImageFile)
+			if err != nil {
+				return fmt.Errorf("failed to open partition image file %s: %w", partitionImageFile, err)
+			}
+
+			defer partitionData.Close() //nolint:errcheck
```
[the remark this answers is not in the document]
frezbo: fixed

pkg/imager/out.go
frezbo: aligned to 1Mib

cmd/installer/pkg/install/install.go
frezbo: fixed

cmd/installer/pkg/install/install.go
frezbo: fixed

internal/app/machined/pkg/runtime/v1alpha1/bootloader/sdboot/sdboot.go
```diff
@@ -295,11 +334,46 @@ func (c *Config) Install(opts options.InstallOptions) (*options.InstallResult, e
...
+			if err := c.generateAssets(opts, constants.EFIMountPoint, ukiPath); err != nil {
+				return err
+			}
+
+			installResult, err = c.setup(opts, ukiPath)
+			if err != nil {
+				return err
+			}
+
+			return err
```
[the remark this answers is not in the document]
frezbo: not correct at line 327 err is assigned to the func err, so we should return it

cmd/installer/pkg/install/install.go
```diff
@@ -230,6 +239,98 @@ func NewInstaller(ctx context.Context, cmdline *procfs.Cmdline, mode Mode, opts
...
+	switch mode {
+	case ModeInstall, ModeUpgrade:
+		info, err := blkid.Probe(bd.File(), blkid.WithSkipLocking(true))
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to probe blockdevice %s: %w", i.options.DiskPath, err)
+		}
+
+		return bd, info, nil
+	case ModeImage:
+		info, err := blkid.ProbePath(i.options.DiskPath)
```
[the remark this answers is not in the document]
frezbo: this is fine since caller calls defer

pkg/makefs/ext4_test.go
```diff
@@ -17,7 +17,6 @@ import (
 
 // TestExt4Reproducibility tests that the ext4 filesystem is reproducible.
 func TestExt4Reproducibility(t *testing.T) {
-	t.Setenv("SOURCE_DATE_EPOCH", "1732109929")
 	t.Setenv("PATH", "/usr/bin:/bin:/usr/sbin:/sbin")
```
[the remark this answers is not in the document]
frezbo: ext4 doesn't need source date epoch set for reproducibility

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
```diff
@@ -181,7 +310,7 @@ func (c *Config) install(opts options.InstallOptions) (*options.InstallResult, e
 		platforms = []string{"arm64-efi"}
 	}
 
-	if runtime.GOARCH == amd64 && opts.Arch == amd64 && !opts.ImageMode {
+	if runtime.GOARCH == amd64 && opts.Arch == amd64 {
```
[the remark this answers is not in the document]
frezbo: the constant is used at multiple places

pkg/makefs/xfs_test.go
```diff
@@ -117,3 +118,55 @@ realtime =none                   extsz=4096   blocks=0, rtextents=0
 		})
 	}
 }
+
+func TestXFSReproducibility(t *testing.T) {
+	t.Setenv("SOURCE_DATE_EPOCH", "1732109929")
```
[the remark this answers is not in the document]
frezbo: `mkfs.xfs` needs both `SOURCE_DATE_EPOCH` and `DETERMINISTIC_UUID` set along with setting a fixed UUID

PR #12443, opened 2025-12-17
fix: make upgrade work with SELinux enforcing=1

dsseng:
Add a test for this

### DIFF
--- a/internal/integration/provision/upgrade.go
+++ b/internal/integration/provision/upgrade.go
@@ -10,15 +10,23 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/cosi-project/runtime/pkg/resource/rtestutils"
+	"github.com/siderolabs/go-procfs/procfs"
+	"github.com/stretchr/testify/assert"
+
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/mgmt/helpers"
 	"github.com/siderolabs/talos/pkg/images"
+	talosclient "github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/config/machine"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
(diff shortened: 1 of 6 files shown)


internal/integration/provision/upgrade.go
```diff
@@ -337,5 +366,6 @@ func init() {
 		&UpgradeSuite{specGen: upgradeCurrentToCurrentBios, track: 0},
 		&UpgradeSuite{specGen: upgradeStableToCurrentPreserveStage, track: 1},
 		&UpgradeSuite{specGen: upgradeCurrentToCurrentNewCmdline, track: 2},
+		&UpgradeSuite{specGen: upgradeCurrentToCurrentEnforcing, track: 2},
```
smira: nit: we can probably put it to track 0 or 1, as track 2 is too long already

internal/integration/provision/upgrade.go
```diff
@@ -226,6 +226,35 @@ func upgradeCurrentToCurrentNewCmdline() upgradeSpec {
...
+		"%s/%s:%s-enforcing",
+		DefaultSettings.TargetInstallImageRegistry,
+		images.DefaultInstallerImageName,
+		DefaultSettings.CurrentVersion,
+	)
+
+	return upgradeSpec{
+		ShortName: fmt.Sprintf("%s-same-ver-enforcing", DefaultSettings.CurrentVersion),
+
+		SourceISOPath:        helpers.ArtifactPath("metal-amd64.iso"),
```
smira, line 251: this doesn't test booting ISO in enforcing mode... that's why we might instead inject via QEMU extra cmdline args

dsseng: Yes, and it seems it has not worked out. Will fix

.github/workflows/ci.yaml
dsseng, line 1: stray changes from rekres - will drop when rebasing

PR #12446, opened 2025-12-17
fix: drop the Omni API URL check on IP address

smira:
I don't see much point in this check, as it's only valuable when joining to a local development instance of Omni, which is pretty nice usecase.

But this check breaks joining to "real" Omni which has hostname in the endpoint.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
@@ -261,12 +261,7 @@ func (m *Maker[T]) applyOmniConfigs() error {
 
 	cfg.APIUrlConfig.URL = parsedURL
 
-	mode, err := runtime.ParseMode(runtime.ModeMetal.String())
-	if err != nil {
-		return err
-	}
-
-	_, err = cfg.Validate(mode)
+	_, err = cfg.Validate(runtime.ModeMetal)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
```diff
@@ -261,12 +261,7 @@ func (m *Maker[T]) applyOmniConfigs() error {
 
 	cfg.APIUrlConfig.URL = parsedURL
 
-	mode, err := runtime.ParseMode(runtime.ModeMetal.String())
-	if err != nil {
-		return err
-	}
-
-	_, err = cfg.Validate(mode)
```
smira, line 269: this is unrelated simplifaction

cmd/talosctl/cmd/mgmt/cluster/create/create_qemu.go
```diff
@@ -143,26 +135,3 @@ func writeMachineconfig(clusterConfigs clusterops.ClusterConfigs, cOps clusterop
...
-
-func checkLoopbackOmniURL(cOps clusterops.Common, clusterConfigs clusterops.ClusterConfigs) error {
-	parsedURL, err := url.Parse(cOps.OmniAPIEndpoint)
-	if err != nil {
-		return err
-	}
-
-	host := parsedURL.Hostname()
-
-	ip, err := netip.ParseAddr(host)
```
smira, line 155: this is the core of the issue - attempt to parse hostname as IP which only works for IPs

rothgar (approved):
I just built and ran this PR and it worked for my Omni API

PR #12448, opened 2025-12-17
test: run virtiofs tests only when virtiofsd is running

shanduur:
Detect if virtiofsd is created, and then run or skip virtiofs volumes tests.


frezbo (approved):
let's also run through qemu-enforcing

PR #12461, opened 2025-12-20
fix: update LinkSelector to accept string or true

yesdevnull:
# Pull Request

## What?

Fixes the schema so the `selector.match` property on `LinkAliasConfig` accepts `true` or a CEL expression.

## Why?

As per the [documentation](https://docs.siderolabs.com/talos/v1.12/networking/configuration/aliases) it's valid to provide either `true` or a CEL expression.

## Acceptance

Uses [`check-yamlschema@0.0.7`](https://pypi.org/project/check-yamlschema/0.0.7/) for validation.

```yaml
apiVersion: v1alpha1
kind: LinkAliasConfig
name: net0
selector:
    match: mac(link.permanent_addr) == "00:1a:2b:3c:4d:5e"
```

✅ yields `document 0: validated`

```yaml
apiVersion: v1alpha1
kind: LinkAliasConfig
name: net0
selector:
    match: true
```

✅ yields `document 0: validated`

```yaml
apiVersion: v1alpha1
kind: LinkAliasConfig
name: net0
selector:
    match: false
```

❌ yields `{'apiVersion': 'v1alpha1', 'kind': 'LinkAliasConfig', 'name': 'net0', 'selector': {'match': False}} is not valid under any of the given schemas`

_P.S. hopefully I updated the code in all the right spots and did it the right way._

### DIFF
--- a/pkg/machinery/config/schemas/config.schema.json
+++ b/pkg/machinery/config/schemas/config.schema.json
@@ -2273,7 +2273,16 @@
     "network.LinkSelector": {
       "properties": {
         "match": {
-          "type": "string",
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "enum": [
+                true
+              ]
+            }
+          ],
           "title": "match",
           "description": "The Common Expression Language (CEL) expression to match the link.\n",
           "markdownDescription": "The Common Expression Language (CEL) expression to match the link.",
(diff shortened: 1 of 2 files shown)


yesdevnull:
Unsure why `conform/commit/gpg-identity` is failing, is it because I'm not part of the Sidero Labs GitHub org?

smira:
This is a wrong change, `true` is a valid CEL expression, but you need to quote it according to YAML rules.

PR #12473, opened 2025-12-23
fix: invalid versions check in talos-bundle

### LINKED ISSUE #12471: `talosctl image talos-bundle` incorrectly handles maximum version
### Description

The command `talosctl image talos-bundle v1.12.0` fails with:
```
invalid argument "v1.12.0" for "talosctl image talos-bundle": tag for the bundle must be within range "v1.11.0-alpha.0" - "v1.12.0"
```
However, `v1.12.0` is within the range

### Root cause

The version validation uses `GTE` for minimum version but only `LT` for the maximum version, when it should be `LTE`
https://github.com/siderolabs/talos/blob/48d242918bc97e6a01434bee6fcdcfa735fd1f5a/cmd/talosctl/cmd/talos/image.go#L213

### Environment

Talosctl:
* Tag: v1.12.0
* SHA: ac91ade2
* Built: 
* Go version: go1.25.5
* OS/Arch: linux/amd64

shanduur:
Fixes #12471

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -198,6 +198,12 @@ var imageSourceBundleCmd = &cobra.Command{
 				panic(err) // panic, this should never happen
 			}
 
+			maximumVersion.Patch = 0
+			maximumVersion.Pre = nil
+			if err := maximumVersion.IncrementMinor(); err != nil {
+				panic(err) // panic, this should never happen
+			}
+
 			// If no version specified, use current version
 			if len(args) == 0 {
 				return nil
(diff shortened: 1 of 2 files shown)


smira (approved):
should we have a simple integration/cli test where we run this command with current version?

cmd/talosctl/cmd/talos/image.go
```diff
@@ -198,6 +198,12 @@ var imageSourceBundleCmd = &cobra.Command{
 				panic(err) // panic, this should never happen
 			}
 
+			maximumVersion.Patch = 0
```
frezbo, line 201: we could just return like in quirks https://github.com/siderolabs/talos/blob/main/pkg/machinery/imager/quirks/quirks.go#L16

shanduur: Quirks are `imager` specific, I don't think this belongs there :/

frezbo (approved):

Issue #2120, opened 2020-05-17
Allow etcd to operate in an IPv6-only environment


Ulexus:
## Feature Request

There are a number of places which still presume etcd will run in an environment which has IPv4 addresses, which is not always the case.  We need to remove these assumptions.

james-callahan:
This seems to still be incorrect with Talos 1.3.0

```json
{"level":"fatal","ts":"2023-01-13T04:59:37.489Z","caller":"etcdmain/etcd.go:204","msg":"discovery failed","error":"--initial-advertise-peer-urls \"http://fe80::ca7:82ff:fe53:bad3:2380\" must be \"host:port\" (address fe80::ca7:82ff:fe53:bad3:2380: too many colons in address)","stacktrace":"go.etcd.io/etcd/server/v3/etcdmain.startEtcdOrProxyV2\n\tgo.etcd.io/etcd/server/v3/etcdmain/etcd.go:204\ngo.etcd.io/etcd/server/v3/etcdmain.Main\n\tgo.etcd.io/etcd/server/v3/etcdmain/main.go:40\nmain.main\n\tgo.etcd.io/etcd/server/v3/main.go:32\nruntime.main\n\truntime/proc.go:225"}
```

smira:
> This seems to still be incorrect with Talos 1.3.0

this is not generated by Talos, `http://` URLs are never used.

PR #12475, opened 2025-12-23
feat!: support multiple values for extraArgs

### LINKED ISSUE #11694: [Feature Request] Support multiple --service-account-issuer flags for zero-downtime endpoint migration
# [Feature Request] Support multiple --service-account-issuer flags for zero-downtime endpoint migration

## Feature Request

Add support for configuring multiple `--service-account-issuer` flags in the Talos API server configuration to enable zero-downtime migration when changing cluster control plane endpoints.

## Problem Statement

By default, Talos automatically sets the `service-account-issuer` equal to `cluster.controlPlane.endpoint`. When the control plane endpoint needs to be changed (e.g., during load balancer migration, DNS changes, or infrastructure updates) or when the service-account-issuer needs to be different than the control plane endpoint (e.g., hosted externally), this causes immediate authentication failures and cluster downtime because:

1. Existing service account tokens were issued by the old issuer URL
2. The API server immediately switches to only accept tokens from the new issuer URL  
3. All running workloads with mounted service account tokens lose authentication until pods are restarted

## Current Behavior

```yaml
# Current Talos configuration
cluster:
  controlPlane:
    endpoint: https://new-endpoint.example.com:6443
```

This results in:
- Immediate authentication failures for existing workloads
- Required pod restarts across the entire cluster
- Operational downtime during endpoint changes

## Proposed Solution

Enable configuration of multiple service account issuers, similar to how Kubernetes API server supports this natively since v1.22:

### Option 1: Array configuration in controlPlane section
```yaml
cluster:
  controlPlane:
    endpoint: https://new-endpoint.example.com:6443
    serviceAccountIssuers:
      - https://old-endpoint.example.com:6443  # Still validate existing tokens
      - https://new-endpoint.example.com:6443  # Generate new tokens (first in list)
```

### Option 2: Enhanced extraArgs support for repeatable flags
```yaml
cluster:
  apiServer:
    extraArgs:
      service-account-issuer: 
        - https://old-endpoint.example.com:6443
        - https://new-endpoint.example.com:6443
```

*Note: This would require Talos to enhance `extraArgs` to detect array values and automatically convert them to multiple flag instances (e.g., `--service-account-issuer=https://old-endpoint.example.com:6443 --service-account-issuer=https://new-endpoint.example.com:6443`). This approach would also benefit other repeatable Kubernetes API server flags.*

## Technical Background

Kubernetes API server has supported multiple `--service-account-issuer` flags since v1.22, where:
- The **first** issuer generates new tokens
- **All** issuers are used to validate existing tokens
- This enables non-disruptive issuer changes per [Kubernetes documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)

## Use Cases

1. **Load Balancer Migration**: Moving from one load balancer to another
2. **DNS Changes**: Updating cluster endpoint DNS without service interruption  
3. **Multi-Region Setup**: Supporting multiple endpoint URLs for the same cluster
4. **Certificate Rotation**: Changing endpoint certificates with different CN/SANs
5. **Infrastructure Migration**: Moving control plane infrastructure

## Benefits

- **Zero-downtime** endpoint migrations
- **Improved operational safety** during infrastructure changes
- **Alignment with Kubernetes best practices**
- **Enhanced cluster reliability** during maintenance operations

## References

- Related issue: [#9609 - Unable to change cluster.endpoint without downtime](https://github.com/siderolabs/talos/issues/9609)
- Kubernetes docs: [kube-apiserver flags](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)
- Service Account documentation: [Managing Service Accounts](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/)

## Implementation Considerations

- Maintain backward compatibility with current single endpoint configuration
- Ensure proper validation of issuer URLs (HTTPS, OIDC compliance)
- Consider configuration precedence (explicit vs. derived from controlPlane.endpoint)
- Integration with existing service account key management

This feature would significantly improve operational workflows for Talos clusters by eliminating forced downtime during common infrastructure operations while following established Kubernetes patterns.

shanduur:
This Pull Request introduces support for specifying multiple values for a single command-line argument within the extraArgs configuration fields. Previously, extraArgs only supported a single string value per key. This change updates the API, configuration handling, and internal logic to allow both single strings (for backward compatibility) and lists of strings.

This affects Etcd, Kubelet, and Kubernetes Control Plane components (APIServer, Controller Manager, Scheduler).

In API Server, passing extra args from `service-account-issuer` will prepend them before default value.

Fixes #11694

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -138,6 +138,12 @@ This includes fields such as `.cluster.apiServer.extraArgs`.
 BREAKING: If you were relying on the resources EtcdConfigs, KubeletConfigs, ControllerManagerConfigs, SchedulerConfigs or APIServerConfigs, the protobuf format has changed from `map<string,string>` to `map<string,message>`.
 """
 
+    [notes.serviceAccountIssuer]
+    title = "Service Account Issuer configuration"
+    description = """\
+In API Server, passing extra args with `service-account-issuer` will prepend them before default value.
+"""
+
 [make_deps]
 
     [make_deps.tools]
--- a/pkg/argsbuilder/argsbuilder_args.go
+++ b/pkg/argsbuilder/argsbuilder_args.go
@@ -53,6 +53,18 @@ func (a Args) Merge(args Args, setters ...MergeOption) error {
 		case MergeOverwrite:
 			a[key] = slices.Clone(val)
 
+		case MergeAppend:
+			existing := make([]string, 0, len(val)+len(a[key]))
+			existing = append(existing, a[key]...)
+			existing = append(existing, val...)
+			a[key] = existing
+
+		case MergePrepend:
+			existing := make([]string, 0, len(val)+len(a[key]))
+			existing = append(existing, val...)
+			existing = append(existing, a[key]...)
+			a[key] = existing
+
 		case MergeAdditive:
 			// 1. Join the existing []string slice into one string so we can Split it.
 			//    This handles cases where a[key] might be ["a", "b"] or ["a,b"].
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -429,6 +429,7 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 		"etcd-keyfile":                     argsbuilder.MergeDenied,
 		"kubelet-client-certificate":       argsbuilder.MergeDenied,
 		"kubelet-client-key":               argsbuilder.MergeDenied,
+		"service-account-issuer":           argsbuilder.MergePrepend,
 		"service-account-key-file":         argsbuilder.MergeDenied,
 		"service-account-signing-key-file": argsbuilder.MergeDenied,
 		"tls-cert-file":                    argsbuilder.MergeDenied,
--- a/pkg/argsbuilder/argsbuilder_interface.go
+++ b/pkg/argsbuilder/argsbuilder_interface.go
@@ -14,6 +14,10 @@ const (
 	MergeAdditive
 	// MergeDenied fail merge if another object has the arg defined.
 	MergeDenied
+	// MergePrepend prepends new values before existing ones.
+	MergePrepend
+	// MergeAppend appends new values after existing ones.
+	MergeAppend
 )
 
 // MergePolicies merge policy map.
--- a/pkg/argsbuilder/argsbuilder_test.go
+++ b/pkg/argsbuilder/argsbuilder_test.go
@@ -23,9 +23,10 @@ func (suite *ArgsbuilderSuite) TestMergeAdditive() {
 	}
 
 	suite.Require().NoError(
-		args.Merge(argsbuilder.Args{
-			"param": {"value2, value10"},
-		},
+		args.Merge(
+			argsbuilder.Args{
+				"param": {"value2, value10"},
+			},
 			argsbuilder.WithMergePolicies(argsbuilder.MergePolicies{
 				"param": argsbuilder.MergeAdditive,
 			}),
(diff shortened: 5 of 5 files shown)


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -43,6 +43,98 @@ func init() {
 	})
 }
 
+// Args represents a map of argument names to their values.
+// docgen:nodoc
+type Args map[string]ArgValue
+
+// ToMap converts Args to a map of string slices.
+func (a *Args) ToMap() map[string][]string {
```
smira:
```suggestion
func (a Args) ToMap() map[string][]string {
```

pkg/machinery/config/types/v1alpha1/v1alpha1_etcdconfig.go
```diff
@@ -29,12 +29,12 @@ func (e *EtcdConfig) CA() *x509.PEMEncodedCertificateAndKey {
 }
 
 // ExtraArgs implements the config.Etcd interface.
-func (e *EtcdConfig) ExtraArgs() map[string]string {
+func (e *EtcdConfig) ExtraArgs() map[string][]string {
 	if e.EtcdExtraArgs == nil {
```
smira: drop the if, `ToMap()` works on `nil` values

api/buf.yaml
```diff
@@ -27,3 +27,7 @@ lint:
 breaking:
   use:
     - FILE
+  ignore_only:
+    FIELD_SAME_TYPE: # NOTE: this should be temporary, right?
+      - resource/definitions/etcd/
+      - resource/definitions/k8s/
```
shanduur: @smira what about this?

smira: just do `make api-descriptors` to say "I did a breaking change, and I know it" :)

hack/release.toml
```diff
@@ -127,6 +127,15 @@ A new `KubeSpanConfig` document has been introduced to configure KubeSpan settin
...
 The old configuration field will continue to work for backward compatibility.
+"""
+
+    [notes.extraArgs]
+        title = "Extra Arguments accept slices in addition to strings"
+        description = """\
+Several Talos configuration fields that previously accepted single string values for extra arguments have been updated to accept slices of strings as well.
+This includes fields such as `.cluster.apiServer.extraArgs`.
+
+BREAKING: If you were relying on the EtcdConfigs, KubeletConfigs, ControllerManagerConfigs, SchedulerConfigs or APIServerConfigs, the protobuf format has changed from `map<string,string>` to `map<string,message>`.
```
smira:
```suggestion
BREAKING: If you were relying on the resources EtcdConfigs, KubeletConfigs, ControllerManagerConfigs, SchedulerConfigs or APIServerConfigs, the protobuf format has changed from `map<string,string>` to `map<string,message>`.
```

pkg/argsbuilder/argsbuilder_args.go
```diff
@@ -49,41 +49,57 @@ func (a Args) Merge(args Args, setters ...MergeOption) error {
 		switch policy {
 		case MergeDenied:
 			return NewDenylistError(key)
+
+		case MergeOverwrite:
+			a[key] = slices.Clone(val)
+
 		case MergeAdditive:
```
smira, line 68: we don't have "MergeAppend" yet, so this is going to be a separate PR?

smira:
I would also pass it through `integration/provision` to test various K8s upgrades

pkg/argsbuilder/argsbuilder_args.go
shanduur: We do not actually need it (unless we do for some future flags).

I think the current setup with `MergeOverwrite` works as expected for #11694.

smira: I think this needs to be documented, and we don't have any tests for it.

My point if the Overwrite removes default issuer, it's fine, but it should be documented that it does from my point of view, and we set it automatically, and "extra arg" would completely nuke automatic value.

smira: so basically not setting the first value as the controlplane endpoint would render all service account tokens invalid

oscrx:
Is this included in 1.12.2?

There might be a regression of https://github.com/siderolabs/talos/issues/2120
When I update from 1.12.1 etcd doesn't start anymore on my ipv6 only cluster

```
2a02:a470:edcd:0:be24:11ff:fe6c:6e3d: {"level":"fatal","ts":"2026-01-22T18:16:07.335868Z","caller":"etcdmain/etcd.go:183","msg":"discovery failed","error":"--initial-advertise-peer-urls \"http://2a02:a470:edcd:0:be24:11ff:fe6c:6e3d:2380\" must be \"host:port\" (address 2a02:a470:edcd:0:be24:11ff:fe6c:6e3d:2380: too many colons in address)","stacktrace":"go.etcd.io/etcd/server/v3/etcdmain.startEtcdOrProxyV2\n\tgo.etcd.io/etcd/server/v3/etcdmain/etcd.go:183\ngo.etcd.io/etcd/server/v3/etcdmain.Main\n\tgo.etcd.io/etcd/server/v3/etcdmain/main.go:40\nmain.main\n\tgo.etcd.io/etcd/server/v3/main.go:31\nruntime.main\n\truntime/proc.go:283"}
```

smira:
@oscrx can you please create a bug report with details?

Issue #12492, opened 2025-12-26
drop support for boards (pre-overlays) in Talos and Image Factory


smira:
See also #12146.

In Talos, drop all support for boards: board code, all handling in the imager paths, installer, etc. There is no effect for that in Talos itself.

In Image Factory:

* Image Factory will no longer build any kind of image for boards (Talos < 1.7).
* deny building any image with schematic referencing a board
* deny creating a schematic with board
* in the UI, drop the board flow - for Talos < 1.7, if someone selects "SBC" in the wizard, show the message on the next screen that only Talos 1.7 is supported.

PR #12494, opened 2025-12-26
fix: drop SBC board code

frezbo:
Boards were deprecated in favor of overlays from Talos 1.7.

Now completely remove all board specific code.

Part of: #12492


pkg/machinery/platforms/sbcs.go
```diff
@@ -1,322 +0,0 @@
...
-// DiskImagePath returns the path to the disk image for the SBC.
-func (s SBC) DiskImagePath(talosVersion string) string {
-	if quirks.New(talosVersion).SupportsOverlay() {
-		return "metal-arm64.raw.xz"
-	}
-
-	return "metal-" + s.BoardName + "-arm64.raw.xz"
-}
-
-// SBCs returns a list of supported Single Board Computers.
```
smira, line 39: I think this should stay, it is used in the Image Factory and Omni to build the SBC list. We can drop the BoardName though

frezbo: ah yes, I forgot we moved it back

PR #12502, opened 2025-12-27
feat: imager usernamespace

### LINKED ISSUE #12498: Imager run as non-root
## Feature Request

Imager should work running as non-root. Filesystems should be populated with root owned files only.

frezbo:
If imager detecs running as non-root re-exec in a user namespace mapper to current user so that files are owned by root.

Fixes: #12498


frezbo:
keeping in draft since not sure if we need this

frezbo:
Can be tested by adding `--security-opt seccomp=unconfined --user=1000` to our `image-` target

frezbo:
after internal discussion we'll use a tarball for extfs ad use protofiles for xfs

Issue #12437, opened 2025-12-17
Move `.machine.network.kubespan` into `KubeSpanConfig`


smira:
Move into new multi-doc document, deprecate and hide fully `.machine.network`.

For reference, existing config: https://docs.siderolabs.com/talos/v1.12/reference/configuration/v1alpha1/config#network

```yaml
apiVersion: v1alpha1
kind: KubeSpanConfig
enabled: true
advertiseKubernetesNetworks: false # default value
allowDownPeerBypass: false # default value
harvestExtraEndpoints: false # default value
mtu: 1420 # default value
filters:
  endpoints: ... # optional
```

PR #12504, opened 2025-12-28
feat: implement KubeSpan multi-document configuration

pranav767:
Resolves https://github.com/siderolabs/talos/issues/12437

# Pull Request

## What? (description)

- Add `KubeSpanConfigV1Alpha1` document type with registry support
- Implement container aggregation with wrapper pattern
- Deprecate .machine.network.kubespan v1alpha1 path

## Why? (reasoning)

Making/applying config for kubespan easier, 
instead of the current scenario which requires:
```
machine:
    network:
        nameservers:
            - 9.8.7.6
            - 8.7.6.5
        searchDomains:
            - example.org
            - example.com

        # # Configures KubeSpan feature.
        # kubespan:
        #     enabled: true # Enable the KubeSpan feature.
```
We can have independent config for Kubespan like:
```
apiVersion: v1alpha1
kind: KubeSpanConfig
enabled: true
advertiseKubernetesNetworks: false
allowDownPeerBypass: false
mtu: 1420
```

Few more Points for my own sanity:

1. The issue talks about `deprecate and hide fully .machine.network.` , but this still supports the previous config way, this just extends and gives priority to the multi-document config way for Kubespan.
2. Created `pkg/machinery/config/types/network/kubespan.go`  which reads yaml and creates `KubeSpanConfigV1Alpha1` (struct) with all values and implements all interfaces.
3. Updated container logic in `pkg/machinery/config/container/container.go` to use a wrapper-in-wrapper pattern instead of just returning `container.v1alpha1Config.Machine()` helps to intercept controller call on `container.Machine().Network().KubeSpan()` and use the desired (new multi-doc config) from `container.documents`
4. `make generate` Created json and copy stuff.
5. There is one confusing bit which I feel should be changed, from docs we have params as
```
enabled
advertiseKubernetesNetworks
allowDownPeerBypass
harvestExtraEndpoints
mtu
filters
```
and following above pattern we have such structs, but https://github.com/siderolabs/talos/blob/main/api/resource/definitions/kubespan/kubespan.proto has
```
// ConfigSpec describes KubeSpan configuration..
message ConfigSpec {
  bool enabled = 1;
  string cluster_id = 2;
  string shared_secret = 3;
  **bool force_routing = 4;**
  bool advertise_kubernetes_networks = 5;
  uint32 mtu = 6;
  repeated string endpoint_filters = 7;
  bool harvest_extra_endpoints = 8;
  repeated common.NetIPPort extra_endpoints = 9;
}
```
also the config for kubespan controller has
https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/kubespan/config.go#L49
`res.TypedSpec().ForceRouting = c.Machine().Network().KubeSpan().ForceRouting()` **forceRouting** instead of **allowDownPeerBypass**,
which leads to https://github.com/siderolabs/talos/compare/main...pranav767:talos:feat/kubespan-multi-document-config?expand=1#diff-2f6f6aab6dbbcc38f191c5f752866e9d8d3dc787f0044c4eb9d7107ea7a1719dR170
Just feels a bit of inconsistency with the use of one parameter with 2 different names.

## Acceptance


pkg/machinery/config/container/container.go
```diff
@@ -186,6 +190,48 @@ func (container *Container) Cluster() config.ClusterConfig {
 	return container.v1alpha1Config.Cluster()
 }
 
+// machineConfigWrapper wraps MachineConfig to override Network().KubeSpan() with multi-doc aggregation.
+type machineConfigWrapper struct {
```
smira: this approach doesn't scale well. 

we have this pattern: https://github.com/siderolabs/talos/blob/154952175ab73ac65722732b146a0ee1c56b2f4d/pkg/machinery/config/container/container.go#L326-L341

pkg/machinery/config/types/network/kubespan.go
```diff
@@ -0,0 +1,204 @@
...
+	//   description: |
+	//     KubeSpan advanced filtering of network addresses.
+	//     Settings are optional and apply only to this node.
+	ConfigFilters *KubeSpanFiltersConfig `yaml:"filters,omitempty"`
+}
+
+// KubeSpanFiltersConfig configures KubeSpan endpoint filters.
+type KubeSpanFiltersConfig struct {
+	//   description: |
+	//     Filter node addresses which will be advertised as KubeSpan endpoints for peer-to-peer Wireguard connections.
```
smira, line 103: this needs more configuration/examples

pkg/machinery/config/types/network/kubespan.go
```diff
@@ -0,0 +1,204 @@
...
+	return pointer.SafeDeref(s.ConfigHarvestExtraEndpoints)
+}
+
+// MTU implements config.KubeSpan interface.
+func (s *KubeSpanConfigV1Alpha1) MTU() uint32 {
+	if s.ConfigMTU != nil {
+		return *s.ConfigMTU
+	}
+
+	return 1420
```
smira: I _think_ we had a constant for it.

pkg/machinery/config/types/network/kubespan.go
```diff
@@ -0,0 +1,204 @@
...
+	if s.ConfigFilters == nil {
+		return nil
+	}
+
+	return s.ConfigFilters
+}
+
+// Endpoints implements config.KubeSpanFilters interface.
+func (f *KubeSpanFiltersConfig) Endpoints() []string {
+	if f == nil {
```
smira: this can be removed (nil slice is an empty slice).

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
smira, line 661: we can now deprecate the whole `.machine.network` tree, so that it disappears from the docs (it's now empty as everything inside it is deprecated).

smira (commented):
* `talosctl cluster create` has `--with-kubepsan` option which should use new multi-doc for Talos >= 1.13 (need new config version contract flag)
* KubeSpanConfigController should have a new testcase for enabling KubeSpan via new multi-doc

pkg/machinery/config/container/container.go
pranav767: Removed the wrapper in wrapper pattern, and followed existing one. :+1:

pkg/machinery/config/types/network/kubespan.go
pranav767: Updated :+1:

pkg/machinery/config/types/network/kubespan.go
pranav767: yes, found it `KubeSpanLinkMTU`

pkg/machinery/config/types/network/kubespan.go
pranav767: removed :+1:

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
pranav767: Updated with `docgen: nodoc`

pranav767:
@smira 
- There were two talosctl commands (`talosctl cluster create` and `talosctl gen config`) using `--with-kubespan`. Both have been updated to use the new multi-document configuration format for Talos ≥ 1.13.
- `KubeSpanConfigController` now includes a new test case for the multi-config–style KubeSpan.

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
```diff
@@ -451,11 +467,14 @@ func (m *Maker[T]) initGenOps() error {
 	}
 
 	if m.Ops.EnableKubeSpan {
-		genOptions = slices.Concat(genOptions,
-			[]generate.Option{generate.WithNetworkOptions(
-				v1alpha1.WithKubeSpan(),
-			)},
```
pranav767, line 457: I'm aware that we're migrating `.machine.network` to multi-document format, but I've kept legacy `v1alpha1` config generation for `Talos < 1.13` as previous versions can't parse new `KubeSpanConfigV1Alpha1`

smira:
Thank you for your work, I'm going to refactor some bits, and merge this PR afterwards.

pranav767:
Great, Thank you for the review @smira 🙌

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
smira: I re-did this flow to pin it more into the generate package.

smira:
Notes for myself:

- [x] fix linting issues
- [x] refactor KubeSpan controller test (modernize)
- [x] add hack/release.toml note
- [x] more tests (?)

Issue #12257, opened 2025-11-19
talosctl images default should accept --version


rothgar:
## Feature Request

Talos supports multiple versions of deploying Kubernetes but the `talosctl images default` command only lists the latest version for that version of talosctl. Similar to `talosctl image source-bundle` it should accept a version of kubernetes that will list the images for that version of kubernetes.

### Description

We may want to rename the commands so they make more sense on what list of containers you're getting. It would be helpful to have a way to list what versions (talos and kubernetes) are valid options for the installed talosctl.

Maybe the flow should be

```
talosctl image talos-bundle [$VERSION] > images.txt
talosctl image k8s-bundle [$VERSION] >> images.txt
cat images.txt | talosctl images cache-create --image-cache-path ./image-cache.oci --images=-
```
Then serving the cache would be
```
talosctl image cache-cert-gen --advertised-address $IP:$PORT
talosctl image cache-serve --address $IP --image-cache-path ./image-cache.oci \
  --tls-cert-file tls.crt --tls-key-file tls.key
```
Then I have a `image-cache-patch.yaml` patch I can apply to a system. Is it possible to also generate a registry-mirror patch automatically with the endpoints and IP address? I _think_ I need to add this patch.
```
machine:
    registries:
        mirrors:
            ghcr.io:
                endpoints:
                    - https://$IP:$PORT
                    - https://ghcr.io/v2/
            registry.k8s.io:
                endpoints:
                    - https://$IP:$PORT
                    - https://registry.k8s.io/v2/
```

rothgar:
I just tried the above steps on a node without internet access (it could only access a local LAN with my machine).

I applied the image-cache-patch with certificate information
```
talosctl apply -f image-cache-patch.yaml -i -n 192.168.10.9
```
And then generated a config and a mirrors patch
```
machine:
  time:
    disabled: true
  registries:
    mirrors:
      ghcr.io:
        endpoints:
          - https://192.168.10.19:5009
          - https://ghcr.io/v2/
      registry.k8s.io:
        endpoints:
          - https://192.168.10.19:5009
          - https://registry.k8s.io/v2/
```
I then applied the config and patch to the machine
```
talosctl apply -f controlplane.yaml -p '@mirrors.yaml' -i -n 192.168.10.9
```
I can see in the cache-serve logs 
```
2025/11/19 15:44:03 http: TLS handshake error from 192.168.10.9:46966: remote error: tls: bad certificate
```
[...]

PR #12512, opened 2025-12-29
feat: add k8s-version parameter to k8s-bundle

shanduur:
Allow overriding K8s version in the  command.

Closes https://github.com/siderolabs/talos/issues/12257

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -138,36 +138,56 @@ var imagePullCmd = &cobra.Command{
 	},
 }
 
-// imageDefaultCmd represents the image k8s-bundle command.
-var imageDefaultCmd = &cobra.Command{
+var imageK8sBundleCmdFlags = struct {
+	k8sVersion     pflag.Value
+	flannelVersion pflag.Value
+	corednsVersion pflag.Value
+	etcdVersion    pflag.Value
+}{
+	k8sVersion:     helpers.Semver(constants.DefaultKubernetesVersion),
+	flannelVersion: helpers.Semver(constants.FlannelVersion),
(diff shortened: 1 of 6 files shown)


cmd/talosctl/cmd/talos/image.go
```diff
@@ -140,34 +140,70 @@ var imagePullCmd = &cobra.Command{
...
+					ProxyConfig:             &v1alpha1.ProxyConfig{},
+				},
+			}),
+			images.VersionsListOptions{
+				KubernetesVersion: k8sVersion,
 			},
-		}))
+		)
+
+		// if k8sVersion is defined, then we don't print components versions - there's no simple way to correlate patch-to-patch.
```
smira: these versions are set by Talos, and they don't depend on Kubernetes version? 

I'm trying to understand why we don't print them

shanduur: I wonder if we should be printing them at all here - should those be part of `talos-bundle`? If so, how do we ensure that we print correct version when overriding Talos version?

website/content/v1.13/reference/cli.md
```diff
@@ -2215,8 +2215,11 @@ talosctl image k8s-bundle [flags]
 ### Options
 
 ```
-  -h, --help                 help for k8s-bundle
-      --provisioner string   include provisioner specific images (default "installer")
```
frezbo: this got dropped?

frezbo (approved):

website/content/v1.13/reference/cli.md
shanduur: Yes, it was no-op after the changes made by Justin in https://github.com/siderolabs/talos/pull/12314

PR #12514, opened 2025-12-29
fix: update VIP config example

smira:
The "name" should be the IP address.

See https://github.com/siderolabs/docs/pull/256

### DIFF
--- a/website/content/v1.13/reference/configuration/network/layer2vipconfig.md
+++ b/website/content/v1.13/reference/configuration/network/layer2vipconfig.md
@@ -20,7 +20,7 @@ title: Layer2VIPConfig
 {{< highlight yaml >}}
 apiVersion: v1alpha1
 kind: Layer2VIPConfig
-name: int0 # IP address to be advertised as a Layer 2 VIP.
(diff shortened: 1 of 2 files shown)


shanduur (approved):

website/content/v1.13/reference/configuration/network/layer2vipconfig.md
```diff
@@ -21,7 +21,7 @@ title: Layer2VIPConfig
 apiVersion: v1alpha1
 kind: Layer2VIPConfig
 name: int0 # IP address to be advertised as a Layer 2 VIP.
-link: enp0s2 # Name of the link to assign the VIP to.
+link: 10.3.0.1 # Name of the link to assign the VIP to.
```
TimJones: This now reads even more confusing...
`name` comment suggests the name should be an IP address but the example gives `int0`
`link` comment suggests it should be a named link but example now gives an IP.

smira: yes, you're right, thank you, I put it completely reverse order now 🤦

Issue #12309, opened 2025-12-01
support for LVM volumes


smira:
Support for provisioning, configuration, management, destruction of LVM volumes.

We might be opinionated and support a subset of LVM features.

xvzf:
This would be amazing. So far we've been planning to run a static pod, provisioning LVM volumes once

gunzy83:
It would be good to be able to provision up to a volume group (VG) so that a CSI like Topolvm or OpenEBS could take over from there.

bbinet:
Given the ongoing interest from the community (see the reactions above), this remains a relevant and wanted feature. Please keep this issue open.

PR #12530, opened 2026-01-01
Setup resource types and proto for lvm support 

pranav767:
# Pull Request

Resolves a part of https://github.com/siderolabs/talos/issues/12309

## What? (description)
Setup Resources and Proto 
Updated with following structs:
```
type PhysicalVolumeSpec struct
type VolumeGroupSpec struct
type LogicalVolumeSpec struct
```

## Why? (reasoning)
Setup structs for LVM support ( physical volumes, volume groups and logical volumes)

## Acceptance


pranav767:
I'm still thinking about , https://github.com/siderolabs/talos/issues/12309#issuecomment-3619158741
If the resources and structs are correct, next task would be setting config , I'd say having a separate config than volumeconfig would be better for lvm. Few questions though
1. Do we include / keep optional / don't include LV's in lvmconfig? 
2. If we include lv's in config who takes authority/precedence (talos/csi) (If talos takes precedence we'd still have to think/write about taking care of lv's from volume groups)

Don't know much about how CSI works with handling lv's (still reading)
We could at the very least setup config which has physical volume and volume groups and think about CSI part during controller 
Can't think straight about how things will work :thinking:  , If I get a clear picture, maybe I can be definite about the structs and config

pranav767:
After playing around a bit with the following design where:

- Talos handles PhysicalVolumes and VolumeGroups
- CSI (TopoLVM) takes over management of Logical Volumes

Talos can easily manage PVs and VGs. I tried setting up configs and controllers for both, and they worked fine
```
:~/siderolabs/talos$ _out/talosctl-linux-amd64 -n 172.20.0.3 get pv
NODE         NAMESPACE   TYPE             ID         VERSION   DEVICE     VG              SIZE          UUID                                     ALLOCATED   TOTAL
172.20.0.3   runtime     PhysicalVolume   /dev/sda   1         /dev/sda   vg-talos-data   10733223936   kA22hY-zhcd-7Z7s-VD8l-dJXR-h3FQ-F3Z8Qq               2559


:~/siderolabs/talos$ _out/talosctl-linux-amd64 -n 172.20.0.3 get Vg
NODE         NAMESPACE   TYPE          ID              VERSION   NAME            UUID                                     TOTALSIZE     FREESIZE      LVCOUNT   PVCOUNT
172.20.0.3   runtime     VolumeGroup   vg-talos-data   1         vg-talos-data   MHLVoA-ZQMe-KNV5-monU-oOOD-CEsj-SpPraB   10733223936   10733223936             1

```

However, when handing this off to TopoLVM, the container / DaemonSet / pod is expected to perform operations such as:
```
lvm vgs
lvm lvcreate
lvm lvremove
```
These commands:

- Interact directly with `/dev`
- Modify host disk state
- Create and manage sockets on the host

To make this work, the CSI pod must:

- Mount `/dev`
- Run as a privileged container
- Execute host LVM binaries

Which feels like giving CSI pod more than enough powers
I tried mounting the host binary `/usr/sbin/lvm` into the container, but consistently ran into errors such as:
```
permission denied
no such file or directory
exec not found
binary exists but cannot execute
```

While it may still be possible to have CSI manage Logical Volumes, this approach feels
- complex for users to configure correctly
- against what talos has (giving csi more powers to manipulate host instead of just having talos controllers those powers)

We can have logical volumes handled via talos, that might increase complexity within talos but I believe it would be more talos friendly (having api's) do your work.

Let me know what would be the best approach to proceed.

davidharkis:
> I'm still thinking about , [#12309 (comment)](https://github.com/siderolabs/talos/issues/12309#issuecomment-3619158741) If the resources and structs are correct, next task would be setting config , I'd say having a separate config than volumeconfig would be better for lvm. Few questions though
> 
> 1. Do we include / keep optional / don't include LV's in lvmconfig?
> 2. If we include lv's in config who takes authority/precedence (talos/csi) (If talos takes precedence we'd still have to think/write about taking care of lv's from volume groups)
> 
> Don't know much about how CSI works with handling lv's (still reading) We could at the very least setup config which has physical volume and volume groups and think about CSI part during controller Can't think straight about how things will work 🤔 , If I get a clear picture, maybe I can be definite about the structs and config

As a user of Talos I've been following topics related to volume management with interest.

What you are describing is possible; however, a CSI driver does not mount or execute the host lvm binary.
Instead, LVM is executed from within the CSI container itself, running container-provided LVM tools that operate on the host’s block devices via /dev.

This can be replicated with a privileged pod that includes `lvm2` and mounts host dev and LVM runtime paths

In fact, the [csi-driver-lvm](https://github.com/metal-stack/csi-driver-lvm/tree/master) project carries this out end-to-end, including the creation VolumeGroups on existing versions of Talos. (I've verified this works on v1.11).

pranav767:
Yes, this would work, but my concern is whether it aligns with Talos’ philosophy.
To enable this, the CSI pod must 

- run privileged
- adjust security labels 
- mount host paths like `/dev` 

which effectively gives it root-level access to the host from within the container.
I’m worried this may go against Talos’ design principles.
If there are existing cases where similar things are done in Talos, it wouldn’t be a problem.

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/resource/meta"
+	"github.com/cosi-project/runtime/pkg/resource/protobuf"
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// PhysicalVolumeType is type of PhysicalVolume resource.
+const PhysicalVolumeType = resource.Type("PhysicalVolumes.block.talos.dev")
```
shanduur: I'd generally prefer to see one type per file.

Also, this probably should be namespaced a little bit more, so users understand that `PhysicalVolumes` are referring to LVM Physical Volumes just by name. E.g. `LVMPhysicalVolumes`.

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+
+// VolumeGroup represents an LVM volume group resource.
+type VolumeGroup = typed.Resource[VolumeGroupSpec, VolumeGroupExtension]
+
+// VolumeGroupSpec describes an LVM Volume Group.
+//
+//gotagsrewrite:gen
+type VolumeGroupSpec struct {
+	Name            string   `yaml:"name" protobuf:"1"`
+	PhysicalVolumes []string `yaml:"physicalVolumes" protobuf:"2"`
```
shanduur: Isn't this a circular dependency between VG and PV?

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+
+// PhysicalVolume represents an LVM physical volume resource.
+type PhysicalVolume = typed.Resource[PhysicalVolumeSpec, PhysicalVolumeExtension]
+
+// PhysicalVolumeSpec describes an LVM Physical Volume.
+//
+//gotagsrewrite:gen
+type PhysicalVolumeSpec struct {
+	DevicePath       string `yaml:"devicePath" protobuf:"1"`
+	VGName           string `yaml:"vgName" protobuf:"2"`
```
shanduur: Can you name this as `VolumeGroupName`?

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+//
+//gotagsrewrite:gen
+type VolumeGroupSpec struct {
+	Name            string   `yaml:"name" protobuf:"1"`
+	PhysicalVolumes []string `yaml:"physicalVolumes" protobuf:"2"`
+	TotalSize       uint64   `yaml:"totalSize" protobuf:"3"`
+	FreeSize        uint64   `yaml:"freeSize" protobuf:"4"`
+	UUID            string   `yaml:"uuid" protobuf:"5"`
+	ExtentSize      uint64   `yaml:"extentSize,omitempty" protobuf:"6"`
+	LVCount         uint32   `yaml:"lvCount,omitempty" protobuf:"7"`
```
shanduur: Can you name this as `LogicalVolumesCount`?

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+	PVCount         uint32   `yaml:"pvCount,omitempty" protobuf:"8"`
```
shanduur: Can you name this as `PhysicalVolumesCount`?

pkg/machinery/resources/block/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+
+// LogicalVolume represents an LVM logical volume resource.
+type LogicalVolume = typed.Resource[LogicalVolumeSpec, LogicalVolumeExtension]
+
+// LogicalVolumeSpec describes an LVM Logical Volume.
+//
+//gotagsrewrite:gen
+type LogicalVolumeSpec struct {
+	Name       string `yaml:"name" protobuf:"1"`
+	VGName     string `yaml:"vgName" protobuf:"2"`
```
shanduur: Can you name this as `VolumeGroupName`?

pkg/machinery/resources/block/volumetype.go
```diff
@@ -11,11 +11,14 @@ type VolumeType int
...
+	VolumeTypePartition      VolumeType = iota // partition
+	VolumeTypeDisk                             // disk
+	VolumeTypeTmpfs                            // tmpfs
+	VolumeTypeDirectory                        // directory
+	VolumeTypeSymlink                          // symlink
+	VolumeTypeOverlay                          // overlay
+	VolumeTypeExternal                         // external
+	VolumeTypePhysicalVolume                   // physical volume
+	VolumeTypeVolumeGroup                      // volume group
+	VolumeTypeLogicalVolume                    // logical volume
```
shanduur: Same as above, this needs more clear namespacing, by including the `LVM`.

shanduur (commented):
What are next steps? Do you want to create controllers in the same PR, or...?

pranav767:
@shanduur , Anything works.
I believe next changes would be setting up machine config (LVM configs in this case) something like `pkg/machinery/config/types/block/lvm_volume_config.go` and then having controllers.
If it would not be difficult for you to review, I'd raise in the same PR.

Can you help me answer this https://github.com/siderolabs/talos/pull/12530#issuecomment-3708515902, There can be 2 approaches 
1. Talos manages both Physical Volume & Volume groups. CSI manages Logical Volumes
2. Talos manages everything.

Depending on the answer we'd have to setup configs for LVM.

frezbo:
> @shanduur , Anything works. I believe next changes would be setting up machine config (LVM configs in this case) something like `pkg/machinery/config/types/block/lvm_volume_config.go` and then having controllers. If it would not be difficult for you to review, I'd raise in the same PR.
> 
> Can you help me answer this [#12530 (comment)](https://github.com/siderolabs/talos/pull/12530#issuecomment-3708515902), There can be 2 approaches
> 
>     1. Talos manages both Physical Volume & Volume groups. CSI manages Logical Volumes
> 
>     2. Talos manages everything.
> 
> 
> Depending on the answer we'd have to setup configs for LVM.

I think talos can manage all of pv,vg and lv and users can chose what talos manages and what CSI manages

pranav767:
Alright, I think I was worried too much about how things will work when there will be conflicts, i.e. having both Logical volumes in configs and telling CSI to manage stuff (conflicts between talos and CSI).

Can just add validations to avoid those scenarios.

pkg/machinery/resources/block/lvm.go
pranav767: Updated with 3 different files :+1:

pkg/machinery/resources/block/lvm.go
pranav767: I don't think there would be circular dependency here, we are not embedding and structs/packages.
`LVMVolumeGroupSpec` uses `[] string` for `PhysicalVolumes` and `LVMPhysicalVolumeSpec` uses a string for `VolumeGroupName`, both are just string identifiers and not actual type references

pkg/machinery/resources/block/volumetype.go
pranav767: Updated :+1:

pkg/machinery/resources/block/lvm.go
pranav767: Updated :+1:

pkg/machinery/resources/block/lvm.go
pranav767: Updated :+1:

pkg/machinery/resources/block/lvm.go
pranav767: Updated :+1:

pkg/machinery/resources/block/lvm.go
pranav767: Updated :+1:

shanduur:
Sorry I didn't respond earlier, there was a bank holiday here in Poland.

I agree with Noel, Talos API should be self sufficient, and allow management of vg, pv and lv through Talos API, but should allow managing them also outside of Talos - e.g. via CSI. One thing that needs to be clarified (and IMHO it should be a starting point for designing the LVM API), is to detect the LVM objects created by the external entities.

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+	//   description: |
+	//     Physical volumes to create.
+	PhysicalVolumes []LVMPhysicalVolumeConfig `yaml:"physicalVolumes,omitempty"`
+	//   description: |
+	//     Volume groups to create.
+	VolumeGroups []LVMVolumeGroupConfig `yaml:"volumeGroups,omitempty"`
+	//   description: |
+	//     Logical volumes to create.
+	LogicalVolumes []LVMLogicalVolumeConfig `yaml:"logicalVolumes,omitempty"`
+}
```
shanduur: Note: If a LV, VG, PV is removed from the config, should it be deleted, whole thing should be recreated, or...?

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+		validationErrors error
+	)
+
+	// Validate config name
+	if s.MetaName == "" {
+		validationErrors = errors.Join(validationErrors, errors.New("name is required"))
+	}
+
+	// Track used devices and names to detect duplicates
+	pvDevices := make(map[string]bool)
```
shanduur: nit: No need to have a bool here:

```suggestion
	pvDevices := make(map[string]struct{})
```

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+	pvNames := make(map[string]string)
```
shanduur: nit: same here, we do not use value anywhere:

```suggestion
	pvNames := make(map[string]struct{})
```

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+			if _, exists := pvNames[pv.Name]; exists {
+				validationErrors = errors.Join(validationErrors, fmt.Errorf("physical volume %q: duplicate name", pv.Name))
+			}
+
+			pvNames[pv.Name] = pv.Device
+		}
+	}
+
+	// Validate volume groups
+	vgNames := make(map[string]*LVMVolumeGroupConfig)
```
shanduur: nit: same here, we do not use value anywhere:

```suggestion
	vgNames := make(map[string]struct{})
```

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+			if _, isDevice := pvDevices[pvRef]; !isDevice {
+				if _, isName := pvNames[pvRef]; !isName {
+					validationErrors = errors.Join(validationErrors, fmt.Errorf("volume group %q: physical volume %q not found", vg.Name, pvRef))
+				}
+			}
+		}
+	}
+
+	// Validate logical volumes
+	lvNamesInVG := make(map[string]map[string]bool)
```
shanduur: nit: same here, we do not use value anywhere:

```suggestion
	lvNamesInVG := make(map[string]map[string]struct{})
```

pkg/machinery/resources/block/lvm.go
shanduur: I meant from the logical standpoint - we will be tracking references by name to other objects in two controllers. Will we be able to detect this details from existing PV/VG created by a workload?

pranav767:
> Sorry I didn't respond earlier, there was a bank holiday here in Poland.
> 
> I agree with Noel, Talos API should be self sufficient, and allow management of vg, pv and lv through Talos API, but should allow managing them also outside of Talos - e.g. via CSI. One thing that needs to be clarified (and IMHO it should be a starting point for designing the LVM API), is to detect the LVM objects created by the external entities.

No problem, yes I got the gist, and I think this might work let me know how you feel about this:

A user can have a LVM config like:
```
physicalVolumes:
  - /dev/sda
  - /dev/sdb
  - /dev/sdc

volumeGroups:
  - name: vg-system
    physicalVolumes:
      - /dev/sda
  - name: vg-apps
    physicalVolumes:
      - /dev/sdb
      - /dev/sdc

logicalVolumes:
  - name: lv-etcd
    volumeGroup: vg-system
    size: 5Gi

csiVolumeGroups:
  - name: vg-apps
```
`LogicalVolumes` can be optional but necessary if talos managed.

For all created resources, we can attach a metadata label such as:
`managed-by: talos | CSI` 
like:
```
~/siderolabs/talos$ _out/talosctl-linux-amd64 -n 172.20.0.3 get LVMLogicalVolumes -o yaml
node: 172.20.0.3
metadata:
    namespace: runtime
    type: LVMLogicalVolumes.block.talos.dev
    id: vg-talos-data-lv-data
    version: 1
    owner: block.LVMConfigController
    phase: running
    created: 2026-01-07T09:38:54Z
    updated: 2026-01-07T09:38:54Z
    labels:
        talos.dev/managed-by: talos
```

1. Talos-managed LVM (PV, VG, LV)

- Talos controllers manage the complete LVM lifecycle.
- Talos APIs expose PV, VG, and LV resources.
- A discovery/reconciliation step ensures Talos can distinguish between:
   - resources managed by Talos
   - resources managed externally (e.g. CSI) (skip making LV's if user opts for CSI)
- The managed-by label is used to avoid conflicts and accidental deletion.
-  Yes, I think if LV's in talos-managed are missing those are supposed to be deleted (just thinking this does not cause accidental deletes :thinking: )

2. CSI managed local volumes
-  User creates PV and VG using above talos config, then create manifests asking for storage via pv & pvc, then use this in a pod.
-  To avoid running the CSI node plugin as a privileged container:
    - lvmd runs on the Talos host ( have a PR for missing lvmd in pkgs repo)
    - CSI communicates with lvmd over gRPC
    - lvmd executes LVM commands using the host-installed lvm binary


I should've moved this to a draft PR :thinking: , already had changes for label/discovery bit, was trying CSI stuff.
What do you think about discovery through labels (metadata) is that good enough (as talos would be responsible for initial stuff PV & VG, we can identify in early stages,, while creation).

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+	//   examples:
+	//     - value: '"/var/lib/postgresql"'
+	Path string `yaml:"path"`
+	//   description: |
+	//     Mount options.
+	Options []string `yaml:"options,omitempty"`
+	//   description: |
+	//     Mount the volume as read-only.
+	ReadOnly bool `yaml:"readOnly,omitempty"`
+}
```
shanduur: You should not specify mount spec here, especially the path. Path will be handled by Talos controller, and will always be a `/var/mnt/<name>`, where `name` is constructed as `<prefix> + $Spec.MetaName`. 

I am also not a fan of opaque parameters in form of mount options - if an option is desired, we should put it by name here.

Take a look at how we are handling the MountSpec in e.g. [ExistingVolumeConfig](https://docs.siderolabs.com/talos/v1.12/reference/configuration/block/existingvolumeconfig#mount) or ExternalVolumeConfig.

pkg/machinery/config/types/block/lvm_config.go
```diff
@@ -0,0 +1,393 @@
...
+
+// LVMVolumeGroupConfig describes an LVM volume group configuration.
+type LVMVolumeGroupConfig struct {
+	//   description: |
+	//     Name of the volume group.
+	Name string `yaml:"name"`
+	//   description: |
+	//     Physical volumes to include in this volume group.
+	PhysicalVolumes []string `yaml:"physicalVolumes"`
```
shanduur: Instead of putting names here, this should be a `cel.Expression`, similar to [Provisioning field in UserVolumeConfig](https://docs.siderolabs.com/talos/v1.12/reference/configuration/block/uservolumeconfig#provisioning).

shanduur (commented):
I do not like this design - it feels a little bit clunky.

If you generate the docs, the document for the LVM will look like this:

<details>
<summary>LVMVolumeConfig</summary>

```yaml
apiVersion: v1alpha1
kind: LVMVolumeConfig
name: database-storage # Name of the LVM configuration.
# Physical volumes to create.
physicalVolumes:
    - device: /dev/sdb # Device path for the physical volume.
# Volume groups to create.
volumeGroups:
    - name: data-vg # Name of the volume group.
      # Physical volumes to include in this volume group.
      physicalVolumes:
        - /dev/sdb

      # # Extent size for the volume group.
      # extentSize: 4MiB
      # extentSize: 8MiB
# Logical volumes to create.
logicalVolumes:
    - name: data-lv # Name of the logical volume.
      volumeGroup: data-vg # Volume group name this logical volume belongs to.
      size: 100GB # Size of the logical volume.
      # Filesystem configuration for the logical volume.
      filesystem:
        type: xfs # Filesystem type. Default is `xfs`.
      # Mount configuration for the logical volume.
      mount:
        path: /var/lib/data # Path where the volume should be mounted.

      # # Encryption configuration for the logical volume.
      # encryption:
      #     provider: luks2 # Encryption provider to use for the encryption.
      #     # Defines the encryption keys generation and storage method.
      #     keys:
      #         - slot: 0 # Key slot number for LUKS2 encryption.
      #           # Key which value is stored in the configuration file.
      #           static:
      #             passphrase: exampleKey # Defines the static passphrase value.
      #
      #           # # KMS managed encryption key.
      #           # kms:
      #           #     endpoint: https://192.168.88.21:4443 # KMS endpoint to Seal/Unseal the key.
      #         - slot: 1 # Key slot number for LUKS2 encryption.
      #           # KMS managed encryption key.
      #           kms:
      #             endpoint: https://example-kms-endpoint.com # KMS endpoint to Seal/Unseal the key.
      #     cipher: aes-xts-plain64 # Cipher to use for the encryption. Depends on the encryption provider.
      #     blockSize: 4096 # Defines the encryption sector size.
      #     # Additional --perf parameters for the LUKS2 encryption.
      #     options:
      #         - no_read_workqueue
      #         - no_write_workqueue
```

</details>

I think it does not represent the LVM relationships between primitives well.

Splitting the resources in the `v1alpha1` will give us better control over them?

```yaml
---
apiVersion: v1alpha1
kind: LVMVolumeGroupConfig
name: vg-pool
physicalVolumes:
    volumeSelector:
        match: 'disk.dev_path == /dev/sda0'
---
apiVersion: v1alpha1
kind: LVMLogicalVolumeConfig
name: lv-data
lvType: linear
provisioning:
    volumeGroup: vg-pool
    maxSize: 50GiB # optional, default 100%
    # minSize: 2.5GiB
---
apiVersion: v1alpha1
kind: UserVolumeConfig
name: lv-flex-data
volumeType: partition
provisioning:
    diskSelector:
        match: 'disk.dev_path == "/dev/vg-pool/lv-data"'
# ...
```

pkg/machinery/resources/block/volumetype.go
```diff
@@ -11,11 +11,14 @@ type VolumeType int
...
-	VolumeTypeOverlay                     // overlay
-	VolumeTypeExternal                    // external
+	VolumeTypePartition         VolumeType = iota // partition
+	VolumeTypeDisk                                // disk
+	VolumeTypeTmpfs                               // tmpfs
+	VolumeTypeDirectory                           // directory
+	VolumeTypeSymlink                             // symlink
+	VolumeTypeOverlay                             // overlay
+	VolumeTypeExternal                            // external
+	LVMVolumeTypePhysicalVolume                   // physical volume
```
smira, line 21: I feel that LVMs are not "volumes" in the way we treat volumes today in Talos.

Talos volumes represent something "mountable", while LVMs are raw disks (before they are properly formatted/used). 

Moreover, I don't think we should start with any implementation work before we land on the "API" of the change (proposal) - what does the machine config look like, how does it work, etc.

pranav767: Yes,also this file talks all about volumes, I think we can have a different file just describing LVM types.
```
// LVMType describes LVM type.
LVMTypePhysicalVolume
LVMTypeVolumeGroup
LVMTypeLogicalVolume
```
Yes, agreed I've been going round and round about implementing this, It would be better to outline how configs look like and basic working.

I liked what @shanduur mentioned above, having seperate configs for each LVM types


1.  Physical Volumes

Physical Volumes will be talos managed since CSI will only manage LV.
We can have both Declarative Create and Imperative Create.

A Physical Volume Config should always have name, that name must be unique.
A `diskSelector` can be used to find the device or know which block pv needs to be created from and talos should be responsible for creating this PV.

```
apiVersion: v1alpha1
kind: LVMPhysicalVolumeConfig
name: pv-sd0
diskSelector:
        match: 'disk.dev_path == /dev/sda0'
```
We can have a LVMDiscovery controller which checks on LVM resources .

Once the PV is created it should have an `inactive` status (Record/track of which LVMDiscovery should know),
If in future the PV is used by a volume group this status should go to `active`, and when it has extents allocated state can go to `in-use`

-  After adding this config, if the PV is not present  -> Create PV.
- For Deletion:
   - If the PV is present and state is active/in-use, we should block anyone trying to remove the cofig.
   - If the PV is present but the status is inactive, we can allow removal of config (but the PV is still not deleted) , (If user tried to add same config again, LVMDiscovery should note PV exists already with a `inactive` status and does nothing.

(I'm having second thoughts about this, maybe we should not allow direct removal of config even if it is not in use, maybe the config removal should only be done via API)
As a user if a config is removed and pv still exists , and I do a  `get config` those configs would be missing, LVMDiscovery might record LVM resources but might get confusing, So I'm more inclined having configs removed only via `talosctl lvm delete pv` that too only `inactive` ones. 


Complete cleanup/ Deletion of PV should be allowed only when it is `inactive` and through API
maybe like `talosctl lvm delete pv-sd0`

For API it should follow what config does, 
We can have 3 funtionalities:

`
1. talosctl lvm create pv <pv-name> --disk <disk-name>
2. talosctl lvm delete pv <pv-name>
3. talosctl get pv 
`

I don't think it will be a good idea to edit any existing pv, maybe through change in config it will possible but not via API



Let me know how you feel about above config and API, its better to get one component of LVM straight before moving on to next.

shanduur: I definitely think that if we go the route to have PV as a separate entity, it should be immutable. But, do we need to have them as a separate entity? Why don't we just rely on the `LVMVolumeGroupConfig` to have a selector, based on which disks belonging to the specific VG are selected and provisioned into PV? Users will have harder time breaking things this way.

pranav767: This could work as well, but what happens when there is no PV with the selector, do we automatically create those PVs? (Don't think that would be the correct answer)
It should go in pending untill such pv appears
But the question being how will we manage PVs

shanduur: > there is no PV with the selector

Selector is only for the disks / devices. Based on that selector, we provision PVs and add them to VG.

pranav767: Alright, I think I get it now, those are disks and will be applicable when talos sees those disks, when seen talos will create those pvs and will be assigned to vg
I'll continue with the idea for VG config next (proper design and co fig and flow)

pranav767:
@shanduur , @smira 
From above chats, we can say `LVMVolumeGroupConfig` can have a selector based on which PV provisioning could be done.

For Volume Groups config can be:
```
apiVersion: v1alpha1
kind: LVMVolumeGroupConfig
name: vg-pool
physicalVolumes:
  volumeSelector:
    match: 'disk.dev_path == /dev/sda0'
managed-by: talos | csi
```

- VG names must be unique across the node.
- At least one disk selector is required for a VG.
- managed-by is a required field with allowed values: (this `managed-by` param could be written in a better way maybe `lv-managed-by` which will be clear talos always has control over vg, and lv could be controlled by csi in some cases)
  - talos
  - csi
  
Although Talos always reconciles and manages the VG object itself, the managed-by field defines who is allowed to manage Logical Volumes (LVs) inside the VG. This helps determine the expected workflow and what configuration comes next.

### VG and PV provisioning flow

1. LVMDiscovery discovers existing LVM resources on the node.
    -   If a VG with the given `name` and `managed-by` value does not exist, Talos creates the VG based on the provided configuration.

2. PV provisioning

  - PVs are provisioned only when the disk selector matches an existing disk.
  - When a matching disk is found, Talos automatically creates a PV using the entire disk and adds it to the VG.
  - If the selector does not match any disk, no PV is created and the VG remains in a pending state.
  - When a matching disk appears later, Talos automatically provisions the PV and attaches it to the VG.
Disk discovery might be already exisiting in talos :thinking: , we could extend this to LVMDiscovery.

### LV ownership and behavior

The `managed-by` field strictly defines who is allowed to create and manage LVs for the VG.

A VG must have exactly one LV owner.

`managed-by: talos`

- Talos is responsible for provisioning and reconciling LVs. 
- LV configuration is required.
- CSI is not allowed to create or modify LVs in this VG.

`managed-by: csi`

- Talos manages the VG and PVs only.
- LV configuration is not required.
- CSI is the sole entity allowed to create, resize, or delete LVs in this VG.
- Talos must not perform LV-level changes for this VG.


Deletion

- VG configuration cannot be removed directly.
- Deletion must be performed explicitly via the Talos API:
`talosctl lvm delete vg vg-pool`

Deletion flow:

- LVMDiscovery checks for active LVs belonging to the VG.
- If active LVs exist, deletion is denied.
- If no active LVs exist:
  - PVs are detached and removed
  - The VG is deleted
  
  
Let me know what you think of above design?

smira:
I feel this is a bit more complicated than what I would like it to be.

First of all, CSIs (as far as I know) today happily manage LVM volumes without any Talos involvement, so I'm not sure why we need to introduce any concept of it to Talos.

Second, I think the configuration should be something that speaks an easier language than the raw LVM language, so it serves the purpose better rather than trying to support all cases. 

So I would start with a usecase for LVM.

Let's imagine I have 4 disks in my machine. One disk will be Talos system disk, let's take it out of the picture. 

I want to combine remaining 3 disks into a logical volume that spans the whole 3 disks. Usecase: I want to give workload access to a volume that can utilize all three disks as a whole. 

What is the machine config I would like to have? How does it get reconciled if a new disk is added? 

I would imagine something that I just match all disks by saying e.g. `!system_disk && disk.transport == 'nvme'` and Talos automatically manages that, creates a named LVM volume, and later on e.g. I use `UserVolumeConfig` with a filesystem `xfs` to get it formatted/mounted. 

But I would like to have a higher level, probably opinionated design which exposes the features in a thought through and constrained way. 

Another case might be RAID 0 (which certainly has different restrictions) and RAID 1. I don't know if there's any real case for slicing a disk, or this should be done some other way.

But I would like a scenario and a solution, not a way to map all LVM features into Talos.

shanduur:
> ### LV ownership and behavior
> 
> The `managed-by` field strictly defines who is allowed to create and manage LVs for the VG.
> 
> A VG must have exactly one LV owner.
> 
> `managed-by: talos`
> 
> - Talos is responsible for provisioning and reconciling LVs. 
> - LV configuration is required.
> - CSI is not allowed to create or modify LVs in this VG.
> 
> `managed-by: csi`
> 
> - Talos manages the VG and PVs only.
> - LV configuration is not required.
> - CSI is the sole entity allowed to create, resize, or delete LVs in this VG.
> - Talos must not perform LV-level changes for this VG.

The ownership concept won't work because we have no real way to enforce / lock VG from use outside of Talos. We would need to mask the devices, not propagate them to kubelet, or do some other not-fun hacks, or e.g. requiring enforcing mode with SELinux. Otherwise it's just a metadata that has no impact and does not block any actions - it is meaningless.

> Deletion
> 
> - VG configuration cannot be removed directly.
> - Deletion must be performed explicitly via the Talos API:
>   `talosctl lvm delete vg vg-pool`

I strongly agree with the deletion mechanism. However we should avoid a standalone `talosctl lvm delete` if possible. Instead, we should extend the `wipe` semantics. If a VG is removed from the config, Talos should refuse to delete it unless a `wipe` API is called, preventing data loss during accidental config updates.


> 2. PV provisioning
> 
>   - PVs are provisioned only when the disk selector matches an existing disk.
>   - When a matching disk is found, Talos automatically creates a PV using the entire disk and adds it to the VG.
>   - If the selector does not match any disk, no PV is created and the VG remains in a pending state.
>   - When a matching disk appears later, Talos automatically provisions the PV and attaches it to the VG.

I agree here. But as @smira said, we need to build user story here, to ensure this is a good enough design, that can be extended in the future when new requirements are brought.

1. As a user running stateful workloads on Talos, I want to define Logical Volume Management (LVM) structures directly in the machine configuration, so that I can create flexible, high-performance local storage pools for services like databases and provide a stable base for the TopoLVM drivers to manage dynamic persistent volumes.
  **The Workflow:** Talos manages the Physical Volumes (PV) and Volume Group (VG). For "internal" needs (UserVolumes), Talos creates a LV. For "dynamic" needs (K8s pods), Talos hands the VG over to a CSI.

> Disk discovery might be already exisiting in talos :thinking: , we could extend this to LVMDiscovery.

This will bloat already complicated controller, let's not do that. We should treat LVM as a 'Layer 2' storage entity. The block devices controller identifies raw block devices; the LVM Controller consumes those.

ifoughal:
> A user can have a LVM config like:
> 
> ```
> physicalVolumes:
>   - /dev/sda
>   - /dev/sdb
>   - /dev/sdc
> 
> volumeGroups:
>   - name: vg-system
>     physicalVolumes:
>       - /dev/sda
>   - name: vg-apps
>     physicalVolumes:
>       - /dev/sdb
>       - /dev/sdc
> 

@pranav767 users must also be able to list the disks, actually this should be a must, since the labels (/dev/sdx) are not persistent. users should list by disk WWN
```
lsblk -no WWN /dev/sda
```

pranav767:
@ifoughal and others following this issue, this likely needs more planning and discussion, I've created a new discussion here https://github.com/siderolabs/talos/discussions/12578 , please let me know your views/opinions.

smira:
Thank you, LVM support was implemented in other set of PRs.

PR #12532, opened 2026-01-02
chore: update slack links

rsmitty:
This PR updates the slack links to reference the auto-inviter at inviter.co.


frezbo:
I think https://github.com/siderolabs/docs/ also needs updating

Issue #12511, opened 2025-12-29
Missing installer image for 1.12


mtb-xt:
## Bug Report
There is no `installer` image published in the release for talos 1.12.0

### Description
The [documentation](https://docs.siderolabs.com/talos/v1.12/configure-your-talos-cluster/lifecycle-management/upgrading-talos) states - 

> Each Talos version has a corresponding installer image, listed on the release page for the version, for example [v1.12.0](https://github.com/siderolabs/talos/releases/tag/v1.12.0).

If we open the release page, there is a list of images (at the bottom), but there is no `installer` image among them. If we check release [1.11.0](https://github.com/siderolabs/talos/releases/tag/v1.11.0), for example, it does have an `installer` image.

Yes, it's a nitpick. It's still annoying.

smira:
Installer image is published:

```
$ crane manifest ghcr.io/siderolabs/installer:v1.12.0 
{"schemaVersion":2,"mediaType":"application/vnd.oci.image.index.v1+json","manifests":[{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","size":593,"digest":"sha256:67167732d5c0f7bc437d0dfe78836057845550e2966cfe97bbcd3a719889f5dd","platform":{"architecture":"amd64","os":"linux"}},{"mediaType":"application/vnd.docker.distribution.manifest.v2+json","size":593,"digest":"sha256:14c62315cd633e89a3a6248fe9a4b0038eccb6930dbc6ea325bf9e649067f412","platform":{"architecture":"arm64","os":"linux"}}]}
```

You shouldn't be using it though most of the time (see [Image Factory](https://factory.talos.dev/)). 

But it's missing in the release notes, and that's a problem.

mtb-xt:
> Installer image is published:

Yup, as I said, it's just a nitpick :)


> You shouldn't be using it though most of the time (see [Image Factory](https://factory.talos.dev/)).

If you don't mind explaining - are we supposed to use the image factory even for AWS images? I thought if it's like a standard installation, then we should avoid adding additional load to the image factory.

smira:
> > You shouldn't be using it though most of the time (see [Image Factory](https://factory.talos.dev/)).
> 
> If you don't mind explaining - are we supposed to use the image factory even for AWS images? I thought if it's like a standard installation, then we should avoid adding additional load to the image factory.

[...]

PR #12539, opened 2026-01-05
fix: print talosctl images to release notes

shanduur:
After changing `talsoctl images k8s-bundle`/`talos-bundle` we stopped printing some of the images to release notes.
This fixes that issue.

Closes https://github.com/siderolabs/talos/issues/12511

### DIFF
--- a/cmd/talosctl/cmd/talos/image.go
+++ b/cmd/talosctl/cmd/talos/image.go
@@ -173,20 +173,13 @@ var imageDefaultCmd = &cobra.Command{
 	},
 }
 
-const (
-	provisionerDocker    = "docker"
-	provisionerInstaller = "installer"
-	provisionerAll       = "all"
-)
+var imageTalosBundleCmdFlags = struct {
+	extensions bool
+	overlays   bool
+}{}
 
-var imageDefaultCmdFlags = struct {
-	provisioner pflag.Value
-}{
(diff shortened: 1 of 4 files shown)


cmd/talosctl/cmd/talos/image.go
```diff
@@ -266,12 +249,26 @@ var imageSourceBundleCmd = &cobra.Command{
...
+
+			for _, extension := range extensions {
+				digestedReferences = append(digestedReferences, fmt.Sprintf("%s@%s", extension.TaggedReference.String(), extension.Digest))
+			}
 		}
 
-		for _, extension := range extensions {
-			digestedReferences = append(digestedReferences, fmt.Sprintf("%s@%s", extension.TaggedReference.String(), extension.Digest))
+		if imageTalosBundleCmdFlags.overlays {
+			overlays, err = artifacts.FetchOfficialOverlays(tag)
```
frezbo, line 264: I'm not sure about hitting ghcr.io, what about air-gapped users, or we should support setting the registry to hit 🤔

shanduur: This is intended to be only used in non-air-gapped envs, to generate required digests that are needed to be moved to the air-gapped env.

frezbo (approved):

PR #12548, opened 2026-01-07
feat: add EnvironmentConfig document

### LINKED ISSUE #12439: Move `.machine.env` into `EnvironmentConfig`
Allow to configure environment variables via multi-doc.

This might be handy to configure e.g. `HTTP_PROXY` in the maintenance mode.

It should be documented that environment changes don't propagate to already running services/containers.

Deprecate `.machine.env`, but we can merge environment from both sources.

```yaml
apiVersion: v1alpha1
kind: EnvironmentConfig
variables:
  HTTP_PROXY: http://10.0.0.1/
```

shanduur:
Add new EnvironmentConfig document for configuring the Env vars.
Deprecate .Machine.Env

```console
$ talosctl -n 172.20.0.2 get env
NODE         NAMESPACE   TYPE          ID         VERSION   VARIABLES
172.20.0.2   runtime     Environment   machined   1         ["GRPC_ENFORCE_ALPN_ENABLED=false","PATH=/usr/bin:/usr/local/sbin:/usr/local/bin:/opt/cni/bin","TCELL_MINIMIZE=1"]
```

```console
$ talosctl  -n 172.20.0.2 get env -o yaml
node: 172.20.0.2
metadata:
    namespace: runtime
    type: Environments.runtime.talos.dev
    id: machined
    version: 1
    owner: runtime.EnvironmentController
    phase: running
    created: 2026-01-12T13:57:14Z
    updated: 2026-01-12T13:57:14Z
spec:
    variables:
        - GRPC_ENFORCE_ALPN_ENABLED=false
        - PATH=/usr/bin:/usr/local/sbin:/usr/local/bin:/opt/cni/bin
        - TCELL_MINIMIZE=1
```

Closes #12439


laurazard:
Is it worth calling out (in the related docs) about multiple `EnvironmentConfig` docs overwriting previous environment vars with the same name? (Plus, it'll be possible to overwrite previous environment vars, but not unset them, I guess)

shanduur:
@laurazard I added the note to the changelog and some more explicit information to the docs.

pkg/machinery/config/types/runtime/environment.go
```diff
@@ -0,0 +1,139 @@
...
+
+// Validate implements config.Validator interface.
+func (s *EnvironmentV1Alpha1) Validate(validation.RuntimeMode, ...validation.Option) ([]string, error) {
+	var err error
+
+	for key := range s.EnvironmentVariables {
+		if !POSIX1EnvKeyRegex.MatchString(key) {
+			err = errors.Join(err, fmt.Errorf("invalid environment variable name: %q", key))
+		}
+	}
```
laurazard, line 141: This validation is new right? If not, disregard everything following this 😅 

I don't think this is a huge problem, I remember at previous places implementing things like this (with a rationale like "surely, there's a POSIX standard, no one is doing things that are too unhinged") and then getting a bunch of complaints about breaking things, iirc. in particular because of weird Java services with crazy environment variables.

Actually, reading the regex, I can tell it's going to fail for variables containing dots (`.`), which people [unfortunately do use](https://github.com/docker-library/tomcat/issues/302).

shanduur: I'm impressed that it's not enforced at the host/shell level... 

I think it shouldn't be issue for us, as it affects only Talos (PID1) and core services, over which we have (almost) full control. `GRPC_*`, `HTTP*` and other env vars we are using are POSIX.1 compliant.

Extension, which might use non-compliant environment variables, are configured with `ExtensionServiceConfig` - we don't have env validation here.

laurazard: That's fair! I was looking through it and figuring out whether it was, I guess the description in https://github.com/siderolabs/talos/issues/12439 threw me off a bit. Thanks for the clarification!

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -186,8 +187,9 @@ func (e *Etcd) Runner(r runtime.Runtime) (runner.Runner, error) {
 
 	env := environment.Get(r.Config())
 
-	if goruntime.GOARCH == "arm64" {
-		env = append(env, "ETCD_UNSUPPORTED_ARCH=arm64")
+	// NOTE: leave it here for future unsupported architectures, so we can know where to add them
+	if slices.Contains([]string{}, goruntime.GOARCH) {
+		env = append(env, "ETCD_UNSUPPORTED_ARCH="+goruntime.GOARCH)
 	}
```
shanduur, line 193: We no longer need this value, `aarch64` is level 1 support.

pkg/machinery/config/types/runtime/environment.go
```diff
@@ -0,0 +1,141 @@
...
+type EnvironmentV1Alpha1 struct {
+	meta.Meta `yaml:",inline"`
+
+	//   description: |
+	//     This field allows for the addition of environment variables.
+	//     All environment variables are set on PID 1 in addition to every service.
+	//     Propagation of environment variables to services is done only at initial service start time.
+	//     To modify environment variables for services, the node must be restarted.
+	//     Multiple values for the same environment variable (in multiple documents) will replace previous values, with the last one taking precedence.
+	//     Fully removing an environment variable can only be achieved by removing it from the document and restarting the machine.
```
laurazard, line 56: It might still be worth to document the validation:

```suggestion
	//     This field allows for the addition of environment variables.
	//     All environment variables are set on PID 1 in addition to every service.
	//     Propagation of environment variables to services is done only at initial service start time.
	//     To modify environment variables for services, the node must be restarted.
	//     Multiple values for the same environment variable (in multiple documents) will replace previous values, with the last one taking precedence.
	//     Fully removing an environment variable can only be achieved by removing it from the document and restarting the machine.
	//     Environment variable names are validated, and should:
	//       - start with an uppercase letter, lowercase letter, or an undescore (_) character, and
	//       - contain only uppercase and lowercase letters, underscore (_) characters, and numbers.
```

Issue #11440, opened 2025-07-28
TrustD certificates don't contain expected SANs


james-callahan:
### Description

Trustd generates and serves a certificate with different SANs (Subject Alternative Names) than I expected.

I expect that trustd certificates would contain the endpoint that was connected to, i.e. the value of `cluster.controlPlane.endpoint`. Instead, the certificates contain SANs of the individual machine's hostnames and ips.
Those ips and hostnames may not be reachable; and for our case, our `cluster.controlPlane.endpoint` is an ELB

Additionally, I consider those IPs/hostnames to be an information leak: I don't expect given a public trustd endpoint that people would be able to figure out our internal ips or internally assigned host ids.

----

In particular this causes us an issue when trying to start new workers, they fail to generate a CSR with `tls: failed to verify certificate`.
```
[  769.756226] [talos] controller failed {"component": "controller-runtime", "controller": "secrets.APIController", "error": "failed to sign API server CSR: 6 error(s) occurred:\n\trpc error: code = DeadlineExceeded desc = context deadline exceeded while waiting for connections to become ready\n\trpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for 10.208.144.4, 52.91.19.58, 2600:1f10:4a76:2210::4, fdd1:5910:5684:c9ed:1112:dd49:ec70:1900, not 54.85.242.149\"\n\trpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate is valid for 3.84.129.115, 10.208.128.4, 2600:1f10:4a76:2200::4, fdd1:5910:5684:706a:3e66:f992:3b31:ab40, not 2600:1f10:4a76:2200:f9e5:2179:26d:2470\"\n\trpc error: code = Unavailable desc = connection error: desc = \"transport: authentica....
```
I note that the controlplane address as given in the machineconfig was [talos.dev](http://talos.dev/).turnkey.engineering.

### Logs

```
$ openssl s_client -connect 3.92.155.143:50001 | openssl x509 -noout -text
Connecting to 3.92.155.143
Can't use SSL_get_servername
depth=0 CN=i-01a4709d29d3b4bdc.ec2.internal
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 CN=i-01a4709d29d3b4bdc.ec2.internal
verify error:num=21:unable to verify the first certificate
verify return:1
[...]

PR #12549, opened 2026-01-07
feat: add authority in GRPC client initialization

smasset-orange:
# Pull Request

## What? (description)
This PR adds an authority option in the initialization of remote Trustd clients

## Why? (reasoning)
I'm trying to add Talos workers to Hosted Control Planes (generated with [Kamaji](https://github.com/clastix/kamaji) with extra [talos-csr-signer](https://github.com/clastix/talos-csr-signer) component deployed).

Their [demo](https://youtu.be/nSGo_72LnmY) uses a different LoadBalancer service on the admin cluster for each Tenant so the current implementation of the trustd GRPC client works well.

In my case I'm trying to use an external LB with a single IP address for all Tenants. As pointed out in [#11440](https://github.com/siderolabs/talos/issues/11440#issuecomment-3131709894) Talos "doesn't support trustd over any kind of LB today". This tries to fix that.

## Acceptance
I've built a custom image with the proposed changes and a hard-coded hostname in [internal/app/machined/pkg/controllers/secrets/api.go#L272](https://github.com/smasset-orange/talos/blob/fix/grpc-authority/internal/app/machined/pkg/controllers/secrets/api.go#L272) instead of the current empty string.

With this, the request is correctly forwarded to the Tenant's CSR signer and the worker joins the cluster.

I'll need guidance to implement the second part of the change : getting the actual hostname configured for the cluster endpoint (as I can't use the existing `address`  parameter which already has replaced it with IP addresses the hostname resolves to).

### DIFF
--- a/internal/app/machined/pkg/controllers/secrets/api.go
+++ b/internal/app/machined/pkg/controllers/secrets/api.go
@@ -269,7 +269,7 @@ func (ctrl *APIController) reconcile(ctx context.Context, r controller.Runtime, 
 				return err
 			}
 		} else {
-			if err := ctrl.generateWorker(ctx, r, logger, rootSpec, endpointsStr, certSANs); err != nil {
+			if err := ctrl.generateWorker(ctx, r, logger, rootSpec, endpointsStr, "", certSANs); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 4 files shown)


shanduur:
I'll merge it as is (given the tests will pass), but the second part is quite tricky - this _might_ technically be done with some additional input to the `secrets.APIController`?

shanduur:
I reverted this change - it shouldn't have been merged to `release-1.10`, but to `main`.

Issue #12553, opened 2026-01-08
talosctl health --wait-timeout does not work


olavst-spk:
### Description
The `--wait-timeout` flag has no effect when set to a value higher than 5 minutes. 

This issue is related to https://github.com/siderolabs/talos/issues/11601 and https://github.com/siderolabs/terraform-provider-talos/issues/241.

#### Steps to reproduce

1. Create a test cluster:
    `talosctl cluster create docker`
2. Simulate slow cluster startup by scaling down coredns:
    `kubectl -n kube-system scale deploy coredns --replicas=0`
3. Run talosctl health with the --wait-timeout flag set to a high value (for example 30 minutes):
    `talosctl health -n 10.5.0.2 --wait-timeout 30m`
4. In a separate terminal, wait for 5 minutes and scale up coredns again:
    `sleep 5m && kubectl -n kube-system scale deploy coredns --replicas=1`

#### Expected result
The healthcheck should succeed since coredns is back up and running within the time that `--wait-timeout` is set to (in this case 30 minutes)

#### Actual result
The healtcheck fails with a timeout after 5 minutes, regardless of what `--wait-timeout` is set to.

### Environment

- Talos version: 
  ```
  Client:
  	Tag:         v1.12.1
  	SHA:         7ea2ef7c
  	Built:       
  	Go version:  go1.25.5
  	OS/Arch:     linux/amd64
  Server:
  	NODE:        10.5.0.2
  	Tag:         v1.12.1
  	SHA:         7ea2ef7c
  	Built:       
  	Go version:  go1.25.5
  	OS/Arch:     linux/amd64
  	Enabled:     RBAC
  ```
- Kubernetes version:
  ```
  Client Version: v1.35.0
  Kustomize Version: v5.7.1
  Server Version: v1.35.0
  ```
- Platform: any

olavst-spk:
I'm working on a PR to fix this by adding a timeout value to the healthcheck functions as suggested in https://github.com/siderolabs/talos/issues/11601

smira:
To clarify: this setting works, but probably it's confusing, as it sets global timeout for all checks, not a per-check timeout.

Issue #11601, opened 2025-08-15
DefaultClusterChecks should be able to accept a variable Timeout value


Joe-Bresee:
### Description
Some talos users may require longer timeout sessions or don't need to wait 10/5 minutes for checks. There is an issue on the [terraform-provider](https://github.com/siderolabs/terraform-provider-talos) issue board [here](https://github.com/siderolabs/terraform-provider-talos/issues/241) asking to be able to set the timeout interval, but for that to work, DefaultClusterChecks needs to be able to accept timeout variables on the talos/ side.

Joe-Bresee:
I'd like this issue to be assigned to myself please

olavst-spk:
@Joe-Bresee Did you get any traction on this? I'm willing to help out with a PR if needed

Joe-Bresee:
@olavst-spk Unfortunately, I haven't received much communication from maintainers regarding this issue.

PR #12554, opened 2026-01-08
fix: fix healthcheck timeout

olavst-spk:
# Pull Request

## What? (description)

Removes the hard-coded 5 minute timeout in DefaultClusterChecks, K8sComponentsReadinessChecks and PreBootSequenceChecks.

## Why? (reasoning)

Currently, each healthcheck has a hard-coded timeout of 5 minutes. Because of this, the `--wait-timeout` parameter has no effect if set to a value larger than 5 minutes (see https://github.com/siderolabs/talos/issues/12553 for more details).

Fixes https://github.com/siderolabs/talos/issues/12553, Fixes https://github.com/siderolabs/talos/issues/11601

## Acceptance

### DIFF
--- a/pkg/cluster/check/default.go
+++ b/pkg/cluster/check/default.go
@@ -23,7 +23,7 @@ func DefaultClusterChecks() []ClusterCheck {
 			func(cluster ClusterInfo) conditions.Condition {
 				return conditions.PollingCondition("all k8s nodes to report ready", func(ctx context.Context) error {
 					return K8sAllNodesReadyAssertion(ctx, cluster)
-				}, 10*time.Minute, 5*time.Second)
+				}, 5*time.Second)
 			},
 
 			// wait for kube-proxy to report ready
@@ -39,7 +39,7 @@ func DefaultClusterChecks() []ClusterCheck {
 					}
 
 					return K8sPodReadyAssertion(ctx, cluster, replicas, "kube-system", "k8s-app=kube-proxy")
-				}, 5*time.Minute, 5*time.Second)
+				}, 5*time.Second)
 			},
 
 			// wait for coredns to report ready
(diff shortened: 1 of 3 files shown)


smira (commented):
I understand the problem, but I think the change here is a bit wrong. 

The `ClusterWaitTimeout` is the global timeout for all checks combined, and we have also per-check timeout. 

So the setting today works, but it is applied on the global level.

I don't know if a proper fix is simply to drop per-check timeout and rely on a global timeout.

shanduur (changes requested):
Hmm, right, so the better way is to add extra flags, that will allow changing individual timeouts, and clarify the `help` on the `--wait-timeout`, so that it's clear that it is a global timeout for all checks combined.

laurazard:
Yeah, there's two separate things IMO which are:
- clarify that `--wait-timeout` is used to **set a global timeout** (i.e. this is as long as I'm willing to wait before it's done)
- add some new flag(s) to configure the per-timeout checks/remove the hardcoded values (or drop the per-check timeouts as per what @smira said)

olavst-spk:
From a user's perspective I think having multiple timeout flags could be confusing. I think most users view the healtcheck as a single operation, even if it is implemented as many separate checks.

olavst-spk:
How about removing the per-check timeouts (by making them infinite) and relying only on the global timeout? Then the current `--wait-timeout` flag should work as expected. 

I can update the PR if you think this is an okay solution.

smira:
> How about removing the per-check timeouts (by making them infinite) and relying only on the global timeout? Then the current `--wait-timeout` flag should work as expected.
> 
> I can update the PR if you think this is an okay solution.

I think that is the proper way. Having shorter timeout for each step makes sense only to fail faster, but really doesn't change anything.

laurazard:
I think dropping per-check timeouts and just doing a global timeout is fine, and an improvement over the current state! Feel free to update the PR to do that @olavst-spk :')

olavst-spk:
I have updated the PR and description now :) 

I decided to remove the timeout parameter from `PollingCondition` since it was only used by these checks.

smira (approved):
LGTM, thank you!

smira:
I don't think it should be backported to 1.12, as it might be a breaking change in some cases, so I would rather keep it in 1.13 to get more exposure.

PR #12563, opened 2026-01-09
feat: use new xfs config file

frezbo:
Use new xfs config file version.


pkg/machinery/constants/constants.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	// DefaultKernelVersion is the default Linux kernel version.
-	DefaultKernelVersion = "6.18.2-talos"
```
smira, line 17: hack/release.toml update

smira (approved):

PR #12582, opened 2026-01-12
fix: drop the persist config flag from gen config

smira:
The `persist` value was locked to be true for a long time now, and Talos doesn't support any other mode (machine config is persisted).

Drop the `gen config` flag and related generate options, as modern Talos doesn't accept `persist: false`.


frezbo (approved):
:cool:

Issue #12308, opened 2025-12-01
opaque mount options for volumes


smira:
Support for adding extra mount options for volumes.

PR #12585, opened 2026-01-13
feat: add opaque mount options support for existing volumes

pranav767:
This PR closes https://github.com/siderolabs/talos/issues/12308

## What? (description)

- Add support for adding extra mount options for volumes.

## Why? (reasoning)

- Allowing users more options for mounting volumes , ref : https://man7.org/linux/man-pages/man8/mount.8.html

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/user_volumes.go
+++ b/internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/user_volumes.go
@@ -46,11 +46,12 @@ func UserVolumeTransformer(c configconfig.Config) ([]VolumeResource, error) {
 		userVolumeResource := VolumeResource{
 			VolumeID:           volumeID,
 			Label:              block.UserVolumeLabel,
-			MountTransformFunc: DefaultMountTransform,
+			MountTransformFunc: HandleUserVolumeMountRequest(userVolumeConfig), // This is overridden for Directory type below.
 		}
 
 		switch userVolumeConfig.Type().ValueOr(block.VolumeTypePartition) {
 		case block.VolumeTypeDirectory:
+			userVolumeResource.MountTransformFunc = DefaultMountTransform
 			userVolumeResource.TransformFunc = NewBuilder().
 				WithType(block.VolumeTypeDirectory).
 				WithMount(block.MountSpec{
@@ -191,6 +192,7 @@ func ExistingVolumeTransformer(c configconfig.Config) ([]VolumeResource, error) 
 
 	for _, existingVolumeConfig := range c.ExistingVolumeConfigs() {
 		volumeID := constants.ExistingVolumePrefix + existingVolumeConfig.Name()
+
 		resources = append(resources, VolumeResource{
 			VolumeID: volumeID,
 			Label:    block.ExistingVolumeLabel,
@@ -336,11 +338,23 @@ func SwapVolumeTransformer(c configconfig.Config) ([]VolumeResource, error) {
 	return resources, nil
 }
 
+// HandleUserVolumeMountRequest returns a MountTransformFunc for user volumes.
+func HandleUserVolumeMountRequest(userVolumeConfig configconfig.UserVolumeConfig) func(m *block.VolumeMountRequest) error {
+	return func(m *block.VolumeMountRequest) error {
+		m.TypedSpec().DisableAccessTime = userVolumeConfig.Mount().DisableAccessTime()
+		m.TypedSpec().Secure = userVolumeConfig.Mount().Secure()
+
+		return nil
+	}
+}
+
 // HandleExistingVolumeMountRequest returns a MountTransformFunc for existing volumes.
 // It sets `VolumeMountRequestSpec.ReadOnly` based on the existing configuration.
 func HandleExistingVolumeMountRequest(existingVolumeConfig configconfig.ExistingVolumeConfig) func(m *block.VolumeMountRequest) error {
 	return func(m *block.VolumeMountRequest) error {
 		m.TypedSpec().ReadOnly = existingVolumeConfig.Mount().ReadOnly()
+		m.TypedSpec().DisableAccessTime = existingVolumeConfig.Mount().DisableAccessTime()
+		m.TypedSpec().Secure = existingVolumeConfig.Mount().Secure()
 
 		return nil
 	}
@@ -350,6 +364,8 @@ func HandleExistingVolumeMountRequest(existingVolumeConfig configconfig.Existing
 func HandleExternalVolumeMountRequest(externalVolumeConfig configconfig.ExternalVolumeConfig) func(m *block.VolumeMountRequest) error {
 	return func(m *block.VolumeMountRequest) error {
 		m.TypedSpec().ReadOnly = externalVolumeConfig.Mount().ReadOnly()
+		m.TypedSpec().DisableAccessTime = externalVolumeConfig.Mount().DisableAccessTime()
+		m.TypedSpec().Secure = externalVolumeConfig.Mount().Secure()
 
 		return nil
 	}
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -34,9 +34,11 @@ import (
 )
 
 type mountContext struct {
-	point     *mount.Point
-	readOnly  bool
-	unmounter func() error
+	point             *mount.Point
+	readOnly          bool
+	disableAccessTime bool
+	secure            bool
+	unmounter         func() error
 }
 
 // MountController performs actual mount/unmount operations based on the MountRequests.
@@ -563,7 +565,7 @@ func (ctrl *MountController) updateTargetSettings(
 	return nil
 }
 
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (ctrl *MountController) handleDiskMountOperation(
 	logger *zap.Logger,
 	mountSource, mountTarget string,
@@ -621,6 +623,14 @@ func (ctrl *MountController) handleDiskMountOperation(
 			mount.WithSelinuxLabel(volumeStatus.TypedSpec().MountSpec.SelinuxLabel),
 		)
 
+		if mountRequest.TypedSpec().DisableAccessTime {
+			opts = append(opts, mount.WithDisableAccessTime())
+		}
+
+		if mountRequest.TypedSpec().Secure {
+			opts = append(opts, mount.WithSecure())
+		}
+
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
 		}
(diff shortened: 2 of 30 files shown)


pkg/machinery/config/types/block/existing_volume_config.go
```diff
@@ -102,11 +102,27 @@ func exampleVolumeSelector2() cel.Expression {
...
+}
+
 // MountSpec describes how the volume is mounted.
 type MountSpec struct {
 	//   description: |
 	//     Mount the volume read-only.
 	MountReadOnly *bool `yaml:"readOnly,omitempty"`
+	//   description: |
+	//     Additional mount options to pass to the mount syscall.
+	//
```
pranav767, line 115: I’ve kept this mount option flexible to allow both boolean flags and key–value pairs (where key–value pairs are expressed as key=value), representing key–value pairs as plain strings feels a bit off—please let me know if there’s a better way to model this.
This will look like
```
mount:
  readOnly: false
  options:
    - noatime        # boolean flag
    - nodiratime     # boolean flag
    - nosuid         # boolean flag
    - nodev          # boolean flag
    - uid=1000       # key=value
    - gid=1000       # key=value
    - size=100m      # filesystem-specific key=value (e.g., tmpfs)
```

website/content/v1.13/reference/configuration/block/existingvolumeconfig.md
```diff
@@ -84,6 +84,13 @@ MountSpec describes how the volume is mounted.
 | Field | Type | Description | Value(s) |
 |-------|------|-------------|----------|
 |`readOnly` |bool |Mount the volume read-only.  | |
```
pranav767, line 86: The docs page currently does not show mount options in the config [here](https://docs.siderolabs.com/talos/v1.12/reference/configuration/block/existingvolumeconfig). I’m guessing this is because it’s an optional parameter, so I haven’t included those extra options here.

pkg/machinery/config/types/block/existing_volume_config.go
smira: We had an internal discussion about this, and the consensus was that we are not happy with a free-form field.

We were considering the following (probably separate) fields, we don't have good names for them, but:

- `DisableAccessTime` (defaults to false) -> triggers proper `no*atime` based on filesystem type
- `Secure` (defaults to true, breaking change) -> triggers `nosuid`, `nodev` (if the filesystems supports it), and it should be explicitly set to `false` (breaking change for 1.13, but in the direction of better security)
- `uid`/`gid` might be interesting, but if we do that, we need to do this as separate fields
- `size` doesn't apply (yet), as there is no support for `tmpfs`

pranav767: I've removed mountOptions and used separate fields : `DisableAccessTime` and `Secure`.

- For `uid`/`gid` we should only apply them for filesystems that support those mount options 
(examples:  `vfat, fat, ntfs, cifs, nfs, tmpfs, virtiofs` ) some of these might not be supported yet by talos but no harm while using validation. We cannot validate this at config time because the filesystem is discovered at runtime, the mount controller can check the filesystem type for this, I can add that logic if this is alright.

Also, there are several other mount options available, let me know if you want additional typed fields.

website/content/v1.13/reference/configuration/block/existingvolumeconfig.md
```diff
@@ -84,6 +84,13 @@ MountSpec describes how the volume is mounted.
...
+|`options` |[]string |Additional mount options to pass to the mount syscall.<br><br>Supports both boolean flags (e.g., "noatime", "ro") and key-value pairs (e.g., "uid=1000"). <details><summary>Show example(s)</summary>mount options:{{< highlight yaml >}}
+options:
+    - noatime
+    - nodiratime
+    - uid=1000
+    - gid=1000
+{{< /highlight >}}</details> | |
```
shanduur: Looks like you missed `make generate docs`

shanduur: Fixed

shanduur:
I think we need E2E test for those options as well before we can merge it.

Edit:
In this file: https://github.com/siderolabs/talos/blob/v1.12.1/internal/integration/api/volumes.go

pranav767:
@shanduur , Added E2E test for both `disableAccessTime` and `secure` mount options.
I can move the PR to draft if further discussion is needed around including additional mount options.

shanduur:
The tests are failing with:

```log
2026/01/16 09:38:02 172.20.1.6: {"component":"controller-runtime","controller":"block.MountController","error":"failed to mount \"e-694b99c\": openfs failed: failed to create root filesystem: FSCONFIG_SET_FLAG failed: invalid argument: key=\"nosuid\"","msg":"2026/01/16 09:38:02.099973 [talos] controller failed","talos-level":"info","talos-service":"machined","talos-time":"2026-01-16T09:38:02.099993382Z"}
```

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/user_volumes.go
```diff
@@ -190,6 +190,20 @@ func ExistingVolumeTransformer(c configconfig.Config) ([]VolumeResource, error)
 
 	for _, existingVolumeConfig := range c.ExistingVolumeConfigs() {
 		volumeID := constants.ExistingVolumePrefix + existingVolumeConfig.Name()
+
+		// Build mount parameters based on configuration
+		var params []block.ParameterSpec
```
shanduur: Those are FS_CONFIG parameters, you want Mount Parameters.

pranav767: Could you give another try for integration test on github actions, I've updated `mount.go` to use mount params,

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/user_volumes.go
```diff
@@ -190,6 +190,20 @@ func ExistingVolumeTransformer(c configconfig.Config) ([]VolumeResource, error)
...
+
+		if existingVolumeConfig.Mount().DisableAccessTime() {
+			params = append(params, block.NewBooleanParameter("noatime"))
+		}
+
+		// Add secure mount options (nosuid, nodev) by default
+		if existingVolumeConfig.Mount().Secure() {
+			params = append(params, block.NewBooleanParameter("nosuid"))
+			params = append(params, block.NewBooleanParameter("nodev"))
+		}
```
shanduur: Okay, I've did some more reading and this is not a correct way to propagate the parameters. You should follow the same path as `ReadOnly` is doing, so the values are properly read and if needed, the volume is remounted. The `params` are read and set only once on the disk provisioning, as they are referring to the filesystem, and can't be updated - you need to reprovision the Volume to do this.

Also, this should be captured in the VolumeMountRequest:
https://github.com/siderolabs/talos/blob/4b274f76159495cc6c2977ec3bbade71e35aade8/internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/user_volumes.go#L339-L347

pranav767: Thanks for the tip @shanduur , 
After reading I've updated all 3 bits for mounts (mounts, mount_request and mount_status)

internal/pkg/mount/v3/point.go
```diff
@@ -75,10 +75,19 @@ func (p *Point) Mount(opts Options) error {
 			return fmt.Errorf("error mounting %q to %q: %w", p.Source(), p.target, err)
 		}
```
pranav767, line 77: I tried setting mount attributes directly using `mount_setattr`, but this failed with `EINVAL` when attempting to apply `MOUNT_ATTR_NOATIME | MOUNT_ATTR_NOSUID | MOUNT_ATTR_NODEV` on mounts created via the new mount API.

Example error:

```
ERROR controller failed {"component":"controller-runtime","controller":"block.MountController","error":"failed to mount \"e-test-data\": failed to mount: 1 error(s) occurred:
        error setting mount attributes on \"/var/mnt/test-data\": setattr failed for fd=70 target=\"/var/mnt/test-data\" flags=36864 attr=unix.MountAttr{Attr_set:0x16, Attr_clr:0x0, Propagation:0x0, Userns_fd:0x0}: invalid argument"}
```

I tested multiple variations:

- Calling mount_setattr both before and after move_mount
- Different flag combinations
- With and without AT_RECURSIVE

In all cases, setting these attributes via `mount_setattr` failed with `EINVAL`.
Using `MS_REMOUNT` works reliably for these flags, while `mount_setattr` does work correctly for `MOUNT_ATTR_RDONLY`.

pranav767:
Dropping some logs which I tried while testing with the changes:
```
~/siderolabs/talos$ cat /tmp/test-runtime-disable.yaml
 ---
 apiVersion: v1alpha1
 kind: ExistingVolumeConfig
 name: test-data
 discovery:
   volumeSelector:
     match: volume.partition_label == "u-test-volume"
 mount:
   disableAccessTime: false
   secure: false

~/siderolabs/talos$ sudo -E _out/talosctl-linux-amd64 patch mc -n 172.20.0.3 --mode=no-reboot -p @/tmp/test-runtime-disable.yaml && sleep 5 && sudo -E _out/talosctl-linux-amd64 read /proc/mounts -n 172.20.0.3 | grep test-data
patched MachineConfigs.config.talos.dev/v1alpha1 at the node 172.20.0.3
Applied configuration without a reboot
/dev/sda1 /var/mnt/test-data xfs rw,seclabel,noatime,inode64,logbufs=8,logbsize=32k,noquota 0 0
```

For Mount options: **disableAccessTime** & **secure** enabled
```

cat /tmp/test-runtime-enable2.yaml 
---
apiVersion: v1alpha1
kind: ExistingVolumeConfig
name: test-data
discovery:
  volumeSelector:
    match: volume.partition_label == "u-test-volume"
mount:
  disableAccessTime: true
  secure: true

sudo -E _out/talosctl-linux-amd64 patch mc -n 172.20.0.3 --mode=no-reboot -p @/tmp/test-runtime-enable2.yaml && sleep 3 && sudo -E _out/talosctl-linux-amd64 read /proc/mounts -n 172.20.0.3 | grep test-data
patched MachineConfigs.config.talos.dev/v1alpha1 at the node 172.20.0.3
Applied configuration without a reboot
/dev/sda1 /var/mnt/test-data xfs rw,seclabel,nosuid,nodev,noatime,inode64,logbufs=8,logbsize=32k,noquota 0 0
```

internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -690,6 +705,52 @@ func (ctrl *MountController) handleDiskMountOperation(
...
+	//nolint:dupl
+	if mountCtx.disableAccessTime != mountRequest.TypedSpec().DisableAccessTime {
+		var err error
+
+		switch mountRequest.TypedSpec().DisableAccessTime {
+		case true:
+			err = mountCtx.point.SetDisableAccessTime(true)
+		case false:
+			err = mountCtx.point.SetDisableAccessTime(false)
+		}
```
shanduur:
```suggestion
err = mountCtx.point.SetDisableAccessTime(mountRequest.TypedSpec().DisableAccessTime)
```

internal/pkg/mount/v3/point.go
```diff
@@ -284,20 +293,60 @@ func (p *Point) RemountReadWrite() error {
...
-		return fmt.Errorf("setattr failed for fd=%d target=%q flags=%d attr=%#+v: %w", fd, p.target, flags, *attr, err)
+	if (mountAttrs & unix.MOUNT_ATTR_NODEV) != 0 {
+		flags |= unix.MS_NODEV
+	}
+
+	if err := unix.Mount("", p.target, "", flags, ""); err != nil {
+		return fmt.Errorf("remount failed for target=%q flags=%#x: %w", p.target, flags, err)
+	}
+
+	return nil
```
shanduur: We specifically moved away from old mount API. Have you found why it's returning EINVAL on `unix.MountSetattr`?

shanduur: > MOUNT_ATTR_NOATIME
>   Do not update access times for (all types of) files
>   on this mount.
>
>   To enable this access-time setting on a mount or
>   mount tree, MOUNT_ATTR_NOATIME must be set in
>   attr_set and MOUNT_ATTR__ATIME must be set in the
>   attr_clr field.

https://man7.org/linux/man-pages/man2/mount_setattr.2.html

pranav767: Yes, thank you I missed this out setting `attr_clr` as  `MOUNT_ATTR__ATIME`.
I'll try again again with new mount API

pranav767: @shanduur , I've updated this to use `unix.MountSetattr`.
Tried this on a existingVolumeConfig, works fine.

shanduur:
@pranav767 thanks for your contribution! I'll take the PR from here and merge it after few minor adjustments.

pranav767:
Alright, thank you @shanduur 🙌

pkg/machinery/config/config/volume.go
smira, line 25: should we have this also for system volumes? or at least to default to Secure=true?

shanduur: Sure, `secure` definitely is a good call.

cmd/talosctl/cmd/talos/image.go
```diff
@@ -558,6 +558,7 @@ var imageIntegrationCmd = &cobra.Command{
 			"docker.io/library/alpine:latest",
 			"ghcr.io/siderolabs/talosctl:latest",
 			"registry.k8s.io/kube-apiserver:v1.27.0",
+			"registry.k8s.io/kube-apiserver:v1.27.1",
```
shanduur, line 561: Added missing image @smira, E2E for Airgapped and ImageCache were failing for those image missing.

shanduur: Extracted to https://github.com/siderolabs/talos/pull/12711

PR #12589, opened 2026-01-13
pcap(talosctl): Add backtick to wrape inline

alagoutte:
When looking talos doc, there is missing backtick for talospcap example 

https://docs.siderolabs.com/talos/v1.7/reference/cli#talosctl-pcap

### DIFF
--- a/cmd/talosctl/cmd/talos/pcap.go
+++ b/cmd/talosctl/cmd/talos/pcap.go
@@ -46,21 +46,21 @@ Default behavior is to decode the packets with internal decoder to stdout:
     talosctl pcap -i eth0
 
 Raw pcap file can be saved with ` + "`--output`" + ` flag:
-
+` + "```
     talosctl pcap -i eth0 --output eth0.pcap
-
+```" + `
 Output can be piped to tcpdump:
-
+" + ````
     talosctl pcap -i eth0 -o - | tcpdump -vvv -r -
-
+```" + `
 BPF filter can be applied, but it has to compiled to BPF instructions first using tcpdump.
 Correct link type should be specified for the tcpdump: EN10MB for Ethernet links and RAW
 for e.g. Wireguard tunnels:
-
+" + ````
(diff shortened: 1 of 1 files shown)


cmd/talosctl/cmd/talos/pcap.go
```diff
@@ -46,21 +46,21 @@ Default behavior is to decode the packets with internal decoder to stdout:
     talosctl pcap -i eth0
 
 Raw pcap file can be saved with ` + "`--output`" + ` flag:
-
```
smira, line 49: this is CLI help, that is `talosctl pcap --help`, it can't use Markdown here.

cmd/talosctl/cmd/talos/pcap.go
```diff
@@ -46,21 +46,21 @@ Default behavior is to decode the packets with internal decoder to stdout:
     talosctl pcap -i eth0
 
 Raw pcap file can be saved with ` + "`--output`" + ` flag:
```
alagoutte, line 48: there is already markdown here ?

cmd/talosctl/cmd/talos/pcap.go
alagoutte: how to get better (web) docs ?

cmd/talosctl/cmd/talos/pcap.go
smira: there might be minimal readable one, but certainly three backticks don't have a place there

cmd/talosctl/cmd/talos/pcap.go
smira: probably look into the way cobra generates markdown docs from the CLI docs and how these are converted

alagoutte: do you have a link about this ?

smira: start looking from https://github.com/pranav767/talos/blob/8184927316c5de7d9b04f21474a60cc791c3d26d/cmd/talosctl/cmd/docs.go#L84-L105

PR #12591, opened 2026-01-13
pcap(talosctl): move talosctl pcap example to Example

alagoutte:
it is before on Long but wrong display on docs website (already use by image cache-create cmd)


alagoutte:
may be backport to 1.12 ?

smira:
> may be backport to 1.12 ?

I think it's fine to keep it till 1.13, thank you

PR #12593, opened 2026-01-14
fix: wipe disk by signatures

### LINKED ISSUE #12491: Talos reset should wipe partition headers before device headers
### Description

When using TPM-based LUKS volume passwords, sequential installations using default partitioning can fail without visible cause due to stale LUKS headers left by the previous installation. Due to a lack of knowledge, this was discovered without logs, so most of this report is conjecture. 

The problem was found as the author muddled through learning Secure Boot and Talos at the same time. Out of an abundance of caution, when rebuilding the practice node, they would erase the volumes via `wipefs -fa /dev/sda[123] /def/sda`, then clear the TPM in BIOS setup. Note the wipe of the partitions before the volume. 

When both the disk and the TPM are cleared in this manner on the hardware available to test (Supermicro X10 vintage), install was reliable with the PCR7 fix in v12.0 (see https://github.com/siderolabs/talos/discussions/12476).

Later, when testing a new configuration, the "reset" functionality was discovered to clear the installation. This was a refreshing discovery that sped the development lifecycle, but added a new problem.

When the node was reset _without_ clearing the TPM, it appears that the reinstall worked correctly. When the TPM was cleared _with_ the reset, the node would boot to the splash screen and not progress. 

Presumably, the reason for this is the same partition geometry was aligned around the leftover LUKS headers and the boot tried to use the headers with the wrong password from the wiped TPM. 

Further investigation led to trying the `wipefs` technique from above, but with the partition table was already cleared, it impossible to wipe the LUKS labels. Finally, after zeroing the entire drive, the install started working again. 

### Resolution Strategy

It seems that when a node is reset, either the labels from each partition should be wiped as above (each partition before the partition map) and/or labels should be searched and erased when a geometry is written. Both would be valuable in various user situations. 

### Environment

- Talos 1.12.0 secureboot ISO (standard, no custom kernel args needed)
- Supermicro X10DRi-TR013 with Infineon SLB9665 TPM, (firmware 5.62.12.13824, TIS interface)
- SecureBoot enabled

smira:
Fixes #12491

In (almost) all places we previously used `FastWipe`, use instead a helper which will try to discover filesystem/partition signatures, and wipe them.

This fixes the issue when a partition re-created in the same place might already hit a scenario when the "old" filesystem is discovered in the same place.

### DIFF
--- a/internal/integration/api/environment.go
+++ b/internal/integration/api/environment.go
@@ -80,6 +80,7 @@ func (suite *EnvironmentSuite) TestEnvironment() {
 		suite.ctx, node, func(nodeCtx context.Context) error {
 			return base.IgnoreGRPCUnavailable(suite.Client.Reboot(nodeCtx))
 		}, 5*time.Minute,
+		suite.CleanupFailedPods,
 	)
 
 	suite.Require().Eventually(func() bool {
@@ -95,6 +96,7 @@ func (suite *EnvironmentSuite) TestEnvironment() {
 		suite.ctx, node, func(nodeCtx context.Context) error {
 			return base.IgnoreGRPCUnavailable(suite.Client.Reboot(nodeCtx))
 		}, 5*time.Minute,
+		suite.CleanupFailedPods,
 	)
 
 	suite.Require().Eventually(func() bool {
--- a/internal/integration/api/update-hostname.go
+++ b/internal/integration/api/update-hostname.go
@@ -137,31 +137,11 @@ func (suite *UpdateHostnameSuite) TestUpdateHostname() {
 }
 
 func (suite *UpdateHostnameSuite) updateHostname(nodeCtx context.Context, newHostname string) {
-	// [TODO]: this should be dropped once Terraform Provider is updated for Talos 1.12
-	// do a negative patch removing v1alpha1 hostname-related config
-	// In "normal" tests config is generated without hostname v1alpha1 parts,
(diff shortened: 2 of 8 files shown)


internal/integration/api/update-hostname.go
```diff
@@ -137,31 +137,11 @@ func (suite *UpdateHostnameSuite) TestUpdateHostname() {
 }
 
 func (suite *UpdateHostnameSuite) updateHostname(nodeCtx context.Context, newHostname string) {
-	// [TODO]: this should be dropped once Terraform Provider is updated for Talos 1.12
```
smira, line 140: I dropped this part for two reasons:

* first, we don't need it anymore (TF provider supports Talos v1.12)
* second, it is not going to work as `machine.network` section is gone now in v1.13 configs, and this was leading to panic I fixed in #12595, but still without the panic this patch would fail to apply

internal/integration/api/environment.go
```diff
@@ -80,6 +80,7 @@ func (suite *EnvironmentSuite) TestEnvironment() {
 		suite.ctx, node, func(nodeCtx context.Context) error {
 			return base.IgnoreGRPCUnavailable(suite.Client.Reboot(nodeCtx))
 		}, 5*time.Minute,
+		suite.CleanupFailedPods,
```
smira, line 83: I added this @shanduur as I saw the test failing as Kubernetes was not cleaning up failed pods quick enough between reboots

PR #12594, opened 2026-01-14
fix: use network config from config-drive even when user_data is missing

maxpain:
When OpenStack config-drive contains `meta_data.json` and `network_data.json` but `no user_data`, the `NetworkConfiguration` method was incorrectly falling back to network metadata service. This happened because `configFromCD` returns `ErrNoConfigSource` when `user_data` is missing, causing the already-read metadata and network config to be discarded.

This fix checks if metadata was successfully read from config-drive before falling back to network, allowing network configuration to be applied even when user_data is not provided (e.g., in OpenStack Ironic bare metal setups where machine config is applied later via talosctl).

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
@@ -357,7 +357,8 @@ func (o *OpenStack) ParseMetadata(
 // Configuration implements the runtime.Platform interface.
 func (o *OpenStack) Configuration(ctx context.Context, r state.State) (machineConfig []byte, err error) {
 	_, _, machineConfig, err = o.configFromCD(ctx, r)
-	if err != nil {
+	// Fall back to network if config drive not found or user_data missing
+	if stderrors.Is(err, errors.ErrNoConfigSource) || stderrors.Is(err, errors.ErrNoUserData) {
 		if err = netutils.Wait(ctx, r); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
```diff
@@ -395,7 +395,11 @@ func (o *OpenStack) NetworkConfiguration(ctx context.Context, st state.State, ch
 	networkSource := false
 
 	metadataConfigDl, metadataNetworkConfigDl, _, err := o.configFromCD(ctx, st)
```
shanduur, line 400: In `configFromCD` we appear to be shadowing/discarding some of the underlying errors. 

Your changes are inferring intent based on `metadataConfigDl` / `metadataNetworkConfigDl` being `nil`. This makes the fallback logic harder to reason about and risks masking partial-failure cases.

It might be clearer and safer to explicitly return (or wrap) sentinel errors (e.g. via [`errors.Join`](https://pkg.go.dev/errors#Join)) from `configFromCD`, and then rely on [`errors.Is`](https://pkg.go.dev/errors#Is) here instead of checking for data nilness. That would decouple error semantics from data shape and make the fallback decision more robust.

smira (commented):
I don't think I can say I like the code today in OpenStack, neither I like the proposed change.

Given that we have no way to test this, there is no meaningful way to test this change.

I _think_ the logic of choosing between CD/network should be separate out and removed from the functions.

Only one of CD/network should be used ever and consistently for any requests.

PR #12600, opened 2026-01-15
fix: sort mirrors and tls configs when generating the machine config

Unix4ever:
As registry mirrors are using map to save all defined options without sort it is not deterministic.

### DIFF
--- a/pkg/machinery/config/generate/registry.go
+++ b/pkg/machinery/config/generate/registry.go
@@ -7,6 +7,8 @@ package generate
 import (
 	"fmt"
 	"net/url"
+	"slices"
+	"strings"
 
 	"github.com/siderolabs/go-pointer"
 
@@ -16,7 +18,7 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/config/types/v1alpha1"
 )
 
-//nolint:gocyclo
+//nolint:gocyclo,cyclop
 func (in *Input) generateRegistryConfigs(machine *v1alpha1.MachineConfig) ([]config.Document, error) {
 	if !in.Options.VersionContract.MultidocNetworkConfigSupported() {
 		// old-style registry config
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/generate/registry.go
```diff
@@ -112,5 +114,20 @@ func (in *Input) generateRegistryConfigs(machine *v1alpha1.MachineConfig) ([]con
 		documents = append(documents, tlsConfig)
 	}
 
+	// sort the TLS config and registry mirrors docs alphabetically by the name
+	slices.SortStableFunc(documents, func(a, b config.Document) int {
+		na, ok := a.(config.NamedDocument)
+		if !ok {
+			return 0
```
smira: This is wrong sort order, as not all documents are named.

I think we should first compare by Kind, and if Kind(s) are equal, sort by name

Unix4ever: yeah, good idea

Unix4ever: in this particular method all the docs are named though, but I'll fix it anyways

pkg/machinery/config/generate/registry.go
```diff
@@ -112,5 +114,17 @@ func (in *Input) generateRegistryConfigs(machine *v1alpha1.MachineConfig) ([]con
 		documents = append(documents, tlsConfig)
 	}
 
+	// sort the TLS config and registry mirrors docs alphabetically by the name
+	slices.SortStableFunc(documents, func(a, b config.Document) int {
+		na, aok := a.(config.NamedDocument)
+		nb, bok := b.(config.NamedDocument)
+
+		if aok || bok {
```
smira: this looks wrong to me...

it should be instead

```go
if c := stirngs.Compare(a.Kind(), b.Kind(); c != 0 {
   return c
}

if aok && bok {
  return strings.Compare(na.Name(), nb.Name)
}

return 0
```

Unix4ever: yeah, I messed up, I wanted to do `if !aok || !bok {`, then it would make sense.
But your option is more readable anyways.

Unix4ever:
> I knew it was Omni, not machinery :D

It is machinery ;)

Issue #12526, opened 2025-12-31
The OOM killer was incorrectly triggered in Talos v1.12.0


StormYudi:
### Description
We are using Cilium + Ceph, and we recently upgraded our Talos version to v1.12.0. Tonight, we discovered that the Cilium container is frequently restarting. We consulted the documentation and it seems to be related to a bug in the new feature described in https://docs.siderolabs.com/talos/v1.12/configure-your-talos-cluster/system-configuration/oom.

### Logs
cilium events shows:
`Normal | SandboxChanged | kubelet | Pod sandbox changed, it will be killed and re-created. | 29 minutes (182 times over 2 hours)`
Upon checking the Talos dashboard, we found that the following events were frequently triggered:
```
 user: warning: [2025-12-31T13:36:46.234797064Z]: [talos] victim processes: {"component": "controller-runtime", "controller": "runtime.OOMController", "processes": [231126]}                             
 user: warning: [2025-12-31T13:36:54.230729064Z]: [talos] OOM controller triggered {"component": "controller-runtime", "controller": "runtime.OOMController"}                                             
 user: warning: [2025-12-31T13:36:54.235373064Z]: [talos] Sending SIGKILL to cgroup {"component": "controller-runtime", "controller": "runtime.OOMController", "cgroup": "/sys/fs/cgroup/kubepods/        
 burstable/podacb30f59-0d20-4521-b6b8-9618cd6a3be6"}                                                                                                                                                      
 user: warning: [2025-12-31T13:36:54.235761064Z]: [talos] victim processes: {"component": "controller-runtime", "controller": "runtime.OOMController", "processes": [231405]}                             
 user: warning: [2025-12-31T13:36:58.230521064Z]: [talos] OOM controller triggered {"component": "controller-runtime", "controller": "runtime.OOMController"}                                             
 user: warning: [2025-12-31T13:36:58.234951064Z]: [talos] Sending SIGKILL to cgroup {"component": "controller-runtime", "controller": "runtime.OOMController", "cgroup": "/sys/fs/cgroup/kubepods/        
 burstable/podacb30f59-0d20-4521-b6b8-9618cd6a3be6"}                                                                                                                                                      
 user: warning: [2025-12-31T13:36:58.235235064Z]: [talos] victim processes: {"component": "controller-runtime", "controller": "runtime.OOMController", "processes": [232586]}                             
[...]

PR #12602, opened 2026-01-15
fix: make OOM controller more precise by considering separate cgroup PSI

dsseng:
This should reduce false triggers due to high IO activity and similar
events increasing global memory PSI despite free memory being available.

Fixes: #12526

### DIFF
--- a/pkg/machinery/constants/constants.go
+++ b/pkg/machinery/constants/constants.go
@@ -1282,12 +1282,16 @@ const (
 	ContainerMarkerFilePath = "/usr/etc/in-container"
 
 	// DefaultOOMTriggerExpression is the default CEL expression used to determine whether to trigger OOM.
-	DefaultOOMTriggerExpression = `memory_full_avg10 > 12.0 && d_memory_full_avg10 > 0.0 && time_since_trigger > duration("500ms")`
+	DefaultOOMTriggerExpression = `(multiply_qos_vectors(d_qos_memory_full_total, {System: 8.0, Podruntime: 4.0}) > 3000.0) ||
+		(memory_full_avg10 > 75.0 && time_since_trigger > duration("10s"))`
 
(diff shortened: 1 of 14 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -1282,13 +1282,19 @@ const (
 	ContainerMarkerFilePath = "/usr/etc/in-container"
 
 	// DefaultOOMTriggerExpression is the default CEL expression used to determine whether to trigger OOM.
-	DefaultOOMTriggerExpression = `memory_full_avg10 > 12.0 && d_memory_full_avg10 > 0.0 && time_since_trigger > duration("500ms")`
+	DefaultOOMTriggerExpression = `((init_memory_full_avg10 > 5.0 && d_init_memory_full_avg10 > 0.0) ||
+		(system_memory_full_avg10 > 15.0 && d_system_memory_full_avg10 > 0.0) ||
```
smira: I like the direction we're moving to, but I wonder if this is the readable/best version possible.

What I was thinking in math terms is:

```
Weights := MAP[QoS: float64]
PSI_weighted := SUM(PSI_by_cgroup[QoS] * Weights[QoS])
d_PSI_weighted := SUM(d_PSI_by_cgroup[QoS] * Weights[QoS])

if PSI_weighted > threshold && d_PSI_weighted > 0 => trigger
```

smira: so this expression would be somewhat similar to ranking expression. 

we can use CEL macros or allow functions to make this readable

dsseng: I might explore making a map of maps to represent those per-class PSIs, yes, will take a look once I make the formula stable

smira: I'm happy to hack together on this...

smira (approved):

PR #12603, opened 2026-01-15
fix: add hostname to endpoints

smira:
Populate endpoint coming from the Kubernetes controlplane endpoint with the hostname (if the endpoint is a hostname).

This should improve cases when hostname is used for the endpoint in terms of SNI, proper resolving of DNS if it's dynamic.

See https://github.com/siderolabs/talos/pull/12556#issuecomment-3755862314

### DIFF
--- a/api/resource/definitions/k8s/k8s.proto
+++ b/api/resource/definitions/k8s/k8s.proto
@@ -93,9 +93,10 @@ message ControllerManagerConfigSpec {
   Resources resources = 9;
 }
 
-// EndpointSpec describes status of rendered secrets.
+// EndpointSpec describes a list of endpoints to connect to.
 message EndpointSpec {
   repeated common.NetIP addresses = 1;
+  repeated string hosts = 2;
 }
 
 // ExtraManifest defines a single extra manifest to download.
--- a/internal/app/machined/pkg/controllers/k8s/static_endpoint.go
+++ b/internal/app/machined/pkg/controllers/k8s/static_endpoint.go
@@ -75,6 +75,7 @@ func (ctrl *StaticEndpointController) Run(ctx context.Context, r controller.Runt
 			var (
 				resolver net.Resolver
 				addrs    []netip.Addr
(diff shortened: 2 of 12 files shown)


smasset-orange:
Thanks for seeing this through.

Successfully tested with Talos nodes running as pods in isolated namespaces behind external LB.

Moving on to generating ISO and testing with Kamaji's hosted control-plane and talos-csr-signer.

smira:
> Thanks for seeing this through.

sorry for not seeing it sooner, looking at your PR, the context that endpoints are just IPs was completely lost.

smasset-orange:
> Thanks for seeing this through.
> 
> Successfully tested with Talos nodes running as pods in isolated namespaces behind external LB.
> 
> Moving on to generating ISO and testing with Kamaji's hosted control-plane and talos-csr-signer.

@smira, sorry for the late reply but I was hoping to find the root cause for some issue before reporting back but got side-tracked.

For some unrelated but yet unknown reason, our openstack based hypervisor can't boot Talos images >= 1.12.0 : very early in the boot sequence, the system crashes and reboots in a loop.

I was able to boot with 1.11.6 though. Is there any chance to backport this PR to 1.11 ? I saw this PR for [backports to 1.12](https://github.com/siderolabs/talos/pull/12731) and dropped a comment there too just in case I can find a way to boot it before the next release.

For the record, I was able to backport it to 1.10 with some (minimal) conflict resolution effort so I can probably do the same for 1.11. Do you accept contributions for backporting efforts ?

smasset-orange:
Opened #12735 for the boot issue

smira:
> I was able to boot with 1.11.6 though. Is there any chance to backport this PR to 1.11 ? I saw this PR for https://github.com/siderolabs/talos/pull/12731 and dropped a comment there too just in case I can find a way to boot it before the next release.

We do not plan to release 1.11.x unless there is a security issue that needs to be fixed.

PR #12605, opened 2026-01-16
feat: update Linux to 6.18.5

smira:
Also test new NVIDIA drivers.


smira:
NVIDIA drivers are not testable due to BIOS boot issue.

smira:
Repurposed this to bump Go & pkgs, I will test NVIDIA separately after BIOS issue is fixed.

PR #12606, opened 2026-01-16
feat: add ProbeConfig for network connectivity probes

mcanevet:
# Pull Request

## What? (description)

Add support for configuring network connectivity probes via machine configuration. This allows users to define TCP probes that validate network connectivity without requiring a default route (0.0.0.0/0).

Example usage:
```yaml
---
apiVersion: v1alpha1
kind: ProbeConfig
name: proxy-check
interval: 1s
failureThreshold: 3
tcp:
  endpoint: proxy.example.com:3128
  timeout: 10s
```

## Why? (reasoning)

Fixes this: https://github.com/siderolabs/talos/discussions/12400

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/network/probe_config.go
+++ b/internal/app/machined/pkg/controllers/network/probe_config.go
@@ -81,7 +81,7 @@ func (ctrl *ProbeConfigController) Run(ctx context.Context, r controller.Runtime
 		}
 
 		if err = r.CleanupOutputs(ctx,
-			resource.NewMetadata(network.NamespaceName, network.ProbeSpecType, "", resource.VersionUndefined),
+			resource.NewMetadata(network.ConfigNamespaceName, network.ProbeSpecType, "", resource.VersionUndefined),
 		); err != nil {
 			return fmt.Errorf("error cleaning up outputs: %w", err)
 		}
@@ -100,7 +100,7 @@ func (ctrl *ProbeConfigController) apply(ctx context.Context, r controller.Runti
 		if err := safe.WriterModify(
 			ctx,
 			r,
-			network.NewProbeSpec(network.NamespaceName, id),
+			network.NewProbeSpec(network.ConfigNamespaceName, network.LayeredID(spec.ConfigLayer, id)),
 			func(r *network.ProbeSpec) error {
 				*r.TypedSpec() = spec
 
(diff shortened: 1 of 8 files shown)


internal/app/machined/pkg/controllers/network/probe_config.go
```diff
@@ -0,0 +1,195 @@
...
+			return ids, err
+		}
+
+		ids = append(ids, id)
+	}
+
+	return ids, nil
+}
+
+func (ctrl *ProbeConfigController) parseMachineConfiguration(cfgProvider talosconfig.Provider, logger *zap.Logger) []network.ProbeSpecSpec {
```
shanduur: Instead of going like this, you should use `cfgProvider.Confif().ProbeConfig()` (or similar).

You won't be relying on a specific type.

You can take a look at the PR I've done recently to get the inspiration:
https://github.com/siderolabs/talos/pull/12548

mcanevet: @shanduur is it better now?

pkg/machinery/config/types/network/probe.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+// ProbeConfigV1Alpha1 is a config document to configure network connectivity probes.
+//
+//	examples:
+//	  - value: exampleProbeConfigV1Alpha1()
+//	alias: ProbeConfig
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/ProbeConfig
+type ProbeConfigV1Alpha1 struct {
```
smira: notes for myself:

- ~It should be `TCPProbeConfig` with `CommonProbeConfig` embedded~
- ~same with interfaces, and they should have `Network` prefix (similar to various flavors of links)~
- ~first, before doing any changes, introduce layered config & merging for probes, update platform apply config controller for it, introduce merge controller~
- ~add hack/release.toml note~

internal/app/machined/pkg/controllers/network/probe_config.go
shanduur: Yeah, that's cool, thanks for your contribution!

smira:
> @smira do you want to make changes in this PR, or will we create a follow-up from your comments?

I will push another commit with layering, but so far I fixed all other stuff.

PR #12607, opened 2026-01-16
fix: use append instead of prepend in service-account-issuer

shanduur:
Changing `.cluster.controlPlane.endpoint=$NEW` will cause old tokens to be no longer valid.
We want to ensure that new tokens are issued using the `.cluster.controlPlane.endpoint=$NEW` value,
but all the existing tokens (issued using `.cluster.controlPlane.endpoint=$OLD`) are still accepted.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -141,7 +141,9 @@ BREAKING: If you were relying on the resources EtcdConfigs, KubeletConfigs, Cont
     [notes.serviceAccountIssuer]
     title = "Service Account Issuer configuration"
     description = """\
-In API Server, passing extra args with `service-account-issuer` will prepend them before default value.
+In API Server, passing extra args with `service-account-issuer` will append them after default value.
+This allows easy migration, e.g. by changing `.cluster.controlPlane.endpoint` to new value, and keeping the old value in
+`.cluster.apiServer.extraArgs["service-account-issuer"]`.
 """
 
 [make_deps]
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -429,7 +429,7 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 		"etcd-keyfile":                     argsbuilder.MergeDenied,
 		"kubelet-client-certificate":       argsbuilder.MergeDenied,
 		"kubelet-client-key":               argsbuilder.MergeDenied,
-		"service-account-issuer":           argsbuilder.MergePrepend,
+		"service-account-issuer":           argsbuilder.MergeAppend,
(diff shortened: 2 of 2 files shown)


smira (approved):

eplightning:
Does this not break setups that want to override the issuer URL (to serve OIDC discovery data from different endpoint, etc.)?

eplightning:
Seems to be a breaking change that makes it impossible to setup workload identity properly. Ideally the behaviour should be just to replace default value if user provided its own values.

frezbo:
> Seems to be a breaking change that makes it impossible to setup workload identity properly. Ideally the behaviour should be just to replace default value if user provided its own values.

This is on purpose so a user doesn't remove the default kubernetes endpoint, Talos just works fine with external OIDC providers, if there's a bug please create an issue

eplightning:
I'll try to post an issue later today, but the problem with appending is that the first value is used as iss claim in the tokens. And that has to match issuer returned by OIDC discovery which in turn needs to be the same as URL requested by the other side (e.g. by Azure Federated identity). 

If there's no way to override first value then workload identity on Azure, AWS, GCP stops working unless you expose control plane on the internet with trusted web PKI.

Managed Kubernetes clusters also override that value so its a standard setup, there's no real reason for the issuer to be equal control plane endpoint.

frezbo:
> I'll try to post an issue later today, but the problem with appending is that the first value is used as iss claim in the tokens. And that has to match issuer returned by OIDC discovery which in turn needs to be the same as URL requested by the other side (e.g. by Azure Federated identity).
> 
> If there's no way to override first value then workload identity on Azure, AWS, GCP stops working unless you expose control plane on the internet with trusted web PKI.
> 
> Managed Kubernetes clusters also override that value so its a standard setup, there's no real reason for the issuer to be equal control plane endpoint.

That is weird, this used to work with AWS and GCP federation with the first value being the cluster endpoint and the second the cloud specific endpoint, let's create a new issue if that's not the case

frezbo:
See https://docs.siderolabs.com/kubernetes-guides/advanced-guides/gcp-workload-identity

eplightning:
Yes that's pretty much the same setup I'm using.

The problem is after this PR (and Talos 1.13 I guess?) this will stop working because `service-account-issuer` specified in `extraArgs` will be appended instead of prepended - which results in service account tokens issued by Kubernetes to be always have iss token claim equal to control plane URL (because iss can only be a single value, and api-server will use firstprovided value).

frezbo:
> Yes that's pretty much the same setup I'm using.
> 
> The problem is after this PR (and Talos 1.13 I guess?) this will stop working because `service-account-issuer` specified in `extraArgs` will be appended instead of prepended - which results in service account tokens issued by Kubernetes to be always have iss token claim equal to control plane URL (because iss can only be a single value, and api-server will use firstprovided value).

let's create a new issue then

PR #12608, opened 2026-01-16
feat: add support for configurable route types

mcanevet:
# Pull Request

## What? (description)

Adds support for specifying route type in network configuration for Talos 1.12+ document-based configs. This enables creation of blackhole routes and other route types (unreachable, prohibit, etc.) in addition to the default unicast routes.

Key changes:
- Add RouteType field to RouteConfig struct in link.go
- Add Type() method to NetworkRouteConfig interface
- Update RouteConfigController.processMachineConfig to use Type from configuration
- Add unit test TestMachineConfigurationWithRouteType for blackhole and unicast routes
- Generate documentation for new RouteType field

Note: This feature is only available in 1.12+ document-based network configuration. v1alpha1 (pre-1.12) configs continue to default to unicast routes as the cfg.Route interface doesn't support the Type() method.

## Why? (reasoning)

In order to workaround the connectivity check that requires to set a default gateway, we'd like to create a blackhole for default gateway, but it is currently not possible because the route type is not exposed.

see. https://github.com/siderolabs/talos/discussions/12400

## Acceptance

### DIFF
--- a/pkg/machinery/config/types/network/link.go
+++ b/pkg/machinery/config/types/network/link.go
@@ -168,6 +168,23 @@ type RouteConfig struct {
 	//   schema:
 	//     type: string
 	RouteTable nethelpers.RoutingTable `yaml:"table,omitempty"`
+	//   description: |
+	//     The route type.
+	//
+	//     If not specified, the route type will be unicast (or multicast for multicast destinations).
+	//     Common types: unicast, local, broadcast, blackhole, unreachable, prohibit.
+	//   values:
+	//     - "local"
+	//     - "broadcast"
+	//     - "unicast"
+	//     - "multicast"
+	//     - "blackhole"
+	//     - "unreachable"
+	//     - "prohibit"
+	//     - "throw"
+	//     - "nat"
+	//     - "xresolve"
+	RouteType nethelpers.RouteType `yaml:"type,omitempty"`
 }
 
 // NewLinkConfigV1Alpha1 creates a new LinkConfig config document.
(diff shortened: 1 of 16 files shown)


pkg/machinery/config/types/network/link.go
```diff
@@ -168,6 +168,14 @@ type RouteConfig struct {
...
 	//     type: string
 	RouteTable nethelpers.RoutingTable `yaml:"table,omitempty"`
+	//   description: |
+	//     The route type.
+	//
+	//     If not specified, the route type will be unicast.
+	//     Common types: unicast, local, broadcast, blackhole, unreachable, prohibit.
+	//   schema:
+	//     type: string
+	RouteType nethelpers.RouteType `yaml:"type,omitempty"`
```
smira, line 187: I guess this is missing the actual nethelpers.go changes.

mcanevet: `RouteType` already exists in pkg/machinery/nethelpers/routetype.go with all the necessary YAML marshaling support and constants (`TypeBlackhole`, `TypeUnreachable`, etc.). Could you clarify what additional changes you'd 
  like to see in the nethelpers package? I'm just using the existing type here.

smira: oops, I missed that we already have it :)

smira:
LGTM, I will do some small changes and fixups last week, thank you for your PR!

smira:
🚧  I think this is not ready to go in any way. 

Blackhole routes are actually treated special way, and attached to the `lo` interface. 

I don't think we should even try to support them in the current config.

We need to have a special document which configures such routes, something like `BlackholeRouteConfig`.

mcanevet:
@smira do you want me to try with a special document like I did for the ProbeConfig here: https://github.com/siderolabs/talos/pull/12606?

smira:
> @smira do you want me to try with a special document like I did for the ProbeConfig here: #12606?

Yes, I think we need to have a specific document like:

```yaml
version: v1alpha1
kind: BlackholeRouteConfig
name: 10.0.0.0/24 # subnet
```

I don't think anything else is needed.

They should be translated into `RouteSpec` attached to the `lo` interface with type == blackhole. 

(I don't know if there's anything else missing that might be useful to configure?)

Preisschild:
> (I don't know if there's anything else missing that might be useful to configure?)

Metric/priority so you can have a fallback route

smira:
Moved to #12863

PR #12611, opened 2026-01-16
refactor: generate GRUB images 

smira:
Simplify the flow a bit by using live partition info,
avoid doing some calculations which are already done in the
partition code.

Remove some steps I believe we don't need to do.

### DIFF
--- a/cmd/installer/pkg/install/install.go
+++ b/cmd/installer/pkg/install/install.go
@@ -499,7 +499,7 @@ func (i *Installer) handleMeta(ctx context.Context, mode Mode, previousLabel str
 			return fmt.Errorf("failed to write back META partition data: %w", err)
 		}
 
-		return gptdev.Sync()
+		return nil
 	default:
 		return fmt.Errorf("unknown image mode: %d", mode)
 	}
@@ -692,11 +692,11 @@ func (i *Installer) formatPartitions(ctx context.Context, mode Mode, parts []par
 			}
 		}
 
-		if err := i.handleGrubBlocklist(gptdev, pt, parts); err != nil {
+		if err := i.handleGrubBlocklist(gptdev, pt); err != nil {
 			return fmt.Errorf("failed to handle GRUB blocklist: %w", err)
 		}
 
-		return gptdev.Sync()
+		return nil
 	default:
 		return fmt.Errorf("unknown image mode: %d", mode)
 	}
@@ -761,7 +761,7 @@ func (i *Installer) handlePartitionDataPopulation(ctx context.Context, idx int, 
 }
 
 //nolint:gocyclo,cyclop
-func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table, partitionOptions []partition.Options) error {
+func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table) error {
 	if i.options.Arch != "amd64" {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


internal/pkg/partition/helpers.go
```diff
@@ -42,8 +42,6 @@ func WipeWithSignatures(bd *block.Device, deviceName string, log func(string, ..
 				),
 			)
 		}
-
```
frezbo: i guess unrelated fix?

cmd/installer/pkg/install/install.go
```diff
@@ -770,17 +770,23 @@ func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table, partit
...
+		}
 	}
 
-	sectorSize := gptdev.GetSectorSize()
+	if biosPartition == nil {
+		return fmt.Errorf("failed to find BOOT partition for GRUB blocklist handling")
+	}
 
-	if err := grub.PatchBlocklistsForDiskImage(sectorSize, efiPartitionInfo[0].Size, i.options.MountPrefix); err != nil {
+	if err := grub.PatchBlocklistsForDiskImage(sectorSize, biosPartition.FirstLBA, i.options.MountPrefix); err != nil {
```
frezbo, line 795: ahh I see, rather than relying on the EFI partition size, we take the firstLBA of boot partition, this is better

frezbo (approved):

internal/pkg/partition/helpers.go
smira: It is somewhat related, but I will split it up

smira: I need to look more, but we wipe vfat signatures, while mkfs.xfs still thinks it's vfat and fails to format

smira: I need to look more, but we wipe vfat signatures, while mkfs.xfs still thinks it's vfat and fails to format

smira: I need to look more, but we wipe vfat signatures, while mkfs.xfs still thinks it's vfat and fails to format

cmd/installer/pkg/install/install.go
smira: Yes, and we avoid the math as well, as GPT LBA is a sector number

smira: Yes, and we avoid the math as well, as GPT LBA is a sector number

cmd/installer/pkg/install/install.go
```diff
@@ -499,7 +499,7 @@ func (i *Installer) handleMeta(ctx context.Context, mode Mode, previousLabel str
 			return fmt.Errorf("failed to write back META partition data: %w", err)
 		}
 
-		return gptdev.Sync()
```
smira, line 502: both .Sync() calls were in the "image" mode, so they flush the pre-allocated file to disk which results in tons of useless I/O, as the file has gaps, and we're going to compress and throw it away in a moment.

So don't call `fsync()` on the disk image.

cmd/installer/pkg/install/install.go
```diff
@@ -770,84 +770,53 @@ func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table, partit
...
+		biosPartition      *gpt.Partition
+	)
 
-	// handle the BIOS GRUB partition
-	biosPartitionInfo := xslices.Filter(partitionOptions, func(p partition.Options) bool {
-		return p.Label == constants.BIOSGrubPartitionLabel
-	})
+	for idx, p := range pt.Partitions() {
+		if p.Name == constants.BIOSGrubPartitionLabel {
+			biosPartition = p
```
smira, line 783: if we find the partition in the live partition table, we find the `FirstLBA` which is the first sector - this is what we need for patching, drop all the math later

cmd/installer/pkg/install/install.go
```diff
@@ -770,84 +770,53 @@ func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table, partit
...
 
-	if len(coreImgData) > int(biosPartitionInfo[0].Size) {
-		return fmt.Errorf("core.img size (%d bytes) exceeds BIOS partition size (%d bytes)", len(coreImgData), biosPartitionInfo[0].Size)
+	if biosPartition == nil {
+		return fmt.Errorf("failed to find BOOT partition for GRUB blocklist handling")
 	}
 
-	partitionImageFile := filepath.Join(i.options.MountPrefix, biosPartitionInfo[0].Label+".img")
-
-	if err := utils.CreateRawDisk(i.options.Printf, partitionImageFile, int64(biosPartitionInfo[0].Size), true); err != nil {
```
smira, line 807: we don't need an intermediate file to have an image of the future partition which contains `core.img` -> just write `core.img` directly

cmd/installer/pkg/install/install.go
```diff
@@ -861,7 +830,7 @@ func (i *Installer) handleGrubBlocklist(gptdev gpt.Device, pt *gpt.Table, partit
 
 	mbr := make([]byte, 446)
 
-	if _, err := bootImg.ReadAt(mbr, 0); err != nil {
+	if _, err := io.ReadFull(bootImg, mbr); err != nil {
```
smira, line 833: `Read()` call is not guaranteed to read all bytes (vs. the `Write` call), so use `ReadFull`

frezbo: i thought newer go fixed the Read(), anyways let's be safe

frezbo (approved):
🆒

PR #12620, opened 2026-01-19
fix: wipe the first/last 1MiB in addition to wiping by signatures

smira:
I got a failure when dual-boot image refuses to format EPHEMERAL partition where `EFI` partition used to be (VFAT).

So until we have a resolution, do this workaround.

### DIFF
--- a/internal/pkg/partition/helpers.go
+++ b/internal/pkg/partition/helpers.go
@@ -42,10 +42,12 @@ func WipeWithSignatures(bd *block.Device, deviceName string, log func(string, ..
 				),
 			)
 		}
-
-		return nil
 	}
 
+	// [TODO]: wipe the first/last 1MiB after wiping by signatures to cover somewhat unknown edge cases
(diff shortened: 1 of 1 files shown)


smira:
Split from #12611

internal/pkg/partition/helpers.go
```diff
@@ -42,10 +42,12 @@ func WipeWithSignatures(bd *block.Device, deviceName string, log func(string, ..
 				),
 			)
 		}
-
-		return nil
 	}
 
+	// [TODO]: wipe the first/last 1MiB after wiping by signatures to cover somewhat unknown edge cases
```
frezbo, line 47: I wonder if I dropped force option to xfs with imager changes, and this was always there? 🤔

smira: good question, but we wipe before format... and I added wipe by signatures instead of wipe 1MiB, so I thought it's my fault

PR #12621, opened 2026-01-19
fix: boot with GRUB

smira:
The problem is that xfs with 6.18 LTS settings is not supported by GRUB yet. It might be supported with newly released 2.14 though.


smira:
Split from #12611

frezbo:
> The problem is that xfs with 6.18 LTS settings is not supported by GRUB yet. It might be supported with newly released 2.14 though.

isn;t bumping and trying new grub worth it?

smira:
> > The problem is that xfs with 6.18 LTS settings is not supported by GRUB yet. It might be supported with newly released 2.14 though.
> 
> isn;t bumping and trying new grub worth it?

probably, but let's split it up. right now let's fix things and make sure we can release IF at any moment. 

new GRUB might have its own set of bugs

PR #12622, opened 2026-01-19
feat: add support for negative max size

### LINKED ISSUE #12405: Support negative max size for volumes (grow, but leave "X GiB" unallocated)
I want EPHEMERAL to occupy most of the disk, but not all of it (reserve some space for swap/user volumes, etc.)

So instead of specifying exact maxSize which might vary, support the following:
* grow to maximum size, but leave "X GiB" unallocated

shanduur:
Add support for negative max size values in volume configuration. Negative max size represents the amount of space to be left free on the device, rather than the size the volume should consume. For example, a max size of "-10GiB" means the volume can grow to the device size minus 10GiB.

Closes #12405

### DIFF
--- a/api/resource/definitions/block/block.proto
+++ b/api/resource/definitions/block/block.proto
@@ -169,6 +169,7 @@ message PartitionSpec {
   string label = 4;
   string type_uuid = 5;
   uint64 relative_max_size = 6;
+  bool negative_max_size = 7;
 }
 
 // ProvisioningSpec is the spec for volume provisioning.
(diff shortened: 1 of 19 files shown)


internal/integration/base/api.go
```diff
@@ -590,6 +591,12 @@ func (apiSuite *APISuite) PatchMachineConfigWithModeSetter(
 		marshaled, err := yaml.Marshal(patch)
 		apiSuite.Require().NoError(err)
 
+		// create temp file and write to it. Log the name of the file
+		tempFile, err := os.CreateTemp("", "patch")
+		apiSuite.Require().NoError(err)
+		_, err = tempFile.Write(marshaled)
```
smira: debug stuff?

shanduur: Yes sir!

shanduur: Removed.

PR #12627, opened 2026-01-20
fix: check if the device is not mounted when wiping

smira:
Open the blockdevice in `O_EXCL` mode when wiping to ensure that we don't wipe a mounted device.

This issue was discovered via #12620, when we wipe a blockdevice which is still mounted ending up in a wrong state.


shanduur (approved):

frezbo:
so do we need #12620 ?

smira:
> so do we need #12620 ?

yes, actually #12620 uncovered a bug in the integration tests which this PR fixes

PR #12630, opened 2026-01-20
fix: overwrite resolver config with machine config

### LINKED ISSUE #12614: Unable to remove default IPv4 DNS resolvers in single-stack IPv6 setup
## Bug Report

I am unable to remove the default IPv4 DNS resolvers (1.1.1.1 and 8.8.8.8) from my talos nodes when using the following ResolverConfig:

```yaml
---
apiVersion: v1alpha1
kind: ResolverConfig
nameservers:
    - address: 2606:4700:4700::64
    - address: 2606:4700:4700::6400
```

### Description

The issue seems to be caused by https://github.com/siderolabs/talos/commit/c288ace7b185cd3fad569c0848afbda7217ac269 . In some cases this behaviour may make sense (eg; dual-stack DHCP setups), but not being able to disable the IPv4 resolvers makes DNS unstable in single-stack IPv6 scenarios.

It might be possible to filter the configured DNS resolvers based on if a route exists for that IP address. I'd be happy to contribute a fix if you agree with that method.

### Logs

```
<node ip>: user: warning: [2026-01-16T22:55:43.734619883Z]: [talos] setting resolvers {"component": "controller-runtime", "controller": "network.ResolverSpecController", "resolvers": ["2606:4700:4700::64", "2606:4700:4700::6400", "1.1.1.1", "8.8.8.8"], "searchDomains": []}
```

### Environment

- Talos version: v1.12.1
- Kubernetes version: v1.35.0
- Platform: amd64 on Proxmox (nocloud image)

smira:
Fixes #12614

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -139,8 +139,8 @@ BREAKING: If you were relying on the resources EtcdConfigs, KubeletConfigs, Cont
 """
 
     [notes.serviceAccountIssuer]
-    title = "Service Account Issuer configuration"
-    description = """\
+        title = "Service Account Issuer configuration"
+        description = """\
 In API Server, passing extra args with `service-account-issuer` will append them after default value.
 This allows easy migration, e.g. by changing `.cluster.controlPlane.endpoint` to new value, and keeping the old value in
 `.cluster.apiServer.extraArgs["service-account-issuer"]`.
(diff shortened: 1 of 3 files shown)


frezbo (approved):

lucasvbeek:
@smira is there any chance this could be included in the 1.12 backports?

smira:
> @smira is there any chance this could be included in the 1.12 backports?

I don't think we can, as it's technically a breaking change for whomever depends on the previous behavior.

lucasvbeek:
> > @smira is there any chance this could be included in the 1.12 backports?
> 
> I don't think we can, as it's technically a breaking change for whomever depends on the previous behavior.

Makes sense, thanks for clarifying and the fix!

Pythoner6:
Am I understanding correctly that this change still means that if Talos e.g. receives only ipv6 dns servers from say dhcpv6, it will still merge that with the default 1.1.1.1 and 8.8.8.8? I feel like if Talos receices *any* configuration (whether from dhcp or elsewhere) I'd personally have expected it to drop the default addresses.

smira:
> Am I understanding correctly that this change still means that if Talos e.g. receives only ipv6 dns servers from say dhcpv6, it will still merge that with the default 1.1.1.1 and 8.8.8.8? I feel like if Talos receices _any_ configuration (whether from dhcp or elsewhere) I'd personally have expected it to drop the default addresses.

yes, things are complicated on many layers, but we are happy to adjust things if you create a specific issue.

PR #12634, opened 2026-01-21
fix: improve warnings in 802.3ad and add fallback for memory modules

shanduur:
- **fix: add warnings to 802.3ad bond**
  In case some settings are missing that might be impacting the usage of 802.3ad, present a warning to users.

- **fix: fallback to /proc/meminfo for memory modules**
  If SMBIOS does not report memory information, fall back to `/proc/meminfo` and expose a dummy memory module as a best-effort approximation.

### DIFF
--- a/internal/app/machined/pkg/controllers/hardware/system.go
+++ b/internal/app/machined/pkg/controllers/hardware/system.go
@@ -10,8 +10,10 @@ import (
 	"strings"
 
 	"github.com/cosi-project/runtime/pkg/controller"
+	"github.com/cosi-project/runtime/pkg/resource"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/prometheus/procfs"
 	"github.com/siderolabs/gen/optional"
 	"github.com/siderolabs/go-smbios/smbios"
 	"go.uber.org/zap"
@@ -70,6 +72,8 @@ func (ctrl *SystemInfoController) Outputs() []controller.Output {
 	}
 }
 
+const memoryModuleUnknown = "UNKNOWN"
+
 // Run implements controller.Controller interface.
 //
 //nolint:gocyclo
@@ -85,6 +89,8 @@ func (ctrl *SystemInfoController) Run(ctx context.Context, r controller.Runtime,
 			return nil
 		case <-r.EventCh():
 		}
+
+		r.StartTrackingOutputs()
 
 		_, err := safe.ReaderGetByID[*runtime.MetaLoaded](ctx, r, runtime.MetaLoadedID)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -152,5 +157,44 @@ func (ctrl *SystemInfoController) Run(ctx context.Context, r controller.Runtime,
...
+		if len(ctrl.SMBIOS.MemoryDevices) == 0 {
+			if err := ctrl.updateUnknownMemoryModule(ctx, r); err != nil {
+				return err
+			}
+		} else if len(ctrl.SMBIOS.MemoryDevices) > 0 && ctrl.haveUnknownMemoryModule {
+			if err := r.Destroy(ctx, hardware.NewMemoryModuleInfo(memoryModuleUnknown).Metadata()); err != nil {
+				return err
+			}
+
+			ctrl.haveUnknownMemoryModule = false
```
smira: I wonder if we should just do `CleanupOutputs` here to avoid manual cleanup?

smira: something like https://github.com/siderolabs/talos/blob/30da0bc19eb699dabf966cce38ef4477add193d4/internal/app/machined/pkg/controllers/block/zswap_status.go#L79-L116

internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -106,51 +111,100 @@ func (ctrl *SystemInfoController) Run(ctx context.Context, r controller.Runtime,
...
 			return nil
 		}); err != nil {
 			return fmt.Errorf("error updating objects: %w", err)
 		}
+	}
 
-		for _, p := range ctrl.SMBIOS.ProcessorInformation {
-			// replaces `CPU 0` with `CPU-0`
-			id := strings.ReplaceAll(p.SocketDesignation, " ", "-")
+	if err := safe.CleanupOutputs[*hardware.Processor](ctx, r); err != nil {
```
smira: I _believe_ it doesn't work this way... I think you need to call it once per cycle? (but even though I wrote it, I can't say that I remember lol)

smira (approved):

internal/app/machined/pkg/controllers/hardware/system.go
shanduur: Isn't it called once per cycle for different types? Technically it's the same as the example you provided.

smira: to be on the safe side, I would put `r.CleanupOutputs()` to the end of the `Run()` loop which requires a bit more code, but it supports multiple output kinds.

`safe.CleanupOutputs` is a syntax sugar on top of it

internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -106,51 +112,99 @@ func (ctrl *SystemInfoController) Run(ctx context.Context, r controller.Runtime,
...
-		if uuidRewriteRes != nil && uuidRewriteRes.TypedSpec().Value != "" {
-			uuidRewrite = uuidRewriteRes.TypedSpec().Value
+		if err := ctrl.reconcileMemoryModules(ctx, r, logger); err != nil {
+			return err
+		}
 
-			logger.Info("using UUID rewrite", zap.String("uuid", uuidRewrite))
+		if err := r.CleanupOutputs(ctx,
+			resource.NewMetadata(hardware.NamespaceName, hardware.ProcessorType, "", resource.VersionUndefined),
+			resource.NewMetadata(hardware.NamespaceName, hardware.MemoryModuleType, "", resource.VersionUndefined),
```
smira, line 130: looks like there is also `SystemInformationType` output? 

and looks like the code which handles this was removed?

shanduur: Gotcha, added test for it as well, so it's not going missing in the future.

PR #12644, opened 2026-01-22
fix(talosctl): pass --k8s-endpoint flag to rotate-ca kubernetes rotation

kvaps:
## What? (description)

Pass the `--k8s-endpoint` flag value to the Kubernetes client during CA rotation.

The flag was defined in the CLI but never actually used - the value was stored in `rotateCACmdFlags.forceEndpoint` but not passed to `kubernetes.Options`, so the Kubernetes client always used the default endpoint from kubeconfig.

## Why? (reasoning)

When rotating Kubernetes CA on a cluster behind NAT or with a virtual IP that is not directly accessible, users need to specify an alternative endpoint. Without this fix, the `--k8s-endpoint` flag has no effect and the rotation fails trying to connect to the unreachable default endpoint.

## Acceptance


smira:
Forking this PR as I can't push back to org branches.

PR #12651, opened 2026-01-23
fix: set GRUB prefix correctly on arm64

smira:
We always unconditionally create `BIOS` partition, even on arm64, so the prefix should be same on all arches.

We don't use `BIOS` on arm64, but still this would be easier to support in the future.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
@@ -121,6 +121,9 @@ func (c *Config) generateGrubImage(opts options.InstallOptions) error {
 		"all_video",
 	}
 
+	const grubPrefix = "(hd0,gpt3)/grub" // EFI, BIOS, BOOT
+
+	// in amd64 mode only, install GRUB BIOS mode
 	if opts.Arch == "amd64" {
 		grub32Modules := []string{
 			"biosdisk",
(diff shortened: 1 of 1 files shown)


smira:
We should pull this into the Image Factory before going 1.0.0.

internal/app/machined/pkg/runtime/v1alpha1/bootloader/grub/install.go
```diff
@@ -153,18 +156,14 @@ func (c *Config) generateGrubImage(opts options.InstallOptions) error {
...
+	var platform string
 
+	// install GRUB in UEFI mode
 	switch opts.Arch {
 	case "amd64":
 		platform = "x86_64-efi"
-		prefix = "(hd0,gpt3)/grub" // EFI, BIOS, BOOT
 	case "arm64":
 		platform = "arm64-efi"
-		prefix = "(hd0,gpt2)/grub" // EFI, BOOT
```
frezbo, line 167: arm only creates EFI, BOOT, so should be gpt2 right?

smira: no, it always creates BIOS - that's the whole point (IF tests confirm that). yes, BIOS is not used, but it's there

frezbo: yeh, might be I dropped it while testing and re-added back

PR #12652, opened 2026-01-23
feat(talosctl): add --skip-verify flag to skip TLS certificate verification

kvaps:
## Summary

Add global `--skip-verify` flag that disables TLS server certificate verification while preserving client certificate authentication.

This is useful when connecting to nodes via IP addresses not listed in the server certificate's SANs (e.g., NAT, VPN, port-forwarding scenarios).

### Changes

- Add `SkipVerify` field to `global.Args` struct
- Add `WithClientSkipVerify` method that creates TLS connection with `InsecureSkipVerify: true` but preserves client certificate
- Add `--skip-verify` global flag to all commands
- Update commands that support `--insecure` to also check for `--skip-verify`:
  - `apply-config`
  - `get`
  - `meta write/delete`
  - `reset`
  - `upgrade`
  - `version`
  - `wipe disk`

### Difference from --insecure

- `--insecure`: Uses maintenance mode API without client authentication
- `--skip-verify`: Uses normal authenticated API but skips server certificate verification

## Test plan

- [ ] Run `talosctl get members --skip-verify -n <ip>` against a node with mismatched SAN
- [ ] Run `talosctl apply-config --skip-verify -n <ip> -f config.yaml`
- [ ] Verify client authentication still works (commands fail without valid talosconfig)

### DIFF
--- a/cmd/talosctl/cmd/talos/apply-config.go
+++ b/cmd/talosctl/cmd/talos/apply-config.go
@@ -98,6 +98,10 @@ var applyConfigCmd = &cobra.Command{
 				return WithClientMaintenance(applyConfigCmdFlags.certFingerprints, f)
 			}
 
+			if GlobalArgs.SkipVerify {
+				return WithClientSkipVerify(f)
+			}
+
 			return WithClient(f)
 		}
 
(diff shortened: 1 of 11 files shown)


smira:
As much as I see the problem, I don't like the idea of including this into `talosctl`. This leads to a risk of submitting sensitive data to a malicious actor pretending to be Talos API.

PR #12665, opened 2026-01-26
feat: resetting a node now requires a confirmation from the user

kaerbr:
# Pull Request

## What? (description)
As known for example from `apt install` has a conformation step before executing the task: `Do you want to continue? [Y/n]`.  
I think this style of confirmation would also be a good addition to the `talosclt reset` command.

Maybe this could even be expanded to commands like `talosctl wipe`, `talosctl shutdown`, ...?

grafik

## Why? (reasoning)
I nearly reset a node of my k8s cluster because I thought `talosctl reboot` was auto completed, instead `talosctl reset` was auto completed.  
With this experience and because this command is of a destroying nature I think a confirmation by the user would make `talosctl reset` safer. (:  

It seems this happened also to other users: https://github.com/siderolabs/talos/issues/10133#issuecomment-2771817866, https://github.com/siderolabs/talos/issues/10133#issuecomment-2771826284

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/reset.go
+++ b/cmd/talosctl/cmd/talos/reset.go
@@ -74,6 +74,7 @@ var resetCmdFlags struct {
 	graceful           bool
 	reboot             bool
 	insecure           bool
+	noConfirm          bool
 	wipeMode           WipeMode
 	userDisksToWipe    []string
 	systemLabelsToWipe []string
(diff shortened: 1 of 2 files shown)


smira:
The `reset` command already doesn't allow to remove the last controlplane without confirmation, and it doesn't destroy user data (only system disk is "reset"). So I'm not sure it's _that_ destructive. 

Plus this PR is certainly a breaking change, as it will break `reset` from a script, and confirmations shouldn't be used in scripts.

kaerbr:
Okay fair point.  
I did not know that there must be a confirmation for the last control plane.

PR #12678, opened 2026-01-29
refactor: change the way one2many proxying is picked

smira:
Instead of defaulting to one2many, list explicitly one2many supported APIs.

The idea is that any new API will only be "normal" gRPC API, so we can flip the switch, and consider one2many APIs as "legacy".

Extracted from #12392


frezbo (approved):

smira:
Pushed another commit to bump PKGS/TOOLS to address OpenSSL 3.6.0 CVEs.

PR #12679, opened 2026-01-29
feat: implement new image service APIs

smira:
These new APIs only support one2one proxying, so they don't have any hacks, and look as regular gRPC APIs.

Old APIs are deprecated, but still supported.

Implement client-side multiplexing in `talosctl`, provide fallback to old APIs for legacy Talos versions.

New APIs include removing an image, importing an image.

Extracted from #12392


cmd/talosctl/cmd/talos/pull/pull.go
```diff
@@ -0,0 +1,98 @@
...
+// ProgressWriter writes pull progress updates to a reporter.
+type ProgressWriter struct {
+	// ongoingPulls keeps track of ongoing pull jobs per node.
+	ongoingPulls map[string]pullJobs
+}
+
+// UpdateJob updates the progress of a pull job for a given node and layer ID.
+//
+// It is supposed to be called whenever there is a progress update for a layer pull.
+func (w *ProgressWriter) UpdateJob(node, layerID string, progress *machine.ImageServicePullLayerProgress) {
```
smira, line 28: TODO: don't do this if the stderr is not a tty

internal/app/images/images.go
```diff
@@ -0,0 +1,283 @@
...
+func (svc *Service) Remove(ctx context.Context, req *machine.ImageServiceRemoveRequest) (*emptypb.Empty, error) {
+	ctx, client, err := containerdInstanceHelper(ctx, req.GetContainerd())
+	if err != nil {
+		return nil, err
+	}
+
+	//nolint:errcheck
+	defer client.Close()
+
+	err = client.ImageService().Delete(ctx, req.ImageRef)
```
smira:
```suggestion
	err = client.ImageService().Delete(ctx, req.ImageRef, image.SynchronousDelete)
```

smira:
Demo for image pull with progess: 

[asciicast](https://asciinema.org/a/20sIISeN3czLtEsr)

PR #12696, opened 2026-02-01
fix: improve OpenStack bare metal network configuration reliability

maxpain:
## Summary

Fixes network configuration issues on OpenStack bare metal (Ironic) servers with bonded interfaces.

Changes:
- Add `WaitForDevicesReady` call before parsing network configuration to ensure network interfaces are available
- Add fallback to `HardwareAddr` when `PermanentAddr` is empty (ethtool may not have populated it yet)
- Add retry loop with exponential backoff when MAC address matching fails
- Remove hardcoded bond parameters (`UpDelay: 200`, `DownDelay: 200`, `LACPRate: fast`) to use kernel defaults matching Ubuntu behavior
- Explicitly set bond interface MAC address from OpenStack metadata to ensure the switch sees the correct MAC

The bond MAC fix addresses a critical issue where Linux assigns the bond the MAC of the first enslaved interface, which may not match the MAC expected by the switch. OpenStack metadata provides the correct MAC via `ethernet_mac_address` field, and we now apply it explicitly.

The retry loop, HardwareAddr fallback and `WaitForDevicesReady` match the pattern used in the NoCloud platform.

## Testing

These changes have been verified in production on our bare metal servers (GCore Bare Metal Cloud) and everything works correctly after the fix.

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
@@ -12,9 +12,12 @@ import (
 	stderrors "errors"
 	"fmt"
 	"log"
+	"net"
 	"net/netip"
 	"strings"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/cosi-project/runtime/pkg/safe"
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/go-procfs/procfs"
@@ -48,8 +51,9 @@ func (o *OpenStack) ParseMetadata(
 	extIPs []netip.Addr,
 	metadata *MetadataConfig,
 	st state.State,
-) (*runtime.PlatformNetworkConfig, error) {
+) (*runtime.PlatformNetworkConfig, bool, error) {
 	networkConfig := &runtime.PlatformNetworkConfig{}
+	needsReconcile := false
 
 	if metadata.Hostname != "" {
 		hostnameSpec := network.HostnameSpecSpec{
@@ -57,7 +61,7 @@ func (o *OpenStack) ParseMetadata(
 		}
 
 		if err := hostnameSpec.ParseFQDN(metadata.Hostname); err != nil {
-			return nil, err
+			return nil, false, err
 		}
 
 		networkConfig.Hostnames = append(networkConfig.Hostnames, hostnameSpec)
@@ -72,7 +76,7 @@ func (o *OpenStack) ParseMetadata(
 			if ip, err := netip.ParseAddr(netsvc.Address); err == nil {
 				dnsIPs = append(dnsIPs, ip)
 			} else {
-				return nil, fmt.Errorf("failed to parse dns service ip: %w", err)
+				return nil, false, fmt.Errorf("failed to parse dns service ip: %w", err)
 			}
 		}
 	}
@@ -86,7 +90,7 @@ func (o *OpenStack) ParseMetadata(
 
 	hostInterfaces, err := safe.StateListAll[*network.LinkStatus](ctx, st)
 	if err != nil {
-		return nil, fmt.Errorf("error listing host interfaces: %w", err)
+		return nil, false, fmt.Errorf("error listing host interfaces: %w", err)
 	}
 
 	ifaces := make(map[string]string)
@@ -103,12 +107,12 @@ func (o *OpenStack) ParseMetadata(
 
 		mode, err := nethelpers.BondModeByName(netLink.BondMode)
 		if err != nil {
-			return nil, fmt.Errorf("invalid bond_mode: %w", err)
+			return nil, false, fmt.Errorf("invalid bond_mode: %w", err)
 		}
 
 		hashPolicy, err := nethelpers.BondXmitHashPolicyByName(netLink.BondHashPolicy)
 		if err != nil {
-			return nil, fmt.Errorf("invalid bond_xmit_hash_policy: %w", err)
+			return nil, false, fmt.Errorf("invalid bond_xmit_hash_policy: %w", err)
 		}
 
 		bondName := fmt.Sprintf("bond%d", bondIndex)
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
```diff
@@ -427,6 +427,11 @@ func (o *OpenStack) KernelArgs(string, quirks.Quirks) procfs.Parameters {
 
 // NetworkConfiguration implements the runtime.Platform interface.
 func (o *OpenStack) NetworkConfiguration(ctx context.Context, st state.State, ch chan<- *runtime.PlatformNetworkConfig) error {
+	// wait for devices to be ready before proceeding, otherwise we might not find network interfaces by MAC
+	if err := netutils.WaitForDevicesReady(ctx, st); err != nil {
```
smira, line 455: I would put it to `ParseMetadata` call (close to where it matters).

maxpain: In NoCloud it's placed in the same spot — at the top of NetworkConfiguration ([nocloud.go:131](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/nocloud.go#L131)). I kept it consistent with that pattern. Happy to move it closer to ParseMetadata if you prefer, either here or in a follow-up PR.

maxpain:
@smira this PR is ready to merge

internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
```diff
@@ -126,12 +130,18 @@ func (o *OpenStack) ParseMetadata(
...
 		}
 
+		if netLink.Mac != "" {
+			mac, err := net.ParseMAC(netLink.Mac)
+			if err != nil {
+				return nil, false, fmt.Errorf("invalid bond MAC address %q: %w", netLink.Mac, err)
+			}
+
+			bondLink.HardwareAddress = nethelpers.HardwareAddr(mac)
+		}
```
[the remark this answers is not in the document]
maxpain: Will add tests for the new behaviors in a follow-up.

internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
```diff
@@ -126,12 +130,18 @@ func (o *OpenStack) ParseMetadata(
 				Mode:       mode,
 				MIIMon:     netLink.BondMIIMon,
 				HashPolicy: hashPolicy,
```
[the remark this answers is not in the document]
maxpain: OpenStack `network_data.json` does not provide `bond_lacp_rate`, `bond_updelay`, or `bond_downdelay` fields — only `bond_miimon` and `bond_xmit_hash_policy` are available in the metadata (see the [Nova spec](https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html)). The suggested `netLink.BondUpDelay` / `netLink.BondDownDelay` / `netLink.BondLACPRate` fields don't exist in the `NetworkConfig` struct. Removing the hardcoded values is intentional — kernel defaults should be used instead. Will update the test fixture accordingly.

internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
```diff
@@ -179,7 +189,12 @@ func (o *OpenStack) ParseMetadata(
 			linkName := ""
 
 			for hostInterface := range hostInterfaces.All() {
-				if strings.EqualFold(hostInterface.TypedSpec().PermanentAddr.String(), netLink.Mac) {
+				macAddress := hostInterface.TypedSpec().PermanentAddr.String()
+				if macAddress == "" {
+					macAddress = hostInterface.TypedSpec().HardwareAddr.String()
+				}
+
+				if strings.EqualFold(macAddress, netLink.Mac) {
```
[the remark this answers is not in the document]
maxpain: Good catch, will add a guard for empty `netLink.Mac`.

internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/testdata/expected.yaml
```diff
@@ -36,17 +36,16 @@ links:
...
         mode: 802.3ad
         xmitHashPolicy: layer2+3
-        lacpRate: fast
+        lacpRate: slow
         arpValidate: none
         arpAllTargets: any
         primaryReselect: always
         failOverMac: none
         miimon: 100
-        updelay: 200
```
smira, line 48: I think this definitely is a bug, as without up/down delay the bond is usually totally useless in 802.3ad

maxpain: The previous hardcoded values (`UpDelay: 200`, `DownDelay: 200`, `LACPRate: fast`) were not based on OpenStack metadata — these fields (`bond_updelay`, `bond_downdelay`, `bond_lacp_rate`) simply don't exist in the OpenStack `network_data.json` schema.

The kernel defaults for bonding are `updelay=0`, `downdelay=0`, `lacp_rate=slow` ([kernel docs](https://docs.kernel.org/networking/bonding.html)), and this is exactly how cloud-init handles the same metadata on Ubuntu — it only passes through what OpenStack provides, without adding its own defaults. For reference, here's a real-world `network_data.json` from GCore Bare Metal Cloud (OpenStack Ironic) for an 802.3ad bond — no `lacp_rate`/`updelay`/`downdelay` fields:

```json
{
  "type": "bond",
  "bond_mode": "802.3ad",
  "bond_miimon": "100",
  "bond_xmit_hash_policy": "layer2+3"
}
```

This works fine on Ubuntu with kernel defaults, so removing the hardcoded values aligns our behavior with cloud-init and avoids overriding what the kernel/switch negotiate naturally.

smira: ok, let's see... 🤞

internal/app/machined/pkg/runtime/v1alpha1/platform/openstack/openstack.go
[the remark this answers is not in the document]
smira: I will probably revert this change, LACP rate might be irrelevant, but it would break existing deployments for sure. I don't think kernel has any defaults for up/down delay.

PR #12699, opened 2026-02-02
feat: add a helper module to generate standard patches

smira:
Allow some patches to be generated correctly according to the version contract of the machine configuration.


smira:
See e.g. https://github.com/siderolabs/omni-infra-provider-proxmox/pull/44

smira:
> I guess this is for omni and infra providers?

yes, my goal is to replace anything in providers which tries to use this

Unix4ever (approved):

PR #12700, opened 2026-02-02
fix: implement merger for PercentageSize

shanduur:
Fixing issue when PercentageSize is used and instead of calling Merge it was trying to merge individual unexported fields.

Blocks #12585


shanduur:
@frezbo I will test it in the E2E of #12585

PR #12707, opened 2026-02-02
fix: stop existing QEMU launcher before starting a new one

utkuozdemir:
Found the issue while working on the bare-metal infra provider integration tests: the pid of a qemu machine would get overwritten by a new launcher process before the old one had fully terminated, orphaning the old launcher and causing both to fight over the same disk image - resulting in lock conflicts and the HTTP power management endpoint becoming unreachable.

To fix it, when creating a new node, ensure any existing launcher process is stopped before starting a new one.

This prevents the race condition where the new launcher overwrites the PID file while the old launcher is still running, causing:
- Orphaned launcher processes (old PID is lost)
- Disk lock conflicts when both launchers try to run QEMU
- HTTP server (for power control) port becoming unreachable after the new launcher crashes

### DIFF
--- a/pkg/provision/providers/qemu/node.go
+++ b/pkg/provision/providers/qemu/node.go
@@ -30,12 +30,19 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"github.com/siderolabs/talos/pkg/machinery/kernel"
 	"github.com/siderolabs/talos/pkg/provision"
+	"github.com/siderolabs/talos/pkg/provision/providers/vm"
 )
 
 //nolint:gocyclo,cyclop
 func (p *provisioner) createNode(ctx context.Context, state *provision.State, clusterReq provision.ClusterRequest, nodeReq provision.NodeRequest, opts *provision.Options) (provision.NodeInfo, error) {
 	arch := Arch(opts.TargetArch)
 	pidPath := state.GetRelativePath(fmt.Sprintf("%s.pid", nodeReq.Name))
+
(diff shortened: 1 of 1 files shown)


pkg/provision/providers/qemu/node.go
```diff
@@ -30,13 +30,20 @@ import (
...
 )
 
 //nolint:gocyclo,cyclop
 func (p *provisioner) createNode(ctx context.Context, state *provision.State, clusterReq provision.ClusterRequest, nodeReq provision.NodeRequest, opts *provision.Options) (provision.NodeInfo, error) {
 	arch := Arch(opts.TargetArch)
 	pidPath := state.GetRelativePath(fmt.Sprintf("%s.pid", nodeReq.Name))
 
+	// Ensure any existing launcher for this node is stopped before starting a new one.
+	// This prevents orphaned launcher processes when the PID file gets overwritten.
+	if err := vm.StopProcessByPidfile(pidPath); err != nil {
```
smira, line 43: I wonder how can it get here when the previous launcher is still running?

I don't think it's possible, as it should fail on state directory already existing?

frezbo: unless state directory creation failed, but I don;t think we would ever get into this state? 🤔

smira: yep, the pid is written just once on "create cluster", and never touched again (if the VM restarts, launcher stays alive)

utkuozdemir: yes, you are right, something else was going on, I'll look into it a bit further. the fix is certainly not this.

PR #12711, opened 2026-02-03
fix: add missing image for integration tests

shanduur:
Added missing image for integration tests.


cmd/talosctl/cmd/talos/image.go
```diff
@@ -558,6 +558,7 @@ var imageIntegrationCmd = &cobra.Command{
 			"docker.io/library/alpine:latest",
 			"ghcr.io/siderolabs/talosctl:latest",
 			"registry.k8s.io/kube-apiserver:v1.27.0",
+			"registry.k8s.io/kube-apiserver:v1.27.1",
```
frezbo, line 561: can't we just use 1.27.0 for all then ? or even 1.27.1 everywhere

PR #12734, opened 2026-02-06
chore: update quic-go import

rsmitty:
This PR bumps the quick-go import to protect from CVE-2025-64702 (found via dependabot alerts).


smira:
I think this is all false positives

smira:
I'll add a bigger bump of tools which will resolve this, I just need to handle one edge case issue.

PR #12743, opened 2026-02-06
fix: default value for BondUseCarrier, resolve #12742

### LINKED ISSUE #12742: Flapping bond in v1.12
## Bug Report
Talos v1.12 with networking set via meta config results in flapping bond.   
Forming bond via kernel command line parameters works normally.

### What is broken

The default value for `use_carrier` bond option changed with talos v1.12 to `false`.   
According to [linux bonding docs](https://docs.kernel.org/networking/bonding.html) the only valid value is `1` / `true`.

Talos v1.11
https://github.com/siderolabs/talos/blob/6dd14300ecad2ad4dd1c63f5550f67cf8973195a/pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go#L1056-L1058

Talos v1.12
https://github.com/siderolabs/talos/blob/54e5b438d8dcf6395e6424808d1155d02abf3bc0/pkg/machinery/config/types/network/bond.go#L409-L411

otahirs:
## Fix bond flapping 
Fix regression introduced in talos v1.12 by changing the default value of `use_carrier` for bond configuration.

fix #12742

### DIFF
--- a/pkg/machinery/config/types/network/bond.go
+++ b/pkg/machinery/config/types/network/bond.go
@@ -429,7 +429,7 @@ func (s *BondConfigV1Alpha1) DownDelay() optional.Optional[uint32] {
 // UseCarrier implements NetworkBondConfig interface.
 func (s *BondConfigV1Alpha1) UseCarrier() optional.Optional[bool] {
 	if s.BondUseCarrier == nil {
-		return optional.None[bool]()
+		return optional.Some(true)
 	}
 
 	return optional.Some(*s.BondUseCarrier)


smira:
I don't think this PR matches the actual intention in the issue, as it has nothing to do with META config.

The use_carrier is already locked to true: https://github.com/siderolabs/talos/blob/9690dbad02cfc8682d697679b655e753039c5254/internal/app/machined/pkg/adapters/network/bond_master_spec.go#L38-L39

PR #12762, opened 2026-02-09
feat(nocloud): make user-data optional, support network-config only mode

dyudin0821:
# Pull Request

## What? (description)
This PR makes the user-data configuration file optional for the NoCloud platform implementation, allowing network configuration to be provided independently via network-config.

Specifically, the change enables scenarios where only initial network configuration is required, without forcing users to provide a full Talos machine configuration.

Key changes:

- Updated configFromNetwork() to handle missing user-data gracefully
- Updated configFromCD() to avoid returning ErrNoConfigSource when user-data is absent
- Added nil/empty checks in Configuration() with appropriate logging
- Fixed handling of empty network-config files by checking both nil and content length
- Added a test case for network-only configuration
- Added a test validating behavior with an empty network-config

## Why? (reasoning)
This creates a limitation for on-prem and virtualization use cases where:
- initial network bootstrapping is required first,
- machine configuration (certificates, tokens, cluster config) should be injected later
- or configuration is delivered via a different secure channel.

Requiring user-data in such cases forces users to provide a minimally valid Talos machine configuration upfront, which may involve generating CA material and tokens earlier than desired.

By allowing network-config to be consumed independently:

- initial network provisioning becomes cleaner and safer,
- NoCloud can be used purely as a network bootstrap mechanism,
- existing behavior remains unchanged for setups that already provide user-data.
- The change is fully backward-compatible and only relaxes validation where appropriate.

## Acceptance


internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
```diff
@@ -172,9 +172,13 @@ func (n *Nocloud) configFromNetwork(ctx context.Context, metaBaseURL string, r s
 
 	log.Printf("fetching machine config from: %q", metaBaseURL+configUserDataPath)
 
-	machineConfig, err = download.Download(ctx, metaBaseURL+configUserDataPath,
-		download.WithErrorOnNotFound(errors.ErrNoConfigSource),
-		download.WithErrorOnEmptyResponse(errors.ErrNoConfigSource))
+	machineConfig, err = download.Download(ctx, metaBaseURL+configUserDataPath)
+	if err != nil {
+		log.Printf("user-data not found (optional): %s", err)
```
smira, line 177: this is certainly wrong, and it's not clear why we start ignoring any error on downloading userdata

smira: we have explicit error `errors.ErrNoConfigSource` for the case when it's not found

smira:
I'm going to close this unless it's proven that there is a problem that needs to be fixed.

PR #12765, opened 2026-02-10
feat: update doc links to docs.siderolabs.com

Slessi:
Update various documentation links that are pointing to `www.talos.dev` to the new docs page at `docs.siderolabs.com`.

Related to siderolabs/omni#1956

Factory/Omni will be updated in a follow-up PR.

### DIFF
--- a/pkg/machinery/platforms/sbcs.go
+++ b/pkg/machinery/platforms/sbcs.go
@@ -47,7 +47,7 @@ func SBCs() []SBC {
 			OverlayImage: "siderolabs/sbc-raspberrypi",
 
 			Label:         "Raspberry Pi Series",
-			Documentation: "/talos-guides/install/single-board-computers/rpi_generic/",
+			Documentation: "/platform-specific-installations/single-board-computers/rpi_generic",
 		},
 		{
 			Name: "revpi_generic",
@@ -66,7 +66,7 @@ func SBCs() []SBC {
 			OverlayImage: "siderolabs/sbc-allwinner",
 
 			Label:         "Banana Pi M64",
-			Documentation: "/talos-guides/install/single-board-computers/bananapi_m64/",
+			Documentation: "/platform-specific-installations/single-board-computers/bananapi_m64",
 		},
 		{
 			Name: "nanopi_r4s",
@@ -75,7 +75,7 @@ func SBCs() []SBC {
 			OverlayImage: "siderolabs/sbc-rockchip",
 
 			Label:         "Friendlyelec Nano PI R4S",
-			Documentation: "/talos-guides/install/single-board-computers/nanopi_r4s/",
+			Documentation: "/platform-specific-installations/single-board-computers/nanopi_r4s",
 
 			MinVersion: semver.MustParse("1.3.0"),
 		},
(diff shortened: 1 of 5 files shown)


hack/release.toml
```diff
@@ -36,7 +36,7 @@ Talos is built with Go 1.25.7.
         title = "External Volumes"
         description = """\
 Talos now supports virtiofs-based external volumes via the new
-[ExternalVolumeConfig](https://www.talos.dev/v1.13/reference/configuration/block/externalvolumeconfig/)
+[ExternalVolumeConfig](https://docs.siderolabs.com/talos/v1.13/reference/configuration/block/externalvolumeconfig)
```
Slessi: Neither the old nor new page here exists, though

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ To remove an environment variable, remove it from the `EnvironmentConfig` docume
 ### External Volumes
 
 Talos now supports virtiofs-based external volumes via the new
-[ExternalVolumeConfig](https://www.talos.dev/v1.13/reference/configuration/block/externalvolumeconfig/)
+[ExternalVolumeConfig](https://docs.siderolabs.com/talos/v1.13/reference/configuration/block/externalvolumeconfig)
```
Slessi: I left most of the changelog untouched, with the exception of this 1.13.0 change which I guess is ok since it isn't released yet.

But FYI there are many links from very old releases like https://www.talos.dev/v1.3/talos-guides/install/single-board-computers/rpi_generic/ which do not work. Anything with a talos version older than 1.6 does not work.

smira: let's drop this change please, as the file is auto-generated, and it's more like an archive. I don't think we should ever change it.

pkg/machinery/platforms/sbcs.go
```diff
@@ -216,7 +214,7 @@ func SBCs() []SBC {
 			OverlayImage: "siderolabs/sbc-rockchip",
 
 			Label:         "Turing RK1",
-			Documentation: "/talos-guides/install/single-board-computers/turing_rk1/",
+			Documentation: "/platform-specific-installations/single-board-computers/turing_rk1",
```
smira, line 217: This (I guess) needs a matching PR to the Image Factory to update the links (this is where it's getting consumed anyways).

Slessi: It does which I'm gonna do immediately next, I have to import a new version of Talos first right?

pkg/cluster/hydrophone/product.go
```diff
@@ -25,7 +25,7 @@ var talos = product{
 	Version:          version.Tag,
 	WebsiteURL:       "https://www.siderolabs.com/",
 	RepoURL:          "https://github.com/siderolabs/talos",
-	DocumentationURL: "https://www.talos.dev",
+	DocumentationURL: "https://docs.siderolabs.com/talos",
```
smira: let's keep the previous URL, I think `www.talos.dev` is a good entrypoint (it gets published to CNCF website), and docs is a one click from it.

hack/release.toml
smira: yeah, documentation sync process doesn't exist

pkg/machinery/platforms/sbcs.go
smira: yep, just bump in `go.mod` for both machinery & Talos itself, thanks!

PR #12772, opened 2026-02-11
fix: fix talosctl debug in enforcing mode

dsseng:
Also allow the system containerd to execute igzip, which is essential
for pulling images


internal/pkg/selinux/policy/selinux/services/system-containerd.cil
```diff
@@ -86,3 +86,9 @@
 ; Manage keys
 (allow containerd_p self (key (view read write search link setattr create)))
 
+; Containerd uses igzip on amd64
```
smira: and pigz on arm64, I think the rule covers them both, but comments might be clariied

internal/pkg/selinux/policy/selinux/services/system-containers.cil
```diff
@@ -54,3 +54,35 @@
 ; Talk to machined
 (allow installer_t machine_socket_t (sock_file (write)))
 (allow installer_t init_t (unix_stream_socket (connectto)))
+
+; talosctl debug container, automatically transition to it
+(type debug_t)
```
smira, line 59: for my education: how does it know that this is debug container?

I mean how does it know it has to transition only for debug container, but not for e.g. `apid` container?

internal/pkg/selinux/policy/selinux/services/system-containerd.cil
dsseng: Both should be covered. Thanks, will fix the comment

internal/pkg/selinux/policy/selinux/services/system-containers.cil
dsseng: lower, there's a typetransition directive, which makes each process spawned by sys_containerd_t launching a binary a debug_t. other containers have their labels explicitly set via container creation options

smira: does it also apply if sys_containerd_t runs `igzip` as an example? should we do an explicit label here?

dsseng: Ah, no, the second operand here is `system_var_t`: `(typetransition sys_containerd_t system_var_t process debug_t)`

So, when `sys_containerd_t` creates a `process` from `system_var_t`, it transitions to `debug_t` unless specified otherwise.

PR #12775, opened 2026-02-12
fix: switch to better Myers algorithm implementation

smira:
See https://github.com/siderolabs/omni/pull/2311 for the rationale.


smira:
Might need some time to figure out correct settings, the PR itself is still good cleanup

smira:
I will do one more to go-kubernetes to finish the puzzle, I hope I can use machinery's method there.

Issue #11695, opened 2025-08-28
Theme: Talos 1.13 Kubernetes Configuration improvements


smira:
## Why?

Our existing configuration in `MachineConfig` is not flexible enough for some usecases, which doesn't allow configuring Kubernetes as flexible as we need.

Having Kubernetes configuration embedded into MachineConfig makes it hard to see all Kubernetes configuration, and look into each component.

## Scope

Kubernetes configuration: core secrets (CAs, etc.), component-specific configuration: kube-apiserver, kube-scheduler, kube-controller-manager, kubelet, kube-proxy.

Kubernetes-related configuration: CNI (Flannel), CoreDNS, bootstrap manifests, etc.

## Goal

Deprecate all Kubernetes in MachineConfig, decompose into multi-doc, and improve each document as we go.

## Current List

A set of features that might require going multi-docs for Kubernetes configuration first.

clementnuss:
and for reference, an example implementation for a k8s config document for etcd encryption over here: #11641

PR #12785, opened 2026-02-12
feat: add EncryptionConfiguration document in a new config/k8s package

clementnuss:
# Pull Request

## What? (description)

offers a config document based solution for the user to provide a custom etcd encryption configuration.
when specified, the custom encryption config is applied as-is band it bypasses the talos etcd encryption config

the config document apiversion `apiserver.config.k8s.io/v1`, and the kind `EncryptionConfiguration` are the same as the upstream config: https://kubernetes.io/docs/reference/config-api/apiserver-config.v1/#apiserver-config-k8s-io-v1-EncryptionConfiguration
, making it straightforward for users to work customize the encryption configuration.

## Why? (reasoning)

- making it possible to specify a custom apiserver encryption config: #10899
- more generally, provide an example for config-doc based kubernetes configuration #11695 
- close #11641 

## Implementation details

created a new k8s package to store multiple config documents related to the apiserver configuration.

a basic validation of the provided `EncryptionConfiguration` is done (a `json.Unmarshal(&cfg, ...)`, but not more. this can be discussed, but I think it is the responsibility of the user to provide a valid configuration document.

I also first wanted to directly include the upstream types instead of parsing to a `Fields: map[string]any`, but[ upstream k8s types use json tags](https://github.com/kubernetes/apiserver/blob/bef6898871aac34e617d3e8e489bf2a7ebc49903/pkg/apis/apiserver/v1/types_encryption.go#L70-L87), and I don't think we could parse those with the current yaml parsing library

## Acceptance


pkg/machinery/config/types/k8s/encryption_configuration.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	cp := make(map[string]any, len(m))
+
+	for k, v := range m {
+		cp[k] = deepCopyValue(v)
+	}
+
+	return cp
+}
```
shanduur: https://pkg.go.dev/maps#Clone

pkg/machinery/config/types/k8s/encryption_configuration.go
```diff
@@ -0,0 +1,140 @@
...
+//
+// EncryptionConfiguration is replaced entirely rather than recursively merged,
+// since the inline map structure doesn't support meaningful partial merges.
+func (d *EncryptionConfigurationDoc) Merge(other any) error {
+	otherDoc, ok := other.(EncryptionConfigurationDoc)
+	if !ok {
+		return fmt.Errorf("unexpected type for merge: %T", other)
+	}
+
+	d.Fields = deepCopyMap(otherDoc.Fields)
```
shanduur:
```suggestion
	d.Fields = maps.Clone(otherDoc.Fields)
```

pkg/machinery/config/types/k8s/encryption_configuration.go
clementnuss: thanks, that's much better 👍🏼

clementnuss: done. there was another use of `deepCopyMap`. I changed that, removed the function and fixed some linter errors.

PR #12790, opened 2026-02-13
fix: memory overuse in imager VFAT

smira:
Use proper `io.Copy` instead of `io.ReadFile`. If the source UKI is huge, it might momentarily consume lots of RAM.


smira:
relevant profile from Image Factory:

image

oguzkilcan (approved):

Issue #12781, opened 2026-02-12
talosctl cluster create qemu not respecting disk options


tracyde:
## Bug Report
When creating a new cluster using `talosctl cluster create qemu --presets iso-secureboot --disks "virtio:50GiB"` disks are not created as expected. Expect controlplane to have a single 10G disk and workers to have a single 50G disk.

What actually happens is controlplane gets single 10G disk but workers each get 3 disks 10G, 6G, and 50G.
 
### Description
I am trying to create a local development environment for kubernetes for software testing. When trying to add multiple pieces of software I am running into issues where I run out of disk space when installing helm charts trying to download images.

I would like to increase the size of the installation disk. I have also tried patching the machineconfig and changing the install disk to "/dev/vdc" but that causes the installation to fail. I have also tried creating a patch for the machineconfig that adds `provisioning` to the `EPHEMERAL` VolumeConfig, but that did not work either.

```
apiVersion: v1alpha1
kind: VolumeConfig
name: EPHEMERAL # Name of the volume.
provisioning:
  diskSelector:
    match: disk.dev_path == '/dev/vdc' && !system_disk
```

### Environment

- Talos version: 1.12.3
- Kubernetes version: any
- Platform: Linux x86_64

Orzelius:
Thank you for the report! I've confirmed the undesired behavior and am working on a fix.

Orzelius:
The issue has been fixed via #12793.

@tracyde the desired behavior for the disks flag is slightly different from what you described in the original issue. The CLI help goes as follows
```
list of disks to create in format "<driver1>:<size1>" (disks after the first one are added only to worker machines)
```
so passing `--disks "virtio:50GiB"` will results in in both the workers and controlplanes having a single 50GiB disk. Any disks added after the first one would be added only to the workers. Hope this clears things up.

PR #12793, opened 2026-02-15
fix: disks flag parsing and handling in create qemu command

Orzelius:
The disks flag Set method was appending new disk requests to the existing ones, which caused duplicate disk entries when custom values for the disks flag were set.

fixes https://github.com/siderolabs/talos/issues/12781


anatoly314:
@Orzelius , any expectation when it will be released?

PR #12805, opened 2026-02-18
fix: add owning inventory annotation to talos manifests

Orzelius:
although the manifests were deployed via ssa the inventory annotation was never added previously

https://github.com/siderolabs/go-kubernetes/pull/43#discussion_r2792962112

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/manifest_apply.go
+++ b/internal/app/machined/pkg/controllers/k8s/manifest_apply.go
@@ -311,6 +311,21 @@ func (ctrl *ManifestApplyController) apply(
 			return fmt.Errorf("error checking resource existence: %w", err)
 		}
 
+		// Set inventory annotation.
+		annotations := obj.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+
+		inventoryAnnotation, inventoryAnnotationSet := annotations["config.k8s.io/owning-inventory"]
+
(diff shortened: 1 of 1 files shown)


internal/app/machined/pkg/controllers/k8s/manifest_apply.go
```diff
@@ -311,6 +311,15 @@ func (ctrl *ManifestApplyController) apply(
 			return fmt.Errorf("error checking resource existence: %w", err)
 		}
 
+		// Set inventory annotation.
+		annotations := obj.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+
+		annotations["config.k8s.io/owning-inventory"] = constants.KubernetesBootstrapManifestsInventoryName
```
frezbo, line 326: if there a case we need to check if annotation already exists or user template already has that?

smira: I guess that would be a user's bug (we can fail on such annotations?)

smira (approved):

internal/app/machined/pkg/controllers/k8s/manifest_apply.go
Orzelius: Yep, that makes sense. I also already have the check in the go-kubernetes library.
Where should I put the check if I want the users to see the error when applying the machine config or is that not possible?

PR #12817, opened 2026-02-19
fix: remove stale endpoints

### LINKED ISSUE #12802: bug: talos-ipv4 EndpointSlice accumulates stale endpoints
# Bug: `talos-ipv4` EndpointSlice accumulates stale endpoints and never prunes them

Analysis by Claude Opus 4.6

## Summary

The `kubeaccess.EndpointController` manages a `talos-ipv4` (and optionally `talos-ipv6`) `EndpointSlice` in the `default` namespace to expose the Talos API (port 50000) to workloads inside Kubernetes. When reconciling the EndpointSlice, the controller **appends** the current control plane IPs to a deep copy of the existing EndpointSlice and deduplicates — but it **never removes** endpoints whose IPs are no longer present in the current endpoint list. Over time, this causes the EndpointSlice to accumulate stale entries from former control plane nodes or changed IPs.

## Affected versions

- Confirmed on **v1.11.6**
- Bug is still present on `main` (as of commit `b8f824525`)

## Steps to reproduce

1. Create a Talos cluster with 3 control plane nodes.
2. Observe `talos-ipv4` EndpointSlice has 3 endpoints (correct).
3. Replace or rescale control plane nodes such that IP addresses change (e.g. node replacement, IP reassignment, scaling down then up).
4. Observe `talos-ipv4` EndpointSlice now has more endpoints than there are control plane nodes — the old IPs persist.

## Root cause

In [`ensureTalosEndpointSlicesTyped`](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L265-L351):

1. The existing EndpointSlice is fetched and [**deep copied**](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L298), preserving all existing endpoints:

   ```go
   newEndpointSlice = oldEndpointSlice.DeepCopy()
   ```

2. The current control plane IPs are [**appended**](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L309-L321) to the copy (which already contains the old endpoints):

   ```go
   for _, addr := range endpointAddrs {
       newEndpointSlice.Endpoints = append(
           newEndpointSlice.Endpoints,
           discoveryv1.Endpoint{...},
       )
   }
   ```

3. The list is [**deduplicated**](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L323-L325):

   ```go
   newEndpointSlice.Endpoints = xslices.Deduplicate(newEndpointSlice.Endpoints, func(e discoveryv1.Endpoint) string {
       return e.Addresses[0]
   })
   ```

Deduplication only collapses entries with the **same** address. It does **not** remove entries whose addresses are no longer in `endpointAddrs`. As a result, stale endpoints from former control plane nodes accumulate indefinitely.

### Note on `cleanupTalosEndpoints`

The [`cleanupTalosEndpoints`](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L354-L375) method only cleans up the deprecated legacy `Endpoints` resource (via `CoreV1().Endpoints()`), **not** the `EndpointSlice` resource. It does not help with this issue.

## Suggested fix

Clear the endpoints list before rebuilding it. After the deep copy on [line 298](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L298), reset the `Endpoints` field so that only current IPs are included:

```go
newEndpointSlice = oldEndpointSlice.DeepCopy()
newEndpointSlice.Endpoints = nil  // ← clear stale endpoints before rebuilding
```

This preserves the ObjectMeta (including `resourceVersion`, needed for the update API call) while ensuring the endpoint list is rebuilt from scratch on every reconciliation.

## Workaround

Delete the EndpointSlice to force the controller to recreate it from scratch (the [`IsNotFound` path](https://github.com/siderolabs/talos/blob/v1.11.6/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go#L283-L296) creates a fresh slice without the deep-copy problem):

```bash
kubectl delete endpointslice talos-ipv4 -n default
```

The controller will recreate it with only the current control plane IPs. This will need to be repeated if control plane membership changes again.

## Impact

- The `talos` Service routes traffic to nodes that are no longer control plane members (or no longer exist), causing connection failures or timeouts for Talos API consumers inside the cluster.
- Affects any cluster where control plane node IPs have changed over time.

shanduur:
In the KubeAccess controller, we never cleaned up stale endpoints.
This resulted in stale IPs being preserved indefinitely in the resulting object.

Fixes #12802

### DIFF
--- a/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
+++ b/internal/app/machined/pkg/controllers/kubeaccess/endpoint.go
@@ -298,31 +298,7 @@ func (ctrl *EndpointController) ensureTalosEndpointSlicesTyped(
 			newEndpointSlice = oldEndpointSlice.DeepCopy()
 		}
 
-		newEndpointSlice.Ports = []discoveryv1.EndpointPort{
-			{
-				Name:     ptr.To("apid"),
-				Port:     ptr.To[int32](constants.ApidPort),
-				Protocol: ptr.To(corev1.ProtocolTCP),
-			},
-		}
-
-		for _, addr := range endpointAddrs.Addresses {
-			newEndpointSlice.Endpoints = append(
-				newEndpointSlice.Endpoints,
-				discoveryv1.Endpoint{
-					Addresses: []string{addr.String()},
-					Conditions: discoveryv1.EndpointConditions{
-						Ready:       ptr.To(true),
-						Serving:     ptr.To(true),
-						Terminating: ptr.To(false),
-					},
-				},
-			)
-		}
-
-		newEndpointSlice.Endpoints = xslices.Deduplicate(newEndpointSlice.Endpoints, func(e discoveryv1.Endpoint) string {
-			return e.Addresses[0]
-		})
+		PopulateEndpointSlice(newEndpointSlice, endpointAddrs)
 
 		if oldEndpointSlice != nil &&
 			(reflect.DeepEqual(oldEndpointSlice.Endpoints, newEndpointSlice.Endpoints) &&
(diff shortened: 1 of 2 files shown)


shanduur:
@frezbo refactored and included tests.

frezbo (approved):

PR #12829, opened 2026-02-20
feat: bring in new SSA manifest sync

smira:
Can be merged after https://github.com/siderolabs/go-kubernetes/pull/50


Orzelius:
@smira I've pushed my commit separate from yours so it's easy to see what has chanced since you touched it. Will squash after review process

Orzelius:
ci seems to be in bad health, will pray for fast recovery and retry tomorrow 🙏

PR #12830, opened 2026-02-20
refactor: drop termui-widgets library

smira:
Use the library built for tview, so that we don't have to have two UI libraries working in parallel in the same TUI.


smira:
[asciicast](https://asciinema.org/a/pnhbJNpi0dQhTXsJ)

smira:
This seems to be almost character-perfect replacement of previous implementation, probably even better in some places.

shanduur (approved):
:cool:

PR #12835, opened 2026-02-22
fix: update path handling on talosctl cgroups

nerakhon:
# Pull Request

## What? (description)

A small fix for [bug](https://github.com/siderolabs/talos/issues/12836) in talosctl that solves breakage of `talosctl cgroups --preset=<preset>` .

## Why? (reasoning)

The call to filepath.join in current code causes breakage when using talosctl on windows due to wrong slash introduced into the embed path.

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/cgroupsprinter/presets.go
+++ b/cmd/talosctl/cmd/talos/cgroupsprinter/presets.go
@@ -7,6 +7,7 @@ package cgroupsprinter
 import (
 	"embed"
 	"io/fs"
+	"path"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -37,7 +38,8 @@ func GetPresetNames() []string {
 
 // GetPreset returns the preset by name.
 func GetPreset(name string) Schema {
-	f, err := presetsFS.Open(filepath.Join("presets", name+".yaml"))
+	// embed.FS always uses / as separator, even on Windows, we need OS-agnostic path joining here
+	f, err := presetsFS.Open(path.Join("presets", name+".yaml"))
 	if err != nil {
 		panic(err) // should not fail
 	}


cmd/talosctl/cmd/talos/cgroupsprinter/presets.go
```diff
@@ -27,7 +27,11 @@ func GetPresetNames() []string {
 
 	presets := xslices.Map(list, func(dirEntry fs.DirEntry) string {
 		// cut extension
-		return strings.TrimSuffix(dirEntry.Name(), filepath.Ext(dirEntry.Name()))
+		name := dirEntry.Name()
+		if idx := strings.LastIndex(name, "."); idx != -1 {
+			return name[:idx]
+		}
+		return name
```
shanduur: Is this change needed? It's doing _almost_ the same thing, except it ignores some edge cases, e.g. dot-prefixed names.

smira: I agree, I think only the last chunk is needed

nerakhon: I've rewritten this to remove the filepath import completely. To me that seemed preferrable. But I leave it to you guys. If you have strong preferrence on minimizing the change. Though I agree that if there was a dot prefixed file this wouldn't work but I kinda assumed that isn't probable.

shanduur: I'd prefer leaving the `path/filepath` import here and do not implement the function ourselves - no need to duplicate efforts that are in stdlib, and make it confusing for future maintenance.

cmd/talosctl/cmd/talos/cgroupsprinter/presets.go
```diff
@@ -37,7 +41,7 @@ func GetPresetNames() []string {
 
 // GetPreset returns the preset by name.
 func GetPreset(name string) Schema {
-	f, err := presetsFS.Open(filepath.Join("presets", name+".yaml"))
+	f, err := presetsFS.Open(path.Join("presets", name+".yaml"))
```
shanduur, line 42: I'd add comment here explaining that embed.FS always uses `/` as path separator, so we can't use function that handles os-specific separators.

Issue #12841, opened 2026-02-23
implement new Install/Upgrade APIs


smira:
```proto
syntax = "proto3";

package machine;

option go_package = "github.com/siderolabs/talos/pkg/machinery/api/machine";
option java_package = "dev.talos.api.machine";

import "common/common.proto";
import "google/protobuf/duration.proto";
import "google/protobuf/timestamp.proto";
import "google/protobuf/empty.proto";

// The InstallService handles installation and upgrade operations.
service InstallService {
  // Install Talos to disk (if it hasn't been installed yet).
  rpc Install(InstallServiceInstallRequest) returns (stream InstallServiceInstallResponse);
  // Upgrade Talos to a new version.
  rpc Upgrade(InstallServiceUpgradeRequest) returns (stream InstallServiceUpgradeResponse);
}

message InstallArtifactsSource {
    string image_name = 1;
}

message InstallDestination {
    string disk = 1;
}

message InstallServiceInstallRequest {
    common.ContainerdInstance containerd = 1;
    InstallArtifactsSource source = 2;
    InstallDestination destination = 3;
}

message InstallServiceInstallResponse {
    oneof response {
        string message = 1;
        int32 exit_code = 2;
    }
}

message InstallServiceUpgradeRequest {
    common.ContainerdInstance containerd = 1;
    InstallArtifactsSource source = 2;
}

message InstallServiceUpgradeResponse {
    oneof response {
        string message = 1;
        int32 exit_code = 2;
    }
}
```

smira:
Design ideas:

* make it minimal - the installer image should be pulled using Image APIs, reboot should be done using reboot APIs
* remove all deprecated/legacy settings
* new APIs just run the installer container and report the progress/result/error/....
* in the future, we can extend to support e.g. installing to RAID1, or support installing from ISO UKIs, ...

smira:
For the `Install` API, we can't do proper testing until the issue #12702 is resolved.

For the `Upgrade` API, we can hook it up into `talosctl upgrade` (while keeping backwards compatibility), and decompose it there into three calls:

* Image.Pull
* Install.Upgrade
* Machine.Reboot

We can update provision integration tests to use new Upgrade API for Talos v1.13+

smira:
Make sure that only a single Upgrade/Install can run at a time.

Install should fail if Talos is already installed.

Upgrade should fail is Talos hasn't been installed yet.

PR #12842, opened 2026-02-23
feat: implement new install/upgrade API

shanduur:
Implement new minimal Install/Upgrade API.

Fixes https://github.com/siderolabs/talos/issues/12841

### DIFF
--- a/cmd/talosctl/cmd/talos/upgrade.go
+++ b/cmd/talosctl/cmd/talos/upgrade.go
@@ -5,40 +5,49 @@
 package talos
 
 import (
-	"cmp"
 	"context"
 	"errors"
 	"fmt"
 	"os"
-	"slices"
 	"strings"
 	"text/tabwriter"
 	"time"
 
-	"github.com/siderolabs/gen/maps"
-	"github.com/siderolabs/gen/xslices"
+	"github.com/blang/semver/v4"
+	"github.com/siderolabs/gen/xerrors"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/peer"
 
+	"github.com/siderolabs/talos/cmd/talosctl/cmd/talos/lifecycle"
+	"github.com/siderolabs/talos/cmd/talosctl/cmd/talos/multiplex"
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/action"
 	"github.com/siderolabs/talos/cmd/talosctl/pkg/talos/helpers"
 	"github.com/siderolabs/talos/pkg/cli"
+	"github.com/siderolabs/talos/pkg/flags"
 	"github.com/siderolabs/talos/pkg/images"
+	"github.com/siderolabs/talos/pkg/machinery/api/common"
 	"github.com/siderolabs/talos/pkg/machinery/api/machine"
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/version"
+	"github.com/siderolabs/talos/pkg/reporter"
 )
 
-var upgradeCmdFlags struct {
+var upgradeCmdFlags = struct {
 	trackableActionCmdFlags
+	imageCmdFlagsType
 
 	upgradeImage string
-	rebootMode   string
-	preserve     bool
-	stage        bool
-	force        bool
-	insecure     bool
+	rebootMode   flags.PflagExtended[machine.RebootRequest_Mode]
+	progress     flags.PflagExtended[reporter.OutputMode]
+
+	force    bool // Deprecated: only used for legacy upgrade path, to be removed in Talos 1.18.
+	insecure bool // Deprecated: only used for legacy upgrade path, to be removed in Talos 1.18.
+	preserve bool // Deprecated: only used for legacy upgrade path, to be removed in Talos 1.18.
+	stage    bool // Deprecated: only used for legacy upgrade path, to be removed in Talos 1.18.
+}{
+	rebootMode: flags.ProtoEnum(machine.RebootRequest_DEFAULT, machine.RebootRequest_Mode_value, machine.RebootRequest_Mode_name),
+	progress:   reporter.NewOutputModeFlag(),
 }
 
 // upgradeCmd represents the processes command.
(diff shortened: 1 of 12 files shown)


internal/app/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,180 @@
...
+	if err := crires.WaitForImageCache(ctx, s.runtime.State().V1Alpha2().Resources()); err != nil {
+		return status.Error(codes.Internal, fmt.Sprintf("failed to wait for the image cache: %v", err))
+	}
+
+	installerImage := req.GetSource().GetImageName()
+	if installerImage == "" {
+		return status.Error(codes.InvalidArgument, "installer image name is required")
+	}
+
+	disk := req.GetDestination().GetDisk()
```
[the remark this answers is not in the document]
shanduur: Handled by the generated code.

internal/app/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if s.runtime.State().Platform().Mode().InContainer() {
+		return status.Error(codes.FailedPrecondition, "upgrade is not supported in container mode")
+	}
+
+	if !s.runtime.State().Machine().Installed() {
+		return status.Error(codes.FailedPrecondition, "machine is not installed")
+	}
+
+	installerImage := req.GetSource().GetImageName()
```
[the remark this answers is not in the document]
shanduur: Handled by the generated code.

internal/app/lifecycle/lifecycle.go
```diff
@@ -0,0 +1,180 @@
...
+
+	if s.runtime.State().Machine().Installed() {
+		return status.Error(codes.AlreadyExists, "machine is already installed")
+	}
+
+	if err := crires.WaitForImageCache(ctx, s.runtime.State().V1Alpha2().Resources()); err != nil {
+		return status.Error(codes.Internal, fmt.Sprintf("failed to wait for the image cache: %v", err))
+	}
+
+	installerImage := req.GetSource().GetImageName()
```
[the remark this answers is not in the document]
shanduur: Handled by the generated code.

api/machine/lifecycle.proto
```diff
@@ -0,0 +1,61 @@
...
+  string image_name = 1;
+}
+
+message InstallDestination {
+  string disk = 1;
+}
+
+message InstallOptions {
+  bool wipe = 1;
+  repeated string extra_kernel_args = 2;
```
smira: this should not be there - we deprecated this completely

api/machine/lifecycle.proto
```diff
@@ -0,0 +1,61 @@
...
+message InstallArtifactsSource {
+  string image_name = 1;
+}
+
+message InstallDestination {
+  string disk = 1;
+}
+
+message InstallOptions {
+  bool wipe = 1;
```
smira: I don't think we need this. if someone needs to wipe, they should wipe using the API I think

api/machine/lifecycle.proto
```diff
@@ -0,0 +1,61 @@
...
+    int32 exit_code = 2;
+  }
+}
+
+message LifecycleServiceInstallResponse {
+  LifecycleServiceInstallProgress progress = 1;
+}
+
+message UpgradeOptions {
+  repeated string extra_kernel_args = 1;
```
smira: same as install - deprecated completely

shanduur: Removed.

api/machine/lifecycle.proto
shanduur: Removed.

api/machine/lifecycle.proto
shanduur: Removed.

api/machine/lifecycle.proto
```diff
@@ -0,0 +1,50 @@
...
+// The LifecycleService handles installation and upgrade operations.
+service LifecycleService {
+  // Install Talos to disk (if it hasn't been installed yet).
+  rpc Install(LifecycleServiceInstallRequest) returns (stream LifecycleServiceInstallResponse);
+  // Upgrade Talos to a new version.
+  rpc Upgrade(LifecycleServiceUpgradeRequest) returns (stream LifecycleServiceUpgradeResponse);
+}
+
+message InstallArtifactsSource {
+  string image_name = 1;
```
smira, line 23: I would really appreciate if we do some short comments for every field for users/future readers.

E.g. `image_name` is a response of `ImageService.Pull` API (after pulling an image).

cmd/talosctl/cmd/talos/upgrade.go
```diff
@@ -56,85 +62,221 @@ var upgradeCmd = &cobra.Command{
...
-		).Run()
-	},
+	containerdInstance, err := upgradeCmdFlags.containerdInstance()
+	if err != nil {
+		return err
+	}
+
+	rep := reporter.New()
+
+	_, err = imagePullInternal(ctx, c, containerdInstance, nodes, upgradeCmdFlags.upgradeImage, rep)
```
smira, line 114: I think the first thing which _should_ fail is the image pull new API, this might be the moment we fall back to old APIs.

Or, alternatively, we could grab versions of all Talos nodes, and if we have at least one below 1.13, we should fall back to "legacy" API

website/content/v1.13/reference/cli.md
```diff
@@ -3329,14 +3329,12 @@ talosctl upgrade [flags]
...
   -e, --endpoints strings          override default endpoints in Talos configuration
-  -f, --force                      force the upgrade (skip checks on etcd health and members, might lead to data loss)
   -h, --help                       help for upgrade
   -i, --image string               the container image to use for performing the install (default "ghcr.io/siderolabs/installer:v1.13.0-alpha.2")
-      --insecure                   upgrade using the insecure (encrypted with no auth) maintenance service
+      --namespace string           namespace to use: "system" (etcd and kubelet images), "cri" for all Kubernetes workloads, "inmem" for in-memory containerd instance (default "system")
   -n, --nodes strings              target the specified nodes
-  -m, --reboot-mode string         select the reboot mode during upgrade. Mode "powercycle" bypasses kexec. Valid values are: ["default" "powercycle"]. (default "default")
+  -m, --reboot-mode string         select the reboot mode during upgrade. Mode "powercycle" bypasses kexec. Values: [default force powercycle] (default "default")
```
smira, line 3337: I wonder if we should hide deprecated flags as well?

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -152,6 +153,7 @@ func (s *Server) Register(obj *grpc.Server) {
 	machine.RegisterMachineServiceServer(obj, s)
 	machine.RegisterImageServiceServer(obj, images.NewService(s.Controller))
 	machine.RegisterDebugServiceServer(obj, &debug.Service{})
+	machine.RegisterLifecycleServiceServer(obj, lifecycle.NewService(s.Controller.Runtime())) // TODO: How do we access logger?
```
smira: I have something I've built for image verification in https://github.com/siderolabs/talos/pull/12853

website/content/v1.13/reference/cli.md
shanduur: Deprecated flags are hidden - we still reboot after an upgrade automatically through CLI (right?)

smira: ah, sorry, now I see it... I thought we deprecated many more flags, but some of them were already deprecated I guess

cmd/talosctl/cmd/talos/upgrade.go
shanduur: I want to reuse the whole internal handler for the imagePull - this way I don't need to reimplement it and just use the same semantics as `talosctl debug` and `talosctl image pull`. It adds a little bit more of internal handling, but it is hidden and (IMHO) easier to maintain.

smira: That's true, what I thought about (and same applies to `talosctl debug`) - the image pull function has automatic fallback to <1.13 APIs, so it would still pull an image, and only when we try to run Upgrade API, it would actually detect a failure and fall back to old upgrade API.

So we have two options here I think:

* check Talos version explicitly before starting any work (we do that in some CLI helpers already, to check version compatibility, but we don't keep the version response anywhere)
* in image pull helper, add an option to skip fall back to legacy API, so that we know immediately it's old Talos, and we fall back to old upgrade API

api/docs.markdown.tmpl
```diff
@@ -43,15 +43,15 @@ description: Talos gRPC API reference.
 | Field | Type | Label | Description |
 | ----- | ---- | ----- | ----------- |
 {{range .Fields -}}
-  | {{.Name}} | [{{.LongType}}](#{{.FullType}}) | {{.Label}} | {{if (index .Options "deprecated"|default false)}}**Deprecated.** {{end}}{{nobr .Description}}{{if .DefaultValue}} Default: {{.DefaultValue}}{{end}} |
+  | {{.Name}} | [{{.LongType}}](#{{.FullType}}) | {{.Label}} | {{if (index .Options "deprecated"|default false)}}**Deprecated.** {{end}}{{nobr .Description | replace "\n\n" "<br><br>"}}{{if .DefaultValue}} Default: {{.DefaultValue}}{{end}} |
```
smira, line 46: I'm not sure what is going on here... if it's a separate change, should we split it out?

internal/app/lifecycle/container.go
```diff
@@ -0,0 +1,338 @@
...
+
+	ctx = leases.WithLease(ctx, l.ID)
+
+	img, err := c8dClient.GetImage(ctx, rc.imageRef)
+	if err != nil {
+		return fmt.Errorf("installer image %q not found in containerd store: %w", rc.imageRef, err)
+	}
+
+	// clean up old container/snapshot
+	if err := cleanupOldContainer(ctx, c8dClient); err != nil {
```
smira: should we do a random `contianerID` like for debug, so that we don't need to cleanup? (as lease thing should do automatic cleanup)?

api/docs.markdown.tmpl
shanduur: Our docs are broken now (at least it looks like it). This replaces `\n` with `<br>`, so they are rendering properly.

smira: I wonder how it affects other conversion procedures... but sounds okay!

internal/app/lifecycle/container.go
shanduur: I think it's done?

smira:
I think this log message look strange

```
[  369.494635] [talos] The Pull option is deprecated, ignored in Lifecycle API, and will be removed in a future release.
```

smira:
I would also add a copy for installer messages to the console - just makes it easier for us to understand what is going on  in case of troubleshooting, or at least print an overall status "install/upgrade sucessful (failed)"

pkg/machinery/api/api.go
```diff
@@ -43,6 +43,7 @@ func TalosAPIdAllAPIs() []protoreflect.FileDescriptor {
 		cosi.File_v1alpha1_state_proto,
 		machine.File_machine_debug_proto,
 		machine.File_machine_image_proto,
+		machine.File_machine_lifecycle_proto,
```
shanduur, line 46: @smira this?

internal/app/machined/pkg/system/services/machined.go
```diff
@@ -45,6 +45,9 @@ var rules = map[string]role.Set{
 
 	"/machine.DebugService/ContainerRun": role.MakeSet(role.Admin),
 
+	"/machine.LifecycleService/Install": role.MakeSet(role.Admin),
+	"/machine.LifecycleService/Upgrade": role.MakeSet(role.Admin),
```
shanduur, line 51: It's also marked here?

pkg/machinery/api/api.go
smira: I guess I'm blind, ignore me. I looked for it, but I didn't see it

internal/app/lifecycle/container.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/constants"
+)
+
+func generateContainerID() (string, error) {
+	b := make([]byte, 8)
+	if _, err := rand.Read(b); err != nil {
+		return "", fmt.Errorf("failed to generate random ID: %w", err)
+	}
+
+	return fmt.Sprintf("debug-%s", hex.EncodeToString(b)), nil
```
smira:
```suggestion
	return fmt.Sprintf("installer-%s", hex.EncodeToString(b)), nil
```

shanduur: Fixed.

cmd/talosctl/cmd/talos/upgrade.go
shanduur: Added the checks, as advised.

api/machine/lifecycle.proto
shanduur: Done

PR #12847, opened 2026-02-24
feat: include hid-multitouch.ko kernel module in rootfs

AndreasFreund:
# Pull Request

## What? (description)
Add hid-multitouch.ko kernel module to hack/modules-*.txt for it to be included in default rootfs.
Was added to kernel build in pkgs/kernel [here](https://github.com/siderolabs/pkgs/pull/1466).

## Why? (reasoning)
Enables passing multi-touch touchscreens to containers as input devices.

## Acceptance
Built metal image and tested on amd64 device with multi-touch capable display attached via USB (& DisplayPort). Talos boots & kernel module loads automatically.


rothgar:
Are you running Talos on tablets/kiosks???

PR #12853, opened 2026-02-24
feat: add image verification endpoint

### LINKED ISSUE #12466: support machine-wide image signature verification policy
Support an image verification policy which applies to any image being pulled by Talos/CRI, and working transparently with registry mirrors or registry auth.

Yes, there is a Kubernetes admission policy plugin, but it doesn't apply to static pods and Talos image pulls.

Yes, there is support in containerd transfer service, but we can't use it right now as it doesn't support registry auth.

We need to figure out some solution which might involve patching containerd/CRI.

smira:
Add support for whole machine-wide image verification configuration.
Configuration is a set of rules applied top-down to the image reference,
each specifying a specific cosign-based identity or static public key
claim.

Talos provides a machined API to verify an image reference, resolving it
to the digest on the way as needed.

Talos itself hooks up in the image verification process, while
containerd CRI plugin accesses same API via the machined socket.

Fixes #12466


internal/app/images/verify.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+
+		return rule.Verifier, rule.Verify
+	}
+
+	return nil, false
+}
+
+func checkSignatures(ctx context.Context, digestRef name.Reference, verifier *security.ImageVerificationVerifierSpec) (verificationResult, error) {
+	trustedMaterial, err := cosign.TrustedRoot()
```
frezbo: we should only get this in online mode

smira: yes, this is total WIP at the moment, I'm resurrecting & rebasing Laura's WIP code

smira:
TODO list:

- [x] deal with deadcode elimination issue
- [ ] ~rework config to support multiple keyless identities~ (cosign doesn't like multiple identities)
- [x] plug into Talos pull process
- [x] plug into containerd
- [x] add a policy for registry.k8s.io and inject it in the CI
- [x] hack/release.toml note
- [ ] ~support for image-cache create to download signatures~ #12892
- [ ] ~test in image-cache, airgapped modes, etc.~ #12892

internal/app/machined/pkg/controllers/security/image_verification_config.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		r.StartTrackingOutputs()
+
+		machineConfig, err := safe.ReaderGetByID[*configres.MachineConfig](ctx, r, configres.ActiveID)
+		if err != nil && !state.IsNotFoundError(err) {
+			return fmt.Errorf("failed to get machine config: %w", err)
+		}
+
+		if machineConfig != nil {
```
frezbo, line 70: return early:



```suggestion
		if machineConfig == nil ||  machineConfig.Config().ImageVerificationConfig() == nil {
		return nil
		}
```

smira: I can't return early actually, as I need to clean up the rules if the config was destroyed

internal/app/images/images.go
```diff
@@ -34,12 +37,14 @@ type Service struct {
 	machine.UnimplementedImageServiceServer
 
 	controller runtime.Controller
+	logger     *zap.Logger
```
shanduur, line 40: It's never used (?)

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -117,6 +118,9 @@ type Server struct {
 	// ShutdownCtx signals that the server is shutting down.
 	ShutdownCtx context.Context //nolint:containedctx
 
+	// Zap logger.
+	Logger *zap.Logger
```
shanduur, line 123: This is never set?

internal/app/images/images.go
smira: it is used in the `Verify()` method (new one) to pass the logger down to the actual verification

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
smira: this is set when instantiating machined API (it is done only once, directly as a struct field)

PR #12865, opened 2026-02-27
feat: bump kube-network-policy to v1.0.0

smira:
The latest release.


smira:
> no generated content?

there shouldn't be, but CI should be mad at me if so :)

PR #12873, opened 2026-03-01
feat: add --cidr-v6 flag for custom IPv6 prefix in QEMU provisioner

### LINKED ISSUE #12872: feat: support custom IPv6 CIDR for QEMU provisioner (--cidr-v6 flag)
## Problem

The QEMU provisioner auto-derives IPv6 ULA prefix from the IPv4 CIDR. Users cannot specify a custom IPv6 prefix.

## Use Case

Joining QEMU-provisioned Talos VMs to remote clusters that use public IPv6 addressing (e.g., `2a01:e11:2440:2435::/80`). This is useful for:
- Development/testing nodes that need to join production-like clusters over public IPv6
- Edge nodes with public IPv6 connectivity
- Hybrid deployments where some nodes are on-prem with public IPv6

## Proposed Solution

Add a `--cidr-v6` flag to `talosctl cluster create qemu` that accepts a custom IPv6 CIDR. If not specified, fall back to the current ULA derivation behavior (backward compatible).

## Example

```bash
talosctl cluster create qemu   --cidr 10.5.0.0/24   --cidr-v6 2a01:e11:2440:2435::/80   --ipv6   --workers 1
```

## Current Workaround

Currently, the only options are:
1. Use Docker provisioner (which has its own limitations with IPv6)
2. Run QEMU manually with custom networking (no talosctl lifecycle management)

tibrezus:
## Summary

- Adds `--cidr-v6` flag to `talosctl cluster create qemu` to specify a custom IPv6 CIDR
- Falls back to ULA derivation (existing behavior) if `--cidr-v6` is not specified
- Enables joining QEMU-provisioned nodes to clusters with public IPv6 addressing

## Changes

- Add `NetworkCIDRv6` field to `Common` options struct in `options.go`
- Add `networkCIDRv6FlagName` constant and flag binding in `cmd.go` / `cmd_qemu.go`
- Modify `initCIDRs()` in `common.go` to use custom IPv6 CIDR when provided

## Example Usage

```bash
talosctl cluster create qemu   --cidr 10.5.0.0/24   --cidr-v6 2a01:e11:2440:2435::/80   --ipv6   --workers 1
```

## Testing

- All existing unit tests pass
- Built talosctl and verified `--help` shows the new flag
- Tested with custom IPv6 CIDR and verified it is used instead of ULA

Closes #12872

## Breaking Changes

None - backward compatible. If `--cidr-v6` is not specified, the existing ULA derivation behavior is used.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
@@ -381,18 +381,33 @@ func (m *Maker[T]) initCIDRs() error {
 	}
 
 	if !cidr4.Addr().Is4() {
-		return errors.New("IPV4 CIDR expected, got IPV6 CIDR")
+		return errors.New("IPv4 CIDR expected for --cidr, got IPv6 CIDR")
 	}
 
-	// use ULA IPv6 network fd00::/8, add 'TAL' in hex to build /32 network, add IPv4 CIDR to build /64 unique network
-	cidr6, err := netip.ParsePrefix(
-		fmt.Sprintf(
-			"fd74:616c:%02x%02x:%02x%02x::/64",
-			cidr4.Addr().As4()[0], cidr4.Addr().As4()[1], cidr4.Addr().As4()[2], cidr4.Addr().As4()[3],
-		),
-	)
-	if err != nil {
-		return fmt.Errorf("error validating cidr IPv6 block: %w", err)
+	var cidr6 netip.Prefix
+
+	if m.Ops.NetworkCIDRv6 != "" {
+		m.Ops.NetworkIPv6 = true
+
+		cidr6, err = netip.ParsePrefix(m.Ops.NetworkCIDRv6)
+		if err != nil {
+			return fmt.Errorf("error validating cidr-v6 block: %w", err)
+		}
+
+		if !cidr6.Addr().Is6() {
+			return errors.New("IPv6 CIDR expected for --cidr-v6, got IPv4 CIDR")
+		}
+	} else {
+		// use ULA IPv6 network fd00::/8, add 'TAL' in hex to build /32 network, add IPv4 CIDR to build /64 unique network
+		cidr6, err = netip.ParsePrefix(
+			fmt.Sprintf(
+				"fd74:616c:%02x%02x:%02x%02x::/64",
+				cidr4.Addr().As4()[0], cidr4.Addr().As4()[1], cidr4.Addr().As4()[2], cidr4.Addr().As4()[3],
+			),
+		)
+		if err != nil {
+			return fmt.Errorf("error validating cidr IPv6 block: %w", err)
+		}
 	}
 
 	var cidrs []netip.Prefix
--- a/cmd/talosctl/cmd/mgmt/cluster/create/cmd_qemu.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/cmd_qemu.go
@@ -33,7 +33,7 @@ func init() {
 	commonFlags := getCommonUserFacingFlags(&cOps)
 	addControlplanesFlag(commonFlags, &cOps.Controlplanes)
 	addTalosVersionFlag(commonFlags, &cOps.TalosVersion, "the desired talos version")
-	commonFlags.StringVar(&cOps.NetworkCIDR, networkCIDRFlagName, "10.5.0.0/24", "CIDR of the cluster network")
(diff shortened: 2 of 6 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
```diff
@@ -379,18 +379,32 @@ func (m *Maker[T]) initCIDRs() error {
...
+		if !cidr6.Addr().Is6() {
+			return errors.New("IPv6 CIDR expected for --cidr-v6, got IPv4 CIDR")
+		}
+	} else {
+		cidr6, err = netip.ParsePrefix(
+			fmt.Sprintf(
+				"fd74:616c:%02x%02x:%02x%02x::/64",
+				cidr4.Addr().As4()[0], cidr4.Addr().As4()[1], cidr4.Addr().As4()[2], cidr4.Addr().As4()[3],
+			),
+		)
```
dsseng, line 407: Consider keeping the comment regarding the meaning of octets 74 61 6c

```suggestion
		// use ULA IPv6 network fd00::/8, add 'TAL' in hex to build /32 network, add IPv4 CIDR to build /64 unique network
		cidr6, err = netip.ParsePrefix(
			fmt.Sprintf(
				"fd74:616c:%02x%02x:%02x%02x::/64",
				cidr4.Addr().As4()[0], cidr4.Addr().As4()[1], cidr4.Addr().As4()[2], cidr4.Addr().As4()[3],
			),
		)
```

cmd/talosctl/cmd/mgmt/cluster/create/cmd_qemu.go
```diff
@@ -34,7 +34,8 @@ func init() {
 	commonFlags := getCommonUserFacingFlags(&cOps)
 	addControlplanesFlag(commonFlags, &cOps.Controlplanes)
 	addTalosVersionFlag(commonFlags, &cOps.TalosVersion, "the desired talos version")
-	commonFlags.StringVar(&cOps.NetworkCIDR, networkCIDRFlagName, "10.5.0.0/24", "CIDR of the cluster network")
+	commonFlags.StringVar(&cOps.NetworkCIDR, networkCIDRFlagName, "10.5.0.0/24", "CIDR of the cluster network (IPv4, ULA network for IPv6 is derived in automated way)")
+	commonFlags.StringVar(&cOps.NetworkCIDRv6, networkCIDRv6FlagName, "", "CIDR of the cluster network for IPv6 (optional, overrides automatic ULA derivation)")
```
dsseng: IMO it does not make much sense to mention IPv6 in the IPv4 CIDR option. Could this maybe be clearer to the user?

```suggestion
	commonFlags.StringVar(&cOps.NetworkCIDR, networkCIDRFlagName, "10.5.0.0/24", "IPv4 CIDR of the cluster network")
	commonFlags.StringVar(&cOps.NetworkCIDRv6, networkCIDRv6FlagName, "", "IPv6 CIDR of the cluster network (optional, defaults to automatic ULA derivation)")
```

dsseng: Actually, `m.Ops.NetworkIPv6 = true` gets enabled by this flag being set. Therefore, we never actually see ULAs. Should we maybe bring the `--ipv6` boolean flag from `talosctl cluster create dev` to here as well? cc @Orzelius

dsseng:
After some discussions we've decided it's still too early to enable IPv6 in the user-facing `qemu` command, as it is not yet working without extra config and will therefore confuse users.

Also, it's been proposed to drop the ULA auto-generation altogether - it is of quite little use, and the user could always supply ULA prefix on their own, given they configured it to do something useful on the upstream side (e.g. host doing NPTv6)

Could we then work on bringing this feature to the `dev` command first, as I'm currently working on adding all the features for an IPv6 testbed? I will use this PR as a base for my experiments and then add a review on how to direct this PR for the enablement steps currently underway

tibrezus:
@dsseng thanks for the reply, I am working on a custom edge platform which is ipv6 first over native cilium CNI, managed to have it working after adding some contributions there.
This is not a **must** but nice to have for people developing on native ipv6 networks which is a growing market.
Happy to contribute and test over the `dev` command.

dsseng:
Yes, I'm mostly working with the same concept to create a testbed to test Talos in IPv6-mostly and IPv6-only scenarios. I'm focusing on testing Cilium as well, and currently implementing NDP RA and BGP to simulate a real network.

tibrezus:
Per your earlier comment, I've redirected everything to dev instead. The qemu subcommand no longer exposes any IPv6 flags (--cidr-v6 or --ipv6). When the IPv6 testbed work is ready, the right combination of flags can be exposed on qemu together.

PR #12876, opened 2026-03-01
fix: do not fail on RO virtiofs

### LINKED ISSUE #12874: ExternalVolumeConfig VirtioFS mounts do not support read-only backings
### Description
1.13.0's new `ExternalVolumeConfig` allows for mounting VirtioFS volumes from a host into the Talos VM (Yay!); however, it does not work when the backing filesystem is read-only. This is true for both `readOnly: true` and `readOnly: false`, even when SELinux is disabled.

`block.MountController` attempts to perform chmod and SELinux xattr labeling, both of which fail when the backing volume is truly read-only.

When `readOnly: true`, SELinux xattr labeling is attempted and fails.
When `readOnly: false` and `selinux=0`, chmod is attempted and fails.

The release notes state virtiofs volumes are unsupported with SELinux enforcing, but this occurs in permissive mode and with SELinux fully disabled.

The controller should skip write operations (xattr, chmod) when `readOnly: true`.

### Logs

**readOnly: true**
Image

**readOnly: false, selinux=0**
Image

### Environment

- Talos version: `v1.13.0-alpha.2`
- Kubernetes version: `1.36.0-alpha.1`
- Platform: `amd64`, running no-cloud secureboot in Proxmox VMs

shanduur:
Fixes #12874

### DIFF
--- a/internal/app/machined/pkg/controllers/block/mount.go
+++ b/internal/app/machined/pkg/controllers/block/mount.go
@@ -22,7 +22,6 @@ import (
 	"github.com/siderolabs/go-blockdevice/v2/swap"
 	"github.com/siderolabs/go-pointer"
 	"go.uber.org/zap"
-	"golang.org/x/sys/unix"
 
 	"github.com/siderolabs/talos/internal/pkg/mount/v3"
 	"github.com/siderolabs/talos/internal/pkg/selinux"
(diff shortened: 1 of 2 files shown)


jessemartin:
Thank you for digging into this @shanduur!

internal/app/machined/pkg/controllers/block/mount.go
jessemartin, line 648: I suspect this is where the `readOnly: true` path was failing -> `m.point.Mount` -> `selinux.SetLabel`.

Looks like you've addressed it with the addition of `FilterSelinuxLabelErrors`. LGTM (for what it counts)!

smira (approved):

PR #12877, opened 2026-03-02
fix: ignore image digest when doing upgrade-k8s

smira:
The `talosctl upgrade-k8s` doesn't support pinning to image digests, but it should ignore any image digests if they already exist in the machine configuration.

### DIFF
--- a/pkg/images/list.go
+++ b/pkg/images/list.go
@@ -6,6 +6,7 @@ package images
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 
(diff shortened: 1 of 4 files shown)


pkg/images/list.go
```diff
@@ -115,6 +116,8 @@ func ListWithOptions(config config.Config, opts VersionsListOptions) Versions {
 }
 
 func mustParseTag(s string) name.Tag {
+	s, _, _ = strings.Cut(s, "@") // ignore digest if present
```
smira, line 119: this is a new issue in 1.13 only (cc @shanduur )

shanduur (approved):

PR #12881, opened 2026-03-02
feat: update Kubernetes to 1.36.0-alpha.2

smira:
See https://github.com/kubernetes/kubernetes/releases/tag/v1.36.0-alpha.2


smira:
Still not updating Go module as Flux package is not compatible at the moment.

PR #12887, opened 2026-03-04
feat: add trusted roots generation to stdpatches

rsmitty:
This PR would add the ability to generate a standardized document for trusted roots. Ran into wanting this with the vsphere provider and I generally feel like we should try to extend the available standard patches here over time.

### DIFF
--- a/pkg/machinery/config/generate/stdpatches/stdpatches.go
+++ b/pkg/machinery/config/generate/stdpatches/stdpatches.go
@@ -6,6 +6,8 @@
 package stdpatches
 
 import (
+	"fmt"
+
 	"go.yaml.in/yaml/v4"
 
 	"github.com/siderolabs/talos/pkg/machinery/config"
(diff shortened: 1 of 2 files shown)


pkg/machinery/config/generate/stdpatches/stdpatches.go
```diff
@@ -35,6 +38,18 @@ func WithStaticHostname(versionContract *config.VersionContract, hostname string
...
+// WithTrustedRoots returns a patch that sets trusted roots in the machine configuration.
+func WithTrustedRoots(versionContract *config.VersionContract, trustedRoots string) ([]byte, error) {
+	if versionContract.MultidocNetworkConfigSupported() {
+		trustedRootsConfig := security.NewTrustedRootsConfigV1Alpha1()
+		trustedRootsConfig.Certificates = trustedRoots
+
+		return patchFromDocument(trustedRootsConfig)
+	}
+
+	return []byte(""), fmt.Errorf("trusted roots patch is not supported for version contract %s", versionContract.String())
```
smira: This could be nil simply instead of []byte("")

smira (approved):

smira:
I added some tests

PR #12888, opened 2026-03-05
fix: environment suite failures

shanduur:
Environment suite tests fail often, especially on AWS/GCP.
This change makes the tests more robust.

### DIFF
--- a/internal/integration/api/environment.go
+++ b/internal/integration/api/environment.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/cosi-project/runtime/pkg/safe"
+	"google.golang.org/grpc/codes"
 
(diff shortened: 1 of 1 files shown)


internal/integration/api/environment.go
```diff
@@ -108,7 +109,16 @@ func (suite *EnvironmentSuite) validateEnvironment(node string, expectedVariable
 	ctx := client.WithNode(suite.ctx, node)
```
smira, line 111: it should be better to do this instead
```suggestion
  	suite.ClearConnectionRefused(suite.ctx, node)
	ctx := client.WithNode(suite.ctx, node)
```

frezbo (approved):

Issue #12885, opened 2026-03-04
feat: development flags for talosctl cluster create docker


battamatt:
## Feature Request

In Talos v1.12+, development flags like` --skip-k8s-node-readiness-check` were moved under the `talosctl cluster create dev` command. However, dev appears hardcoded to the QEMU provisioner.

However, `talosctl cluster create dev` appears strictly coupled to the QEMU provisioner. There is currently no apparent way to use these development flags when targeting the Docker provisioner.

### Use Case
We use cluster.network.cni.name: none in our machine configuration because we install Cilium via kubectl immediately after bootstrap. Because the nodes never reach a Ready state without a CNI, talosctl cluster create docker hangs indefinitely and eventually times out. In v1.11, we bypassed this using --skip-k8s-node-readiness-check, but this flag is no longer available for the Docker-specific subcommand.

### To Reproduce
Try to skip readiness check on Docker:
talosctl cluster create docker --skip-k8s-node-readiness-check
Result: Error: unknown flag: --skip-k8s-node-readiness-check

Try to use the dev command with Docker:
talosctl cluster create dev --provisioner docker
Result: no unknown flag error but defaults to QEMU

### Expected Behavior / Proposed Solution
`talosctl cluster create docker` should support the same development flags as the `dev` subcommand, or `dev` should allow a `--provisioner docker` flag.

smira:
Thanks for reporting it, we will look into bringing back `--provisioner=docker` in `talosctl  cluster dev`.

Please note that you can go directly to docker, you don't need `talosctl`: https://docs.siderolabs.com/talos/v1.12/platform-specific-installations/local-platforms/docker#running-talos-in-docker-manually

This way you can build your own flow as you like it.

`talosctl cluster create` is not a production deployment tool in general.

smira:
@Orzelius I wonder if we can solve this more easily though by skipping the checks on node readiness if the CNI is set to none...

battamatt:
> [@Orzelius](https://github.com/Orzelius) I wonder if we can solve this more easily though by skipping the checks on node readiness if the CNI is set to none...

That would be enough for our use case at least.

Orzelius:
> [@Orzelius](https://github.com/Orzelius) I wonder if we can solve this more easily though by skipping the checks on node readiness if the CNI is set to none...

I'd like to avoid bringing the dev flags to the docker provider if possible so this solution is perfect!

anthosz:
Hello,

[...]

PR #12896, opened 2026-03-06
fix: cluster create hanging when CNI is disabled

Orzelius:
skip some of the readiness checks if CNI is disabled

fixes https://github.com/siderolabs/talos/issues/12885

### DIFF
--- a/pkg/cluster/check/default.go
+++ b/pkg/cluster/check/default.go
@@ -9,8 +9,12 @@ import (
 	"slices"
 	"time"
 
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
+
 	"github.com/siderolabs/talos/pkg/conditions"
 	"github.com/siderolabs/talos/pkg/machinery/config/machine"
+	"github.com/siderolabs/talos/pkg/machinery/resources/config"
 )
 
 // DefaultClusterChecks returns a set of default Talos cluster readiness checks.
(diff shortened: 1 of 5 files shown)


cmd/talosctl/cmd/mgmt/cluster/create/create.go
```diff
@@ -180,6 +181,12 @@ func bootstrapCluster(ctx context.Context, clusterAccess *access.Adapter, cOps c
 		checks = slices.Concat(check.PreBootSequenceChecks(), check.K8sComponentsReadinessChecks())
 	}
 
+	if configBundle.ControlPlaneCfg.Cluster().Network().CNI().Name() == "none" {
```
smira: I think this is incomplete fix, I think a way better/easier way is to check whether CNI is disabled directly, and skip node readiness check (this would work any way the check is called)

smira: I think it might be easier to go to K8s node readiness check itself,  query BootstrapManifestConfig resource: https://github.com/laurazard/talos/blob/f79a0f06426a44ce61767e4ca99da261e32a6cfe/internal/app/machined/pkg/controllers/k8s/control_plane.go#L356-L357

And check this field - if it's false, skip node readiness check...

pkg/cluster/check/default.go
```diff
@@ -45,6 +56,13 @@ func DefaultClusterChecks() []ClusterCheck {
 			// wait for coredns to report ready
 			func(cluster ClusterInfo) conditions.Condition {
 				return conditions.PollingCondition("coredns to report ready", func(ctx context.Context) error {
+					if CNIDisabled, err := cniDisabledStatus(ctx, cluster); err != nil {
```
smira:
```suggestion
					if cniDisabled, err := cniDisabledStatus(ctx, cluster); err != nil {
```

pkg/cluster/check/default.go
```diff
@@ -45,6 +56,13 @@ func DefaultClusterChecks() []ClusterCheck {
...
+						return err
+					} else if CNIDisabled {
+						// skip check if CNI is disabled
+						return nil
```
smira: we have a specific "Skip" error which will show correct "SKIP" status

PR #12935, opened 2026-03-10
chore(ci): update nvidia test to use gpu-operator

frezbo:
Update NVIDIA tests to use GPU Operator.


smira:
let's also add `hack/release.toml` note please!

PR #12938, opened 2026-03-10
feat: implement routing rules support

### LINKED ISSUE #7184: Support for routing rules and tables

shanduur:
Add RoutingRuleConfig multi-doc config type for management of routing rules.
KubeSpan now uses COSI resources instead of direct kernel management.

Closes #7184


pkg/machinery/config/types/network/routing_rule.go
```diff
@@ -0,0 +1,287 @@
...
+	if s.RuleTable == nethelpers.TableLocal || s.RuleTable == nethelpers.TableMain || s.RuleTable == nethelpers.TableDefault {
+		errs = errors.Join(errs, fmt.Errorf("cannot use reserved table %d in routing rule", s.RuleTable))
+	}
+
+	if s.RuleFwMask != 0 && s.RuleFwMark == 0 {
+		errs = errors.Join(errs, errors.New("fwMask requires fwMark to be set"))
+	}
+
+	return nil, errs
+}
```
shanduur, line 217: I'm not sure about the validation rules - if there's something too tight or too loose, I'll happily adjust them.

smira:
Promoting to `misc-1` - it runs KubeSpan test as well

hack/release.toml
```diff
@@ -278,6 +278,12 @@ Talos now enables [CDI](https://github.com/cncf-tags/container-device-interface)
 CDI spec files under `/run/cdi`.
 """"
 
+    [notes.routing_rules]
+        title = "Routing Rules Support"
+        description = """\
+Talos now supports routing rules via the new `RoutingRuleConfig` machine config document.
```
smira, line 304: 🎉

internal/app/machined/pkg/controllers/network/routing_rule_config.go
```diff
@@ -0,0 +1,190 @@
...
+
+		action := ruleCfg.Action()
+		if action == nethelpers.RoutingRuleActionUnspec {
+			action = nethelpers.RoutingRuleActionUnicast
+		}
+
+		rule.Action = action
+
+		// determine family from src/dst prefixes
+		rule.Family = ctrl.determineFamily(src, dst)
```
smira: I'm thinking about the case when neither src/dst is set, in this case I want the rule to apply to both ipv4 & ipv6, so I think we need to "copy" such rule to be both ipv4/ipv6 (imagine e.g. the rule like "if iifname == 'eth0'" - this should apply to both families?

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
```diff
@@ -0,0 +1,321 @@
...
+			if ctrl.matchesRule(&existingRules[i], spec) {
+				// rule already exists
+				return nil
+			}
+		}
+
+		// create the rule
+		msg := ctrl.buildRuleMessage(spec)
+
+		if err := conn.Rule.Add(msg); err != nil {
```
smira, line 178: does it work if you _change_ the rule? so `ctrl.matchesRule` returns "false", but the rule e.g. has the same priority?

or in other words, what if I update the rule config but keep the priority?

internal/app/machined/pkg/controllers/network/routing_rule_status.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+
+			priority := pointer.SafeDeref(rule.Attributes.Priority)
+
+			table := uint32(rule.Table)
+			if rule.Attributes.Table != nil {
+				table = *rule.Attributes.Table
+			}
+
+			id := network.RoutingRuleID(family, src, dst, priority)
```
smira: I haven't looked into this, but what is the Linux internal "unique ID" for the rule? is it priority? family+priority? can you have two rules with the same priority?

I'm asking this as the resource ID we use should match whatever is the Linux "uniqueness" condition

pkg/machinery/config/types/network/routing_rule.go
```diff
@@ -0,0 +1,286 @@
...
+
+	//   description: |
+	//     A unique name for this routing rule.
+	//     Used as the resource identifier.
+	//
+	//   examples:
+	//    - value: >
+	//       "from-10.0.0.0/8-to-custom-table"
+	//   schemaRequired: true
+	MetaName string `yaml:"name"`
```
smira: also comes back to the question about uniqueness of rule. 

if e.g. the priority is the unique constraint, we could push the name to be the priority, and then config merging works naturally by not allowing you to have non-unique priorities in the first place

pkg/machinery/config/types/network/routing_rule.go
smira: I think we should validate priority and prohibit at least Linux default priorities:

```
0:      from all lookup local
32766:  from all lookup main
32767:  from all lookup default
```

I guess it should be `>0 && < 32766` (at least?)

smira:
I think it's really 🆒 , I added some overall design feedback

internal/app/machined/pkg/controllers/network/routing_rule_status.go
shanduur: Yeah, priority is the uniqueness constraint here, at least according to [manpage](https://man7.org/linux/man-pages/man8/ip-rule.8.html).

pkg/machinery/config/types/network/routing_rule.go
shanduur: Right, there is even a race condition now. Let me fix it.

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
shanduur: Yes! Fixed!

pkg/machinery/config/types/network/routing_rule.go
shanduur: Done!

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
```diff
@@ -0,0 +1,359 @@
...
+// Outputs implements controller.Controller interface.
+func (ctrl *RoutingRuleSpecController) Outputs() []controller.Output {
+	return nil
+}
+
+// Run implements controller.Controller interface.
+//
+//nolint:gocyclo
+func (ctrl *RoutingRuleSpecController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	// No rtnetlink multicast group for rule changes, so we use periodic polling
```
smira: I wonder if we could simply add RoutingRuleStatus as the input here to act on rule changes, while avoiding ticking in both controllers?

smira: and I can see the group here: https://github.com/torvalds/linux/blob/80234b5ab240f52fa45d201e899e207b9265ef91/include/uapi/linux/rtnetlink.h#L700

does it not work?

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
smira: There's also Rule.Replace() which should probably atomically replace without ever dropping it?

And it might allow us to remove ErrExist below

pkg/machinery/config/types/network/routing_rule.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// Clone implements config.Document interface.
+func (s *RoutingRuleConfigV1Alpha1) Clone() config.Document {
+	return s.DeepCopy()
+}
+
+// Name implements config.NamedDocument interface.
+func (s *RoutingRuleConfigV1Alpha1) Name() string {
+	return strconv.FormatUint(uint64(s.RulePriority), 10)
```
smira: I think this unfortunately is going to break strategic patch merging - it has a static assumption that `name:` is the key for multi-doc configs. So we can't make it a custom YAML key

pkg/machinery/config/types/network/routing_rule.go
```diff
@@ -0,0 +1,289 @@
...
+	//   description: |
+	//     Source address prefix to match.
+	//     If empty, matches all sources.
+	//
+	//   examples:
+	//    - value: >
+	//       "10.0.0.0/8"
+	//   schema:
+	//     type: string
+	RuleSrc string `yaml:"src,omitempty"`
```
smira: we can use a custom `Prefix` type defined in this package, it already does proper YAML marshal/unmarshal, and embed `netip.Prefix`, so you can skip validating for correctness here

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
shanduur: The issue with Replace is that it duplicates the rules if called multiple times for the same values - I'm not sure why:

image

smira: interesting, so the actual unique key is something else?

smira: I read the Linux code a bit, and looks like it kind of has no uniqueness constraint at all, so we can't really replace, we can only add and delete. Replace seems just to instert duplicates.

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
shanduur: 🤦🏻 Looks like I'm just bad at searching.

smira: The constant in Go says "V4" only, but looks like it should handle both according to kernel headers.

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
shanduur: There is no uniqueness constraint _strictly defined_ but it looks like priority should be unique (by conventions) according to docs.

internal/app/machined/pkg/controllers/network/routing_rule_spec.go
shanduur: Looks like it does not work for IPv6 - I'm looking into the kernel source what else can I pass to make the watch work for IPv6 routing rules.

smira: In the header it's same constant for both 🤷

pkg/machinery/config/schemas/config.schema.json
```diff
@@ -2761,6 +2761,100 @@
...
+            "RoutingRuleConfig"
+          ],
+          "title": "kind",
+          "description": "kind is the kind of the resource.\n",
+          "markdownDescription": "kind is the kind of the resource.",
+          "x-intellij-html-description": "\u003cp\u003ekind is the kind of the resource.\u003c/p\u003e\n"
+        },
+        "name": {
+          "type": "string",
+          "title": "name",
```
binary-zero-one, line 2786: The property is "name", but the description includes "priority"

shanduur: Yes. That’s because we are doing YAML merges based on the `name` field. Thus the value of name is the priority.

website/content/v1.13/schemas/config.schema.json
```diff
@@ -2761,6 +2761,100 @@
...
+        },
+        "table": {
+          "type": "string",
+          "title": "table",
+          "description": "The routing table to look up if the rule matches.\n",
+          "markdownDescription": "The routing table to look up if the rule matches.",
+          "x-intellij-html-description": "\u003cp\u003eThe routing table to look up if the rule matches.\u003c/p\u003e\n"
+        },
+        "action": {
+          "type": "integer",
```
rxchard, line 2813: This indicates "integer" for `action`, yet the description and docs show "unicast" (`= 1`?) as a possible value.

PR #12939, opened 2026-03-10
refactor: drop apid file socket

smira:
This was yet another socket with implicit auth - remove it completely by reworking the only usecase for it - cluster-side health checks. Now these health checks build a "regular" network Talos API client (as they anyways work only controlplane nodes).

Refactor the check for controlplane nodes to use resources instead of machine config directly (as machine config might not be always present).

### DIFF
--- a/pkg/cluster/local.go
+++ b/pkg/cluster/local.go
@@ -6,18 +6,36 @@ package cluster
 
 import (
 	"context"
-	"errors"
+	"fmt"
+	"time"
 
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials/insecure"
+	"github.com/cosi-project/runtime/pkg/safe"
+	"github.com/cosi-project/runtime/pkg/state"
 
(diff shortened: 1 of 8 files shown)


pkg/cluster/local.go
```diff
@@ -6,41 +6,87 @@ package cluster
...
 //
 // Client implements ClientProvider interface.
 func (c *LocalClientProvider) Client(endpoints ...string) (*client.Client, error) {
-	if len(endpoints) > 0 {
-		return nil, errors.New("custom endpoints not supported with LocalClientProvider")
+	if c.client != nil {
+		return c.client, nil
+	}
+
+	ctx := context.TODO()
```
shanduur, line 50: Should we pass context here instead of creating detached one?

smira: we don't have a context in the interface (this is part of a common interface), so I can't really pass it here (without changing the interface)

PR #12942, opened 2026-03-10
fix: prevent stale discovered volumes reads

smira:
This pulls in a fix https://github.com/siderolabs/go-blockdevice/pull/148


internal/integration/api/volumes.go
```diff
@@ -92,7 +92,7 @@ func (suite *VolumesSuite) testDiscoveredVolumes(node string) {
 			Names: []string{"xfs"},
 		},
 		"EPHEMERAL": {
-			Names: []string{"xfs", ""},
```
smira, line 95: this empty string I removed was a workaround for the issue we are seeing - sometimes the EPHEMERAL was not discovered after being just provisioned

frezbo: this is cool, was an annoying stuff

PR #12976, opened 2026-03-16
fix: introduce pull attempt stall detection for image pull 

smira:
Also unlock not found (404) retries for installer image pulls.

### DIFF
--- a/internal/pkg/containers/image/stallguard/stallguard_test.go
+++ b/internal/pkg/containers/image/stallguard/stallguard_test.go
@@ -1,0 +1,114 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package stallguard_test
+
+import (
+	"net"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


ipuustin:
The issue this PR attempts to solve is this:

1. During a network installation, Talos autoconfigures a network interface using DHCP, getting an IP address
2. Talos starts downloading the installer image
3. Talos configures the network interface based on the downloaded MachineConfiguration and the IP address is changed
4. The download loses the source address, leading to TCP retries and eventually to a timeout

A more robust fix would be to first configure networking to the MC level and only afterwards start the download, but having a better retry system will most likely fix the issue after a timeout.

internal/pkg/containers/image/stallguard/stallguard_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	client, server := net.Pipe()
+
+	t.Cleanup(func() {
+		assert.NoError(t, client.Close())
+		assert.NoError(t, server.Close())
+	})
+
+	const (
+		idleTimeout = 100 * time.Millisecond
```
frezbo, line 51: i know these are very low timeout, but should we just wrap this in synctest? so time.Sleep is not actual time? 🤔

frezbo (approved):

internal/pkg/containers/image/stallguard/stallguard_test.go
smira: this is network read deadline, this is Linux thing, I don't think synctest can do this.

PR #12999, opened 2026-03-18
fix(machined): clear stale bond ARP/NS targets on decode

nberlee:
# Pull Request

## What? (description)
Reset ARPIPTargets and NSIP6Targets at the start of BondMasterSpec.Decode.

## Why? (reasoning)
Without this, repeated decode calls on the same struct can retain old target entries after config removes them, which makes linkspec of the bond interface drift from current bond configuration.

Add a regression test that decodes a payload with targets, then decodes a payload without target attributes into the same struct and asserts both slices are empty.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/adapters/network/bond_master_spec_test.go
+++ b/internal/app/machined/pkg/adapters/network/bond_master_spec_test.go
@@ -5,6 +5,7 @@
 package network_test
 
 import (
+	"net/netip"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/adapters/network/bond_master_spec_test.go
```diff
@@ -31,3 +32,31 @@ func TestBondMasterSpec(t *testing.T) {
...
+	}
+
+	initialEncoded, err := networkadapter.BondMasterSpec(&initial).Encode()
+	require.NoError(t, err)
+
+	cleared := network.BondMasterSpec{
+		Mode: nethelpers.BondModeActiveBackup,
+	}
+
+	clearedEncoded, err := networkadapter.BondMasterSpec(&cleared).Encode()
```
smira, line 50: thanks for the fix, but I wonder if there is same problem for Status, and I also wonder where do we decode twice in the same spec? (In general, we should clear the spec if we do this)

nberlee: [WriterModify](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/network/link_status.go#L227) in link_status.go  passes the existing persisted resource into the callback, so status.BondMasterSpec retains values from the previous reconcile cycle. If an attribute is absent in a later netlink message, old values silently persist and therefor accumulate.

In `link_spec.go` this isn't an issue since [existingBond is a fresh local variable](https://github.com/siderolabs/talos/blob/main/internal/app/machined/pkg/controllers/network/link_spec.go#L431) each time.

Updated the fix to reset the entire spec

smira: thank you for the fix!

PR #13010, opened 2026-03-20
feat: update Kubernetes to 1.36.0-beta.0

smira:
Update to the latest available release.


go.mod
```diff
@@ -29,17 +29,17 @@ replace github.com/containerd/containerd/v2 => github.com/smira/containerd/v2 v2
 
 // Kubernetes dependencies sharing the same version.
 require (
-	k8s.io/api v0.35.2
```
smira, line 32: 0.36.0 modules are not compatible with SSA flux modules yet (I guess we will bump closer to K8s release when Flux catches up)

PR #13027, opened 2026-03-23
fix: unset rlimits for extension services

smira:
See https://github.com/siderolabs/talos/discussions/13012

The containerd's default OCI spec sets NOFILE rlimit to 1024, unset it to simply let machined defaults take over.

### DIFF
--- a/internal/app/machined/pkg/system/services/extension.go
+++ b/internal/app/machined/pkg/system/services/extension.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


WinterNis:
Hi 👋 

Is there a chance that this fix gets backported to `v1.12` or it will be `v1.13+` only ? :D

Thanks !

smira:
> Hi 👋
> 
> Is there a chance that this fix gets backported to `v1.12` or it will be `v1.13+` only ? :D
> 
> Thanks !

This feels like a change big enough not be backportable, so it's 1.13 only.

PR #13030, opened 2026-03-24
feat: add validation of detached LUKS header

Orzelius:
now luks headers are validated for tampering


smira:
@Orzelius as a nit - it's better to name the pull request as `feat:` or `fix:`, as it changes the way Talos works

Also PR title should reflect the idea of the change - "add validation of detached LUKS header", the fact that blockdevice library got updated is clear from the diff

Also we should link to the blockdevice PR getting pulled.

PR #13031, opened 2026-03-24
test: improve maintenance API provision tests

smira:
Add a test that covers all maintenance APIs in general.

Add a test for transition from SideroLink.

Also some small 🐛  fixes.


cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu.go
```diff
@@ -20,7 +20,6 @@ import (
 	"go.yaml.in/yaml/v4"
 
 	"github.com/siderolabs/talos/cmd/talosctl/cmd/mgmt/cluster/create/clusterops"
-	"github.com/siderolabs/talos/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/siderolinkbuilder"
```
smira, line 23: moved this to a package available for all code

internal/app/lifecycle/lifecycle.go
```diff
@@ -88,7 +89,7 @@ func (s *Service) Install(req *machine.LifecycleServiceInstallRequest, ss grpc.S
 		return status.Error(codes.InvalidArgument, fmt.Sprintf("invalid disk path: %v", err))
 	}
 
-	s.logger.Info("starting installation", zap.String("installer_image", installerImage), zap.String("disk", targetDisk))
+	log.Printf("starting installation: installer_image=%s, disk=%s\n", installerImage, targetDisk)
```
smira, line 92: pushed the log message to the console so we can see important transitions like installation

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -1347,6 +1347,10 @@ func k8slogs(ctx context.Context, req *machine.LogsRequest) (chunker.Chunker, io
 }
 
 func getContainerInspector(ctx context.Context, namespace string, driver common.ContainerDriver) (containers.Inspector, error) {
+	if namespace == "" {
```
smira, line 1350: small validation to prevent invalid requests from reaching the API

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -2551,23 +2557,19 @@ func tryLockUpgradeMutex(ctx context.Context, etcdClient *etcd.Client) (unlock f
...
-		UDP6:          req.L4Proto.Udp6,
-		UDPLite:       req.L4Proto.Udplite,
-		UDPLite6:      req.L4Proto.Udplite6,
-		Raw:           req.L4Proto.Raw,
-		Raw6:          req.L4Proto.Raw6,
-		PID:           req.Feature.Pid,
-		NoHostNetwork: !req.Netns.Hostnetwork,
-		AllNetNs:      req.Netns.Allnetns,
-		NetNsName:     req.Netns.Netns,
+		TCP:           req.GetL4Proto().GetTcp(),
```
smira, line 2559: fix for panic on nil pointer dereference

internal/app/machined/internal/server/v1alpha1/v1alpha1_time.go
```diff
@@ -60,6 +62,10 @@ func (r *TimeServer) Time(ctx context.Context, in *emptypb.Empty) (reply *timeap
 func (r *TimeServer) TimeCheck(ctx context.Context, in *timeapi.TimeRequest) (reply *timeapi.TimeResponse, err error) {
 	var t time.Time
 
+	if in.GetServer() == "" {
```
smira, line 65: small validation to prevent a cryptic error later

internal/app/machined/pkg/controllers/runtime/maintenance_service_inform.go
```diff
@@ -94,6 +94,11 @@ func (ctrl *MaintenanceServiceInformController) Run(ctx context.Context, r contr
 			return fmt.Errorf("failed to get maintenance service config: %w", err)
 		}
 
+		if cfg == nil {
```
smira, line 97: fix for panic in the controller if the maintenance config hasn't been established

internal/app/machined/pkg/runtime/sequencer.go
```diff
@@ -53,6 +53,7 @@ const (
 var sequenceTakeOver = map[Sequence]map[Sequence]struct{}{
 	SequenceInitialize: {
 		SequenceMaintenanceUpgrade: {},
+		SequenceReboot:             {},
```
smira, line 56: allow to Reboot while in Initialize (this is important to reboot in "maintenance" mode which is technically in initialize sequence)

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -245,11 +246,27 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 }
 
 // HealthFunc implements the HealthcheckedService interface.
-func (o *APID) HealthFunc(runtime.Runtime) health.Check {
+func (o *APID) HealthFunc(r runtime.Runtime) health.Check {
 	return func(ctx context.Context) error {
+		cfg, err := safe.ReaderGetByID[*runtimeres.APIServiceConfig](ctx, r.State().V1Alpha2().Resources(), runtimeres.APIServiceConfigID)
```
smira, line 251: if `apid` is now in maintenance over SideroLink, we should healthcheck over SideroLink IP, as it doesn't listen on localhost

pkg/machinery/go.mod
```diff
@@ -8,7 +8,7 @@ replace github.com/mdlayher/ethtool => github.com/siderolabs/ethtool v0.4.0-side
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/containerd/go-cni v1.1.13
-	github.com/cosi-project/runtime v1.14.0
+	github.com/cosi-project/runtime v1.14.1
```
smira, line 11: pull in a fix for COSI Watch API panic (nil pointer deref in protobuf)

internal/app/lifecycle/lifecycle.go
Orzelius: Is it not possible to make the s.logger log to console here instead? This loses the structured logging and log levels, or is it fine in this area?

internal/integration/provision/maintenance_siderolink.go
```diff
@@ -0,0 +1,442 @@
...
+
+	body, err := io.ReadAll(reader)
+	suite.Require().NoError(err)
+
+	bootID := strings.TrimSpace(string(body))
+
+	_, err = io.Copy(io.Discard, reader)
+	suite.Require().NoError(err)
+
+	suite.Require().NoError(reader.Close())
```
Orzelius, line 432: any particular reason for `suite.Require().NoError(reader.Close())` if there's the deferred close above?

smira: I think to verify that Close() is clear, and `defer` should handle early abort...

internal/app/lifecycle/lifecycle.go
smira: I left `s.logger` also for logs which are not printed to the console (they go to `talosctl logs machined`), but sometimes all you have to debug is the console log, and install/upgrade is an important transition that needs to pop up to the console. In fact, `log.Printf` also goes to `talosctl logs machined`.

`log.Printf` is a legacy (old Go stdlog) which we use throughout the code most of the time, and we don't have zap-based replacement for it (yet), but we could do it eventually.

Issue #13020, opened 2026-03-21
nocloud: on-link IPv4 route uses IPv6 source address in dual-stack config


maxpain:
## Bug Description

When using the `nocloud` platform with a netplan v2 `network-config` that has both IPv4 and IPv6 addresses on the same interface, the on-link IPv4 auxiliary route (to the gateway) is created with an IPv6 source address, causing `EINVAL` from the kernel.

## Environment

- Talos version: v1.12.6
- Platform: `nocloud` (CIDATA ISO)
- VMM: Cloud Hypervisor

## network-config (netplan v2)

```yaml
version: 2
ethernets:
  nic0:
    match:
      macaddress: "02:xx:xx:xx:xx:xx"
    dhcp4: false
    dhcp6: false
    addresses:
      - 2a0c:6ec0:b00:c:0:1::/96
      - 10.0.0.0/32
    routes:
      - to: 0.0.0.0/0
        via: 169.254.0.1
        on-link: true
      - to: ::/0
        via: fe80::1
        on-link: true
    nameservers:
      addresses:
        - 1.1.1.1
        - 2606:4700:4700::1111
```

## Error

```
[talos] controller failed {
  "controller": "network.RouteSpecController",
  "error": "error adding route: netlink receive: invalid argument, message {
    Family:2 DstLength:32 Dst:169.254.0.1
    Src:2a0c:6ec0:b00:c:0:1::
    ...
  }"
}
```

`Family:2` (AF_INET) route with an IPv6 `Src` — kernel rejects this.

## Root Cause

In `internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go`, the on-link route handling blindly uses `eth.Address[0]` as the source address:

```go
if route.OnLink && gw.Is4() {
    // This assumes an interface with multiple routes will never have multiple statically set ips.
    ipPrefix, err := netip.ParsePrefix(eth.Address[0])
    // ...
    routeSpec := network.RouteSpecSpec{
        Source: ipPrefix.Addr(),       // ← IPv6 address!
        Family: nethelpers.FamilyInet4, // ← IPv4 family
    }
}
```

When the IPv6 address is listed first in `addresses`, `eth.Address[0]` is IPv6, but the route family is IPv4 → invalid argument.

## Suggested Fix

Find the first address matching the gateway's address family instead of using `Address[0]`:

```go
if route.OnLink && gw.Is4() {
    var source netip.Addr
    for _, addr := range eth.Address {
[...]

PR #13040, opened 2026-03-25
fix: incorrect route source for on-link routes

Orzelius:
when processing on-link routes, the source address was incorrectly set to the first address of the interface. This caused issues when the interface had multiple addresses, as the source address may not have been valid for the route. The source address is now set to an empty string, which allows the kernel to automatically select the appropriate source address for the route.

fixes https://github.com/siderolabs/talos/issues/13020

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
@@ -101,7 +101,7 @@ type Ethernet struct {
 		OnLink bool   `yaml:"on-link,omitempty"`
 	} `yaml:"routes,omitempty"`
 	RoutingPolicy []struct { // TODO
-		From  string `yaml:"froom,omitempty"`
+		From  string `yaml:"from,omitempty"`
 		Table uint32 `yaml:"table,omitempty"`
 	} `yaml:"routing-policy,omitempty"`
 }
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/testdata/in-v2-cloud-init.yaml
```diff
@@ -6,13 +6,17 @@ network:
         macaddress: "00:20:6e:1f:f9:a8"
       dhcp4: true
       addresses:
-        - 192.168.14.2/24
         - 2001:1::1/64
+        - 192.168.14.2/24
```
Orzelius, line 9: Changed the order to test that it still behaves as expected

internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
```diff
@@ -101,7 +101,7 @@ type Ethernet struct {
 		OnLink bool   `yaml:"on-link,omitempty"`
 	} `yaml:"routes,omitempty"`
 	RoutingPolicy []struct { // TODO
-		From  string `yaml:"froom,omitempty"`
+		From  string `yaml:"from,omitempty"`
```
Orzelius, line 104: typo fix

smira:
I think it doesn't handle the onlink IPv6 routes as described in the ticket. 

I would probably add new data file as a separate test to make debugging easier.

Orzelius:
> I think it doesn't handle the onlink IPv6 routes as described in the ticket.
> 
> I would probably add new data file as a separate test to make debugging easier.

I've added the exact example from the issue and it seems to work fine

internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/testdata/in-v2-cloud-init.yaml
smira, line 9: I would have preferred to skip changing existing input files, as this serves as a quick regression that they are processed same (or different way). If you change them, it's hard to notice if the expected output changed in an unexpected way.

internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/metadata.go
```diff
@@ -734,16 +734,9 @@ func applyNetworkConfigV2Ethernet(name string, eth Ethernet, networkConfig *runt
 		networkConfig.Routes = append(networkConfig.Routes, routeSpec)
 
 		if route.OnLink && gw.Is4() {
```
smira: this code is wrong still - if the route is `on-link`, we should create an on-link route to the gateway as destination, but do it for both IPv4 & IPv6

internal/app/machined/pkg/runtime/v1alpha1/platform/nocloud/testdata/in-v2-cloud-init.yaml
Orzelius: yeah, makes sense

Orzelius:
third time's the charm!

smira (approved):

PR #13041, opened 2026-03-25
feat: add NTS support for NTP time sync

### LINKED ISSUE #8779: Use NTS for NTP time sync
See https://github.com/beevik/nts

Our default time.cloudflare.com supports NTS, probably should be optional.

eleboucher:
# Pull Request

## What? (description)

Adds optional NTS (RFC 8915) support using beevik/nts. Controlled by a useNTS bool on the TimeSyncConfig document — when enabled, all NTP queries go through authenticated TLS sessions instead of plain NTP.
Closes #8779
## Why? (reasoning)

Plain NTP is unauthenticated — a MITM can mess with a node's clock. That matters because a skewed clock can break TLS cert validation, log ordering, and distributed consensus. NTS adds TLS authentication to NTP queries so you know the time response actually came from the server you asked.

## Acceptance


pkg/machinery/config/types/network/time_sync.go
```diff
@@ -78,6 +78,12 @@ type NTPConfig struct {
 	//     Specifies time (NTP) servers to use for setting the system time.
 	//     Defaults to `time.cloudflare.com`.
 	Servers []string `yaml:"servers,omitempty"`
+	//   description: |
+	//     Enables NTS (Network Time Security) for NTP queries.
+	//     NTS provides authenticated and encrypted time synchronization using TLS.
+	//     When enabled, all NTP servers must support NTS (RFC 8915).
+	//     Defaults to `false`.
+	UseNTS bool `yaml:"useNTS,omitempty"`
```
smira:
```suggestion
	UseNTS *bool `yaml:"useNTS,omitempty"`
```

It should be a pointer in the config (to support correct merge). 

I think the default (not in this config document) should be to enable NTS with time.cloudflare.com.

This field should propagate to `TimeServerSpec`/`TimeServerStatus` via the controller(s), and be correctly merged.

Also I think we should validate that this value can bet set to `true` only if the Servers are hostnames.

eleboucher: Addressed your comments! thank you for the quick review

pkg/machinery/resources/network/timeserver_spec.go
```diff
@@ -28,6 +28,7 @@ const TimeServerID resource.ID = "timeservers"
 type TimeServerSpecSpec struct {
 	NTPServers  []string    `yaml:"timeServers" protobuf:"1"`
 	ConfigLayer ConfigLayer `yaml:"layer" protobuf:"2"`
+	UseNTS      *bool       `yaml:"useNTS,omitempty" protobuf:"3"`
```
smira: we don't need to use pointers here (it's important only in the config when it gets patched - merged)

```suggestion
	UseNTS      bool       `yaml:"useNTS,omitempty" protobuf:"3"`
```

pkg/machinery/config/config/network.go
```diff
@@ -125,6 +125,7 @@ type NetworkTimeSyncConfig interface {
 	Disabled() bool
 	Servers() []string
 	BootTimeout() time.Duration
+	UseNTS() *bool
```
smira: this should return final computed value without pointers

```suggestion
	UseNTS() bool
```

pkg/machinery/config/types/network/time_sync.go
```diff
@@ -189,3 +205,13 @@ func (s *TimeSyncConfigV1Alpha1) Servers() []string {
 
 	return nil
 }
+
+// UseNTS implements config.NetworkTimeSyncConfig interface.
+// Returns nil if not explicitly configured.
+func (s *TimeSyncConfigV1Alpha1) UseNTS() *bool {
+	if s.TimeNTP == nil {
```
smira:
```suggestion
	return pointer.SafeDeref(s.TimeNTP)
```

internal/app/machined/pkg/controllers/network/timeserver_merge.go
```diff
@@ -37,6 +37,10 @@ func NewTimeServerMergeController() controller.Controller {
 				} else {
 					// otherwise, replace the lists
 					final = *spec.TypedSpec()
+
+					if final.UseNTS == nil {
```
smira: I think we shouldn't pass NTS across layers - just keep the replace strategy

internal/app/machined/pkg/controllers/network/timeserver_merge.go
smira, line 36: here we should merge NTS, but I think enable if both specs require it 

final.UseNTS = final.UseNTS && spec.UseNTS

pkg/machinery/config/types/network/time_sync.go
dsseng: > Also I think we should validate that this value can bet set to `true` only if the Servers are hostnames.

Is there a strict requirement? It only seems to require TLS, and now TLS certificates for IP addresses are generally available e.g. from Let's Encrypt.

I haven't found any on public lists, but found no evidence they should not be allowed. Shouldn't be a big deal either way I guess, as most users are likely to make use of public services with DNS. WDYT?

pkg/machinery/config/types/network/time_sync.go
```diff
@@ -78,6 +81,12 @@ type NTPConfig struct {
 	//     Specifies time (NTP) servers to use for setting the system time.
 	//     Defaults to `time.cloudflare.com`.
 	Servers []string `yaml:"servers,omitempty"`
+	//   description: |
+	//     Enables NTS (Network Time Security) for NTP queries.
+	//     NTS provides authenticated and encrypted time synchronization using TLS.
+	//     When enabled, all NTP servers must be specified as hostnames (not IP addresses).
```
dsseng: And perhaps if enabled all the servers must also be NTS-capable, am I right?

internal/pkg/ntp/ntp.go
```diff
@@ -131,6 +140,20 @@ func (syncer *Syncer) SetTimeServers(timeServers []string) {
...
+		newSet := make(map[string]struct{}, len(timeServers))
+		for _, s := range timeServers {
+			newSet[s] = struct{}{}
+		}
+
+		for host := range syncer.ntsSessions {
+			if _, ok := newSet[host]; !ok {
+				delete(syncer.ntsSessions, host)
+			}
+		}
```
dsseng: Could this potentially be used for cleaner code?

```suggestion
		for host := range syncer.ntsSessions {
			if !slices.Contains(timeServers, host) {
				delete(syncer.ntsSessions, host)
			}
		}
```

smira:
As a general note - we will hold this PR until we get into 1.14 cycle closer to the end of this month, as it changes the PAI (configuration).

internal/pkg/ntp/ntp.go
eleboucher: sorry i'm still not used to all the slices functions even though they are here for a long time now

smira (approved):

smira:
thank you for your work, 🆒 , I'm going to make small adjustments and merge this

pkg/machinery/config/types/network/time_sync.go
smira: I think we can relax later if we want to, if there is a real timeserver with NTS and no hostname?

pkg/machinery/resources/network/timeserver_spec.go
```diff
@@ -27,6 +27,7 @@ const TimeServerID resource.ID = "timeservers"
 //gotagsrewrite:gen
 type TimeServerSpecSpec struct {
 	NTPServers  []string    `yaml:"timeServers" protobuf:"1"`
+	UseNTS      bool        `yaml:"useNTS" protobuf:"3"`
```
smira:
```suggestion
	UseNTS      bool        `yaml:"useNTS,omitempty" protobuf:"3"`
```

PR #13044, opened 2026-03-26
feat: add graceful power off support to QEMU VM launcher

utkuozdemir:
The QEMU VM launcher's /poweroff HTTP endpoint now accepts an optional grace-period query parameter (Go duration format, e.g. "5m"). When set, it sends an ACPI power button event via the QEMU monitor socket instead of immediately killing the process, allowing the guest OS to shut down cleanly. If the guest does not shut down within the grace period, the process is force-killed as a fallback.

Without the parameter, the behavior is unchanged (immediate kill).

### DIFF
--- a/pkg/provision/providers/vm/controller.go
+++ b/pkg/provision/providers/vm/controller.go
@@ -4,10 +4,13 @@
 
 package vm
 
+import "time"
+
 // Controller interface should be implemented by the VM to be controlled via the API.
 type Controller interface {
 	PowerOn() error
 	PowerOff() error
+	PowerOffWithGracePeriod(gracePeriod time.Duration) error
 	Reboot() error
 	PXEBootOnce() error
 	Status() Status
--- a/pkg/provision/providers/qemu/controller.go
+++ b/pkg/provision/providers/qemu/controller.go
@@ -6,6 +6,7 @@ package qemu
 
 import (
 	"sync"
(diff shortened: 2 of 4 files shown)


pkg/provision/providers/vm/controller.go
```diff
@@ -4,10 +4,12 @@
 
 package vm
 
+import "time"
+
 // Controller interface should be implemented by the VM to be controlled via the API.
 type Controller interface {
 	PowerOn() error
-	PowerOff() error
+	PowerOff(gracePeriod time.Duration) error
```
utkuozdemir: Or do we want to not change this API for compatibility?

Orzelius: I don't know about compatibility but the code looks fine to me 🍏

shanduur: Maybe we can have PowerOffWithTimout?

utkuozdemir: IIRC we did breaking changes in this package in the past - we don't worry about backward-compatibility like we do with `machinery`.

But I like `PowerOffWithTimout` - it'd keep existing code working.

utkuozdemir: done as `PowerOffWithGracePeriod`.

PR #13061, opened 2026-04-01
fix: drop talosctl install

shanduur:
It does not make sense without SideroLink, and will not work,
as it requires Admin role.


cmd/talosctl/cmd/talos/install.go
frezbo, line 1: what was using this then?

PR #13072, opened 2026-04-03
feat: add client-side Kubernetes node drain to reboot and upgrade commands

shanduur:
Add --drain and --drain-timeout flags to `talosctl reboot` (default off)
and `talosctl upgrade` (default on) that cordon and drain the Kubernetes
node before rebooting, then wait for Ready and uncordon after it comes
back. When --drain is enabled, --wait is forced to true.

### DIFF
--- a/cmd/talosctl/pkg/talos/nodedrain/nodedrain.go
+++ b/cmd/talosctl/pkg/talos/nodedrain/nodedrain.go
@@ -1,0 +1,199 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package nodedrain provides reusable Kubernetes node drain, cordon, and uncordon
+// operations for use by talosctl commands (upgrade, reboot).
+package nodedrain
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/safe"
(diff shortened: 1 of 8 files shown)


cmd/talosctl/pkg/talos/nodedrain/nodedrain.go
```diff
@@ -0,0 +1,199 @@
...
+//
+// The context must target a single node (via client.WithNode) because COSI
+// State/Get does not support one-to-many proxying.
+func GetKubernetesNodeName(ctx context.Context, c *client.Client) (string, error) {
+	nodenameRes, err := safe.StateGet[*k8s.Nodename](
+		ctx,
+		c.COSI,
+		resource.NewMetadata(k8s.NamespaceName, k8s.NodenameType, k8s.NodenameID, resource.VersionUndefined),
+	)
+	if err != nil {
```
smira, line 58: I wonder if (for the future proof) we should treat "not found" (no NodeName resource) as "skip drain".

More like a thought, it's an edge case, and users can skip drain via the flag.

smira:
@shanduur please add some size to this task!

cmd/talosctl/pkg/talos/nodedrain/nodedrain.go
shanduur: I don't want it to fail silently and prefer it to stay this way. It's better to inform user that the drain cannot complete, instead of performing the reboot without notification about drain failure.

smira: yep, this makes sense, I was thinking more of the case (which we don't have now), when e.g. there is no Kubernetes node associated... but let's make it explicit for now. 

We have also `--skip-node-registration` option for the kubelet.

PR #13075, opened 2026-04-04
fix: don't set xattrs while decompressing extensions

smira:
When decompressing extensions, we might not be able to set xattrs (e.g. running rootless), so instead of setting xattrs, save them in memory and push to mksquashfs as pseudo definitions.

### DIFF
--- a/internal/pkg/extensions/compress.go
+++ b/internal/pkg/extensions/compress.go
@@ -12,6 +12,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
+	"strings"
 
 	"github.com/siderolabs/talos/pkg/machinery/imager/quirks"
 )
@@ -47,7 +49,7 @@ func initramfsPaths(quirks quirks.Quirks) []string {
 //
 // Components which should be placed to the initramfs are moved to the initramfsPath.
 // Ucode components are moved into a separate designated location.
-func (ext *Extension) Compress(ctx context.Context, squashPath, initramfsPath string, quirks quirks.Quirks) (string, error) {
+func (ext *Extension) Compress(ctx context.Context, squashPath, initramfsPath string, quirks quirks.Quirks, xattrsMap map[string]string) (string, error) {
 	if err := ext.handleUcode(initramfsPath, quirks); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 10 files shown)


smira:
This is WIP (it doesn't work at the moment).

pkg/archiver/untar.go
```diff
@@ -87,9 +91,13 @@ func Untar(ctx context.Context, r io.Reader, rootPath string) error {
...
 		}
 
-		if hdr.PAXRecords["SCHILY.xattr.security.selinux"] != "" {
-			if err = xattr.LSet(path, "security.selinux", []byte(hdr.PAXRecords["SCHILY.xattr.security.selinux"])); err != nil {
-				return fmt.Errorf("error setting selinux xattr for %q: %w", path, err)
+		if hdr.PAXRecords[constants.TarPaxHeaderSELinux] != "" {
+			if xattrsMap != nil {
+				xattrsMap[path] = hdr.PAXRecords[constants.TarPaxHeaderSELinux]
+			} else {
+				if err = xattr.LSet(path, "security.selinux", []byte(hdr.PAXRecords[constants.TarPaxHeaderSELinux])); err != nil {
```
frezbo: do we have a case where er have to set this when untar-ing, I believe this code doesn't ever get called at runtime 🤔 ?

pkg/machinery/constants/constants.go
```diff
@@ -1329,6 +1329,9 @@ const (
 
 	// ImageLabelVerified is the label key for the verified image label.
 	ImageLabelVerified = "talos.dev/verified"
+
+	// TarPaxHeaderSELinux is the name of the PAX header for storing SELinux labels in tar archives.
+	TarPaxHeaderSELinux = "SCHILY.xattr.security.selinux"
```
Orzelius: It's not really specific to tar files, is it?


```suggestion
	// TarPaxHeaderSELinux is the name of the PAX header for storing SELinux labels.
	TarPaxHeaderSELinux = "SCHILY.xattr.security.selinux"
```

pkg/archiver/untar.go
smira: this is a shared code, used also by `talosctl cp`, so I didn't want to break it, but it's a question if it should or should not set the label in that path

internal/pkg/extensions/compress.go
```diff
@@ -70,12 +72,57 @@ func (ext *Extension) Compress(ctx context.Context, squashPath, initramfsPath st
...
+func (ext *Extension) xattrPseudoFlags(xattrsMap map[string]string) ([]string, error) {
+	if xattrsMap == nil {
+		return nil, nil
+	}
+
+	flags := []string{"-xattrs-exclude", ".*"} // exclude all xattrs by default
+
+	for path, xattrValue := range xattrsMap {
+		if strings.HasPrefix(path, ext.RootfsPath()) {
+			relativePath, err := filepath.Rel(ext.RootfsPath(), path)
```
smira, line 115: check that the file exists, it might have been moved by other operations

pkg/archiver/untar.go
smira: I looked further, it was called from another unused function, so I dropped it as well, so now the `xattrsMap` is always used

pkg/machinery/constants/constants.go
smira: thanks, updated

internal/pkg/extensions/compress.go
smira: done

smira:
> I'm wondering what were the steps you took for manual testing?

I disabled masking embedded config here: https://github.com/siderolabs/talos/blob/d697f5538a7a624a1ac7bafdfebc67dd9418c434/internal/app/machined/pkg/controllers/config/acquire.go#L159-L162

Built a UKI with embedded config, booted from it (disabling bootloader), and did `talosctl ls -l /usr/local/etc/config.yaml` to see the labels. If masking is enabled (default), I will not see anything.

Issue #12952, opened 2026-03-12
feat: support http network probes


mottetm:
### Description

Historically, it has been difficult to isolate network configuration issues when deploying new clusters. The TCPProbeConfig provides a clean solution to highlight such issues and we love it. Unfortunately, in environments where egress is routed through an HTTP proxy, it cannot provide the same level of information — because the probe is not proxy-aware, the connectivity test will fail even when the target is reachable through the proxy, producing a false negative in the console.

### Proposed Solution

Add a separate HTTPProbeConfig that accepts an HTTP endpoint as target and succeeds if the server responds. Unlike TCPProbeConfig, it should honour the machine's proxy configuration, making it usable in proxy-gated environments.

smira:
I think this feature does make sense, I will add it to the epic, but this will not make it into Talos 1.13 most likely, as we're less than one week from going beta.

pranav767:
Working on this issue, Hope this will be straightforward to what is already done as `TCPProbeConfig` , following the same model as [tcp_probe](https://github.com/siderolabs/talos/blob/main/pkg/machinery/config/types/network/tcp_probe.go)

Plan: 

- Create a new `HTTPProbeConfigV1Alpha1` config document that issues an HTTP GET  to a URL and succeeds with any response i.e. working network.
- Update resource layer : Extend `ProbeSpecSpec` to include the new HTTP probe, update `network.proto` with the new fields, and regenerate the protobuf and deep-copy files.
- Implement the Controller Logic: Wire up the new configuration in `probe_config.go` and implement the `probeHTTP()` runner logic. Crucially, this will use `httpdefaults.PatchTransport` to ensure the HTTP client natively honors the machine's proxy environment variables.

smira:
Done via #12336

PR #13082, opened 2026-04-06
feat: add HTTP network probe support

pranav767:
Add support for HTTP/HTTPS connectivity probes alongside the existing TCP probes.
Resolves issue : https://github.com/siderolabs/talos/issues/12952

## What? (description)

Adds `HTTPProbeConfig` alongside the existing `TCPProbeConfig`. It sends an HTTP GET to a configured URL and succeeds if the server responds with any status code. Proxy configuration is honoured automatically.

## Why? (reasoning)

TCP probes only verify a port is reachable. HTTP probes can verify the full request path including proxies, TLS, and the application itself. This is especially useful in proxy-gated environments where a TCP probe would succeed (the proxy port is open) even if the actual target is unreachable.

## Acceptance


pkg/machinery/config/types/network/http_probe.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+// HTTPProbeConfigV1Alpha1 is a config document to configure network HTTP connectivity probes.
+//
+//	examples:
+//	  - value: exampleHTTPProbeConfigV1Alpha1()
+//	alias: HTTPProbeConfig
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/HTTPProbeConfig
+type HTTPProbeConfigV1Alpha1 struct {
```
shanduur, line 51: Should we have a field optional `maxRedirects`, that allows following redirects, 0 meaning no redirects (default), and allowing to set it to higher value?

shanduur:
As a general note - we will hold this PR until we get into 1.14 cycle closer to the end of this month, as it changes the PAI (configuration).

pkg/machinery/config/types/network/http_probe.go
pranav767: How would that help? Do we need to think about following redirects?
Are you thinking of the multiple proxies case, one sitting behind another :thinking:

shanduur: Right, on a second thought, we don't need this. It will break locality of probes, and might introduce unexpected behavior. Especially that we ignore the response codes.

smira:
As a general note - let's hold this until 1.13.0 is out (merge for 1.14.0-alpha.0)

shanduur (approved):

pkg/machinery/config/types/network/http_probe.go
```diff
@@ -0,0 +1,163 @@
...
+	MetaName string `yaml:"name"`
+	//nolint:embeddedstructfieldcheck
+	CommonProbeConfig `yaml:",inline"`
+	//   description: |
+	//     HTTP or HTTPS URL to probe. The probe succeeds if the server responds with any HTTP status code.
+	//   examples:
+	//    - value: >
+	//       "https://example.com"
+	//   schemaRequired: true
+	HTTPEndpoint string `yaml:"url"`
```
smira: if that is URL, I think we should try to use proper type: https://github.com/siderolabs/talos/blob/837a9ed077156ad00a1d31e731cf396c466bf6f6/pkg/machinery/config/types/meta/url.go#L9-L13

internal/app/machined/pkg/controllers/network/internal/probe/probe.go
```diff
@@ -152,3 +157,32 @@ func (runner *Runner) probeTCP(ctx context.Context) error {
...
+
+	ctx, cancel := context.WithTimeout(ctx, runner.Spec.HTTP.Timeout)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, runner.Spec.HTTP.URL, nil)
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Do(req)
```
smira, line 180: I think we should check the status code (?)

shanduur: I think we need e.g. CEL expression to validate the subset of response to give the control to the user?

smira: not sure... I wonder if we should do standard >= 200 && <= 299?

smira: or accept 3xx as well

smira: but hard to say, on other hand it's just a probe, and we can document that it's just connectivity check, and status is ignored

pranav767: I was thinking,if we are considering the HTTP probe purely from a connectivity perspective, what benefit does it actually give to validate the response code?

For example, if the probe hits a proxy or load balancer and gets a 502 Bad Gateway or a 401 Unauthorized, that still definitively proves the DNS, TCP, and outbound HTTP path are fully functional i.e. the network is healthy. If we try to validate status codes, we'll end up with false negatives where networks are marked as 'failing' just because an upstream service is having issues.

pranav767: Just saw you went with 
```
HTTP responses with status 200-399 are considered successful, while connection and transport errors are treated as failures.
```
👍

internal/app/machined/pkg/controllers/network/internal/probe/probe.go
```diff
@@ -121,11 +124,11 @@ func (runner *Runner) run(ctx context.Context, notifyCh chan<- Notification, log
 
 // probe runs a probe.
 func (runner *Runner) probe(ctx context.Context) error {
-	var zeroTCP network.TCPProbeSpec
-
 	switch {
-	case runner.Spec.TCP != zeroTCP:
+	case runner.Spec.TCP != (network.TCPProbeSpec{}):
```
smira, line 128: (not sure if it's better)

we also have a generic `value.IsZero()` helper function

internal/integration/api/probe-config.go
```diff
@@ -195,6 +199,51 @@ func (suite *ProbeConfigSuite) TestProbeStatus() {
...
+			asrt.Equal(5*time.Second, spec.TypedSpec().HTTP.Timeout)
+			asrt.Equal(networkres.ConfigMachineConfiguration, spec.TypedSpec().ConfigLayer)
+		},
+	)
+
+	// Update the probe config
+	probeConfig.ProbeFailureThreshold = 5
+	suite.PatchMachineConfig(nodeCtx, probeConfig)
+
+	rtestutils.AssertResource(nodeCtx, suite.T(), suite.Client.COSI, "http:"+probeURL,
```
smira, line 238: do we want to assert on ProbeStatus like we do for TCP?

btw, this probe should fail in airgapped mode as well.

smira (approved):

internal/app/machined/pkg/controllers/network/internal/probe/probe.go
shanduur: IMHO it's more convenient this way.

internal/integration/api/probe-config.go
shanduur: Fixed

PR #13122, opened 2026-04-15
test: allow more tests to run in FIPS strict mode

smira:
Remove the skip statements/rework the code to allow FIPS builds to do Wireguard by wrapping Wireguard operations into `fips140.WithoutEnforcement` blocks.

Using Wireguard (or not using it) is still a user's choice, but this allows tests to run in strict mode.

There might be more fixes required for FIPS strict, right now being blocked by Go issue with X25119 which is going to be backported to Go 1.26.3.

### DIFF
--- a/internal/app/machined/pkg/adapters/kubespan/identity.go
+++ b/internal/app/machined/pkg/adapters/kubespan/identity.go
@@ -12,9 +12,11 @@ import (
 
 	"github.com/mdlayher/netx/eui64"
 	"github.com/siderolabs/gen/value"
+	"go.uber.org/zap"
 	"go4.org/netipx"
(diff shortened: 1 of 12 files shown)


go.mod
```diff
@@ -336,7 +336,7 @@ require (
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
 	github.com/moby/docker-image-spec v1.3.1 // indirect
 	github.com/moby/locker v1.0.1 // indirect
-	github.com/moby/spdystream v0.5.0 // indirect
+	github.com/moby/spdystream v0.5.1 // indirect
```
smira, line 339: I bumped this due to upstream CVE which hasn't been fully published yet.

frezbo (approved):

PR #13123, opened 2026-04-15
fix: encoding for extra args fields in resources

smira:
This is a fixup for #12475

Split the protobuf id for extraArgs fields to use new value, so that we don't have unmarshal failure when using newer machinery with older Talos or vice versa.

Also pull in a fix https://github.com/siderolabs/go-talos-support/pull/15


api/resource/definitions/etcd/etcd.proto
```diff
@@ -17,9 +17,9 @@ message ConfigSpec {
   repeated string advertise_valid_subnets = 1;
   repeated string advertise_exclude_subnets = 2;
   string image = 3;
-  map<string, ArgValues> extra_args = 4;
```
oguzkilcan, line 20: Shouldn't you deprecate field number 4 using keyword `reserved` instead of outright deleting it?

frezbo (approved):

api/resource/definitions/etcd/etcd.proto
smira: I can't really do this, as it's generated from the Go struct.

It's a good point that we should support this annotation though.

oguzkilcan: I feel like doing it this way breaks backwards compatibility of protobuf.

api/resource/definitions/etcd/etcd.proto
```diff
@@ -17,9 +17,9 @@ message ConfigSpec {
...
   repeated string listen_valid_subnets = 5;
   repeated string listen_exclude_subnets = 6;
+  map<string, ArgValues> extra_args = 7;
 }
```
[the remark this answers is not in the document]
smira: good point, but we can't do that now (it's generated).

api/resource/definitions/etcd/etcd.proto
smira: `reserved` is a hint to prevent field reuse, it doesn't affect protobuf marshal/unmarshal in any way.

PR #13126, opened 2026-04-15
feat(ci): rework to schedule daily runs after a cron

frezbo:
This prevents us from building and pushing artifacts and replacing then for each run.


smira (approved):
🤞

PR #13143, opened 2026-04-16
feat: update dependencies

smira:
Sync pkgs/tools, update Go dependencies.


internal/integration/api/constants.go
```diff
@@ -9,7 +9,7 @@ package api
 const (
 	// NvidiaGPUOperatorChartVersion is the version of the NVIDA device plugin chart to use
 	// renovate: datasource=helm versioning=helm depName=gpu-operator registryUrl=https://helm.ngc.nvidia.com/nvidia
-	NvidiaGPUOperatorChartVersion = "v25.10.1"
+	NvidiaGPUOperatorChartVersion = "v26.3.0"
```
frezbo: this is a broken version, see https://github.com/NVIDIA/gpu-operator/pull/2261, there's no release with the fix yet

frezbo (approved):

internal/integration/api/constants.go
smira: oops, thank you, I will revert back

frezbo:
might also address this while we're here https://github.com/advisories/GHSA-77fj-vx54-gvh7/dependabot?query=user:siderolabs

shanduur (approved):

PR #13147, opened 2026-04-17
fix: return failed precondition on upgrade when not installed

smira:
This check was in maintenance Upgrade API for Talos <= 1.12, so keep it in the "normal" API as well.

It always makes sense - the upgrade would fail if Talos is not installed, but that failure in legacy Upgrade API is async and not reported properly back.

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -503,6 +503,10 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (*mach
 		return nil, err
 	}
 
+	if !s.Controller.Runtime().State().Machine().Installed() {
+		return nil, status.Errorf(codes.FailedPrecondition, "Talos is not installed")
+	}
+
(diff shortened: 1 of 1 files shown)


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -503,6 +503,10 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (*mach
 		return nil, err
 	}
 
+	if !s.Controller.Runtime().State().Machine().Installed() {
+		return nil, status.Errorf(codes.FailedPrecondition, "Talos is not installed")
```
frezbo, line 507: do we put the correct talos name here also 😅

frezbo (approved):

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
smira: that is exact copy-pasted from 1.12 :-D

Unix4ever (approved):

utkuozdemir (approved):

shanduur (approved):

PR #13149, opened 2026-04-17
fix: do not publish fatal event on expected sequence termination

utkuozdemir:
Sequences that intentionally terminate the machine (reboot, shutdown, upgrade, etc.) return a reboot signal to the sequencer when they complete. This was being treated as a sequence failure and published as a fatal event, which downstream consumers (including the machine stage tracking) interpreted as the machine entering reboot, even when it was actually shutting down.

The reboot signal is now recognized as expected termination and no longer emitted as a fatal event.

### DIFF
--- a/internal/app/machined/pkg/controllers/runtime/machine_status.go
+++ b/internal/app/machined/pkg/controllers/runtime/machine_status.go
@@ -422,8 +422,10 @@ func (ctrl *MachineStatusController) watchEvents() {
 						newStage = runtime.MachineStageRebooting
 					}
 				case machineapi.SequenceEvent_NOOP:
-					if event.Error != nil && event.Error.Code == common.Code_FATAL {
-						// fatal errors lead to reboot
+					shuttingDown := currentSequence == v1alpha1runtime.SequenceShutdown.String()
+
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
```diff
@@ -140,6 +140,13 @@ func (c *Controller) Run(ctx context.Context, seq runtime.Sequence, data any, se
 
 	err = c.run(ctx, seq, phases, data)
 	if err != nil {
+		if runtime.IsRebootError(err) {
+			// RebootError is the expected termination signal for reboot/shutdown/upgrade
```
smira: I would rather filter this in the MachineStatus controller, as this change might have some unintended consequences

smira (approved):

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_controller.go
utkuozdemir: 👍  just gonna add a test before merging.

PR #13165, opened 2026-04-20
fix: propagate route table down to the resource

### LINKED ISSUE #13153: LinkConfig doesn't correctly propagate table property to RouteConfig resource
### Description
I'd like to be able to configure fully custom routing for Wireguard, and the default routes break network connectivity via the ethernet interface when setting allowed IPs to `0.0.0.0/0`.

Being able to disable the default subnet route on all kinds of interfaces (e.g. `192.168.255.0/24` if the IP is `192.168.255.11/24`) would be great too, so I can create all routes manually in a different routing table.

I suggest a simple boolean field inside the `WireguardConfig` and `LinkConfig` like:
```yaml
kind: WireguardConfig
name: wg0
peers: [...]
disableDefaultRoutes: true
routes: [...]
```

smira:
Fixes #13153

### DIFF
--- a/internal/app/machined/pkg/controllers/network/route_config.go
+++ b/internal/app/machined/pkg/controllers/network/route_config.go
@@ -328,7 +328,7 @@ func (ctrl *RouteConfigController) processMachineConfig(linkConfigs []cfg.Networ
 				route.Family = nethelpers.FamilyInet4
 			}
 
-			route.Table = nethelpers.TableMain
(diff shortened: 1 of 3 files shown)


frezbo (approved):

silenium-dev:
Is this fix going to be part of 1.13.0?

smira:
yes,

image

PR #13181, opened 2026-04-21
fix(talosctl): always uncordon after drain on reboot/upgrade error

shanduur:
fix(talosctl): ensure uncordon runs after reboot/upgrade errors

Use defer blocks and error joining to guarantee uncordon cleanup
runs regardless of reboot/upgrade success or failure. Prevents nodes
from staying cordoned when operations fail.

Also added gRPC keepalive params to prevent timeout issues during
long operations.

### DIFF
--- a/cmd/talosctl/cmd/talos/reboot.go
+++ b/cmd/talosctl/cmd/talos/reboot.go
@@ -57,7 +57,7 @@ var rebootCmd = &cobra.Command{
 	},
 }
 
-func rebootRun(opts []client.RebootMode) error {
+func rebootRun(opts []client.RebootMode) (retErr error) {
 	rep := reporter.New(
 		reporter.WithOutputMode(rebootCmdFlags.progress.Value()),
 	)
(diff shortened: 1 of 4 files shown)


smira (approved):

cmd/talosctl/cmd/talos/reboot.go
```diff
@@ -78,13 +78,15 @@ func rebootRun(opts []client.RebootMode) error {
 		return err
 	}
 
-	if err := rebootInternal(rebootCmdFlags.wait, rebootCmdFlags.debug, rebootCmdFlags.timeout, rep, opts...); err != nil {
-		return err
-	}
+	defer func() {
+		if uncordonErr := WithClientAndNodes(func(ctx context.Context, c *client.Client, _ []string) error {
```
smira, line 82: I guess if `ctx` is canceled at this point, we should not uncordon (e.g. if someone pressed ^C)?

cmd/talosctl/pkg/talos/action/node.go
```diff
@@ -216,12 +218,37 @@ func (a *nodeTracker) runPostCheckWithRetry(preActionBootID string) error {
...
 
+//nolint:gocyclo
 func (a *nodeTracker) handleEvents(eventCh chan client.EventResult, actorID string) error {
+	// On slow-boot platforms (e.g. arm64 kexec-disabled), MachineReadyEventFn
+	// (RUNNING + etcd healthy) may never arrive before the outer timeout.
+	// Poll postCheckFn (boot ID changed + node reachable) as a fallback so we
+	// don't block forever waiting for an event that requires full cluster health.
+	var tickerCh <-chan time.Time
+
+	if a.preActionBootID != "" && a.tracker.postCheckFn != nil {
```
smira: pinging @utkuozdemir  on these changes as he know this code path the best ;)

smira (approved):

cmd/talosctl/cmd/talos/reboot.go
shanduur: This is handled internally by `WithClientAndNodes` (context handling is added in `(c *Args) WithClientNoNodes(...)`), right?

smira: yep, there is a bigger problem, but this change is good

utkuozdemir (approved):

PR #13194, opened 2026-04-23
feat: add customHeaders support to cluster.network.cni

mcanevet:
## What? (description)

Add a `customHeaders` field to `CNIConfig` (`cluster.network.cni`) that allows specifying arbitrary HTTP headers to be sent when fetching CNI manifest URLs.

The new field is a `map[string]string` and is only valid when `cni.name` is `custom`. It is rejected by validation for `flannel` and `none` CNI types.

## Why? (reasoning)

Some CNI manifests are hosted behind authenticated endpoints — private registries, internal artifact servers, or GitHub repos requiring a bearer token. Without this field there is no way to fetch such manifests during cluster bootstrap, forcing users to mirror manifests to a public location.

The headers are forwarded to `ExtraManifest.ExtraHeaders` for every CNI URL, so the existing manifest-fetching path handles them without any new download logic.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane.go
@@ -394,11 +394,14 @@ func NewControlPlaneExtraManifestsController() *ControlPlaneExtraManifestsContro
 
 				spec := k8s.ExtraManifestsConfigSpec{}
 
+				cniHeaders := cfgProvider.Cluster().Network().CNI().CustomHeaders()
+
 				for _, url := range cfgProvider.Cluster().Network().CNI().URLs() {
 					spec.ExtraManifests = append(spec.ExtraManifests, k8s.ExtraManifest{
-						Name:     url,
-						URL:      url,
-						Priority: "05", // push CNI to the top
+						Name:         url,
+						URL:          url,
+						Priority:     "05", // push CNI to the top
+						ExtraHeaders: cniHeaders,
 					})
 				}
 
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_test.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_test.go
@@ -923,6 +923,65 @@ func (suite *K8sControlPlaneSuite) TestReconcileKubeProxyModeLegacy() {
 	)
 }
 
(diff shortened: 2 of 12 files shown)


smira:
Sorry, but I don't want this change. 

First of all, it mutates old config we want to deprecate.

Second, you can use `extraManifests`, which already supports headers, and it's equivalent.

Third, downloading manifests isn't a good idea unless the URL is locked to a specific tag/sha (immutable).

Fourth, it's best done outside of Talos, as it's a cluster operation, not a machine operation.

mcanevet:
Thanks for the detailed review. All four points make sense and I agree, happy to close this out.

Two clarifying questions for my own understanding, if you don't mind:

On point 3 (immutable URLs): is this a general caveat that also applies to `extraManifests`, or is it more specifically an argument against the CNI fetching path? Put differently: is `extraManifests` with a tag/SHA-pinned URL still considered an acceptable bootstrap pattern, or is the concern broader than just URL pinning?

On point 4 (cluster vs. machine): the practical issue I keep hitting is that Talos's bootstrap healthcheck stays unhealthy until nodes report Ready, which requires a CNI to be in place. So CNI effectively has to be provisioned during bootstrap rather than applied afterwards by an external tool. In that light, what's the recommended pattern when the CNI manifests live behind auth? Is `extraManifests` (with pinned URLs) the intended bootstrap shim, or is there a different path you'd point users toward?

Not pushing back, just want to make sure I apply the right pattern next time.

smira:
> On point 3 (immutable URLs): is this a general caveat that also applies to extraManifests, or is it more specifically an argument against the CNI fetching path? Put differently: is extraManifests with a tag/SHA-pinned URL still considered an acceptable bootstrap pattern, or is the concern broader than just URL pinning?

Yes, but promoting URL-based manifests seems like not an exactly great flow.

> On point 4 (cluster vs. machine): the practical issue I keep hitting is that Talos's bootstrap healthcheck stays unhealthy until nodes report Ready, which requires a CNI to be in place. So CNI effectively has to be provisioned during bootstrap rather than applied afterwards by an external tool. In that light, what's the recommended pattern when the CNI manifests live behind auth? Is extraManifests (with pinned URLs) the intended bootstrap shim, or is there a different path you'd point users toward?

This is fixed in Talos 1.13+, node readiness healthcheck is skipped if CNI is none.

PR #13211, opened 2026-04-25
fix: volume mount permissions

### LINKED ISSUE #11946: Research Talos mounts
Make sure we have mounts that are `rw` always set with `noexec,nosuid,nodev` and we are only executing binaries from mounts are readonly

Identify if we also need `noatime`

EPHEMERAL/`/opt/cni` etc would be exemptions.

This is more of a generic routine check to make sure we haven't missed anything

frezbo:
Make Talos volumes mount options more stricter.

Fixes: #11946


internal/app/machined/pkg/controllers/block/mount.go
```diff
@@ -625,6 +626,10 @@ func (ctrl *MountController) handleDiskMountOperation(
 			opts = append(opts, mount.WithSecure())
 		}
 
+		if mountRequest.TypedSpec().MountOptionNoExec {
+			opts = append(opts, mount.WithMountAttributes(unix.MOUNT_ATTR_NOEXEC))
+		}
+
 		if mountRequest.TypedSpec().ReadOnly {
 			opts = append(opts, mount.WithReadOnly())
 		}
```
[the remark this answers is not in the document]
frezbo: this is okay, we need to set it unconditionally

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -1456,11 +1456,13 @@ func MountEphemeralPartition(runtime.Sequence, any) (runtime.TaskExecutionFunc,
...
-			vol, _ := cfg.Volumes().ByName(constants.EphemeralPartitionLabel)
-			mountRequest.TypedSpec().Secure = vol.Mount().Secure()
-		}
+		// /var hosts state only — containerd extracts layers there but
+		// container exec resolves through the overlay rootfs at
+		// /run/containerd/.../rootfs which is a separate mount with its
+		// own flags. Per-mount-flag rule means flags on /var don't
+		// propagate to overlay or bind children. Secure applies the full
+		// nosuid+nodev+noexec triplet.
+		mountRequest.TypedSpec().Secure = true
```
smira: This looks like a change that can't be reverted with a config, should we instead flip the default in the config?

frezbo: ah yes, that works, i'll flip the default, but also i don't see why `/var` should allow running binaries at all, for uservolumes this still needs explicitly setting, though I would propose it to be even strict

internal/pkg/selinux/policy/selinux/services/udev.cil
```diff
@@ -60,9 +60,13 @@
 
 (allow kernel_t modprobe_exec_t (file (execute)))
 (allow kernel_t udev_t (process (all))) ; including transition
+; modprobe_exec_t lives on rootfs squashfs (NOSUID)
```
smira: do we need to add it for any process that init execs? why nosuid here? is the executable file marked as suid?

smira (commented):
I really like where it goes, but probably it needs a bit more thinking on compatibility

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
smira: My point is to make it `secure = true` by default, but if anyone hits a wall, they can revert that via config

internal/pkg/selinux/policy/selinux/services/udev.cil
frezbo: it's a different selinux thing, it considers mounts with nosuid a different https://github.com/SELinuxProject/selinux-notebook/blob/9888b84b8e76c18c6141812d7ffd9b3e2fad6f9e/src/object_classes_permissions.md?plain=1#L1928-L1930

smira: what I mean - should the rule be more generic? As we have many binaries being executed from `nosuid`? What about e.g. extension services?

frezbo: this is explicitly for squashfs from what i understood, @DmitriyMV for visibility

smira: I guess @dsseng is the correct tag ;)

But my point still holds - we have many more binaries in the squashfs, e.g. what if the CNI calls `iptables`?

smira: Or should we allow domain transitions from `nosuid` mount in general?

frezbo: so these are all allowing init -> process transitions, and for the udev one allowing udev -> modprobe

These are all known ones we're allowing explicitly

CNI calling iptables is not part of these rules at all

@dsseng can correct me if my understanding is wrong

dsseng: CNI shouldn't be the case, because it should execute without transitions IIRC

However, given we anyway have explicit permissions to transition when executing certain executables, these rules should be redundant and we should be able to allow nosuid transitions by default - they should still need transition permission as always, and we granted it before

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/system_volumes.go
```diff
@@ -133,6 +133,7 @@ func GetEphemeralVolumeTransformer(inContainer bool) volumeConfigTransformer {
 						UID:                 0,
 						GID:                 0,
 						ProjectQuotaSupport: cfg.Machine().Features().DiskQuotaSupportEnabled(),
+						Secure:              extraVolumeConfig.Mount().Secure(),
```
smira, line 137: we probably need a unit-test here

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/system_volumes.go
```diff
@@ -133,6 +133,7 @@ func GetEphemeralVolumeTransformer(inContainer bool) volumeConfigTransformer {
...
 					}).
 					WithLocator(labelVolumeMatch(constants.EphemeralPartitionLabel)).
```
smira, line 139: should we do `secure` also for `STATE` ?

smira (approved):
let's add some points to the PR!

internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/system_volumes.go
frezbo: ah yes

PR #13213, opened 2026-04-27
fix: prepend 'service-account-issuer' instead of appending

### LINKED ISSUE #13210: GCP Workload Identity Federation no longer works with Talos 1.13.0-rc.0
### Description

After upgrading to Talos 1.13.0-rc.0, the GCP Workload Identity Federation (and extremely likely all other similar solutions in Azure, AWS and so on) stopped working because of an `iss` claim mismatch. 

This seems to be due to `service-account-issuer` api-server config now being appended instead of prepended - which changes the `iss` claim for the issued service account tokens. This is because api-server will use first value as an issuer value, while all remaining values are simply additional issuers that are allowed during its own validation - that has no effect on validation done by federated services (`iss` is always a single value, so that's expected).
The PR that introduced this change: https://github.com/siderolabs/talos/pull/12607

This can be reproduced by creating a new Talos cluster and following guide here https://docs.siderolabs.com/kubernetes-guides/advanced-guides/gcp-workload-identity which no longer works, with an error that can be seen in the logs below.

### Logs

```
root@workload-identity-test:~# gcloud storage ls gs://<redacted>
ERROR: (gcloud.storage.ls) There was a problem refreshing your current auth tokens: ('Error code invalid_grant: `issuer` claim in OIDC discovery document does not match the issuer specified in the request.', '{"error":"invalid_grant","error_description":"`issuer` claim in OIDC discovery document does not match the issuer specified in the request."}')
```

Claims of an JWT token projected into container:

```
{
  "aud": [
    "//iam.googleapis.com/projects/<redacted>/locations/global/workloadIdentityPools/<redacted>/providers/<redacted>"
  ],
  "exp": 1777141419,
  "iat": 1777137819,
  "iss": "https://<redacted-controlplane-address>",
  "jti": "f07a7ca9-e7d0-48c9-9f84-115bae4bb9e9",
  "kubernetes.io": {
    "namespace": "default",
    "node": {
      "name": "build-controlplane-1",
      "uid": "6c5c8390-37de-481b-82b1-0a0134293e86"
    },
    "pod": {
      "name": "workload-identity-test",
      "uid": "59cc2f98-91a2-4961-a420-e6bd8a485756"
    },
    "serviceaccount": {
      "name": "workload-identity",
      "uid": "c7a93136-d01e-41f8-b2a6-469b72ef3cc8"
    }
  },
  "nbf": 1777137819,
  "sub": "system:serviceaccount:default:workload-identity"
}
```

### Environment

```
Client:
        Tag:         v1.12.7
        SHA:         91c63991eb669602e379ff653ecc20212834390e

        Built:       
        Go version:  go1.26.2-X:nodwarf5
        OS/Arch:     linux/amd64
Server:
        NODE:        172.18.0.3
        Tag:         v1.13.0-rc.0
        SHA:         1f949d9a
        Built:       
        Go version:  go1.26.2
        OS/Arch:     linux/arm64
        Enabled:     RBAC
```

- Kubernetes version: [`kubectl version`]

```
Client Version: v1.35.4
Kustomize Version: v5.7.1
Server Version: v1.35.3

```
- Platform: Hetzner Cloud

shanduur:
Fixes #13210


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -1176,6 +1176,13 @@ type ControlPlaneConfig struct {
 	//     pattern: "^https://"
 	Endpoint *Endpoint `yaml:"endpoint"`
 	//   description: |
+	//    EndpointIsDefaultIssuer indicates whether the control plane endpoint should be used as the default issuer for service account tokens.
+	//    If true, the API server will use the control plane endpoint as the issuer in the service account token issuer discovery document and in the `iss` claim of issued tokens.
+	//    All values passed to `service-account-issuer` will be appended to the issuer list, and the control plane endpoint will be used as the default if no `service-account-issuer` is specified.
+	//    If false, the control plane endpoint will be used as the default issuer only if no `service-account-issuer` values are specified.
+	//    All values passed to `service-account-issuer` will be prepended before the control plane endpoint in the issuer list.
+	EndpointIsDefaultIssuer bool `yaml:"endpointIsDefaultIssuer,omitempty"`
```
smira, line 1184: This should be `*bool` to make merge work correctly.

And I guess it should be `true` by default.

shanduur: I think it should be `false` to preserve the behavior from before 1.13.

PR #13253, opened 2026-04-30
feat: implement etcd encryption config (kube-apiserver)

### LINKED ISSUE #8702: Ability to rotate secret encryption keys used in apiserver
### Description
While trying to implement the procedure as described at
https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#rotating-a-decryption-key

(due to accidental publishing of a full machineconfig to a public github repository)

I was blocked by the inability to modify the encryptionconfig.yaml file to have the older keys in it to transition to a newly generated key.

This seems like something that would be nice to be supported in talosctl, or at least, the ability to add multiple encryption secrets to the machineconfig for the purposes of key rotation. 

I've attempted the procedure manually by trying to specify a custom encryptionconfig file that has been inserted via files[], but apparently that argument is protected by talos and I am unable to override it.

### LINKED ISSUE #10899: ETCD encryption configuration

smira:
This replaces existing fixed field for etcd encryption with a completely flexbile configuraiton which exactly matches upstream kube-apiserver configuration.

The default machine configuration generated still retains previous defaults.

New configuration allows:

* rotating etcd encryption secrets
* implementing any encryption policies (e.g. encrypting configmaps).

Fixes #10899

Fixes #8702


pkg/machinery/config/generate/secrets/bundle.go
```diff
@@ -181,13 +204,52 @@ func NewBundleFromConfig(clock Clock, c config.Config) *Bundle {
...
+					for _, key := range provider.Secretbox.Keys {
+						secrets.SecretboxEncryptionSecret = key.Secret
+
+						break outerLoop
+					}
+				}
+
+				// if it's not the first provider, ignore it
+				break
+			}
```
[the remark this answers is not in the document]
smira: This is correct flow - we only accept the first configured provider for "secrets" resource.

pkg/machinery/config/generate/kubernetes.go
```diff
@@ -0,0 +1,43 @@
...
+									"name":   "key1",
+									"secret": in.Options.SecretsBundle.Secrets.SecretboxEncryptionSecret,
+								},
+							},
+						},
+					},
+				},
+				"resources": []any{
+					"secrets",
+				},
```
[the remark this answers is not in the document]
smira: This was done on purpose to prevent reading identity (not encrypted) secrets ever, this should be a better default.

pkg/machinery/config/contract.go
```diff
@@ -214,3 +214,8 @@ func (contract *VersionContract) KubeSpanMultidocConfig() bool {
 func (contract *VersionContract) HostDNSMultidocConfig() bool {
 	return contract.Greater(TalosVersion1_13)
 }
+
+// MultidocKubernetesConfigSupported returns true if version of Talos should use multi-doc Kubernetes config.
+func (contract *VersionContract) MultidocKubernetesConfigSupported() bool {
```
shanduur, line 219: Is the `Supported` suffix necessary? Other contract checks do not follow this pattern.

pkg/machinery/config/contract.go
```diff
@@ -214,3 +214,8 @@ func (contract *VersionContract) KubeSpanMultidocConfig() bool {
...
+	return contract.Greater(TalosVersion1_13)
```
shanduur, line 220: ?

```suggestion
	return contract.Greater(TalosVersion1_14)
```

shanduur (approved):

pkg/machinery/config/contract.go
smira: I followed this pattern: https://github.com/siderolabs/talos/blob/631a1bc5e1230b14ddc018e7f5116b2d0ebe8821/pkg/machinery/config/contract.go#L184-L185

pkg/machinery/config/contract.go
smira: `contract = (major, minor)` (without patch), so this is correct, it's basically `>= 1.14`

PR #13261, opened 2026-05-01
fix: deadlock in the makefs ext4 with populated source

### LINKED ISSUE #13256: imager: Ext4() deadlocks when source-directory tar exceeds pipe buffer (pkg/makefs/ext4.go)
`pkg/makefs/ext4.go::Ext4` keeps `pr` open while waiting for the tar producer goroutine. When the consumer subprocess exits before tar finishes writing, tar blocks on `pipe_write` and `Ext4` never returns. Surfaces in `imager` populating IMAGECACHE from a non-trivial `imageCache`.

Triggered by `output.kind: image` with `input.imageCache.{ociPath,tarballPath}` set: `pkg/imager/out.go:354` (`Install` with `ModeImage`) calls `makefs.Ext4` + `WithSourceDirectory` for the IMAGECACHE partition. ISO output uses a different filesystem path and is unaffected.

Introduced by `70e67787d`; `v1.12` unaffected. `origin/main` at `dedb7a96c` still affected.

### End-user CLI

```bash
talosctl images default | talosctl images cache-create \
    --image-cache-path /tmp/cache --images=-

docker run --rm --privileged \
    -v /dev:/dev -v /tmp/out:/out -v /tmp/cache:/cache:ro \
    ghcr.io/siderolabs/imager:v1.13.0 \
    metal --image-cache /cache --output /out
```

Hangs indefinitely after `creating ext4 filesystem on .../IMAGECACHE.img with args: [... -d - ...]`. Exits only when killed externally.

### Go test

Drop into `pkg/makefs/ext4_largesource_test.go`:

```go
package makefs_test

import (
	"context"
	"os"
	"path/filepath"
	"testing"
	"time"

	"github.com/stretchr/testify/require"

	"github.com/siderolabs/talos/pkg/makefs"
)

func TestExt4ReproducerLargeInput(t *testing.T) {
	t.Setenv("PATH", "/usr/bin:/bin:/usr/sbin:/sbin")

	tmpDir := t.TempDir()
	tempFile := filepath.Join(tmpDir, "ext4.img")

	f, err := os.Create(tempFile)
	require.NoError(t, err)
	require.NoError(t, f.Truncate(512*1024*1024))
	require.NoError(t, f.Close())

	sourceDirectory := filepath.Join(tmpDir, "source")
	require.NoError(t, os.MkdirAll(sourceDirectory, 0o755))
	require.NoError(t, os.WriteFile(
		filepath.Join(sourceDirectory, "data.bin"),
		make([]byte, 1*1024*1024),
		0o644,
	))

	done := make(chan error, 1)
	go func() {
		done <- makefs.Ext4(context.Background(), tempFile,
			makefs.WithLabel("REPROTEST"),
			makefs.WithForce(true),
			makefs.WithSourceDirectory(sourceDirectory),
		)
	}()

	select {
	case err := <-done:
		t.Logf("Ext4 returned: %v", err)
	case <-time.After(20 * time.Second):
		t.Fatalf("DEADLOCK: makefs.Ext4 did not return within 20s")
	}
}
```

`go test -v -timeout 1m -run TestExt4ReproducerLargeInput ./pkg/makefs/` → fails at 20 s.

smira:
Close the pipe on error/abort.

Fixes #13256

### DIFF
--- a/.kres.yaml
+++ b/.kres.yaml
@@ -1636,6 +1636,17 @@ spec:
             VIA_MAINTENANCE_MODE: true
             WITH_CONFIG_PATCH: '@hack/test/patches/image-cache.yaml:@hack/test/patches/image-verification.yaml'
             WITH_ISO: true
+        - name: e2e-image-cache-disk
+          command: e2e-qemu
+          withSudo: true
+          environment:
+            GITHUB_STEP_NAME: ${{ github.job}}-e2e-image-cache-disk
+            IMAGE_REGISTRY: registry.dev.siderolabs.io
+            REGISTRY_MIRROR_FLAGS: no
+            SHORT_INTEGRATION_TEST: yes
+            VIA_MAINTENANCE_MODE: true
+            USE_DISK_IMAGE: true
+            WITH_CONFIG_PATCH: '@hack/test/patches/image-cache.yaml:@hack/test/patches/image-verification.yaml'
         - name: e2e-image-cache-encrypted
           command: e2e-qemu
           withSudo: true
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2026-04-29T12:24:24Z by kres 980313d.
+# Generated on 2026-05-01T15:48:11Z by kres 1762ab2.
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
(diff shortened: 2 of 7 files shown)


smira:
NB: I want to add an integration test for image cache + disk image

utkuozdemir (approved):

.kres.yaml
```diff
@@ -1636,6 +1636,17 @@ spec:
             VIA_MAINTENANCE_MODE: true
             WITH_CONFIG_PATCH: '@hack/test/patches/image-cache.yaml:@hack/test/patches/image-verification.yaml'
             WITH_ISO: true
+        - name: e2e-image-cache-disk
```
frezbo, line 1639: this seems like we should make these parallel, maybe for later

smira: yes, we could make the cache-create "shared" step, but requires some work

PR #13262, opened 2026-05-01
fix(ci): zfs test

frezbo:
Expect the zvol file to be present eventually.


smira (approved):
promote to extensions?

smira (approved):

PR #13265, opened 2026-05-01
fix: re-enable kexec on arm64

smira:
The upstream kernel bug should be fixed now.

### DIFF
--- a/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu.go
+++ b/cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/qemu.go
@@ -157,9 +157,8 @@ func (m *Qemu) AddExtraGenOps() error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


smira:
on QEMU/darwin/arm64 this fails to bring 2nd CPU online

smira:
> on QEMU/darwin/arm64 this fails to bring 2nd CPU online

no problem on AWS though:

```
[    0.002555] CPU1: Booted secondary processor 0x0000000001 [0x413fd0c1]
[    0.002679] smp: Brought up 1 node, 2 CPUs
[    0.002691] SMP: Total of 2 processors activated.
[    0.002693] CPU: All CPU(s) started at EL1
```

frezbo (approved):

shanduur (approved):

smira:
> on QEMU/darwin/arm64 this fails to bring 2nd CPU online

I mean after kexec, but not sure it's a regression. The VM works still fine, just with 1 CPU. `with-bootloader=false` disables kexec, so it shouldn't be a real development issue.

PR #13267, opened 2026-05-01
feat: add host-namespace debug profile

frezbo:
Add a PROFILE_HOST_NS debug profile that runs `talosctl debug` in a fork of
the host mount namespace instead of a containerd task. The session gets an
overlay root (host `/` as lower + disk-backed upper) with a Nix tools image's
`/nix` overlaid in, so host binaries (zpool, mkfs.xfs, …) run at their native
paths without nsenter while Nix-provided tools are on PATH.
The `talosctl debug --host-ns` flag is gated behind the `sidero.debug` build
tag (WITH_DEBUG=true); the machine API profile is always available.

### DIFF
--- a/internal/app/debug/debug.go
+++ b/internal/app/debug/debug.go
@@ -51,7 +51,9 @@ func (s *Service) ContainerRun(srv grpc.BidiStreamingServer[machine.DebugContain
 		return status.Errorf(codes.InvalidArgument, "expected debug container spec")
 	}
 
-	if spec.GetProfile() != machine.DebugContainerRunRequestSpec_PROFILE_PRIVILEGED {
+	switch spec.GetProfile() { //nolint:exhaustive
+	case machine.DebugContainerRunRequestSpec_PROFILE_PRIVILEGED, machine.DebugContainerRunRequestSpec_PROFILE_HOST_NS:
+	default:
 		return status.Errorf(codes.InvalidArgument, "unsupported debug container profile: %s", spec.GetProfile())
 	}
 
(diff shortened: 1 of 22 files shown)


pkg/machinery/constants/constants.go
```diff
@@ -758,6 +758,12 @@ const (
 	// CgroupSystemDebug is the cgroup name for debug processes.
 	CgroupSystemDebug = CgroupSystem + "/debug"
 
+	// DebugHostNsImage is the default image used for PROFILE_HOST_NS debug sessions.
+	// It provides a Nix environment whose store paths are self-contained and do not
+	// conflict with the Talos host's library paths when bind-mounted into the forked
+	// mount namespace.
+	DebugHostNsImage = "docker.io/nixos/nix:latest"
```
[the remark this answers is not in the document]
frezbo: this is fine for now

smira (commented):
this is crazy amount of code, should we put it under `debug` build tag? 

I wonder if we should hide the flag, and hide the server implementation under a build tag, so it's not available in general? this is powerful, but imho too much crazy to understand all consequences for a production run

frezbo:
> this is crazy amount of code, should we put it under `debug` build tag?
> 
> I wonder if we should hide the flag, and hide the server implementation under a build tag, so it's not available in general? this is powerful, but imho too much crazy to understand all consequences for a production run

After discussion, let's put this under a debug flag and only enable for CI tests or along similar lines

hack/release.toml
```diff
@@ -434,6 +434,19 @@ Use CSI drivers instead — a CSI node plugin performs the attach/mount itself i
 unaffected by the sandbox. For iSCSI, `kubernetes-csi/csi-driver-iscsi` (or `democratic-csi`) consumes a
 target the same way. All in-tree (non-CSI) volume plugins are deprecated for the kubelet and support for them
 may be removed in a later release.
+"""
+
+    [notes.debugHostNS]
+        title = "Host Namespace Debugging"
+        description = """\
+Debug builds of `talosctl` gain a `talosctl debug --host-ns` flag that runs the debug session in the host mount
```
smira: I don't think we should put it to the release notes (at least now), as it's only available in the debug builds, and we don't distribute them

internal/app/debug/debug.go
```diff
@@ -51,8 +51,12 @@ func (s *Service) ContainerRun(srv grpc.BidiStreamingServer[machine.DebugContain
 		return status.Errorf(codes.InvalidArgument, "expected debug container spec")
 	}
 
-	if spec.GetProfile() != machine.DebugContainerRunRequestSpec_PROFILE_PRIVILEGED {
-		return status.Errorf(codes.InvalidArgument, "unsupported debug container profile: %s", spec.GetProfile())
+	switch spec.GetProfile() {
+	case machine.DebugContainerRunRequestSpec_PROFILE_PRIVILEGED, machine.DebugContainerRunRequestSpec_PROFILE_HOST_NS:
+	case machine.DebugContainerRunRequestSpec_PROFILE_UNSPECIFIED:
```
smira:
```suggestion
	default:
```

I don't think we need a specific entry for zero value and any other value - just use InvalidArgument always

pkg/machinery/constants/constants.go
```diff
@@ -769,6 +769,18 @@ const (
...
```
smira, line 776: I guess this image should be in the airgapped set of images instead of alpine now (have you tested with airgapped?)

smira (approved):

Issue #9383, opened 2024-09-26
host DNS forwarding to kube should use IPv4/IPv6 endpoints based on pod subnets


smira:

dsseng:
Resolved by #13274. Does not seem to require any splitting based on subnets, as pods will use available addresses to reach the DNS resolver and should skip ones they got no routes to

PR #13274, opened 2026-05-02
feat: support conditional start of IPv6 dns servers

### LINKED ISSUE #9384: add default IPv6 DNS resolvers

dsseng:
This PR does those things:
- Raise IPv6 listener on link-local address for dns (both TCP and UDP).
- Update kubelet's `resolv.conf` IPv4/IPv6 endpoints.

Closes #9384

Link: https://github.com/siderolabs/talos/pull/9596

### DIFF
--- a/api/resource/definitions/network/network.proto
+++ b/api/resource/definitions/network/network.proto
@@ -259,6 +259,7 @@ message HostDNSConfigSpec {
   repeated common.NetIPPort listen_addresses = 2;
   common.NetIP service_host_dns_address = 3;
   bool resolve_member_names = 4;
+  common.NetIP service_host_dns_address_v6 = 5;
 }
 
 // HostnameSpecSpec describes node hostname.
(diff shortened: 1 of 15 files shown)


dsseng:
Rebased #9596

I tested this with experimental IPv6-only cluster (in Docker), and this change ~~fixes HostDNS and makes it work with single-stack IPv6~~ makes DNS traffic use HostDNS

Before: CoreDNS directly connects to upstream, not using HostDNS, skipping HostDNS cache
After: CoreDNS connects to HostDNS, which then serves requests via its cache or contacts upstreams

dsseng:
This issue if likely more relevant: https://github.com/siderolabs/talos/issues/9383

#9384 is more about 'Talos should list Google and Cloudflare DNS IPv6 addresses together with IPv4 in the defaults'

smira (approved):

binary-zero-one:
I apologize for revisiting this discussion. The description specifies "Raise IPv6 listener on **link-local address**", yet a ULA address is being utilized, just like in the initial PR.
Are there specific architectural or networking constraints necessitating the use of a ULA, or would it be feasible to employ an L-L address instead? E.g, `fe80:5461:6c6f:7320:4f53:2044:4e53:1` or perhaps a configuration with zeroed subnet bits `fe80::5461:6c6f:7344:4e53`, which results in "TalosDNS" being encoded within the interface bits.

smira:
We use ULA as it has to be routable from Kubernetes pod (coredns) down to the host.

PR #13277, opened 2026-05-04
feat: run depmod with verification on rootfs build 

smira:
Verify that the set of modules in the rootfs with the kernel combined
doesn't have any unresolved symbols.

Uses https://github.com/siderolabs/pkgs/pull/1530

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -648,6 +648,7 @@ FROM tools AS depmod-amd64
 WORKDIR /staging
 COPY hack/modules-amd64.txt .
 COPY --from=pkg-kernel-amd64 /usr/lib/modules usr/lib/modules
+COPY --from=pkg-kernel-amd64 /boot/System.map /staging
 RUN <<EOF
 set -euo pipefail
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -648,14 +648,25 @@ FROM tools AS depmod-amd64
...
 RUN <<EOF
 set -euo pipefail
 
 KERNEL_VERSION=$(ls usr/lib/modules)
 
 xargs -a modules-amd64.txt -I {} install -D usr/lib/modules/${KERNEL_VERSION}/{} /build/usr/lib/modules/${KERNEL_VERSION}/{}
 
-depmod -b /build/usr ${KERNEL_VERSION}
+# check if the output of the command is empty, as depmod doesn't fail and just prints a warning
+DEPMOD_OUTPUT=$(depmod -b /build/usr -F /staging/System.map --errsyms -w ${KERNEL_VERSION} 2>&1)
```
frezbo, line 660: what about extensions? or should imager do this verification too since it also rebuilds?

smira: This would require us to preserve System.map, but I think we could try to validate during the extension build... This is not trivial

smira:
For the backporters: backport just the 2nd commit (a fix).

frezbo (approved):

shanduur (approved):

Issue #13257, opened 2026-04-30
v1.13.0: bond0 unreachable after migrating from legacy machine.network.interfaces to multi-doc BondConfig (ICMP Redirect Host loop, kernel local table missing entry)


Deenyoro:
### Description

After migrating a worker node's network config from the legacy `machine.network.interfaces[]` form to the new multi-doc form (`LinkConfig` + `BondConfig` + `RoutingRuleConfig`), the node comes back Ready after the reboot, then drops to NotReady about 60 seconds later. The bond0 management IP becomes locally undeliverable: incoming pings to it are answered with `ICMP Redirect Host(New nexthop: <self IP>)`, which is what you see when the kernel's `local` routing table (table 255) is missing the auto-populated `local <IP> dev bond0` entry that should appear when an address is bound to an interface.

The userspace side of Talos thinks everything is fine. `talosctl get addresses` shows the bond IPs bound. `talosctl get routingrules` shows the new custom rule installed correctly. `talosctl get routes` shows both the bond0 default in main and the bond2 default in table 100. The kernel just never gets the local-route side of the address binding, so the box is unreachable on its own management IP within about a minute of the apply finishing.

Reverting the VM snapshot back to the legacy inline config restores normal behavior on the same hardware and same Talos binary, so this looks like a config-shape / reconciliation issue specific to converting an in-place node from legacy `machine.network.interfaces[]` to multi-doc.

The intent of the apply was to add a bond2 carrying a second uplink with a default route in routing table 100 plus a source-based rule, and move bond0/bond1 out of legacy `interfaces[]` into multi-doc `BondConfig` so the whole worker network section is canonical v1.13 (the docs note that `BondConfig` conflicts with `LinkConfig` of the same name, so I wanted to avoid any chance of mixed-mode on the same node).

### Pre-apply config (legacy inline, on a node that had been running this for months)

```yaml
machine:
  network:
    interfaces:
      - interface: eth0
        dhcp: false
      - interface: eth1
        dhcp: false
      - interface: eth2
        dhcp: false
      - interface: bond0
        bond:
          interfaces: [eth0]
          mode: active-backup
        mtu: 9000
        addresses: [10.0.0.10/24]
        routes:
          - network: 0.0.0.0/0
            gateway: 10.0.0.1
            metric: 25
      - interface: bond1
        bond:
          interfaces: [eth1]
          mode: active-backup
        mtu: 9000
        addresses: [192.168.10.10/24]
[...]

PR #13281, opened 2026-05-05
fix(network): skip reserved routing rule priorities

shanduur:
Prevent config controller from generating specs for rules at kernel-reserved
priorities (local, main, default), stopping the spec controller from ever
tearing them down. Add protocol check (RTPROT_STATIC only) to ensure Talos
only manages its own rules, never touching kernel or third-party agent rules.

Part of #13257

### DIFF
--- a/internal/app/machined/pkg/controllers/network/routing_rule_config.go
+++ b/internal/app/machined/pkg/controllers/network/routing_rule_config.go
@@ -17,10 +17,18 @@ import (
 	"go.uber.org/zap"
 
 	cfg "github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/constants"
 	"github.com/siderolabs/talos/pkg/machinery/nethelpers"
 	"github.com/siderolabs/talos/pkg/machinery/resources/config"
 	"github.com/siderolabs/talos/pkg/machinery/resources/network"
 )
+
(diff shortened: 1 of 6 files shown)


frezbo (approved):

internal/app/machined/pkg/controllers/network/routing_rule_config.go
```diff
@@ -109,13 +118,24 @@ func (ctrl *RoutingRuleConfigController) apply(ctx context.Context, r controller
...
 ) []network.RoutingRuleSpecSpec {
 	rules := make([]network.RoutingRuleSpecSpec, 0, len(ruleConfigs))
 
 	for _, ruleCfg := range ruleConfigs {
 		var rule network.RoutingRuleSpecSpec
 
+		priority := ruleCfg.Priority()
+
+		if _, reserved := reservedConfigRulePriorities[priority]; reserved {
+			logger.Warn("skipping routing rule at reserved priority",
```
smira: do we need more validation at the config level? or how did the rule slip through into this layer?

smira (approved):

internal/app/machined/pkg/controllers/network/routing_rule_config.go
shanduur: I’m not sure how could this happen - I tried reproducing that with manifests provided in the original issue, but I’m not able to do that.

PR #13286, opened 2026-05-05
feat: add golangci-lint fmt target

shanduur:
Adds Docker target + Makefile target for running golangci-lint
formatter across codebase. Auto-fixes lint issues instead of just
reporting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1268,6 +1268,31 @@ RUN --mount=type=cache,target=/.cache,id=talos/.cache,sharing=locked go tool git
 WORKDIR /src
 RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool github.com/siderolabs/importvet/cmd/importvet github.com/siderolabs/talos/...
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1268,6 +1268,15 @@ RUN --mount=type=cache,target=/.cache,id=talos/.cache,sharing=locked go tool git
 WORKDIR /src
 RUN --mount=type=cache,target=/.cache,id=talos/.cache go tool github.com/siderolabs/importvet/cmd/importvet github.com/siderolabs/talos/...
 
+# clean golangci-lint fmt output
+# exclude files populated by the `base` stage (gendata build args, generated os-release)
+# so running this target doesn't dirty the source tree with build-time values.
+FROM scratch AS lint-go-fmt
```
smira: does `run` format? does it not fail? (I never tried it)

shanduur: Fixed.

smira (approved):
you can do `go tool golangci-lint ...` in local mode as well, as it exactly matches the one used inside Docker, but this is good as well

PR #13287, opened 2026-05-05
feat: update containerd to 2.3.0

smira:
Drop the fork of Go package, as we don't need it anymore.

Adjust other calls, as containerd package now has another reachable DCE violation via text/template.

Also update Linux to 6.18.26 (via pkgs).


smira:
Config migration issue as a follow-up https://github.com/siderolabs/talos/issues/13290

Right now containerd automatically migrates v2 config to v3, so migration is not an immediate blocker.

internal/app/machined/pkg/system/runner/containerd/opts.go
```diff
@@ -41,3 +48,87 @@ func WithCustomSeccompProfile(override func(*specs.LinuxSeccomp)) oci.SpecOpts {
...
+	}
+
+	return results
+}
+
+// WithImageConfigStripped is a reduced version of WithImageConfig which skips WithUser call.
+//
+// The function oci.WithUser has issues with deadcode elimination in containerd >= 2.3.0 due to
+// a call chain that includes a call to text/template.
+func WithImageConfigStripped(image oci.Image) oci.SpecOpts {
```
smira, line 86: This is a reduced copy from containerd OCI package.

I had to do that, as oci.WithImageConfig reaches into oci.WithUser which tries to mount container filesystem which in turn imports `text/template` and disables dead code elimination.

internal/app/machined/pkg/system/services/apid.go
```diff
@@ -238,7 +238,7 @@ func (o *APID) Runner(r runtime.Runtime) (runner.Runner, error) {
 				oci.WithMounts(mounts),
 				oci.WithRootFSPath(filepath.Join(constants.SystemLibexecPath, o.ID(r))),
 				oci.WithRootFSReadonly(),
-				oci.WithUser(fmt.Sprintf("%d:%d", constants.ApidUserID, constants.ApidUserID)),
+				oci.WithUIDGID(constants.ApidUserID, constants.ApidUserID),
```
smira, line 241: same DCE handling, we never needed `WithUser` as we're using static UID/GID, so use proper (simple) wrapper

go.mod
```diff
@@ -19,11 +19,6 @@ replace (
 	golang.zx2c4.com/wireguard/wgctrl => github.com/siderolabs/wgctrl-go v0.0.0-20251029173431-c4fd5f6a4e72
 )
 
-// deadcode elimination fix replacement: https://github.com/siderolabs/talos/issues/11296
-// upstream PR: https://github.com/containerd/containerd/pull/12175
```
smira, line 23: https://github.com/containerd/containerd/pull/12175 PR got merged, so I dropped the fork while updating to 2.3.0, but another DCE violation led to a series of fixes here

internal/app/machined/pkg/system/runner/containerd/opts.go
```diff
@@ -41,3 +48,87 @@ func WithCustomSeccompProfile(override func(*specs.LinuxSeccomp)) oci.SpecOpts {
...
+	return func(ctx context.Context, client oci.Client, c *containers.Container, s *specs.Spec) error {
+		ic, err := image.Config(ctx)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
smira: this is part of the change, the actual behavior is tested indirectly via integration test (they would break if this is broken)

internal/app/machined/pkg/system/runner/containerd/opts.go
```diff
@@ -41,3 +48,87 @@ func WithCustomSeccompProfile(override func(*specs.LinuxSeccomp)) oci.SpecOpts {
...
+			s.Process.Args = slices.Concat(config.Entrypoint, config.Cmd)
+
+			cwd := config.WorkingDir
+			if cwd == "" {
+				cwd = "/"
+			}
+
+			s.Process.Cwd = cwd
+		}
+
```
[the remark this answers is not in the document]
smira: Talos is Linux-only, so this is fine.

frezbo (approved):

PR #13292, opened 2026-05-06
fix: make dnsd retry listening

dsseng:
IPv6 address takes longer time to become available, so without this
retry dnsd was not available via IPv6 address of the host

### DIFF
--- a/pkg/provision/providers/vm/dnsd.go
+++ b/pkg/provision/providers/vm/dnsd.go
@@ -5,12 +5,16 @@
 package vm
 
 import (
+	"errors"
 	"fmt"
 	"log/slog"
 	"net"
 	"os"
+	"syscall"
+	"time"
 
 	"github.com/miekg/dns"
+	"github.com/siderolabs/go-retry/retry"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/siderolabs/talos/pkg/provision"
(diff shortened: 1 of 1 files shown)


shanduur (approved):

pkg/provision/providers/vm/dnsd.go
```diff
@@ -43,9 +47,24 @@ func DNSd(ips []net.IP, resolvConfPath string) error {
 			Handler: dns.HandlerFunc(forwardHandler(mode, log, config)),
 		}
 
-		log.Info("starting DNS forwarder server")
+		// IPv6 address takes longer to become ready, and therefore might need a retry
+		return retry.Constant(
```
smira, line 51: how is this address assigned?

for IPv4, it's on the bridge? with IPv6 I guess it should be also on the bridge?

smira (approved):

pkg/provision/providers/vm/dnsd.go
dsseng: I assume the kernel runs DAD before making the address available, and that's why we initially cannot listen, since the kernel hasn't yet confirmed the address is non-conflicting and may be used. I have not observed it in details though

dsseng: Yes. This address is global in kernel terms, and not marked with `nodad`, so the kernel checks for duplicates first

dsseng: image

Yes, so the host first runs DAD, so the static address is initialized as `tentative`, and only later is assigned

PR #13295, opened 2026-05-07
feat: add LVM status controller

shanduur:
Add resource types for physical volumes, volume groups, and logical
volumes to track LVM inventory and state. Provides read-only discovery
through the resource API for visibility into existing LVM setup.

### Manual Testing

Creating LVM objects:

```bash
cat <<'EOF' | sudo -E _out/talosctl-linux-amd64 -n 172.20.0.6 debug docker.io/library/alpine:latest --args /bin/sh
set -e

NSENTER="nsenter --target 1 --mount --uts --ipc --net --pid"

echo "==> Creating PVs"
$NSENTER pvcreate -y /dev/vdb /dev/vdc /dev/vdd

echo "==> Creating VG"
$NSENTER vgcreate vgdata /dev/vdb /dev/vdc /dev/vdd

echo "==> Creating LV"
$NSENTER lvcreate -n lvdata -l 100%FREE vgdata

echo "==> Listing LVM state"
$NSENTER pvs
$NSENTER vgs
$NSENTER lvs
EOF
```

Removing LVM objects:

```bash
cat <<'EOF' | sudo -E _out/talosctl-linux-amd64 -n 172.20.0.6 debug docker.io/library/alpine:latest --args /bin/sh
set -e

NSENTER="nsenter --target 1 --mount --uts --ipc --net --pid"

echo "==> Removing logical volume"
$NSENTER lvremove -y /dev/vgdata/lvdata || true

echo "==> Removing volume group"
$NSENTER vgremove -y vgdata || true

echo "==> Wiping physical volumes"
$NSENTER pvremove -y /dev/vdb /dev/vdc /dev/vdd || true

echo "==> Done"
$NSENTER pvs || true
$NSENTER vgs || true
$NSENTER lvs || true
EOF
```


pkg/machinery/resources/storage/storage.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"github.com/cosi-project/runtime/pkg/resource"
+
+	"github.com/siderolabs/talos/pkg/machinery/resources/v1alpha1"
+)
+
+//go:generate go tool github.com/siderolabs/deep-copy -type LVMLogicalVolumeStatusSpec -type LVMPhysicalVolumeStatusSpec -type LVMVolumeGroupStatusSpec -header-file ../../../../hack/boilerplate.txt -o deep_copy.generated.go .
+
+// NamespaceName contains storage resources.
+const NamespaceName resource.Namespace = v1alpha1.NamespaceName
```
smira: we could probably actually use proper namespace name, it was a bug when I originally started `block` that I put it to the wrong namespace, but it was too late when I discovered it

internal/app/machined/pkg/controllers/storage/storage.go
```diff
@@ -0,0 +1,13 @@
...
+
+// Package storage provides the controllers related to LVM and storage virtualization.
+package storage
+
+import "time"
+
+// pollInterval is how often the LVM status controllers re-scan the system in
+// the absence of input events. LVM has no inotify-style change feed, so a
+// poll is the only way to catch attribute changes (e.g. grow operations).
+const pollInterval = 30 * time.Second
```
smira: how bad it is to poll on interval? would it e.g. wake up a sleeping disk just to check LVM metadata?

internal/app/machined/pkg/controllers/storage/lvm_physical_volume_status.go
```diff
@@ -0,0 +1,126 @@
...
+func (ctrl *LVMPhysicalVolumeStatusController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	ticker := time.NewTicker(pollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		case <-ticker.C:
```
smira: can a PV change without ever writing to blockdevice? we have a flow which goes into DiscoverdVolumes and it notifies on any process opening a blockdevice for write, we could feed it here? 

(happy to discuss more)

internal/app/machined/pkg/controllers/storage/storage.go
shanduur: `pvs -a` enumerates every block device so it will open each and read sector 0 / PV label region to check for LVM signature. On spun-down HDDs that wakes the  disk every poll - so could probably significantly increase the wear of the device.

shanduur: I'll fix that.

internal/app/machined/pkg/controllers/storage/lvm_physical_volume_status.go
shanduur: PV not, but some kernel DM states can - e.g. active/inactive, suspended, open-count, kernel major/minor - lives in device-mapper, not on disk. So thing slike `lvchange -ay/-an`, `dmsetup
  suspend`, `mount/umount` can flip these w/o touching PV metadata. Reflected in `lv_active`/`lv_suspended`/`lv_device_open` columns.

smira: but I guess that would trigger DM blockdevice changes? We also listen for these (we listen for all blockdevice events).

pkg/machinery/resources/storage/storage.go
shanduur: Fixed.

internal/app/machined/pkg/controllers/storage/storage.go
shanduur: Dropped the polling.

internal/app/machined/pkg/controllers/storage/lvm_refresh_trigger.go
```diff
@@ -0,0 +1,80 @@
...
+// Run implements controller.Controller interface.
+func (ctrl *LVMRefreshTriggerController) Run(ctx context.Context, r controller.Runtime, _ *zap.Logger) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		}
+
+		if err := safe.WriterModify(
```
smira, line 168: I wonder if we should throttle this (the easiest way possible is to add something like this after this line):

This way we would coalesce multiple updates into a single event, as e.g. BlockDevice -> Disk -> DV will always propagate in a sequence

```go
select {
case <-time.After(time.Second): // or something bigger
case <-ctx.Done():
  return nil
}
```

smira (approved):

PR #13303, opened 2026-05-07
feat: revert update CoreDNS to 1.14.3

smira:
This reverts commit 2b6c06ef51fdd3c7ffcbf621eacbb37d798e5228.

The published image is linux/amd64 only.


smira:
See https://github.com/kubernetes/k8s.io/pull/9421

frezbo (approved):

PR #13304, opened 2026-05-07
feat: dockerfile improvements

frezbo:
Use better syntax and avoid generating un-necessary layers.

* Split out doc generation from `generate` so it doesn't run twice
* Use `--link` and `RUN --mount=from` syntax more
* Inline more commands and reduce COPY steps and use bind mounts with RUN
* Add lint stage for tools
* Parallelize lint stages
* Move `cloud-image-uploader` to tools
* Create `integration-test` target similar to talosctl, and update ci to use them


Makefile
```diff
@@ -568,14 +564,16 @@ unit-tests-race: ## Performs unit tests with race detection enabled.
...
+	@$(MAKE) local-$(INTEGRATION_TEST)-linux-arm64 DEST=$(ARTIFACTS) PLATFORM=linux/arm64 WITH_RACE=true PUSH=false
 
-$(ARTIFACTS)/$(INTEGRATION_TEST_DEFAULT_TARGET)-arm64:
-	@$(MAKE) local-$(INTEGRATION_TEST_DEFAULT_TARGET)-arm64 DEST=$(ARTIFACTS) PLATFORM=linux/arm64 WITH_RACE=true PUSH=false
+$(INTEGRATION_TEST)-darwin-arm64:
+	@$(MAKE) local-$(INTEGRATION_TEST)-darwin-arm64 DEST=$(ARTIFACTS) PUSH=false
 
-$(ARTIFACTS)/$(INTEGRATION_TEST):
-	@$(MAKE) local-$(INTEGRATION_TEST)-targetarch DEST=$(ARTIFACTS)
+$(INTEGRATION_TEST): $(INTEGRATION_TEST)-$(OPERATING_SYSTEM)-$(ARCH) ## Builds the integration test binary for the host OS/arch.
```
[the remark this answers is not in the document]
frezbo: this is fine, no real users of darwin/amd64 for now

api/docs/api.md
smira, line 1: did we not have it before?

smira: I think this is wrong path, should go to website/...

frezbo: nope, we were just keeping it as a stage, meaning make generate docs would always regenerate

smira (approved):

api/docs/api.md
frezbo: ah I see, I'll fix

frezbo: fixed

smira (approved):

PR #13308, opened 2026-05-08
fix: handle empty GCP operation errors

immanuwell:
## What?

Small guard in the GCP cloud image uploader. If Compute returns an operation with `error` set but no detailed `errors` entries, return a normal `gcp: operation failed` error instead of indexing into an empty slice. Also fixes the tiny `faild` typo, oops.

## Why?

Repro is pretty small: build a `compute.Operation` with `HTTPStatusCode: 200` and `Error: &compute.OperationError{}`. The old path tried `op.Error.Errors[0].Message` and panicked. New test covers this exact thing, so it should stay fixed.

## Acceptance

- [x] issue linked if applicable: none found
- [x] tests included
- [ ] conformance (`make conformance`) attempted locally; commit-body is fixed, but `gpg-identity` requires a Sidero Labs org signer
- [x] lint: `sudo make lint`
- [x] unit check: `cd hack/cloud-image-uploader && go test ./...`

Tiny patch, no big rewrite. just making the error path less spicy.

### DIFF
--- a/tools/cloud-image-uploader/gcp.go
+++ b/tools/cloud-image-uploader/gcp.go
@@ -186,12 +186,8 @@ func (u *GCPUploder) registerImage(arch string) error {
 			return fmt.Errorf("gcp: failed to get operation: %w", err)
 		}
 
-		if op.HTTPStatusCode != http.StatusOK {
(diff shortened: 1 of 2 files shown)


smira (approved):
thank you!

smira:
This got merged, GitHub is out of sync with itself.

PR #13309, opened 2026-05-08
feat: implement KernelModuleStatus

### LINKED ISSUE #13305: Expose modules.builtin as an API
For CVEs that affect built in kernel modules it's hard for customers to know what modules are included with talos unless they look at the pkgs code or read modules.builtin file. Reading the file isn't available from clusters managed by Omni so we should expose the data as an API similarly to how we do with `loadedkernelmodules`

It might be hard to indicate if a module has been patched (and what it has been patched for) because the kernel version may not change even if modules have been patched multiple times.

smira:
Resolves https://github.com/siderolabs/talos/issues/13305

- Introduces a new resource `KernelModuleStatuses.runtime.talos.dev`, which surfaces the state of all Linux kernel modules (both built-in and dynamically loaded).
- Retains the `LoadedKernelModule.runtime.talos.dev` resource for backward-compatibility.
- Reworks the reconcililiation trigger mechanism with a `kobject` watcher, in place of the previous one based on `filehash`.
    - Reuses the watcher that was previously only used [for watching block devices](https://github.com/immanuwell/talos/blob/574298ec11a9aaeb3375733cc539c308bbdc4984/internal/app/machined/pkg/controllers/block/internal/kobject/kobject.go). Now moved up in the source tree.
    - An event is emitted also for each built-in module as well, so on startup there's a burst of events that would trigger many reconciliations with not much benefit. To handle this gracefully, I've re-used the [RateLimitedTrigger](https://github.com/siderolabs/talos/blob/649a384a961c64c036dcc0b82ecbc102b951d2e2/internal/app/machined/pkg/controllers/network/watch/trigger.go#L26) from `internal/app/machined/pkg/controllers/network/watch` (now moved up in the source tree).
- Built-in modules are pre-loaded into the controller's state on startup. These never change throughout the controllers lifetime.

Co-authored-by: Andrey Smirnov <andrey.smirnov@siderolabs.com>, Maja Bojarska <maja.bojarska@siderolabs.com>


pkg/machinery/resources/runtime/loaded_kernel_module.go
```diff
@@ -45,7 +47,6 @@ type LoadedKernelModuleExtension struct{}
 func (LoadedKernelModuleExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 	return meta.ResourceDefinitionSpec{
 		Type:             LoadedKernelModuleType,
```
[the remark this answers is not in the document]
majabojarska: This seems fine to me, we now want the CLI alias to point to `KernelModuleStatus` by default. If somebody needs to retrieve `LoadedKernelModule` resources, they still can do it by using the actual resource's name:
```sh
talosctl get --nodes 172.20.0.2 loadedkernelmodule
``` 

or via the fully qualified name:
```sh
talosctl get --nodes 172.20.0.2 loadedkernelmodules.runtime.talos.dev 
```

or alternatively, through the Go lib.

@shanduur what's your opinion on this, since you wrote the original implementation?

pkg/machinery/resources/runtime/kernel_module_type.go
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package runtime
+
+//go:generate go tool github.com/dmarkham/enumer -type=KernelModuleType -linecomment -text
+
+// KernelModuleType represents whether a kernel module is built into the kernel or dynamically loadable.
+type KernelModuleType int
```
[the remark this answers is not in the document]
majabojarska: Fixed https://github.com/siderolabs/talos/pull/13309/commits/7e249f14c998dd989bae541a2e3f86ea43ec5da8

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,364 @@
...
+)
+
+// KernelModuleStatusController presents both built-in and dynamically loaded Linux kernel modules as resources in the runtime.
+type KernelModuleStatusController struct {
+	V1Alpha1Mode machineruntime.Mode
+
+	ProcModulesDirPath     string
+	ModulesBuiltinFilePath string
+
+	// This never changes during the lifetime of the controller, so we can cache it here to avoid re-reading modules.Builtin on every reconcile loop.
```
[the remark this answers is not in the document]
majabojarska: Fixed https://github.com/siderolabs/talos/pull/13309/commits/44e3574e246938d9c165a83e28266b15a8e4b95d

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/siderolabs/talos/internal/app/machined/pkg/controllers/internal/trigger"
+	machineruntime "github.com/siderolabs/talos/internal/app/machined/pkg/runtime"
+	"github.com/siderolabs/talos/pkg/machinery/resources/runtime"
+)
+
+// KernelModuleStatusController presents both built-in and dynamically loaded Linux kernel modules as resources in the runtime.
+type KernelModuleStatusController struct {
+	V1Alpha1Mode machineruntime.Mode
+
+	ProcModulesDirPath     string
```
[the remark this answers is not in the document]
majabojarska: Fixed https://github.com/siderolabs/talos/pull/13309/commits/4e81bdd5a2e926bac4298e9ee5c6d5961d23cf2b

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,364 @@
...
+	ctx context.Context, r controller.Runtime, rateLimitedTrigger trigger.RateLimitedTrigger, watchCh <-chan *kobject.Event, errCh <-chan error,
+) error {
+	// Initial reconcile to expose resources immediately.
+	rateLimitedTrigger.QueueReconcile()
+
+	for {
+		select {
+		case ev := <-watchCh:
+			// We want only dynamic module load/unload events
+			if ev.Subsystem != "module" {
```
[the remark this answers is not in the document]
majabojarska: Addressed in https://github.com/siderolabs/talos/pull/13309/commits/11d82294dd7d504e1efa56dcf044e07183df8f4b

smira:
Can you please squash the commits (at least before merging) into some reasonable set of commits which has its own purpose?

internal/app/machined/pkg/controllers/internal/kobject/kobject.go
```diff
@@ -65,7 +68,7 @@ func (w *Watcher) Close() error {
 }
 
 // Run the watcher, returns the channel of events.
-func (w *Watcher) Run(logger *zap.Logger) <-chan *Event {
+func (w *Watcher) Run() <-chan *Event {
```
smira: I wonder if we should add a filter here for event type, so that we don't even send events to the channel if we are not interested in t?

majabojarska:
> Can you please squash the commits (at least before merging) into some reasonable set of commits which has its own purpose?

Yes, that's the plan before merging.

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,372 @@
...
+		// no modules in containers
+		return nil
+	}
+
+	if ctrl.ProcModulesPath == "" {
+		ctrl.ProcModulesPath = "/proc/modules"
+	}
+
+	if ctrl.ModulesBuiltinFilePath == "" {
+		release, err := KernelRelease()
```
smira: we can simplify this a lot, we have a constant in `pkg/machinery/constants` for the kernel version, no need to read this in runtime

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,372 @@
...
+		if err != nil {
+			return fmt.Errorf("error determining kernel release: %w", err)
+		}
+
+		ctrl.ModulesBuiltinFilePath = fmt.Sprintf("/lib/modules/%s/modules.builtin", release)
```
smira: same here, it could be just a constant that returns this path (I think we already have a constant for /lib/modules

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,372 @@
...
+
+func (ctrl *KernelModuleStatusController) reconcile(ctx context.Context, r controller.Runtime) error {
+	r.StartTrackingOutputs()
+
+	// These don't change after startup, but must be applied on every reconcile to ensure they are present in the runtime and not deleted by CleanupOutputs.
+	if err := ctrl.reconcileBuiltinModules(ctx, r); err != nil {
+		return fmt.Errorf("error reconciling built-in modules: %w", err)
+	}
+
+	if err := ctrl.reconcileDynamicModules(ctx, r); err != nil {
```
smira, line 155: I think for `builtin` we should also `Stat` the `/sys/module/<name>` path to see if the module is active (we can assume that dynamic are active I guess)

internal/integration/api/kernel.go
```diff
@@ -63,6 +64,19 @@ func (suite *KernelSuite) TestCmdline() {
...
 }
 
+// TestKernelModuleStatus tests the runtime.KernelModuleStatus controller.
+func (suite *KernelSuite) TestKernelModuleStatus() {
+	node := suite.RandomDiscoveredNodeInternalIP()
+	ctx := client.WithNode(suite.ctx, node)
+
+	suite.T().Logf("using node %s", node)
+
+	list, err := safe.StateListAll[*runtime.KernelModuleStatus](ctx, suite.Client.COSI)
```
smira, line 74: we already had some tests for loaded kernel modules in extensions tests, we should use this new resource instead

smira: https://github.com/Spuxy/talos/blob/820a9fa59f7a8096fce77f7f7ac1c9dc0215331f/internal/integration/base/api.go#L516-L532

pkg/machinery/resources/runtime/kernel_module_state.go
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package runtime
+
+//go:generate go tool github.com/dmarkham/enumer -type=KernelModuleState -linecomment -text
```
smira: there's a high cost of running a generator, can you run it once with https://github.com/Spuxy/talos/blob/820a9fa59f7a8096fce77f7f7ac1c9dc0215331f/pkg/machinery/resources/runtime/security_status.go#L25-L26 ? (or move this `//go:generate` combined into `runtime.go`)

pkg/machinery/resources/runtime/kernel_module_status.go
```diff
@@ -0,0 +1,77 @@
...
+	// Type indicates whether the kernel module is built-in or dynamically loaded.
+	Type KernelModuleType `yaml:"type" protobuf:"1"`
+	// Size is the size of the kernel module in bytes.
+	Size int `yaml:"size" protobuf:"2"`
+	// ReferenceCount is the number of references to this kernel module.
+	ReferenceCount int `yaml:"referenceCount" protobuf:"3"`
+	// Dependencies lists the names of kernel modules that this module depends on.
+	Dependencies []string `yaml:"dependencies,omitempty" protobuf:"4"`
+	// State is the operational state of the kernel module.
+	State *KernelModuleState `yaml:"state,omitempty" protobuf:"5"`
```
smira: I think we could drop the pointer here and make default value "NotActive" (or something like that), and use `/sys/module` to turn builtin into Active appropriately?

pkg/machinery/resources/runtime/loaded_kernel_module.go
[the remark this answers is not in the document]
shanduur: That's fine. I'd even go a step further and drop the `loadedkernelmodules.runtime.talos.dev`. There is no API stability guarantee for most of the Status/Spec. Just mention it in the `release.toml` that `KernelModuleStatus` is replacing `LoadedKernelModules`.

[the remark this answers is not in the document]
shanduur: After discussing that, mark it deprecated, remove in another release.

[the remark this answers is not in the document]
majabojarska: All clear, thanks!

pkg/machinery/resources/runtime/kernel_module_state.go
Spuxy: @smira be careful, you have mentioned my fork in this link, Just FYI to be not confused rn

smira: yes, my editor was on your branch, but it doesn't really matter here

pkg/machinery/resources/runtime/kernel_module_status.go
majabojarska: Reworked in https://github.com/siderolabs/talos/pull/13309/changes/8cdd7999bb60ab9093845ece0f178de19a8ca722

pkg/machinery/resources/runtime/kernel_module_state.go
majabojarska: I've consolidated the `go:generate` calls to `enumer` into a single one for all relevant resources under `pkg/machinery/resources/runtime/` into `runtime.go` in the same pkg.

https://github.com/siderolabs/talos/pull/13309/changes/12728b1840ced483b2400b859e701938b8b0aace

internal/integration/api/kernel.go
majabojarska: Reworked in https://github.com/siderolabs/talos/pull/13309/changes/70216f1a98894a120d6df163365fd0de52930dcf

Please do let me know if I missed anything, thanks!

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
majabojarska: Support for Active/Inactive state for built-in modules implemented in https://github.com/siderolabs/talos/pull/13309/changes/8cdd7999bb60ab9093845ece0f178de19a8ca722

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
majabojarska: Reworked in https://github.com/siderolabs/talos/pull/13309/changes/896b55b60c6e0c1820de54d87a9daed194541198

internal/app/machined/pkg/controllers/runtime/kernel_module_status.go
majabojarska: Reworked in https://github.com/siderolabs/talos/pull/13309/changes/7d4c75242aad5b560ebb98264e1ffd9da27f159f

internal/app/machined/pkg/controllers/internal/kobject/kobject.go
majabojarska: Filtering on `.Run()` implemented in https://github.com/siderolabs/talos/pull/13309/changes/246c926abdb1eef0e2947eb7050fd1f01caeac57

I've considered a variadic param approach initially (like `Run(subsystems ...string)`), but:
- Anywhere this watcher is used, we need to focus on just one specific subsystem (e.g. `module`, `block`).
- A simple string equality check is faster and simpler, than a `slices.Contains` check.

internal/integration/base/api.go
```diff
@@ -509,47 +508,18 @@ func (apiSuite *APISuite) AssertServicesRunning(ctx context.Context, node string
 	}
 }
 
-// AssertExpectedModules verifies that expected kernel modules are loaded on the node.
-func (apiSuite *APISuite) AssertExpectedModules(ctx context.Context, node string, expectedModules map[string]string) {
+// AssertExpectedModules verifies that expected kernel modules are loaded on the node as dynamic type and active state.
+func (apiSuite *APISuite) AssertExpectedModules(ctx context.Context, node string, expectedModules []string) {
```
smira, line 512: 👍 let's promote to `integration/extensions` and `integration/nvidia` to verify that the tests pass (if you haven't already)

smira (commented):
👍  (can't approve my own PR lol)

majabojarska (approved):
Approving myself since Andrey did in the previous comment, but cannot rubber stamp his own PR.

PR #13313, opened 2026-05-08
feat: implement DNS over HTTP(S)

### LINKED ISSUE #13310: Support DNS over HTTPS
Same as #9999, but add support for DNS over HTTPS.

smira:
Fixes #13310


internal/pkg/dns/doh/doh.go
```diff
@@ -0,0 +1,227 @@
...
+		ForceAttemptHTTP2:     true,
+		IdleConnTimeout:       90 * time.Second,
+		MaxIdleConns:          10,
+		TLSHandshakeTimeout:   dialTimeout,
+		ExpectContinueTimeout: time.Second,
+	}
+
+	// Pick up Talos-managed root CAs (refreshed live via machine config) and
+	// honor HTTPS_PROXY/NO_PROXY changes for outbound DoH connections.
+	httpdefaults.PatchTransport(transport)
```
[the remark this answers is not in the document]
smira: great point, fixed!

frezbo (approved):

internal/pkg/dns/doh/doh.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/coredns/coredns/plugin/pkg/proxy"
+	"github.com/coredns/coredns/request"
+	"github.com/miekg/dns"
+
+	"github.com/siderolabs/talos/pkg/httpdefaults"
+)
+
+// dialTimeout caps the time spent establishing the underlying TCP+TLS
+// connection to an upstream DoH server.
+const dialTimeout = 5 * time.Second
```
Unix4ever: Isn't it a bit too low? Maybe we should have at least 10 seconds?
Or we could make both timeouts configurable

internal/pkg/dns/doh/doh.go
```diff
@@ -0,0 +1,234 @@
...
+// The opts parameter (TCP/UDP preference, etc.) is irrelevant for DoH — DoH
+// always runs over HTTP/2 — and is therefore ignored.
+func (p *Proxy) Connect(ctx context.Context, state request.Request, _ proxy.Options) (*dns.Msg, error) {
+	if state.Req == nil {
+		return nil, errors.New("doh: nil request")
+	}
+
+	cmsg, err := convertToCodeberg(state.Req)
+	if err != nil {
+		p.fails.Add(1)
```
Unix4ever: Could wrap connect with another function and avoid having this `p.fails.Add` on every `return err`

Unix4ever (approved):

internal/pkg/dns/doh/doh.go
smira: the overall DNS request timeout is actually 4.5 seconds, so it might be too high. 

we go by the timeout of the application trying to resolve, so that if resolving fails, it doesn't say it failed to talk to the host DNS, but rather make host DNS return a proper response back to the app (before it times out)

smira: updated timeout/constants

internal/pkg/dns/doh/doh.go
smira: thanks, fixed

shanduur (approved):

Unix4ever (approved):

PR #13327, opened 2026-05-11
docs: drop controlplane endpoint examples

### LINKED ISSUE #13325: Misleading in the logging related documentation
### Description

[MachineConfig](https://docs.siderolabs.com/talos/v1.13/reference/configuration/v1alpha1/config#logging) documentation says that logging destinations are either `tcp` or `udp`. 

However, endpoints examples clearly says that it might be `https` as well.

Image

This might cause some misleading on what protocols might be actually used here.

smira:
They seem to be redundant, and as they attach to a common `Endpoint` type, the doc generator attaches it to the logging endpoint, which is wrong.

This seems to be the easiest way out.

Fixes #13325

### DIFF
--- a/pkg/machinery/config/types/v1alpha1/v1alpha1_examples.go
+++ b/pkg/machinery/config/types/v1alpha1/v1alpha1_examples.go
@@ -323,18 +323,6 @@ func machineSeccompExample() []*MachineSeccompProfile {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


smira:
I need to rework this a bit... still incorrect

frezbo (approved):

shanduur (approved):

PR #13330, opened 2026-05-11
fix: enforce strict QoS ordering in OOM victim selection

### LINKED ISSUE #13329: Talos OomAction triggered for kube-apiserver
### Description
Since upgrading to talos v1.13.0 I get way more oomactions, using the default v1alpha1/OomConfig. The initial dump of `talosctl get oomactions -o yaml` can be found in [this gist](https://gist.github.com/SironheartAcc/f83c55dd6a2ab3426bcdd2e93dc0e30c)

Be aware, this is a complete dump, but there are 5 mentions of `/usr/local/bin/kube-apiserver` in there for services being killed.

### Logs
will be provided once needed.

### Environment

- Talos version: v1.13.0
- Kubernetes version: v1.36.0
- Platform: Metal

eleboucher:
# Pull Request

## What? (description)

The CEL ranking only weighted BestEffort 2x over Burstable, so a large Burstable pod (e.g. kube-apiserver) could outscore a smaller BestEffort pod and get killed first.

Do the class ordering in Go instead: pick the lowest-importance class with any eligible cgroup, then the highest-scoring one inside it.

## Why? (reasoning)

Fixes  https://github.com/siderolabs/talos/issues/13329

Closes #13384

## Acceptance


smira:
This goes against the original design intention, I can see the problem, but this might not be the right solution.

eleboucher:
> This goes against the original design intention, I can see the problem, but this might not be the right solution.

well looking at the comment in the constant, the current algorithm isn't doing what it's supposed to — a 2:1 weight isn't enough, a big burstable pod (kube-apiserver) still outscores small besteffort ones. what we can do is put a magic number, like the linux kernel does with oom_score_adj, so the class always wins over memory:
```
  memory_max.hasValue() ? 0.0 :
    {Besteffort: 2.0, Burstable: 1.0, Guaranteed: 0.0, Podruntime: 0.0, System: 0.0}[class] *
       (1.0e18 + double(memory_current.orValue(0u)))
```
map values are priority levels, 1.0e18 makes the band gap unbridgeable by memory_current.

smira:
I don't see the problem right now, it does what it is supposed to do, everything else is left to the operator.

Let's first see what #13329 has exactly, and then move into the discussion of the fix, and only after that move into the actual code.

smira:
The plan is to incorporate the change, but make it configurable (and enabled by default).

smira:
> The plan is to incorporate the change, but make it configurable (and enabled by default).

Thanks, @eleboucher, this was a not to myself, I didn't mean to make you do more work :)

eleboucher:
> > The plan is to incorporate the change, but make it configurable (and enabled by default).
> 
> Thanks, @eleboucher, this was a not to myself, I didn't mean to make you do more work :)

always happy to contribute to talos

smira (approved):
thank you!

PR #13345, opened 2026-05-13
feat: redact more machine config secrets and audit redactors

utkuozdemir:
The legacy machine config now redacts registry credentials, mutual TLS client keys, WireGuard private keys, the Equinix Metal and Hetzner Cloud VIP API tokens, and disk encryption static passphrases.

The same fields are redacted in the multi-doc successors. SideroLink now also clears the unique connection token, not only the join token in the API URL.

A new test walks every registered document kind, picks out string and byte fields whose yaml tag name looks like a secret, and checks that the document's Redact replaces them. New kinds that miss one will fail the test.

### DIFF
--- a/pkg/machinery/config/types/registry_audit_test.go
+++ b/pkg/machinery/config/types/registry_audit_test.go
@@ -1,0 +1,248 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package types_test
+
+import (
+	"reflect"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/config/internal/registry"
+	_ "github.com/siderolabs/talos/pkg/machinery/config/types" // register all document kinds
+)
+
+// TestRegistryRedactAudit walks every registered document kind, looks for
+// fields whose yaml tag name suggests a secret, and asserts that:
+//
+//  1. The document implements config.SecretDocument.
+//  2. Calling Redact actually replaces the field's value.
+//
+// Detection is a pure case-insensitive substring match on the yaml tag against
+// a small keyword list. Tags in this codebase are camelCase, so no separator
+// handling is needed. New documents added in the future will be picked up
+// automatically without an allowlist to maintain.
+//
+// The frozen v1alpha1 Config is registered too and gets walked here, but in
+// practice it never grows new fields, so its Redact is also covered by
+// v1alpha1_redact_test.go for direct coverage of its existing surface.
+func TestRegistryRedactAudit(t *testing.T) {
+	t.Parallel()
+
+	kinds := registry.Kinds()
+	require.NotEmpty(t, kinds, "no document kinds registered, blank imports in types.go are missing")
+
+	for _, kind := range kinds {
+		t.Run(kind, func(t *testing.T) {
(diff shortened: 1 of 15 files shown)


pkg/machinery/config/types/registry_audit_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/config/internal/registry"
+	_ "github.com/siderolabs/talos/pkg/machinery/config/types" // register all document kinds
+)
+
+// TestRegistryRedactAudit walks every registered document kind, looks for
+// fields whose yaml tag name suggests a secret, and asserts that:
+//
+//  1. The document implements config.SecretDocument.
+//  2. Calling Redact actually replaces the field's value.
```
utkuozdemir, line 25: I thought this would be a good idea, pls let me know if you think otherwise.

pkg/machinery/config/types/registry_audit_test.go
```diff
@@ -0,0 +1,248 @@
...
+				out = append(out, finding{
+					path: fieldPath,
+					set:  func(s string) { fv.SetBytes([]byte(s)) },
+					get:  func() string { return string(fv.Bytes()) },
+				})
+
+				continue
+			}
+
+			out = append(out, findSecretFields(fv, fieldPath)...)
```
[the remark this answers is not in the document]
utkuozdemir: `fv` is assigned like `fv := v.Field(i)`, it does not get reassigned every iteration.

pkg/machinery/config/types/registry_audit_test.go
```diff
@@ -0,0 +1,248 @@
...
+					"field %s.%s was not redacted by Redact", kind, f.path)
+			}
+		})
+	}
+}
+
+// secretKeywords is the case-insensitive substring set used to flag suspect yaml tags.
+// Deliberately narrow: bare "auth" and "key" would generate false positives
+// (authorizationConfig, publicKey, encryptionKeySize, ...).
+var secretKeywords = regexp.MustCompile(`(?i)(password|passphrase|token|secret|privatekey|presharedkey|apikey)`)
```
utkuozdemir, line 85: Open for suggestions on this.

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
```diff
@@ -144,6 +144,63 @@ func (c *Config) Redact(replacement string) {
...
+				registry.RegistryAuth.RegistryAuth = redactStr(registry.RegistryAuth.RegistryAuth)
+				registry.RegistryAuth.RegistryIdentityToken = redactStr(registry.RegistryAuth.RegistryIdentityToken)
+			}
+
+			if registry.RegistryTLS != nil {
+				if registry.RegistryTLS.TLSClientIdentity != nil {
+					registry.RegistryTLS.TLSClientIdentity.Key = redactBytes(registry.RegistryTLS.TLSClientIdentity.Key)
+				}
+
+				registry.RegistryTLS.TLSCA = redactBytes(registry.RegistryTLS.TLSCA)
```
smira: this is not secret, it's a CA validation (public certificate).

smira (approved):

pkg/machinery/config/types/v1alpha1/v1alpha1_provider.go
utkuozdemir: Did it for consistency, because it was redacted in the new document: https://github.com/siderolabs/talos/blob/main/pkg/machinery/config/types/cri/registry_tls.go#L194-L196

I can instead also remove that redaction, wdyt?

smira: yeah, I wonder if I over-did this... this is not a secret 🤦‍♂️

PR #13347, opened 2026-05-13
feat: implement static host resolving via host DNS 

### LINKED ISSUE #13141: StaticHostConfig not used when forwardKubeDNSToHost: true
## Bug Report

Notice this may not be a bug but simply not supported, yet i expected this to work.

### Description

For dev purposes i'm using a bunch on fake dns names (like keycloak.example.org) and some of these need to be resolved from within the cluster (oidc cycle for instance). I thought to add the mappings using a StaticHostConfig (and i confirmed the mapping is in the talos hosts /etc/hosts file and then when trying to resolve the name from a pod it sends the request to coredns which delegates to the upstream (talos dns server). This indeed occurs but talos doesn't take the local mappings into account but just forwards to the configured upstreams, as a result it fails to resolve.

How can i best proceed; Run (another) dns server locally and add as upstream? Or patch coredns?

### Logs

talosctl --nodes $CTL3_IP  logs dns-resolve-cache --follow | grep foo.bar.baz
```
192.168.50.184: 2026-04-16T13:46:14.150Z DEBUG making dns request {"component": "dns-resolve-cache", "data": ";; opcode: QUERY, status: NOERROR, id: 6941\n;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0\n\n;; QUESTION SECTION:\n;foo.bar.baz.\tIN\t A\n", "proto": "udp", "question": "foo.bar.baz.", "local_addr": "169.254.116.108:53", "remote_addr": "10.244.141.197:46797", "upstream": "192.168.50.1:53"}
192.168.50.184: 2026-04-16T13:46:14.154Z DEBUG dns response {"component": "dns-resolve-cache", "data": ";; opcode: QUERY, status: NOERROR, id: 6941\n;; flags: rd; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0\n\n;; QUESTION SECTION:\n;foo.bar.baz.\tIN\t A\n", "proto": "udp", "question": "foo.bar.baz.", "local_addr": "169.254.116.108:53", "remote_addr": "10.244.141.197:46797", "data": ";; opcode: QUERY, status: NXDOMAIN, id: 6941\n;; flags: qr rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 0\n\n;; QUESTION SECTION:\n;foo.bar.baz.\tIN\t A\n\n;; AUTHORITY SECTION:\n.\t504\tIN\tSOA\ta.root-servers.net. nstld.verisign-grs.com. 2026041601 1800 900 604800 86400\n"}
```

### Environment

- Talos version: v1.12.6
- Kubernetes version: v1.35.2

smira:
Fixes https://github.com/siderolabs/talos/issues/13141


smira:
Note: some refactoring (moving code around) is split as the first commit.

internal/app/machined/pkg/controllers/network/static_host.go
```diff
@@ -0,0 +1,171 @@
...
+		names := []string{
+			normalizeHostName(hostnameStatus.TypedSpec().Hostname),
+			normalizeHostName(hostnameStatus.TypedSpec().FQDN()),
+		}
+
+		for _, name := range names {
+			if name == "" {
+				continue
+			}
+
```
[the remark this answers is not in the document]
smira: I think this is fine.

internal/pkg/dns/manager.go
```diff
@@ -220,6 +223,28 @@ func (s *addrResolver) ResolveAddr(ctx context.Context, qType uint16, name strin
...
+
+func (s *staticHostResolver) ResolveAddr(ctx context.Context, qType uint16, name string) (iter.Seq[netip.Addr], bool) {
+	name = strings.TrimRight(name, ".")
+
+	key := strings.ToLower(strings.TrimRight(name, "."))
+
+	addrs, err := s.shr.ReadStaticHosts(ctx, key)
+	if err != nil {
+		return nil, false
+	}
```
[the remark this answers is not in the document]
smira: I think it's fine.

internal/pkg/dns/static_host.go
```diff
@@ -0,0 +1,67 @@
...
+	if idx == -1 {
+		return h.next.ServeDNS(ctx, wrt, msg)
+	}
+
+	req := request.Request{W: wrt, Req: msg}
+
+	addrs, ok := h.mapper.ResolveAddr(ctx, req.QType(), req.Name())
+	if !ok {
+		return h.next.ServeDNS(ctx, wrt, msg)
+	}
```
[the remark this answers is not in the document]
smira: fixed, thanks

internal/pkg/dns/cache.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		// we should probably call newProxy.Healthcheck() if there are too many errors
+		c.logger.Warn("error serving dns request", zap.Error(err))
+	}
+
+	if clientWrite(code) {
+		return
+	}
+
+	// Something went wrong
```
frezbo, line 56: hmm what went wrong here?

smira: this is actually part of the refactoring (I haven't touched this code, I just moved it around), it's the first commit which just splits the code into multiple files.

smira: but I guess it means that we can't serve the DNS request, so the chain wrote no response back, and we need to write our own error response

frezbo: let's add this as a comment, which is more helpful

frezbo (approved):

frezbo (approved):

shanduur (approved):

PR #13355, opened 2026-05-14
fix: restore some shared (and some lower tier slave) mount propagation

### LINKED ISSUE #13352: Kubelet cannot update projected volumes after restart
### Description
In case one restarts kubelet projected volume updates are no longer visible in pods. E.g. service account tokens won't be refreshed.

Talos starts kubelet in containerd container and bind mounts /var/lib/kubelet from host into the container. It is mounted as shared bind mount, so all tmpfs mounts for projected volumes are propaged back to host. But on kubelet restart it gets broken. Because rbind is not used none of the tmpfs under /var/lib/kubelet are mounted back to new kubelet container and they are overmounted again by kubelet. So kubelet updates its own new tmpfs volume but containers see tmpfs from of the previous kubelet.

Looks like the bug was introduced here: 
https://github.com/siderolabs/talos/commit/04e3e87adcbd24ee0d82dce4cc27121d34d316f4
But I didn't test it.

### Environment

- Talos version: 1.13.2
- Kubernetes version: 1.35, 1.36, 1.34
- Platform: vmware

smira:
Fixes #13352

### DIFF
--- a/internal/app/machined/pkg/system/services/kubelet.go
+++ b/internal/app/machined/pkg/system/services/kubelet.go
@@ -162,12 +162,12 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: constants.PodResolvConfPath, Source: constants.PodResolvConfPath, Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/etc/cni", Source: "/etc/cni", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"bind", "rw"}},
+		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"rbind", "rslave", "rw"}},
+		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rslave", "rw"}},
+		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/log/containers", Source: "/var/log/containers", Options: []string{"bind", "rw"}},
 		{Type: "bind", Destination: "/var/log/pods", Source: "/var/log/pods", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: constants.UserVolumeMountPoint, Source: constants.UserVolumeMountPoint, Options: []string{"rbind", "ro"}},
+		{Type: "bind", Destination: constants.UserVolumeMountPoint, Source: constants.UserVolumeMountPoint, Options: []string{"rbind", "rslave", "ro"}},
 	}
 
 	if _, err := os.Stat("/sys/kernel/security"); err == nil {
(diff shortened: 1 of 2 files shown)


internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -162,12 +162,12 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
...
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"bind", "rw"}},
+		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"rbind", "rslave", "rw"}},
+		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rslave", "rw"}},
+		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"rbind", "rshared", "rw"}},
 		{Type: "bind", Destination: "/var/log/containers", Source: "/var/log/containers", Options: []string{"bind", "rw"}},
 		{Type: "bind", Destination: "/var/log/pods", Source: "/var/log/pods", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: constants.UserVolumeMountPoint, Source: constants.UserVolumeMountPoint, Options: []string{"rbind", "ro"}},
```
smira, line 170: using `rslave` here, so that mount propagates host -> kubelet, but the other way around never works

internal/app/machined/pkg/system/services/kubelet.go
```diff
@@ -162,12 +162,12 @@ func (k *Kubelet) Runner(r runtime.Runtime) (runner.Runner, error) {
 		{Type: "bind", Destination: constants.PodResolvConfPath, Source: constants.PodResolvConfPath, Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/etc/cni", Source: "/etc/cni", Options: []string{"bind", "ro"}},
 		{Type: "bind", Destination: "/usr/libexec/kubernetes", Source: "/usr/libexec/kubernetes", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/run", Source: "/run", Options: []string{"bind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/containerd", Source: "/var/lib/containerd", Options: []string{"rbind", "rw"}},
-		{Type: "bind", Destination: "/var/lib/kubelet", Source: "/var/lib/kubelet", Options: []string{"bind", "rw"}},
```
smira, line 167: kubelet "native" mount point.

frezbo (approved):
Can we catch this with a test, from the report we should be able to generate a test

smira:
> Can we catch this with a test, from the report we should be able to generate a test

I'm not sure how to do this exactly, as we need to trigger kubelet to update the service account after restart? I mean can we trigger this update without waiting for service account to become stale?

frezbo:
> > Can we catch this with a test, from the report we should be able to generate a test
> 
> I'm not sure how to do this exactly, as we need to trigger kubelet to update the service account after restart? I mean can we trigger this update without waiting for service account to become stale?

if we use a pod with projected service account with a very low ttl, then restart kubelet then check again?

internal/integration/api/kubelet.go
```diff
@@ -0,0 +1,249 @@
...
+// TestProjectedVolumeUpdatesSurviveKubeletRestart creates a pod with a
+// downwardAPI projected volume exposing the pod's labels, restarts the
+// kubelet on the pod's node, then patches a label on the pod and asserts
+// that the projected file inside the pod is updated.
+//
+// The bug from #13352 manifests as: after kubelet restart, the new kubelet
+// writes projected-volume updates into a tmpfs that is invisible to running
+// pods (because /var/lib/kubelet was bind-mounted without rbind/rshared),
+// so the pod keeps reading the pre-restart value forever.
+func (suite *KubeletSuite) TestProjectedVolumeUpdatesSurviveKubeletRestart() {
```
smira, line 74: @frezbo this is a regression test

frezbo (approved):
🆒

PR #13363, opened 2026-05-15
chore: drop e2e step dependencies

smira:
This triggers a rebuild of an integration test every time a job runs, even though it's an artifact already.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -579,6 +579,9 @@ $(INTEGRATION_TEST)-darwin-arm64:
 
 $(INTEGRATION_TEST): $(INTEGRATION_TEST)-$(OPERATING_SYSTEM)-$(ARCH) ## Builds the integration test binary for the host OS/arch.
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -600,7 +600,7 @@ $(ARTIFACTS)/cilium: | $(ARTIFACTS)
 
 external-artifacts: $(ARTIFACTS)/kubectl $(ARTIFACTS)/kubestr $(ARTIFACTS)/helm $(ARTIFACTS)/cilium
 
-e2e-%: $(INTEGRATION_TEST) external-artifacts ## Runs the E2E test for the specified platform (e.g. e2e-docker).
```
smira, line 603: another idea is to make `$(INTEGRATION_TEST)` depend on a proper file, but that means locally you have to rebuild it every time, bad either way

frezbo: we now have a `make integration-test` which would build the integration binary always

smira: yes... I will refactor the target instead

frezbo:
> This triggers a rebuild of an integration test every time a job runs, even though it's an artifact already.

hmm it shouldn't, all external artifacts use dependency as `: | $(ARTIFACTS)` so if the download step put the file, it shouldn't download again

frezbo (approved):

PR #13365, opened 2026-05-15
chore: update image signer

frezbo:
Update image signer and use version stamp file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ GENERATE_VEX ?= latest
 
 KRES_IMAGE ?= ghcr.io/siderolabs/kres:latest
 CONFORMANCE_IMAGE ?= ghcr.io/siderolabs/conform:latest
-IMAGE_SIGNER_RELEASE ?= v0.2.0
+IMAGE_SIGNER_RELEASE ?= v0.3.2
 
 PKG_APPARMOR ?= $(PKGS_PREFIX)/apparmor:$(PKGS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -703,13 +703,14 @@ clean: ## Cleans up all artifacts.
 image-list: ## Prints a list of all images built by this Makefile with digests.
 	@echo -n installer-base$(IMAGE_NAME_SUFFIX) talos$(IMAGE_NAME_SUFFIX) imager$(IMAGE_NAME_SUFFIX) talosctl$(IMAGE_NAME_SUFFIX) talosctl-all$(IMAGE_NAME_SUFFIX) | xargs -d ' ' -I{} sh -c 'echo $(REGISTRY_AND_USERNAME)/{}:$(IMAGE_TAG_IN)' | xargs -I{} sh -c 'echo {}@$$(crane digest {})'
 
-$(ARTIFACTS)/image-signer: $(ARTIFACTS) ## Downloads image-signer binary
+$(ARTIFACTS)/image-signer-$(IMAGE_SIGNER_RELEASE): | $(ARTIFACTS) ## Downloads image-signer binary
 	@curl -sSL https://github.com/siderolabs/go-tools/releases/download/$(IMAGE_SIGNER_RELEASE)/image-signer-$(OPERATING_SYSTEM)-$(ARCH) -o $(ARTIFACTS)/image-signer
```
smira: I think it's cleaner/more robust to download to the same path, with version attached - this way it will be downloaded exactly once (and normal dependencies would work)

Orzelius (approved):

PR #13366, opened 2026-05-15
fix: streamline config validation flow

smira:
And specifically, for Kubernetes versions.

Fixes https://github.com/siderolabs/talos/discussions/13361

First of all, refactor (confusing) method names for config validation, deprecate the old ones and introduce two new entrypoints:

* `ValidateAsClient` - used in the `talosctl validate` CLI (when outside of Talos)
* `ValidateAtRuntime` - used by Talos itself, and _includes_ `ValidateAsClient` as well.

This avoids previous pattern than you have to call two methods: `Validate` and `RuntimeValidate` on the config and join their results.

Move Kubernetes version validation from `RuntimeValidate` into regular `Validate`, but guard it under the `!Local` condition to make it run only inside the Talos context. This exposes it cases when we can't call `ValidateAtRuntime`, e.g. in the `installer` on upgrade, when access to host Talos resource state is not guaranteed.

### DIFF
--- a/cmd/installer/cmd/installer/install.go
+++ b/cmd/installer/cmd/installer/install.go
@@ -60,7 +60,7 @@ func runInstallCmd(ctx context.Context) (err error) {
 	} else {
 		var warnings []string
 
-		warnings, err = config.Validate(p.Mode())
+		warnings, err = config.ValidateAsClient(p.Mode())
 		if err != nil {
 			return fmt.Errorf("machine configuration is invalid: %w", err)
 		}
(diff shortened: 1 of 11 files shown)


smira:
tl;dr the issue from #13361 was that Talos _allowed_ an upgrade with unsupported Kubernetes version (as the installer did only client-side validation), and in the end got into a state when it doesn't accept machine config changes, while allows to boot with invalid config.

This should close the gap, and:

* don't allow to upgrade if config is invalid
* don't allow to boot with invalid config

frezbo (approved):

Orzelius (approved):
nice! much cleaner

PR #13376, opened 2026-05-18
feat: enable Flannel nftables mode

smira:
Use native nftables (it uses `nft` CLI internally) instead of going via `iptables-nft` shim.


frezbo (approved):

shanduur (approved):
🆒

PR #13378, opened 2026-05-18
feat: add quirks for RegistryAuthConfig support

Unix4ever:
Config is supported starting from Talos version 1.12.0.

### DIFF
--- a/pkg/machinery/imager/quirks/quirks.go
+++ b/pkg/machinery/imager/quirks/quirks.go
@@ -336,3 +336,15 @@ func (q Quirks) SupportsFactoryTalosctlDownload() bool {
 
 	return q.v.GTE(minTalosVersionFactoryTalosctlDownload)
 }
+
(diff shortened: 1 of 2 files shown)


smira (commented):
we don't need this, it's already gated with network multi-doc https://github.com/siderolabs/talos/blob/ecb7d45883bc1de46c4adbe37dcf556c59e78a25/pkg/machinery/config/generate/registry.go#L21-L22

But also you can't really do authenticated Image Factory without it, as it will not be able to do e.g. upgrade in maintenance mode without multi-doc

Unix4ever:
> we don't need this, it's already gated with network multi-doc
> 
> https://github.com/siderolabs/talos/blob/ecb7d45883bc1de46c4adbe37dcf556c59e78a25/pkg/machinery/config/generate/registry.go#L21-L22
> 
> But also you can't really do authenticated Image Factory without it, as it will not be able to do e.g. upgrade in maintenance mode without multi-doc

A bit confusing but alright. Easier for me actually :)

PR #13381, opened 2026-05-18
feat: support 4k sector size disk images

### LINKED ISSUE #13227: Allow downloading metal images from factory with 4k sector size.
## Feature Request
Allow downloading metal images from factory with 4k sector size.

### Description
We use metal images to deploy Talos on multiple hardware platforms. Some of the disks / virtual disks from RAID controllers are presented as 4k-native, so unpacking raw metal images over such disks does not bring desired result, as no correct partitions are produced.
It would be nice if Talos Factory allowed to download raw images for 4k-native disks.

smira:
By default, we create still standard 512-bytes sector size images, but optionally sector size can be set to any value.

Fixes #13227


frezbo (approved):

smira:
There is a broken Canal test, but unrelated to this PR.

PR #13387, opened 2026-05-19
feat(machined): live-grow partition and XFS filesystem when backing disk is expanded

mcanevet:
## What?

When a virtual machine's disk is expanded at the hypervisor level (e.g.
`qemu-img resize`, vSphere disk extension), Talos now automatically grows
the EPHEMERAL partition and its XFS filesystem without requiring a reboot.

Changes:

- **`volume_manager.go`**: `VolumeManagerController` adds a 30-second ticker
  that calls `Grow()` on volumes in `VolumePhaseReady`. When growth is detected,
  the GPT partition is extended via `BLKPG_RESIZE_PARTITION` (safe for mounted
  partitions) and the XFS filesystem is grown inline via `xfs_growfs`.
  `ctx.Done()` now returns a non-nil error so the rruntime adapter restarts the
  controller instead of treating it as permanently finished.

- **`grow.go`**: `Grow()` signature changed from `error` to `(bool, error)` —
  returns `true` when the partition actually grew, so `XFSGrow` is only called
  when needed. Phase advancement removed from `Grow()`; that is now the
  caller's responsibility.

- **`volumes.go`**: integration test `TestLiveDiskResize` that truncates the
  QEMU backing file, sends a `block_resize` notification via the HMP monitor,
  and asserts that `EPHEMERAL` `VolumeStatus.Size` increases within 5 minutes
  without a reboot.

## Why?

Talos nodes on QEMU (and other hypervisors) can have their virtual disks
expanded while running. Today that extra space is invisible until a full
reinstall or reboot. This change makes Talos respond to available partition
growth on the next 30-second tick, which is already the cadence used by the
volume manager for other housekeeping.

The `BLKPG_RESIZE_PARTITION` ioctl is explicitly designed for resizing mounted
partitions in-place; `xfs_growfs` handles live XFS expansion. Both operations
are non-destructive and safe to run while the filesystem is mounted.

## Acceptance

- [x] you linked an issue (if applicable) — no upstream issue; this is a
      quality-of-life improvement for VM operators
- [x] you included tests — `TestLiveDiskResize` integration test (QEMU only);
      `grow_test.go` updated to match new `Grow()` signature and assert the
      phase-not-modified contract
- [x] you ran conformance (`make conformance`) — passes (GPG Identity check
      fails as expected for external contributors)
- [x] you formatted your code (`make fmt`) — gofumpt fix applied manually
      (`make lint-fmt` hit a BuildKit cache bug on the dev machine)
- [x] you linted your code (`make lint`) — passes
- [ ] you generated documentation (`make docs`) — not applicable, no new
      user-facing configuration
- [x] you ran unit-tests (`make unit-tests`) — verified via
      `go test ./internal/app/machined/pkg/controllers/block/...`; full
      `make unit-tests` blocked by disk space on dev machine, disk-operation
      tests require root and skip gracefully without it

### DIFF
--- a/internal/app/machined/pkg/controllers/block/volume_manager.go
+++ b/internal/app/machined/pkg/controllers/block/volume_manager.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"math"
 	"slices"
-	"time"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/resource"
@@ -19,7 +18,6 @@ import (
 	"github.com/cosi-project/runtime/pkg/state"
 	"github.com/siderolabs/gen/optional"
 	"github.com/siderolabs/gen/value"
-	"github.com/siderolabs/gen/xerrors"
 	"github.com/siderolabs/gen/xslices"
 	"go.uber.org/zap"
 
@@ -33,6 +31,7 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/resources/hardware"
 	"github.com/siderolabs/talos/pkg/machinery/resources/runtime"
 	"github.com/siderolabs/talos/pkg/machinery/resources/secrets"
+	"github.com/siderolabs/talos/pkg/makefs"
 )
 
 // VolumeManagerController manages volumes in the system, converting VolumeConfig resources to VolumeStatuses.
@@ -132,22 +131,11 @@ func (ctrl *VolumeManagerController) Run(ctx context.Context, r controller.Runti
 		deviceReadyRequest  int
 	)
 
-	retryTicker := time.NewTicker(30 * time.Second)
-	defer retryTicker.Stop()
-
-	shouldRetry := false
-
 	for {
 		select {
 		case <-r.EventCh():
 		case <-ctx.Done():
 			return nil
-		case <-retryTicker.C:
-			if !shouldRetry {
-				continue
-			}
-
-			shouldRetry = false
 		}
 
 		// if devices are not ready, we can't provision and locate most volumes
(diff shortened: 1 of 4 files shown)


internal/app/machined/pkg/controllers/block/volume_manager.go
```diff
@@ -135,19 +135,19 @@ func (ctrl *VolumeManagerController) Run(ctx context.Context, r controller.Runti
...
 		case <-r.EventCh():
 		case <-ctx.Done():
-			return nil
+			// Return a non-nil error so the rruntime adapter restarts this
+			// controller rather than treating a nil return as "done forever".
+			// The adapter is expected to stop restarting once its own context
+			// is canceled (e.g. on machined shutdown).
+			return errors.New("VolumeManagerController restarting")
 		case <-retryTicker.C:
-			if !shouldRetry {
```
smira, line 146: these changes don't look good to me, returning an error causes an exponential backoff, we don't want it here

internal/app/machined/pkg/controllers/block/volume_manager.go
```diff
@@ -534,17 +531,47 @@ func (ctrl *VolumeManagerController) processVolumeConfig(ctx context.Context, lo
 			// normal state machine
 			switch volumeContext.Status.Phase {
 			case block.VolumePhaseReady:
-				// nothing to do, ready
+				grew, err := volumes.Grow(ctx, logger, volumeContext)
```
smira: we don't need to do this here

we have a way to detect disk size changes if we want to via a blockdevice event from the kernel as I would guess (I haven't checked, but I guess it should?), and we could actually pin potential grow candidates and try to grow only those which are on the disk which got changed.

Growing on timer is a bad idea imho.

smira (commented):
the approach in the PR is still wrong, if we have time, we will rework this

I did some manual testing, and Talos already correctly recognizes disk size change, and reports new disk size. 

So a proper implementation should act on Disk changes (it already does), and remember the disk size at the moment of last grow attempt, and compare it with new disk size - if there  is a diff (and only then) a grow attempt is performed (remembering the new disk size anyways).

smira:
Thank you, this is going to be replaced with #13874.

Issue #13348, opened 2026-05-13
refactor kube-scheduler and kube-controller-manager configs


smira:
Split them out of v1alpha1_types, deprecate old config.

Rework config generation.

Add config for controller-manager.

PR #13395, opened 2026-05-19
feat: introduce new KubeSchedulerConfig

smira:
Move old `.cluster.scheduler` into new multi-doc document.

Changes vs. the old config:

* image is required (previous default "use bundled Kubernetes version"
  is gone)
* volumes field is removed (I don't see any usecase for that, with custom
  scheduler config available).

Part of https://github.com/siderolabs/talos/issues/13348


internal/app/machined/pkg/controllers/k8s/control_plane_final.go
```diff
@@ -0,0 +1,119 @@
...
+			MapMetadataOptionalFunc: func(in *k8s.SchedulerConfig) optional.Optional[*k8s.SchedulerConfig] {
+				if in.Metadata().ID() != k8s.SchedulerConfigID {
+					return optional.None[*k8s.SchedulerConfig]()
+				}
+
+				return optional.Some(k8s.NewSchedulerConfig(k8s.FinalSchedulerConfigID))
+			},
+			TransformFunc: func(ctx context.Context, r controller.Reader, logger *zap.Logger, in *k8s.SchedulerConfig, out *k8s.SchedulerConfig) error {
+				// clear the spec
+				*out.TypedSpec() = k8s.SchedulerConfigSpec{}
```
smira, line 42: this is the change to push non-trivial transformation into an easily testable controller - allows to write unit-tests for it

pkg/cluster/kubernetes/kubelet.go
```diff
@@ -180,16 +182,8 @@ func extractKubeletVersionSuffix(imageRef string) string {
 func upgradeKubeletPatcher(
 	options UpgradeOptions,
 	kubeletSpec *k8s.KubeletSpec,
-) func(config *v1alpha1config.Config) error {
```
smira, line 183: using generic patches (see stdpatches changes in this PR) so that we can handle old/new config format in a sane way; in this PR only the kube-scheduler got special treatment, but this provides groundwork for the future

pkg/machinery/compatibility/kubernetes_image.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/siderolabs/talos/pkg/machinery/version"
+)
+
+// ValidateKubernetesImageTag validates the Kubernetes image tag format.
+func ValidateKubernetesImageTag(imageRef string) error {
```
smira, line 15: moved this out of `v1alpha_validation.go` to provide shared/common place for these validations

pkg/machinery/config/generate/stdpatches/guess.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// GuessVersionContractKubeControllerManager attempts to guess the version contract for kube-controller-manager configuration based on the provided machine configuration.
+func GuessVersionContractKubeControllerManager(cfg config.Container) *config.VersionContract {
+	return config.TalosVersionCurrent
+}
+
+// GuessVersionContractKubeScheduler attempts to guess the version contract for kube-scheduler configuration based on the provided machine configuration.
+func GuessVersionContractKubeScheduler(cfg config.Container) *config.VersionContract {
+	if hasDocument(k8s.KubeSchedulerConfig, cfg) {
```
smira, line 29: every other one is a stub for now (as we don't have different machine configs yet)

pkg/machinery/go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/siderolabs/talos/pkg/machinery
 
-go 1.26.1
+go 1.26.2
 
 // forked ethtool introduces missing APIs
 replace github.com/mdlayher/ethtool => github.com/siderolabs/ethtool v0.6.0-sidero
 
+replace github.com/cosi-project/runtime => github.com/smira/os-runtime v0.0.0-20260520152655-1ba4717975e7
```
smira: this is https://github.com/cosi-project/runtime/pull/667

frezbo (approved):

PR #13399, opened 2026-05-20
feat: update default Kubernetes version to 1.36.1

smira:
Using the latest available version.

PR #13402, opened 2026-05-21
fix: bump Kubernetes to 1.36.1 in one more place

smira:
Fix for #13399


frezbo (approved):
provision.go also has 1.3.6.0 is that fine?

smira:
> provision.go also has 1.3.6.0 is that fine?

this one is fixed for a specific release of Talos, we should change it in sync with Talos release we are testing with

PR #13408, opened 2026-05-21
feat: add vex and sbom support to quirks

Slessi:
Add SupportsSBOM and SupportsVEX to machinery quirks.

### DIFF
--- a/pkg/machinery/imager/quirks/quirks.go
+++ b/pkg/machinery/imager/quirks/quirks.go
@@ -336,3 +336,27 @@ func (q Quirks) SupportsFactoryTalosctlDownload() bool {
 
 	return q.v.GTE(minTalosVersionFactoryTalosctlDownload)
 }
+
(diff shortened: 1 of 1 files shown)


pkg/machinery/imager/quirks/quirks.go
```diff
@@ -336,3 +336,27 @@ func (q Quirks) SupportsFactoryTalosctlDownload() bool {
 
 	return q.v.GTE(minTalosVersionFactoryTalosctlDownload)
 }
+
+var minTalosVersionSBOM = semver.MustParse("1.11.0")
+
+// SupportsSBOM returns true if the Talos version supports SBOMS from image factory.
```
frezbo, line 342: shouldn't factory report these via some api, since factory version also has to support SBOM's, do we assume factory is always latest?

frezbo (approved):

smira (commented):
we don't need

PR #13412, opened 2026-05-21
feat: add bnxt_re module to the rootfs

smira:
See https://github.com/siderolabs/pkgs/pull/1562


frezbo (approved):
good PPP is going to be in some extension

smira:
> good PPP is going to be in some extension

PPP is `=y`, it's builtin

frezbo:
> > good PPP is going to be in some extension
> 
> PPP is `=y`, it's builtin

I keep forgetting that 😅

PR #13418, opened 2026-05-22
fix: memorymodules resource reporting

frezbo:
Since we're creating memorymodules with the devicelocator as id, this was causing talos to misreport memory information if there are multiple sticks attached to the same devicelocator, fix by using the banklocator information also for the id.

go-smbios bump is no-op, just dependency bumps in go-smbios.


smira (approved):

majabojarska:
Not backporting to 1.12, since it would require bumping the Go toolchain to >=1.26.3 and redoing a fair amount of builds.

```sh
0.623 go: github.com/siderolabs/go-smbios@v0.3.4 requires go >= 1.26.3 (running go 1.25.11; GOTOOLCHAIN=local)
```

frezbo:
> Not backporting to 1.12, since it would require bumping the Go toolchain to >=1.26.3 and redoing a fair amount of builds.
> 
> ```shell
> 0.623 go: github.com/siderolabs/go-smbios@v0.3.4 requires go >= 1.26.3 (running go 1.25.11; GOTOOLCHAIN=local)
> ```

we can ignore the go change and just pick up the fix only, the library bump was not really needed for this PR

majabojarska:
> > Not backporting to 1.12, since it would require bumping the Go toolchain to >=1.26.3 and redoing a fair amount of builds.
> > ```shell
> > 0.623 go: github.com/siderolabs/go-smbios@v0.3.4 requires go >= 1.26.3 (running go 1.25.11; GOTOOLCHAIN=local)
> > ```
> 
> we can ignore the go change and just pick up the fix only, the library bump was not really needed for this PR

Ack, will do that

smira:
but this reminds me - @frezbo - please don't bump Go version in `go.mod` of libraries until you _really_ need to. This causes this cascading dependency. go.mod declares only the minimum version, not the one which is going to be used.

frezbo:
> but this reminds me - @frezbo - please don't bump Go version in `go.mod` of libraries until you _really_ need to. This causes this cascading dependency. go.mod declares only the minimum version, not the one which is going to be used.

yup, my mistake, also i should have made this two separate commits

PR #13420, opened 2026-05-22
fix: perform backwards-compatible kernel args cleanup

smira:
Two changes:

* `nvme_core.io_timeout` is AWS-specific, move into AWS platform code
* `init_on_alloc` is enabled by default in the kernel config, so a kernel arg is no-op and confusing (as removing it won't disable it)

### DIFF
--- a/website/content/v1.14/reference/kernel.md
+++ b/website/content/v1.14/reference/kernel.md
@@ -16,11 +16,6 @@ Several of these are enforced by the Kernel Self Protection Project [KSPP](https
 * `slab_nomerge`: required by KSPP
 * `pti=on`: required by KSPP
 
-**Recommended** parameters:
-
-* `init_on_alloc=1`: advised by KSPP, enabled by default in kernel config
-* `init_on_free=1`: advised by KSPP, enabled by default in kernel config
-
 ### Available Talos-specific parameters
 
 #### `ip`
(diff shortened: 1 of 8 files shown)


website/content/v1.14/reference/kernel.md
```diff
@@ -18,8 +18,7 @@ Several of these are enforced by the Kernel Self Protection Project [KSPP](https
 
 **Recommended** parameters:
 
-* `init_on_alloc=1`: advised by KSPP, enabled by default in kernel config
-* `init_on_free=1`: advised by KSPP, enabled by default in kernel config
+* `init_on_free=1`: advised by KSPP, disabled by default in kernel config (`init_on_alloc=1` is enabled by default, but can be disabled with `init_on_alloc=0`)
```
frezbo: shouldn't be two separate points?

smira: how exactly you see it? `init_on_alloc=1` is the default, while `init_on_free=0` is the default

frezbo: this is all so confusing, previously was two separate points, making into one made it a bit confusing

> but can be disabled with `init_on_alloc=0`)

i think needs `-init_on_alloc init_on_alloc=0`

smira: This is the whole point of my change - no longer needed, as it's not in the kernel args anymore.

frezbo: let then drop it from docs, so it's less confusing, users can always enable kernel parameters that they want

frezbo (approved):

website/content/v1.14/reference/kernel.md
smira: yeah, good point, I will

PR #13421, opened 2026-05-22
fix: drop modprobe path and enforce usermode helper

smira:
See

* https://github.com/siderolabs/pkgs/pull/1565
* https://github.com/siderolabs/pkgs/pull/1569

Also bump Linux to 6.18.33 on the way.


frezbo (approved):
🤞

smira:
> One caveat worth flagging: this stops the controller loop, but functionally the kernel is now in BR_USER_STP mode waiting for a /sbin/bridge-stp daemon that doesn't exist in Talos. STP packets won't actually be processed. If real STP needs to work, the proper fix is kernel-side — either patch br_stp_start() to detect the no-op call_usermodehelper case and fall back to BR_KERNEL_STP, or carve out an exception for /sbin/bridge-stp in the static usermode helper. Worth raising upstream or in the pkgs PR.

This is a TODO for me.

internal/app/machined/main.go
```diff
@@ -344,7 +346,17 @@ func main() {
...
+	case "init", "machined":
+		// fall through to the main machined entrypoint
 	default:
+		// unknown name
+		if !containermode.InContainer() {
+			kmsg.SetupLogger(nil, "machined", nil) //nolint:errcheck // best effort logging to kmsg
+		}
+
+		log.Printf("unknown executable name %q (args %v)", os.Args[0], os.Args[1:])
+
```
frezbo, line 358: so we fail?

smira: yep, we need to - otherwise if we don't fail, kernel would think userspace bridge STP works. so we fail on whatever we don't support

frezbo: makes sense, I assumed it was a hard fail

frezbo (approved):

Issue #13424, opened 2026-05-24
Allow overriding kubelet kubeconfig tls-server-name when KubePrism is enabled (SNI-based ingress passthrough)


IvanHunters:
## Use case

Talos workers in one cloud, kube-apiserver hosted as a pod in a separate management cluster (e.g. Kamaji-managed control plane) and exposed to workers via `nginx-ingress` in `ssl-passthrough` mode on a per-cluster DNS hostname (`<cluster>.example.com:443`).

`cluster.controlPlane.endpoint` is set to that hostname, so KubePrism forwards `127.0.0.1:7445` → `<cluster>.example.com:443`. The apiserver certificate has `<cluster>.example.com` and `IP:127.0.0.1` in SANs.

## Problem

When `machine.features.kubePrism.enabled: true`, [`internal/app/machined/pkg/controllers/secrets/kubelet.go`](https://github.com/siderolabs/talos/blob/v1.13.0/internal/app/machined/pkg/controllers/secrets/kubelet.go) hardcodes the kubelet endpoint as `https://127.0.0.1:<port>`:

```go
case cfgProvider.Machine().Features().KubePrism().Enabled():
    localEndpoint, err := url.Parse(
        fmt.Sprintf("https://127.0.0.1:%d", cfgProvider.Machine().Features().KubePrism().Port()))
```

This makes Go's TLS client send `ServerName = "127.0.0.1"` in the ClientHello. Behind KubePrism (TCP-level proxy), the byte stream reaches the SNI-routing nginx ingress, which sees SNI=`127.0.0.1`, finds no matching ingress, and serves its default fake certificate (`DNS:ingress.local`, no IP SANs). kubelet then fails TLS validation:

```
Post "https://127.0.0.1:7445/api/v1/nodes": tls: failed to verify certificate:
x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs
```

The apiserver cert is valid for `127.0.0.1`, but kubelet never receives it: the proxy in the path serves a different cert because it routes by SNI.

This is independent of the closed regression in #8254 (which addressed the apiserver-side SAN). Here the apiserver SAN is correct; the path-on-the-wire is the issue.

## Proposal

Add a `tlsServerName` (or `serverName`) field to KubePrism config:

```yaml
machine:
  features:
    kubePrism:
      enabled: true
      port: 7445
      tlsServerName: cluster-xyz.example.com   # NEW
```

[...]

PR #13425, opened 2026-05-24
feat: allow overriding kubelet kubeconfig tls-server-name when KubePrism is enabled

IvanHunters:
# Pull Request

Refs siderolabs/talos#13424

## What? (description)

Add a new optional field `tlsServerName` to the KubePrism feature config:

```yaml
machine:
  features:
    kubePrism:
      enabled: true
      port: 7445
      tlsServerName: cluster-xyz.example.com   # new, optional, default empty
```

When set, the generated kubelet kubeconfig carries `clusters[0].cluster.tls-server-name` set to that value, so the kubelet uses it for SNI and certificate hostname verification. KubePrism keeps binding to `127.0.0.1:<port>` and the kubelet keeps dialing `https://127.0.0.1:<port>` as the TCP destination. Only the TLS SNI/hostname that the kubelet sends and verifies against is decoupled from the TCP target.

When the field is empty (default), behavior is unchanged.

## Why? (reasoning)

In `internal/app/machined/pkg/controllers/secrets/kubelet.go` the endpoint is hardcoded:

```go
localEndpoint, err := url.Parse(fmt.Sprintf("https://127.0.0.1:%d", cfgProvider.Machine().Features().KubePrism().Port()))
```

This forces SNI=`127.0.0.1` over the proxied connection. In setups where KubePrism's upstream is reached through an SNI-routing L4 proxy (for example nginx-ingress in `ssl-passthrough` mode in front of a Kamaji-hosted apiserver), SNI=`127.0.0.1` does not match any ingress route and the proxy serves a fallback certificate without IP SANs, so the kubelet rejects with:

```
cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs
```

Setting `tls-server-name` in the kubeconfig makes Go's `tls.Config.ServerName` (and therefore SNI) be the configured value, while the TCP destination stays loopback. This is the natural Kubernetes-native way to decouple the two, and aligns with how kubeconfigs already support this field.

## Changes

- New optional `tlsServerName` field on `KubePrism` in `pkg/machinery/config/types/v1alpha1`.
- New `TLSServerName() string` method on the `config.KubePrism` provider interface.
- New `EndpointTLSServerName` field on `secrets.KubeletSpec` resource (protobuf id `6`, `omitempty`), so the kubeconfig writer has access to the value.
- `secrets.KubeletController` populates `EndpointTLSServerName` from the KubePrism config when KubePrism is enabled, and leaves it empty otherwise.
- `k8s.KubeletServiceController.writePKI` sets `cluster.TLSServerName` on the bootstrap kubeconfig.
- `k8s.KubeletServiceController.updateKubeconfig` sets `cluster.TLSServerName` when refreshing the kubelet runtime kubeconfig.
- Regenerated `v1alpha1_types_doc.go` and updated the website config reference for `v1.14`.

## Acceptance

Notes on the unchecked items:

- `make conformance` and `make lint` go through the project's Docker/buildkit pipeline and were not run locally; CI will exercise them.
- Unit tests for the touched packages compile and run cleanly locally (`go test ./pkg/machinery/config/...` passes; secrets/k8s controllers compile cleanly with `GOOS=linux`).
- The new test case `TestReconcileKubePrismTLSServerName` in `internal/app/machined/pkg/controllers/secrets/kubelet_test.go` asserts that `EndpointTLSServerName` is propagated when configured.

## Test plan

- Build a Talos image with this change.
- On a cluster where KubePrism's upstream apiserver is fronted by an SNI-routing L4 proxy, set `machine.features.kubePrism.tlsServerName` to the apiserver's hostname.
- Verify `/etc/kubernetes/kubelet.conf` and `/etc/kubernetes/bootstrap-kubeconfig` contain `tls-server-name: <hostname>` under the cluster entry.
- Verify the kubelet successfully connects and registers the node.
- Verify that leaving `tlsServerName` unset produces an unchanged kubeconfig (no `tls-server-name` key) and behaves identically to current Talos.

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
+++ b/internal/app/machined/pkg/controllers/k8s/kubelet_service.go
@@ -208,7 +208,7 @@ func (ctrl *KubeletServiceController) Run(ctx context.Context, r controller.Runt
 			return err
 		}
 
-		if err = ctrl.updateKubeconfig(secretSpec.Endpoint, secretSpec.AcceptedCAs, logger); err != nil {
+		if err = ctrl.updateKubeconfig(secretSpec.Endpoint, secretSpec.EndpointTLSServerName, secretSpec.AcceptedCAs, logger); err != nil {
 			return err
 		}
 
@@ -306,6 +306,7 @@ func (ctrl *KubeletServiceController) writePKI(secretSpec *secrets.KubeletSpec) 
 		Clusters: map[string]*clientcmdapi.Cluster{
 			"local": {
 				Server:                   secretSpec.Endpoint.String(),
+				TLSServerName:            secretSpec.EndpointTLSServerName,
 				CertificateAuthorityData: acceptedCAs,
 			},
 		},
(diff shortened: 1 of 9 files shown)


smira:
Merged into #13820, thank you!

PR #13426, opened 2026-05-25
chore: script for tracking fixes made in upstream toolchain/tools/pkgs

frezbo:
Add a script to trace which commit in talos has a matching fix from upstream toolchain/tools/pkgs.


smira (approved):
prob

PR #13427, opened 2026-05-25
feat: add LVMService for VG/LV/PV removal

shanduur:
Operator-facing API to clear LVM metadata left behind after volume
teardown - pvremove is the only way to wipe an LVM label so the
underlying disk can be reused.

- storage.LVMService with LogicalVolumeRemove, VolumeGroupRemove and
  PhysicalVolumeRemove (unary, single target per call). Multi-node
  fan-out is client-side via multiplex.Unary; google.protobuf.Empty
  cannot be augmented by apid's one-to-many proxy path.
- internal/pkg/lvm: typed sentinels (ErrNotFound, ErrInUse,
  ErrNotEmpty, ErrOpen, ErrInvalidCommand, ErrInitFailed) classified
  in lvm.run. Stderr matchers are derived from lvm2 source and back
  the cases ECMD_FAILED=5 cannot disambiguate. ExecError carries
  the raw exit code and stderr for server-side logging without
  surfacing them to API clients.
- internal/app/lvmd: handler logs exit code + stderr on failure;
  client sees only the sanitized gRPC status.
- talosctl wipe lv|vg|pv subcommands with --insecure for maintenance.
  vg cascades to its LVs; documented in the long help.
- Integration tests provision LVM via nsenter, then exercise the
  full RPC chain; cleanup treats NotFound as success.


internal/app/lvmd/lvmd.go
```diff
@@ -0,0 +1,197 @@
...
+	all = append(all, fields...)
+	all = append(all, zap.Error(err))
+
+	if exec, ok := errors.AsType[*lvm.ExecError](err); ok {
+		all = append(
+			all,
+			zap.Int("lvm_exit_code", exec.ExitCode),
+			zap.ByteString("lvm_stderr", exec.Stderr),
+		)
+	}
```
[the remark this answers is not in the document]
shanduur: Outdated knowledge.

api/machine/lvm.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package storage;
+
+import "google/protobuf/empty.proto";
+
+option go_package = "github.com/siderolabs/talos/pkg/machinery/api/storage";
+option java_package = "dev.talos.api.storage";
+
+// LVMService manages LVM resources on a Talos node.
```
smira, line 10: can we put it under `api/machine` please? 

so that it becomes `machine.*` service - this way it's consistent with other APIs, and proxying for gRPC would work correctly for Omni

cmd/talosctl/cmd/talos/wipe.go
```diff
@@ -80,6 +84,137 @@ func wipeMethodValues() []string {
...
+// path (which can't augment Empty with per-node metadata).
+//
+// Maintenance mode is single-node by definition, so the fan-out collapses
+// to a single direct call there.
+//
+// The request is built per goroutine via `newReq` rather than shared by
+// pointer: protobuf messages cache marshaling state internally, and
+// concurrent reads of the same *Req across goroutines can race
+// (e.g. lazy memo updates inside generated GetXxx accessors).
+func fanOutLVMRemove[Req any, Resp any](
```
smira: can we use a common helper https://github.com/siderolabs/talos/blob/b88f16a529ff69a1b77ac3f6c14d90922399912e/cmd/talosctl/cmd/talos/multiplex/unary.go#L16-L18 ?

or adjust the helper if it's not good enough?

smira (approved):

PR #13428, opened 2026-05-25
fix: reset QEMU UEFI variable store when disk is wiped

Unix4ever:
When the guest disk is wiped externally and boot media is re-attached, UEFI keeps the previous Talos install's Boot#### entries in the variable store and tries them first, failing to fall through to the newly-attached ISO/USB/UKI. Recreate flash1.img alongside flash0.img in that path so the firmware starts from a clean variable store.


smira (approved):

smira:
This will probably allow us to run Sidero Metal/bare-metal-provisioner tests with UEFI - this was the problem previously (I think)... at least I 🙏  so

PR #13449, opened 2026-05-27
fix: create parent directories in ExtractTarGz for nested tar entries

immanuwell:
Same gap that was just fixed in `pkg/archiver/untar.go` (#1e31deda3) also exists in `cmd/talosctl/pkg/talos/helpers/archive.go` — the function used by `talosctl cp`.

`ExtractTarGz` uses `os.Mkdir` and never pre-creates parent dirs before writing files/symlinks. A valid tar archive with entries like `nested/path/file.txt` but no explicit dir headers will blow up with:

```
error creating file "nested/path/file.txt" mode -rw-r--r--: open nested/path/file.txt: no such file or directory
```

## Reproduce

```go
var buf bytes.Buffer
gzw := gzip.NewWriter(&buf)
tw := tar.NewWriter(gzw)
tw.WriteHeader(&tar.Header{Name: "nested/path/file.txt", Mode: 0o644, Size: 5})
tw.Write([]byte("hello"))
tw.Close(); gzw.Close()

// fails before this fix
helpers.ExtractTarGz("/tmp/out", io.NopCloser(bytes.NewReader(buf.Bytes())))
```

## Changes

- `os.Mkdir` -> `os.MkdirAll` for dir entries
- pre-create parent dirs before symlinks and regular files
- `%s` -> `%w` in error wrap
- drop stale TODO (already handled by `safepath.CleanPath`)
- add regression test

### DIFF
--- a/cmd/talosctl/pkg/talos/helpers/archive.go
+++ b/cmd/talosctl/pkg/talos/helpers/archive.go
@@ -74,7 +74,7 @@ func ExtractTarGz(localPath string, r io.ReadCloser) error {
 				break
 			}
 
-			return fmt.Errorf("error reading tar header: %s", err)
+			return fmt.Errorf("error reading tar header: %w", err)
 		}
 
 		hdrPath := safepath.CleanPath(hdr.Name)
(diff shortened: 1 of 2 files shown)


smira:
I don't think this is valid, as this operates only on archives created by Talos, and they have intermediate entries. 

If you have a bug report/reproducer for the problem, we can reconsider this, but for now I'm going to close it.

PR #13451, opened 2026-05-27
fix: suppress ICMP redirects by default

smira:
Make one of the CIS benchmark items enabled by default.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -139,6 +139,15 @@ Talos now supports mounting and provisioning `btrfs` filesystem for user volumes
 Support for `btrfs` is enabled by installing `btrfs` system extension.
 """
 
+    [notes.sendRedirects]
+        title = "ICMP send_redirects Disabled by Default"
+        description = """\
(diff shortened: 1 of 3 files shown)


frezbo (approved):

hack/release.toml
```diff
@@ -137,6 +137,15 @@ Talos is now compatible with etcd v3.6.x only (the default version was etcd 3.6.
 Talos now supports mounting and provisioning `btrfs` filesystem for user volumes and existing volumes.
 
 Support for `btrfs` is enabled by installing `btrfs` system extension.
+"""
+
+    [notes.sendRedirects]
+        title = "ICMP send_redirects Disabled by Default"
+        description = """\
+Talos now sets `net.ipv4.conf.all.send_redirects=0` and `net.ipv4.conf.default.send_redirects=0` by default,
```
frezbo, line 145: no need for ipv6 then?

smira: it looks to be IPv4 only feature afaik

PR #13459, opened 2026-05-28
fix: align flannel MTU with kubespan to avoid permanent fragmentation

### LINKED ISSUE #13457: flannel vs kubespan MTU
This is both a bug report and a feature request, I suppose.

## Bug Report

Flannel is not aware of kubespan in the path and sets an MTU that causes permanent packet fragmentation, potentially leading to degraded network performance.

### Description

Original issue:
When deploying a talos cluster (v1.12) I noticed massive packet fragmentation.

It turns out that when using the default `flannel` with `kubespan` enabled, flannel sets an `MTU` of `1450` and `kubespan` uses the default of `1420`. This will cause permanent fragmentation, as flannel is not aware of kubespan in the networking path.

If kubespan is enabled: `flannel` should use an MTU of `kubespan.MTU - 50` instead of `interface.MTU - 50` to avoid permanent fragmentation. The default being 1420 for kubespan MTU, flannel should use an MTU of `1420 - 50 (VXLAN header size) = 1370` by default.

Alternatively (better): talos machineconfig should offer admins to manually set the MTU used by flannel or point to the master interface.

Related issue:
We have `rook-ceph` running in hostnet mode, talking to other node IPs directly. Kubespan has a very aggressive nftables rule, that forces all traffic regardless of origin to use kubespan, if the destination IP is a kubespan endpoint (the node IP of another node). As `rook-ceph` runs on hostnet, it uses the default MTU of whatever interface it binds to, in this case one with MTU of 1500.

The aggressive nft rule of kubespan captures more traffic than it should, breaking PMTU in the process. This can pose a problem for high throughput, time sensitive services like rook-ceph.

I am not 100% sure of this, gut there is another related issue, I think this can also "poison" interfaces that have a much higher MTU (say 9000):
if traffic traverses such an interface, and is then (qua nft rule, based on IP `dst`) routed via kubespan, this will trigger an ICMP "fragmentation needed" response.
This will affect _all_ interfaces on the routing path to take on the minimum MTU signaled.
If an upstream interface is then used for traffic that is NOT affected by the kubespan NFT rule, the "poisoned" MTU still applies interface wide, causing fragmentation even on traffic completely unrelated to kubespan.
On top of this, this MTU clamping is on a timer and will expire - causing the process to periodically repeat itself, which every time can cause a massive network congestion on busy/high volume routes (again, rook-ceph being an example as it is sensitive to this kind of thing).

### Logs

from a privileged debug pod on a worker node:

1. flannel MTU (1450):
```
$ ip link show flannel.1
14: flannel.1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UNKNOWN mode DEFAULT group default
    link/ether b2:00:dd:c4:a0:1c brd ff:ff:ff:ff:ff:ff
```

2. kubespan MTU (1420):
```
ip link show kubespan
12: kubespan: <POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1420 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/none
```

This graph shows the FRAG sockets in use by that an affected machine:

Image

### Environment

- Talos version: `v1.12.7`
- Kubernetes version: `v1.35.4`
- Platform: bare metal (workers), EC2 (controlplane)

fsgh42:
When kubespan is enabled, flannel auto-detects its underlay MTU from
the host's primary interface (typically 1500) and creates flannel.1
at MTU 1450 (= 1500 - 50 VXLAN overhead). Encapsulated VXLAN packets
are then routed through kubespan's WireGuard tunnel (default MTU
1420), forcing permanent fragmentation and degraded throughput.

Fix it in two parts:

* Render `Backend.MTU` in flannel's net-conf.json from a new
  `cluster.network.cni.flannel.mtu` field. Counter-intuitively,
  flannel's `Backend.MTU` represents the *underlay* transport MTU,
  not the resulting flannel.1 MTU — flannel internally subtracts
  the VXLAN overhead when creating the device. See:
  https://github.com/flannel-io/flannel/blob/master/Documentation/backends.md#vxlan

* In ControlPlaneBootstrapManifestsController, default
  `Backend.MTU` to the kubespan link MTU when kubespan is enabled
  and the field is unset. Explicit user override always wins.

With the default kubespan MTU of 1420, flannel.1 ends up at 1370,
so encapsulated VXLAN traffic fits within the WireGuard MTU without
fragmentation. When kubespan is disabled and the field is unset,
behavior is unchanged: `Backend.MTU` is omitted from net-conf.json
and flannel auto-detects from the external interface as before.

Fixes #13457

Signed-off-by: Fritz Schaal <fritz.schaal@siderolabs.com>
Assisted-by: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/api/resource/definitions/k8s/k8s.proto
+++ b/api/resource/definitions/k8s/k8s.proto
@@ -80,6 +80,7 @@ message BootstrapManifestsConfigSpec {
   bool flannel_kube_network_policies_enabled = 19;
   string flannel_kube_network_policies_image = 20;
   string cni_name = 21;
+  uint32 flannel_mtu = 22;
 }
 
 // ConfigStatusSpec describes status of rendered secrets.
--- a/internal/app/machined/pkg/controllers/k8s/control_plane.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane.go
@@ -306,6 +306,25 @@ func NewControlPlaneSchedulerController() *ControlPlaneSchedulerController {
 	)
 }
 
+// resolveFlannelMTU returns the underlay MTU flannel should advertise as
+// Backend.MTU in its net-conf.json. Explicit user override wins; otherwise,
+// if KubeSpan is enabled, default to the KubeSpan link MTU so that the
+// encapsulated VXLAN traffic fits within the WireGuard tunnel without
+// permanent fragmentation. Returns 0 when no override applies, which omits
+// the field from net-conf.json (flannel then auto-detects from the external
+// interface as it does today).
+func resolveFlannelMTU(cfgProvider talosconfig.Config) uint32 {
+	if mtu := cfgProvider.Cluster().Network().CNI().Flannel().MTU(); mtu != 0 {
+		return mtu
+	}
+
+	if ks := cfgProvider.NetworkKubeSpanConfig(); ks != nil && ks.Enabled() {
+		return ks.MTU()
+	}
+
+	return 0
+}
+
 // ControlPlaneBootstrapManifestsController manages k8s.BootstrapManifestsConfig based on configuration.
 type ControlPlaneBootstrapManifestsController = transform.Controller[*config.MachineConfig, *k8s.BootstrapManifestsConfig]
 
@@ -379,6 +398,7 @@ func NewControlPlaneBootstrapManifestsController() *ControlPlaneBootstrapManifes
 					FlannelKubeServicePort:            flannelKubeServicePort,
 					FlannelKubeNetworkPoliciesEnabled: cfgProvider.Cluster().Network().CNI().Flannel().KubeNetworkPoliciesEnabled(),
 					FlannelKubeNetworkPoliciesImage:   images.KubeNetworkPolicies.String(),
+					FlannelMTU:                        resolveFlannelMTU(cfgProvider),
 
 					TalosAPIServiceEnabled: cfgProvider.Machine().Features().KubernetesTalosAPIAccess().Enabled(),
 
--- a/internal/app/machined/pkg/controllers/k8s/internal/k8stemplates/flannel.go
+++ b/internal/app/machined/pkg/controllers/k8s/internal/k8stemplates/flannel.go
@@ -147,12 +147,14 @@ func FlannelConfigMapTemplate(spec *k8s.BootstrapManifestsConfigSpec) runtime.Ob
 		Backend        struct {
 			Type string `json:"Type"`
 			Port int    `json:"Port"`
+			MTU  uint32 `json:"MTU,omitempty"`
 		} `json:"Backend"`
 	}
 
 	netConf.EnableNFTables = new(true)
 	netConf.Backend.Type = "vxlan"
 	netConf.Backend.Port = 4789
+	netConf.Backend.MTU = spec.FlannelMTU
 
 	hasIPv4 := false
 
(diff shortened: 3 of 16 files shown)


fsgh42:
tested like so:

prerequisite: I have local registry proxies running at port 5000, 5001 etc

```
make talosctl
make initramfs kernel
make installer-base IMAGE_REGISTRY=127.0.0.1:5007 PUSH=true
make imager        IMAGE_REGISTRY=127.0.0.1:5007 PUSH=true INSTALLER_ARCH=targetarch
make installer     IMAGE_REGISTRY=127.0.0.1:5007 PUSH=true
set INSTALLER_TAG $(git describe --tag --always --dirty --match 'v[0-9]*')
sudo --preserve-env=HOME _out/talosctl-linux-amd64 \
  --talosconfig=_out/talosconfig \
  cluster create dev \
  --name=dev \
  --cidr=172.20.0.0/24 \
  --registry-mirror docker.io=http://172.20.0.1:5000 \
  --registry-mirror registry.k8s.io=http://172.20.0.1:5001 \
  --registry-mirror gcr.io=http://172.20.0.1:5003 \
  --registry-mirror ghcr.io=http://172.20.0.1:5004 \
  --registry-mirror factory.talos.dev=http://172.20.0.1:5005 \
  --registry-mirror quay.io=http://172.20.0.1:5006 \
  --registry-mirror 127.0.0.1:5007=http://172.20.0.1:5007 \
  --install-image=127.0.0.1:5007/siderolabs/installer:$INSTALLER_TAG \
  --controlplanes 3 --workers 2 --with-bootloader=false \
  --with-kubespan

sudo chown "$USER:$USER" _out/talosconfig _out/talosctl-linux-amd64  --talosconfig=_out/talosconfig kubeconfig _out/kubeconfig

export TALOSCONFIG=$PWD/_out/talosconfig
export KUBECONFIG=$PWD/_out/kubeconfig
```

As kubespan is on by default, we'd expect the new flannel MTU to now be 1370:

```
$ _out/talosctl-linux-amd64 -n 172.20.0.2 get links kubespan -oyaml | yq .spec.mtu
1420
$ kubectl -n kube-system get cm kube-flannel-cfg -o yaml | yq .data
cni-conf.json: |-
  {
    "name": "cbr0",
    "cniVersion": "1.0.0",
    "plugins": [
      {
        "type": "flannel",
        "delegate": {
          "hairpinMode": true,
          "isDefaultGateway": true
        }
      },
      {
        "type": "portmap",
        "capabilities": {
          "portMappings": true
        }
      }
    ]
  }
net-conf.json: |-
  {
    "Network": "10.244.0.0/16",
    "EnableNFTables": true,
    "Backend": {
      "Type": "vxlan",
      "Port": 4789,
      "MTU": 1420
    }
  }
$ _out/talosctl-linux-amd64 -n 172.20.0.2 get links flannel.1 -oyaml | yq .spec.mtu
1370
```

this looks correct to me.

fsgh42:
I also ran 
```
for TGT in fmt lint unit-tests generate docs ; make $TGT ; end
```

to update generated files and docs. Some linting and testing errors surfaced, but they seem unrelated to my changes. Let me know if there is more I can do.

smira:
thank you, it makes sense, but let's hold on this until we refactor the config, I don't want to introduce new changes to old config for now

fsgh42:
> thank you, it makes sense, but let's hold on this until we refactor the config, I don't want to introduce new changes to old config for now

I'll happily leave it to you guys :+1:

smira:
Closing this to re-open from a fork.

PR #13463, opened 2026-05-28
feat: update pkgs, tools, Go dependencies

smira:
Brings in new LLVM & gcc built binaries.

Update Go dependencies and test dependencies.


.github/workflows/lock.yaml
frezbo, line 1: this seems weird?

.github/workflows/stale.yaml
frezbo, line 1: same here

frezbo (approved):

.github/workflows/lock.yaml
smira: these are stale files - I added them manually first (before kres), so we had two instances of these, one with `.yaml` and another with `.yml` - I cleaned up the stale ones, the kres ones are still here

frezbo (approved):

PR #13465, opened 2026-05-29
fix: add --fail to image-signer curl, fix %w error wrapping in two controllers

immanuwell:
## What? (description)

- Add `--fail` to the `image-signer` curl download (Makefile)
- Fix `%s` -> `%w` in `fmt.Errorf` calls in `cmdline.go` and `watchdog_timer.go`

## Why? (reasoning)

PR #13447 added `--fail` to curl calls to catch HTTP errors, but the `image-signer` download got missed. Without `--fail`, a 404 silently writes HTML to the binary, `chmod +x` still succeeds, and `make sign-images` blows up with "Exec format error" later - annoying to debug.

Reproduce:

```sh
IMAGE_SIGNER_RELEASE=v0.0.0-bogus make _out/image-signer-v0.0.0-bogus
# before: exits 0, file is an HTML error page
# after:  curl exits non-zero right away
```

The `%s` -> `%w` fixes are consistency cleanup - `cmdline.go` and `watchdog_timer.go` use `%w` everywhere except in a couple spots. Broken error wrapping silently breaks `errors.Is`/`errors.As` chains.

## Acceptance

- [ ] you linked an issue (if applicable) - n/a
- [ ] you included tests (if applicable) - n/a
- [ ] you generated documentation (`make docs`) - n/a


smira (approved):
thank you!

PR #13466, opened 2026-05-29
fix: drop installer from bundle

shanduur:
Drop installer image from talos bundle - this image is no longer
published, and should not be present in release notes.


cmd/talosctl/cmd/talos/image.go
```diff
@@ -591,7 +591,6 @@ var imageTalosBundleCmd = &cobra.Command{
 
 		sources := images.ListSourcesFor(tag)
 
-		fmt.Printf("%s\n", sources.Installer)
```
smira, line 594: should we drop only if 1.14+?

smira (approved):

PR #13472, opened 2026-05-29
feat: add custom logging convention linter

shanduur:
Enforce structured logging rules in CI and keep exceptions explicit
while Talos still carries legacy logging paths.

- add a standalone and golangci-lint plugin-backed log linter
- wire a custom golangci-lint binary into Docker and make targets
- enable repo-wide logging rules with scoped allowlists
- fix some of current logging violations needed to pass the new checks


Unix4ever (approved):
We should use it in Omni too :)

internal/app/machined/pkg/system/services/machined.go
```diff
@@ -216,7 +216,7 @@ func (s *machinedService) Main(ctx context.Context, _ runtime.Runtime, logWriter
 			logWriter, zapcore.DebugLevel,
 			logging.WithColoredLevels(),
 		),
-	)
+	).With(logging.Component("machined"))
```
smira, line 219: this is a root logger, I'm not sure if it should have a component at this point

smira: hold on, ignore me, it's not a root logger here

Dockerfile
```diff
@@ -1322,31 +1322,42 @@ COPY --link --from=integration-test-provision-linux-build /src/integration.test
 # All depend on lint-go-config (gating) by bind-mounting /verified from it.
 # Cache mounts: Go cache shared (concurrency-safe), lint cache locked (golangci-lint corruption protection).
 
+FROM base AS lint-golangci-lint-custom
+RUN --mount=type=cache,target=/.cache,id=talos/.cache,sharing=shared \
```
smira, line 1341: is there a way to build it as a plugin some other way, so that `go tool golangci-lint` still works (or `go tool golangci-lint-custom`? it's a very good shortcut to run it in the editor outside of the Dockerfile?

shanduur: You can extract the logger config to separate file, and call `go tool loglinter` directly.

internal/app/machined/pkg/system/services/machined.go
shanduur: IMHO even a root logger should have a component. It is overridden whenever we will call `.With(logging.Component("other-component"))`, and we have a more strict convention that is enforced everywhere.

Dockerfile
shanduur: If you meant the `go tool golangci-lint`, there is a `make golangci-lint-custom`, after which you can call `_out/custom-gcl run --config .golangci.yml`

PR #13481, opened 2026-06-01
fix: recreate dns server and listeners on host DNS runner restart

jleeh:
## Problem

On a node running Talos v1.13.2 we hit a permanent host-DNS outage. `dns-resolve-cache` logged this in a tight loop and never recovered:

```
dns-resolve-cache-runners: Failed service '&dns.Runner{srv:(*dns.Server)(0x174496be4360), ...}'
  (5.99 failures of 5.00), restarting: false, error: dns: server already started
```

Note the `*dns.Server` pointer is identical across the entire multi-hour run — the same server object is restarted forever. Nothing stayed bound to `127.0.0.53:53`, so host DNS was down until the node was rebooted. Toggling `machine.features.hostDNS.enabled` false→true (which tears the runner down and recreates it) also clears it.

## Root cause

`Runner` builds one `*dns.Server` — together with its `net.Listener`/`net.PacketConn` — once in `NewRunner` and reuses it on every supervisor restart. Both are single-use:

- When `ActivateAndServe()` returns on its own (a transient serve error), `Serve()` takes the first `select` case and returns the error **without** calling `close()`, so `ShutdownContext()` never runs and miekg/dns leaves the server flagged `started`.
- `suture` restarts the runner, `ActivateAndServe()` runs against the already-started server and fails immediately with `dns: server already started` — and that path again skips `close()`.
- Every retry repeats this; the runner exhausts its failure budget and gives up. (miekg/dns also closes the listener when the serve loop exits, so the listener can't be reused either.)

## Fix

Construct the `dns.Server` and its listeners lazily via an opts factory invoked on **every** `Serve()` call, so each (re)start gets fresh, restartable instances. The result channel is also buffered so the serve goroutine can't leak if `Serve()` returns via the close timeout.

## Behaviour change

Runner option construction (which binds sockets) now happens inside `Serve()` rather than eagerly in `RunAll`. A persistent construction error therefore surfaces through the supervisor event hook and is retried, rather than being returned synchronously from `RunAll` as `StatusNew`'s error. This also makes a transient bind failure (e.g. address briefly in use during reconfiguration) self-heal instead of being reported once.

## Testing

- `go test ./internal/pkg/dns/...` passes (incl. `goleak`).
- Added `TestRunnerRestart`, which starts/stops/restarts a `Runner` and asserts it never returns `dns: server already started`. The two start cycles bind distinct ephemeral ports, confirming fresh listeners per restart. This test fails against the previous reuse-based implementation.


jleeh:
Thanks for the review. Responses to the three points (pushed `6ea50fb`):

**`runnner.go` — `close()` 5s vs `Serve` 1s wait:** these are sequential, not concurrent. `close()` runs to completion first, and `ShutdownContext` is what makes `ActivateAndServe()` return — so by the time `close()` returns, the serve goroutine has already sent on `errCh` (now buffered) and the post-close `select` reads it immediately. The 1s timeout only fires if serve has still not returned *after* shutdown completed — an edge case, and this `select` is unchanged by the PR. The one related change here is buffering `errCh`, which removes the goroutine leak the old unbuffered channel had on that timeout path. Left as-is.

**`manager.go` — `RunAll` yields `StatusNew`/`nil` even though opts construction can fail:** deliberate trade-off. The listeners and `*dns.Server` are single-use and must be rebuilt per (re)start, so binding has to move into `Serve()`; there is nothing to bind, and therefore no error to surface, at `RunAll` time. Persistent failures are now reported via the supervisor event hook and retried instead of failing the reconcile. Note the status was already optimistic before this change — `"running"` only ever reflected the *initial* bind, and suture-managed restarts after that were never re-surfaced to the controller either. Surfacing live runner health back to the manager/controller is a worthwhile follow-up but a larger change than this outage fix; happy to do it separately if preferred.

**`dns_test.go` — comment overstates "actually serving":** fixed. Reworded to state the assertion checks the runner did not exit early (e.g. `"already started"` on the second start) rather than implying it verifies live serving.

PR #13500, opened 2026-06-02
refactor: replace the callback strategy for most commands

### LINKED ISSUE #13492: refactor: mixed maintenance/normal API commands
There are some commands which have mixed maintenance/"normal" APIs, these should be also refactored and adjusted for client-side multiplexing.

### LINKED ISSUE #13493: refactor: users of helpers.ForEachResource
Use `client.WithNode`, it's almost there, but requires updating each command

### LINKED ISSUE #13498: talosctl auto-completetion does not work as expected
## Bug Report

doing `talosctl` <TAB> auto-completes, but not `talosctl -n <node ip>` <TAB>

smira:
Fixes #13493

Fixes #13498 

Fixes #13492

Instead of using `With*` helpers, switch to a generic `ClientFactory` which can build any kind of clients in any possible scenario:

* clients with "normal" mode, proxying over endpoints using `WithNdoe`
* clients in maintenance mode, where each node becomes an endpoint

Fix numerous bugs and inconsistencies (probably introduced some new bugs on the way 🤣 )

The next PR will remove all `With*` helpers, refactor execution tracker, all streaming commands following the same pattern, bringing the whole code to a state of consistency.

### DIFF
--- a/cmd/talosctl/cmd/talos/patch.go
+++ b/cmd/talosctl/cmd/talos/patch.go
@@ -24,7 +24,6 @@ import (
 	"github.com/siderolabs/talos/pkg/machinery/client"
 	"github.com/siderolabs/talos/pkg/machinery/config/configpatcher"
 	"github.com/siderolabs/talos/pkg/machinery/constants"
-	"github.com/siderolabs/talos/pkg/machinery/resources/config"
 )
 
 var patchCmdFlags struct {
@@ -70,20 +69,8 @@ func extractMachineConfigBody(mc resource.Resource) ([]byte, error) {
 	return []byte(bodyStr), nil
 }
 
-func patchFn(c *client.Client, patches []configpatcher.Patch) func(context.Context, string, resource.Resource, error) error {
-	return func(ctx context.Context, node string, mc resource.Resource, callError error) error {
-		if callError != nil {
-			return fmt.Errorf("%s: %w", node, callError)
-		}
-
-		if mc.Metadata().Type() != config.MachineConfigType {
-			return fmt.Errorf("%s: unsupported resource type: %s", node, mc.Metadata().Type())
-		}
-
-		if mc.Metadata().ID() != config.ActiveID {
-			return nil
-		}
-
+func patchFn(patches []configpatcher.Patch) func(context.Context, *client.Client, string, resource.Resource) error {
+	return func(ctx context.Context, c *client.Client, node string, mc resource.Resource) error {
 		body, err := extractMachineConfigBody(mc)
 		if err != nil {
 			return err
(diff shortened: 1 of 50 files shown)


cmd/talosctl/cmd/talos/upgrade.go
```diff
@@ -370,7 +359,6 @@ func init() {
 	upgradeCmdFlags.addTrackActionFlags(upgradeCmd)
 	upgradeCmd.Flags().BoolVar(&upgradeCmdFlags.legacy, "legacy", false, "force use of legacy upgrade method")
 	upgradeCmd.Flags().BoolVarP(&upgradeCmdFlags.force, "force", "f", false, "force the upgrade (skip checks on etcd health and members, might lead to data loss)")
-	upgradeCmd.Flags().BoolVar(&upgradeCmdFlags.insecure, "insecure", false, "upgrade using the insecure (encrypted with no auth) maintenance service")
```
smira, line 373: I dropped this, this was hacky, and now we have proper integration tests for upgrade in maintenance

cmd/talosctl/cmd/talos/reset.go
```diff
@@ -187,7 +178,6 @@ func resetGetActorID(ctx context.Context, c *client.Client, req *machineapi.Rese
 func init() {
 	resetCmd.Flags().BoolVar(&resetCmdFlags.graceful, "graceful", true, "if true, attempt to cordon/drain node and leave etcd (if applicable)")
 	resetCmd.Flags().BoolVar(&resetCmdFlags.reboot, "reboot", false, "if true, reboot the node after resetting instead of shutting down")
-	resetCmd.Flags().BoolVar(&resetCmdFlags.insecure, "insecure", false, "reset using the insecure (encrypted with no auth) maintenance service")
```
smira, line 190: I dropped this for the same reason - it's hacky, and we have proper integration test now

cmd/talosctl/cmd/talos/patch.go
```diff
@@ -70,20 +69,8 @@ func extractMachineConfigBody(mc resource.Resource) ([]byte, error) {
 	return []byte(bodyStr), nil
 }
 
-func patchFn(c *client.Client, patches []configpatcher.Patch) func(context.Context, string, resource.Resource, error) error {
-	return func(ctx context.Context, node string, mc resource.Resource, callError error) error {
-		if callError != nil {
```
smira, line 75: these checks were moved into a new helper function

cmd/talosctl/cmd/talos/patch.go
```diff
@@ -134,33 +121,29 @@ var patchCmd = &cobra.Command{
...
-			}
+		ctx := cmd.Context()
 
-			for _, node := range GlobalArgs.Nodes {
-				nodeCtx := client.WithNodes(ctx, node) //nolint:staticcheck // to be refactored next
-				if err := helpers.ForEachResource(nodeCtx, c, nil, patchFn(c, patches), patchCmdFlags.namespace, args...); err != nil {
-					return err
-				}
-			}
+		clientFactory := NewClientFactory(&editCmdFlags)
```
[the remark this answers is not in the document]
pranav767: Not related to above comment, but I can see `commit suggestion` button enabled for me. Don't know if it actually does commit or gives error.

frezbo:
This fixes completion? or was a wrong issue linked?

smira:
> This fixes completion? or was a wrong issue linked?

yes, it does. it was failing silently, as we don't build a client when `--nodes` is not set.

`talosctl get -n ` should auto-complete now; `talosctl -n ` doesn't auto-complete as it's not a global flag.

cmd/talosctl/pkg/talos/global/factory.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	f.client = c
+
+	return client.WithNode(ctx, node), c, nil
+}
+
+// BuildClientFirstNode is a helper which builds a client for the first node in the list.
+func (f *ClientFactory) BuildClientFirstNode(ctx context.Context) (context.Context, *client.Client, error) {
+	// the nodes is list is non-empty (see NewClientFactory), so this is safe
```
frezbo:
```suggestion
	// the nodes list is non-empty (see NewClientFactory), so this is safe
```

cmd/talosctl/pkg/talos/helpers/resources.go
```diff
@@ -7,95 +7,74 @@ package helpers
...
+	args []string,
 ) error {
 	if len(args) == 0 {
 		return errors.New("not enough arguments: at least 1 is expected")
 	}
 
 	resourceType := args[0]
 
-	var resourceID string
+	// ignore args[1], always patch in current machine config resource ID
```
frezbo, line 34: then make exact args as 1, isn't it easier?

frezbo (approved):

cmd/talosctl/pkg/talos/helpers/resources.go
smira: I don't want to break compatibility, previously it worked like `talosctl patch mc v1apha1` and `talosctl patch`. 

I don't think it's a good idea that we had such syntax, but we can keep it working with this simple workaround

PR #13504, opened 2026-06-03
fix: classify installer and imager exits

### LINKED ISSUE #13503: Surface installer/imager errors through Exit Codes
Currently, the installer/imager exits with a generic `exit(1)` on any failure. This makes it impossible for callers (CI, automation, users, orchestration layers) to reliably distinguish between different failure modes.

As a result:
* All errors are indistinguishable at the process level
* Automation cannot react differently to different failure types
* Debugging requires log parsing instead of structured signals

We need to introduce a strict, stable exit-code mapping driven by tagged errors.

* Define a set of **API-stable exit codes**
* Exit codes must be:
  * Semantically defined
  * Backward compatible (meaning must never change)
  * Documented as part of the public API

Each exit code represents a **category of failure**, not a single error.

Use tagged errors to map runtime errors to exit codes:

* Errors should be tagged using:
    ```go
    xerrors.NewTagged[Tag](err)
    ```

* Each `Tag` maps to one or more exit codes
* Each exit code maps to one or more tags
* Any error without a tag is treated as an **unknown error**
* Unknown errors must map to a predefined fallback exit code (e.g. `EXIT_UNKNOWN_ERROR`)

On process termination:

1. Inspect error chain
2. Extract all `Tag`s via `xerrors`
3. Resolve corresponding exit code:

   * If one or more tags exist -> use mapped exit code(s)
   * If multiple tags map to different codes -> define deterministic resolution strategy (e.g. priority order or first-match)
   * If no tags exist -> use `EXIT_UNKNOWN_ERROR`
4. Exit with resolved code instead of `exit(1)`

## Requirements

* Exit codes must be centrally defined and versioned as API contract
* Tag -> exit code mapping must be explicit and documented
* Exit codes must remain stable across releases
* Tags should be extensible without breaking existing mappings
* Must support multiple tags per error
* Must define deterministic behavior for multiple matching exit codes

shanduur:
Make automation distinguish invalid input, unsupported configs, host
environment issues, dependency failures, I/O problems, and install
runtime failures instead of treating every failure as exit code 1.

Fixes #13503

### DIFF
--- a/cmd/installer/cmd/imager/root.go
+++ b/cmd/installer/cmd/imager/root.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/siderolabs/gen/xerrors"
 	"github.com/siderolabs/gen/xslices"
 	"github.com/spf13/cobra"
 	"go.yaml.in/yaml/v4"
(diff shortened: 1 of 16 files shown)


oguzkilcan (approved):

frezbo (commented):
🆒

smira (approved):

PR #13506, opened 2026-06-03
fix: handle cluster-scoped resources with a namespace correctly

### LINKED ISSUE #13453: v1.13 resourceManager unexpected object
## Bug Report

After upgrading from talos v1.12.7 to v1.13.3 the update of inline manifests (talosctl upgrade-k8s) fails with errors like:
```
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/argocd-server
```
I tried removing the object but then I only get `returned "created" for unexpected object` and then it goes back to `"unchanged"`. Is there some special handling required for cluster-scoped objects?

### Description

In order to bootstrap our clusters we apply CNI, CSI and ArgoCD via inline manifests. This has worked really well but it looks like we hit some corner case with the new SSA in 1.13.

### Logs

```
updating manifests
error applying manifests: resourceManager returned "unchanged" for unexpected object CustomResourceDefinition/applications.argoproj.io
resourceManager returned "unchanged" for unexpected object CustomResourceDefinition/applicationsets.argoproj.io
resourceManager returned "unchanged" for unexpected object CustomResourceDefinition/appprojects.argoproj.io
resourceManager returned "unchanged" for unexpected object ClusterRole/argocd-application-controller
resourceManager returned "unchanged" for unexpected object ClusterRole/argocd-applicationset-controller
resourceManager returned "unchanged" for unexpected object ClusterRole/argocd-server
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/argocd-application-controller
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/argocd-applicationset-controller
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/argocd-argocd-vault
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/argocd-server
resourceManager returned "unchanged" for unexpected object ClusterRoleBinding/vmware-system-csi-csi-vsphere-vault
```

### Environment

- Talos version: v1.13.3
- Kubernetes version: 1.35.5
- Platform: vmware

Orzelius:
bump to-kubernetes that contains a fix and remove the ns on cluster-scoped resources such that they get correctly entered into the inventory

closes #13453

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/manifest_apply.go
+++ b/internal/app/machined/pkg/controllers/k8s/manifest_apply.go
@@ -278,16 +278,6 @@ func (ctrl *ManifestApplyController) apply(
 		gvk := obj.GroupVersionKind()
 		objName := fmt.Sprintf("%s/%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind, obj.GetName())
 
-		objMeta, err := object.RuntimeToObjMeta(obj)
-		if err != nil {
-			return nil, fmt.Errorf("failed to retrieve object metadata of %q: %w", objName, err)
-		}
-
-		// check if the resource is already in the inventory, if so, skip applying it
-		if inv.Contains(objMeta) {
-			continue
-		}
-
 		mapping, err := mapper.RESTMapping(obj.GroupVersionKind().GroupKind(), obj.GroupVersionKind().Version)
 		if err != nil {
 			switch {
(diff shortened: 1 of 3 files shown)


frezbo (approved):

internal/app/machined/pkg/controllers/k8s/manifest_apply.go
```diff
@@ -319,6 +309,19 @@ func (ctrl *ManifestApplyController) apply(
...
+			obj.SetNamespace("")
+		}
+
+		objMeta, err := object.RuntimeToObjMeta(obj)
+		if err != nil {
+			return nil, fmt.Errorf("failed to retrieve object metadata of %q: %w", objName, err)
+		}
+
+		// check if the resource is already in the inventory, if so, skip applying it
+		if inv.Contains(objMeta) {
```
smira, line 336: can we also cleanup from the inventory if it has already namespaced entry for cluster-scoped resource?

smira: I'm not sure if we need to do it here or in `go-kubernetes`, but the user report was that even if you remove `namespace:` for cluster-scoped resources, they seem to stay in the inventory after repeated calls to `talosctl upgrade-k8s`?

Orzelius: Done! 

Good idea to remove on our side for users for whom the incorrect entry was created

internal/app/machined/pkg/controllers/k8s/manifest_apply.go
```diff
@@ -319,6 +309,33 @@ func (ctrl *ManifestApplyController) apply(
...
 			dr = dyn.Resource(mapping.Resource)
+
+			objMeta, err := object.RuntimeToObjMeta(obj)
+			if err != nil {
+				return nil, fmt.Errorf("failed to retrieve object metadata of %q: %w", objName, err)
+			}
+
+			// Talos v1.13 inserted the cluster-wide resources to the inventory with a namespace if they had one set, remove them from the inventory
+			if obj.GetNamespace() != "" && inv.Contains(objMeta) {
+				fmt.Printf("removing %s from inventory", obj.GetName())
```
smira: certainly not `fmt.Printf`, we have a logger, and should have a context on why this is happening

bit I think it should be actually in go-kubernetes on sync, as it also applies to Omni's Kubernetes manifest sync same way, and this fix will not fix all of it

Orzelius: ah my bad accidentally left the debug in

Orzelius: go-kubernetes is not affected as even before the fix it used the correct reference returned by the server when inserting the objects into the inventory

https://github.com/siderolabs/go-kubernetes/pull/60/changes#diff-9cec99821cdb493b1a4f31edde5952578351e6ef8aa24bb77ade7e1480cf2fedL181-L185

shanduur (approved):

PR #13507, opened 2026-06-03
feat: add unix scheme for log sender

Hugome:
# Pull Request

## What? (description)

Adding support for UNIX scheme in log sender implementation

## Why? (reasoning)

Can be use with a local running collector (Vector or Logstash) to buffer log while the network is starting.
Can also be use to avoid exposing IP when decorating/transforming logs with a local collector (Talos -> Vector (transform/add meta) -> and then send to OTLP/...)

## Acceptance


smira:
Talos writes all logs to `/var/log/*` - this is the best way to collect them via any kind of pod running on the same node. 

This will include system logs, and pod logs, etc. It's the same process as any other "standard" Linux distribution.

We don't want to extend the way Talos sends the logs, as it's more complex/fragile than simply picking it up from the filesystem.

PR #13508, opened 2026-06-03
fix: etcd client leak in the (legacy) Upgrade API

smira:
This doesn't affect new upgrade API.

There were two leaks in the validate path, usually they are harmless, as the Upgrade triggers a reboot, so a couple of leaking clients doesn't matter, but if they API is called repeatedly (and it fails to upgrade), the leak accumulates until `etcd` container runs out of file descriptor, eventually preventing new connections to `etcd` from being established.

Also put the etcd pre-checks under the `!preserve` condition (means the API requests wipe). The wipe behavior was disabled by default for a long time, and all etcd pre-checks only make sense if the wipe is called, otherwise upgrade doesn't affect etcd membership in any way.

### DIFF
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -531,11 +531,13 @@ func (s *Server) Upgrade(ctx context.Context, in *machine.UpgradeRequest) (*mach
 		return nil, fmt.Errorf("error validating installer image %q: %w", in.GetImage(), err)
 	}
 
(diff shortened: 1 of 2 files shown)


shanduur (approved):

frezbo (approved):
I'm surprised no linters caught this, i guess we need better linters that catches these, I guess copilot could have caught this

smira:
> I'm surprised no linters caught this, i guess we need better linters that catches these, I guess copilot could have caught this

this is also super old/legacy code path not touched for ages

Unix4ever (approved):

oguzkilcan (approved):

PR #13509, opened 2026-06-03
feat: bump Go to 1.26.4

frezbo:
Bump Go to 1.26.4

See https://github.com/siderolabs/pkgs/pull/1575


pkg/machinery/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/siderolabs/talos/pkg/machinery
 
-go 1.26.2
+go 1.26.4
```
smira: a nit, but let's not bump machinery aggressively - as it gets imported into other packages (even not our packages), it would force them to upgrade (while there is not real need to do so)

smira (approved):

shanduur (approved):

PR #13510, opened 2026-06-03
feat: enable NRI by default

frezbo:
This makes deploying gpu-operator easier and aligns with default expectations.

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -26,6 +26,26 @@ etcd: v2.6.12
 Talos is built with Go 1.26.4.
 """
 
+    [notes.cri-nri]
+        title = "Containerd NRI"
+        description = """\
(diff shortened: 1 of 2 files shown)


hack/release.toml
```diff
@@ -24,6 +24,13 @@ containerd: 2.3.1
 etcd: v2.6.12
 
 Talos is built with Go 1.26.4.
+"""
+
+    [notes.cri-nri]
+        title = "Containerd NRI"
+        description = """\
+Talos now stops disabling NRI (Node Resource Interface) by default and NRI is available to use
```
smira: I think needs to be rephrased: "Talos enables NRI by default in containerd, it can be disabled with the following config patch ..."

smira (approved):

PR #13511, opened 2026-06-04
fix: pass DisableAccessTime from VolumeConfig to EPHEMERAL mount spec

abrekhov:
## Problem

The EPHEMERAL volume transformer in `GetEphemeralVolumeTransformer` reads `Secure` from `extraVolumeConfig.Mount()` but silently omits `DisableAccessTime`. This means the following VolumeConfig has no effect:

```yaml
apiVersion: v1alpha1
kind: VolumeConfig
name: EPHEMERAL
mount:
  disableAccessTime: true
```

The `disableAccessTime` field was added to the mount spec in Talos 1.13 and is correctly wired through the rest of the stack (`MountRequest`, `mount.go:624`, `SetDisableAccessTime`), but the transformer never sets it on the `MountSpec` for EPHEMERAL.

## Root cause

```go
// system_volumes.go ~line 131
WithMount(block.MountSpec{
    ...
    Secure:            extraVolumeConfig.Mount().Secure(),    // passed
    // DisableAccessTime: extraVolumeConfig.Mount().DisableAccessTime()  missing
})
```

## Fix

One line, consistent with how `Secure` is already handled.

## Context

Discovered while investigating XFS + overlayfs D-state issues on a Talos cluster running kernel 6.18. The `noatime` mount option on `/var` (EPHEMERAL) reduces XFS journal pressure by eliminating atime update transactions on every container read — one of the contributing factors to `xlog_wait` induced D-state hangs under heavy containerd workloads.

### DIFF
--- a/internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/system_volumes.go
+++ b/internal/app/machined/pkg/controllers/block/internal/volumes/volumeconfig/system_volumes.go
@@ -135,6 +135,7 @@ func GetEphemeralVolumeTransformer(inContainer bool) volumeConfigTransformer {
 						GID:                 0,
 						ProjectQuotaSupport: cfg.Machine().Features().DiskQuotaSupportEnabled(),
 						Secure:              extraVolumeConfig.Mount().Secure(),
+						DisableAccessTime:   extraVolumeConfig.Mount().DisableAccessTime(),
 					}).
 					WithLocator(labelVolumeMatch(constants.EphemeralPartitionLabel)).
 					WithFunc(func(vcs *block.VolumeConfigSpec) error {


abrekhov:
Closing — on closer inspection this is not a bug. `cfg.Volumes().ByName()` returns `VolumeConfig` whose `Mount()` interface only exposes `Secure()` by design. `DisableAccessTime` belongs to the `UserVolumeMountConfig` interface and is intentionally not part of system volume configuration. Sorry for the noise.

smira:
It might be good idea to support noatime though

abrekhov:
I can reopen then

smira:
Closing this, I will open a PR to implement noatime.

PR #13516, opened 2026-06-04
chore: refactor tests to use debug api

frezbo:
Refactor tests to use debug api.

### DIFF
--- a/internal/integration/base/debug_exec.go
+++ b/internal/integration/base/debug_exec.go
@@ -1,0 +1,134 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration_api
+
+package base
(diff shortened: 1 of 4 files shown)


internal/integration/base/debug_exec.go
```diff
@@ -0,0 +1,122 @@
...
+				return out.String(), 0, fmt.Errorf("debug container stream closed without an exit code")
+			}
+
+			return out.String(), 0, fmt.Errorf("error receiving debug container output: %w", err)
+		}
+
+		switch resp := msg.GetResp().(type) {
+		case *machineapi.DebugContainerRunResponse_StdoutData:
+			out.Write(resp.StdoutData)
+		case *machineapi.DebugContainerRunResponse_ExitCode:
```
smira, line 112: I wonder if we need to restructure this to read all messages always, but remember the last message (e.g. exit code).

As we might seem to exit before msg.Recv returns EOF, so I wonder if that might leave the stream "hanging"

smira (approved):
🆒

PR #13517, opened 2026-06-04
feat: update Flannel to v0.28.5

smira:
See https://github.com/flannel-io/flannel/releases/tag/v0.28.5

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -22,6 +22,7 @@ Linux: 6.18.34
 Kubernetes: 1.36.1
 containerd: 2.3.1
 etcd: v2.6.12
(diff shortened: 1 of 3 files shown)


frezbo (approved):
there was some generated file right?

smira:
> there was some generated file right?

that was docs I believe which is updated

frezbo:
> > there was some generated file right?
> 
> that was docs I believe which is updated

Wasn't there a manifest that was go generated?

smira:
> Wasn't there a manifest that was go generated?

This was dropped since we moved to typed manifests - no text anymore.

PR #13525, opened 2026-06-05
fix: mark more resources as sensitive

smira:
Mark more resources which might contain user-supplied as sensitive.


shanduur (approved):

pkg/machinery/resources/k8s/kubelet_spec.go
```diff
@@ -49,6 +49,7 @@ func (KubeletSpecExtension) ResourceDefinition() meta.ResourceDefinitionSpec {
 		Type:             KubeletSpecType,
 		Aliases:          []resource.Type{},
 		DefaultNamespace: NamespaceName,
+		Sensitivity:      meta.Sensitive,
```
frezbo, line 52: what;s sensitive in kubelet? 🤔

frezbo (approved):

PR #13528, opened 2026-06-05
feat: expose disk and BIOS firmware versions

rothgar:
Adds firmware-version fields to NVMe disks and surfaces the BIOS
version on the SystemInformation resource.

  block.DiskSpec.FirmwareVersion
    requires new go-blockdevice
    Transport == "nvme"; empty for other transports.

  hardware.SystemInformationSpec.BIOSVersion
    Sourced from SMBIOS (ctrl.SMBIOS.BIOSInformation.Version)

### DIFF
--- a/internal/app/machined/pkg/controllers/block/disks.go
+++ b/internal/app/machined/pkg/controllers/block/disks.go
@@ -258,6 +258,8 @@ func (ctrl *DisksController) analyzeBlockDevice(
 			d.TypedSpec().Symlinks = nil
 		}
 
+		d.TypedSpec().FirmwareVersion = props.FirmwareRevision
+
 		return nil
 	})
 }
--- a/internal/app/machined/pkg/controllers/hardware/system.go
+++ b/internal/app/machined/pkg/controllers/hardware/system.go
@@ -151,6 +151,7 @@ func (ctrl *SystemInfoController) reconcileSystemInformation(ctx context.Context
 
 	if err := safe.WriterModify(ctx, r, hardware.NewSystemInformation(hardware.SystemInformationID), func(res *hardware.SystemInformation) error {
 		hwadapter.SystemInformation(res).Update(&ctrl.SMBIOS.SystemInformation, uuidRewrite)
+		res.TypedSpec().BIOSVersion = ctrl.SMBIOS.BIOSInformation.Version
 
 		return nil
 	}); err != nil {
--- a/api/resource/definitions/block/block.proto
+++ b/api/resource/definitions/block/block.proto
@@ -94,6 +94,7 @@ message DiskSpec {
   repeated string secondary_disks = 16;
   string uuid = 17;
   repeated string symlinks = 18;
+  string firmware_version = 19;
 }
 
 // EncryptionKey is the spec for volume encryption key.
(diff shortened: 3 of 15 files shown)


internal/app/machined/pkg/controllers/block/disks.go
```diff
@@ -317,3 +320,22 @@ func parseEnv(r io.Reader) map[string]string {
...
+		return ""
+	}
+
+	ctrl := id
+
+	if i := strings.IndexByte(id, 'n'); i > 0 {
+		ctrl = id[:i]
+	}
+
+	data, err := os.ReadFile(filepath.Join("/sys/class/nvme", ctrl, "firmware_rev"))
```
smira: this shouldn't be here => it should go to `go-blockdevice` library

internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -149,8 +168,11 @@ func (ctrl *SystemInfoController) reconcileSystemInformation(ctx context.Context
 		logger.Info("using UUID rewrite", zap.String("uuid", uuidRewrite))
 	}
 
+	biosVersion := readSysfsString(cmp.Or(ctrl.BIOSVersionPath, DefaultBIOSVersionPath))
```
smira: we already have it via go-smbios, doesn't make sense to read it https://github.com/siderolabs/go-smbios/blob/063f5dc16e0a86d4dbb8290d197bdb134d9bafac/smbios/bios_information.go#L14

internal/app/machined/pkg/controllers/hardware/system.go
```diff
@@ -161,12 +183,16 @@ func (ctrl *SystemInfoController) reconcileSystemInformation(ctx context.Context
 }
 
 func (ctrl *SystemInfoController) reconcileProcessors(ctx context.Context, r controller.Runtime) error {
+	// microcode is system-wide on x86, read once and apply to every entry
+	microcode := readSysfsString(cmp.Or(ctrl.MicrocodePath, DefaultMicrocodePath))
```
smira: microcode is not system-wide, it _should_ be in sync, but is not guaranteed to be

here mixing SMBIOS source and Linux source, bad idea.

rothgar: Is there another resource we could put this info? It would be helpful to be able to query from the API when new bugs or CVEs come out. I did test this with embedded microcode and with microcode from a system extension and it read the effective microcode properly.

smira: I don't have an immediate answer, it should be designed properly. We can change the source of querying the data to be Linux kernel, implement proper watching to support hot-plug, etc.. But right now it doesn't fit in the current design from my point of view.

rothgar: I ended up adding it to a new `logicalcpus` resource with some other available information

smira: let's not mix in right now - it's possible to add CPU support, but let's do it properly in a separate PR.

internal/app/machined/pkg/controllers/block/disks.go
rothgar: I updated in go-blockdevice should I wait for a 2.0.30 release and then update go.mod with the release?

smira: it was bumped independent of this PR, all good now

smira (approved):

PR #13529, opened 2026-06-05
feat: add declarative LVM logical volume provisioning

### LINKED ISSUE #12309: support for LVM volumes

### LINKED ISSUE #12800: [FR] Managing lvm-raid/zfs-raid using talos
## Feature Request|
Hello! 
We often use local lvm/zfs raids in our cluster, and at the moment, the only option is to provision them using k8s DS — something like this:
```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: openebs-lvm-vg-reinitializer
  namespace: kube-system
  labels:
    app: openebs-lvm-vg-reinitializer
spec:
  selector:
    matchLabels:
      app: openebs-lvm-vg-reinitializer
  template:
    metadata:
      labels:
        app: openebs-lvm-vg-reinitializer
    spec:
      hostPID: true
      nodeSelector:
        nodepool-type: worker
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: storage.example.com/type
                operator: NotIn
                values:
                - sds
      containers:
      - name: lvm-vg-reinit-container
        image: openebs/lvm-driver:1.8.0
        securityContext:
          privileged: true
        volumeMounts:
        - name: device-dir
          mountPath: /dev
        command: ["/bin/bash", "-c"]
        args:
          - |
              set -o pipefail;
              echo "--- PHASE 1: CLEANUP ---"
              for vg in $(vgs --noheadings -o vg_name); do
                echo "Removing VG: $vg"
                vgremove -ff -y "$vg" || true
              done
              for pv in $(pvs --noheadings -o pv_name); do
                echo "Removing PV: $pv"
                pvremove -ff -y "$pv" || true
              done
              MIN_SIZE_BYTES=$((1000 * 1024 * 1024 * 1024))
              for DEV_PATH in $(lsblk -I 259 -bdno NAME | grep 'n1$'); do
                echo "Wiping signatures on /dev/${DEV_PATH}"
                wipefs -a "/dev/${DEV_PATH}"
              done
              echo "--- PHASE 2: NVME PROVISIONING (INDEX + 1) ---"
              for DEV in $(lsblk -I 259 -bdno NAME | grep 'n1$'); do
                  DEV_SIZE=$(lsblk -bdno SIZE "/dev/${DEV}")
                  if [ "$DEV_SIZE" -lt "$MIN_SIZE_BYTES" ]; then
                      echo "Skipping /dev/${DEV} (size: ${DEV_SIZE} bytes < ${MIN_SIZE_BYTES} bytes)"
                      continue
                  fi
                  if [[ $DEV =~ nvme([0-9]+)n1 ]]; then
                      RAW_INDEX="${BASH_REMATCH[1]}"
                      NEW_INDEX=$((RAW_INDEX + 1))
                      FORMATTED_INDEX=$(printf "%02d" "$NEW_INDEX")
                      VG_NAME="vg-nvme-${FORMATTED_INDEX}"
                      echo "Initializing /dev/${DEV} as ${VG_NAME} (HW Index: $RAW_INDEX)"
                      pvcreate -f "/dev/${DEV}"
                      vgcreate -y "${VG_NAME}" "/dev/${DEV}"
                  fi
              done
              echo "--- FINAL LVM STATE ---"
              vgs
              echo "Re-initialization complete. Sleeping indefinitely..."
              sleep infinity
      hostNetwork: true
      serviceAccountName: openebs-lvm-node-sa
      volumes:
      - name: device-dir
        hostPath:
          path: /dev
          type: Directory
```
 But as you can see, it's quite a dangerous thing in the cluster, and even though it's only been there for 5-10 minutes, neither my bosses nor the guys from the security team like it :(
### Description
It would be great if we could prepare such raids using Talos multi-doc files, for example, as is currently done with userVolumes.
It would be really cool if you could also use disc selectors.
```yaml
diskSelector:
      size: " < 1TB"
      type: "nvme"
```
This is very convenient for such operations and makes it easier to understand the configuration! 

Thank you in advance for your reply, and have a nice day!

shanduur:
An LVMLogicalVolumeConfig multi-doc kind reconciles into an
LVMLogicalVolumeSpec and then an on-disk LV via lvcreate once the parent
VG is assembled.

Layouts: linear, raid0, raid1, raid10. RAID layouts accept nillable,
type-validated `mirrors` (raid1/raid10, default 1) and `stripes`
(raid0/raid10) fields; unset stripes auto-resolve to all available PVs
(raid10: PVcount/(mirrors+1)) at reconcile, where per-type PV-count
gates skip-and-retry when the VG is too small. `type` is a typed enum.

Size converges both ways within a 1% slack:
- grow via lvextend when the target exceeds the observed size; percent
  sizing recomputes the target from the live VG size, so the LV also
  grows when the VG itself is extended
- shrink is never performed (risks data loss); the request is surfaced
  as an LVMValidationError instead

LVMValidationError is shared with the PV-spec controller; per-VG errors
key by VG name, per-LV errors by "<vg>/<lv>" (VG names cannot contain
"/", so they never collide).

Rework the LVM integration suite onto pure Talos config + COSI
resources, dropping the privileged-pod / nsenter lvcreate path. Covers
a UserVolume (type=disk) on an LV spanning three PVs, all four layouts,
grow by byte and percent size, and activation across reboot. The
default qemu e2e variant gains a fifth extra disk so raid10 (needs four
block PVs) is exercised.'

Closes https://github.com/siderolabs/talos/issues/12800
Closes https://github.com/siderolabs/talos/issues/12309


frezbo (approved):

frezbo:
let's also test enforcing tests

pkg/machinery/config/types/storage/lvm_logical_volume_config.go
```diff
@@ -0,0 +1,222 @@
...
+	//     Logical volume name.
+	//
+	//     Must be 1-63 chars: ASCII letters, digits, hyphens, underscores.
+	MetaName string `yaml:"name"`
+	//   description: |
+	//     Logical volume layout.
+	//   values:
+	//     - linear
+	//     - raid1
+	LVType string `yaml:"type"`
```
smira: make it typed (like `nethelpers`).

smira (approved):
🆒  🧑‍🍳 💾

PR #13531, opened 2026-06-06
fix: stop follow reader on renamed files

immanuwell:
## What? (description)

Handle `fsnotify.Rename` the same way as `fsnotify.Remove` in `pkg/follow`.
Added a regression test for the renamed-file case too.

## Why? (reasoning)

Talos rotates log files with `os.Rename`, so this is a real path, not just a lab repro.
Before this, a follow reader could just sit there til the context timed out after the watched path got renamed away. Kinda meh for log streaming.
After this change it returns `file was removed while watching` right away, same as the remove case.

Repro:
1. Open a file with `follow.NewReader`.
2. Read past the current EOF.
3. Rename the file.
4. Before this fix, the read waits til ctx timeout and ends as `unexpected EOF`.
5. After this fix, it returns `file was removed while watching` right away.

## Acceptance

### DIFF
--- a/pkg/follow/follow.go
+++ b/pkg/follow/follow.go
@@ -162,7 +162,7 @@ func (r *Reader) notify() {
 				case r.notifyCh <- nil:
 				default:
 				}
-			case fsnotify.Remove:
+			case fsnotify.Remove, fsnotify.Rename:
 				select {
 				case r.notifyCh <- errors.New("file was removed while watching"):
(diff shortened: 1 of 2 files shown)


smira:
I'm not sure what exactly does this fix? 

Reading logs via logs API (`talosctl logs`) doesn't touch the disk. Is this a fix for a problem which doesn't exist?

PR #13536, opened 2026-06-07
feat: allow to disable access time for EPHEMERAL partition

smira:
A small potential optimization to avoid touching atime on filesystem access.


frezbo (approved):

smira:
```
55 38 259:1 / /var rw,nosuid,nodev,noexec,noatime shared:32 - xfs /dev/nvme0n1p1 rw,seclabel,inode64,logbufs=8,logbsize=32k,prjquota
```

PR #13544, opened 2026-06-08
feat: #13272 add default probes for kube-apiserver

### LINKED ISSUE #13272: Provide default probes for controlplane static pods
## Feature Request
Currently its not possible to set probes on the control plane pods rendered by Talos, this would be useful when using a service to provide an L3 VIP to access the cluster.

### Description
I think either allowing just the probes to be specified generically allowing a whole pod patch to configure control plane nodes would be very useful.

majabojarska:
Resolves: https://github.com/siderolabs/talos/issues/13272

Adds default, non-configurable probes to kube-apiserver.

Probe periods, timeouts, failure thresholds follow what kubeadm does out of the box ([1](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/staticpod/utils.go#L256-L279), [2](https://github.com/kubernetes/kubernetes/blob/2f5e22e8bb66836f767715d99a83662250968bf5/cmd/kubeadm/app/phases/controlplane/manifests.go?plain=1#L67-L69)). The key difference is that Talos disables kube-apiserver [anonymous auth](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests), which we work around with [anonymous auth configuration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-authenticator-configuration). Kubeadm doesn't have this problem, as it never disables anonymous auth.

The `AnonymousAuthConfigurableEndpoints` feature gate is introduced as alpha with Kubernetes v1.31 (disabled by default), and beta with v1.32 (enabled by default) (see [feature gate docs](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features)). However, since as the time of writing the oldest version we support is v1.32, this feature does not handle feature compatibility concerns that would otherwise arise with <v1.32 (feature absent or disabled by default).

One caveat is that OIDC and configurable auth are mutually exclusive ([src](https://github.com/kubernetes/kubernetes/blob/9e570c4124691bd0d765e780d0225655f99a905e/pkg/kubeapiserver/options/authentication.go#L863-L866)). Therefore, if OIDC is enabled (e.g. via `extraArgs`):
- we cannot enable configurable auth, and therefore we cannot configure kube-apiserver probes - otherwise they would always get HTTP 401 responses.
- we disable anonymous auth altogether (current approach).

As tested locally, the kube-apiserver Pods are rendered with default probes:

```yaml
apiVersion: v1
kind: Pod
metadata:
# ...
  name: kube-apiserver-talos-default-controlplane-1
  namespace: kube-system
# ...
    image: registry.k8s.io/kube-apiserver:v1.36.1
    imagePullPolicy: IfNotPresent
    livenessProbe:
      failureThreshold: 8
      httpGet:
        host: localhost
        path: /livez
        port: 6443
        scheme: HTTPS
      initialDelaySeconds: 10
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 15
    name: kube-apiserver
    readinessProbe:
      failureThreshold: 3
      httpGet:
        host: localhost
        path: /readyz
        port: 6443
        scheme: HTTPS
      periodSeconds: 1
      successThreshold: 1
      timeoutSeconds: 15
# ...
    startupProbe:
      failureThreshold: 24
      httpGet:
        host: localhost
        path: /livez
        port: 6443
        scheme: HTTPS
      initialDelaySeconds: 10
      periodSeconds: 10
      successThreshold: 1
      timeoutSeconds: 15
# ...
```

### DIFF
--- a/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
+++ b/internal/app/machined/pkg/controllers/k8s/control_plane_static_pod.go
@@ -291,10 +291,8 @@ func (ctrl *ControlPlaneStaticPodController) manageAPIServer(ctx context.Context
 	}
 
 	builder := argsbuilder.Args{
-		"admission-control-config-file": {filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml")},
-		"allow-privileged":              {"true"},
-		// Do not accept anonymous requests by default. Otherwise the kube-apiserver will set the request's group to system:unauthenticated exposing endpoints like /version etc.
-		"anonymous-auth":                     {"false"},
+		"admission-control-config-file":      {filepath.Join(constants.KubernetesAPIServerConfigDir, "admission-control-config.yaml")},
+		"allow-privileged":                   {"true"},
 		"api-audiences":                      {cfg.ControlPlaneEndpoint},
 		"bind-address":                       {"0.0.0.0"},
 		"client-ca-file":                     {filepath.Join(constants.KubernetesAPIServerSecretsDir, "ca.crt")},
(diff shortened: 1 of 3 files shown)


shanduur (approved):

majabojarska:
`integration/qemu` passed on https://github.com/siderolabs/talos/pull/13544/changes/ff8adb9febbb007d3c91114978fb877fa10e5ae6

Removing label to speed up post-rebase CI.

PR #13552, opened 2026-06-09
fix: increment time epoch on wall-clock jump when time sync is disabled

wasd171:
# Pull Request

## What? (description)

This PR adds lightweight **wall-clock jump detection** to `time.SyncController` for cases where active time synchronization is disabled (most notably in `PLATFORM=container`).

When `syncDisabled` is true, the controller now runs a small periodic check (default interval: 1 minute) that compares **wall-clock elapsed time** against **monotonic elapsed time**. If the absolute difference exceeds the existing `ntp.EpochLimit` threshold (15 minutes), it increments `time.Status.Epoch` and logs the event.

Key implementation details:

- `wallClockJumpDetector` strips the monotonic clock reading via `time.Now().Round(0)` so that `Sub` compares true wall-clock deltas
- The detector is **only** instantiated when `syncDisabled == true`; it is stopped and reset immediately when sync becomes enabled
- If a jump is detected while controller inputs are temporarily unavailable, the pending jump is retained and applied once the next reconcile cycle completes
- No new machine configuration options are introduced. No certificate lifetimes or refresh intervals are changed

Added test coverage:

- `TestWallClockJumpDetector`: unit tests for the detector math (forward jump, backward jump, below-threshold, equal-to-threshold).
- `TestSyncWallClockJumpSuite`: integration tests verifying that:
  - a detected jump increments `Epoch` when sync is disabled
  - pending jumps survive temporary input unavailability
  - the detector stops cleanly when sync is re-enabled

## Why? (reasoning)

Talos internal service certificates (Talos API server, `trustd`, maintenance service) are short-lived (`24h`) and refreshed by controllers that already depend on `time.Status`:

```go
// time status isn't fetched, but the fact that it is in dependencies means
// that certs will be regenerated on time sync/jump
```

On normal platforms, an NTP syncer detects large clock discontinuities and bumps `time.Status.Epoch`, which triggers certificate regeneration. In container mode, however, Talos forcibly disables NTP:

```go
if ctrl.V1Alpha1Mode == ModeContainer {
    syncDisabled = true
}
```

Because of this, after a host sleep / VM pause (e.g. Docker Desktop on macOS), the wall clock can jump forward by hours, but `time.Status.Epoch` never changes. The certificate controllers therefore never receive a reconcile trigger, and `apid` continues to serve an expired certificate until enough **active** runtime passes for the next monotonic refresh ticker to fire. In practice this means `talosctl` and any tooling that relies on the Talos API fails immediately after resume.

Rather than increasing certificate lifetimes (which weakens security posture) or adding new configuration knobs, this PR restores the **intended semantics** of `time.Status.Epoch`: signal large wall-clock discontinuities so that downstream controllers (including certificate controllers) can react promptly. This makes the container provisioner behavior consistent with the architectural contract already present in the codebase.

## Acceptance

- [x] you ran conformance (`make conformance`) <- but I am not a member of `siderolabs` GitHub org
- [x] you ran unit-tests (`make unit-tests`) <- full suite has failed (unrelated things), my tests passed via `make unit-tests TESTPKGS=github.com/siderolabs/talos/internal/app/machined/pkg/controllers/time`

### DIFF
--- a/internal/app/machined/pkg/controllers/time/internal/clock/export_test.go
+++ b/internal/app/machined/pkg/controllers/time/internal/clock/export_test.go
@@ -1,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package clock
+
+import "time"
+
+// SetWallClock is used in test to simulate jumps.
+func (d *WallClockJumpDetector) SetWallClock(t time.Time) {
+	d.lastWallClock = wallClockOnly(t)
+}
--- a/internal/app/machined/pkg/controllers/time/internal/clock/jump.go
+++ b/internal/app/machined/pkg/controllers/time/internal/clock/jump.go
@@ -1,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package clock provides clock-related privimites, e.g. jump detection.
+package clock
+
+import (
+	"context"
+	"time"
+
+	"github.com/siderolabs/gen/channel"
+)
+
+// DefaultJumpDetectionInterval is the default interval for wall-clock jump detection.
+const DefaultJumpDetectionInterval = time.Minute
+
+// WallClockJumpDetector detects wall-clock jumps by comparing elapsed time of wall clock and monotonic clock.
+type WallClockJumpDetector struct {
+	lastWallClock      time.Time
+	lastMonotonicClock time.Time
+	interval           time.Duration
+	threshold          time.Duration
+}
+
+// NewWallClockJumpDetector creates a new WallClockJumpDetector.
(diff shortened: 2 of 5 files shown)


smira:
I see the problem, I like the idea, but I think the implementation can be simplified. I will try to work on that.

smira:
@wasd171 I updated the PR, I think it should be still functionally same

smira (approved):

internal/app/machined/pkg/controllers/time/internal/clock/export_test.go
```diff
@@ -0,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package clock
+
+import "time"
+
+// SetWallClock is used in test to simulate jumps.
```
frezbo, line 9: can't we use synctest?

frezbo: ah I see below, cool

smira: we do use it, but with synctest time advances consistently in both clocks, so we can't "jump"

frezbo (approved):

shanduur (approved):

Issue #13554, opened 2026-06-10
Vendoring broken in 1.13.4: `github.com/ryanuber/columnize: no required module provides package github.com/ryanuber/columnize`


kastl-ars:
Dear maintainers,

## Bug Report

The vendoring seems to be broken in 1.13.4:

```
INFO:obs-service-go_modules:Using go.mod found at /tmp/tmp73a9h5j0/talosctl-1.13.4/go.mod
INFO:obs-service-go_modules:go mod download
INFO:obs-service-go_modules:go mod vendor
ERROR:obs-service-go_modules:go: github.com/siderolabs/talos/cmd/talosctl/cmd/talos imports
        github.com/ryanuber/columnize: no required module provides package github.com/ryanuber/columnize; to add it:
        go get github.com/ryanuber/columnize
ERROR:obs-service-go_modules:go mod vendor failed
```

Kind Regards,
Johannes

smira:
@shanduur can you take a look please, I guess missing go mod tidy, but I don't think we can fix it till the next release.

shanduur:
The dependency is also not present in `main`.

shanduur:
Ah, I see. Our release process skips calling go mod tidy to find those issues. I'll add a CI entry to validate it.

PR #13556, opened 2026-06-10
ci: verify go.mod tidiness in generate target

shanduur:
Run `go mod tidy` on root and machinery modules during the
generate flow and emit the results. CI `make check-dirty`
then fails on any go.mod/go.sum drift.

Ref. https://github.com/siderolabs/talos/issues/13554

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -400,7 +400,29 @@ FROM scratch AS microsoft-db-keys
 COPY --link --from=microsoft-secureboot-database /DB/Certificates/MicCor*.der /db/
 COPY --link --from=microsoft-secureboot-database /DB/Certificates/microsoft*.der /db/
 
+# The base target provides a container that can be used to build all Talos
+# assets.
+
+FROM build-go AS base
+COPY ./cmd ./cmd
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -400,7 +400,29 @@ FROM scratch AS microsoft-db-keys
...
+WORKDIR /src
+
+FROM base AS go-mod-tidy
+RUN --mount=type=cache,target=/.cache,id=talos/.cache go mod tidy
+WORKDIR /src/pkg/machinery
+RUN --mount=type=cache,target=/.cache,id=talos/.cache go mod tidy
+WORKDIR /src
+
 FROM --platform=${BUILDPLATFORM} scratch AS generate
+COPY --link --from=go-mod-tidy /src/go.mod /go.mod
```
frezbo: this has the side effect that any go.mod changes will invalidate the cache right? Well I guess that's expected and required

shanduur: Yes, unfortunately it will.

majabojarska (approved):

PR #13557, opened 2026-06-10
feat: drop apply config method reboot

smira:
Reasons:

* most of the changes are applied without reboot
* the reboot check is already inconsistent - sometimes reboot is required, but it is not properly detected
* eventually all configuration documents can be applied without a reboot
* reboot coupled with apply config doesn't allow to do graceful reboot with cordon/drain etc.

An apply config followed by a proper reboot is a better flow than automatic reboot triggered without details.

### DIFF
--- a/api/machine/machine.proto
+++ b/api/machine/machine.proto
@@ -126,7 +126,7 @@ service MachineService {
 // node.
 message ApplyConfigurationRequest {
   enum Mode {
-    REBOOT = 0;
+    REBOOT = 0 [deprecated = true, (common.remove_deprecated_enum_value) = "v2.0"]; // Deprecated: use AUTO or NO_REBOOT instead
     AUTO = 1;
     NO_REBOOT = 2;
     STAGED = 3;
--- a/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
+++ b/internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
@@ -23,7 +23,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -129,8 +128,7 @@ type Server struct {
 
 	server *grpc.Server
 
-	applyConfigMu       sync.Mutex
-	applyConfigRebooted atomic.Bool
+	applyConfigMu sync.Mutex
 }
 
 func (s *Server) checkSupported(feature runtime.ModeCapability) error {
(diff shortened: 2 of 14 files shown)


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -187,8 +185,8 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
 	}
 	defer s.applyConfigMu.Unlock()
 
-	if s.applyConfigRebooted.Load() {
-		return nil, status.Error(codes.FailedPrecondition, "configuration was already applied and node is rebooting, cannot apply more configuration changes until next reboot")
+	if in.Mode == machine.ApplyConfigurationRequest_REBOOT { //nolint:staticcheck // error for removed mode
+		return nil, status.Error(codes.Unimplemented, "the REBOOT mode is not supported, please use AUTO or NO_REBOOT modes instead")
 	}
```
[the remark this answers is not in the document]
smira: yes, this is correct, but I don't see any other to handle this (?)

internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -228,37 +226,25 @@ func (s *Server) ApplyConfiguration(ctx context.Context, in *machine.ApplyConfig
...
-	}
-
 	//nolint:exhaustive
 	switch in.Mode {
 	// --mode=try
 	case machine.ApplyConfigurationRequest_TRY:
+		fallthrough
+	// --mode=auto
+	case machine.ApplyConfigurationRequest_AUTO:
+		in.Mode = machine.ApplyConfigurationRequest_NO_REBOOT
```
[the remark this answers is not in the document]
smira: nice catch!

frezbo (approved):

api/machine/machine.proto
```diff
@@ -126,7 +126,7 @@ service MachineService {
 // node.
 message ApplyConfigurationRequest {
   enum Mode {
-    REBOOT = 0;
+    REBOOT = 0 [deprecated = true, (common.remove_deprecated_enum_value) = "v2.0"]; // use AUTO or NO_REBOOT instead
```
shanduur: Nit: Maybe we can say in the comment as well:

```suggestion
    REBOOT = 0 [deprecated = true, (common.remove_deprecated_enum_value) = "v2.0"]; // Deprecated: use AUTO or NO_REBOOT instead
```

website/content/v1.14/reference/api.md
```diff
@@ -4750,7 +4750,7 @@ rpc upgrade
 
 | Name | Number | Description |
 | ---- | ------ | ----------- |
-| REBOOT | 0 |  |
+| REBOOT | 0 | use AUTO or NO_REBOOT instead |
```
shanduur:
```suggestion
| REBOOT | 0 | Deprecated: use AUTO or NO_REBOOT instead |
```

shanduur (approved):

Unix4ever (approved):

api/machine/machine.proto
smira: thanks, adjusted it this way

PR #13562, opened 2026-06-11
feat: #13243 Support multiple discovery service configs

### LINKED ISSUE #13243: Feature Request: Multiple Discovery Service Endpoint Sources
## Feature Request

Requesting `cluster.discovery.registries.service.endpoint` in the machineconfig be converted to a list with automatic failover mechanisms in place

### Description

Because we run a cross-region KubeSpan mesh and the kubernetes registry was deprecated in Talos 1.12, the external discovery service is now the only discovery source we rely on. Notably, #13237 would enable us to turn k8s registry discovery back on and achieve that discovery redundancy again.

But unless/until that is addressed, we would like to configure multiple discovery services across failure domains, mixing self-hosted + public service endpoints for the highest reliability. Right now the `cluster.discovery.registries.service.endpoint` in the machineconfig is single endpoint by design. I'm wondering if it's possible to convert that to a list (or add an additional `fallbackEndpoints` field) and add client-side round-robin or fallback mechanisms in case one of those sources goes down.

Clients would likely need to register with all sources (rather than just the active one) so a failover doesn't result in the client pulling empty/outdated data.

majabojarska:
Resolves https://github.com/siderolabs/talos/issues/13243

Multiple discovery service endpoints can now be configured via the new, multi-doc `DiscoveryServiceConfig` document(s).

- v1alpha1 `.cluster.discovery` has been deprecated.
- The v1alpha1 `.cluster.discovery` configuration tree is mutually exclusive with multi-doc DiscoveryServiceConfigs. This is validated during config apply.
- Right now the only way to enable Kubernetes discovery is via the legacy configuration. This is intentional, as Kubernetes discovery support is being phased out from Talos since K8s 1.22. This also implies that the end user is limited to using the legacy discovery *service* configuration, if they wish to enable Kubernetes discovery.
- Affiliates will be discovered many times via multiple discovery service endpoints, but COSI will deduplicate them at the resource level, as the `service/***` identifier is the Affilliate's unique identifier.

For example, removing `.cluster.discovery` and adding the following patch to the `machineconfig`:
```yaml
---
apiVersion: v1alpha1
kind: DiscoveryServiceConfig
name: discovery-1
endpoint: https://discovery.talos.dev/
---
apiVersion: v1alpha1
kind: DiscoveryServiceConfig
name: discovery-2
endpoint: https://discovery.other.com/
```

will result in on-the-fly initialization of multiple discovery service clients, which will discover the same non-local nodes twice (assuming the discovery services are routable and healthy). However, once the corresponding Affilliates are written, they will be deduplicated (assuming both discovery services identify them through the same `id`):
```sh
talosctl get --nodes 172.20.0.2 affiliate --namespace cluster-raw                                                                          19:48:09
NODE         NAMESPACE     TYPE        ID                                                    VERSION   HOSTNAME                       MACHINE TYPE   ADDRESSES
172.20.0.2   cluster-raw   Affiliate   service/4h0d3NeXtVNavEXxs8A2Ir8Zw8VCc7FcZnDWpYH81PD   1         talos-default-controlplane-2   controlplane   ["172.20.0.3"]
172.20.0.2   cluster-raw   Affiliate   service/KZhvd6398He7vZSwo5RT6UYiaxmi3D2TdWHx56hGCHE   1         talos-default-worker-2         worker         ["172.20.0.6"]
172.20.0.2   cluster-raw   Affiliate   service/elHsGl8vKHnBvBCmDjQKOWvF0XyniqPvaxCcNzLJGnR   1         talos-default-worker-1         worker         ["172.20.0.5"]
172.20.0.2   cluster-raw   Affiliate   service/s9jWsc1nKEfDNfg1SaxENSqutjn4VZ7gJnH6wPgAKPS   1         talos-default-controlplane-3   controlplane   ["172.20.0.4"]
```

---

[Discovery service documentation](https://docs.siderolabs.com/talos/v1.13/configure-your-talos-cluster/system-configuration/discovery) will be updated here https://github.com/siderolabs/docs/pull/606


pkg/machinery/config/config_schema_test.go
```diff
@@ -160,9 +160,13 @@ func newV1Alpha1Config(t *testing.T, modifications func(config *v1alpha1.Config)
 		rawModifications(data)
 	}
 
-	// deprecated field
+	// deprecated fields
 	delete(data, "persist")
 
+	if cluster, ok := data["cluster"].(map[string]any); ok {
```
majabojarska: Note to self: Is this the correct way to approach this deprecation?

smira: no, you should modify the machine config generation to use new config, and this test should pass, you don't need to modify it ever

pkg/machinery/resources/cluster/config.go
```diff
@@ -20,20 +20,20 @@ const ConfigType = resource.Type("DiscoveryConfigs.cluster.talos.dev")
...
-	ServiceEndpointInsecure   bool   `yaml:"serviceEndpointInsecure,omitempty" protobuf:"5"`
-	ServiceEncryptionKey      []byte `yaml:"serviceEncryptionKey" protobuf:"6"`
-	ServiceClusterID          string `yaml:"serviceClusterID" protobuf:"7"`
+	RegistryKubernetesEnabled bool `yaml:"registryKubernetesEnabled" protobuf:"2"`
+	RegistryServiceEnabled    bool `yaml:"registryServiceEnabled" protobuf:"3"`
+
+	ServiceEncryptionKey []byte `yaml:"serviceEncryptionKey" protobuf:"6"`
+	ServiceClusterID     string `yaml:"serviceClusterID" protobuf:"7"`
+	// Maps DiscoveryServiceConfig name, to endpoint
+	ServiceEndpoints map[string]string `yaml:"serviceEndpoints" protobuf:"8"`
```
majabojarska: note: We might need to change this into a slice. Could be slice of endpoint strings, or a struct with name/endpoint.

pkg/machinery/resources/cluster/config.go
```diff
@@ -20,20 +20,20 @@ const ConfigType = resource.Type("DiscoveryConfigs.cluster.talos.dev")
...
-// ConfigSpec describes KubeSpan configuration.
+// ConfigSpec describes Discovery configuration.
 //
 //gotagsrewrite:gen
 type ConfigSpec struct {
-	DiscoveryEnabled          bool   `yaml:"discoveryEnabled" protobuf:"1"`
-	RegistryKubernetesEnabled bool   `yaml:"registryKubernetesEnabled" protobuf:"2"`
-	RegistryServiceEnabled    bool   `yaml:"registryServiceEnabled" protobuf:"3"`
-	ServiceEndpoint           string `yaml:"serviceEndpoint" protobuf:"4"`
-	ServiceEndpointInsecure   bool   `yaml:"serviceEndpointInsecure,omitempty" protobuf:"5"`
```
majabojarska, line 34: note: Omni needs the old fields for backward compatibility, until Omni is capable of using the new multi-doc config.

pkg/machinery/resources/cluster/config.go
majabojarska: Changed into a `[]ServiceEndpoint`, now it serializes into a `repeated` field in pb.

pkg/machinery/resources/cluster/config.go
majabojarska: Added the fields back, brought back cluster config controller support, added test coverage ensuring both old and new fields are populated correctly from the DiscoveryServiceConfigs.

majabojarska:
Integration passed on [da47ba54e85c1951478caad3356b0529686f8901](https://github.com/siderolabs/talos/pull/13562/changes/da47ba54e85c1951478caad3356b0529686f8901). Removing the `qemu/integration` tag to reduce unnecessary strain on runner.

pkg/machinery/config/generate/init.go
```diff
@@ -189,13 +189,13 @@ func (in *Input) init() ([]config.Document, error) {
 		}
 	}
 
-	if in.Options.DiscoveryEnabled != nil {
-		cluster.ClusterDiscoveryConfig = &v1alpha1.ClusterDiscoveryConfig{
+	if in.Options.DiscoveryEnabled != nil && *in.Options.DiscoveryEnabled && !in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```
smira: we try to maintain one-by-one compatibility with old version, so I think we should drop part of the condition to be strictly compatible
```suggestion
	if in.Options.DiscoveryEnabled != nil && !in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```

pkg/machinery/config/generate/worker.go
```diff
@@ -123,13 +123,13 @@ func (in *Input) worker() ([]config.Document, error) {
 		}
 	}
 
-	if in.Options.DiscoveryEnabled != nil {
-		cluster.ClusterDiscoveryConfig = &v1alpha1.ClusterDiscoveryConfig{
+	if in.Options.DiscoveryEnabled != nil && *in.Options.DiscoveryEnabled && !in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```
smira: same here

```suggestion
	if in.Options.DiscoveryEnabled != nil && !in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```

pkg/machinery/config/generate/worker.go
```diff
@@ -157,6 +157,15 @@ func (in *Input) worker() ([]config.Document, error) {
 
 	documents := []config.Document{v1alpha1Config}
 
+	if in.Options.DiscoveryEnabled != nil && *in.Options.DiscoveryEnabled && in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```
smira: nit, a bit of a shortcut:

```suggestion
	if pointer.SafeDeref(in.Options.DiscoveryEnabled) && in.Options.VersionContract.DiscoveryServiceMultidocConfig() {
```

pkg/machinery/config/config/config.go
smira, line 13: it looks like we haven't dropped the legacy interface via `Cluster().` path?

pkg/machinery/config/types/cluster/discovery_service.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Check interfaces.
+var (
+	_ config.DiscoveryServiceConfig       = &DiscoveryServiceConfigV1Alpha1{}
+	_ config.NamedDocument                = &DiscoveryServiceConfigV1Alpha1{}
+	_ config.Validator                    = &DiscoveryServiceConfigV1Alpha1{}
+	_ container.V1Alpha1ConflictValidator = &DiscoveryServiceConfigV1Alpha1{}
+)
+
+// DiscoveryServiceConfigV1Alpha1 is a config document to configure a discovery service.
```
smira, line 46: I wonder if we should move this to its own group, it's not strictly networking, might be "cluster" group? 

So we could put there some cluster-level stuff later, like ID & Secret which is used in the discovery?

smira (approved):

smira:
I added some labels to mostly test the "discovery disabled" case

pkg/machinery/config/types/cluster/discovery_service.go
majabojarska: Sounds good. I placed it in networking since it's closely related to KubeSpan, but a "cluster" group makes perfect sense.

pkg/machinery/config/config/config.go
majabojarska: I've dropped the legacy discovery service nested interface. Afaik the rest needs to stay, as long as we don't have an alternative for configuring Kubernetes discovery
https://github.com/siderolabs/talos/pull/13562/changes/df23d4d92a403f11f5a38ffba23427d158858687

pkg/machinery/config/types/cluster/discovery_service.go
majabojarska: Moved to .../types/cluster

website/content/v1.14/reference/configuration/cluster/discoveryserviceconfig.md
```diff
@@ -0,0 +1,36 @@
+---
```
smira, line 1: this should be removed, and we need to update `talosctl docs` to index a new `cluster` folder as well

majabojarska: Addressed in https://github.com/siderolabs/talos/pull/13562/commits/d3d451387e7f5187ebab899e641cb58987ef69a3

PR #13571, opened 2026-06-12
feat: add imager flag to set the SecureBoot key enrollment mode

mcanevet:
## What? (description)

Add a `--secureboot-enroll-keys` flag to the `imager` command, mapping to the existing `sdBootEnrollKeys` profile field (`off` / `manual` / `if-safe` / `force`) for both the disk-image and ISO outputs. The value is written to systemd-boot's `loader.conf` as `secure-boot-enroll`.

The default is unchanged (`if-safe`); the flag is purely additive. Previously this field could only be set by hand-writing a raw imager profile YAML.

## Why? (reasoning)

`if-safe` auto-enrolls SecureBoot keys only inside a VM, so on bare-metal the keys are never enrolled unattended. Network provisioning flows with no console operator (PXE, Tinkerbell, Metal³, …) need `force`, which enrolls once while the UEFI firmware is in setup mode (sd-boot's setup-mode gate prevents it from ever overwriting an already-enrolled platform key).

Discussed in https://github.com/siderolabs/talos/discussions/12568 — @smira confirmed adding a flag (and exposing it in Image Factory later); @frezbo confirmed the imager already supports the value via the profile.

## Acceptance

- [x] you linked an issue (if applicable) — tracked via discussion #12568
- [x] you generated documentation (`make docs`) — N/A, imager flags are not part of generated docs

### DIFF
--- a/cmd/installer/cmd/imager/root.go
+++ b/cmd/installer/cmd/imager/root.go
@@ -51,6 +51,7 @@ var cmdFlags struct {
 	SecurebootIncludeWellKnownCerts bool
 	SecurebootSignerAddress         string
 	PCRSignerAddress                string
+	SecurebootEnrollKeys            string
 }
 
 // rootCmd represents the base command when called without any subcommands.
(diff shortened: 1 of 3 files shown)


smira (approved):
thank you!

PR #13577, opened 2026-06-12
feat: introduce KubeProxyConfig multi-doc

smira:
Allow to provide resources for kube-proxy, and also provide default resources (bumping the QoS class).

Allow to customize kube-proxy using the config, clarify the settings.


pkg/machinery/config/types/k8s/proxy.go
```diff
@@ -0,0 +1,234 @@
...
+	//     Please note that kube-proxy is configured with a configuration file,
+	//     so most flags have no effect.
+	//   schema:
+	//     type: object
+	//     additionalProperties:
+	//       oneOf:
+	//         - type: string
+	//         - type: array
+	//           items:
+	//             type: string
```
frezbo, line 94: should we mark args as deprecated ?

go.mod
```diff
@@ -190,6 +191,7 @@ require (
 	k8s.io/klog/v2 v2.140.0
 	kernel.org/pub/linux/libs/security/libcap/cap v1.2.78
 	sigs.k8s.io/hydrophone v0.7.0
+	sigs.k8s.io/yaml v1.6.0
```
frezbo, line 195: 😓

frezbo (approved):

pkg/machinery/config/types/k8s/proxy.go
frezbo: I was thinking we could convert to config on the fly and use it internally 🤔

smira: The way I have it right now is that legacy v1alpha1 config uses args, to keep compatibility.

New config goes into structured config, so args are mostly ignored, but there are a few args which don't exist in theconfig.

go.mod
smira: This is fine - it was actually an indirect dependency all the way, so nothing really changed in this aspect.

smira:
I will land this after https://github.com/siderolabs/talos/pull/13590 lands.

PR #13588, opened 2026-06-15
feat: implement filesystem trim support

### LINKED ISSUE #8314: fstrim and monitoring for disks

smira:
It is not enabled by default, it can be configured either on global level, or for each volume specifically.

Once enabled, an automatic stable schedule is created for each volume which is not dependent on reboots.

Fixes #8314


internal/app/machined/pkg/controllers/block/volume_trim.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	if ctrl.TrimFunc == nil {
+		ctrl.TrimFunc = fstrim.Fstrim
+	}
+
+	timer := time.NewTimer(time.Hour)
+	defer timer.Stop()
+
+	timer.Stop()
```
frezbo, line 83: is this intentional? if the time is stopped, what's it use? 🤔

smira: this timer fires on the next scheduled trim, so initially we have no schedule at all, so we stop it initially, another way is to make it `nil` and recreate it, this is another way, I think this looks more clean (?)

frezbo (approved):

PR #13590, opened 2026-06-16
fix: ensure consistent manifest apply order

smira:
Fist, pull in a fix from go-kubernetes to show apply order in the exact order manifests were applied.

Second, in the sync code, promote configmaps and secrets to the third (custom) stage, before other resources, but after other cluster-wide resources.


smira:
The go-kubernetes fix is https://github.com/siderolabs/go-kubernetes/pull/61

Orzelius (approved):

frezbo (approved):

PR #13591, opened 2026-06-16
feat: add UnattendedInstall config and controller

shanduur:
Replace the deprecated `.machine.install` _v1alpha1_ field with a
`UnattendedInstall` multi-doc config kind driven by a COSI controller.
The controller exposes install progress via `UnattendedInstallStatus`
and triggers reboot through `RebootRequest` + `RebootController` instead
of calling the sequencer directly, decoupling install from the legacy
boot sequence.

Add `UnattendedInstallCondition` to `CRI` and `Trustd` service pre-start
conditions, delaying their startup until the install completes.
This prevents `EPHEMERAL`-holding services from blocking teardown
during the subsequent reboot. The `Boot` sequence always runs
startEverything; services block on their own conditions rather than
requiring an external lifecycle controller.


frezbo:
I  think legacyBIOSSupport is something really old and we should drop it, I don't think any code uses that

pkg/machinery/config/types/runtime/unattended_install.go
```diff
@@ -0,0 +1,209 @@
...
+
+	//   description: |
+	//     Indicates if MBR partition should be marked as bootable (active).
+	//     Should be enabled only for the systems with legacy BIOS that doesn't support GPT partitioning scheme.
+	InstallLegacyBIOSSupport *bool `yaml:"legacyBIOSSupport,omitempty"`
+
+	//   description: |
+	//     Indicates if legacy GRUB bootloader should use kernel cmdline from the UKI instead of building it on the host.
+	//     This changes the way cmdline is managed with GRUB bootloader to be more consistent with UKI/systemd-boot.
+	InstallGrubUseUKICmdline *bool `yaml:"grubUseUKICmdline,omitempty"`
```
smira: I think we should lock it to `true` in the new config, we don't need to keep backwards compatibility in the new machine config document.

pkg/machinery/config/types/runtime/unattended_install.go
```diff
@@ -0,0 +1,209 @@
...
+	//     - true
+	//     - yes
+	//     - false
+	//     - no
+	Wipe *bool `yaml:"wipe,omitempty"`
+
+	//   description: |
+	//     Indicates if MBR partition should be marked as bootable (active).
+	//     Should be enabled only for the systems with legacy BIOS that doesn't support GPT partitioning scheme.
+	InstallLegacyBIOSSupport *bool `yaml:"legacyBIOSSupport,omitempty"`
```
smira: same as Noel, I think we can skip it for now, not sure if anyone still needs this, this is only for legacy/BIOS boot flow on some old hardware.... if someone needs that still (I don't think we even test it today), it could be in the API (?)

internal/app/machined/pkg/controllers/runtime/unattended_install.go
```diff
@@ -0,0 +1,351 @@
...
+func (ctrl *UnattendedInstallController) Run(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-r.EventCh():
+		}
+
+		// no on-disk install in container mode.
+		if ctrl.V1Alpha1Mode == v1alpha1runtime.ModeContainer {
```
smira: totally a nit, we could move this closer to the beginning of the function and do `return nil` <- this signals the runtime to stop running the controller

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -157,6 +157,13 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
...
 		return nil
 	default:
+		// When the UnattendedInstall multi-document config is present, install is driven by the
+		// UnattendedInstallController instead of the install sequence (which also reboots). Skip the
+		// sequence entirely so install/reboot is not performed here.
+		if r.Config() != nil && r.Config().UnattendedInstallConfig() != nil {
+			return nil
+		}
+
 		if !r.State().Machine().Installed() || r.State().Machine().IsInstallStaged() {
```
smira: I wonder if we should skip the install if there is no unattended or no machine.install at all? Just keep running (assuming something calls install API?)

internal/app/machined/pkg/system/services/cri.go
```diff
@@ -74,10 +75,16 @@ func (c *CRI) PostFunc(runtime.Runtime, events.ServiceState) (err error) {
...
 // Condition implements the Service interface.
 func (c *CRI) Condition(r runtime.Runtime) conditions.Condition {
-	return conditions.WaitForAll(
+	conds := []conditions.Condition{
 		network.NewReadyCondition(r.State().V1Alpha2().Resources(), network.AddressReady, network.HostnameReady, network.EtcFilesReady),
 		files.NewEtcFileCondition(r.State().V1Alpha2().Resources(), "machine-id", constants.CRIConfig),
-	)
+	}
+
+	if r.Config() != nil && r.Config().UnattendedInstallConfig() != nil {
```
smira: I wonder if we could skip this condition here, and populate the resource from the controller with something like "skipped" which would be a success? 

so that we don't bind the knowledge about all conditions into the services, just always wait for some form "finalized" state of the install?

pkg/machinery/config/generate/init.go
```diff
@@ -34,12 +36,6 @@ func (in *Input) init() ([]config.Document, error) {
 		MachineCA:       in.Options.SecretsBundle.Certs.OS,
 		MachineCertSANs: in.AdditionalMachineCertSANs,
 		MachineToken:    in.Options.SecretsBundle.TrustdInfo.Token,
-		MachineInstall: &v1alpha1.InstallConfig{
```
smira, line 40: nit, you can also use `nilIf` helper if you want to avoid branching a bit

website/content/v1.14/reference/configuration/v1alpha1/installconfig.md
```diff
@@ -0,0 +1,90 @@
+---
+description: |
+    InstallConfig represents the installation options for preparing a node.
```
smira: need to add `//docgen:nodoc` to the struct, otherwise docgen thinks it's a new multi-doc (but it isn't)

smira (approved):
I like this a lot, but I also think we should excercise this a bit more in the tests.

E.g. we have some tests which boot from a disk image - we could make `talosctl cluster create` skip the UnattendedInstallConfig there to test this case

internal/app/machined/pkg/system/services/cri.go
shanduur: I think that binary condition is not a good idea to have a binary indicator in the resource. This is also the current best practice guideline in Kubernetes to stop having `ready: true` indicator, and instead rely on conditions. We do not have this kind of granularity when it comes to resource statuses, but skipping binary indicators will give use a better control in the future.

pkg/machinery/config/types/runtime/unattended_install.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/config/container"
+	"github.com/siderolabs/talos/pkg/machinery/config/internal/registry"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/meta"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/siderolabs/talos/pkg/machinery/config/validation"
+)
+
+// UnattendedInstallKind is an UnattendedInstall config document kind.
+const UnattendedInstallKind = "UnattendedInstall"
```
smira:
```suggestion
const UnattendedInstallKind = "UnattendedInstallConfig"
```

.kres.yaml
```diff
@@ -93,7 +93,7 @@ spec:
     integration/provision-0: "Run provision integration tests for track 0 (previous→stable, BIOS, trustedboot upgrade)"
     integration/provision-1: "Run provision integration tests for track 1 (stable→current, preserve+stage, enforcing)"
     integration/provision-2: "Run provision integration tests for track 2 (current→current, extra-cmdline)"
-    integration/provision-3: "Run provision integration tests for track 3 (maintenance API basic, maintenance API + SideroLink)"
```
smira, line 96: we already have some tests which do not require an install (disk image), we could automatically disable the install config to verify the behavior

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer.go
```diff
@@ -157,7 +157,21 @@ func (*Sequencer) Install(r runtime.Runtime) []runtime.Phase {
 	case runtime.ModeContainer:
 		return nil
 	default:
-		if !r.State().Machine().Installed() || r.State().Machine().IsInstallStaged() {
+		// When the UnattendedInstall multi-document config is present, install is driven by the
```
smira: TODO: for the next release, drop the `Install` sequence completely and convert `.machine.install` into UnattendedInstallConfig

internal/integration/provision/maintenance_unattended.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// SuiteName ...
+func (suite *MaintenanceUnattendedSuite) SuiteName() string {
+	return fmt.Sprintf("provision.UpgradeSuite.MaintenanceUnattended-TR%d", suite.track)
+}
+
+// TestUnattendedInstall provisions a node with UnattendedInstallConfig
+// and verifies install + reboot + cluster formation.
+func (suite *MaintenanceUnattendedSuite) TestUnattendedInstall() {
```
smira: I think it matches any flow with `cluster create ... --with-apply-config` which includes tests like `e2e-iso`, and `-iso` and PXE.

in `maintenance_siderolink.go` we can patch the test to skip `UnattendedInstallConfig`, as it installuse using the API

smira (approved):

PR #13594, opened 2026-06-16
feat: extract kube-apiserver config into multi-doc configs

### LINKED ISSUE #8933: Add support for structured authentication configuration
## Feature Request

Add support for structured authentication configuration to add authentication providers to the cluster.

### Description

I'm looking into using the Structured Authentication Configuration feature in my Talos cluster. The `APIServerConfig` does not support adding this config file for `kube-apiserver`.

Feature gate

| Feature | Default | Stage | Since | Until |
| --- | --- | --- | --- | --- |
| StructuredAuthenticationConfiguration |false | Alpha | 1.29 | 1.29 |
| StructuredAuthenticationConfiguration |true | Beta | 1.30 | - |

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

It requires a file provided as `--authentication-config` to `kube-apiserver`.

> To use structured authentication configuration, you specify the path to the authentication configuration using the --authentication-config command line argument in the API server. 

https://kubernetes.io/blog/2024/04/25/structured-authentication-moves-to-beta/

### LINKED ISSUE #11324: Unable to use Anonymous Authenticator Configuration (mutually exclusive with --anonymous-auth)
## Bug Report
Unable to use Anonymous Authenticator Configuration (mutually exclusive with --anonymous-auth)

### Description
Currently, it doesn't appear possible to use [Anonymous Authenticator Configuration](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-authenticator-configuration), as this is mutually exclusive with the kube-apiserver `--anonymous-auth` flag - something we don't appear to be able to exclude from apiserver args.

We're not looking to disable anonymous auth completely in this case, but to enable anonymous auth only for select endpoints. ie.

```
        apiVersion: apiserver.config.k8s.io/v1beta1
        kind: AuthenticationConfiguration
        anonymous:
          enabled: true
          conditions:
          - path: "/livez"
          - path: "/readyz"
          - path: "/healthz"
```

I do see this [feature request](https://github.com/siderolabs/talos/issues/11315), so expect this may be resolved alongside that. We're not so keen on enabling anonymous auth globally, so this is quite important to us.

### Logs
```
E0708 03:09:05.998492       1 run.go:72] "command failed" err="anonymous: Forbidden: --anonynous-auth flag cannot be set when anonymous field is configured in authentication configuration file"
```

### Environment

```
Client:
        Tag:         v1.9.6
        SHA:         d6f6e92b
        Built:       
        Go version:  go1.23.8
        OS/Arch:     linux/amd64
Server:
        NODE:        nsw-artm-opsk8stest-ctrl1
        Tag:         v1.9.6
        SHA:         d6f6e92b
        Built:       
        Go version:  go1.23.8
        OS/Arch:     linux/amd64
        Enabled:     RBAC
```
- Kubernetes version: [`kubectl version`]
```
Client Version: v1.32.4
Kustomize Version: v5.5.0
Server Version: v1.32.4
```
- Platform: metal

### LINKED ISSUE #13328: kube-apiserver audit-log-* extraArgs silently dropped (no error, no warning)
## Bug Report

`cluster.apiServer.extraArgs` accepts `audit-log-path`, `audit-log-maxage`, `audit-log-maxbackup`, `audit-log-maxsize`, and `audit-policy-file` without error, but silently ignores them. The hardcoded `/var/log/audit/kube/kube-apiserver.log` remains in effect.

## Reproduce (v1.12.2)

```yaml
cluster:
  apiServer:
    extraArgs:
      audit-log-path: /var/lib/audit/kube-apiserver.log
```

```
$ talosctl apply-config -f patch.yaml         # succeeds
$ talosctl get staticpod kube-apiserver -o yaml | grep audit-log-path
        - --audit-log-path=/var/log/audit/kube/kube-apiserver.log
```

### Description

`control_plane_static_pod.go` `mergePolicies` has no entries for any `audit-*` key. `argsbuilder.Merge()` has no default branch, so unregistered keys fall through with no assignment and no error.

### Logs

no logs, silent drop

### Possible fix (pick one)

- Add the five `audit-*` keys to `mergePolicies` as `MergeOverwrite`.
OR
- Add a default branch to `argsbuilder.Merge()` returning an error on unregistered keys.

## Env
Talos v1.12.2 · kernel 6.18.5-talos · k8s v1.35.0

### Environment

- Talos version: Talos v12.2 (kernel 6.18.5-talos)
- Kubernetes version: v1.35.0
- Platform: SBC rockchip

### LINKED ISSUE #13452: refactor kube-apiserver config
Audit flags used, expose more configuration available as config files.

Do multi-doc for each apiserver config document available. Get rid of deprecated flags.

smira:
Fixes #13452

Fixes #8933 

Fixes #11324 

Fixes #13328


hack/test/patches/usernamespace.yaml
```diff
@@ -13,3 +9,8 @@ apiVersion: v1alpha1
 kind: SysctlConfig
 params:
   user.max_user_namespaces: "11255"
+---
+apiVersion: v1alpha1
+kind: KubeAPIServerConfig
+extraArgs:
+  feature-gates: UserNamespacesSupport=true
```
frezbo: i think this is beta now, so we can drop

internal/app/machined/pkg/controllers/k8s/internal/k8stemplates/apiserver.go
```diff
@@ -77,3 +84,190 @@ func APIServerEncryptionConfig(rootK8sSecrets *secrets.KubernetesRootSpec) (runt
...
+					Resources:      resources,
+					SecurityContext: &v1.SecurityContext{
+						AllowPrivilegeEscalation: new(false),
+						ReadOnlyRootFilesystem:   new(true),
+						Capabilities: &v1.Capabilities{
+							Drop: []v1.Capability{"ALL"},
+							// kube-apiserver binary has cap_net_bind_service=+ep set.
+							// It does not matter if ports < 1024 are configured, the setcap flag causes a capability dependency.
+							// https://github.com/kubernetes/kubernetes/blob/5b92e46b2238b4d84358451013e634361084ff7d/build/server-image/kube-apiserver/Dockerfile#L26
+							Add: []v1.Capability{"NET_BIND_SERVICE"},
```
frezbo, line 231: is this still the case?

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/internal/k8stemplates/apiserver.go
smira: looks like still the same https://github.com/kubernetes/kubernetes/blob/6bf6ebe1a0f92f0256d700dd99b57df3362e677a/build/server-image/kube-apiserver/Dockerfile#L26

hack/test/patches/usernamespace.yaml
smira: it's even GA in 1.36, dropping it

PR #13595, opened 2026-06-16
chore: update `pkgs` and `tools` (2026-06-16)

majabojarska:
Updates `pkgs` and `tools` to latest at the time of writing.


smira (approved):
hack/release.toml needs to be updated as well... 

but also it'd be nice (can be another/separate PR to try run 1.5.0-rc.latest), 1.5.0 should land in time for 1.14.0

hack/release.toml
```diff
@@ -23,6 +23,12 @@ Kubernetes: 1.36.1
 containerd: 2.3.1
 etcd: v3.7.0-rc.0
 Flannel: v0.28.5
+runc: 1.5.0-rc.3
+ZFS: 2.4.3
```
frezbo: just runc is enough in talos

frezbo (approved):

hack/release.toml
smira: we usually don't promote extension versions into Talos release notes.

ZFS -> SOCI.

smira (approved):

PR #13598, opened 2026-06-17
refactor: simplify trustd/apid rootfs setup

smira:
These services run from an empty rootfs, and instead of bind-mounting `/sbin/init` (multi-service binary) into it outside of container startup flow, simply use a container mount to bind the binary into the initially empty rootfs.

### DIFF
--- a/internal/app/machined/pkg/system/services/utils.go
+++ b/internal/app/machined/pkg/system/services/utils.go
@@ -12,7 +12,6 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/siderolabs/talos/internal/pkg/containermode"
-	mount "github.com/siderolabs/talos/internal/pkg/mount/v3"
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/services/utils.go
```diff
@@ -24,16 +23,6 @@ func prepareRootfs(id string) error {
 		return fmt.Errorf("failed to create rootfs %q: %w", rootfsPath, err)
 	}
 
-	executablePath := filepath.Join(rootfsPath, id)
```
frezbo, line 27: this might change some selinux contexts, so worth running the enforcing test

smira: good idea!

smira: it passed

frezbo (approved):

PR #13607, opened 2026-06-18
fix: bump number of open files for etcd

smira:
The default OCI spec contains a limit of 1024.

See https://github.com/kubernetes/kubernetes/issues/111622


frezbo (approved):

Orzelius (approved):

majabojarska:
`oci.WithRlimit` added to the containerd API in v2.3.0, while we're on v2.2.x in 1.12 with no way to upgrade to v2.3.x without upgrading Golang 1.26.x. Rejecting from 1.12 backports.

PR #13610, opened 2026-06-18
chore: bump deps (go, k8s, docker)

majabojarska:
- Partially based on renovate: https://github.com/siderolabs/talos/pull/10045/changes/3c845a0b55e85f9b211857fa4dc0539638babe9f
- For Go deps, I've bumped only the direct dependencies, and then `go get` and `go mod tidy` for each dir containing `go.mod`.
- There are some changes around YAML marshalling/unmarshalling, which reflect the difference in behavior between `go.yaml.in/yaml/v4` versions rc.4 and rc.5.


shanduur (approved):

pkg/machinery/config/encoder/encoder_test.go
```diff
@@ -545,7 +546,12 @@ func (suite *EncoderSuite) TestConcurrent() {
 
 func decodeToMap(data []byte) (map[any]any, error) {
 	raw := map[any]any{}
+
 	err := yaml.Unmarshal(data, &raw)
+	if err != nil && strings.Contains(err.Error(), "no documents in stream") {
```
frezbo, line 551: https://github.com/yaml/go-yaml/issues/342 this was fixed upstream, maybe we wait for rc6 or not update now

PR #13615, opened 2026-06-19
feat: isolate cri, kubelet and pods in a sandbox namespace

frezbo:
Run CRI services in a shared PID+mount namespace anchored
by sandboxd, so workloads can no longer reach machined's PID-1.

sandboxd as PID 1 of the namespace serves launch requests from machined
over a control socket, forking cri and kubelet directly into the namespace
and passing back each service's pidfd. If it dies the kernel tears down
the namespace and it is recreated, without rebooting the node.

sandboxd runs in its own least-privilege SELinux domain (sandboxd_t),
entered via the exec-label mechanism, instead of resharing init_t.


internal/pkg/selinux/policy/selinux/services/sandboxd.cil
```diff
@@ -0,0 +1,52 @@
...
+(allow sandboxd_t procfs_t (dir (mounton)))
+
+; stdout/stderr are machined's console (kernel_t fd referencing the rootfs_t
+; /dev/console char device); sandboxd logs startup errors there.
+(allow sandboxd_t kernel_t (fd (use)))
+(allow sandboxd_t rootfs_t (chr_file (read write getattr ioctl append open)))
+
+; Control + per-service SEQPACKET sockets are created by machined (init_t); the
+; init_t pidfd/stdio fds are also handed across. sandboxd talks over them.
+(allow sandboxd_t init_t (unix_stream_socket (read write getattr getopt setopt shutdown)))
```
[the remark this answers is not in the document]
frezbo: This is correct as written. In SELinux, AF_UNIX `SOCK_SEQPACKET` and `SOCK_STREAM` both map to the `unix_stream_socket` object class — only `SOCK_DGRAM` maps to `unix_dgram_socket`. So `unix_stream_socket` is the right class for the SEQPACKET control/per-service sockets.

This was also confirmed empirically: the rule was written from the actual audit denials observed when running `sandboxd_t` permissive on a node, which reported `tclass=unix_stream_socket` for the SEQPACKET sockets, e.g.:

```
avc: denied { read write } ... scontext=...sandboxd_t... tcontext=...init_t... tclass=unix_stream_socket
```

With this rule in place, sandboxd↔machined works under enforcing with no socket denials.

rothgar:
Does this affect calling the API from workloads? https://docs.siderolabs.com/kubernetes-guides/advanced-guides/talos-api-access-from-k8s

smira:
> Does this affect calling the API from workloads? [docs.siderolabs.com/kubernetes-guides/advanced-guides/talos-api-access-from-k8s](https://docs.siderolabs.com/kubernetes-guides/advanced-guides/talos-api-access-from-k8s)

no, it shouldn't - it's all the way API/network-based.

Edit: we verify it in the integration tests as well.

frezbo:
> Does this affect calling the API from workloads? https://docs.siderolabs.com/kubernetes-guides/advanced-guides/talos-api-access-from-k8s

No, this is more isolating processes and it's fd's

website/content/v1.14/reference/configuration/runtime/workloadisolationconfig.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+
+
+
+
+{{< highlight yaml >}}
+apiVersion: v1alpha1
+kind: WorkloadIsolationConfig
```
smira: This is more of a UX question: should we instead create a `SecurityProfileConfig` document (more generic one), where we'd have more toggles:

* workloadIsolation: true/false (with explicit true in the v1.14+ config)
* lockOptCniBin: true (bad name, but the idea is to drop overlay on /opt/cni/bin, should work with Flannel out of the box I believe)
* something else (?)

smira: (I don't mean we should lock /opt/cni/bin for this PR, just thinking about what we can add there)

frezbo: yup this makes more sense

frezbo: if we're leaning towards a `SecurityProfileConfig` we could even has similar like k8s, baseline, restricted etc

smira: I think toggles for each option work nice for now vs. levels? At least until we don't have enough toggles available

website/content/v1.14/reference/configuration/runtime/workloadisolationconfig.md
shanduur, line 1: This is a leftover, right?

shanduur (approved):

website/content/v1.14/reference/configuration/runtime/workloadisolationconfig.md
frezbo: yup, missed that

PR #13625, opened 2026-06-20
fix: allow CNI SELinux runtime operations

howdude230:
## What

Allow SELinux policy operations needed by CNI agents and pod container runtime helpers on Talos 1.13:

- BPF filesystem/file/dir writes from pod domains.
- CNI state writes from pod domains.
- pod containerd to pod unix stream socket `connectto`.
- pod relabel operations on pod containerd runtime files.
- bounded CNI/initramfs/udev startup operations observed while CNI plugins and module helpers run.
- bounded read-only access for host diagnostics surfaces that expose process/rootfs/udev labels.

The generated embedded `policy.33` is regenerated from the updated CIL sources.

## Why

With SELinux enabled in permissive mode on Talos v1.13.4, Cilium/Multus-style networking can emit AVCs for legitimate CNI/BPF/runtime operations. Several of the needed rules already existed as commented Cilium/CNI policy hints; this makes those paths explicit and adds the missing bounded startup/read cases.

## Validation

- `git diff --check`
- `docker build --target selinux-generate --build-arg TOOLS_PREFIX=ghcr.io/siderolabs/tools --build-arg TOOLS=v1.13.0-4-ga06bb31 -o /tmp/talos-policy .`
- Verified `internal/pkg/selinux/policy/policy.33` matches the regenerated policy output.


internal/pkg/selinux/policy/selinux/services/machined.cil
```diff
@@ -94,8 +94,13 @@
 (allow kernel_t initramfs_t (fd (use)))
 ; Make machined go into proper context
 (allow initramfs_t bin_exec_t (fs_classes (ro))) ; Find init
+(allow initramfs_t bin_exec_t (file (execute execute_no_trans)))
 (allow initramfs_t init_exec_t (fs_classes (ro)))
 (allow initramfs_t init_exec_t (file (execute)))
+(allow initramfs_t cni_plugin_t (fs_classes (ro)))
+(allow initramfs_t cni_plugin_t (file (execute execute_no_trans)))
+(allow initramfs_t modprobe_exec_t (fs_classes (ro)))
+(allow initramfs_t modprobe_exec_t (file (execute execute_no_trans)))
```
dsseng, line 103: Could you please attach the AVC logs of these events (without this patch applied)?

Does the test build contain any extensions (or merely built via Factory)? If so, please rebuild using only the tooling in Talos repo.

I am currently tracking a bug which causes many denials to `initramfs_t`, while no process actually runs with this label. This bug is somehow bound to overlayfs created when working with system extensions, including a metadata one embedded by Image Factory, but has not yet been tracked down to the exact cause

internal/pkg/selinux/policy/selinux/services/udev.cil
```diff
@@ -112,3 +112,4 @@
 (allow udev_t bin_exec_t (file (execute execute_no_trans)))
 (allow udev_t kernel_t (system (ipc_info)))
 (allow udev_t any_p (sem (all)))
+(allow udev_t initramfs_t (fd (use)))
```
dsseng, line 115: Again, might stem from the same bug, so please attach a log. We should not work around this bug, as it will create a lot of wrong rules in the policy

internal/pkg/selinux/policy/selinux/services/cri.cil
```diff
@@ -211,6 +212,16 @@
 ; `install` also seems to test the plugin
 (allow pod_t cni_plugin_t (file (execute execute_no_trans)))
 
+; CNI agents manage pinned BPF programs/maps owned by containerd processes.
+(allow pod_p pod_containerd_t (bpf (prog_run)))
+(allow pod_p sys_containerd_t (bpf (prog_run)))
+
+; Host diagnostics mounted into pods can expose proc entries labelled with the
+; owning process domain instead of procfs_t.
+(allow pod_p init_t (file (getattr open read)))
```
dsseng, line 221: Curious what these are. But given it is effectively RO access to machined procfs, should be relatively harmless

Also wondering how is this going to coexist with #13615 - perhaps we will have to make Cilium not try to contact init anyway (or it will read from a stub one)

internal/pkg/selinux/policy/selinux/common/processes.cil
```diff
@@ -128,13 +128,14 @@
 
 ; ; cilium-agent
 ; (allow any_p self (perf_event (all)))
-; (allow any_p bpf_t (fs_classes (rw)))
+(allow any_p bpf_t (fs_classes (rw)))
```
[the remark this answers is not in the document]
dsseng: This makes sense. Maybe we should even consider creating a separate context for bpf-capable pods and assign it to cilium-agent via its daemonset.

internal/pkg/selinux/policy/selinux/common/processes.cil
```diff
@@ -128,13 +128,14 @@
...
 
 ; fio
 (allow any_p self (shm (all)))
 
 ; /run/flannel, created by containerd
 (allow any_p pod_containerd_run_t (fs_classes (rw)))
+(allow pod_p pod_containerd_run_t (fs_classes (relabelfrom relabelto)))
```
dsseng, line 138: What is the cause for this? But generally should be okay

dsseng (commented):
Thanks for your PR, could you please attach more logs for some insights on the new rules?

internal/pkg/selinux/policy/selinux/services/cri.cil
frezbo: I saw this too just run `ps aux` from a kubectl debug pod with enforcing=1

dsseng: Yes, but I believe most workloads should be okay not examining host processes. And even if SELinux now allows that, namespace isolation would likely prevent them from seeing what they need, so I'm curious why do they need that

frezbo: I was just pointing out something i observed, it's intended behavior and I don't see why we should allow it now

internal/pkg/selinux/policy/selinux/services/machined.cil
dsseng: Could you please rebase and re-test? #14055 should've fixed this

PR #13632, opened 2026-06-22
fix: race in etcd member add

smira:
This seems to be relatively new (might be etcd 3.7 new behavior), but I noticed some of the clusters in the CI failing to build the full etcd cluster with:

```
[   38.183306] [talos] retrying error: error adding member: etcdserver: Peer URLs already exists
```

So looks like adding a member succeeds at some point, while Talos e.g. walks aways on a timeout.

Try to workaround this and assume that the new member is us.

### DIFF
--- a/internal/app/machined/pkg/system/services/etcd.go
+++ b/internal/app/machined/pkg/system/services/etcd.go
@@ -27,6 +27,7 @@ import (
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/siderolabs/gen/xslices"
 	"github.com/siderolabs/go-retry/retry"
+	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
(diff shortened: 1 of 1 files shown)


smira:
I don't have a reliable way to reproduce this though...

internal/app/machined/pkg/system/services/etcd.go
```diff
@@ -306,6 +308,21 @@ func addMember(ctx context.Context, r runtime.Runtime, addrs []string, name stri
 
 	add, err := client.MemberAddAsLearner(ctx, addrs)
 	if err != nil {
+		if errors.Is(err, rpctypes.ErrPeerURLExist) {
```
frezbo, line 311: should we file an issue upstream too? or discuss this first

smira: I don't think it's a bug in etcd... I will circle back to this once I fix other stuff

frezbo (approved):

PR #13643, opened 2026-06-23
feat(talosctl): wait for volume detach before reboot on upgrade

eleboucher:
# Pull Request

## What? (description)

talosctl upgrade and reboot drain the node before rebooting, but the drain only waits for pods to be evicted, not for their CSI volumes to be unmounted and detached.

This adds a post-drain wait for the node's VolumeAttachments to clear before reboot, via two new flags on both commands:

  - --wait-for-volume-detach (default true, requires --drain)
  - --volume-detach-timeout (default 5m)

Best-effort: a timeout is reported but does not block the reboot. Implemented as nodedrain.WaitForVolumeDetach, called per-node from drainNodes after CordonAndDrain.

## Why? (reasoning)
Volume teardown (kubelet unmount + attach-detach controller detach) continues after a pod is evicted. A kexec reboot can outrun it, orphaning the mount and pinning the volume in the node's volumesInUse. That blocks detach, so the volume can't  reattach elsewhere and the rescheduled pod hits a Multi-Attach error until the mount is manually cleared. Waiting for VolumeAttachments to clear before reboot closes the race.

## Acceptance

### DIFF
--- a/cmd/talosctl/cmd/talos/drain.go
+++ b/cmd/talosctl/cmd/talos/drain.go
@@ -30,7 +30,10 @@ type nodeUpdate struct {
 // and performs cordon + drain on all of them in parallel.
 //
 // It returns a map of talosIP -> k8sNodeName for use in the uncordon phase.
-func drainNodes(ctx context.Context, clientFactory *global.ClientFactory, drainTimeout time.Duration, rep *reporter.Reporter) (map[string]string, error) {
+//
+// When waitForVolumeDetach is set, each node also waits (up to volumeDetachTimeout)
+// for its CSI VolumeAttachments to clear after draining; best-effort, never fatal.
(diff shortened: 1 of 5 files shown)


smira:
I feel we can't put everything in talosctl, we can handle some common cases, but in the end every Kubernetes cluster might require its own set of actions to do a pre-reboot action. 

Talos API is modular now, and upgrade and reboot APIs are decoupled, so one can do any kind of drain as desired. 

E.g. `kubectl <something> && talosctl upgrade && kubectl <something else>` is fine. Or any other custom flow built directly on top of the APIs

But we can't really get into a job of supporting any custom drain flow.

smira:
@eleboucher are you interested in more modular `talosctl upgrade` CLI?

Today one could do the following:

1. Perform the upgrade while the machine is running. Nothing changes until the reboot which will actually apply the upgrade.
2. Drain/wait/... (whatever custom procedure is).
3. Reboot.
4. Uncordon/... (whatever custom procedure is).

Talos API offers several primitive operations:

* pull an installer image (you can choose whether to use in-memory containerd, or the one back by /var)
* run the installer (perform an upgrade), once again choosing which one to use
* reboot (with options).

Everything else is just packaging things together in the CLI

PR #13650, opened 2026-06-25
feat: native BGP support via embedded GoBGP

### LINKED ISSUE #11126: Native GoBGP Integration for BGP Configuration via MachineConfig
## Feature Request
 Native GoBGP Integration for BGP Configuration via MachineConfig

### Description
Hi Talos team 👋

I’ve been thinking — instead of relying on the FRR system extension, what if Talos integrated [GoBGP](https://github.com/osrg/gobgp) directly into its core? This would allow native BGP configuration through machineconfig, which would be fantastically convenient for modern L3-based infrastructure setups.

#### Use Case

I’m currently building my own cloud platform, where L3 routing is extended directly down to the Talos host. Each hypervisor node is:
	•	Connected to two ToR/Leaf switches,
	•	Establishing BGP unnumbered sessions (using IPv6 LLA),
	•	Advertising its /32 node IP via a loopback interface,
	•	Receiving BGP routes from components like kube-ovn, MetalLB, or Cilium,
	•	And redistributing those routes back into the fabric.

This setup also makes the Kubernetes control-plane IP highly available by design, and enables proper ECMP-based load balancing — something that traditional VIP (L2/ARP) approaches can’t offer.

For context, Nokia [has already implemented](https://www.youtube.com/watch?v=7prUnxglfCk) this internally using the [FRR system extension](https://github.com/abckey/frr-talos-extension), but having native GoBGP integration in Talos would make the setup much cleaner and more maintainable.

Image

### Request

Would the Talos team be open to implementing native GoBGP support for this?
Or would you prefer that I contribute a PR myself? It might take me a bit more time to implement, but the logic seems relatively straightforward.

Either way — thank you for building such an awesome OS! 🚀

#### Related issues
https://github.com/siderolabs/talos/issues/4334
https://github.com/siderolabs/extensions/issues/625

frezbo:
Add native BGP to Talos, configured through machine config, so L3/CLOS fabrics no longer need the FRR system extension (siderolabs/talos#11126). A new BGPConfig document starts an embedded GoBGP speaker that originates the node's loopback addresses as host routes and installs the routes it learns from its neighbors into the kernel FIB — Talos itself fills the role FRR delegates to Zebra, reusing the existing RouteSpec controller.

The speaker supports both numbered and unnumbered (IPv6 link-local) peering with eBGP "external" neighbors, advertises IPv4 prefixes over IPv6 link-local next-hops (RFC 8950), installs ECMP routes across multiple next-hops, and runs BFD for sub-second failure detection. Learned routes can carry a preferred source address (the FRR `set src` equivalent) so traffic is sourced from the node's loopback even though unnumbered uplinks have no address of their own. Session state is exposed as a BGPPeerStatus resource.

Fixes: #11126


pkg/provision/providers/qemu/node.go
```diff
@@ -428,3 +442,36 @@ func getRandomMacAddress() string {
...
+			CNIConfList: fabricCNIConfList(bridge, mtu),
+		})
+	}
+
+	return uplinks
+}
+
+// fabricCNIConfList returns a per-uplink CNI conflist: an L2-only bridge (no IPAM, no gateway, no
+// masquerade) plus tc-redirect-tap, so the uplink comes up IPv6-link-local only for unnumbered BGP.
+func fabricCNIConfList(bridge string, mtu int) string {
```
dsseng, line 470: Cool, I might want to experiment with something like peer-to-peer links between nodes, without the host connected to them. I think current BGP status should be enough to autoconfigure such fast paths for inter-node traffic

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
dsseng, line 1: Great, may be worth a separate commit?

cmd/talosctl/cmd/mgmt/bgp_launch.go
```diff
@@ -0,0 +1,333 @@
...
+
+	_, err = srv.AddPath(apiutil.AddPathRequest{Paths: []*apiutil.Path{{Family: family, Nlri: nlri, Attrs: attrs}}})
+
+	return err
+}
+
+// fabricSendRAs periodically emits IPv6 Router Advertisements on the named interface (router-lifetime 0:
+// presence / link-local discovery only, not a default gateway) so unnumbered peers learn the host's
+// link-local. The interface is resolved in-loop and retried, so it tolerates the interface appearing
+// after the peer starts (the per-uplink CNI bridges are created when the nodes attach).
```
dsseng, line 282: I planned on adding the same thing, and it should perhaps be not only specific to BGP, any IPv6 network should advertise its presence via RA for the clients to enable IPv6 and extras such as DHCPv6

dsseng: But it's abstract enough for me to extend the code after this is merged and I go on with those IPv6 changes

internal/app/machined/pkg/controllers/network/bgp.go
```diff
@@ -0,0 +1,891 @@
...
+	"github.com/siderolabs/talos/internal/app/machined/pkg/controllers/internal/trigger"
+	"github.com/siderolabs/talos/internal/app/machined/pkg/controllers/network/watch"
+	talosconfig "github.com/siderolabs/talos/pkg/machinery/config/config"
+	"github.com/siderolabs/talos/pkg/machinery/nethelpers"
+	configresource "github.com/siderolabs/talos/pkg/machinery/resources/config"
+	"github.com/siderolabs/talos/pkg/machinery/resources/network"
+)
+
+// bgpListenPort is the standard BGP TCP port.
+const bgpListenPort = 179
```
dsseng: cmd/talosctl/cmd/mgmt/bgp_launch.go also contains this, and we should perhaps just move the BGP standard port to constants for clarity

internal/app/machined/pkg/controllers/network/route_status.go
```diff
@@ -119,6 +124,24 @@ func (ctrl *RouteStatusController) Run(ctx context.Context, r controller.Runtime
...
+				status.NextHops = nil
+
+				for _, nh := range route.Attributes.Multipath {
+					nhGateway, _ := netip.AddrFromSlice(nh.Gateway)
+
+					if nh.Via != nil {
+						nhGateway, _ = netip.AddrFromSlice(nh.Via.Addr)
+					}
+
+					weight := uint32(nh.Hop.Hops) + 1 // the kernel encodes the next-hop weight as (weight - 1)
```
dsseng: It's not really clear to me, why does the weight depend on (number of?) hops? I guess in the default BGP setup we should keep all weights at 1 (or otherwise equal) unless we have some specific input to prefer a certain path

Will read the code in the IDE later and maybe realize what is `nh.Hop.Hops` responsible for

dsseng (commented):
🆒 going to play with this and Cilium BGP soon-ish, great feature

frezbo:
> 🆒 going to play with this and Cilium BGP soon-ish, great feature

yeh, next I want to add in the veth pair support and add tests with cilium and stuff

cmd/talosctl/cmd/mgmt/bgp_launch.go
frezbo: yeh, i'll keep as is for now, later you can extract

cmd/talosctl/cmd/mgmt/cluster/create/clusterops/configmaker/internal/makers/common.go
frezbo: the patches one?

internal/app/machined/pkg/controllers/network/bgp.go
```diff
@@ -0,0 +1,889 @@
...
+func (ctrl *BGPController) reconcile(ctx context.Context, r controller.Runtime, logger *zap.Logger) error {
+	cfg, err := safe.ReaderGetByID[*configresource.MachineConfig](ctx, r, configresource.ActiveID)
+	if err != nil && !state.IsNotFoundError(err) {
+		return fmt.Errorf("error getting machine config: %w", err)
+	}
+
+	var bgpConfig talosconfig.NetworkBGPPeerConfig
+
+	if cfg != nil {
+		bgpConfigs := cfg.Config().NetworkBGPPeerConfigs()
```
smira: I think we have an interface problem here - the document is not named, so there should be only a single instance of it; the method should return nil if the document is not present

internal/app/machined/pkg/controllers/network/bgp.go
```diff
@@ -0,0 +1,889 @@
...
+
+	return netip.Addr{}
+}
+
+// buildPeer translates a config neighbor into a gobgp peer.
+// resolveNeighborPeer resolves a neighbor's BGP address. Numbered peers use the configured address;
+// unnumbered peers resolve their single link-local neighbor from the kernel neighbor table (populated
+// via Router Advertisements) and use a zoned address (fe80::x%iface). Returns false if an unnumbered
+// peer is not yet discovered.
+func resolveNeighborPeer(neighbor talosconfig.NetworkBGPNeighbor, logger *zap.Logger) (resolvedPeer, bool) {
```
smira: I feel like we should move these small helpers functions into a new `internal/bgp` sub-package and add some unit-tests for them to prevent regressions

internal/app/machined/pkg/controllers/network/export_test.go
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package network
+
+// Test exports for unexported BGP/route helpers (consumed by the external network_test package).
+var (
+	RouteSpecForTest       = routeSpec
```
smira: see above, let's better encapsulate into a small sub-package

internal/app/machined/pkg/controllers/network/router_advertisement.go
```diff
@@ -0,0 +1,210 @@
...
+// Name implements controller.Controller interface.
+func (ctrl *RouterAdvertisementController) Name() string {
+	return "network.RouterAdvertisementController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *RouterAdvertisementController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: configresource.NamespaceName,
```
smira: I feel we have two consumers of the config here, and to prevent unnecessary weke-ups we should insert an intermediate resource and controller:

* `MachineConfig` -> `BPGPeerConfig` resource, probably enhance with data we care about
* consume `BGPPeerConfig` directly

pkg/machinery/config/config/config.go
```diff
@@ -34,6 +34,7 @@ type Config interface { //nolint:interfacebloat
 	NetworkProbeConfigs() []NetworkCommonProbeConfig
 	NetworkBlackholeRouteConfigs() []NetworkBlackholeRouteConfig
 	NetworkRoutingRuleConfigs() []NetworkRoutingRuleConfig
+	NetworkBGPPeerConfigs() []NetworkBGPPeerConfig
```
smira: I guess it should be this

```suggestion
	NetworkBGPPeerConfigs() NetworkBGPPeerConfig
```

website/content/v1.14/reference/configuration/network/bgppeerconfig.md
```diff
@@ -0,0 +1,104 @@
...
+apiVersion: v1alpha1
+kind: BGPPeerConfig
+localASN: 65001 # Local autonomous system number for the BGP speaker.
+# Names of the interfaces whose addresses are originated into BGP as host routes (/32, /128).
+advertise:
+    - dummy0
+multipath: true # Enable ECMP (multipath) for routes learned from multiple neighbors.
+# BGP neighbors to peer with.
+neighbors:
+    - interface: enp1s0 # Interface name for an unnumbered (IPv6 link-local) session. Mutually exclusive with `address`.
```
smira: we use the name `link` exclusively everywhere in the machine config now

also not clear if we support aliases

pkg/machinery/config/config/config.go
dsseng: Hm, I guess we should support multiple peers with different configs, so might be worth keeping it a slice and work in that direction at some point

frezbo: is there a real use case? with BFD and a CLOS i think it's only required for a single one, at least if we need multiple, we need to have new listener for each and more complex, at least i'll keep a single one for now only

dsseng: maybe for complex multihoming setups it could be useful, but probably it's a task for later

internal/app/machined/pkg/controllers/network/router_advertisement.go
```diff
@@ -0,0 +1,210 @@
...
+		go ctrl.runSender(senderCtx, iface, logger)
+	}
+
+	// enable accept_ra=2 on each unnumbered interface so we learn the neighbor's link-local.
+	r.StartTrackingOutputs()
+
+	for _, iface := range sortedKeys(interfaces) {
+		id := kernel.Sysctl + "." + fmt.Sprintf("net/ipv6/conf/%s/accept_ra", iface)
+
+		if err = safe.WriterModify(ctx, r, runtimeres.NewKernelParamSpec(runtimeres.NamespaceName, id), func(spec *runtimeres.KernelParamSpec) error {
```
smira: I think should we create KernelParamDefaultSpec to make it user-overridable, as KernelParamSpecs today are written by the config? (probably it's not the first place we do it wrong)

frezbo: if that's the case it will break BGP, so we're left with a broken state right? I mean if the user overrode this?

smira: yep, but if the user _wants_ it, what can we do?

in fact, today with the same scenario (user sets sysctl manually), this controller will simplify fail to write the resource (as it's owned by another controller) and go into forever crash loop

frezbo: yeh, i see, fixed

smira (approved):
I think this is very cool, looking forward to see how it unwraps 😎

dsseng (approved):
great!

PR #13657, opened 2026-06-25
fix: `cli.md` codeblocks

majabojarska:
https://github.com/siderolabs/talos/pull/13073 introduced a regression to codeblock fencing (\`\`\`), breaking downstream static site generation in siderolabs/docs. This commit fixes the generation logic to fence code snippets.

### DIFF
--- a/website/content/v1.14/reference/cli.md
+++ b/website/content/v1.14/reference/cli.md
@@ -714,17 +714,23 @@ If it is not installed already, you can install it via your OS's package manager
 
 To load completions in your current shell session:
 
-	source <(talosctl completion bash)
+```
+source <(talosctl completion bash)
+```
 
 To load completions for every new session, execute once:
 
 #### Linux:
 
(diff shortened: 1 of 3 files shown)


website/content/v1.14/reference/cli.md
```diff
@@ -3535,20 +3555,24 @@ Generated bundle contains the following debug information:
 
 - For each node:
 
-	- Kernel logs.
```
smira, line 3538: this is valid Markdown though, it's just Mintlify being exceptionally dumb

smira (approved):

website/content/v1.14/reference/cli.md
majabojarska: ~Yep~

majabojarska: I think it was fine with the list, it's the `>`, `<(` patterns it crashes on, which is fair, since those should be fenced. I'll improve the logic to not fence this list.

PR #13658, opened 2026-06-25
fix: building integration test binary on darwin

frezbo:
Fix building integration test binary on darwin.

### DIFF
--- a/internal/app/machined/pkg/controllers/cluster/config.go
+++ b/internal/app/machined/pkg/controllers/cluster/config.go
@@ -7,8 +7,6 @@ package cluster
 import (
 	"context"
 	"encoding/base64"
-	"net"
-	"net/url"
 
 	"github.com/cosi-project/runtime/pkg/controller"
 	"github.com/cosi-project/runtime/pkg/controller/generic/transform"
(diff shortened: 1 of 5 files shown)


smira:
Hm... Do we have something in machinery which doesn't build on darwin?

frezbo:
> Hm... Do we have something in machinery which doesn't build on darwin?

yeh, the xfs stuff uses a lot of linux syscalls

frezbo:
> > Hm... Do we have something in machinery which doesn't build on darwin?
> 
> yeh, the xfs stuff uses a lot of linux syscalls

I moved things around so they are not in machinery anymore

internal/pkg/discovery/discovery.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"net"
+	"net/url"
+
+	clustertypes "github.com/siderolabs/talos/pkg/machinery/config/types/cluster"
+)
+
+// NormalizeEndpoint normalizes a discovery service endpoint URL into a host:port address plus insecure flag.
+func NormalizeEndpoint(rawEndpoint string) (addr string, insecure bool, err error) {
```
smira: I think we should instead move this into pkg/machinery/resources/cluster - this would be much cleaner instead of creating a one-off package here

smira: or even into `machinery/config/types/cluster` where the validate method lives

frezbo: fixed

smira:
Let's do a rekor bump as a separate PR please for backporting

smira (approved):

PR #13662, opened 2026-06-26
feat: allow user managed etc files

frezbo:
Allow user managed etc files using `EtcFileConfig` machine config document.


pkg/machinery/config/types/runtime/etc_file.go
```diff
@@ -0,0 +1,187 @@
...
+	"apparmor.d/",
+	"ca-certificates/",
+	"lvm/",
+	"pki/",
+	"selinux/",
+	"ssl/",
+}
+
+func validateEtcFilePath(path string) error {
+	if slices.Contains(managedEtcFiles, path) {
```
smira, line 181: I think we should also deny things like "../", and any "."

Also not sure if we should allow directories at all? Would we create them if they are missing?

But in general, I think this should be a "clean" path at least

internal/app/machined/pkg/controllers/files/etc_file_config.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+
+		return false, fmt.Errorf("error getting spec %q: %w", id, err)
+	}
+
+	if spec.Metadata().Phase() == resource.PhaseRunning {
+		return true, nil
+	}
+
+	managedBy, _ := spec.Metadata().Annotations().Get(files.EtcFileManagedByAnnotation)
```
smira: I'm not sure what is exactly going on here... multiple controllers write to the same resource? this is already protected by the owner in the metadata

not sure why we try to remove a resource?

smira: there is a pattern e.g. in this controller https://github.com/siderolabs/talos/blob/fd8a6f77071bb957a0362bfcb4f408b3d19549ab/internal/app/machined/pkg/controllers/block/mount_request.go#L63-L67

Which does exactly this - reconcile config with resources, and handle teardown

The only caveat is that we should ignore outputs not owned by us

pkg/machinery/resources/files/files.go
```diff
@@ -12,3 +12,12 @@ const NamespaceName resource.Namespace = "files"
 
 // SourceFileAnnotation is used to annotate a file resource with the source file path(s).
 const SourceFileAnnotation = "talos.dev/source-file"
+
+// EtcFileManagedByAnnotation marks whether an EtcFileSpec is managed by Talos or by user configuration.
+const EtcFileManagedByAnnotation = "talos.dev/managed-by"
```
smira: can we simply use Owner which is set by the controller? if we name the controller something like "UserEtcFileController" it should be pretty obvious?

smira (commented):
I think we need an integration test

pkg/machinery/resources/files/files.go
frezbo: ah I see

internal/app/machined/pkg/controllers/files/etc_file_config.go
frezbo: i just wanted to see if an existing file in /etc was overwritten by a user provided one from the etcfile yaml

pkg/machinery/config/types/runtime/etc_file.go
frezbo: > I think we should also deny things like "../", and any "."

I'll add more tests also

> Also not sure if we should allow directories at all? Would we create them if they are missing?

yeh the etcfile controller does mkdir internally for filepath.Base, so we could deny them

frezbo: anyways it'll never escape since we use xfs.Root

smira: it might be still nice to avoid confusion (just in case)

frezbo: fixed

internal/app/machined/pkg/controllers/files/etc_file_config.go
frezbo: fixed

pkg/machinery/resources/files/files.go
frezbo: fixed

smira (approved):

PR #13667, opened 2026-06-28
feat: udevd controller and udev rules config document

frezbo:
This moves udevd into it's own controller and implements a new `UdevRulesConfig` machine document.

In the process also harden SELinux permissions, introduce a new `udev_hwdb_t` for the `hwdb.bin` file.

Now only `hwdb.bin` is writable, and `/usr/lib/udev` is no longer an overlay mount.

`99-talos.rules` is synced by controller in an anonymous tmpfs and the file is `move_mounted`'ed as `ro` in `/usr/lib/udev/rules.d`.

This should also increase the security posture.


smira:
@frezbo should have a size please

smira (approved):

PR #13676, opened 2026-06-29
feat: extract `ClusterId` and `ClusterSecret` to multi-doc `DiscoveryIdentityConfig`

### LINKED ISSUE #13654: Move `.ClusterID` and `.ClusterSecret` to multi-doc `DiscoveryIdentityConfig`
https://github.com/siderolabs/talos/blob/960875b66bfc97a6e4c0eb1d09bd58ef5406b203/pkg/machinery/config/types/v1alpha1/v1alpha1_types.go#L310-L317

majabojarska:
Resolves https://github.com/siderolabs/talos/issues/13654

Extracts v1alpha1 `ClusterConfig.ClusterID` and `ClusterConfig.ClusterSecret` into a multi-doc `DiscoveryIdentityConfig`.


majabojarska:
After integration was done, I've only updated docs/comments. Removed the labels not strain the runners unnecessarily.

hack/release.toml
```diff
@@ -116,6 +117,7 @@ List of changes:
 - Deprecated `.cluster.network` in the v1alpha1 config; use the `KubeNetworkConfig` document for Kubernetes network configuration; Flannel can be configured using the `KubeFlannelCNIConfig` document.
 - Deprecated `.cluster.coreDNS` in the v1alpha1 config; use the `KubeCoreDNSConfig` document for CoreDNS configuration.
 - Deprecated `.cluster.discovery` in the v1alpha1 config; use the `DiscoveryServiceConfig` document for discovery service configuration. The v1alpha1 config and `DiscoveryServiceConfig` are mutually exclusive.
+- Deprecated `.cluster.secret` and `cluster.id` in the v1alpha1 config; use the `DiscoveryIdentityConfig` document for discovery identity configuration. The v1alpha1 config and `DiscoveryIdentityConfig` are mutually exclusive.
```
smira: nit: we can fix up later, but probably `.cluster.discovery` related changes deserve their own block in release notes (they are not strictly coupled with Kubernetes).

internal/app/machined/pkg/controllers/cluster/config.go
```diff
@@ -62,12 +63,17 @@ func NewConfigController() *ConfigController {
 						})
 					}
 
-					res.TypedSpec().ServiceEncryptionKey, err = base64.StdEncoding.DecodeString(cfg.Cluster().Secret())
+					identity := cfg.DiscoveryIdentityConfig()
```
smira, line 66: we can actually move this pre-check to `MapMetadataOptionalFunc` - it is now redundant, as it checks for `Config().Cluster() != nil`, but looks like nothing consumes this anymore, we could instead check `DiscoveryIdentityConfig() != nil`

internal/app/machined/pkg/controllers/cluster/config.go
smira, line 51: from previous change, but I think this should be moved out of `if` - if there is no config for discovery service, we should wipe it

internal/app/machined/pkg/controllers/cluster/info.go
```diff
@@ -36,7 +36,10 @@ func NewInfoController() *InfoController {
 				return optional.Some(cluster.NewInfo())
 			},
 			TransformFunc: func(ctx context.Context, r controller.Reader, logger *zap.Logger, cfg *config.MachineConfig, info *cluster.Info) error {
-				info.TypedSpec().ClusterID = cfg.Config().Cluster().ID()
+				if identity := cfg.Config().DiscoveryIdentityConfig(); identity != nil {
+					info.TypedSpec().ClusterID = identity.ClusterID()
```
smira, line 40: need an `else` clause to clear the value if the config is missing

internal/app/machined/pkg/controllers/kubespan/config.go
```diff
@@ -56,8 +56,11 @@ func NewConfigController() *ConfigController {
 						}
 					}
 
-					res.TypedSpec().ClusterID = c.Cluster().ID()
-					res.TypedSpec().SharedSecret = c.Cluster().Secret()
+					if identity := c.DiscoveryIdentityConfig(); identity != nil {
```
smira: same here, probably move to `MapMetadataOptionalFunc` - KubeSpan can't function without identity, so identity missing => no KubeSpan

pkg/machinery/config/config/cluster.go
smira: this and `Secret()` should be dropped

pkg/machinery/config/types/cluster/discovery_identity.go
```diff
@@ -0,0 +1,135 @@
...
+			MetaAPIVersion: "v1alpha1",
+		},
+		MetaClusterID:     clusterID,
+		MetaClusterSecret: clusterSecret,
+	}
+}
+
+func exampleDiscoveryIdentityConfigV1Alpha1() *DiscoveryIdentityConfigV1Alpha1 {
+	return NewDiscoveryIdentityConfigV1Alpha1(
+		"luc6cM3kvc0+Y4FlScZ3EHcaadnSdjJ9ECsvhV5mFLA=",
```
smira: nit - let's use invalid values here, so that nobody copy-pastes that into real config

pkg/machinery/config/types/cluster/discovery_identity.go
```diff
@@ -0,0 +1,135 @@
...
+	return s.MetaClusterSecret
+}
+
+// Validate implements config.Validator interface.
+func (s *DiscoveryIdentityConfigV1Alpha1) Validate(validation.RuntimeMode, ...validation.Option) ([]string, error) {
+	if s.MetaClusterID == "" {
+		return nil, errors.New("clusterID is required")
+	}
+
+	if s.MetaClusterSecret == "" {
```
smira, line 132: I think we should check that it base64-decodes, and the expected length (?). At least we should do this for the secret - as it should be a valid AES encryption key

pkg/machinery/config/config/cluster.go
majabojarska: Done

hack/release.toml
majabojarska: Moved to a new section.

pkg/machinery/config/types/cluster/discovery_identity.go
majabojarska: Done, used values non-parsable from base64 into 32 bytes:
```go
	return NewDiscoveryIdentityConfigV1Alpha1(
		"cluster-id-base64-encoded-32-bytes",
		"cluster-secret-base64-encoded-32-bytes",
	)
```

pkg/machinery/config/types/cluster/discovery_identity.go
majabojarska: Added validation with appropriate unit tests.

internal/app/machined/pkg/controllers/cluster/config.go
majabojarska: What about Kubernetes discovery? The K8s discovery controller ([push](https://github.com/siderolabs/talos/blob/0d6199fa3279d32b0eca0d93f30029d20146dbc1/internal/app/machined/pkg/controllers/cluster/kubernetes_push.go#L1), [pull](https://github.com/siderolabs/talos/blob/0d6199fa3279d32b0eca0d93f30029d20146dbc1/internal/app/machined/pkg/controllers/cluster/kubernetes_pull.go#L1)) doesn't need the discovery identity nor discovery service configuration, but if we wipe `cluster.Config` on `DiscoveryIdentityConfig` being `nil`, then we prevent `RegistryKubernetesEnabled` from being set to `true` [here](https://github.com/siderolabs/talos/blob/0d6199fa3279d32b0eca0d93f30029d20146dbc1/internal/app/machined/pkg/controllers/cluster/config.go#L98).

Is my understanding correct?

smira: I think technically it was the case before as well (?)

smira: I mean even it was unnecessary dependency, it was still locked to it.

majabojarska: As discussed on call: 
- We can't check for discovery service config or discovery identity config being empty/nil in `MapMetadataOptionalFunc`, as it would prevent the `cluster.Config` output from being created (output) altogether, which would prevent K8s discovery from running when discovery service/identity config were empty/nil, even K8s discovery was enabled through the machine config, as we would not produce and `cluster.Config` to consume.
- Decoupling (as in splitting) K8s discovery configuration from Discovery Service configuration in the cluster.Config would resolve the above issue, which I'll look into in a subsequent PR.

majabojarska:
To be squashed before merging. Keeping diffs separate to make the second round of review easier.

pkg/machinery/config/types/cluster/discovery_identity.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/siderolabs/talos/pkg/machinery/config/types/meta"
+	"github.com/siderolabs/talos/pkg/machinery/config/types/v1alpha1"
+	"github.com/siderolabs/talos/pkg/machinery/config/validation"
+)
+
+// DiscoveryIdentityKind is a discovery identity config document kind.
+const (
+	DiscoveryIdentityKind = "DiscoveryIdentityConfig"
+	ClusterIDLenBytes     = 32
+	ClusterSecretLenBytes = 32
```
majabojarska: note: replace with `constants.DefaultClusterIDSize`, `constants.DefaultClusterSecretSize`

pkg/machinery/config/types/cluster/discovery_identity.go
```diff
@@ -0,0 +1,164 @@
...
+		return errors.New("cluster identity is already configured in .cluster.id/.cluster.secret of the v1alpha1 config")
+	}
+
+	return nil
+}
+
+// ValidateBase64WithLen validates that the given string is a valid base64 encoded
+// string and that it decodes to the expected length in bytes.
+func ValidateBase64WithLen(base64Str string, wantLenBytes int) error {
+	decoded, err := base64.StdEncoding.DecodeString(base64Str)
```
majabojarska: Tests are failing because clusterID uses URLEncoding, while secret uses StdEncoding.
https://github.com/siderolabs/talos/blob/39e4e774fe919bf21a50c8493e9c6712ef29177b/pkg/machinery/config/generate/secrets/bundle.go#L390-L404

Decoding in validation needs to reflect the used encoding.

smira (approved):
🆒

PR #13679, opened 2026-06-30
fix: flaky tests

frezbo:
For serviceaccountsuite use latest talosctl
For file atomicity use Rename.

### DIFF
--- a/internal/integration/api/serviceaccount.go
+++ b/internal/integration/api/serviceaccount.go
@@ -275,7 +275,7 @@ func (suite *ServiceAccountSuite) createTestJob(ns, name, serviceAccount, node s
 						"containers": []map[string]any{
 							{
 								"name":  "talosctl",
-								"image": "ghcr.io/siderolabs/talosctl:v1.12.4", // sync with cmd/talos/image.go list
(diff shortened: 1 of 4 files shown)


internal/integration/api/serviceaccount.go
```diff
@@ -271,7 +271,7 @@ func (suite *ServiceAccountSuite) creteTestJob(ns, name, serviceAccount, node st
 						"containers": []map[string]any{
 							{
 								"name":  "talosctl",
-								"image": "ghcr.io/siderolabs/talosctl:v1.12.4", // sync with cmd/talos/image.go list
```
smira, line 278: in the airgapped test we rely on this - so if we bump, we need to keep it in sync (e.g. use 1.13.5)

internal/integration/api/serviceaccount.go
```diff
@@ -271,7 +271,7 @@ func (suite *ServiceAccountSuite) creteTestJob(ns, name, serviceAccount, node st
...
+								"image": "ghcr.io/siderolabs/talosctl:v1.13.5", // sync with cmd/talos/image.go list
```
smira: I don't see the image list being updated for it?

smira: it'd be great to verify with airgapped pipeline

smira (approved):

PR #13693, opened 2026-07-02
chore: speed up lefthook with globs

majabojarska:
Currently we run all jobs every single time, which for Talos is fairly time consuming. It's best to skip the jobs that do not relate to the currently staged changes.

Here the change is straightforward, as we maintain the `Makefile` and `lefthook.yml` by hand. I'll try to reflect this back into Kres `CompileLefthook` callables.

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -6,21 +6,33 @@ pre-commit:
         jobs:
           - name: generate
             run: make generate
+            glob:
+              - "pkg/**"
+              - "api/**"
+              - "hack/boilerplate.txt"
             env:
               USERNAME: siderolabs
             stage_fixed: true
           - name: docs
             run: make docs
+            glob:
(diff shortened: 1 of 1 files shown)


smira (approved):

lefthook.yml
```diff
@@ -6,21 +6,33 @@ pre-commit:
...
+              - "cmd/talosctl/**"
+              - "pkg/machinery/config/**"
             env:
               USERNAME: siderolabs
             stage_fixed: true
           - name: fmt
             run: make fmt
+            glob:
+              - "api/**/*.proto"
+              - "pkg/provision/api/**/*.proto"
```
majabojarska, line 28: Note: This is correct, as the only things `make fmt` does now is format protobuf.

shanduur (approved):

PR #13694, opened 2026-07-02
feat: move `.machine.kernel` into multi-doc

### LINKED ISSUE #12441: Move `.machine.kernel` (module config) into multi-doc
We only allow kernel module loading there, so we can decompose into a specific document.

See https://docs.siderolabs.com/talos/v1.12/reference/configuration/v1alpha1/config#kernel

```yaml
version: v1alpha1
kind: KernelModuleConfig
name: btrfs # module name
parameters: # optional
  - param1
  ...
```

majabojarska:
Resolves https://github.com/siderolabs/talos/issues/12441

- Adds a new `KernelModuleConfig` multi-doc document.
- Deprecates v1alpha1 `.machine.kernel`


pkg/machinery/config/container/container.go
```diff
@@ -276,6 +276,29 @@ func (container *Container) SysfsConfig() map[string]string {
...
+// with the multi-doc documents taking precedence over the legacy config on module name conflicts
+// (enforced by the KernelModuleConfigController, which writes one resource per module name).
+func (container *Container) KernelModuleConfigs() []config.KernelModuleConfig {
+	var modules []config.KernelModuleConfig
+
+	// v1alpha1 has the lowest priority
+	if container.v1alpha1Config != nil &&
+		container.v1alpha1Config.MachineConfig != nil &&
+		container.v1alpha1Config.MachineConfig.MachineKernel != nil { //nolint:staticcheck // legacy configuration
+		for _, module := range container.v1alpha1Config.MachineConfig.MachineKernel.KernelModules { //nolint:staticcheck // legacy configuration
```
majabojarska: note: simplify this deref if possible

hack/release.toml
```diff
@@ -285,6 +285,7 @@ List of changes:
 
 - Deprecated `.machine.sysctls` in the v1alpha1 config; use the `SysctlConfig` document for kernel sysctl configuration.
 - Deprecated `.machine.sysfs` in the v1alpha1 config; use the `SysfsConfig` document for sysfs configuration.
+- Deprecated `.machine.kernel` in the v1alpha1 config; use the `KernelModuleConfig` document for kernel module configuration.
```
majabojarska, line 288: note to self once integration passes: Mention this is the multi-doc machinery `KernelModuleConfig`, not the v1alpha1 `KernelModuleConfig`. The naming can be confusing.

shanduur (approved):

hack/release.toml
smira: users don't see `KernelModuleConfig` in the v1alpha1 old doc (this is Talos internal name only)

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -240,6 +240,10 @@ type MachineConfig struct {
 	//     Configures the kernel.
 	//   examples:
 	//     - value: machineKernelExample()
+	//
```
smira: you can drop the lines 239-242 as well - they are no longer in use

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -2602,18 +2606,38 @@ type LoggingDestination struct {
...
 // KernelConfig struct configures Talos Linux kernel.
+//
+// docgen:nodoc
+//
+// Deprecated: Use multi-doc `KernelModuleConfig` instead.
 type KernelConfig struct {
 	// description: |
 	//   Kernel modules to load.
+	//
+	// docgen:nodoc
```
smira: I think it's not required to do nodoc for every field, just for the struct (as struct becomes orphaned, it needs nodoc to exlude it from machine config docs)

smira (approved):
also: 

* [this](https://github.com/siderolabs/talos/blob/f783f6636b33750af25e37bd5d6c79fc2698acc9/hack/test/patches/extensions.yaml#L1-L33) can be refactored now
* and [this](https://github.com/siderolabs/talos/blob/f783f6636b33750af25e37bd5d6c79fc2698acc9/hack/test/tfvars/nvidia.yaml#L1-L8)

There might be more, but these are the ones I remember

The one in TF vars might be tricky - TF provider uses old machinery, so it might need to wait, but the extensions one should be gtg

frezbo (approved):

majabojarska:
Needs https://github.com/siderolabs/talos-vex/pull/82

PR #13700, opened 2026-07-02
feat: implement service account configuration

smira:
Implement full configuration for Kubernetes service accounts, supporting the full set of issuing/accepted configurations.


.kres.yaml
```diff
@@ -104,6 +104,7 @@ spec:
     integration/qemu-encrypted-vip: "Run QEMU integration tests with disk encryption and virtual IP"
     integration/qemu-enforcing: "Run QEMU integration tests on enforcing variant"
     integration/qemu-race: "Run QEMU integration tests with the race detector"
+    integration/qemu-13: "Run QEMU integration tests with Talos version contract v1.13"
```
smira, line 107: so the idea is to generate pre-1.14 configs (no multi-doc Kubernetes), and run all our tests with "old" config to verify we have no regressions

frezbo (approved):

PR #13701, opened 2026-07-02
fix: flaky serviceaccount suite test

frezbo:
We were not clearing connection refused for the controlplane nodes which might have been due to side effect of other tests, in our case this was due to the rotateca tests. Now explicitly clear the error for both workers and controlplane nodes.

### DIFF
--- a/internal/integration/api/serviceaccount.go
+++ b/internal/integration/api/serviceaccount.go
@@ -64,7 +64,7 @@ func (suite *ServiceAccountSuite) SetupTest() {
 	// make sure API calls have timeout
 	suite.ctx, suite.ctxCancel = context.WithTimeout(context.Background(), 5*time.Minute)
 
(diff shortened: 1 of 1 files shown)


internal/integration/api/serviceaccount.go
```diff
@@ -117,13 +121,9 @@ func (suite *ServiceAccountSuite) TestValid() {
 	_, err = suite.creteTestJob("kube-system", name, name, node)
 	suite.Assert().NoError(err)
 
-	err = suite.waitForJobReady(2*time.Minute, "kube-system", name)
-	suite.Assert().NoError(err)
+	defer suite.DeleteResource(suite.ctx, jobGVR, "kube-system", name) //nolint:errcheck
```
smira: ensure resource is deleted seems to be gone?

smira (approved):

internal/integration/api/serviceaccount.go
frezbo: is that really needed? i'll add back just in case

Issue #13708, opened 2026-07-05
akamai: dual-homed (public + VPC) nodes install a competing default route via the internet-less VPC gateway


macole16:
## Summary

On the `akamai` platform, a Talos node attached to **both** a public interface and a **VPC** gets a default route (`0.0.0.0/0`) installed on **both** NICs at the same metric (1024, table `main`). Because they share the same `(table, dst=default, priority)` key, which one "wins" is a race. When the **VPC gateway wins** — and the VPC gateway has **no internet egress** — the node silently loses outbound connectivity and becomes unreachable (dead apid + ICMP), while still showing "running" at the hypervisor. It's intermittent and boot-timing-dependent.

## Root cause

The `akamai` platform driver (`internal/app/machined/pkg/runtime/v1alpha1/platform/akamai/akamai.go`, v1.13.5) sets up **addresses only**:

- it emits **no IPv4 routes** (the only route it produces is an IPv6 link-local one), and
- it hardcodes `LinkName: "eth0"` on every address spec.

On current images (Talos v1.13.5, kernel 6.18.36) Linode NICs enumerate as `ens3`/`ens4`, so the platform address layer doesn't even match a link and is effectively inert. All real IPv4 addressing + routing therefore falls to Talos's **default per-link `dhcp4` operator**, which runs on every physical NIC that lacks an explicit `machine.network.interfaces` entry. On a dual-homed node that means a `dhcp4` operator on **both** the public and the VPC NIC — and Linode's VPC DHCP serves a router option, so Talos installs a competing default route via the internet-less VPC gateway.

The driver also does **not** read the newer **"Linode Interfaces"** model's per-interface `default_route` field from metadata (`akametadata.NetworkData` is consumed for addresses/ranges only), so that field can't be used to suppress the VPC default either (confirmed with a live canary: a node on `interface_generation=linode` with the VPC interface's `default_route.ipv4=false` still installed the VPC gateway as its default).

## Evidence (Talos v1.13.5, `akamai`, dual-homed node)

```
$ talosctl get operatorspecs
NAMESPACE   TYPE           ID
network     OperatorSpec   dhcp4/ens3     # public NIC
network     OperatorSpec   dhcp4/ens4     # VPC NIC  <-- Talos runs a dhcp4 operator here too

$ talosctl get routespecs inet4/10.20.0.1//1024 -o yaml
    gateway:     10.20.0.1     # VPC gateway (no internet egress)
    outLinkName: ens4
    priority:    1024
[...]

Issue #13578, opened 2026-06-12
controller-runtime(akamai): metadata parsing fails on nodes using vpc interfaces


demosdemon:
### Description

When launching a talos instance on Linode inside a new vpcs causes an error loop.

### Logs

```
[talos] restarting platform network config {"component": "controller-runtime", "controller": "network.PlatformConfigController", "interval": "3.707644929s", "error": "parse metadata: ParseAddr(\"invalid Prefix::1\"): each colon-separated field must have at least one digit (at \"invalid Prefix::1\")"}
```

https://github.com/siderolabs/talos/blob/3fc981c570205e3a71df32ad1a607c9730beaf1c/internal/app/machined/pkg/runtime/v1alpha1/platform/akamai/akamai.go#L119-L124

### Environment

- Talos version: 1.13.4
- Kubernetes version: 1.36.1
- Platform: Akamai Linode w/ new generation interfaces

Linode VPCs do not support ipv6. The search for the v6 gateway causes this to fail.

**Example**:

```
interfaces.id: ###
interfaces.purpose: vpc
interfaces.vpc.id: ###
interfaces.vpc.label: talos-devl
interfaces.vpc.subnet.id: ###
interfaces.vpc.subnet.ipv4.instance_address: 10.239.111.253/32
interfaces.vpc.subnet.ipv4.nat_1_1: ###.###.###.###
interfaces.vpc.subnet.ipv4.subnet_range: 10.239.96.0/20
interfaces.vpc.subnet.label: talos-devl
ipv4.public: ###.###.###.###
ipv4.vpc: 10.239.111.253/32
```

smira:
This is a contributed code, we don't have tests on Akamai - grab the metadata, add it to the test case for Akamai, and ensure Talos translates that into proper network config, and create a PR with a fix.

If you don't know how to adjust the code, creating a PR with a new set of data from Akamai might help us to fix this.

demosdemon:
A bigger issue is actually https://github.com/linode/go-metadata/issues/146. Could hack around it but DHCP still works.

PR #13728, opened 2026-07-07
feat: add ignoreRoutes to DHCPv4Config document

macole16:
## What

Add an `ignoreRoutes` option to the new-style `DHCPv4Config` machine-config document. When set, the DHCPv4 operator uses the lease for addressing but installs none of the routes it carries: both the Router-option default gateway and any RFC 3442 classless static routes. The connected route for the leased address is still configured from the address itself, so the interface remains usable.

This mirrors the existing `ignoreHostname` field end to end (config document, operator spec, and `dhcpparse.ParseDHCP4Ack`). The behavior matches systemd-networkd's `UseRoutes=false`.

```yaml
apiVersion: v1alpha1
kind: DHCPv4Config
name: enp0s2
ignoreRoutes: true
```

## Why

Relates to #13708 and #13578.

On a dual-homed Linode node (a public NIC plus a VPC NIC), the VPC's DHCP server always serves a default gateway, even when that VPC has no internet egress. If Talos runs a DHCP4 operator on the VPC interface, it installs a competing default route that can win the race and black-hole the node's traffic. Today the only way to avoid this is to declare the interface statically with `dhcp: false`. `ignoreRoutes` lets the interface keep using DHCP for its address while declining the routes, which matters for dynamically provisioned nodes where a static per-node address is awkward.

I wrote up the investigation behind this in a comment on #13708: https://github.com/siderolabs/talos/issues/13708#issuecomment-4898228510 (it also may cover the independent dual-homed report in #13578).

## Scope

Kept deliberately small:

- New-style `DHCPv4Config` document only. I would be glad to add the same option to the v1alpha1 `dhcpOptions` and/or a DHCPv6 counterpart if that is preferred.
- No platform-driver changes.

## Testing

- `dhcpparse` unit tests for both suppression paths (the Router-option default gateway and RFC 3442 classless static routes); the existing tests continue to assert routes are produced when routes are not ignored.
- Config document marshal round-trip and JSON schema.
- Controller mapping test asserting `ignoreRoutes` maps to the operator spec's `SkipRoutes`.

## Live validation on hardware

Beyond the unit tests, I built a `v1.14.0-alpha.2` image with this change and validated it end to end on a real dual-homed Linode node — one public NIC (`eth0`) and one VPC NIC (`eth1`) whose DHCP server serves a default gateway with no internet egress. Same node, three configurations (`talosctl get routespecs` / `operatorspecs`):

**1. Address-only baseline** — VPC NIC static (`dhcp: false`), today's workaround. A single default route, via the public gateway:

```
RouteSpec     inet4/<public-gw>//100
OperatorSpec  dhcp4/eth0
```

**2. DHCP on the VPC NIC, no `ignoreRoutes`** — reproduces the problem. The VPC's DHCP-served default route is installed alongside the public one, so there are two default routes and the internet-less VPC gateway can win:

```
RouteSpec     inet4/<vpc-gw>//1024      # competing default route
RouteSpec     inet4/<public-gw>//100
OperatorSpec  dhcp4/eth0
OperatorSpec  dhcp4/eth1
```

**3. DHCP on the VPC NIC, `ignoreRoutes: true`** — the fix. The DHCP4 operator still runs on `eth1` and the interface still gets its lease address, but the DHCP-carried default route is not installed, so we are back to a single default route:

```
RouteSpec     inet4/<public-gw>//100    # vpc-gw route no longer installed
OperatorSpec  dhcp4/eth0
OperatorSpec  dhcp4/eth1                # still running
AddressStatus eth1/<vpc-addr>/24        # lease still acquired
```

The only difference between (2) and (3) is `ignoreRoutes: true`: the lease and the operator are unchanged, only the route install is suppressed.

### DIFF
--- a/api/resource/definitions/network/network.proto
+++ b/api/resource/definitions/network/network.proto
@@ -138,6 +138,7 @@ message DHCP4OperatorSpec {
   uint32 route_metric = 1;
   bool skip_hostname_request = 2;
   ClientIdentifierSpec client_identifier = 3;
+  bool skip_routes = 4;
 }
 
 // DHCP6OperatorSpec describes DHCP6 operator options.
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -131,9 +131,11 @@ List of changes:
 """
 
     [notes.dhcpsearchdomains]
-        title = "DHCP Search Domains"
+        title = "DHCP"
         description = """\
 DHCPv4 search domains are now applied to the resolver configuration.
+
+DHCPv4 configuration now supports `ignoreRoutes` option to ignore routes provided by DHCPv4 servers.
 """
 
 [notes.EPHEMERAL]
--- a/internal/app/machined/pkg/controllers/network/operator/dhcp4.go
+++ b/internal/app/machined/pkg/controllers/network/operator/dhcp4.go
@@ -35,6 +35,7 @@ type DHCP4 struct {
 	routeMetric         uint32
 	clientIdentifier    network.ClientIdentifierSpec
 	skipHostnameRequest bool
+	skipRoutes          bool
 	requestMTU          bool
 
 	lease *nclient4.Lease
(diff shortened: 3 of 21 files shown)


smira (approved):
thank you!

PR #13731, opened 2026-07-08
feat: support booting from MD RAID1 array

### LINKED ISSUE #11626: support for `md` volumes
Minimum:

* ensure that e.g. user volumes can be provisioned on top of MD array (add an integration test, fix issues if found)

Full:

* provide a way to create MD arrays over non-system disks.

User story:

I have many local disks, and I would like to make EPHEMERAL span across all disks as RAID 0 array.

shanduur:
Add `metadata` enum (1.0/1.2, default 1.0) to RAIDArrayConfig and plumb
it through MDArraySpec into `mdadm --create --metadata`. 1.0 keeps the
superblock at the member's end, so the array's partition table stays
firmware-visible and the mirror can be a boot/system disk.

Install-to-MD works via the auto-generated UnattendedInstallConfig: its
disk selector matches the array's md-name-<name> symlink, so Talos
installs onto /dev/mdN built across the primaries.

- add `--primary-disks` flag: N identical primary disks per QEMU VM
- add QEMU_SYSTEM_DISKS env (default 1) + e2e-md-boot CI run
- add md-boot-array test patch (raid1, metadata 1.0, size-gated
  member selector so user/extra disks are not absorbed)
- document boot constraints in release notes: raid1 + metadata 1.0 only

Closes https://github.com/siderolabs/talos/issues/11626


frezbo (approved):

majabojarska (approved):
🚀

Orzelius:
this deserves some points!

PR #13748, opened 2026-07-13
test: apid tls provider

neiltalap:
# Pull Request

## What? (description)

Add unit tests for the apid TLS provider package (`internal/app/apid/pkg/provider`), replacing the empty placeholder test.

Coverage includes:

- `certificateProvider.Update`: valid certs, missing client cert, invalid server/client keys, invalid/empty CAs, client cleared on update, cert rotation
- `TLSConfig.ServerConfig` / `ClientConfig`: mutual TLS vs server-only, client config present vs `nil`

## Why? (reasoning)

This package is responsible for building apid's server and client TLS configs from COSI `secrets.API` resources. It previously had no real unit tests, only an empty placeholder for coverage estimation.

## Acceptance

- [x] go test ./internal/app/apid/pkg/provider/ -count=1 -v

### DIFF
--- a/internal/app/apid/pkg/provider/provider_test.go
+++ b/internal/app/apid/pkg/provider/provider_test.go
@@ -2,27 +2,50 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package provider
+package provider_test
 
 import (
+	"context"
 	stdlibtls "crypto/tls"
 	"testing"
+	"time"
 
+	"github.com/cosi-project/runtime/pkg/resource"
+	"github.com/cosi-project/runtime/pkg/state"
+	"github.com/cosi-project/runtime/pkg/state/impl/inmem"
+	"github.com/cosi-project/runtime/pkg/state/impl/namespaced"
 	"github.com/siderolabs/crypto/x509"
 	"github.com/stretchr/testify/suite"
 
+	"github.com/siderolabs/talos/internal/app/apid/pkg/provider"
 	"github.com/siderolabs/talos/pkg/machinery/resources/secrets"
 )
 
-type CertificateProviderSuite struct {
+type TLSConfigSuite struct {
 	suite.Suite
+
+	ctx       context.Context //nolint:containedctx
+	ctxCancel context.CancelFunc
+
+	resources state.State
 }
 
-func TestCertificateProviderSuite(t *testing.T) {
-	suite.Run(t, new(CertificateProviderSuite))
+func TestTLSConfigSuite(t *testing.T) {
+	suite.Run(t, new(TLSConfigSuite))
 }
 
-func (suite *CertificateProviderSuite) newAPICerts(withClient bool) (*secrets.API, []byte) {
+func (suite *TLSConfigSuite) SetupTest() {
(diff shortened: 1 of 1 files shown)


internal/app/apid/pkg/provider/provider_test.go
```diff
@@ -2,13 +2,229 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-package provider_test
+package provider
```
smira: sorry, but our policy is to never test private parts of the package, the test should test only public surface, so the package name should be `provider_test`

neiltalap: Got it. What is the philosophy behind it? I might start using it myself if it is well-formed and justified.

Will change the test to only test exported functions/public surfaces today or tomorrow.

smira: The philosophy is that it doesn't make any sense to test something which is not being called from the outside. There might be more restructuring required to get unit-tests going properly, but going via private methods is not the way to go.

neiltalap: > The philosophy is that it doesn't make any sense to test something that is not being called from the outside. There might be more restructuring required to get unit tests going properly, but going via private methods is not the way to go.

More TDD style, an orthodox approach to tests.

A pragmatic approach is to test the internal functions too if they are complex enough, but orthodox philosophy solves it by stating to restructure that internal function to be simpler so it doesn't need testing.

Very interesting.

Pushed an update, you can check on it whenever you have time.

internal/app/apid/pkg/provider/provider_test.go
```diff
@@ -4,11 +4,205 @@
...
+
+	cfg, err := provider.NewTLSConfig(suite.ctx, suite.resources, skipClientCertVerify)
+	suite.Require().NoError(err)
+
+	return cfg
+}
+
+// updateAPI applies mutate to the current secrets.API resource in state.
+func (suite *TLSConfigSuite) updateAPI(mutate func(*secrets.API)) {
+	r, err := suite.resources.Get(suite.ctx, resource.NewMetadata(secrets.NamespaceName, secrets.APIType, secrets.APIID, resource.VersionUndefined))
```
smira: use `safe.StateGet` helpers, much easier

internal/app/apid/pkg/provider/provider_test.go
```diff
@@ -4,11 +4,205 @@
...
+	suite.Require().NoError(err)
+
+	api := r.(*secrets.API).DeepCopy().(*secrets.API) //nolint:forcetypeassert
```
smira: no need to deepcopy, Get returns a copy

internal/app/apid/pkg/provider/provider_test.go
neiltalap: I used StateGetResource, turned out to be shorter, and removed some redundant checks too.

I must note that COSI is very unintuitive to navigate. Tracing function calls keeps running into interfaces or whatnot instead of the actual implementations.

neiltalap:
@smira Hi, I pushed the changes some time ago. Forgot to tag you, so I am tagging you now. I am not exactly sure what the convention for tagging and all is for this repository. And what priority external PRs take.

smira (approved):
thank you, I did some test improvements on top of your changes

Issue #13475, opened 2026-05-30
Add support for ipv6-only clusters on scaleway


miran248:
### Description

If you try to create an instance on scaleway without ipv4 connectivity, it will not read the provided cloud-init patches, i assume it's because the metadata endpoint [here](https://github.com/siderolabs/talos/blob/v1.13.3/internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go) is hardcoded, ipv6 endpoint is `fd00:42::42`, see [cloud-init reference](https://docs.cloud-init.io/en/latest/reference/datasources/scaleway.html).

### Logs

These were all over the place, until i added the ipv4 address to the machine (in addition to ipv6), at which point the configuration was applied and cluster was established.

```
[ 1758.816655] [talos] failed looking up "time.cloudflare.com", ignored {"component": "controller-runtime", "controller": "time.SyncController", "error": "lookup time.cloudflare.com on 8.8.8.8:53: dial udp 8.8.8.8:53: connect: network is unreachable"}
[ 1758.981262] [talos] DHCP request/renew failed {"component": "controller-runtime", "controller": "network.OperatorSpecController", "operator": "dhcp4", "error": "unable to receive an offer: got an error while the discovery request: no matching response packet received", "link": "eth0"}
[ 1787.307947] [talos] retrying error: failed to download config from "http://169.254.42.42/user_data/cloud-init": Get "http://169.254.42.42/user_data/cloud-init": dial tcp :552->169.254.42.42:80: connect: network is unreachable
```

### Environment

- Talos version: `v1.13.3`
- Kubernetes version: `v1.36.1`
- Platform: scaleway

Cluster config uses the new apis but is otherwise more or less standard stuff (well, apart from prioritizing ipv6 instead of ipv4).
[scaleway patches](https://github.com/miran248/terraform-talos-modules/blob/851b951b7d670dde294dd815c8384e50de8c314f/modules/scaleway-pool/locals.tf) were taken from your docs, image is standard, uploaded via [script](https://github.com/miran248/terraform-talos-modules/blob/851b951b7d670dde294dd815c8384e50de8c314f/packer/justfile) and then [initialized](https://github.com/miran248/terraform-talos-modules/blob/851b951b7d670dde294dd815c8384e50de8c314f/modules/scaleway-image/main.tf) via terraform.

(should this be marked as a feature request?)

smira:
[...]

PR #13752, opened 2026-07-13
feat(machined): add IPv6-only support for Scaleway platform

miran248:
## What? (description)

Add IPv6-only support for Scaleway platform. Also adds `RoutedIPEnabled` to metadata.
Solves https://github.com/siderolabs/talos/issues/13475

## Why? (reasoning)

Previously the scaleway platform only probed the IPv4 metadata endpoint (169.254.42.42), which is unreachable on IPv6-only instances. This change tries IPv4 with a 15-second timeout, then falls back to the IPv6 metadata endpoint (fd00:42::42).

Also adds support for Scaleway's routed IP mode (`RoutedIPEnabled`), where public IPs are configured statically rather than via DHCP.

## Acceptance

### DIFF
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
@@ -8,26 +8,57 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
+	"time"
 
 	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
 
 	"github.com/siderolabs/talos/pkg/download"
 )
 
+// Metadata extends the SDK's instance.Metadata with fields not yet present in the SDK.
+type Metadata struct {
+	instance.Metadata
+
+	// RoutedIPEnabled is true when the instance is in routed IP mode. In this mode
+	// IPs from PublicIpsV4/V6 are configured statically rather than via DHCP.
+	RoutedIPEnabled bool `json:"routed_ip_enabled"`
+}
+
 const (
-	// ScalewayMetadataEndpoint is the local Scaleway endpoint.
+	// ScalewayMetadataEndpoint is the local Scaleway IPv4 metadata endpoint.
 	ScalewayMetadataEndpoint = "http://169.254.42.42/conf?format=json"
-	// ScalewayUserDataEndpoint is the local Scaleway endpoint for the config.
+	// ScalewayMetadataEndpointIPv6 is the local Scaleway IPv6 metadata endpoint.
+	ScalewayMetadataEndpointIPv6 = "http://[fd00:42::42]/conf?format=json"
+	// ScalewayUserDataEndpoint is the local Scaleway IPv4 endpoint for the config.
 	ScalewayUserDataEndpoint = "http://169.254.42.42/user_data/cloud-init"
+	// ScalewayUserDataEndpointIPv6 is the local Scaleway IPv6 endpoint for the config.
+	ScalewayUserDataEndpointIPv6 = "http://[fd00:42::42]/user_data/cloud-init"
+
+	// metadataIPv4Timeout is how long to wait for the IPv4 metadata endpoint before
+	// falling back to IPv6. Long enough for dual-stack instances where IPv4 needs a
+	// few seconds to come up, short enough not to delay IPv6-only instances too much.
+	metadataIPv4Timeout = 15 * time.Second
 )
 
-func (u *Scaleway) getMetadata(ctx context.Context) (*instance.Metadata, error) {
-	metaConfigDl, err := download.Download(ctx, ScalewayMetadataEndpoint)
+func (u *Scaleway) getMetadata(ctx context.Context) (*Metadata, error) {
+	probeCtx, cancel := context.WithTimeout(ctx, metadataIPv4Timeout)
+	metaConfigDl, err := download.Download(probeCtx, ScalewayMetadataEndpoint,
+		download.WithTimeout(metadataIPv4Timeout))
+
+	cancel()
+
+	if err != nil {
+		log.Printf("IPv4 metadata unreachable (%s), falling back to IPv6", err)
+
+		metaConfigDl, err = download.Download(ctx, ScalewayMetadataEndpointIPv6)
+	}
+
 	if err != nil {
 		return nil, fmt.Errorf("error fetching metadata: %w", err)
 	}
 
-	var meta instance.Metadata
+	var meta Metadata
 	if err = json.Unmarshal(metaConfigDl, &meta); err != nil {
 		return nil, err
 	}
--- a/internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
+++ b/internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
@@ -8,21 +8,55 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/netip"
+	"time"
 
 	"github.com/scaleway/scaleway-sdk-go/api/instance/v1"
+	"github.com/siderolabs/go-retry/retry"
 
 	"github.com/siderolabs/talos/pkg/download"
 )
 
 const (
-	// ScalewayMetadataEndpoint is the local Scaleway endpoint.
+	// ScalewayMetadataEndpoint is the local Scaleway IPv4 metadata endpoint.
 	ScalewayMetadataEndpoint = "http://169.254.42.42/conf?format=json"
-	// ScalewayUserDataEndpoint is the local Scaleway endpoint for the config.
+	// ScalewayMetadataEndpointIPv6 is the local Scaleway IPv6 metadata endpoint.
+	ScalewayMetadataEndpointIPv6 = "http://[fd00:42::42]/conf?format=json"
+	// ScalewayUserDataEndpoint is the local Scaleway IPv4 endpoint for the config.
 	ScalewayUserDataEndpoint = "http://169.254.42.42/user_data/cloud-init"
+	// ScalewayUserDataEndpointIPv6 is the local Scaleway IPv6 endpoint for the config.
+	ScalewayUserDataEndpointIPv6 = "http://[fd00:42::42]/user_data/cloud-init"
+
+	// endpointAttemptTimeout bounds a single attempt against one address family. An instance
+	// that only has one of them fails over to the other after this long, rather than spending
+	// the whole retry budget on an endpoint it can never reach.
+	endpointAttemptTimeout = 5 * time.Second
 )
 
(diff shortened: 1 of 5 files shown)


internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
```diff
@@ -8,26 +8,57 @@ import (
...
 	"github.com/siderolabs/talos/pkg/download"
 )
 
+// Metadata extends the SDK's instance.Metadata with fields not yet present in the SDK.
+type Metadata struct {
+	instance.Metadata
+
+	// RoutedIPEnabled is true when the instance is in routed IP mode. In this mode
+	// IPs from PublicIpsV4/V6 are configured statically rather than via DHCP.
+	RoutedIPEnabled bool `json:"routed_ip_enabled"`
```
smira: let's wait for the official SDK support, this field looks confusing enough

miran248: Will ask them on slack.

miran248:
Just a quick update, that unnecessary `RoutedIPEnabled` was removed yesterday, i've also deployed two clusters, to try out both stacks.

smira:
Some review notes:

## IPv4 loop emits a default route per public IP 

The IPv6 loop guards with defaultV6RouteSet; the IPv4 loop doesn't. With two entries in public_ips_v4 carrying different gateways you get two RouteSpecSpecs.

## Dropping DHCP4 loses resolvers / time servers / MTU 

Previously any instance with a v4 IP got a DHCP4 operator. Now needDHCP4 is only set when a gateway is absent, so routed-IP instances get operators: [] (visible in the expected-v2.yaml diff), and ParseMetadata never populates Resolvers/TimeServers. 

## Halved retry budget 

`retry.Exponential(3*time.Minute, …)` fits ~7–8 attempts; strict alternation spends half on an unreachable family, leaving ~4 real tries for a single-stack instance. Sticking to whichever endpoint last answered would preserve it.

miran248:
Synced the fork (and terraform providers), applied the changes, built new images, created two clusters (both single-stacks) - they were both provisioned successfully.
(also converted more configs to the new api format and they also worked fine, loving the new direction!)

smira (commented):
Review pass over the Scaleway IPv6-only change. Package tests pass; the overall approach (alternating endpoints, static routed-IP config, SLAAC-provided IPv6 before metadata fetch) looks sound. Six comments inline — the two I'd consider blocking are the `retry.ExpectedError(ErrNoConfigSource)` wrap (3 minute stall on every boot without user data) and the missing IPv6 resolver for IPv6-only instances (default resolvers are IPv4-only, so DNS is dead on exactly the instances this PR enables).

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/scaleway.go
```diff
@@ -185,12 +219,13 @@ func (s *Scaleway) Configuration(ctx context.Context, r state.State) ([]byte, er
...
 
-	log.Printf("fetching machine config from %q", ScalewayUserDataEndpoint)
+	log.Printf("fetching machine config from %q or %q", ScalewayUserDataEndpoint, ScalewayUserDataEndpointIPv6)
 
-	return download.Download(ctx, ScalewayUserDataEndpoint,
+	return downloadAlternating(ctx, ScalewayUserDataEndpoint, ScalewayUserDataEndpointIPv6,
 		download.WithLowSrcPort(),
-		download.WithErrorOnNotFound(errors.ErrNoConfigSource),
-		download.WithErrorOnEmptyResponse(errors.ErrNoConfigSource))
+		download.WithErrorOnNotFound(retry.ExpectedError(errors.ErrNoConfigSource)),
```
smira, line 226: Wrapping `ErrNoConfigSource` in `retry.ExpectedError` turns "no user data" into a *retryable* error: `retry.retry` finds the `expectedError` via `errors.As` even through the `fmt.Errorf` wrap that `download.Download` adds, so a 404 (and the 204/empty case on the next line) no longer aborts the download — it keeps retrying until the default `download` budget of 3 minutes expires.

Concretely: the common Scaleway flow of booting an instance with no user data and applying config in maintenance mode now stalls ~3 minutes in `AcquireController.loadFromPlatform` before `errors.Is(err, ErrNoConfigSource)` finally matches (via `ErrorSet.Is`) and the node proceeds to maintenance. Every other platform passes the bare `ErrNoConfigSource` here.

If the goal is only "keep trying the other address family before giving up", bound it with `download.WithTimeout(...)` instead of inheriting the 3 minute default.

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/scaleway.go
```diff
@@ -54,32 +89,55 @@ func (s *Scaleway) ParseMetadata(metadata *instance.Metadata) (*runtime.Platform
...
+		// /32 routed IPs require a host route to the gateway before the default route.
+		networkConfig.Routes = append(networkConfig.Routes,
+			staticRoute(nethelpers.FamilyInet4, netip.PrefixFrom(gw, gw.BitLen()), netip.Addr{}, 3*network.DefaultRouteMetric),
+			staticRoute(nethelpers.FamilyInet4, netip.Prefix{}, gw, 2*network.DefaultRouteMetric),
+		)
+
+		defaultV4RouteSet = true
+	}
 
 	if len(metadata.PublicIpsV4) > 0 {
```
smira, line 142: The DHCP4 operator is the only operator emitted, and it is gated on `len(metadata.PublicIpsV4) > 0`. On the IPv6-only instance this PR targets, no operator is created at all, so the platform contributes no resolvers, no time servers and no MTU.

`ResolverConfigController.getDefault` then falls back to `constants.DefaultPrimaryResolver` / `DefaultSecondaryResolver` = `1.1.1.1` / `8.8.8.8` — both unreachable from an IPv6-only node. DNS is dead, so image pulls and `pool.ntp.org` fail, i.e. the node comes up with connectivity but cannot join.

Compare `aws.go` (`hasIPv6` branch), which adds an explicit IPv6 resolver spec (plus DHCP6) precisely for this case; Scaleway needs the equivalent.

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/metadata.go
```diff
@@ -8,21 +8,55 @@ import (
...
+	options = append(options,
+		download.WithEndpointFunc(func(context.Context) (string, error) {
+			endpoint := endpoints[attempt%len(endpoints)]
+			attempt++
+
+			return endpoint, nil
+		}),
+		download.WithRetryOptions(retry.WithAttemptTimeout(endpointAttemptTimeout)),
+	)
+
```
smira, line 54: `ipv4Endpoint` is passed as the `endpoint` argument while `WithEndpointFunc` alternates families. `download.Download` uses that argument only for the error text — `fmt.Errorf("failed to download config from %q: %w", endpoint, err)` — so every retry log line and the final error names `http://169.254.42.42/...` even when the failing attempt actually went to `[fd00:42::42]`. Debugging an IPv6-only boot shows exclusively the IPv4 URL.

It also interacts badly with `retry.ErrorSet.Append`, which de-duplicates by message string: distinct v4 and v6 failures that render identically collapse into a single entry.

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/scaleway.go
```diff
@@ -54,32 +89,55 @@ func (s *Scaleway) ParseMetadata(metadata *instance.Metadata) (*runtime.Platform
...
-		Type:        nethelpers.TypeUnicast,
-		Family:      nethelpers.FamilyInet4,
-		Priority:    4 * network.DefaultRouteMetric,
+	networkConfig.Routes = []network.RouteSpecSpec{
+		staticRoute(nethelpers.FamilyInet4, metadataRoute, netip.Addr{}, 4*network.DefaultRouteMetric),
 	}
 
-	route.Normalize()
-	networkConfig.Routes = []network.RouteSpecSpec{route}
+	// Instances always run in routed IP mode: public IPs are attached to the instance as
```
smira, line 102: Nothing actually reads `routed_ip_enabled` — the SDK's `instance.Metadata` has no such field, so the `"routed_ip_enabled": true` line added to `testdata/metadata-v4.json` is silently discarded by `json.Unmarshal` and covers nothing (the PR description says `RoutedIPEnabled` was added to metadata, but it isn't).

The only discriminator left is `v4.Gateway == ""`. If any legacy NAT-mode instance does advertise a gateway, Talos statically assigns the *public* IP to eth0 and sets `SkipRoutes: true`, dropping the DHCP-supplied default route in favour of a default via a gateway that is not on-link — the node loses all connectivity. Worth keying off `provisioning_mode`, or actually parsing the flag, rather than assuming.

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/scaleway.go
```diff
@@ -54,32 +89,55 @@ func (s *Scaleway) ParseMetadata(metadata *instance.Metadata) (*runtime.Platform
...
+	// host routes and configured statically. The `routed_ip_enabled` flag is deprecated in
+	// the API and always true, so there is no legacy NAT mode left to detect.
+	// See https://github.com/scaleway/scaleway-sdk-go/issues/3247.
+	var defaultV4RouteSet bool
+
+	for _, v4 := range metadata.PublicIpsV4 {
```
smira, line 108: `metadata.PublicIP` (the legacy IPv4-only field, still populated in fixtures v1-v3) is no longer consulted for `ExternalIPs`; only `PublicIpsV4` is.

On an instance whose metadata carries `public_ip` but an empty/absent `public_ips_v4`, the public IPv4 now disappears from `ExternalIPs`, so it is no longer added as a node address nor to the API server certificate SANs and `talosctl` over the public IP breaks. That path used to be covered by the `metadata.PublicIP.Family == "inet"` branch that was removed.

internal/app/machined/pkg/runtime/v1alpha1/platform/scaleway/scaleway.go
```diff
@@ -54,32 +89,55 @@ func (s *Scaleway) ParseMetadata(metadata *instance.Metadata) (*runtime.Platform
...
+		addr, err := address.IPPrefixFrom(v4.Address, v4.Netmask)
```
smira, line 109: These errors are now returned from inside a loop over *all* `public_ips_v4` entries (same for `netip.ParseAddr(v4.Gateway)` below, and for the v6 loop). A single malformed or empty entry — e.g. a detached IP with `"address": ""` — makes `ParseMetadata` fail outright, so `NetworkConfiguration` returns an error and the node gets no platform network config at all: no links, no hostname, no platform metadata. Previously an empty address was simply skipped.

Skipping unparseable entries with a log line would degrade more gracefully.

PR #13756, opened 2026-07-14
fix: triggered e2e flakes

frezbo:
More fixes for flaky tests.

### DIFF
--- a/internal/integration/api/rotate.go
+++ b/internal/integration/api/rotate.go
@@ -53,6 +53,14 @@ func (suite *RotateCASuite) TearDownTest() {
 	}
 }
 
+func (suite *RotateCASuite) newRotationContext() context.Context {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+
+	suite.T().Cleanup(cancel)
+
+	return ctx
+}
+
 // TestTalos updates Talos CA in the cluster.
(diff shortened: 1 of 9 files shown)


internal/integration/api/rotate.go
```diff
@@ -59,6 +59,10 @@ func (suite *RotateCASuite) TestTalos() {
 		suite.T().Skip("cluster information is not available")
 	}
 
+	if testing.Short() {
```
smira: this is a fast test, should we skip it?

smira (approved):

internal/integration/api/rotate.go
frezbo: this is fast, but then other tests wait on cluster being back to healthy which takes a bit 

> • RotateCASuite.SetupTest sets one shared 10-minute context.

  TestTalos does two CA rotations, then calls:

  suite.AssertClusterHealthy(suite.ctx)

  In that failure, repeated config application left one kubelet Finished. Cluster health kept polling until suite.ctx expired, exactly 10 minutes after the test began.

  The rotation itself was done. The useless wait was health-checking a kubelet that was never going to restart.


shouldn't this be fixed now?

frezbo: this was wrong hallucination

PR #13761, opened 2026-07-14
chore: (1.12) bump tools and pkgs

majabojarska:


majabojarska:
Needs https://github.com/siderolabs/tools/pull/558 first

tools/structprotogen/go.mod
frezbo, line 1: we don't really need to update tools for a backport

frezbo (approved):

tools/structprotogen/go.mod
majabojarska: Ah good to know. I can keep it as-is right, won't break anything?

PR #13774, opened 2026-07-15
fix: enable `noexec` for EPHEMERAL only for new machines

### LINKED ISSUE #13720: make the noexec for EPHEMERAL opt-in (no-op for upgrades)
Follow-up on https://github.com/siderolabs/talos/pull/13211

There is at least one more (besides Longhorn v1) which might be broken by noexec var:  https://www.vcluster.com/docs/vcluster/troubleshoot/noexec-emptydir-volumes

So the proposal is to flip the defaults:

* without any explicit machine config, noexec defaults to false for EPHEMERAL
* instead, for 1.14+ machine configs, we would inject VolumeConfig document with `noexec: true`

Result:

* users upgrading to 1.14 will not be broken
* new clusters 1.14 if using Longhorn/vcluster might need an adjustment of the base machine config

frezbo:
Enable `noexec` for EPHEMERAL for newly created machines only.

Fixes: #13720

### DIFF
--- a/hack/release.toml
+++ b/hack/release.toml
@@ -141,12 +141,16 @@ DHCPv4 configuration now supports `ignoreRoutes` option to ignore routes provide
 [notes.EPHEMERAL]
         title = "noexec on EPHEMERAL (/var)"
         description = """\
-The EPHEMERAL volume (`/var`) is now mounted with `noexec` in addition to the existing `nosuid` and `nodev`,
-blocking binary execution from `/var`.
+Talos 1.14 clusters now default the EPHEMERAL volume (`/var`) to `noexec` in addition to the existing `nosuid` and `nodev`
+mount options through generated machine configuration.
 
-Workloads that exec binaries placed under `/var` will break.
-For example, Longhorn v1's `instance-manager` exec's engine binaries the `engine-image` DaemonSet drops under `/var/lib/longhorn/engine-binaries/`,
-which now fails with `permission denied`. Affected users can opt out via a `VolumeConfig` document:
+Existing machines are not affected on upgrades.
+
+Note: Workloads that execute binaries placed under `/var` can break on new machines.
+Longhorn v1 and [vCluster](https://www.vcluster.com/docs/vcluster/troubleshoot/noexec-emptydir-volumes) are known to be affected.
+For example, Longhorn v1's `instance-manager` executes engine binaries that the `engine-image` DaemonSet places under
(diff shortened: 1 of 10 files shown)


hack/release.toml
```diff
@@ -142,9 +142,11 @@ DHCPv4 configuration now supports `ignoreRoutes` option to ignore routes provide
         title = "noexec on EPHEMERAL (/var)"
         description = """\
 The EPHEMERAL volume (`/var`) is now mounted with `noexec` in addition to the existing `nosuid` and `nodev`,
-blocking binary execution from `/var`.
+blocking binary execution from `/var` for newly generated machine configurations.
```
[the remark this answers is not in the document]
smira: I would phrase this around Talos 1.14 cluster default to this (via machine config)

hack/release.toml
```diff
@@ -142,9 +142,11 @@ DHCPv4 configuration now supports `ignoreRoutes` option to ignore routes provide
...
 
-Workloads that exec binaries placed under `/var` will break.
+Existing machines are not affected on upgrades.
+
+Note: Workloads that exec binaries placed under `/var` will break on new machines.
 For example, Longhorn v1's `instance-manager` exec's engine binaries the `engine-image` DaemonSet drops under `/var/lib/longhorn/engine-binaries/`,
 which now fails with `permission denied`. Affected users can opt out via a `VolumeConfig` document:
```
smira: we can add vcluster here (see the ticket)

pkg/machinery/config/config/volume.go
```diff
@@ -110,7 +110,7 @@ func (emptyVolumeMountConfig) DisableAccessTime() bool {
 }
 
 func (emptyVolumeMountConfig) Secure() bool {
-	return true
+	return false
```
smira: as Maja adds more volume types, I think we should gate this under `EPHEMERAL` name, otherwise default to true

pkg/machinery/config/generate/block.go
```diff
@@ -15,8 +15,14 @@ func (in *Input) generateBlockConfigs() []config.Document {
 		return nil
 	}
 
+	secure := true
+
+	ephemeralConfig := block.NewVolumeConfigV1Alpha1()
+	ephemeralConfig.MetaName = constants.EphemeralPartitionLabel
+	ephemeralConfig.MountSpec.MountSecure = &secure
```
smira:
```suggestion
	ephemeralConfig.MountSpec.MountSecure = new(true)
```

smira (approved):

PR #13781, opened 2026-07-16
feat: refactor Kubernetes configs into `KubeNodeConfig`

### LINKED ISSUE #13715: refactor configs into `KubeNodeConfig`
This config document is for all machine types.

This refactors parts of `.machine.kubelet`, and `.machine.labels`, `.machine.annotations` and `.machine.taints`, and also `.cluster.allowSchedulingOnControlplanes`:

```yaml
apiVersion: v1alpha1
kind: KubeNodeConfig
skipNodeRegistration: false # default, if set, all other settings don't make sense
registerWithFQDN: fase # default (.machine.kubelet.registerWithFQDN
nodeIP: # mirrors exactly .machine.kubelet.nodeIP
   - 10.0.0.0/24
   - !10.3.0.0/24
labels: # .machine.nodeLabels
   foo: bar
annotations:  # .machine.nodeAnnotations
   foo: bar
taints: # .machine.nodeTaints
   foo: bar
```

The setting `.cluster.allowSchedulingOnControlplanes` gets moved into default `NoSchedule` taint on controlplanes.

We should update the kubelet to pass the initial taints/labels via kubelet's config as `registerWith....` stuff - this way it becomes explicit that the taint was there. 

When migrating from v1alpha1, we can make `.cluster.allowSchedulingOnControlplanes` to report a taint

smira:
List of changes:

* `.machine.nodeLabels`, `.machineNodeAnnotations`, `.machine.nodeTaints` moved into `KubeNodeConfig`
* now the controlplane label and no schedule taint are explicit
* removed `.cluster.allowSchedulingOnControlplanes` (remove the taint as needed)
* `.machine.kubelet.nodeIP` + `skipNodeRegistration` + `registerWithFQDN` moved into `KubeNodeConfig`

Fixes #13715


shanduur (approved):

internal/app/machined/pkg/controllers/k8s/node_label_spec.go
```diff
@@ -77,12 +77,8 @@ func (ctrl *NodeLabelSpecController) Run(ctx context.Context, r controller.Runti
 
 		nodeLabels := map[string]string{}
 
-		if cfg != nil && cfg.Config().Machine() != nil {
-			maps.Copy(nodeLabels, cfg.Config().Machine().NodeLabels())
-
-			if cfg.Config().Machine().Type().IsControlPlane() {
-				nodeLabels[constants.LabelNodeRoleControlPlane] = ""
-			}
+		if cfg != nil && cfg.Config().K8sNodeConfig() != nil {
```
frezbo, line 80: we don;t need to set the controlplane label here?

frezbo (approved):

internal/app/machined/pkg/controllers/k8s/node_label_spec.go
smira: nope, as it either comes via shim in the v1alpha1 legacy path, or it explicitly goes via machine config in the new path

PR #13786, opened 2026-07-17
fix: zero MD superblock via block wipe on destroy

shanduur:
mdadm --zero-superblock only clears a superblock it can still parse
on the exact device given, so it is a silent no-op on a stale or
corrupt superblock, or on a member detached from the array. Such a
leftover gets auto-assembled into a phantom /dev/md0 that holds the
member disk busy, so the next array built from it comes up degraded
(1 of 2 mirrors), goes inactive, and never recovers.

Wipe the on-disk superblock regions directly through the block wipe
primitive (WipeWithSignatures: secure discard + signature wipe +
first/last 1MiB), reaching every metadata location (1.1 start, 1.2
4KiB, 1.0/0.90 end) regardless of whether mdadm can parse it. Destroy
and Shrink route through ZeroSuperblock, so both inherit the fix.

### DIFF
--- a/internal/pkg/md/array.go
+++ b/internal/pkg/md/array.go
@@ -6,11 +6,17 @@ package md
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
+
+	"github.com/siderolabs/go-blockdevice/v2/block"
+
+	"github.com/siderolabs/talos/internal/pkg/partition"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


internal/pkg/md/array.go
```diff
@@ -323,14 +329,52 @@ func (md *MD) Stop(ctx context.Context, device string) error {
...
+	}
+
+	return errs
+}
+
+// wipeSuperblock zeroes the MD superblock on a single member device using the
+// standard Talos block wipe primitive (secure discard + signature wipe + first
+// and last 1MiB fast wipe), which reaches every possible superblock location.
+func wipeSuperblock(ctx context.Context, member string) error {
+	bd, err := block.NewFromPath(member, block.OpenForWrite())
```
frezbo, line 364: shouldn't we create a lock when opening for wiping?

frezbo: nevermind, I see below, i thought it was an NewFromPath options

internal/pkg/md/array.go
```diff
@@ -323,14 +329,52 @@ func (md *MD) Stop(ctx context.Context, device string) error {
...
+
+	defer bd.Close() //nolint:errcheck
+
+	if err := bd.RetryLockWithTimeout(ctx, true, time.Minute); err != nil {
+		return fmt.Errorf("lock block device: %w", err)
+	}
+
+	defer bd.Unlock() //nolint:errcheck
+
+	return partition.WipeWithSignatures(bd, member, nil)
```
frezbo, line 377: let's add a focused test for this to verify all signatures are wiped

frezbo (approved):

internal/pkg/md/array.go
smira: we could also wipe twice if we want to get rid of all data - as the first wipe would destroy only the MD data, the second one should see the underlying structure (if any).

PR #13796, opened 2026-07-20
feat: support system volume wiping

### LINKED ISSUE #10679: Volume wipe without reboot

majabojarska:
Resolves https://github.com/siderolabs/talos/issues/10679

Adds support for wiping a specific Talos system volume by id (e.g. `EPHEMERAL`, `KUBELET`, ...). For now, we will support wiping only partitions and disks. Support for wiping directory-backed system volumes will come at a later point.


internal/app/machined/internal/server/v1alpha1/v1alpha1_server.go
```diff
@@ -725,6 +725,142 @@ func (s *Server) Reset(ctx context.Context, in *machine.ResetRequest) (reply *ma
...
+		if volumeStatus.TypedSpec().Type != block.VolumeTypePartition {
+			return nil, status.Errorf(codes.InvalidArgument, "volume %q is not a partition-backed volume (type: %v)", id, volumeStatus.TypedSpec().Type)
+		}
+
+		result = append(result, volumeStatus)
+	}
+
+	return result, nil
+}
+
```
majabojarska: note to self: All of these need to be factored out of `v1alpha1_server.go`. `VolumeWipe` can also become lighter by delegating all of the wiping to some util placed in `internal/pkg/partition/wipe.go` or similar. The endpoint handler should mostly just do param validation and authz.

PR #13797, opened 2026-07-20
fix: identify etcd members by ID instead of hostname in health check

faymard:
## What? (description)

The `EtcdConsistentAssertion` health check (`pkg/cluster/check/etcd.go`) now compares etcd members across nodes using their etcd member ID (stable) instead of their hostname (etcd's `Name` field).

## Why? (reasoning)

An etcd member's hostname can be temporarily empty right after a reboot, while the node re-learns peer names via a raft proposal, even though the etcd cluster membership is otherwise perfectly consistent and healthy. Since the check compared members using a tuple that included the hostname, this transient empty field caused the check to fail even though the etcd cluster was fully functional. The etcd member ID, on the other hand, is stable from member creation and does not disappear on reboot, making it a reliable identifier for this comparison.

The scenario where etcd members loose their names but keep their IDs has been seen during an upgrade from 1.8 to 1.9, then repeated on every step to 1.12

## Acceptance

### DIFF
--- a/pkg/cluster/check/etcd.go
+++ b/pkg/cluster/check/etcd.go
@@ -63,35 +63,42 @@ func EtcdConsistentAssertion(ctx context.Context, cl ClusterInfo) error {
 		return cmp.Compare(a.node, b.node)
 	})
 
+	// members are identified by their etcd member ID, which is stable across reboots;
+	// the hostname (member name) is not used for identification, as a member can
+	// temporarily report an empty name right after a reboot, before it re-learns
+	// peer names via a raft proposal, even though cluster membership is otherwise consistent.
 	type data struct {
-		hostname  string
-		id        uint64
 		isLearner bool
 	}
 
-	knownMembers := map[data]struct{}{}
+	knownMembers := map[uint64]data{}
 
 	for i, message := range memberResponses {
 		if i == 0 {
 			// Fill data using first message
 			for _, member := range message.Members {
-				knownMembers[data{member.Hostname, member.Id, member.IsLearner}] = struct{}{}
+				knownMembers[member.Id] = data{isLearner: member.IsLearner}
 			}
 
 			continue
 		}
 
 		if len(message.Members) != len(knownMembers) {
-			expected := maps.ToSlice(knownMembers, func(k data, v struct{}) string { return k.hostname })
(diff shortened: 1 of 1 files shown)


smira:
It's not clear what is the problem this PR is trying to solve - the checks are expected to fail & recover, and no failure is hard failure. So removing a hostname check doesn't seem to solve anything if that is a transient failure.

faymard:
> It's not clear what is the problem this PR is trying to solve - the checks are expected to fail & recover, and no failure is hard failure. So removing a hostname check doesn't seem to solve anything if that is a transient failure.

I struggled to find relevant issues, also I tried to reproduce the issue earlier today but could not do it. 

The thing that I might have forgotten to describe is that in that particular case, the ETCD cluster is always healthy and fully available on all nodes. It seems to be a metadata issue rather than a "real" issue. What a real issue is or is not is up to our own appreciation I guess.
The fix that this PR offers allows to check the consistency of the memberlist without relying on possibly missing metadata that does not impact the health of the cluster.

I'm planning to do the upgrade from 1.12 to 1.13 in a few days/weeks depending of how fast I end the first phase of rollouts. On that occasion I will make sure to reproduce the issue so I can explain the PR better.

Until then should I put this PR back on draft mode?

smira:
Once again, the problem is not clear if the issue is transient.

The upgrade doesn't touch etcd in any way in modern Talos, except for upgrading etcd version if the default version is used.

faymard:
> The upgrade doesn't touch etcd in any way in modern Talos, except for upgrading etcd version if the default version is used.

This is what happened, during the upgrade from Talos Linux 1.8 to 1.12 the etcd version bumped from `3.5.16` to `3.6.11`.

> the issue is transient

We had to reboot control plane nodes to have complete metadata on `talosctl etcd members` on all nodes and make `talosctl health` return a healthy status. So I would not call this a transient issue to be fair

Maybe this issue resides somewhere else. I'll put this back to draft and see if 1.12->1.13 gives the same scenario

smira:
The upgrade 1.8 -> 1.12 is not supported and should be denied. Either way this change in the best case masks a real issue. 

I'm closing this PR for now.

faymard:
Sorry for the confusion, it was obviously 1.8->1.12 going through each minor version latest patch.

PR #13798, opened 2026-07-20
feat: drop kubernetes flexvolume mounts

frezbo:
Kubernetes FlexVolume has been deprectaed since 1.23 and nothing will be using it anymore.


internal/integration/k8s/tink.go
```diff
@@ -439,27 +439,16 @@ func (suite *TinkSuite) getTinkManifests(namespace, serviceName, ssName, talosIm
...
-			{
-				MountPoint: constants.EphemeralMountPoint,
-				Size:       "6Gi",
-			},
+	for _, overlayMount := range []overlayMountSpec{
+		{
+			MountPoint: constants.StateMountPoint,
+			Size:       "100Mi",
 		},
-		xslices.Map(
```
smira, line 453: is this related?

internal/integration/k8s/tink.go
```diff
@@ -439,27 +439,16 @@ func (suite *TinkSuite) getTinkManifests(namespace, serviceName, ssName, talosIm
 		Size       string
 	}
 
-	for _, overlayMount := range append(
```
smira, line 442: should run qemu full cycle to test this...

smira (approved):

PR #13802, opened 2026-07-20
fix(talosctl): uncordon nodes when a drain fails

lexfrei:
## What? (description)

When `talosctl upgrade` or `talosctl reboot --drain` drains the Kubernetes node before rebooting and the drain fails after cordoning one or more nodes, those nodes are left `SchedulingDisabled`. The early `return` on a drain error fires before the uncordon defer is registered, and `drainNodes` returns a `nil` map on error — discarding even the nodes it successfully cordoned — so nothing uncordons them.

This makes `drainNodes` return the partially-populated map alongside the error, and registers the uncordon defer before the drain error check in both callers, so a failed drain restores the nodes it cordoned instead of stranding them.

## Why? (reasoning)

`drainNodes` records each node name before cordoning it, then cordons and evicts. If cordoning succeeds but the drain then fails (eviction timeout, PDB blocking, apiserver hiccup), the node is already cordoned. The old `return nil, err` threw away that set, and the callers returned before their uncordon defer was in scope, so the operator was left with `SchedulingDisabled` nodes after an already-failed command.

The reboot decision is unchanged: `upgrade` still aborts on a failed drain (the staged image is applied by a later reboot, and whether to reboot is the caller's call). This PR only ensures a failed drain does not leave nodes cordoned.

## Acceptance

- [ ] you linked an issue (if applicable) — N/A
- [ ] you included tests (if applicable) — the drain/upgrade/reboot paths take a live `*global.ClientFactory` + gRPC + k8s clientset with no mock seam, and the package has no unit tests for this flow
- [ ] you ran conformance (`make conformance`) — not run (containerized)
- [x] you formatted your code (`make fmt`) — `gofmt` clean
- [ ] you linted your code (`make lint`) — not run (needs repo container plugin)
- [ ] you generated documentation (`make docs`) — N/A
- [ ] you ran unit-tests (`make unit-tests`) — not run (containerized)

Verified locally: `go build ./cmd/talosctl/...`, `go vet ./cmd/talosctl/cmd/talos/...`, `gofmt`.

One behaviour here goes past the original request: interrupting a drain with Ctrl+C. A drain runs under a context that SIGINT cancels, so without this the uncordon this PR adds would fail before it starts, and an interrupted drain would leave the nodes SchedulingDisabled. The uncordon on that path therefore runs on a context detached from the caller's, with a one minute ceiling. A second Ctrl+C still aborts immediately, because the signal handler restores the default disposition after the first one. If you would rather keep this PR to the plain uncordon fix, I can move that part to a separate PR.

### DIFF
--- a/cmd/talosctl/cmd/talos/drain.go
+++ b/cmd/talosctl/cmd/talos/drain.go
@@ -30,6 +30,12 @@ type nodeUpdate struct {
 // and performs cordon + drain on all of them in parallel.
 //
 // It returns a map of talosIP -> k8sNodeName for use in the uncordon phase.
+//
+// On error the map is still returned (partially populated): each node name is
+// recorded before that node is cordoned, so the map holds every node that may
+// have been cordoned before the failure. A caller that proceeds past a failed
+// drain (e.g. the upgrade path, which has already staged a new image on disk)
+// can then still uncordon those nodes instead of leaving them SchedulingDisabled.
 func drainNodes(ctx context.Context, clientFactory *global.ClientFactory, drainTimeout time.Duration, rep *reporter.Reporter) (map[string]string, error) {
 	// For kubeconfig - build a random endpoint client (to go to the controlplane).
 	c, err := clientFactory.BuildRandomEndpointClient(ctx)
@@ -103,7 +109,10 @@ func drainNodes(ctx context.Context, clientFactory *global.ClientFactory, drainT
 	<-aggregatorDone
 
 	if err != nil {
-		return nil, err
+		// Return the partially-populated map alongside the error: it holds the
+		// nodes that were cordoned before the failure, so a caller that continues
+		// past a failed drain can still uncordon them.
+		return k8sNames, err
 	}
 
 	return k8sNames, nil
--- a/cmd/talosctl/cmd/talos/drain.go
+++ b/cmd/talosctl/cmd/talos/drain.go
@@ -29,7 +29,10 @@ type nodeUpdate struct {
 // drainNodes runs Phase 1: resolves the Kubernetes node name for each Talos node
 // and performs cordon + drain on all of them in parallel.
 //
-// It returns a map of talosIP -> k8sNodeName for use in the uncordon phase.
+// It returns a map of talosIP -> k8sNodeName for use in the uncordon phase. The
+// map is returned on error too: each name is recorded before that node is
+// cordoned, so it holds every node that may have been cordoned before the
+// failure and the caller can restore them.
 func drainNodes(ctx context.Context, clientFactory *global.ClientFactory, drainTimeout time.Duration, rep *reporter.Reporter) (map[string]string, error) {
 	// For kubeconfig - build a random endpoint client (to go to the controlplane).
 	c, err := clientFactory.BuildRandomEndpointClient(ctx)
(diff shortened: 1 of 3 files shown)


smira:
I'm not sure if this is correct behavior - if the drain fails, the decision to reboot or not is dependent on the caller. 

Nothing is lost either way - the upgrade is staged, and a simple reboot would apply it. 

`talosctl` now supports `--no-reboot` flag, so any flow can be implemented from the outside, e.g.:

1. Drain whatever way you prefer
2. upgrade --no-reboot
3. Reboot the way you would like to
4. Un-cordon or do any other actions

lexfrei:
Fair, the reboot-vs-not decision is the caller's, and `--no-reboot` composes it cleanly. I've dropped the reboot-on-drain-failure change.

I've narrowed the PR to an independent issue it turned up. On a failed drain, both `upgrade` and `reboot --drain` left the already-cordoned nodes `SchedulingDisabled`: the early return fired before the uncordon defer was registered, and `drainNodes` returned a nil map on error, so even the nodes it did cordon were never restored. It now returns the partial set and uncordons before aborting, with the reboot behavior unchanged (upgrade still aborts on a failed drain). If you'd rather handle that separately, I can close this.

lexfrei:
Rebased on current main. No conflicts, main's changes here were in `upgradeInternal` and this patch touches `upgradeViaLifecycleService`, so the rebase left the diff as it was.

Then I fixed a problem in my own patch. The uncordon on the abort path still went through `WaitForNodeReady`. That is right after a reboot, but on a failed drain nothing restarted, so there is nothing to wait for. Usually it returns at once, since a drain that fails on a PodDisruptionBudget or an unevictable pod leaves the node Ready. When the node is not Ready though, and a wedged kubelet stalling evictions is one way to get both at once, it polls for the full `--timeout`, 30m by default, and then leaves the node cordoned anyway. It skips the readiness wait on that path now, and runs detached from the caller's context so an interrupted drain still uncordons. That second part is called out in the description, since it goes past the original ask.

This path is not opt-in on upgrade: `--drain` defaults to true there, and to false on `reboot`. So an aborted drain is what an operator meets the first time a PodDisruptionBudget stalls an upgrade.

Otherwise the scope is the narrow one from the earlier review. This only stops a failed drain from leaving nodes SchedulingDisabled. Upgrade still aborts when the drain fails, and there is no skip-the-reboot logic in here.

One limitation worth flagging: that detached context has its own one minute ceiling, and if it expires the nodes stay cordoned, which is the state this PR exists to prevent. The error text does not currently distinguish that from an unreachable control plane, so an operator would go looking at the cluster for a problem that is not there. Happy to make the message say which one it hit, if you want it.

I also left the kubeconfig refetch alone. The abort-path uncordon refetches it over the Talos API instead of reusing the clientset `drainNodes` already built. That is the dominant cost in the path's timeout budget, and one more way for the restore to fail if the control plane is unreachable. Returning the clientset from `drainNodes` would remove both, but it is past what this PR is for.

The failure path this PR fixes has no test. `internal/integration/cli/reboot.go` covers the drain happy path, and nothing covers a drain that fails. I have not added one. There is no precedent to copy in that suite: no test under `internal/integration/cli/` creates Kubernetes objects, so the failure injection would be the first of its kind there. The helpers are within reach though, `base.K8sSuite` sits in the same test binary, and `RebootSuite` embeds only `base.CLISuite` today. If you want that test, tell me what shape you want and I will write it.

No build, test or lint job has ever run on this PR. Every check on it is a `conform/*` commit or license-header check, and the workflow runs sit in `action_required`. If someone can approve a run, that would be the first real signal on this branch. What I ran locally instead, on the current commit: `go build ./cmd/talosctl/...`, `go vet` on the touched packages, `go test -race ./cmd/talosctl/cmd/talos/... ./cmd/talosctl/pkg/talos/...`, and `golangci-lint` with the repo config. All clean, though the race run only exercises tests that already existed in those packages, none of which touch the drain lifecycle. I ran the linter without the `loglinter` and `kubeimportlinter` plugins, since both are `type: module` and need a golangci-lint built through the repo Makefile; neither applies here, one checks logger usage and the other aliasing of versioned Kubernetes imports, and this diff adds no imports and no logging.

One note on the red check: `conform/commit/gpg-identity` builds its keyring from siderolabs org members, so an outside contributor's key is never in it. The signature itself verifies fine locally.

PR #13803, opened 2026-07-20
fix: support persistent volumemount requests for services

frezbo:
When restarting CRI we shouldn't teardown and re-create `/opt`, we need to persist it. Support transient and persistent volume requests.

Service destroy or shutdown removes both.

### DIFF
--- a/internal/app/machined/pkg/controllers/block/volume_config.go
+++ b/internal/app/machined/pkg/controllers/block/volume_config.go
@@ -43,30 +43,43 @@ func (ctrl *VolumeConfigController) Name() string {
 
 // Inputs implements controller.Controller interface.
 func (ctrl *VolumeConfigController) Inputs() []controller.Input {
-	return []controller.Input{
-		{
+	inputs := make([]controller.Input, 0, 4+len(constants.Overlays))
+
+	inputs = append(inputs,
+		controller.Input{
 			Namespace: config.NamespaceName,
 			Type:      config.MachineConfigType,
 			ID:        optional.Some(config.ActiveID),
 			Kind:      controller.InputWeak,
 		},
-		{
+		controller.Input{
 			Namespace: runtime.NamespaceName,
 			Type:      runtime.MetaKeyType,
 			ID:        optional.Some(runtime.MetaKeyTagToID(meta.StateEncryptionConfig)),
 			Kind:      controller.InputWeak,
 		},
-		{
+		controller.Input{
 			Namespace: block.NamespaceName,
 			Type:      block.VolumeMountRequestType,
 			Kind:      controller.InputDestroyReady,
 		},
-		{
+		controller.Input{
 			Namespace: block.NamespaceName,
 			Type:      block.VolumeConfigType,
 			Kind:      controller.InputDestroyReady,
 		},
+	)
+
+	for _, overlay := range constants.Overlays {
+		inputs = append(inputs, controller.Input{
+			Namespace: block.NamespaceName,
+			Type:      block.MountStatusType,
+			ID:        optional.Some(overlay.Path),
+			Kind:      controller.InputWeak,
+		})
 	}
+
+	return inputs
 }
 
 // Outputs implements controller.Controller interface.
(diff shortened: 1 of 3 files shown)


internal/app/machined/pkg/system/persistent_volumes_test.go
```diff
@@ -0,0 +1,205 @@
...
+	services.LoadAndStart(service)
+	require.Eventually(t, func() bool {
+		return serviceStateFor(services, serviceID) == events.StateRunning.String()
+	}, time.Second, 10*time.Millisecond)
+
+	require.NoError(t, services.Stop(t.Context(), serviceID))
+
+	assertVolumeRequest(t, rt, "service/"+serviceID+"-"+transientVolume, false)
+	assertVolumeRequest(t, rt, "service/"+serviceID+"-"+persistentVolume, true)
+
```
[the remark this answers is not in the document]
frezbo: yeh, wrong PR description, i meant service unload

internal/app/machined/pkg/system/services/cri.go
```diff
@@ -123,6 +126,17 @@ func (c *CRI) Volumes(r runtime.Runtime) []string {
 	return volumes
 }
 
+// PersistentVolumes implements the system.PersistentVolumeMountService interface.
+func (c *CRI) PersistentVolumes(r runtime.Runtime) []string {
+	if r.State().Platform().Mode().InContainer() {
+		return nil
+	}
+
+	return xslices.Map(constants.Overlays, func(target constants.SELinuxLabeledPath) string {
```
smira: I don't think this is enough with new system volumes stuff. 

I almost wonder if we should actually move all volume stuff outside of this flow?

smira: there is actually an easier way to pin those volumes to be mounted I think

smira: basically this block does equivalent of the same with way less code:

https://github.com/siderolabs/talos/blob/5bc52c94a9909041390817530089820fe7305fe8/internal/app/machined/pkg/controllers/block/volume_config.go#L95-L109

internal/app/machined/pkg/controllers/block/volume_config.go
```diff
@@ -154,6 +164,73 @@ func (ctrl *VolumeConfigController) Run(ctx context.Context, r controller.Runtim
...
+	return nil
+}
+
+func (ctrl *VolumeConfigController) shouldPinOverlayMountRequests(ctx context.Context, r controller.Reader, cfg configconfig.Config) (bool, error) {
+	if cfg == nil || cfg.Machine() == nil || ctrl.V1Alpha1Mode.InContainer() {
+		return false, nil
+	}
+
+	// Overlay upper and work directories live under /var, so wait until EPHEMERAL is mounted.
+	ephemeralMountStatus, err := safe.ReaderGetByID[*block.VolumeMountStatus](ctx, r, constants.EphemeralPartitionLabel)
```
smira: is this needed? should the mount request itself only succeed when the parent is ready?

internal/app/machined/pkg/controllers/block/volume_config.go
```diff
@@ -154,6 +164,73 @@ func (ctrl *VolumeConfigController) Run(ctx context.Context, r controller.Runtim
...
+
+			return nil
+		}); err != nil {
+			return fmt.Errorf("failed to create overlay mount request %q: %w", overlayPath, err)
+		}
+
+		return nil
+	}
+
+	ready, err := r.Teardown(ctx, mountRequest.Metadata())
```
smira: also not sure why do we need to tear it down? what is the goal here? the request will keep it alive, but it doesn't prevent the volume from being torn down

internal/app/machined/pkg/controllers/block/volume_config.go
frezbo: yes, it only checks on `VolumeStatus` not volumemountstatus, otherwise we try to create overlay before /var is mounted

smira: yes, but I think we should handle this case already correctly without issue? the controllers will figure out that /opt can't be mounted before /var

And same with CRI volume now - it's a separate one

frezbo:
closing this, since the fix is trivial and will be part of cri customizarion

PR #13808, opened 2026-07-21
test: stabilize AWS readiness and Talos 1.13 QEMU config

frezbo:
The AWS arm64 integration script accepted one successful `kubectl version`
probe and then issued another unprotected request. A Classic ELB can route
those requests to different backends, so the second request could still hit
an unhealthy API server. Require three consecutive successful probes and
fail explicitly when the bounded readiness window expires.

Failure:
https://github.com/siderolabs/talos/actions/runs/29802021729/job/88544787679

The Talos 1.13 QEMU variant generates kubelet configuration in the legacy
`.machine.kubelet` field. The current ephemeral NVMe fixture adds a separate
`KubeletConfig`, which made validation report both an empty kubelet image and
conflicting kubelet configuration. Use the pre-1.14 fixture only for the
1.13 variant while leaving current variants unchanged.

Failure:
https://github.com/siderolabs/talos/actions/runs/29802259625/job/88545465808

The AWS script passes shell syntax validation. The generated v1.13 worker
configuration validates in metal mode with the pre-1.14 fixture, which is
identical to the fixture used before the KubeletConfig refactor.


pkg/rotate/pki/kubernetes/kubernetes.go
```diff
@@ -325,6 +333,110 @@ func (r *rotator) dropOldCA(ctx context.Context) error {
...
+	if err = restartKubeProxyPods(ctx, clientset, 3*time.Minute, time.Second); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func waitForRootCA(ctx context.Context, client kubernetes.Interface, expectedCA []byte, timeout, interval time.Duration) error {
+	err := wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
+		configMap, err := client.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(ctx, "kube-root-ca.crt", metav1.GetOptions{})
```
smira: I think this is broken when kube-proxy is missing

pkg/rotate/pki/kubernetes/kubernetes.go
```diff
@@ -325,6 +333,110 @@ func (r *rotator) dropOldCA(ctx context.Context) error {
...
+	})
+	if err != nil {
+		return fmt.Errorf("error waiting for kube-system/kube-root-ca.crt to contain the new CA: %w", err)
+	}
+
+	return nil
+}
+
+func restartKubeProxyPods(ctx context.Context, client kubernetes.Interface, timeout, interval time.Duration) error {
+	pods, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(ctx, metav1.ListOptions{LabelSelector: "k8s-app=kube-proxy"})
```
smira: I feel this is wrong way to attack this problem in the first place... if we want the pod to restart on configmap change, we should do something else, or the pod should itself act on such changes.

smira:
nit: I would like PR titles and descriptions to be more descriptive, this is like the third PR with a title like this one, and it's very hard to track them as they have no meaning.

If the PR is fixing tests, it should be `test:`. 

If the PR is a fix to CA rotation, it should be `fix` and separated out.

If there is a problem we're fixing, let's bring in the problem (log, failure), so it's trackable.

frezbo:
> nit: I would like PR titles and descriptions to be more descriptive, this is like the third PR with a title like this one, and it's very hard to track them as they have no meaning.
> 
> If the PR is fixing tests, it should be `test:`.
> 
> If the PR is a fix to CA rotation, it should be `fix` and separated out.
> 
> If there is a problem we're fixing, let's bring in the problem (log, failure), so it's trackable.

I should let LLM write commits messages and give up on my silly ones 😅

smira (approved):
we should promote to respective pipelines?

PR #13810, opened 2026-07-21
fix: update the vulnerability dates and description

frezbo:
Both of the vulnerability issues is due to `govulncheck` database being wrong or due to it's parsing it wrong.

The gobgp we're using has the fix and the report for openpgp is completely wrong, both code paths are not used at all.


majabojarska (approved):

smira:
nit: that should be a chore, not a fix; it doesn't fix anything from user perspective

PR #13812, opened 2026-07-21
fix: preserve trailing rate-limited trigger events

frezbo:
RateLimitedTrigger used a nonblocking send to an unbuffered channel. An event arriving while the worker forwarded the previous wakeup was dropped. If that event was the final RTM_NEWLINK transition, LinkStatus kept a stale operational state and RequireUp DHCP never started.

The natural race is narrow. Reproduce it deterministically by retaining the unbuffered channel and adding a 250ms sleep immediately after `t.trigger.QueueReconcile()` in the worker. Then run:

  make initramfs PLATFORM=linux/amd64
  _out/talosctl-darwin-arm64 cluster create dev \
    --remote-endpoint=storm.goose-piano.ts.net:50100 \
    --name dhcp-repro --cidr 172.16.0.0/24 \
    --control-plane-port 6444 --controlplanes 3 --workers 2 \
    --with-bootloader=false --skip-injecting-config \
    --skip-unattended-install-config --wait=false

This left one node without a DHCP lease even though the guest reported carrier and the host TAP and veth were both UP,LOWER_UP. The e2e-cilium job hit the same failure on control plane 3 and timed out waiting for the node Talos API:

https://github.com/siderolabs/talos/actions/runs/29801925248/job/88544517041

Before this fix, TestRateLimitedTriggerQueuesTrailingEvent also times out because the mock receives only the first reconciliation. A one-slot channel makes the test pass by retaining one pending dirty event while still coalescing bursts. With the same 250ms QEMU fault injection, all five nodes then receive DHCP leases.

Validated with the regression test for 100 iterations, the trigger package race test, the Linux TestLinkStatusSuite, and the five-node QEMU A/B test.

### DIFF
--- a/internal/app/machined/pkg/controllers/internal/trigger/trigger.go
+++ b/internal/app/machined/pkg/controllers/internal/trigger/trigger.go
@@ -33,7 +33,7 @@ func NewRateLimitedTrigger(ctx context.Context, trigger Trigger, rateLimit rate.
 	t := &RateLimitedTrigger{
 		trigger: trigger,
 		limiter: rate.NewLimiter(rateLimit, burst),
-		ch:      make(chan struct{}),
+		ch:      make(chan struct{}, 1),
 	}
 
 	go t.run(ctx)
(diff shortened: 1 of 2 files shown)


smira (commented):
I like the fix, but the test looks to be totally strange starting from the way it is written down to what it verifies. 

In theory the reproducer should be setting low rate limit, and doing two triggers, and ensuring that two arrive out - the limiter should not suppress the second one.

Basically forcing it to be stuck in `.Wait` for some time

frezbo:
> I like the fix, but the test looks to be totally strange starting from the way it is written down to what it verifies.
> 
> In theory the reproducer should be setting low rate limit, and doing two triggers, and ensuring that two arrive out - the limiter should not suppress the second one.
> 
> Basically forcing it to be stuck in `.Wait` for some time

fixed

smira (approved):

PR #13816, opened 2026-07-21
feat: add veth pair support

### LINKED ISSUE #12859: Feature Request: support defining veth inerface pairs in machine config
## Feature Request

Currently using a daemonset to create veth interfaces along with using `bridge` command to assign vlans when attached to a linux bridge with vlan_filtering enabled.

### Description

Current commands used to configure network:
```
# create veth interface pair
ip link add name ${interface-name} type veth peer name ${interface-name}-port

#  configure vlans on bridge (add all vlans; vlan 1 is untagged and added by default)
bridge vlan add vid 2-4094 dev ${bridge-name} self

# configure vlans for veth interface (assign vlan tags to port [switch] end)
bridge vlan add vid 3000 dev ${interface-name}-port pvid untagged
```

related: #9117

frezbo:
Add a new `VethConfig` machine config document to support veth pairs.

Fixes: #12859


internal/app/machined/pkg/controllers/network/veth_pair_spec.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func vethEndpointLinkSpec(name, peerName string, endpoint network.VethPairEndpointSpec) network.LinkSpecSpec {
+	var multicast *bool
+
+	if endpoint.Multicast != nil {
+		multicast = new(*endpoint.Multicast)
+	}
+
+	return network.LinkSpecSpec{
```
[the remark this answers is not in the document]
frezbo: wrong, new go supports this

internal/app/machined/pkg/controllers/network/veth_config.go
```diff
@@ -0,0 +1,123 @@
...
+
+func vethPairLink(cfg talosconfig.NetworkVethLinkConfig) network.VethPairEndpointSpec {
+	spec := network.VethPairEndpointSpec{
+		Up:  cfg.Up().ValueOr(true),
+		MTU: cfg.MTU().ValueOrZero(),
+	}
+
+	if multicast, ok := cfg.Multicast().Get(); ok {
+		spec.Multicast = new(multicast)
+	}
```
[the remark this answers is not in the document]
frezbo: wrong, new go supports this

internal/app/machined/pkg/controllers/network/link_spec.go
```diff
@@ -195,6 +269,166 @@ func findLink(links []rtnetlink.LinkMessage, name string, allowAliases bool) *rt
...
+		*links, err = conn.Link.List()
+		if err != nil {
+			return fmt.Errorf("error listing links: %w", err)
+		}
+	}
+
+	return multiErr.ErrorOrNil()
+}
+
+func (ctrl *LinkSpecController) cleanupVethPairStates(
```
smira: I think it's crazy that this code is here... I don't think this controller should ever produce any state

pkg/machinery/config/container/container.go
```diff
@@ -95,6 +96,22 @@ func New(documents ...config.Document) (*Container, error) {
 						conflictingDocuments[conflictingID] = documentID
 					}
 				}
+
+				if linkConfig, ok := d.(config.NetworkCommonLinkConfig); ok {
```
smira, line 100: I think this is too specific for a single type, instead we should in the common block handle the fact that Veth basically has two names - one for itself and another for the peer

pkg/machinery/resources/network/link_spec.go
```diff
@@ -99,6 +100,13 @@ type VRFSlave struct {
 	MasterName string `yaml:"masterName,omitempty" protobuf:"1"`
 }
 
+// VethSpec identifies the expected peer of a veth endpoint.
+//
+//gotagsrewrite:gen
+type VethSpec struct {
+	PeerName string `yaml:"peerName,omitempty" protobuf:"1"`
```
smira, line 107: link_status should have a matching change

pkg/machinery/config/container/container.go
frezbo: ah that's easier

smira: we could test for a specific small interface for the document to claim another name I think

internal/app/machined/pkg/controllers/network/link_spec.go
smira: the change here seems to be too complicated compared to what it should be imho

I don't see why veth is so much special - it's basically two LinkSpecs. what is there that needs to be handled very different?

The status of veth pairs should be/can be observed via LinkStatus, as this would cover links which were not created by Talos

smira: Or we could create VethPairStatus out of LinkStatus if we want to

pkg/machinery/resources/network/veth_pair_state.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/cosi-project/runtime/pkg/resource/typed"
+
+	"github.com/siderolabs/talos/pkg/machinery/proto"
+)
+
+// VethPairStateType is the type of VethPairState resource.
+const VethPairStateType = resource.Type("VethPairStates.net.talos.dev")
+
+// VethPairState records a veth pair managed by LinkSpecController.
+type VethPairState = typed.Resource[VethPairStateSpec, VethPairStateExtension]
```
smira:
```suggestion
type VethPairStatus = typed.Resource[VethPairStateSpec, VethPairStateExtension]
```

we use Status everywhere

Not sure what is the ID of the resource? Do we create to resources per pair?

smira (approved):

PR #13817, opened 2026-07-21
feat: add dedicated CRI configuration documents

frezbo:
Add `CRICustomizationConfig` and `CRIBaseRuntimeSpecConfig` documents with automatic CRI regeneration and restart. Deprecate `machine.files` and `machine.baseRuntimeSpecOverrides` while preserving legacy compatibility.


pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -236,12 +241,17 @@ type MachineConfig struct {
 	//
 	//    It can be used to set some default container settings which are not configurable in Kubernetes,
 	//    for example default ulimits.
-	//    Note: this change applies to all newly created containers, and it requires a reboot to take effect.
+	//    Applying, updating, or removing these overrides restarts CRI automatically.
```
smira: docgen:nodoc missing, and it should be trimmed down to exclude examples and all descriptions

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -236,12 +241,17 @@ type MachineConfig struct {
...
+	//    A machine reboot is not required.
+	//
+	//    Deprecated: use the `CRIBaseRuntimeSpecConfig` configuration document instead.
 	//  examples:
 	//    - name: override default open file limit
 	//      value: machineBaseRuntimeSpecOverridesExample()
 	//  schema:
 	//    type: object
+	//
+	// Deprecated: Use the CRIBaseRuntimeSpecConfig configuration document instead.
```
smira, line 229: duplicate?

pkg/machinery/config/types/v1alpha1/v1alpha1_types.go
```diff
@@ -174,9 +174,14 @@ type MachineConfig struct {
 	//     In the case of `overwrite`, and `append`, `path` must be a valid file.
 	//     If an `op` value of `append` is used, the existing file will be appended.
 	//     Note that the file contents are not required to be base64 encoded.
+	//
+	//     Deprecated: use dedicated configuration documents such as `EtcFileConfig` and
```
smira: same here

pkg/machinery/config/types/cri/base_runtime_spec.go
```diff
@@ -0,0 +1,120 @@
...
+//	alias: CRIBaseRuntimeSpecConfig
+//	schemaRoot: true
+//	schemaMeta: v1alpha1/CRIBaseRuntimeSpecConfig
+type CRIBaseRuntimeSpecConfigV1Alpha1 struct {
+	meta.Meta `yaml:",inline"`
+
+	//   description: |
+	//     Overrides for the default OCI runtime specification used by CRI containers.
+	//
+	//     These values are merged with the deprecated
```
smira: let's instead make it conflict with v1alpha1 and remove this crazy statement - there is no reason to have both

pkg/machinery/config/types/cri/customization.go
```diff
@@ -0,0 +1,110 @@
...
+// Content implements config.CRICustomizationConfig interface.
+func (cfg *CRICustomizationConfigV1Alpha1) Content() string {
+	return cfg.CustomizationContent
+}
+
+// Validate implements config.Validator interface.
+func (cfg *CRICustomizationConfigV1Alpha1) Validate(validation.RuntimeMode, ...validation.Option) ([]string, error) {
+	var validationErrors error
+
+	if err := labels.ValidateQualifiedName(cfg.MetaName); err != nil {
```
smira, line 108: we should reject "customization" ?

pkg/machinery/config/config/config.go
```diff
@@ -82,6 +82,8 @@ type Config interface { //nolint:interfacebloat
 	RegistryAuthConfigs() map[string]RegistryAuthConfig
 	RegistryTLSConfigs() map[string]RegistryTLSConfig
 	ImageCacheConfig() ImageCacheConfig
+	CRIBaseRuntimeSpecConfig() CRIBaseRuntimeSpecConfig
```
smira, line 86: we should imho remove old accessors via v1alpha1, make v1alpha1 implement new interfaces(s) like we do in v1alpha1_k8s_bridge.go for K8s configs or similar for network configs

pkg/machinery/config/config/config.go
```diff
@@ -82,6 +82,8 @@ type Config interface { //nolint:interfacebloat
...
+	CRICustomizationConfigs() map[string]CRICustomizationConfig
```
smira: I'm not sure why this is a map vs. a slice?

pkg/machinery/config/types/cri/base_runtime_spec.go
frezbo: I was not sure if we should merge

frezbo:
`.machine.files` old interface is still kept

internal/integration/api/apply-config.go
```diff
@@ -234,14 +235,14 @@ func (suite *ApplyConfigSuite) TestApplyNoOpCRIPatch() {
 
 	// this CRI patch is a no-op, as NRI is already disabled by default, this verifies that CRI config generation handles it correctly.
 	cfgDataOut := suite.PatchV1Alpha1Config(provider, func(cfg *v1alpha1.Config) {
-		cfg.MachineConfig.MachineFiles = xslices.Filter(cfg.MachineConfig.MachineFiles, func(file *v1alpha1.MachineFile) bool {
-			return file.FilePath != "/etc/cri/conf.d/20-customization.part"
+		cfg.MachineConfig.MachineFiles = xslices.Filter(cfg.MachineConfig.MachineFiles, func(file *v1alpha1.MachineFile) bool { //nolint:staticcheck // test deprecated machine files compatibility
```
smira: should we rewrite these tests to follow new documents instead and rely on the fact that cri is now restarted on the fly and things are applied?

e.g. we could wait for a restart of the CRI here and then try immediately to test the effect?

internal/integration/api/common.go
```diff
@@ -252,16 +253,29 @@ func (suite *CommonSuite) TestBaseOCISpec() {
...
+					"soft": 1024,
+				},
+			},
+		},
+	}
+
 	suite.PatchMachineConfig(nodeCtx, &v1alpha1.Config{
 		MachineConfig: &v1alpha1.MachineConfig{
-			MachineBaseRuntimeSpecOverrides: meta.Unstructured{
+			MachineBaseRuntimeSpecOverrides: meta.Unstructured{ //nolint:staticcheck // test deprecated compatibility
```
smira: same question here, why do we test legacy ones here? we should have some unit-test coverage that they conflict?

internal/app/machined/pkg/runtime/v1alpha1/v1alpha1_sequencer_tasks.go
```diff
@@ -555,10 +552,6 @@ func WriteUserFiles(runtime.Sequence, any) (runtime.TaskExecutionFunc, string) {
 
 			// CRI configuration customization
 			if f.Path() == filepath.Join("/etc", constants.CRICustomizationConfigPart) {
-				if err = injectCRIConfigPatch(ctx, r.State().V1Alpha2().Resources(), []byte(f.Content())); err != nil {
```
smira, line 558: I would add a small comment here saying it's going via the machine config cri_bridge (just for anyone who would ask why it looks so strange)

internal/app/machined/pkg/system/services/cri.go
```diff
@@ -57,18 +59,29 @@ func (c *CRI) Client() (*containerd.Client, error) {
...
 
 // PreFunc implements the Service interface.
-func (c *CRI) PreFunc(context.Context, runtime.Runtime) error {
-	return os.MkdirAll(defaults.DefaultRootDir, os.ModeDir)
+func (c *CRI) PreFunc(ctx context.Context, r runtime.Runtime) error {
+	if err := os.MkdirAll(defaults.DefaultRootDir, 0o750); err != nil {
+		return err
+	}
+
+	if r.State().Platform().Mode().InContainer() {
```
smira, line 73: leftover code? this seems to be doing nothing

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod_test.go
```diff
@@ -59,6 +60,51 @@ func (suite *ControlPlaneStaticPodSuite) TestReconcileDefaults() {
...
+		suite.Ctx(),
+		suite.T(),
+		suite.State(),
+		staticPodIDs,
+		func(*k8s.StaticPod, *assert.Assertions) {},
+	)
+
+	ctest.UpdateWithConflicts(suite, v1alpha1.NewService("etcd"), func(service *v1alpha1.Service) error {
+		service.TypedSpec().Healthy = false
+		service.TypedSpec().Unknown = true
```
smira: this is strange state, and I'm not sure how this test relates to the PR?

hack/release.toml
```diff
@@ -67,6 +66,43 @@ machine:
...
+        title = "CRI Customization Configuration"
+        description = """\
+Talos now supports customizing the CRI containerd configuration with named `CRICustomizationConfig`
+documents. Each document contains a TOML fragment; fragments are merged in lexicographical order by name.
+Applying, updating, or removing these documents updates the generated CRI configuration and restarts CRI
+automatically.
+
+The legacy `/etc/cri/conf.d/20-customization.part` machine-file configuration remains supported during the
+deprecation period and is exposed under the reserved name `customization`. A `CRICustomizationConfig` document
+cannot use that name.
```
smira, line 109: I would add a small note like below that reboot is not required

internal/app/machined/pkg/controllers/cri/service.go
```diff
@@ -0,0 +1,175 @@
...
+func (ctrl *ServiceController) Name() string {
+	return "cri.ServiceController"
+}
+
+// Inputs implements controller.Controller interface.
+func (ctrl *ServiceController) Inputs() []controller.Input {
+	return []controller.Input{
+		{
+			Namespace: files.NamespaceName,
+			Type:      files.EtcFileStatusType,
```
smira, line 41: I _believe_ this doesn't work as it should, it's duplicate input, we should just drop the `ID` and leave only one entry

pkg/machinery/config/config/cri.go
```diff
@@ -4,9 +4,23 @@
...
-)
+import "github.com/siderolabs/crypto/x509"
+
+// LegacyCRICustomizationConfigName is the name assigned to the legacy
+// /etc/cri/conf.d/20-customization.part machine file.
+const LegacyCRICustomizationConfigName = "customization"
+
+// CRIBaseRuntimeSpecConfig configures the base OCI runtime specification for CRI containers.
+type CRIBaseRuntimeSpecConfig interface {
+	Overrides() map[string]any
```
smira, line 16: I would suggest to add here unique "marker" methods, like

`CRIBaseRuntimeSpecConfigSignal()` & `CRICustomizationConfigSignal()` to avoid accidentally matching any other interface/document, as the names Content/Overrides are really generic

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod_test.go
smira: ok, I got it... but I think Unknown should be false

smira (commented):
🆒  left some minor comments

internal/app/machined/pkg/controllers/k8s/control_plane_static_pod_test.go
frezbo: yeh, this brought in an issue that static pod majifests were deleted

internal/app/machined/pkg/controllers/cri/service.go
frezbo: good one

internal/app/machined/pkg/system/services/cri.go
frezbo: yup, leftover

internal/integration/api/common.go
```diff
@@ -252,32 +251,20 @@ func (suite *CommonSuite) TestBaseOCISpec() {
...
-	suite.AssertRebooted(
-		suite.ctx, node, func(nodeCtx context.Context) error {
-			return base.IgnoreGRPCUnavailable(suite.Client.Reboot(nodeCtx))
-		}, assertRebootedRebootTimeout,
-		suite.CleanupFailedPods,
-	)
+	}
 
-	suite.ClearConnectionRefused(suite.ctx, node)
+	suite.PatchMachineConfig(nodeCtx, ociRuntimeOverride)
```
smira, line 267: should we wait for CRI restart here?

frezbo: yeh make sense, I can pull in the same assertions from apply-config.go

smira (approved):

